PR #2, opened 2018-09-03
pkg/render: add basic template & parsing support

squeed:
Add a simple rendering engine that scrapes directories for manifest files, does a glolang template substitution on them, then renders them to `Unstructured` k8s objects.


danwinship:
Is this based on something from openshift-ansible?

squeed:
Mostly lifted from kubectl.

squeed:
That's pretty heavy-weight, mostly because we don't need to use codecs. It's not a lot of code for the actual parsing bit.

Still might need a more ergonomic interface for passing template vars, but this will do for now.

PR #6, opened 2018-09-21
Add render-to-disk binary

squeed:
This adds a separate mode that works entirely offline - it consumes configuration from disk, and writes a single yaml file with all desired objects.

### DIFF
--- a/pkg/operator/handler.go
+++ b/pkg/operator/handler.go
@@ -11,7 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func MakeHandler(manifestDir string) sdk.Handler {
+func MakeHandler(manifestDir string) *Handler {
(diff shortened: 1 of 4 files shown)


pkg/operator/handler.go
```diff
@@ -23,6 +23,12 @@ type Handler struct {
 	ManifestDir string
 }
 
+func (h *Handler) SetConfig(conf *v1.NetworkConfig) {
+	h.syncLock.Lock()
+	h.config = conf
+	h.syncLock.Unlock()
```
danwinship: preferred style would be

    h.syncLock.Lock()
    defer h.syncLock.Unlock()
    h.config = conf

? Sure, it doesn't matter yet, but it might given future changes to the function

danwinship:
lgtm I think

PR #8, opened 2018-09-26
Openshift-sdn support

squeed:
This renders and installs all the manifests needed for openshift-sdn support.


manifests/network/openshift-sdn/001-rbac.yaml
```diff
@@ -0,0 +1,89 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: sdn
```
danwinship: since this is non-namespaced should it have a more distinctive name?

manifests/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -0,0 +1,111 @@
...
+      hostPID: true
+      priorityClassName: system-node-critical
+      containers:
+      - name: openvswitch
+        image: {{.Image}}
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
```
danwinship: This all seems kind of terrible... We can't have an actual `.sh` file?

manifests/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -0,0 +1,111 @@
+{{if .InstallOVS}}
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: ovs
+  namespace: openshift-sdn
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the openvswitch daemon.
```
danwinship: `Open vSwitch` ?

manifests/network/openshift-sdn/sdn.yaml
```diff
@@ -0,0 +1,199 @@
...
+              file="${f}"
+              break
+            done
+          fi
+          # Use the same config as the node, but with the service account token
+          oc config "--config=${file}" view --flatten > /tmp/kubeconfig
+          oc config --config=/tmp/kubeconfig set-credentials sa "--token=$( cat /var/run/secrets/kubernetes.io/serviceaccount/token )"
+          oc config --config=/tmp/kubeconfig set-context "$( oc config --config=/tmp/kubeconfig current-context )" --user=sa
+          # Launch the network process
+          exec openshift start network --config=/etc/origin/node/node-config.yaml --kubeconfig=/tmp/kubeconfig --loglevel=${DEBUG_LOGLEVEL:-2}
```
danwinship: `openshift start network` doesn't exist in git master. You need to use the `openshift-sdn` binary

pkg/operator/openshift_sdn.go
```diff
@@ -5,29 +5,94 @@ import (
...
+	for _, net := range operConfig.ClusterNetworks {
+		ippools = append(ippools, netv1.ClusterNetworkEntry{CIDR: net.CIDR, HostSubnetLength: net.HostSubnetLength})
+	}
+
+	clusterNet := netv1.ClusterNetwork{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "network.openshift.io/v1",
+			Kind:       "ClusterNetwork",
+		},
+		ObjectMeta: metav1.ObjectMeta{Name: "default"},
```
danwinship: `Name: netv1.ClusterNetworkDefault`

pkg/operator/openshift_sdn_test.go
```diff
@@ -3,37 +3,105 @@ package stub
...
+	errs := h.validateOpenshiftSDN()
 	g.Expect(errs).To(HaveLen(0))
 
-	objs, err := h.renderOpenshiftSDN(OpenshiftSDNConfig.Spec.DefaultNetwork.OpenshiftSDNConfig)
+	// Make sure the OVS daemonset isn't created
+	truth := true
+	sdnConfig.UseExternalOpenvswitch = &truth
+	objs, err := h.renderOpenshiftSDN(sdnConfig)
 	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(objs).To(HaveLen(7))
```
danwinship: No, that's way too fragile. We'll have to change the number here every time we add or remove an object from the rendering.

Write a function that looks through the objects and returns the OVS daemonset, and then expect the result `To(BeNil())` here and `NotTo(BeNil())` below.

pkg/operator/openshift_sdn_test.go
```diff
@@ -3,37 +3,105 @@ package stub
...
+
+	// enable openvswitch
+	sdnConfig.UseExternalOpenvswitch = nil
+	objs, err = h.renderOpenshiftSDN(sdnConfig)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(objs).To(HaveLen(8))
+
+	// Check that the Network is first and namespace second
```
danwinship: Why is that important?

pkg/operator/openshift_sdn_test.go
```diff
@@ -3,37 +3,105 @@ package stub
...
+	port := uint32(66666)
+	sdnconfig.VXLANPort = &port
+	errExpect("invalid VXLANPort 66666")
+
+	config.ClusterNetworks = nil
+	errExpect("ClusterNetworks cannot be empty")
+}
 
-	g.Expect(objs).To(HaveLen(1))
+func errExpect(g *GomegaWithT, e error, substr string) {
```
danwinship: I don't think you're actually using this function?

pkg/render/render.go
```diff
@@ -27,6 +27,13 @@ func MakeRenderData() RenderData {
 	}
 }
 
+// AddData is a convenience method to treat render data as strings
+func (d *RenderData) AddData(data map[string]string) {
```
danwinship: I don't think you're using this either?

manifests/network/openshift-sdn/001-rbac.yaml
squeed: definitely, good point!

manifests/network/openshift-sdn/sdn-ovs.yaml
squeed: copied directly from openshift-ansible!

I think the point is that this can be changed without also taking an openshift PR.

pkg/operator/openshift_sdn_test.go
squeed: It's important that the namespace is before any namespaced components. I didn't feel like writing a more flexible ordering test, since this won't be changing, but I'll definitely add some clarification.

manifests/network/openshift-sdn/sdn-ovs.yaml
squeed: copied directly from openshift-ansible too

squeed:
@danwinship updated, PTAL. Thanks!

manifests/network/openshift-sdn/000-ns.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-sdn
+  labels:
+    name: openshift-sdn
+  annotations:
+    openshift.io/node-selector: "" #override default node selector
+    openshift.io/description: "Openshift SDN components"
```
danwinship: "OpenShift"

pkg/operator/openshift_sdn_test.go
```diff
@@ -1,39 +1,120 @@
...
 	}
+	config := h.config.Spec
+	sdnConfig := config.DefaultNetwork.OpenshiftSDNConfig
 
-	errs := h.validateOpenshiftSDN(OpenshiftSDNConfig.Spec.DefaultNetwork.OpenshiftSDNConfig)
+	errs := h.validateOpenshiftSDN()
 	g.Expect(errs).To(HaveLen(0))
 
-	objs, err := h.renderOpenshiftSDN(OpenshiftSDNConfig.Spec.DefaultNetwork.OpenshiftSDNConfig)
+	_ = func(objs []*uns.Unstructured, kind, namespace, name string) {
```
danwinship: What is this? You're creating an anonymous function and then throwing it away without ever calling it?

pkg/operator/testutil_test.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"fmt"
+
+	"github.com/onsi/gomega/types"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// Fun matcher for testing the presence of Kubernetes objects
+func HaveKubernetesID(kind, namespace, name string) types.GomegaMatcher {
+
```
danwinship: gratuitous blank line

danwinship:
(Assuming the anon function is just cruft, and you delete it, and fix the other two nits, then lgtm)

pkg/operator/openshift_sdn_test.go
squeed: whoops, forgot to run GC.

squeed:
Updated, thanks for the review.

manifests/network/openshift-sdn/sdn.yaml
```diff
@@ -0,0 +1,199 @@
...
+          name: host-var-run-ovs
+          readOnly: true
+        - mountPath: /var/run/kubernetes/
+          name: host-var-run-kubernetes
+          readOnly: true
+        # We mount our socket here
+        - mountPath: /var/run/openshift-sdn
+          name: host-var-run-openshift-sdn
+        # CNI related mounts which we take over
+        - mountPath: /host/opt/cni/bin
```
bmeng: Why we mount the path with prefix `/host` for this only? But not be same as others.

Found this long time ago, but I am not sure about the reason. It looks a little weird.

PR #12, opened 2018-09-27
Rename from openshift-network-operator to cluster-network-operator

squeed:


squeed:
Change is essentially a no-op, merging.

PR #13, opened 2018-09-27
install: generate proper install manifests

squeed:
This is so the CVO tooling can pick up and install the operator directly.

@abhinavdahiya


install/00_namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-cluster-network-operator
+  labels:
+    name: openshift-cluster-network-operator
```
abhinavdahiya: https://github.com/openshift/machine-config-operator/blob/3a73b1d142eb1efb3fad5b160915e6ddf3d38311/install/00_namespace.yaml#L7
`openshift.io/run-level: "1"` might be needed as `openshift-apiserver` might not be up.

squeed: definitely, good catch

install/03_deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+    matchLabels:
+      k8s-app: cluster-network-operator
+  template:
+    metadata:
+      labels:
+        k8s-app: cluster-network-operator
+    spec:
+      containers:
+      - name: cluster-network-operator
+        image: openshift/origin-cluster-network-operator:v4.0.0
```
abhinavdahiya: This should match the image in `image-references`

squeed:
fair enough, though for `create`, it doesn't really make any difference.

PR #15, opened 2018-10-02
Restructure to be picked up by the CVO:

squeed:
* Manifests directory is special, not install
* New filename pattern
* Add dockerfile label
* Minor fixes


squeed:
Whoops, don't merge yet - the runlevel is now 7.

PR #17, opened 2018-10-03
WIP: openshift-sdn controller

squeed:
This renders the controller for openshift-sdn, along with its configuration.

*TODO:*
- [ ] Get openshift/origin#21090 merged
- [x] Write CRDs


squeed:
@danwinship

bindata/network/openshift-sdn/controller.yaml
```diff
@@ -0,0 +1,69 @@
...
+    metadata:
+      labels:
+        app: sdn-controller
+    spec:
+      containers:
+      - name: sdn-controller
+        image: {{.Image}}
+        command: ["hypershift", "openshift-network-controller"]
+        args:
+        - "-v=4"
```
danwinship: I did that in dind but we don't want that debug level by default in production

squeed:
This code is still a WIP; I had to grant the sdn-controller `admin` access. Still need to dial back the RBAC permissions.

Also, I'm using a local build of Origin to get the new controller code, which hasn't merged yet.

squeed:
When writing the RBAC policy for the controller, I see two weird accesses:

```
serviceaccounts "sdn-controller" is forbidden: User "system:serviceaccount:openshift-sdn:sdn-controller" cannot get resource "serviceaccounts" in API group "" in the namespace "openshift-infra"
F1004 13:33:01.480094       1 interfaces.go:396] unable to get token for service account: secrets is forbidden: User "system:serviceaccount:openshift-sdn:sdn-controller" cannot list resource "secrets" in API group "" in the namespace "openshift-infra"
```

If I grant access, then this all works. But it seems strange that they should be required...

danwinship:
> `serviceaccounts "sdn-controller" is forbidden: User "system:serviceaccount:openshift-sdn:sdn-controller" cannot get resource "serviceaccounts" in API group "" in the namespace "openshift-infra"`

Looks like that is coming from the [openshift-controller-manager code](https://github.com/openshift/origin/blob/02a8e4d6e3267fec7d29daee8ffb17e31d9c4308/pkg/cmd/openshift-controller-manager/controller/interfaces.go#L104) that I'm reusing.

squeed:
Yup. From what I can tell, it seems to use its kubeconfig to bootstrap a different service account. Strange, but it works.

PR #18, opened 2018-10-09
Add support for Kuryr-Kubernetes

dulek:
Kuryr-Kubernetes is an OpenStack project that implements CNI through
OpenStack Neutron. It's ideal for Kubernetes/OpenShift deployments
running on OpenStack.

This commit aims to implement support for Kuryr-Kubernetes in
cluster-network-operator.


dulek:
@celebdor, @tzumainn: You might want to take a sneak peek on this. Also @tomasedovic might be interested, but I can't tag him for some reason.

bindata/network/kuryr/005-daemon.yaml
```diff
@@ -0,0 +1,170 @@
...
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: kuryr-cni
+  namespace: kuryr
+  annotations:
+# TODO(dulek): Not really sure if we need to hardcode that now or not?
+#    image.openshift.io/triggers: |
+#      [
+#        {"from":{"kind":"ImageStreamTag","name":"node:v3.11"},"fieldPath":"spec.template.spec.initContainers[?(@.name==\"install-cni-plugins\")].image"},
```
celebdor: It seems the triggers are gone from openshiftsdn bindata/network

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,146 @@
...
+	data.Data["OpenStackProjectId"] = *c.OpenStackCredentials.ProjectId
+	data.Data["OpenStackUserDomainName"] = *c.OpenStackCredentials.UserDomainName
+	data.Data["OpenStackUsername"] = *c.OpenStackCredentials.Username
+	data.Data["KuryrProject"] = c.KuryrProject
+
+	// TODO(dulek): Figure out how do I get those
+	data.Data["PodSubnet"] = ""
+	data.Data["PodSecurityGroups"] = ""
+	data.Data["ServiceSubnet"] = ""
+	data.Data["WorkerNodesSubnet"] = ""
```
celebdor: I was thinking that since they will be created by terraform in openshift/installer, that probably openshift/installer should create a ConfigMap asset that we mount also on /etc/kuryr/ and leverage oslo config functionality of reading multiple config files.

@squeed @dcbw Would that be an installer + operator idiomatic flow for the SDN needing things created during deployment?

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// daemon DaemonSet data
+	if c.DaemonEnableProbes == nil {
```
celebdor: We should default probes to true, shouldn't we?

pkg/operator/kuryr.go
celebdor: (obviously we would only mount the configmap on the controller, I forgot to specify :P )

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,146 @@
...
+		data.Data["ControllerEnableProbes"] = *c.ControllerEnableProbes
+	}
+	if c.ControllerProbesPort == nil {
+		data.Data["ControllerProbesPort"] = 8082
+	} else {
+		data.Data["ControllerProbesPort"] = *c.ControllerProbesPort
+	}
+
+	// TODO(dulek): How do I get image names correctly?
+	data.Data["OpenShiftSDNImage"] = ""
```
celebdor: Why do you need the openshiftSDNImage?

bindata/network/kuryr/005-daemon.yaml
dulek: That's why I'm confused - they don't really have an established solution for getting the image name. I don't think that env var thing is going to last.

pkg/operator/kuryr.go
dulek: Another way could be to make openshift/installer put that onto the NetworkConfig object its (I assume?) creating. See the manifests/0007_10_cluster-network-operator_01_crd.yaml file. I'm assuming whole user-facing Kuryr configuration is read from there.

pkg/operator/kuryr.go
dulek: I'm trying to make this compatible with OSP 13 first - in 13 we don't have probes on kuryr-daemon yet.

pkg/operator/kuryr.go
dulek: To get the OpenShift-grade CNI binaries (mainly loopback) and then to be able to run the DNS container. See 005-daemon.yaml and initContainer and sky-dns container there.

bindata/network/kuryr/001-rbac.yaml
```diff
@@ -0,0 +1,84 @@
...
+- kind: ServiceAccount
+  name: kuryr
+  namespace: kuryr
+
+---
+# Rather than PATCHing the SecurityContextConstraints object, we can bind to 
+# it. This is undocumented, I'm copying it from openshift-sdn.
+# TODO: When the openshift-api control plane is splut out, we will need to
+# remove this and replace it with something (sigh) hardcoded
+apiVersion: authorization.openshift.io/v1
```
squeed: This can now be removed!

bindata/network/kuryr/004-config.yaml
```diff
@@ -0,0 +1,462 @@
...
+#       [cni_health_server] (Rocky)
+
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kuryr-config
+  namespace: kuryr
+data:
+  kuryr.conf: |+
```
squeed: Any reason to keep these commented-out lines? Not sure what purpose they serve.

bindata/network/kuryr/005-daemon.yaml
```diff
@@ -0,0 +1,173 @@
...
+        - name: openvswitch
+          hostPath:
+            path: /var/run/openvswitch
+        - name: host-config
+          hostPath:
+            path: /etc/origin/node
+        - name: host-sysconfig-node
+          hostPath:
+            path: /etc/sysconfig/origin-node
+        - name: host-var-run
```
squeed: If you can get rid of bind-mounting `/var/run`, that would be good. Someday I'll get rid of it for openshift-sdn...

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,63 @@
...
+            port: {{ default .ControllerProbesPort 8082 }}
+            scheme: HTTP
+          timeoutSeconds: 5
+        livenessProbe:
+          httpGet:
+            path: /alive
+            port: {{ default .ControllerProbesPort 8082 }}
+          initialDelaySeconds: 15
+{{ end }}
+        # FIXME(dulek): This shouldn't be required, but without it selinux is
```
squeed: weird!

manifests/0007_10_cluster-network-operator_03_deployment.yaml
```diff
@@ -28,6 +28,12 @@ spec:
         env:
         - name: SDN_IMAGE
           value: "docker.io/openshift/origin-node:v4.0.0"
+        - name: NODE_IMAGE
+          value: "docker.io/openshift/origin-node:v4.0.0-local"
```
squeed: Before this is merged, we should get rid of the `-local` and `latest` image references.

pkg/apis/networkoperator/v1/types.go
```diff
@@ -119,6 +123,59 @@ type OpenshiftSDNConfig struct {
...
+	ProjectId         string `json:"projectId"`
+}
+
+// KuryrConfig configures the Kuryr-Kubernetes SDN
+type KuryrConfig struct {
+	// The base64-encoded CA certificate to use when connecting to OpenStack.
+	// +optional
+	OpenStackCACertificate string `json:"openStackCACertificate"`
+
+	OpenStackCredentials *OpenStackCredentials `json:"openStackCredentials"`
```
squeed: Hmm, I wonder if the network configuration should have secrets. @crawford, any thoughts?

pkg/apis/networkoperator/v1/types.go
```diff
@@ -119,6 +123,59 @@ type OpenshiftSDNConfig struct {
...
+	// If true, kuryr-daemon readiness and liveness probes are enabled.
+	// +optional
+	DaemonEnableProbes *bool `json:"daemonEnableProbes"`
+
+	// The port kuryr-daemon will listen for readiness and liveness requests.
+	// +optional
+	DaemonProbesPort *uint16 `json:"daemonProbesPort"`
+
+	// The CNI bin directory on the hosts, usually /opt/cni/bin.
+	CNIBinDir string `json:"cniBinDir"`
```
squeed: Users shouldn't be allowed to configure this; we should treat it as a constant we own.

pkg/apis/networkoperator/v1/types.go
```diff
@@ -119,6 +123,59 @@ type OpenshiftSDNConfig struct {
...
+
+	// If true, kuryr-controller readiness and liveness probes are enabled.
+	// +optional
+	ControllerEnableProbes *bool `json:"controllerEnableProbes"`
```
squeed: Is there any reason for probes to be disabled?

pkg/apis/networkoperator/v1/types.go
```diff
@@ -119,6 +123,59 @@ type OpenshiftSDNConfig struct {
...
+	// The port kuryr-controller will listen for readiness and liveness requests.
+	// +optional
+	ControllerProbesPort *uint16 `json:"controllerProbesPort"`
+
+	// TODO(dulek): Might want to make that a list (scope?)
+	PodSecurityGroups string `json:"podSecurityGroups"`
+
+	PodSubnet string `json:"podSubnet"`
+
+	ServiceSubnet string `json:"serviceSubnet"`
```
squeed: Is there a reason that the IP addresses can't come from the networks above? I would prefer that. It's OK if Kuryr only supports a single ClusterCIDR.

pkg/apis/networkoperator/v1/types.go
```diff
@@ -119,6 +123,59 @@ type OpenshiftSDNConfig struct {
...
+
+	KuryrProject string `json:"kuryrProject"`
+
+	WorkerNodesSubnet string `json:"workerNodesSubnet"`
+
+	DaemonImage string `json:"daemonImage"`
```
squeed: I don't think images should be configurable.

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,147 @@
...
+
+	objs := []*uns.Unstructured{}
+
+	// generate master network configuration
+	ippools := []netv1.ClusterNetworkEntry{}
+	for _, net := range operConfig.ClusterNetworks {
+		ippools = append(ippools, netv1.ClusterNetworkEntry{CIDR: net.CIDR, HostSubnetLength: net.HostSubnetLength})
+	}
+
+	clusterNet := netv1.ClusterNetwork{
```
squeed: a ClusterNetwork is just used for `openshift-sdn` - do you need to create it? I don't think so.

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,147 @@
...
+		// This is very unlikely
+		return nil, errors.Wrap(err, "failed to transmutate ClusterNetwork")
+	}
+	objs = append(objs, obj)
+
+	// render the manifests on disk
+	data := render.MakeRenderData()
+
+	// OpenStack credentials
+	data.Data["OpenStackCACertificate"] = c.OpenStackCACertificate
```
squeed: I wonder if we can come up with a nicer way to do this - sorry that the library makes this so ugly.

pkg/operator/kuryr.go
```diff
@@ -0,0 +1,147 @@
...
+	data.Data["DaemonEnableProbes"] = c.DaemonEnableProbes
+	data.Data["DaemonProbesPort"] = c.DaemonProbesPort
+	data.Data["CNIBinDir"] = c.CNIBinDir
+
+	// kuryr-controller Deployment data
+	data.Data["ControllerEnableProbes"] = c.ControllerEnableProbes
+	data.Data["ControllerProbesPort"] = c.ControllerProbesPort
+
+	// TODO(dulek): How do I get image names correctly?
+	data.Data["NodeImage"] = os.Getenv("NODE_IMAGE")
```
squeed: AFAICT, this is the correct way to do node images. It would be your team's responsibility to build a pipeline for the kuryr images,.

bindata/network/kuryr/001-rbac.yaml
dulek: Done, can you tell my why this is no longer required?

bindata/network/kuryr/004-config.yaml
dulek: This is standard way of providing OpenStack services config template. It was generated by OpenStack tools and operators are used to see it that way. We can discuss squeezing it, but for now I'd rather keep it like that through exploratory period.

bindata/network/kuryr/005-daemon.yaml
dulek: @tomassedovic Hey, do you remember why this /var/run mount is required for SkyDNS?

bindata/network/kuryr/006-controller.yaml
dulek: Yeah, but we've gone through that in openshift-ansible, so copying the workaround here for now.

pkg/apis/networkoperator/v1/types.go
dulek: Okay, great.

pkg/apis/networkoperator/v1/types.go
dulek: We need it as an option because probes support it depends on Kuryr version being used. We could switch to some autodetection, but let's start with that.

dulek: Okay, seems like we won't support that old Kuryr, so whatever.

pkg/apis/networkoperator/v1/types.go
dulek: Uh, this one is complicated. So I should be able to fetch required Neutron subnet ID's using serviceNetwork if I'm allowed to connect to cloud provider (OpenStack) from inside the operator.

With clusterNetworks we work totally differently, so I should probably go with the idea you've described and catch multiple in the validation.

dulek:
Okay, this now creates Kuryr-required OpenStack Neutron networks in the initContainer in kuryr-controller. Still needs some fixes (grep for FIXME and TODO) but I've successfully run this on DevStack + Kubernetes and it worked. :)

@squeed, I'd appreciate second round of reviews, I think all of your remarks are now addressed.

@luis5tb, @celebdor: Can you guys please take a look if that makes sense from Kuryr's point of view? I'll definitely need @luis5tb help to correctly add all the if clauses into kuryr.conf ConfigMap to support every mode (Subnet, Multitenant, NetworkPolicy).

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,268 @@
...
+          # Init neutron client
+          config.init(('--config-file', '/etc/kuryr/kuryr.conf'))
+          neutron = utils.get_neutron_client()
+
+          # TODO(dulek): Error handling.
+          # TODO(dulek): Using cluster_id in names in case there are multiple clusters
+          #              on single OpenStack cloud.
+
+
+          def create_network(net_name):
```
luis5tb: ensure_network?

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,268 @@
...
+                      'network': {
+                          'name': net_name,
+                          'project_id': project_id,
+                      }})['network']['id']
+              else:
+                  net_id = networks['networks'][0]['id']
+              return net_id
+
+
+          def create_subnetpool(subnetpool_name, cidrs):
```
luis5tb: ensure_subnetpool

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,268 @@
...
+              pod_subnet_id = create_subnet('kuryr-pod-subnet', pod_net_id,
+                                            subnetpool_id=pod_sp_id)
+              # FIXME(dulek): This is missing setting DNS server for the pod subnet.
+              data['neutron_defaults']['pod_subnet_pool'] = pod_sp_id
+          elif mode == 'Subnet':
+              pod_subnet_id = create_subnet('kuryr-pod-subnet', pod_net_id,
+                                            cidr=pod_net_cidrs[0])
+          data['neutron_defaults']['pod_subnet'] = pod_subnet_id
+
+          if mode == 'namespace':
```
luis5tb: can mode be Multitenant and namespace? I believe the namespace mode also needs the pod_sg_id

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,268 @@
...
+              # FIXME(dulek): I should also open traffic from those groups to node and
+              #               master.
+
+              data['namespace_sg'] = {
+                  'sg_allow_from_namespace': from_namespace_sg_id,
+                  'sg_allow_from_default': from_default_sg_id,
+              }
+
+          config = ""
+          for section, options in data.items():
```
luis5tb: does this have the extra information needed? like enabling the required kuryr handlers/drivers?

danwinship:
> to correctly add all the if clauses into kuryr.conf ConfigMap to support every mode (Subnet, Multitenant, NetworkPolicy)

Subnet/Multitenant/NetworkPolicy is specifically a configuration option of the openshift-sdn plugin, not of OpenShift networking in general. If you don't already have a kuryr.conf option that can be set to "subnet", "multitenant", or "networkpolicy", then you should not have a field in `KuryrConfig` for that.

bindata/network/kuryr/006-controller.yaml
dulek: Not yet, this will be handled in the the ConfigMap Jinja2 template. Here only go the options that the script generates.

dulek:
@danwinship: We do support equivalent modes in Kuryr, but of course the configuration options required to turn them on are named differently (and it's actually a combination of options). Do you suggest that I should create a new enum instead of reusing the OpenShift-SDN one?

danwinship:
@dulek hm... well, you *shouldn't* support those modes. No other non-openshift-sdn plugin does.

But anyway, `KuryrConfig` should look like the kuryr configuration, not like the OpenShift SDN configuration. So if "it's actually a combination of options", then you should expose those options in `KuryrConfig`, and people can configure them accordingly.

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,326 @@
...
+          import os
+
+          from kuryr.lib import utils
+          import netaddr
+          from neutronclient.common import exceptions as n_exc
+          import requests
+
+          from kuryr_kubernetes import config
+
+          svc_net_cidr = '{{ .ServiceSubnetCIDR }}'
```
squeed: This could change at any time, but the InitContainer won't be re-triggered.

One way to solve this would be to move this code in to the network operator. Does that seem possible?

dulek:
@danwinship: We were told a while ago that to state that Kuryr is supported and default option in OpenShift on OpenStack deployments we should support all the modes of OpenShift-SDN, so that's why we've developed support for them. I'm totally fine with creating our own enum, with option names named in more Kuryr-native way, but those will basically work the same as OpenShift-SDN modes. I'd love to get @celebdor opinion here, for me both ways are okay.

bindata/network/kuryr/006-controller.yaml
dulek: Just for the record, we've agreed to implement this in the operator itself with gophercloud. This should be fairly straightforward.

danwinship:
> We were told a while ago that to state that Kuryr is supported and default option in OpenShift on OpenStack deployments we should support all the modes of OpenShift-SDN

Sigh. I'm not sure where that requirement came from. OpenShift SDN's modes are stupid and only make sense as a historical accident. Especially, there's no good reason to have separate "subnet" and "networkpolicy" modes, which are identical except that one supports NetworkPolicy and one doesn't. This was done in OpenShift SDN for a mix of technical and support-related reasons that don't apply to kuryr.

Also, you can't compatibly implement OpenShift-SDN-style "multitenant" mode anyway, because `oc adm pod-network` only works if you are actually using OpenShift SDN.

FTR, https://jira.coreos.com/browse/SDN-198 shows how we're hoping to reimplement things in OpenShift SDN and OVN in the future.

pkg/network/bootstrap.go
```diff
@@ -0,0 +1,19 @@
+package network
+
+import (
+	"github.com/pkg/errors"
+
+	netv1 "github.com/openshift/cluster-network-operator/pkg/apis/networkoperator/v1"
+)
+
+// Bootstrap creates resources required by SDN on the cloud.
+func Bootstrap(conf *netv1.NetworkConfigSpec) (map[string]interface{}, error) {
```
squeed: Bootstrap should definitely return a properly typed struct. 

It also should reconcile idempotently against OpenStack (or whatever), creating or updating resources as required.

It should also commit its changes - the correct way to do this would be the `NetworkConfig.Status` field, but I could be convinced to store it just as part of the Applied configuration.

pkg/network/kuryr.go
```diff
@@ -0,0 +1,151 @@
...
+
+	objs = append(objs, manifests...)
+	return objs, nil
+}
+
+// validateKuryr checks that the Kuryr specific configuration is basically sane.
+func validateKuryr(conf *netv1.NetworkConfigSpec) []error {
+	out := []error{}
+	kc := conf.DefaultNetwork.KuryrConfig
+	osc := conf.DefaultNetwork.KuryrConfig.OpenStackCredentials
```
squeed: If KuryrConfig is nil, this will panic / segfault.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			return "", errors.Wrap(err, "failed to create network")
+		}
+		return netObj.ID, nil
+	}
+}
+
+func ensureSubnet(client *gophercloud.ServiceClient, name, netId,
+	cidr string, gatewayIp *string) (string, error) {
+	subnetId, err := subnets.IDFromName(client, name)
```
squeed: Is it possible to ensure that the cidr and gateway of the existing subnet are correct? We should (eventually) handle subnet expansion, at least for ServiceNetwork.

Gopkg.toml
```diff
@@ -64,3 +64,7 @@ required = [
 [[constraint]]
   name = "github.com/vishvananda/netlink"
   version = "1.0.0"
+
+[[constraint]]
```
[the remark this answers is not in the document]
dulek: Sure thing, will do.

bindata/network/kuryr/005-daemon.yaml
```diff
@@ -0,0 +1,98 @@
...
+        type: infra
+        openshift.io/component: network
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      hostNetwork: true
+      serviceAccountName: kuryr
+      initContainers:
+      - name: install-cni-plugins
+        image: {{ .NodeImage }}
```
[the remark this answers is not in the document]
dulek: Yes, that's exactly what we want, this step only installs those generic plugin binaries, mainly loopback that is required for other plugins to work correctly.

kuryr-cni plugin binary is installed by kuryr-cni container entrypoint because it's basically a bash script that executes `runc exec` on kuryr-cni container while passing the CNI_* env vars. Kuryr is a Python application and it's extremely problematic to make it a binary that can be easily moved onto the host. Some more info about what solutions we've explored is here: https://opensource.com/article/18/10/how-kubernetize-openstack-service#injecting-the-kuryr-cni-executable

pkg/network/bootstrap.go
dulek: Done.

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,60 @@
...
+          httpGet:
+            path: /alive
+            port: {{ default 8082 .ControllerProbesPort }}
+          initialDelaySeconds: 15
+{{ end }}
+        # FIXME(dulek): This shouldn't be required, but without it selinux is
+        #               complaining about access to kuryr.conf.
+        securityContext:
+          privileged: true
+          runAsUser: 0
```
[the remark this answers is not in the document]
dulek: Done.

pkg/network/kuryr_bootstrap.go
dulek: Oh yes, we might want to look it up based on CIDR instead, good point.

bindata/network/kuryr/004-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+    token_file = /var/run/secrets/kubernetes.io/serviceaccount/token
+    ssl_verify_server_crt = true
+    controller_ha = false
+    controller_ha_elector_port = 16401
+    watch_retry_timeout = 3600
+    pod_vif_driver = nested-vlan
+    vif_pool_driver = nested
+    multi_vif_drivers = noop
+
+    enabled_handlers = vif,lb,lbaasspec,policy,pod_label
```
luis5tb: you are missing namespace here

bindata/network/kuryr/004-config.yaml
```diff
@@ -0,0 +1,70 @@
...
+    pod_security_groups_driver = policy
```
luis5tb: I'm also missing the namespace subnet driver:
pod_subnets_driver = namespace

dulek:
@squeed: Okay, I've added support for looking up openshift-installer-created OpenStack resources through tags, bootstrap is now tagging create resources and looking them up through tags. I've also added BootstrapData struct and split the gophercloud addition into base commit. Also I've tested this on small OpenStack cloud and it configured and deployed Kuryr as expected, so seems like it works.

This might need addition of some security groups manipulations in the bootstrap phase, but we're not yet 100% sure how to approach it due changes to openshift/installer. This should be fine to be addressed as a bug after the freeze (as lack of connectivity is clearly a bug ;)).

dulek:
What is added in new version is support to set up aforementioned security groups and our new finding - setting up OpenShift API LB as it turns out openshift/installer is not willing to do that for us. :disappointed:.

Yeah, I'm using emojis in PR comments, so I guess that's good time to go to sleep.

bindata/network/kuryr/004-config.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+    [namespace_subnet]
+    pod_router = {{ .WorkerNodesRouter }}
+    pod_subnet_pool = {{ .PodSubnetpool }}
+
+    [neutron_defaults]
+    service_subnet = {{ .ServiceSubnet }}
+    project = {{ default "\"\"" .KuryrProject }}
+    pod_security_groups = {{ default "default" .PodSecurityGroups }}
+    #network_device_mtu = 1500
```
celebdor: do we need this in KuryrConfig so people can customize it?

dulek: Discussed on IRC, we'll try to do some autodetection and handle it as a bug later.

bindata/network/kuryr/006-controller.yaml
```diff
@@ -0,0 +1,59 @@
...
+            port: {{ default 8082 .ControllerProbesPort }}
+            scheme: HTTP
+          timeoutSeconds: 5
+        livenessProbe:
+          httpGet:
+            path: /alive
+            port: {{ default 8082 .ControllerProbesPort }}
+          initialDelaySeconds: 15
+{{ end }}
+        # FIXME(dulek): This shouldn't be required, but without it selinux is
```
squeed: Now that this is a ConfigMap, I bet privileged isn't required anymore. Worth a try, but can be a followup.

dulek: We had those selinux issues when that was a ConfigMap. I can totally try removing it and retest. Let's assume it's a follow up, it's trivial.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,620 @@
...
+		return "", errors.New("router not found")
+	} else {
+		return "", errors.New("multiple matching routers")
+	}
+}
+
+func getRouterPorts(client *gophercloud.ServiceClient, routerId string) ([]ports.Port, error) {
+	page, err := ports.List(client, ports.ListOpts{DeviceID: routerId}).AllPages()
+	if err != nil {
+		return []ports.Port{}, errors.Wrap(err, "failed to get port list")
```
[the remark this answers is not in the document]
dulek: Thanks! I've noticed it elsewhere but forgot to fix it here.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,620 @@
...
+	})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create Octavia client")
+	}
+
+	// We need first usable IP from services CIDR
+	// This will get us the first one (subnet IP)
+	ip = svcNet.IP
+	// And this second one (first usable)
+	ip[3] |= 0x01
```
squeed: Interesting. Why is the openshift API service special? Is this just needed to bootstrap the kuryr-controller?

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,620 @@
...
+	listenerId, err := ensureLbListener(lbClient, "kuryr-api-loadbalancer-listener", poolId, 443)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create OpenShift API loadbalancer listener")
+	}
+	log.Printf("OpenShift API loadbalancer listener %s present", listenerId)
+
+	// We need to list all master ports and add them to the LB pool
+	log.Print("Creating OpenShift API loadbalancer pool members")
+	r, _ := regexp.Compile("^master-port-[0-9]+$")
+	portList, err := listPortsMatchingPattern(client, tag, r)
```
squeed: While master nodes are somewhat special now, we do intend for them to be more cattle-like in the future. Does doing this all require the set of master nodes to be constant?

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,620 @@
...
+
+	netv1 "github.com/openshift/cluster-network-operator/pkg/apis/networkoperator/v1"
+)
+
+func tagResource(client *gophercloud.ServiceClient, resource, id, tag string) ([]string, error) {
+	tagOpts := attributestags.ReplaceAllOpts{Tags: []string{tag}}
+	return attributestags.ReplaceAll(client, resource, id, tagOpts).Extract()
+}
+
+func ensureNetwork(client *gophercloud.ServiceClient, name, tag string) (string, error) {
```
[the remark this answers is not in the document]
dulek: Renamed to ensureOpenStackNetwork (seems a bit more appropriate), along with rest of the functions. godoc comments are added to each function now.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,620 @@
...
+	page, err := networks.List(client, networks.ListOpts{Name: name, Tags: tag}).AllPages()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get network list")
+	}
+	nets, err := networks.ExtractNetworks(page)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to extract networks list")
+	}
+	if len(nets) == 1 {
```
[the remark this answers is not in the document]
dulek: I've added that as a possibility and we're failing in that case. That's the best we can do, it's up to user to check out what happened with his cloud. In other case we would just add new resource for every reconciliation and that would be disastrous.

pkg/network/kuryr_bootstrap.go
dulek: Yes, it's needed for the kuryr-controller. We agreed to check out if we can use the openshift-sdn approach and let kuryr-controller configure that LB on its own.

pkg/network/kuryr_bootstrap.go
dulek: If the above comment about moving management of openshift/kubernetes API LB will get resolved, this will get supported automatically by controller. Otherwise I'll need to add tiny logic that will look up for pool members that have no corresponding ports and delete them. If that would be done in reconciliation, we're good.

dulek:
I've addressed all the above comments. Current version was verified on OpenStack cloud with Octavia LBaaS installed and creates resources and configures Kuryr as expected.

dcbw:
A fundamental question about this PR; why is the network operator itself doing a bunch of network configuration with Neutron? Why is that not in a daemonset that is rendered by the Network Operator instead?

dcbw:
Also RE the images, you need to have whatever images you need coming from a repo that gates on origin CI. That typically means an openshift/xxxx public repo that has all the normal OpenShift prow tests including e2e-aws. That image then needs to be pushed to the public OpenShift image repositories where it will be accessible by the CI infrastructure so that it can be tested.

We have the same issue with the ovn-kubernetes and multus images and are working through this process.

celebdor:
> Also RE the images, you need to have whatever images you need coming from a repo that gates on origin CI. That typically means an openshift/xxxx public repo that has all the normal OpenShift prow tests including e2e-aws. That image then needs to be pushed to the public OpenShift image repositories where it will be accessible by the CI infrastructure so that it can be tested. We don't ship things in OSE that don't also have an OCP equivalent. Or at least that's what I've understood from Clayton.
> 
> So docker.io/kuryr/xxxx images aren't going to work. And whatever images the operator configures, _must_ be in manifest/image-references.
> 
> We have the same issue with the ovn-kubernetes and multus images and are working through this process.

We are as well, we already have the repos on openshift/ We were waiting to be feature complete on upstream kuryr-kubernetes today and get the go ahead to populate those repos.

celebdor:
> A fundamental question about this PR; why is the network operator itself doing a bunch of network configuration with Neutron? Why is that not in a daemonset that is rendered by the Network Operator instead?

We actually were doing that in a previous version of the PR and IIRC it was suggested to move it to the operator. That makes sense to me since the resources that are created are infrastructure prerequisites for running Kuryr-Kubernetes. In openshift-ansible they are created by the openstack playbook. In openshift/installer we were told that we should definitely not create infra there depending on which SDN plugin was to be installed.
 
> 
> The network operator has not (at least until now) been responsible for configuring networking between nodes or creating the cluster network itself; that's all the responsibility of the network plugin.

From the OpenStack perspective, connectivity between the nodes is part of the infra responsibility, we are not setting up an overlay or a routed network. What is the advantage of having a separate daemonset that, like the operator, will monitor kubernetes and OpenStack and will create the resources that are prerequisites (a few of them are verified in the readiness check of the kuryr controller pod)?

squeed:
Moving that code in to the network operator was my suggestion. I don't want that code in the installer, since it could have to react to changes in the network configuration. The next iteration was a bunch of python code in a daemonset. I'd rather just execute that code directly in the network operator - saves a step, and is easier to follow errors.

If, someday, the kuryr controller is more advanced, then we could move the code to there. But I'm comfortable with the network operator having this sort of code.

squeed:
I will lgtm this when the image references are correct.

dulek:
Resolved the merge conflicts, I'm still working with OSBS to setup Kuryr images builds in OpenShift infra.

dulek:
This is WiP, seems like I haven't rebased it correctly.

dulek:
Okay, sorry, this was easier than I thought.

sample-kuryr-config.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: "networkoperator.openshift.io/v1"
+kind: "NetworkConfig"
+metadata:
+  name: "default"
+  namespace: cluster-network-operator
+spec:
+  serviceNetwork: "10.3.0.0/24"
+  clusterNetworks:
+    - cidr: "10.4.0.0/16"
+      hostSubnetLength: 9
```
pecameron: I think we are now using HostPrefix instead of hostSubnetLength

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,802 @@
...
+	lbClient, err := openstack.NewLoadBalancerV2(provider, gophercloud.EndpointOpts{})
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create Octavia client")
+	}
+
+	// We need first usable IP from services CIDR
+	// This will get us the first one (subnet IP)
+	ip = iputil.FirstUsableIP(*svcNet)
+	ipStr = ip.String()
+	log.Printf("Creating OpenShift API loadbalancer with IP %s", ipStr)
```
luis5tb: Does the next means we are creating our own lb in front of the master, regardless of the one created by the installer? If that so, this will only be used by the kuryr components, right?

dulek: I believe we talked this through on IRC, but let's repeat it here for the record. Currently the LB created by installer is a service VM. They plan to move to Octavia, but there's no ETA yet and they will still need to support deployments without Octavia. It makes sense to make Kuryr responsible for that LB.

sample-kuryr-config.yaml
dulek: Thanks, I forgot to update the sample.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,844 @@
...
+}
+
+// Tries to create an OpenStack security group rule on sgId SG. Ignores an
+// error if such rule already exists.
+func ensureOpenStackSgRule(client *gophercloud.ServiceClient, sgId, remoteSgId string) error {
+	opts := rules.CreateOpts{
+		SecGroupID: sgId,
+		EtherType: rules.EtherType4,
+		Direction: rules.DirIngress,
+		RemoteGroupID: remoteSgId,
```
luis5tb: use subnetpool cird instead, for scalability reasons

luis5tb: On a second look, it this even used? or just the following method ensureOpenStackSgRuleCIDR? In that case, perhaps it can be removed

dulek: Yeah, I forgot to replace ensureOpenStackSgRule with ensureOpenStackSgRuleCIDR. Fixed on my local copy, will update this PR once openshift/api commit is merged and I can finally make this pass all the tests.

dulek:
This should finally have a chance to pass gate as openshift/api#284 is merged now.

bindata/network/kuryr/001-rbac.yaml
```diff
@@ -0,0 +1,66 @@
...
+rules:
+- apiGroups: [""]
+  resources:
+  - namespaces
+  - nodes
+  - endpoints
+  - services
+  - services/status
+  - pods
+  verbs: ["*"]
```
squeed: This is... a lot. Do you need the ability to write to services and pods? If so, then that's Ok, but it would be nice to pare this down if possible.

bindata/network/kuryr/002-crds.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
squeed: Take this as a TODO to add some kind of validation and pretty printing for these CRDs.

bindata/network/kuryr/004-daemon.yaml
```diff
@@ -0,0 +1,98 @@
...
+      - name: install-cni-plugins
+        image: {{ .NodeImage }}
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -ex
+          # Take over network functions on the node
+          rm -Rf /host-cni-bin/*
```
squeed: This was a silly holdover from 3.11 - you shouldn't be removing other CNI binaries any more. Either remove only your plugin or just delete this line.

bindata/network/kuryr/005-controller.yaml
```diff
@@ -0,0 +1,60 @@
...
+            port: {{ default 8082 .ControllerProbesPort }}
+            scheme: HTTP
+          timeoutSeconds: 5
+        livenessProbe:
+          httpGet:
+            path: /alive
+            port: {{ default 8082 .ControllerProbesPort }}
+          initialDelaySeconds: 15
+{{ end }}
+        # FIXME(dulek): This shouldn't be required, but without it selinux is
```
squeed: Weird. Maybe you can fix this by just symlinking `/etc/kuryr` to somewhere else in the Dockerfile?

bindata/network/kuryr/001-rbac.yaml
luis5tb: Actually we do, Kuryr is using services annotations to differentiate between the k8s side and what has been applied on the OpenStack side

pkg/network/bootstrap.go
```diff
@@ -0,0 +1,40 @@
...
+	ServiceSubnet     string
+	PodSubnetpool     string
+	WorkerNodesRouter string
+	WorkerNodesSubnet string
+	PodSecurityGroups []string
+	ClusterID         string
+	OpenStackCloud    clientconfig.Cloud
+}
+
+type BootstrapData struct {
```
squeed: BootstrapData is the *result* of bootstrapping, right? Then I'd rather be called something like `BootstrapResult` or `BootstrapStatus`. Thoughts?

pkg/network/bootstrap.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Bootstrap creates resources required by SDN on the cloud.
+func Bootstrap(conf *operv1.NetworkSpec, client client.Client) (*BootstrapData, error) {
+	emptyData := BootstrapData{}
+	switch conf.DefaultNetwork.Type {
+	case operv1.NetworkTypeKuryr:
+		return bootstrapKuryr(conf, client)
+	case operv1.NetworkTypeOpenShiftSDN:
+		return &emptyData, nil
```
squeed: Probably easier just to return nil?

pkg/network/kuryr.go
```diff
@@ -0,0 +1,110 @@
...
+	return out
+}
+
+// isKuryrChangeSafe currently returns an error if any changes are made.
+// In the future we'll support changing some stuff.
+func isKuryrChangeSafe(prev, next *operv1.NetworkSpec) []error {
+	pn := prev.DefaultNetwork.KuryrConfig
+	nn := next.DefaultNetwork.KuryrConfig
+
+	if reflect.DeepEqual(pn, nn) {
```
squeed: It would probably be nice to add a TODO to allow whatever safe changes there might be, like probes and the like.

bindata/network/kuryr/001-rbac.yaml
dulek: Yup, we don't yet use CRD's to keep our information, instead we annotate pods and services. This is because Kuryr was there before CRD's.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+
+func GetClusterID(kubeClient client.Client) (string, error) {
+	cluster := &confv1.Infrastructure{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "config.openshift.io/v1", Kind: "Infrastructure"},
+		ObjectMeta: metav1.ObjectMeta{Name: InfrastructureCRDName},
+	}
+
+	err := kubeClient.Get(context.TODO(), client.ObjectKey{Name: InfrastructureCRDName}, cluster)
+	if err != nil {
+		return "", err
```
squeed: Can you add a descriptive `errors.Wrapf` here?

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+	}
+
+	err := kubeClient.Get(context.TODO(), client.ObjectKey{Namespace: CloudsSecretNamespace, Name: CloudsSecret}, secret)
+	if err != nil {
+		return emptyCloud, err
+	}
+
+	content, ok := secret.Data[CloudsSecretKey]
+	if !ok {
+		return emptyCloud, fmt.Errorf("OpenStack credentials secret %v did not contain key %v",
```
squeed: we use `github.com/pkg/errors.Errorf` since it gives pretty tracebacks.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+			CloudsSecret, CloudsSecretKey)
+	}
+	var clouds clientconfig.Clouds
+	err = yaml.Unmarshal(content, &clouds)
+	if err != nil {
+		return emptyCloud, fmt.Errorf("failed to unmarshal clouds credentials stored in secret %v: %v", CloudsSecret, err)
```
squeed: Likewise, `errors.Wrapf` here and in the rest of this file.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+// network and subnet, pods subnetpool, security group and load balancer for
+// OpenShift API. Besides that it looks up router and subnet used by OpenShift
+// worker nodes (needed to configure Kuryr) and makes sure there's a routing
+// between them and created subnets. Also SG rules are added to make sure pod
+// subnet can reach nodes and nodes can reach pods and services. The data is
+// returned to populate Kuryr's configuration.
+func bootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*BootstrapData, error) {
+	log.Print("Kuryr bootstrap started")
+
+	clusterID, err := GetClusterID(kubeClient)
```
squeed: gotta check this `err`.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,845 @@
...
+	log.Print("Ensuring services network")
+	svcNetId, err := ensureOpenStackNetwork(client, "kuryr-service-network", tag)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create service network")
+	}
+	log.Printf("Services network %s present", svcNetId)
+
+	// Service subnet
+	// We need last usable IP from this CIDR. We use first subnet, we don't support multiple entries in Kuryr.
+	_, svcNet, err := net.ParseCIDR(conf.ServiceNetwork[0])
```
squeed: check this err.

bindata/network/kuryr/002-crds.yaml
dulek: Added to backlog. Currently we don't have that upstream either.

pkg/network/bootstrap.go
dulek: Makes sense to me, changing now.

bindata/network/kuryr/005-controller.yaml
dulek: I just removed it, we'll deal with it if we'll encounter it again.

pkg/network/bootstrap.go
dulek: Right, for some reason I made all the other functions in the chain to receive bare struct. Changed them to be able to use a pointer.

bindata/network/kuryr/001-rbac.yaml
squeed: Gotcha. I'd still like to see the list of verbs made explicit, since the ability to create and delete is almost certainly not necessary.

pkg/network/bootstrap.go
```diff
@@ -0,0 +1,39 @@
...
+	switch conf.DefaultNetwork.Type {
+	case operv1.NetworkTypeKuryr:
+		return bootstrapKuryr(conf, client)
+	case operv1.NetworkTypeOpenShiftSDN:
+		return nil, nil
+	case operv1.NetworkTypeOVNKubernetes:
+		return nil, nil
+	}
+
+	return nil, errors.Errorf("unknown or unsupported NetworkType: %s", conf.DefaultNetwork.Type)
```
squeed: We need to support unknown network types, so this should just be `return nil, nil`.

pkg/network/kuryr_bootstrap.go
```diff
@@ -0,0 +1,851 @@
+package network
```
squeed: Minor organizational request: Can you move this to `pkg/platform/openstack/`? I think `pkg/network` is getting too big.

squeed:
This looks very good! Almost there. Only two things left to do:
- Move the bootstrapping code to another package
- Document :-). There needs to be an entry in the README for every possible Kuryr configuration knob. It would also be good to see some kind of overview of how the cloud network is set up, for later debugging help.

bindata/network/kuryr/001-rbac.yaml
dulek: Okay, seems like removing create and delete makes total sense.

squeed:
This looks good to me. @danwinship, would you mind giving this a quick review?

danwinship:
I pushed some fixups to the README. We don't want to reference every plugin type in every possible location throughout the file. Also some wording improvements in the kuryr section. (Still working on the review of the actual code.)

pkg/network/kuryr.go
```diff
@@ -0,0 +1,112 @@
...
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
+	"github.com/openshift/cluster-network-operator/pkg/render"
+)
+
+// renderKuryr returns manifests for Kuryr SDN.
+// This creates
+// - the ClusterNetwork object
```
danwinship: no it doesn't

pkg/network/kuryr.go
```diff
@@ -0,0 +1,112 @@
...
+// - the kuryr-daemon daemonset
+// and some other small things.
+func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult, manifestDir string) ([]*uns.Unstructured, error) {
+	c := conf.DefaultNetwork.KuryrConfig
+	b := bootstrapResult.Kuryr
+
+	objs := []*uns.Unstructured{}
+
+	// render the manifests on disk
+	data := render.MakeRenderData()
```
danwinship: That comment was apparently copied from openshift_sdn.go, where it is also wrong. Please remove it.

bindata/network/kuryr/004-daemon.yaml
```diff
@@ -0,0 +1,97 @@
...
+        openshift.io/component: network
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      hostNetwork: true
+      serviceAccountName: kuryr
+      priorityClassName: system-node-critical
+      initContainers:
+      - name: install-cni-plugins
+        image: {{ .NodeImage }}
```
danwinship: This may be OK for now but kuryr should be using its own image here, which contains whatever CNI plugins it needs, rather than assuming that the openshift-sdn image will have the right ones.

pkg/util/ip/addr.go
```diff
@@ -43,3 +43,21 @@ func lastIP(subnet net.IPNet) net.IP {
 	}
 	return end
 }
+
+// lastUsableIP returns second to last IP of a subnet
+func LastUsableIP(subnet net.IPNet) net.IP {
+	// FIXME(dulek): This should have proper IPv6 support, but for now whatever, Kuryr doesn't support it yet.
+	ip := lastIP(subnet)
+	// And this second to last (last usable)
+	ip[len(subnet.IP)-1] &= 0xFE
```
danwinship: The last IP is guaranteed to be an odd number, so `& 0xFE` just means `- 1`

Also `len(subnet.IP)` should really be `len(ip)`. Also, the comment is ungrammatical. ("And this second to last")

pkg/util/ip/addr.go
```diff
@@ -43,3 +43,21 @@ func lastIP(subnet net.IPNet) net.IP {
...
+	return ip
+}
+
+// FirstUsableIP returns second  IP of a subnet
```
danwinship: (extra space between "second" and "IP")

pkg/util/ip/addr.go
```diff
@@ -43,3 +43,21 @@ func lastIP(subnet net.IPNet) net.IP {
...
+func FirstUsableIP(subnet net.IPNet) net.IP {
+	// FIXME(dulek): This should have proper IPv6 support, but for now whatever, Kuryr doesn't support it yet.
+	ip := subnet.IP
+	// And this second one (first usable)
+	ip[len(ip)-1] |= 0x01
```
danwinship: as above, the base IP is guaranteed to be an even number, so this is really `+ 1`

sample-kuryr-config.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: "operator.openshift.io/v1"
```
danwinship: kill this file. We don't need a sample of each network type.

squeed:
> I didn't look much at the bootstrap code. I worry that synchronously running all of that during the reconciliation loop may cause problems? Also, does it handle the fact that it will be run more than once? It really seems like the reconciliation loop is not the right place to be doing cluster bootstrap...

But it does actually have "reconcile" logic - if the expected network configuration changes (either because the network configuration evolves, or because someone deletes something they shouldn't have), then it correctly reconciles that.

Admittedly, it's no Terraform, but it's a start.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -177,8 +177,15 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// Bootstrap any resources
+	bootstrapResult, err := network.Bootstrap(&operConfig.Spec, r.client)
+	if err != nil {
+		log.Printf("Failed to bootstrap: %v", err)
```
squeed: You should have a SetDegraded call here. Something like
`r.status.SetDegraded(..., "BootstrapError", ...)`

squeed: Also, for user-facing error messages, let's use "reconcile platform networking resources" rather than "bootstrap" as the terminology here.

bindata/network/kuryr/004-daemon.yaml
dulek: @danwinship: Okay, please help me on this one. So this is taken directly from openshift-ansible where playbooks of SDN's were responsible for injecting all the "general" plugins, in particular loopback, which was required for CNI to work. Is that not the case anymore?

manifests/image-references
```diff
@@ -38,3 +38,11 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-ovn-kubernetes:4.2
+  - name: kuryr-cni
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-kuryr-cni:4.2
```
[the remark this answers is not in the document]
dulek: I'll take a look on what steps are needed to publish those on Quay. Currently the images are available in CI and that's probably why images job is passing on the patch.

bindata/network/kuryr/004-daemon.yaml
danwinship: Some day, the SDN image will go away entirely (when we've fully migrated to ovn-kubernetes). So in the long run you should not depend on it.

At the moment you could also get the `loopback` and `host-local` plugins from the `container-networking-plugins-unsupported` image, although that seems like a bug; they should presumably be in the `container-networking-plugins-supported` image instead...

dulek: Fair enough, so I guess I can leave it like it is for now and update once this is sorted out.

manifests/image-references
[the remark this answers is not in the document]
dulek: Images are now published.

https://quay.io/repository/openshift/origin-kuryr-controller?tag=latest&tab=tags

https://quay.io/repository/openshift/origin-kuryr-cni?tag=latest&tab=tags

squeed:
OK, this looks good to me. @danwinship, can you take a look as well?

danwinship:
Yeah, I'm fine with it (though I'd intended for @dulek to just squash my README fixups into his commit)

dulek:
@danwinship, @squeed: I just squashed that readme commit.

dulek:
image

PR #19, opened 2018-10-17
Openshift-sdn support

squeed:
* Renders the NodeConfig directly
* Renders the ControllerConfig directly
* Adds controller CRDs
* Adds RBAC scaffolding
* Adds controller, sdn, and ovs daemonsets


squeed:
This is ready for final review.

pkg/apis/networkoperator/v1/types.go
```diff
@@ -134,6 +137,21 @@ type OVNKubernetesConfig struct {
 // NetworkType describes the network plugin type to configure
 type NetworkType string
 
+// ProxyConfig defines the configuration knobs for kubeproxy
+// All of these are optional and have sensible defaults
+type ProxyConfig struct {
+	// The period that iptables rules are refreshed.
+	// Default: 30s
+	IptablesSyncPeriod string `json:"iptablesSyncPeriod,omitempty"`
```
squeed: @dcbw suggests that we also want the `minSyncPeriod`

danwinship:
>     * Adds controller CRDs

And operator-based installs already no longer have the network types in the apiserver, right? So there's no conflict there?

bindata/network/openshift-sdn/003-rbac-controller.yaml
```diff
@@ -0,0 +1,160 @@
+# Only for testing against vanilla k8s
+# will go away when running against the real installer
```
danwinship: "will go away" because the namespace will already be created by someone else, right? The comment makes it sound like we don't need the namespace when using the real installer...

bindata/network/openshift-sdn/002-rbac.yaml
```diff
@@ -58,31 +58,11 @@ metadata:
...
-
----
-# Rather than PATCHing the SecurityContextConstraints object, we can bind to 
-# it. This is undocumented, thanks to evb who showed me this trick
-# TODO: When the openshift-api control plane is splut out, we will need to
-# remove this and replace it with something (sigh) hardcoded
-apiVersion: authorization.openshift.io/v1
-kind: ClusterRoleBinding
-metadata:
-  name: sdn-privileged
```
danwinship: what's up with the changes here?

bindata/network/openshift-sdn/003-rbac-controller.yaml
```diff
@@ -0,0 +1,160 @@
...
+  name: sdn-controller
+  namespace: openshift-infra
+
+---
+# Only grant access to the specific config map for leader election
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: openshift-sdn-controller-leaderelection
+  namespace: kube-system
```
danwinship: The current version of the controller PR uses the openshift-sdn namespace for this

bindata/network/openshift-sdn/003-rbac-controller.yaml
```diff
@@ -0,0 +1,160 @@
...
+  kind: Role
+  name: openshift-sdn-controller-leaderelection
+subjects:
+- kind: ServiceAccount
+  name: sdn-controller
+  namespace: openshift-sdn
+
+---
+# The controller synthesizes its own authentication, so we need to let it
+# read *and* list screts
```
danwinship: typo "screts"

pkg/apis/networkoperator/v1/types.go
```diff
@@ -48,7 +48,10 @@ type NetworkConfigSpec struct {
 	// +optional
 	DeployKubeProxy *bool `json:"deployKubeProxy,omitempty"`
 
-	// TODO: KubeProxyConfig
+	// KubeProxyConfig lets us configure desired proxy configuration.
+	// If not specified, sensible defaults will be chosen by Openshift directly.
```
danwinship: Open**S**hift

pkg/apis/networkoperator/v1/types.go
```diff
@@ -134,6 +137,21 @@ type OVNKubernetesConfig struct {
...
+	// The period that iptables rules are refreshed.
+	// Default: 30s
+	IptablesSyncPeriod string `json:"iptablesSyncPeriod,omitempty"`
+
+	// The address to "bind" on
+	// Defaults to 0.0.0.0
+	BindAddress string
+
+	// Any additional arguments to pass to the kubeproxy process
+	ProxyArguments map[string][]string
```
danwinship: Maybe the right way to handle kube-proxy configuration is to just *only* have ProxyArguments, and then the SDN node process can convert that into a [insert name of upstream kube-proxy config object] using the same functions kube-proxy itself does (so as to get defaults, etc) to extract the values of any fields it cares about.

In particular:
- As dcbw said, we need iptables minSyncPeriod too
- There's no good argument for exposing the iptables sync period fields but *not* exposing the proxy mode. (And likewise various other fields)
- The current system would let you, eg, set `BindAddress` to one value in this struct and then to a different value in the `ProxyArguments`

pkg/operator/openshift_sdn.go
```diff
@@ -109,3 +102,95 @@ func sdnPluginName(n v1.SDNMode) string {
...
+		Network: cpv1.NetworkControllerConfig{
+			NetworkPluginName:  sdnPluginName(c.Mode),
+			ClusterNetworks:    ippools,
+			ServiceNetworkCIDR: h.config.Spec.ServiceNetwork,
+			VXLANPort:          vxlanPort,
+		},
+
+		LeaderElection: configv1.LeaderElection{
+			Namespace: "openshift-sdn",
+			Name:      "openshift-sdn-controller",
```
danwinship: Neither of these fields gets read. If we're not providing a way to default the *other* fields (timeouts or whatever they are) then you can skip the LeaderElection section)

pkg/operator/openshift_sdn.go
```diff
@@ -109,3 +102,95 @@ func sdnPluginName(n v1.SDNMode) string {
...
+	}
+
+	cfg := cpv1.OpenShiftControllerManagerConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "openshiftcontrolplane.config.openshift.io/v1",
+			Kind:       "OpenShiftControllerManagerConfig",
+		},
+		// no ObjectMeta - not an API object
+
+		Controllers: []string{"openshift.io/sdn"},
```
danwinship: you don't need to set this either

manifests/image-references
```diff
@@ -10,3 +10,7 @@ spec:
     from:
       kind: DockerImage
       name: docker.io/openshift/origin-node:v4.0.0
+  - name: hypershift
+    from:
+      kind: DockerImage
```
pecameron: @squeed in the "docker purge" is this getting changed?

pkg/operator/openshift_sdn.go
```diff
@@ -109,3 +102,95 @@ func sdnPluginName(n v1.SDNMode) string {
...
+
+// controllerConfig builds the contents of controller-config.yaml
+// for the controller
+func (h *Handler) controllerConfig() (string, error) {
+	c := h.config.Spec.DefaultNetwork.OpenshiftSDNConfig
+
+	// generate master network configuration
+	ippools := []cpv1.ClusterNetworkEntry{}
+	for _, net := range h.config.Spec.ClusterNetworks {
+		ippools = append(ippools, cpv1.ClusterNetworkEntry{CIDR: net.CIDR, HostSubnetLength: net.HostSubnetLength})
```
pecameron: This looks like the old (and still valid) format. The new multiple cidr supports has an extended syntax for the HostSubnetLength, and it is a list. Actually, what is the new network cidr syntax?

pkg/operator/openshift_sdn.go
```diff
@@ -109,3 +102,95 @@ func sdnPluginName(n v1.SDNMode) string {
...
+	}
+
+	var vxlanPort uint32 = 4789
```
pecameron: Is 4789 a constant? Should the admin be able to specify a value?

sample-config.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: "networkoperator.openshift.io/v1"
+kind: "NetworkConfig"
+metadata:
+  name: "default"
+  namespace: cluster-network-operator
+spec:
+  serviceNetwork: "172.30.0.0/16"
+  clusterNetworks:
+    - cidr: "10.128.0.0/14"
+      hostSubnetLength: 9
```
pecameron: Should we test a multiple cidr case?

bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -0,0 +1,191 @@
...
+        pluginName:
+          type: string
+        clusterNetworks:
+          type: array
+          items:
+            type: object
+            properties:
+              CIDR:
+                type: string
+                pattern: '^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])/([1-9]|[12][0-9]|3[0-2])$'
```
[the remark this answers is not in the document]
danwinship: That's not "4.0", that's "ovn-kubernetes". This is the config object for OpenShift SDN only.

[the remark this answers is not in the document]
pecameron: @danwinship Didn't @JacobTanenbaum add multiple cluster cidr support to sdn and then port it to ovn? I think there was a modified syntax as well.

bindata/network/openshift-sdn/002-rbac.yaml
squeed: These are no longer needed because we are running in a special mode where we no longer need access to the privileged SecurityContextConstraint.

pkg/apis/networkoperator/v1/types.go
squeed: I didn't think the proxymode was configurable - https://github.com/openshift/origin/blob/master/pkg/cmd/server/kubernetes/network/options/options.go#L60 ?

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: openshift-sdn
   labels:
     name: openshift-sdn
+    openshift.io/run-level: "0"
   annotations:
     openshift.io/node-selector: "" #override default node selector
```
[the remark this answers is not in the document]
squeed: no, but we need them in the daemonsets, good point.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -57,31 +65,14 @@ spec:
...
-              echo "error: No existing node-config.yaml, exiting" 2>&1
-              exit 1
-            fi
-          done
+
+          # Create the node config
+          sed "s/%%NODENAME%%/${NODENAME}/" /config/sdn-config.yaml.in > /tmp/sdn-config.yaml
 
           # Take over network functions on the node
           rm -Rf /etc/cni/net.d/80-openshift-network.conf
```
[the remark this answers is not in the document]
squeed: yes, this is all openshift-sdn specific.

pkg/operator/openshift_sdn.go
squeed: It's just the default, but can be overridden.

sample-config.yaml
squeed: we do, in the unit tests. These are the openshift-ansible defaults, so I'd like to keep them.

pkg/apis/networkoperator/v1/types.go
danwinship: That's the default value. The call to `cmdflags.Resolve` at the end then overrides the defaults using the values from `ProxyArguments`.

pkg/operator/openshift_sdn.go
squeed: Not sure what you mean; this is the exact network configuration type as consumed by the openshift-sdn controller...

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -28,12 +37,11 @@ spec:
       serviceAccountName: sdn
       hostNetwork: true
       hostPID: true
-      priorityClassName: system-node-critical
       containers:
       # The network container launches the openshift-sdn process, the kube-proxy, and the local DNS service.
       # It relies on an up to date node-config.yaml being present.
       - name: sdn
-        image: {{.Image}}
+        image: {{.NodeImage}}
```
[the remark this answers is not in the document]
squeed: yup.

squeed:
> > ```
> > * Adds controller CRDs
> > ```
> 
> And operator-based installs already no longer have the network types in the apiserver, right? So there's no conflict there?

Good point, I have to track that down. The operator runs before the openshift-api, so there's no immediate conflict. I'm not sure if there is a later problem when openshift-api comes up.

squeed:
> > ```
> > * Adds controller CRDs
> > ```
> 
> And operator-based installs already no longer have the network types in the apiserver, right? So there's no conflict there?

Good point. When the operator runs, there isn't an openshift-api server with which to conflict. I need to track down what happens when the openshift-api server comes online.

danwinship:
> > And operator-based installs already no longer have the network types in the apiserver, right? So there's no conflict there?
> 
> Good point, I have to track that down.

I believe https://github.com/openshift/origin/pull/21184 did this

pkg/apis/networkoperator/v1/types.go
squeed: Oh duh.

squeed:
Re-pushed to trigger CI.

squeed:
@danwinship good suggestions about the kube-proxy configuration. Mind if I fix in a follow-up PR?

bindata/network/openshift-sdn/003-rbac-controller.yaml
squeed: Turns out it won't go away; updated the comment.

squeed:
Okay, I'm confused. how do we set `minSyncPeriod`? It's not plumbed through in the config object. Is it just an argument?

danwinship:
oh... sorry, yeah, that's kube-proxy only. the node sdn code only cares about `syncPeriod`, not `minSyncPeriod`

squeed:
Well, I think this is ready for merge. I've tested it a few times with `hack/local-up-cluster.sh`, and it works. Once this is merged (and the other small PRs are merged), we can actually get this in the installer.

README.md
```diff
@@ -26,7 +26,8 @@ BUILDCMD="docker build" ./hack/build-image.sh
 There are some premade Kubernetes manifests to run a development build. After setting the image URL to something sane in the daemonset, do:
 
 ```
-oc create -f ./install/*.yaml
+oc create -f manifests/
+oc create -f sample-config.yaml
 ```
 
 Then you can watch the daemonset with `kubectl -n openshift-cluster-network-operator logs <PODID>`.
```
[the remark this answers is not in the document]
danwinship: what he said; we basically always use `oc` rather than `kubectl` in OpenShift docs currently.

PR #21, opened 2018-10-30
pkg/operator: better reconciliation

squeed:
We need to merge annotations, rather than just stomping on them.


squeed:
Goal is to remove all this when https://github.com/operator-framework/operator-sdk/issues/385 is done.

pkg/operator/reconcile.go
```diff
@@ -122,3 +121,49 @@ func MergeServiceForUpdate(current, updated *uns.Unstructured) error {
 
 	return nil
 }
+
+// mergeAnnotationsExcept merges the annotations between current and updated,
+// preferring those in "updated" if there's a conflict. Keys listed in exceptKeys
+// will have the conflict order reversed; current will win.
```
danwinship: "merges the annotations" and then does what with them?

> mergeAnnotationsExcept copies all of the annotations from `current` to `updated`, except in cases where `updated` already has that annotation, unless the annotation's key appears in `exceptKeys`.

or maybe

> mergeAnnotationsExcept copies annotations from `current` to `updated` where either the annotation does not already appear in `updated`, or else the annotation's key is in `exceptKeys`.

although the naming of both the function and `exceptKeys` doesn't make a lot of sense with either of those descriptions... It's kind of a weird function...

pkg/operator/reconcile.go
```diff
@@ -68,19 +68,25 @@ func (h *Handler) ReconcileObject(ctx context.Context, object *uns.Unstructured)
...
 	}
 
 	if err := MergeServiceForUpdate(current, updated); err != nil {
 		return err
 	}
+
+	// For all object types, merge annotations.
+	// Run this last, in case any of the more specific merge logic has
+	// changed "updated"
+	mergeAnnotationsExcept(current, updated, nil)
```
danwinship: This looks like it should do the wrong thing for the deployment revision annotation, since you're passing `nil` for `exceptKeys` here. Except that it doesn't because `updated` has already been mutated to have `current`s value, but that's really non-obvious. Anyway, maybe you should just only call mergeAnnotationsExcept here if it wasn't a Deployment or Service

squeed:
FWIW, I would love to use https://github.com/openshift/library-go/tree/master/pkg/operator/resource/resourcemerge, but it has a pretty different paradigm from operator-sdk.

squeed:
@danwinship refactored a bit; I got rid of the `mergeAnnotationsExcept` bit and just fixed up the special-case annotation directly in the object-specific code. I still have the slightly strange updating flow, but, it works and I try to comment it.

This should all go away when the library-go reconciler plays nice with the operator-sdk.

pkg/operator/reconcile.go
```diff
@@ -89,36 +94,88 @@ const (
...
+func mergeAnnotations(current, updated *uns.Unstructured) {
+	updatedAnnotations := updated.GetAnnotations()
+	curAnnotations := current.GetAnnotations()
+
+	if curAnnotations == nil {
+		curAnnotations = map[string]string{}
+	}
+
+	if updatedAnnotations == nil {
+		updatedAnnotations = map[string]string{}
```
danwinship: this isn't needed is it?

pkg/operator/reconcile.go
```diff
@@ -89,36 +94,88 @@ const (
...
+func mergeLabels(current, updated *uns.Unstructured) {
+	updatedLabels := updated.GetLabels()
+	curLabels := current.GetLabels()
+
+	if curLabels == nil {
+		curLabels = map[string]string{}
+	}
+
+	if updatedLabels == nil {
+		updatedLabels = map[string]string{}
```
danwinship: (likewise)

pkg/operator/reconcile.go
squeed: as long as the object has an ObjectMeta, no. `GetAnnotations` could theoretically return nil, since it wraps unstructured.

danwinship: But you're not assigning to `updatedAnnotations`, you're only reading from it, so `nil` and empty are equivalent.

squeed: but yeah, we shouldn't do weird contortions for unreachable invalid code; updated.

squeed:
@danwinship updated; can you give this an lgtm if you're satisfied?

squeed:
The operator isn't even in production yet; bah.

PR #23, opened 2018-10-31
Manifests: installer updates

squeed:
We need to tweak the installer manifests slightly:

* Need to be a DaemonSet
* Need locking
* Rename the files a bit


Dockerfile
```diff
@@ -10,6 +10,7 @@ COPY --from=build-env /go/src/github.com/openshift/cluster-network-operator/_out
 COPY --from=build-env /go/src/github.com/openshift/cluster-network-operator/_output/linux/amd64/cluster-network-renderer /bin/cluster-network-renderer
 COPY manifests /manifests
 COPY bindata /bindata
+COPY testbindata /testbindata
```
danwinship: We really want test data in the image? Is that what other people are doing?

squeed: I'm not sure. I haven't really decided how to do e2e testing. I could get rid of this you like.

danwinship: I don't really know...

squeed:
Bah, it looks like I'll have to completely refactor the operator to pick up a small change to the operator-sdk leaderelection logic.

Gonna close this PR, as the new one will be quite different.

PR #24, opened 2018-10-31
pkg/reconcile: ServiceAccount-specific logic

squeed:
ServiceAccounts have a token that is automatically generated, but we were deleting it every time we reconciled. Let the auto-generated one win.

Also, test that all objects from openshift-sdn reconcile to themselves.


squeed:
I figure there's a 50% chance that Dan is just gonna make me write the "merge list by name key" logic anyways, even though we don't need it.

PR #25, opened 2018-11-02
Refactor the operator for operator-sdk v0.1.0

squeed:
This has the following changes:
- operator-sdk v0.1.0 structure
- Convert from Deployment to DaemonSet
- Add leader election
- Validate configuration before applying


pkg/controller/networkconfig/networkconfig_controller.go
```diff
@@ -0,0 +1,132 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// Configuration objects must have a single name
+const CONFIG_OBJECT_NAME = "default"
+
+// The periodic resync interval.
+// We will re-run the reconciliation logic, even if the network configuration
+// hasn't changed.
+var ResyncPeriod = 5 * time.Minute
```
knobunc: I know this is the same as before, but I think @deads2k has told me that a resync that low is pointless...  @deads2k am I mis-remembering?  What number would you advise?

PR #26, opened 2018-11-05
Vxlan port hack

squeed:
The API object definition we're vendoring hasn't been merged in to Origin yet, and it has a field rename. So just set both fields.


danwinship:
You don't actually need to set `vxlanPort` because that's already set by the normal serialization of the object.

Is it possible to convert directly from `cpv1.OpenShiftControllerManagerConfig` to `uns.Unstructured` rather than round-tripping through yaml? (It looks like maybe no?)

squeed:
I'd set both out of an overabundance of caution; but no, we're not going to go backwards in time, so I'll remove it.

Any utility method that converts things to Unstructured just does it with a serialization - deserialization behind the scenes anyways :-)

PR #27, opened 2018-11-06
manifests: change to using installer-created apiserver environment.

squeed:
We override `KUBERNETES_SERVICE_HOST|PORT` since we can't use the service IP. We previously used 127.0.0.1, but that doesn't work during bootstrap since the apiserver is still running on the bootstrap node. Switch to a new configmap created by the installer.


danwinship:
ok, i guess?
/lgtm

PR #28, opened 2018-11-07
Installer + RHCOS fixes

squeed:
Fixes to bring the operator working with the new installer:

- some kubeconfig hackery
- path changes due to SELinux and installer divergence
- need to enable ip forwarding (wat)

ref openshift/installer#600

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -107,7 +107,7 @@ spec:
           path: /sys
       - name: host-config-openvswitch
         hostPath:
-          path: /etc/origin/openvswitch
+          path: /var/lib/openvswitch
       tolerations:
       - operator: "Exists"
 {{- end}}
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -49,6 +49,9 @@ spec:
           #!/bin/bash
           set -euo pipefail
 
+          # enable ip forwarding
+          sysctl -w net.ipv4.ip_forward=1
+
           # if another process is listening on the cni-server socket, wait until it exits
           trap 'kill $(jobs -p); exit 0' TERM
           retries=0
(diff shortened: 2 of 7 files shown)


bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -107,7 +107,7 @@ spec:
           path: /sys
       - name: host-config-openvswitch
         hostPath:
-          path: /etc/origin/openvswitch
```
squeed: This is because the selinux policy forbids writing to this file.

sample-config.yaml
```diff
@@ -3,9 +3,9 @@ kind: "NetworkConfig"
 metadata:
   name: "default"
 spec:
-  serviceNetwork: "172.30.0.0/16"
+  serviceNetwork: "10.3.0.0/16"
   clusterNetworks:
-    - cidr: "10.128.0.0/14"
+    - cidr: "10.2.0.0/16"
```
danwinship: Where did these new defaults come from?

Putting `serviceNetwork` immediately after `clusterNetworks[0].cidr` means that you can't grow that CIDR. I guess since we support multiple clusterNetwork CIDRs now, this isn't fatal, but it's annoying, since it means that if you want to grow from the default, you'll probably end up with, eg, pods on `10.2.0.0/16`, services on `10.3.0.0/16`, and then more pods on `10.4.0.0/16`.

More generically, it is nice to be able to distinguish pod IPs from service IPs at a glance, and using such similar CIDR ranges makes that harder.

If we're changing defaults, we should also consider dropping `hostSubnetLength` back down to 8 since that results in nicer IP range allocations and I think 253 is already more pods-per-node than we support (scaling-wise) on expected cluster configurations anyway.

Also, if we change any of the defaults there will need to be docs updates

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -49,6 +49,9 @@ spec:
           #!/bin/bash
           set -euo pipefail
 
+          # enable ip forwarding
+          sysctl -w net.ipv4.ip_forward=1
```
danwinship: Since RHCOS only exists to support OpenShift, should it enable this itself by default?

cmd/cluster-network-operator/main.go
```diff
@@ -33,6 +45,30 @@ func main() {
 	// TODO: Expose metrics port after SDK uses controller-runtime's dynamic client
 	// sdk.ExposeMetricsPort()
 
+	// Hack: the network operator can't use the apiserver service ip, since theres
+	// no network. We also can't hard-code it to 127.0.0.1, because we run during
+	// bootstrap. Instead, we bind-mount in the kubelets kubeconfig, but just
+	// use it to get the apiserver url.
```
danwinship: there are two apostrophes missing in this comment

sample-config.yaml
squeed: Good point - I was going off the new installer defaults, but I think those are old Tectonic defaults. I'll change the installer to use the usual Openshift ones, and then update the sample object here (which is just for documentation anyways).

bindata/network/openshift-sdn/sdn.yaml
squeed: Probably, yes. This was in openshift-ansible before, so I moved it to the operator.

I'm looking for a reasonable rhcos package to put this in.

squeed:
I'll fix them in a followup.

knobunc:
Thanks

PR #29, opened 2018-11-09
openshift-sdn: Add stupid loop to wait out crio SELinux relabeling.

squeed:
As discovered in openshift/installer#600, there is some murky race within crio with regards to SELinux relabeling. Unfortunately, the SDN is one of the first containers to run, so it loses the race.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -36,6 +36,24 @@ spec:
           #!/bin/bash
           set -euo pipefail
 
+          # wait out the SELinux relabeling race on crio
+          retries=0
+          while true; do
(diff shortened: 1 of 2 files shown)


knobunc:
Horrifying, but does what it says on the tin.

/lgtm

sjenning:
@mrunalp and I are looking into this

mrunalp:
We should probably add ls -lZ in this code to see the SELinux label changes. We don't start runc till rootfs is setup and volumes are setup in crio. i.e. there is no code in crio that changes the volume labels after the container is started.

squeed:
Thanks for looking in to this. As a fun extra, Luca discovered that the machine control agent is bind-mounting /. I wonder if we're racing with that.

PR #36, opened 2018-11-13
manifests: switch network operator to deployment

abhinavdahiya:
https://github.com/openshift/cluster-version-operator/pull/38#issuecomment-438465640

@deads2k suggests that a deployment should be able to schedule even when node is not-ready.

### DIFF
--- a/manifests/0000_07_cluster-network-operator_03_daemonset.yaml
+++ b/manifests/0000_07_cluster-network-operator_03_daemonset.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   name: cluster-network-operator
   namespace: openshift-cluster-network-operator


abhinavdahiya:
@deads2k 
```console
sh-4.2$ oc -n openshift-cluster-network-operator get pods
NAME                                       READY     STATUS    RESTARTS   AGE
cluster-network-operator-5cb656c65-s2s5n   0/1       Pending   0          30s
sh-4.2$ oc -n openshift-cluster-network-operator get pods cluster-network-operator-5cb656c65-s2s5n
NAME                                       READY     STATUS    RESTARTS   AGE
cluster-network-operator-5cb656c65-s2s5n   0/1       Pending   0          45s
sh-4.2$ oc -n openshift-cluster-network-operator describe pods cluster-network-operator-5cb656c65-s2s5n
Name:               cluster-network-operator-5cb656c65-s2s5n
Namespace:          openshift-cluster-network-operator
Priority:           0
PriorityClassName:  <none>
Node:               <none>
Labels:             k8s-app=cluster-network-operator
                    pod-template-hash=176212721
Annotations:        <none>
Status:             Pending
IP:
Controlled By:      ReplicaSet/cluster-network-operator-5cb656c65
Containers:
  cluster-network-operator:
    Image:      registry.svc.ci.openshift.org/ci-op-tl6yk2r1/stable@sha256:7210108e0a9bddbe8773dc23d53f92d3ff6ebb2b8e158a1ebc1fa1ba74dcc6bd
    Port:       <none>
    Host Port:  <none>
    Command:
      /bin/cluster-network-operator
      --url-only-kubeconfig=/etc/kubernetes/kubeconfig
    Limits:
      cpu:     20m
      memory:  50Mi
    Requests:
      cpu:     20m
      memory:  50Mi
    Environment:
      NODE_IMAGE:        registry.svc.ci.openshift.org/ci-op-tl6yk2r1/stable@sha256:2915dc4741e62def2bf15886012e2d0ad3cb5dea0db2f4b516275c3537220000
      HYPERSHIFT_IMAGE:  registry.svc.ci.openshift.org/ci-op-tl6yk2r1/stable@sha256:d8c99f7fc22bd8e387ef5cbf85d69c4f1ed59bb17da0944beb33a333d963add7
      POD_NAME:          cluster-network-operator-5cb656c65-s2s5n (v1:metadata.name)
    Mounts:
      /etc/kubernetes/kubeconfig from host-kubeconfig (ro)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-s29cp (ro)
Conditions:
  Type           Status
  PodScheduled   False
Volumes:
  host-kubeconfig:
    Type:          HostPath (bare host directory volume)
    Path:          /etc/kubernetes/kubeconfig
    HostPathType:
  default-token-s29cp:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-s29cp
    Optional:    false
QoS Class:       Guaranteed
Node-Selectors:  node-role.kubernetes.io/master=
Tolerations:     node-role.kubernetes.io/master:NoSchedule
                 node.kubernetes.io/memory-pressure:NoSchedule
                 node.kubernetes.io/not-ready:NoSchedule
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
  Warning  FailedScheduling  4s (x15 over 53s)  default-scheduler  0/3 nodes are available: 3 node(s) were not ready.
```

squeed:
I assume we'll have to add some tolerations too...

deads2k:
@aveshagarwal @sjenning I thought it may have been because the scheduler wasn't starting on the bootstrap node, but if it is, then it appears something is still wrong  :(

aveshagarwal:
> @deads2k suggests that a deployment should be able to schedule even when node is not-ready.

I think that is only applicable to daemonsets, not for deployments. If we want to have deployments scheduled to not-ready nodes, it would require adding not-ready tolerations to them.

aveshagarwal:
See here: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/daemon/util/daemonset_util.go#L57-#L61

aveshagarwal:
Though it does not seem like a normal pattern to have deployments forced to schedule on not-ready nodes.

aveshagarwal:
also can you share the output of "oc describe nodes" for master nodes to see what is going on with them and what do they have w.r.t. taints?

aveshagarwal:
@deads2k what is the reason we are using deployments and not daemonsets, which are specifically (at least one of the main reasons) for bootstrapping purposes?

aveshagarwal:
@deads2k also in 1.12, DS are scheduled by the default scheduler so should work as far as scheduler is running.

aveshagarwal:
> node.kubernetes.io/not-ready:NoSchedule

Or at the minimum try to change it to to add a new one: node.kubernetes.io/not-ready:NoExecute and see if helps.

ravisantoshgudimetla:
What version of kube are we on? This could be related to https://github.com/openshift/cluster-version-operator/pull/35#issuecomment-429148379, where we have not enabled TaintNodesByCondition and instead the pod created from deployment go through `CheckNodeConditions` pre

deads2k:
@aveshagarwal if you have a daemonset for an operator, it runs on every node.  That's fine with 5 nodes.  It's not fine with 500 nodes.  In addition, you cannot scale them.  The rollout functionality on updates isn't the same, so lease acquisition is weird.

Basically, deployments are the rigth resource.  Let's figure out what's wrong in the scheduler or deployment that prevents it from scheduling.

squeed:
fwiw, the network operator daemonset only runs on the masters, so it's not so bad.

aveshagarwal:
> @aveshagarwal if you have a daemonset for an operator, it runs on every node. That's fine with 5 nodes. It's not fine with 500 nodes. In addition, you cannot scale them. The rollout functionality on updates isn't the same, so lease acquisition is weird.

> Basically, deployments are the rigth resource. Let's figure out what's wrong in the scheduler or deployment that prevents it from scheduling.

I think there is nothing wrong in scheduler as thats how the scheduler has been so far (until 1.11). By default, scheduler eliminates unschedulable nodes (including not-ready), and as deployments are scheduled by the default scheduler, deployments can not be scheduled on unschedulable nodes by the default scheduler. This behavior is different for DS because DS controller does the actual scheduling of DS pods (upto 1.11) so schedules pods on unschedulable nodes (including not-ready).

I think the behavior you are looking for deployments should be achievable in 1.12. Because in 1.12, TaintNodesByCondition is enabled by default and node conditions are managed by using taints. That means no matter what are the conditions on nodes, pods (irrespective of whether via deployments or DS, job, cronjob, statefulsets etc) can be scheduled on any node as far as the pods are assigned right tolerations.

deads2k:
> Because in 1.12, TaintNodesByCondition is enabled by default and node conditions are managed by using taints

Can you enable it in 1.11?  Is there a reason not to?

aveshagarwal:
> > Because in 1.12, TaintNodesByCondition is enabled by default and node conditions are managed by using taints
> 
> Can you enable it in 1.11? Is there a reason not to?

I would not be comfortable in enabling alpha feature in 1.11 because

1. In 1.12, the feature has seen some testing so there is a bit of confidence about its stability.
2. That would also mean bringing those patches that have gone into kube to make it more stable. Dont remember off the top of my head what those are.

Also if the operator stuff is for 4.0 that means 1.12, why would we want to enable the alpha feature in 1.11?

squeed:
I've already run the operator as a deployment against a vanilla 1.13 cluster. I can test 1.12.

aveshagarwal:
> I've already run the operator as a deployment against a vanilla 1.13 cluster. I can test 1.12.

Yes please.

aveshagarwal:
> I've already run the operator as a deployment against a vanilla 1.13 cluster. I can test 1.12.

Did you try with nodes with not-ready conditions?

abhinavdahiya:
only 2 tests failed Woohoo! with 1.12 rebase in this seems to be ready for review.
cc @squeed 

failures from [e2e-aws](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/36/pull-ci-openshift-cluster-network-operator-master-e2e-aws/661)
```console
Failing tests:

[sig-storage] Dynamic Provisioning DynamicProvisioner should provision storage with different parameters [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] Volume limits should verify that all nodes have volume limits [Suite:openshift/conformance/parallel] [Suite:k8s]
```

/retest

squeed:
We should switch the openshift-sdn controller to a Deployment now too.

squeed:
I can see from the e2e run that the network operator succeeded, but bootstrap didn't succeed. So that's a flake (albeit a troubling one).

sjenning:
the file needs to be renamed as well

PR #37, opened 2018-11-14
openshift-sdn: use the crio CRI socket

squeed:
We query the CRI endpoint directly. The code defaults to the dockershim, but now we're using crio.

### DIFF
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -195,6 +195,12 @@ func nodeConfig(conf *netv1.NetworkConfig) (string, error) {
 			ClientCA:    "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
 			BindAddress: bindAddress + ":10251", // port is unused
 		},
+
+		// Openshift-sdn calls the CRI endpoint directly; point it to crio
+		KubeletArguments: legacyconfigv1.ExtendedArguments{
+			"container-runtime":          {"remote"},
+			"container-runtime-endpoint": {"/var/run/crio/crio.sock"},
+		},
 	}
 
 	if kpc != nil && kpc.IptablesSyncPeriod != "" {


pkg/network/openshift_sdn.go
```diff
@@ -195,6 +195,12 @@ func nodeConfig(conf *netv1.NetworkConfig) (string, error) {
 			ClientCA:    "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
 			BindAddress: bindAddress + ":10251", // port is unused
 		},
+
+		// Openshift-sdn calls the CRI endpoint directly; point it to crio
+		KubeletArguments: legacyconfigv1.ExtendedArguments{
```
smarterclayton: So ultimately this is a rhcos / rhel / installer problem - why is crio not at the standard CRI path instead of having to know directly about crio?  @mrunalp for 4.0 it would be better if crio was at the path the kubelet defaults to.

smarterclayton: CRI is an interface, I should have to know it’s crio.

squeed:
This should resolve a lot of the installer's networking issues.

pkg/network/openshift_sdn.go
derekwaynecarr: Kubelet still special cases a crio sock for cadvisor , I’m not following Clayton’s comments.

wking: > Kubelet still special cases a crio sock for cadvisor , I’m not following Clayton’s comments.

CRI should allow the kubelet (and other consumers) to run on top of a given CRI implementation (e.g. CRI-O), without having to break the abstraction and care about which implementation it is.

derekwaynecarr: @wking i agree, that is not the state of the world at this time.  it will be 6mos or more before that is true.

PR #38, opened 2018-11-19
Remove ip_forward sysctl

squeed:
This addresses @danwinship's review comments in #28.

The sysctl was moved to openshift/machine-config-operator#178.

The default network blocks are being changed in openshift/installer#658

some extra apostrophe's we're added.


squeed:
Interesting...

Looking in to the test issues, but I suspect these are bad flakes. If this change were bad, the whole cluster would just never come up.

PR #42, opened 2018-11-26
Fix capitalization of config file data

### LINKED ISSUE #20: Should capitalize "OpenShift" correctly in config data
There's a lot of inconsistency between "OpenShift" and "Openshift" in internal type names, but for anything that might actually show up in a config object that an end user might have to edit, it seems to me like we should be consistently correct? (eg, regardless of what the name of `NetworkTypeOpenshiftSDN` is, the value should be `"OpenShiftSDN"` not `"OpenshiftSDN"`)

danwinship:
It's annoying to have the wrong capitalization in the code, but at least if someone types it wrong, it will fail to compile. But in config files, if you get the naming wrong, the config may be accepted but then misinterpreted. So we should capitalize things in the standard ways.

Fixes #20


danwinship:
@squeed sorry, had the rebased version all queued up and ready to go after #41 landed but then forgot to push it

PR #44, opened 2018-11-28
Initial support for ovn-kubernetes

pecameron:
This sets up the structural support for ovn-kubernetes overlay networks.
This is very early, lots more changes in the works.
It supplies pkg/network/{ovn_kubernetes.go,}

And modifies pkg/network/{cluster_config.go,render.go} to reference ovn

It supplies the needed bindata/network/ovn-kubernetes/ files.

The operator comes up and creates the ovn environment

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ It follows the [Controller pattern](https://godoc.org/github.com/kubernetes-sigs
 
 Most users will be able to use the top-level OpenShift Config API, which has a [Network type](https://github.com/openshift/api/blob/master/config/v1/types_network.go#L26). The operator will automatically translate the `Network.config.openshift.io` object in to a `Network.operator.openshift.io`.
 
-When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
+When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed network plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
 
 # Configuring
 The network operator gets its configuration from two objects: the Cluster and the Operator configuration. Most users only need to create the Cluster configuration - the operator will generate its configuration automatically. If you need finer-grained configuration of your network, you will need to create both configurations.
@@ -44,6 +44,8 @@ spec:
   - 172.30.0.0/16
 ```
 
+Alternatively, ovn-kubernetes can be configured by setting `networkType: OVNKubernetes`.
+
 *Corresponding Operator Config*
 This configuration is the auto-generated translation of the above Cluster configuration.
 ```yaml
@@ -62,8 +64,10 @@ spec:
   - 172.30.0.0/16
 ```
 
+(For ovn-kubernetes, `type: OVNKubernetes`.)
+
 ## Configuring IP address pools
-Users must supply at least two address pools - one for pods, and one for services. These are the ClusterNetwork and ServiceNetwork parameter. Some network plugins, such as OpenShiftSDN, support multiple ClusterNetworks. All address blocks must be non-overlapping. You should select address pools large enough to fit your anticipated workload.
+Users must supply at least two address pools - one for pods, and one for services. These are the ClusterNetwork and ServiceNetwork parameter. Some network plugins, such as OpenShiftSDN and OVNKubernetes, support multiple ClusterNetworks. All address blocks must be non-overlapping. You should select address pools large enough to fit your anticipated workload. Each pool must be able to hold 1 or more hostPrefix allocations.
 
 For future expansion, multiple `serviceNetwork` entries are allowed by the configuration but not actually supported by any network plugins. Supplying multiple addresses is invalid.
 
@@ -74,12 +78,12 @@ hostPrefix: 23
 ```
 means nodes would get blocks of size `/23`, or 512 addresses.
 
-IP address pools are always read from the Cluster configuration and propagated "downwards" in to the Operator configuration. Any changes to the Operator configuration will be ignored.
+IP address pools are always read from the Cluster configuration and propagated "downwards" into the Operator configuration. Any changes to the Operator configuration are ignored.
 
 Currently, changing the address pools once set is not supported. In the future, some network providers may support expanding the address pools.
 
 
-Example
+Example:
 ```yaml
 spec:
   serviceNetwork:
@@ -96,7 +100,7 @@ Users must select a default network provider. This cannot be changed. Different 
 
 The network type is always read from the Cluster configuration.
 
-Currently, the only understood value for network Type is `OpenShiftSDN`.
+Currently, the only understood values for network Type are `OpenShiftSDN` and `OVNKubernetes`.
 
 Other values are ignored. If you wish to use use a third-party network provider not managed by the operator, set the network type to something meaningful to you. The operator will not install or upgrade a network provider, but all other Network Operator functionality remains.
 
@@ -122,8 +126,30 @@ spec:
       useExternalOpenvswitch: false
 ```
 
+### Configuring OVNKubernetes
+OVNKubernetes supports the following configuration options, all of which are optional:
+* `MTU`: The MTU to use for the geneve overlay. The default is the MTU of the node that the cluster-network-operator is first run on, minus 100 bytes for geneve overhead. If the nodes in your cluster don't all have the same MTU then you may need to set this explicitly.
+
+These configuration flags are only in the Operator configuration object.
+
+Example:
+```yaml
+spec:
+  defaultNetwork:
+    type: OVNKubernetes
+    ovnKubernetesConfig:
+      mtu: 1400
+```
+
 ## Configuring kube-proxy
-Users may customize the kube-proxy configuration. None of these settings are required:
+Some plugins (like OpenShift SDN) have a built-in kube-proxy, some plugins require a standalone kube-proxy to be deployed, and some (like ovn-kubnernetes) don't use kube-proxy at all.
+
+The deployKubeProxy flag can be used to indicate whether CNO should deploy a standalone kube-proxy, but for supported network types, this will default to the correct value automatically.
+
+The configuration here can be used for third-party plugins with a separate kube-proxy process as well.
+
+For plugins that use kube-proxy (whether built-in or standalone), you can configure the proxy via kubeProxyConfig
+
 
 * `iptablesSyncPeriod`: The interval between periodic iptables refreshes. Default: 30 seconds. Increasing this can reduce the number of iptables invocations.
 * `bindAddress`: The address to "bind" to - the address for which traffic will be redirected.
--- a/bindata/network/ovn-kubernetes/000-ns.yaml
+++ b/bindata/network/ovn-kubernetes/000-ns.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: ovnkube.sh in the OVN image currently hardcodes this namespace name
+  name: openshift-ovn-kubernetes
+  labels:
+    openshift.io/run-level: "0"
+  annotations:
+    openshift.io/node-selector: "beta.kubernetes.io/os=linux"
+    openshift.io/description: "OVN Kubernetes components"
--- a/bindata/network/ovn-kubernetes/002-rbac.yaml
+++ b/bindata/network/ovn-kubernetes/002-rbac.yaml
@@ -1,0 +1,52 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ovn-kubernetes-node
+  namespace: openshift-ovn-kubernetes
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openshift-ovn-kubernetes-node
+rules:
+- apiGroups: [""]
+  resources:
+  - endpoints
+  - namespaces
+  - nodes
+  - pods
+  - services
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: ["extensions", "networking.k8s.io"]
+  resources:
+  - networkpolicies
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: [""]
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: openshift-ovn-kubernetes-node
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: openshift-ovn-kubernetes-node
+subjects:
+- kind: ServiceAccount
+  name: ovn-kubernetes-node
+  namespace: openshift-ovn-kubernetes
--- a/bindata/network/ovn-kubernetes/003-rbac-controller.yaml
+++ b/bindata/network/ovn-kubernetes/003-rbac-controller.yaml
@@ -1,0 +1,59 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ovn-kubernetes-controller
+  namespace: openshift-ovn-kubernetes
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openshift-ovn-kubernetes-controller
+rules:
+- apiGroups: [""]
+  resources:
+  - namespaces
+  - nodes
+  - pods
+  verbs:
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups: [""]
+  resources:
+  - services
+  - endpoints
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: ["extensions", "networking.k8s.io"]
+  resources:
+  - networkpolicies
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: [""]
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: openshift-ovn-kubernetes-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: openshift-ovn-kubernetes-controller
+subjects:
+- kind: ServiceAccount
+  name: ovn-kubernetes-controller
+  namespace: openshift-ovn-kubernetes
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -1,0 +1,303 @@
+# ovnkube-master
+# daemonset version 3
+# starts master daemons, each in a separate container
+# it is run on the master node(s)
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: ovnkube-master
+  # namespace set up by install
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the ovn-kubernetes controller (master) networking components.
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: ovnkube-master
+  strategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        name: ovnkube-master
+        component: network
+        type: infra
+        openshift.io/component: network
+        beta.kubernetes.io/os: "linux"
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      # Requires fairly broad permissions - ability to read all services and network functions as well
+      # as all pods.
+      serviceAccountName: ovn-kubernetes-controller
+      hostNetwork: true
+      hostPID: true
+      containers:
+      # firewall rules for ovn - assumed to be setup
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6641 -j ACCEPT
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6642 -j ACCEPT
+      # ovs flow for ovn (geneve)
+      # /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+
+      # run-ovn-northd - v3
+      - name: run-ovn-northd
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "run-ovn-northd"]
+
+        volumeMounts:
+        # ovn db is stored in the pod in /etc/openvswitch (default ovs tool location)
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+        - name: OVN_LOG_NORTHD
+          value: "-vconsole:info"
+        - name: OVN_NET_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: net_cidr
+        - name: OVN_SVC_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: k8s_apiserver
+        ports:
+        - name: healthz
+          containerPort: 10257
+        # TODO: Temporarily disabled until we determine how to wait for clean default
+        # config
+        # livenessProbe:
+        #   initialDelaySeconds: 10
+        #   httpGet:
+        #     path: /healthz
+        #     port: 10257
+        #     scheme: HTTP
+        lifecycle:
+      # end of container
+
+      # nb-ovsdb - v3
+      - name: nb-ovsdb
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "nb-ovsdb"]
+
+        volumeMounts:
+        # ovn db is stored in the pod in /etc/openvswitch (default ovs tool location)
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+        - name: OVN_LOG_NB
+          value: "-vconsole:info -vfile:info"
+        - name: OVN_NET_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: net_cidr
+        - name: OVN_SVC_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: k8s_apiserver
+        ports:
+        - name: healthz
+          containerPort: 10256
+        # TODO: Temporarily disabled until we determine how to wait for clean default
+        # config
+        # livenessProbe:
+        #   initialDelaySeconds: 10
+        #   httpGet:
+        #     path: /healthz
+        #     port: 10256
+        #     scheme: HTTP
+        lifecycle:
+      # end of container
+
+      # sb-ovsdb - v3
+      - name: sb-ovsdb
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "sb-ovsdb"]
+
+        volumeMounts:
+        # ovn db is stored in the pod in /etc/openvswitch (default ovs tool location)
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+        - mountPath: /var/run/kubernetes/
+          name: host-var-run-kubernetes
+
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+        - name: OVN_LOG_SB
+          value: "-vconsole:info -vfile:info"
+        - name: OVN_NET_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: net_cidr
+        - name: OVN_SVC_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: k8s_apiserver
+        ports:
+        - name: healthz
+          containerPort: 10255
+        # TODO: Temporarily disabled until we determine how to wait for clean default
+        # config
+        # livenessProbe:
+        #   initialDelaySeconds: 10
+        #   httpGet:
+        #     path: /healthz
+        #     port: 10255
+        #     scheme: HTTP
+        lifecycle:
+      # end of container
+
+      - name: ovnkube-master
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "ovn-master"]
+
+        volumeMounts:
+        # ovn db is stored in the pod in /etc/openvswitch (default ovs tool location)
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+        - mountPath: /var/run/kubernetes/
+          name: host-var-run-kubernetes
+
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+        - name: OVN_MASTER
+          value: "true"
+        - name: OVNKUBE_LOGLEVEL
+          value: "4"
+        - name: OVN_NET_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: net_cidr
+        - name: OVN_SVC_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: k8s_apiserver
+        ports:
+        - name: healthz
+          containerPort: 10254
+        # TODO: Temporarily disabled until we determine how to wait for clean default
+        # config
+        # livenessProbe:
+        #   initialDelaySeconds: 10
+        #   httpGet:
+        #     path: /healthz
+        #     port: 10254
+        #     scheme: HTTP
+        lifecycle:
+      # end of container
+
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+        beta.kubernetes.io/os: "linux"
+      volumes:
+      - name: host-modules
+        hostPath:
+          path: /lib/modules
+      - name: host-var-lib-ovs
+        hostPath:
+          path: /var/lib/openvswitch
+      - name: host-var-run-ovs
+        hostPath:
+          path: /var/run/openvswitch
+      - name: host-var-run-kubernetes
+        hostPath:
+          path: /var/run/kubernetes
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        operator: "Exists"
+        effect: "NoSchedule"
+      - key: "node.kubernetes.io/not-ready"
+        operator: "Exists"
+        effect: "NoSchedule"
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -1,0 +1,243 @@
+---
+# ovnkube-node
+# daemonset version 3
+# starts node daemons for ovs and ovn, each in a separate container
+# it is run on all nodes
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: ovnkube-node
+  # namespace set up by install
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the ovn-kubernetes per node networking components.
+spec:
+  selector:
+    matchLabels:
+      app: ovnkube-node
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: ovnkube-node
+        component: network
+        type: infra
+        openshift.io/component: network
+        beta.kubernetes.io/os: "linux"
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      # Requires fairly broad permissions - ability to read all services and network functions as well
+      # as all pods.
+      serviceAccountName: ovn-kubernetes-node
+      hostNetwork: true
+      hostPID: true
+      containers:
+
+      # ovsdb-server and ovs-switchd daemons
+      - name: ovs-daemons
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "ovs-server"]
+
+        securityContext:
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
+        - mountPath: /lib/modules
+          name: host-modules
+          readOnly: true
+        - mountPath: /run/openvswitch
+          name: host-run-ovs
+        - mountPath: /var/run/openvswitch
+          name: host-var-run-ovs
+        - mountPath: /sys
+          name: host-sys
+          readOnly: true
+        - mountPath: /etc/openvswitch
+          name: host-config-openvswitch
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+
+
+      # firewall rules for ovn - assumed to be setup
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6641 -j ACCEPT
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6642 -j ACCEPT
+      # ovs flow for ovn (geneve)
+      # /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+      # The network container launches the ovn-k8s-cni-overlay process, the kube-proxy, and the local DNS service.
+      # It relies on an up to date node-config.yaml being present.
+      - name: ovn-controller
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "ovn-controller"]
+
+        securityContext:
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
+        # Directory which contains the host configuration.
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+        # CNI related mounts which we take over
+        - mountPath: /host/opt/cni/bin
+          name: host-opt-cni-bin
+        - mountPath: /etc/cni/net.d
+          name: host-etc-cni-netd
+        - mountPath: /var/lib/cni/networks/ovn-k8s-cni-overlay
+          name: host-var-lib-cni-networks-ovn-kubernetes
+
+        resources:
+          requests:
+            cpu: 100m
+            memory: 300Mi
+        env:
+        - name: OVN_DAEMONSET_VERSION
+          value: "3"
+        - name: OVNKUBE_LOGLEVEL
+          value: "4"
+        - name: OVN_NET_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: net_cidr
+        - name: OVN_SVC_CIDR
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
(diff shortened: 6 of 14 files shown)


pecameron:
@squeed am I on the right track with this? PTAL
I will continue to flesh this out over the next few days, so making changes is easy now.

danwinship:
> Signed-off-by: root root@wsfd-netdev31.ntdv.lab.eng.bos.redhat.com

lol. You don't need to use Signed-off-by in OpenShift/Kubernetes repos, and if you did, then you're doing it wrong. :)

pecameron:
@danwinship Oops! Thanks for spotting that. Does this seem to be organized reasonably?

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,161 @@
...
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openshift-sdn-controller
+rules:
+- apiGroups: [""]
+  resources:
+  - namespaces
```
danwinship: As with the node-side ClusterRole, you need to double-check this against the ovn-kubernetes code.

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,161 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: openshift-sdn-controller
+subjects:
+- kind: ServiceAccount
+  name: sdn-controller
+  namespace: openshift-infra
+
+---
+# Only grant access to the specific config map for leader election
```
danwinship: You don't need this object or the next one because ovn-kube doesn't use etcd-based leader election

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,161 @@
...
+  kind: Role
+  name: openshift-sdn-controller-leaderelection
+subjects:
+- kind: ServiceAccount
+  name: sdn-controller
+  namespace: openshift-sdn
+
+---
+# The controller synthesizes its own authentication, so we need to let it
+# read *and* list secrets
```
danwinship: I don't think you need this one either...

bindata/network/ovn-kubernetes/controller.yaml
```diff
@@ -0,0 +1,74 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: ovn-kubernetes
+  name: sdn-controller-config
+data:
+  controller-config.yaml: |-
+{{.NetworkControllerConfig | indent 4}}
```
danwinship: So this is a template for a ConfigMap. In the OpenShift SDN case, it gets filled in with the YAML config data that the OpenShift SDN controller needs, and then that data gets mounted as `/config/controller-config.yaml` in the SDN controller pod, and the pod spec below invokes `hypershift openshift-network-controller --config=/config/controller-config.yaml`.

You need to redo this in a way that works for ovn-kube. Since ovn-kube doesn't take a config file, you'll probably want to use the ConfigMap to fill in pod environment variables rather than a file (see the ConfigMap docs in kubernetes if you're not familiar with this), and define a bunch of single variables here like `K8S_APISERVER` and `CLUSTER_SUBNET` and such, and then in the pod spec, suck the ConfigMap into the environment and then use the environment variables in the `ovnkube` command line.

Oh, actually, ovnkube-master.yaml and ovnkube.yaml below are already written to be pulling data out of a configmap called `ovn-config`, so that's what you want to be creating a template for here.

bindata/network/ovn-kubernetes/controller.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: sdn-controller
+  namespace: ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This deployment runs the openshift SDN networking controller.
```
danwinship: This is the SDN equivalent of `ovnkube --net-controller`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,478 @@
...
+      # Requires fairly broad permissions - ability to read all services and network functions as well
+      # as all pods.
+      serviceAccountName: ovn
+      hostNetwork: true
+      hostPID: true
+      #tolerations:
+      #- key: node-role.kubernetes.io/master
+      # effect: Schedule
+      containers:
+      # firewall rules for ovn - assumed to be setup
```
danwinship: Nope. You can't assume that. If you don't set them up, they won't be set up.

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -150,6 +150,10 @@ type OVNKubernetesConfig struct {
 	// The MTU to use for the tunnel interface
 	// Default is 1400
 	MTU *uint32 `json:"mtu,omitempty"`
+
+	// If true, then assume the nodes already have a running openvswitch.
+	// +optional
+	UseExternalOpenvswitch *bool `json:"useExternalOpenvswitch,omitempty"`
```
danwinship: do we support that for OVN?

pecameron:
@danwinship Thanks for taking a look through this start on OVN. I will be integrating your suggestions in the next pass at this.

bindata/network/ovn-kubernetes/001-crd.yaml
```diff
@@ -0,0 +1,191 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: clusternetworks.network.openshift.io
```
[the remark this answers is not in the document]
pecameron: @danwinship I assume OVN will need at least some of this. I will need some help figuring out what is actually needed.

bindata/network/ovn-kubernetes/002-rbac.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - egressnetworkpolicies
+  - hostsubnets
+  - netnamespaces
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: [""]
+  resources:
+  - namespaces
```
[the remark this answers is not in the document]
pecameron: OK, I will need some help here as well.

bindata/network/ovn-kubernetes/controller.yaml
pecameron: @danwinship ovnkube-master.yaml and ovnkube.yaml are copied from ovn-kubernetes and have not yet be edited. I am not sure how we will keep the daemonsets in sync between the operator and ovn-kubernetes.

bindata/network/ovn-kubernetes/controller.yaml
pecameron: Deleting the file.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: The operator will need to set up the firewall on the host.

pkg/apis/networkoperator/v1/networkconfig_types.go
pecameron: @danwinship Why do we support it for sdn? Do the same reasons apply for OVN?

bindata/network/ovn-kubernetes/001-crd.yaml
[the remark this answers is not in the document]
danwinship: ovn-kubernetes does not need any of these types. For most of what's in ClusterNetwork, HostSubnet, and NetNamespace, ovn-kubernetes uses command-line flags and information in the OVN database. Egress IPs and EgressNetworkPolicy are not currently supported in ovn-kubernetes, and when they are, they won't be implemented in exactly the same way as in openshift-sdn.

pkg/apis/networkoperator/v1/networkconfig_types.go
danwinship: I think we support it for OpenShift SDN because some people are using DPDK and so need to be running the DPDK-ified OVS. But I wasn't really asking "*should* we support it with ovn-kubernetes?" so much as "*can* we support it with ovn-kubernetes?", or are OVN and ovn-kubernetes more tightly bound than OVS and OpenShift SDN are?

At the least it would presumably have to be "UseExternalOVN" rather than "UseExternalOpenvswitch", since if the user is going to replace OVS, they need to ensure they also run the version of OVN that goes with it, not necessarily the version of OVN that we're shipping.

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,161 @@
+# We need to create this because the controller expects it to exist, and we're
+# the first openshift-specific code running. Other code will manipulate the
```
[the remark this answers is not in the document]
squeed: This object can be dropped; it was just for openshift-sdn controller.

bindata/network/ovn-kubernetes/001-crd.yaml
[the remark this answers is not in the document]
bmeng: If so, do we have a way to get the networking info on the cluster which is running with ovn? Eg, the cluster subnet range, the service subnet range, etc...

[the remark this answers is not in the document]
danwinship: You'll be able to get them from the network operator's config object. ClusterNetwork is basically going to be totally redundant in 4.0, but we're keeping it around for compatibility.

README.md
```diff
@@ -6,7 +6,7 @@ It follows the [Controller pattern](https://godoc.org/github.com/kubernetes-sigs
 
 Most users will be able to use the top-level OpenShift Config API, which has a [Network type](https://github.com/openshift/api/blob/master/config/v1/types_network.go#L26). The operator will automatically translate the `Network.config.openshift.io` object in to a `NetworkConfig.networkoperator.openshift.io`.
 
-When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
+When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN or OVN plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
```
danwinship: Let's say "an installed network plugin".

In general, we want the documentation to be generic. We don't want to say "OpenShift SDN or ovn-kubernetes" all over the place, because if we do that, then when we add Kuryr support we're going to have to say "OpenShift SDN, ovn-kubernetes, or Kuryr" in a bunch of those places, etc, etc, etc.

README.md
```diff
@@ -43,9 +43,25 @@ spec:
   serviceNetwork:
   - 172.30.0.0/16
 ```
+Alternatively, OVN can be configured as follows:
```
danwinship: I don't think it's useful to repeat the entire config object with only one line changed. Just say something like "Alternatively, OVN could be chosen by specifying `networkType: OVNKubernetes`". (And drop the converted operator config below entirely.)

Although also, we need to figure out what the deal is with README vs Tech Preview-ness, etc. We don't want too much OVN-pimping at this point.

I guess Kuryr is going to have the same problem... maybe we should have `README.OVN.md` and `README.Kuryr.md`? Then we can go into as much detail as we want without derailing the main "narrative".

README.md
```diff
@@ -43,9 +43,25 @@ spec:
...
+    hostPrefix: 23
+  networkType: OVNKubernetes
+  serviceNetwork:
+  - 172.30.0.0/16
+```
 
 *Corresponding Operator Config*
 This configuration is the auto-generated translation of the above Cluster configuration.
+Note: the hostPrefix is converted into hostSubnetLength (hostSubnetLength = 32 - hostPrefix).
+When OVNKubernetes is used it is converted back to hostPrefix format and appended to the cidr.
```
danwinship: Converting the configuration from the generic form into the form needed by the network plugins is what the CNO does. You don't need to call out specific examples. (Likewise with the similar hostSubnetLength comment a bit later.)

README.md
```diff
@@ -60,25 +76,41 @@ spec:
...
+Users must supply at least two address pools - one for pods, and one for services. These are the ClusterNetworks and ServiceNetwork parameter. Some network plugins, such as OpenShiftSDN, and OVNKubernetes, support multiple ClusterNetworks. All address blocks must be non-overlapping. You should select address pools large enough to fit your anticipated workload. Each pool must be able to hold 1 or more hostPrefix allocations.
 
 For future expansion, multiple `serviceNetwork` entries are allowed by the configuration but not actually supported by any network plugins. Supplying multiple addresses is invalid.
 
-Each `clusterNetwork` entry has an additional required parameter, `hostPrefix`, that specifies the address size to assign to assign to each individual node.  For example,
+Each `clusterNetwork` entry has an additional required parameter, `hostPrefix`, that specifies the address size to assign to each individual node.  For example,
```
danwinship: (This PR is not going to merge before 4.0 branches, so "bugfixes" like this and non-OVN-specific clarifications like the "hostPrefix allocations" comment above should be split off into a separate PR.)

README.md
```diff
@@ -60,25 +76,41 @@ spec:
...
+means nodes would get blocks of size `/23`, or 512 addresses. This is converted into hostSubnetLength in the networkoperator.
 
-IP address pools are always read from the Cluster configuration and propagated "downwards" in to the Operator configuration. Any changes to the Operator configuration will be ignored.
+IP address pools are always read from the Cluster configuration and propagated "downwards" into the Operator configuration. Any changes to the Operator configuration are ignored.
 
 Currently, changing the address pools once set is not supported. In the future, some network providers may support expanding the address pools.
 
 
-Example
+OpenShiftSDN or OVNKubernetes Example:
```
danwinship: We've already said that not all plugins support multiple cidrs. So just leave this as "Example"

README.md
```diff
@@ -117,8 +149,25 @@ spec:
...
+spec:
+  defaultNetwork:
+    type: OVNKubernetes
+    ovnKubernetesConfig:
+      genevePort: 6081
+      mtu: 1400
+```
+
+## Configuring openshift-sdn kube-proxy
+Users may customize the openshift-sdn kube-proxy configuration. Ovn is a proxy and doesn't use kube-proxy. None of these settings are required:
```
danwinship: This isn't quite right. In theory the configuration here can be used for third-party plugins with a separate kube-proxy process as well, as discussed in the `deployKubeProxy` paragraph.

So we need to explain that
1. Some plugins (like OpenShift SDN) have a built-in kube-proxy, some plugins require a standalone kube-proxy to be deployed, and some (like OVN and Kuryr) don't use kube-proxy at all.
2. The `deployKubeProxy` flag can be used to indicate whether CNO should deploy a standalone kube-proxy, but for supported network types, this will default to the correct value automatically.
3. For plugins that use kube-proxy (whether built-in or standalone), you can configure the proxy via `kubeProxyConfig`

(If you split out a pre-4.0-fixes-and-clarifications-only PR, this change could be made there, since the idea of it is still correct even if we don't actually support OVN yet.)

README.md
```diff
@@ -146,15 +195,15 @@ You can view and edit that object with
...
 
 ## Unsafe changes
 Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested. 
 
 ### Safe changes to apply:
 It is safe to edit the following fields in `NetworkConfig.Spec`:
-* deployKubeProxy
-* all of kubeProxyConfig
+* deployKubeProxy (OpenShiftSDN only)
+* all of kubeProxyConfig (OpenShiftSDN only)
```
danwinship: drop this change

README.md
```diff
@@ -191,6 +240,9 @@ Making all defaults explicit makes it possible to prevent unsafe changes when a
 
 **Some values must be carried forward from the previous configuration.**
 
+The DefaultNetwork type must not be changed. Once OpenShiftSDN or OVNKubernetes is set it is permanent.
+The serviceNetwork must not be changed. Any existing clusterNetworks definitions must not be changed.
```
danwinship: This doesn't really fit in this subsection, and it isn't needed anyway. The docs already say previously that "most of the operator's configuration parameters are not changeable".

README.md
```diff
@@ -94,7 +126,7 @@ Users must select a default network provider. This cannot be changed. Different
 
 The network type is always read from the Cluster configuration.
 
-Currently, the only supported value for network Type is `OpenShiftSDN`.
+Currently, the only supported values for network Type are `OpenShiftSDN`, and `OVNKubernetes`.
```
danwinship: (This is an example of a place where we *do* want to enumerate the plugins, since we're specifically talking about what the supported types are.)
(But no comma before "and")

bindata/network/ovn-kubernetes/002-rbac.yaml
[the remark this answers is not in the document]
danwinship: From looking at the ovn-kube code, it looks like all of the kubernetes client calls are in `go-controller/pkg/kube/kube.go`:
```
44:	_, err := k.KClient.Core().Pods(pod.Namespace).Patch(pod.Name, types.MergePatchType, []byte(patchData))
55:	_, err := k.KClient.Core().Nodes().Patch(node.Name, types.MergePatchType, []byte(patchData))
70:	_, err := k.KClient.Core().Namespaces().Patch(ns.Name,
81:	pod, err := k.KClient.Core().Pods(namespace).Get(name, metav1.GetOptions{})
90:	return k.KClient.Core().Pods(namespace).Get(name, metav1.GetOptions{})
95:	return k.KClient.Core().Pods(namespace).List(metav1.ListOptions{})
103:	return k.KClient.Core().Pods(namespace).List(options)
108:	return k.KClient.Core().Nodes().List(metav1.ListOptions{})
113:	return k.KClient.Core().Nodes().Get(name, metav1.GetOptions{})
118:	return k.KClient.Core().Services(namespace).Get(name, metav1.GetOptions{})
124:	return k.KClient.Core().Endpoints(namespace).List(metav1.ListOptions{})
130:	return k.KClient.Core().Namespaces().Get(name, metav1.GetOptions{})
135:	return k.KClient.Core().Namespaces().List(metav1.ListOptions{})
140:	return k.KClient.Networking().NetworkPolicies(namespace).List(metav1.ListOptions{})
```

so, eg, the `Pods(...).Patch(...)` call at line 44 is in `SetAnnotationOnPod()`, which is called from `addLogicalPort()` in `pkg/ovn/pods.go`, which is called from `WatchPods()` in `pkg/ovn/ovn.go`, which is called from `Run()`, which is called from `cmd/ovnkube/ovnkube.go` if `netController` is true. So it's called from the master/controller. So `003-rbac-controller.yaml` needs to include the "patch" verb for the "pod" resource (in the "" (core) apiGroup; everything here except NetworkPolicy is in the core apiGroup).

bindata/network/ovn-kubernetes/002-rbac.yaml
```diff
@@ -0,0 +1,150 @@
...
+  name: cluster-admin
+  kind: ClusterRole
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: ServiceAccount
+  name: ovn
+  namespace: openshift-ovn
+
+---
+# service to expose the ovn-master pod
```
danwinship: This doesn't belong in this file, which is otherwise all just permissions stuff

bindata/network/ovn-kubernetes/controller.yaml
danwinship: The details of the daemonsets presumably will not change much, so there will not need to be much "keeping in sync". And if there is, then there is. We need to have our own daemonset here, which is independent of the upstream one. It's a fork, not a copy.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
danwinship: If the operator needs to do that, then this PR needs to *make* the operator do that.
One way to do that would be to add an `initContainer` to the pod that runs the necessary iptables commands.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        ports:
+        - name: healthz
+          containerPort: 10257
+        # TODO: Temporarily disabled until we determine how to wait for clean default
```
danwinship: need to figure out what that's about

pkg/network/cluster_config.go
```diff
@@ -41,7 +41,7 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 		}
 		size, _ := cidr.Mask.Size()
 		// The comparison is inverted; smaller number is larger block
-		if cnet.HostPrefix < uint32(size) {
+		if cnet.HostPrefix <= uint32(size) {
```
danwinship: This was discussed and we decided it was correct as it is; you're allowed to have a cluster network that's only big enough for a single node.

pkg/network/cluster_config.go
```diff
@@ -59,14 +59,16 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
...
 	}
 
 	return nil
 }
 
-// MergeClusterConfig merges the cluster configuration in to the real
+// MergeClusterConfig merges the cluster configuration into the real
 // CRD configuration.
+//PHIL need to include the ovn version...
 func MergeClusterConfig(operConf *netopv1.NetworkConfigSpec, clusterConf configv1.NetworkSpec) {
```
danwinship: This shouldn't need any OVN-specific logic

pkg/network/ovn_kubernetes.go
```diff
@@ -0,0 +1,220 @@
...
+		sc.MTU = &mtu
+	}
+}
+
+func networkPluginName() string {
+	return "ovn-kubernetes"
+}
+
+//PHIL -- what is this?
+// controllerOVNConfig builds the contents of controller-config.yaml
```
danwinship: This creates the configuration file used by the openshift-sdn controller process. In the OVN daemonset, the config options will all come from environment variables whose values come from the ConfigMap, which is all set up in `renderOVNKubernetes` so there's no equivalent of this function. (Likewise with `nodeOVNConfig`.)

bindata/network/ovn-kubernetes/001-crd.yaml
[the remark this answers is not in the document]
squeed: Just to follow back on this, @bmeng, you can get this from the `Network.config.openshift.io` object.

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,92 @@
+---
+# The controller also requires a service account in openshift-infra
```
squeed: This ServiceAccount is not necessary - it is openshift-sdn specific.

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,92 @@
...
+  name: ovn-kubernetes-controller
+rules:
+- apiGroups: [""]
+  resources:
+  - namespaces
+  verbs:
+  - list
+  - get
+  - watch
+- apiGroups: ["network.openshift.io"]
```
squeed: this apigroup is not necessary.

bindata/network/ovn-kubernetes/003-rbac-controller.yaml
```diff
@@ -0,0 +1,92 @@
...
+  kind: ClusterRole
+  name: ovn-kubernetes-controller
+subjects:
+- kind: ServiceAccount
+  name: ovn-controller
+  namespace: openshift-ovn
+
+---
+# bind the ovn-kubernetes-controller role to the service-account in openshift-infra
+apiVersion: rbac.authorization.k8s.io/v1
```
squeed: remove this binding.

bindata/network/ovn-kubernetes/controller.yaml
```diff
@@ -0,0 +1,71 @@
...
+  selector:
+    matchLabels:
+      app: ovn-controller
+  template:
+    metadata:
+      labels:
+        app: ovn-controller
+    spec:
+      containers:
+      - name: ovn-controller
```
squeed: This is the openshift controller and should not be run.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+# it is run on the master node(s)
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: ovnkube-master
+  # namespace set up by install
+  namespace: openshift-ovn
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the ovn-kubernetes networking components.
```
squeed: Clarify that this is the controller.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+spec:
+  progressDeadlineSeconds: 600
+  replicas: 1
+  revisionHistoryLimit: 10
+  selector:
+    matchLabels:
+      name: ovnkube-master
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
```
squeed: remove the `max*` parameters and use the defaults.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+      # firewall rules for ovn - assumed to be setup
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6641 -j ACCEPT
+      # iptables -A OS_FIREWALL_ALLOW -p tcp -m state --state NEW -m tcp --dport 6642 -j ACCEPT
+      # ovs flow for ovn (geneve)
+      # /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+
+      # run-ovn-northd - v3
+      - name: run-ovn-northd
+        image: {{.OvnImage}}
+        imagePullPolicy: {{.OvnImagePolicy}} # IfNotPresent
```
squeed: Hmm, why is this configurable? The default `imagePullPolicy` should be fine.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+
+        command: ["/root/ovnkube.sh", "run-ovn-northd"]
+
+        securityContext:
+          runAsUser: 0
```
squeed: there is a bug with cri-o where `runAsUser: 0` doesn't work with `privileged: true` - remove this please.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
+        - mountPath: /etc/sysconfig/origin-node
+          name: host-sysconfig-node
+          readOnly: true
+        # Mount the entire run directory for socket access for Docker or CRI-o
+        # TODO: remove
+        - mountPath: /var/run
```
squeed: remove this mount - it better be not needed.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-run
+        # Run directories where we need to be able to access sockets
+        - mountPath: /var/run/dbus/
```
squeed: do we need this? 99% chance we don't.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-run-dbus
+          readOnly: true
+        # ovn db is stored in the pod in /etc/openvswitch
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
```
squeed: these two mountpaths are duplicated. Keep teh `/var/lib` one; storing data in `/etc` is deprecated.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/log/openvswitch/
+          name: host-var-log-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+       #  readOnly: false
+        - mountPath: /var/run/kubernetes/
```
squeed: Unlikely we need this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+        - mountPath: /var/run/dbus/
+          name: host-var-run-dbus
+          readOnly: true
+        # ovn db is stored in the pod in /etc/openvswitch
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/log/openvswitch/
```
squeed: Why do we need `/var/log/openvswitch`?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+       #  readOnly: false
+        - mountPath: /var/run/kubernetes/
+          name: host-var-run-kubernetes
+          readOnly: true
+        # We mount our socket here
+        - mountPath: /var/run/ovn-kubernetes
+          name: host-var-run-ovn-kubernetes
+        - mountPath: /etc/origin/node
+          name: host-origin-node
+        # CNI related mounts which we take over
```
squeed: Not needed; only the node process needs this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+        - mountPath: /host/opt/cni/bin
+          name: host-opt-cni-bin
+        - mountPath: /etc/cni/net.d
```
squeed: remove this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
+        - mountPath: /etc/sysconfig/origin-node
+          name: host-sysconfig-node
+          readOnly: true
+        # Mount the entire run directory for socket access for Docker or CRI-o
+        # TODO: remove
+        - mountPath: /var/run
```
squeed: Remove this

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-run
+        # Run directories where we need to be able to access sockets
```
squeed: remove this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+        - mountPath: /var/run/dbus/
+          name: host-var-run-dbus
+          readOnly: true
+        # ovn db is stored in the pod in /etc/openvswitch
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
```
squeed: remove this

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/log/openvswitch/
+          name: host-var-log-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+       #  readOnly: false
+        - mountPath: /var/run/kubernetes/
```
squeed: is this needed?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-var-run-kubernetes
+          readOnly: true
+        # We mount our socket here
+        - mountPath: /var/run/ovn-kubernetes
+          name: host-var-run-ovn-kubernetes
+        - mountPath: /etc/origin/node
+          name: host-origin-node
+        # CNI related mounts which we take over
+        - mountPath: /host/opt/cni/bin
```
squeed: remove this

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-opt-cni-bin
+        - mountPath: /etc/cni/net.d
```
squeed: remove this

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          name: host-etc-cni-netd
+        - mountPath: /var/lib/cni/networks/ovn-k8s-cni-overlay
```
squeed: remove this

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+      # end of container
+
+      # sb-ovsdb - v3
+      - name: sb-ovsdb
+        image: {{.OvnImage}}
+        imagePullPolicy: {{.OvnImagePolicy}} # IfNotPresent
+
+        command: ["/root/ovnkube.sh", "sb-ovsdb"]
+
+        securityContext:
```
squeed: Does this need to be privileged?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          runAsUser: 0
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
```
squeed: Please audit all the volume mounts... Only those absolutely necessary should be kept.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+        lifecycle:
+      # end of container
+
+      - name: ovnkube-master
+        image: {{.OvnImage}}
+        imagePullPolicy: {{.OvnImagePolicy}} # IfNotPresent
+
+        command: ["/root/ovnkube.sh", "ovn-master"]
+
+        securityContext:
```
squeed: I don't think this needs to be privileged.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,449 @@
...
+          runAsUser: 0
+          # Permission could be reduced by selecting an appropriate SELinux policy
+          privileged: true
+
+        volumeMounts:
```
squeed: again with the volume mount audit, please.

pkg/network/cluster_config.go
```diff
@@ -59,14 +59,16 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 	}
 
 	if clusterConfig.NetworkType == "" {
+		//PHIL OpenShiftSDN or OVNKubernetes
```
squeed: extra comment?

pkg/network/ovn_kubernetes.go
```diff
@@ -0,0 +1,220 @@
...
+
+	legacyconfigv1 "github.com/openshift/api/legacyconfig/v1"
+	cpv1 "github.com/openshift/api/openshiftcontrolplane/v1"
+	netv1 "github.com/openshift/cluster-network-operator/pkg/apis/networkoperator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/render"
+)
+
+// NodeNameMagicString is substituted at runtime for the
+// real nodename
+const NodeNameMagicString = "%%NODENAME%%"
```
squeed: Doubt this is needed.

pkg/network/ovn_kubernetes.go
squeed: yup, delete this function.

bindata/network/ovn-kubernetes/001-crd.yaml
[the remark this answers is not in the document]
danwinship: > @danwinship Where do the OVN objects go? the types.go has defs for both sdn and ovn.

The types in this file are not related to the network operator's configuration types. They are resource types which are needed by openshift-sdn itself. ovn-kubernetes does not require any CRDs to be created, so there is nothing to be done here for ovn-kubernetes.

README.md
danwinship: No, `hostSubnetLength` is *not* wrong here. We're not talking about OpenShift SDN's configuration, we're talking about the cluster network operator's configuration, and the cluster network operator's configuration uses a `hostSubnetLength` field, regardless of what plugin type you are using.

bindata/network/ovn-kubernetes/001-crd.yaml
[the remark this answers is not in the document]
dcbw: @pecameron ovn-kube doesn't need anything in this file; it can be removed.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,343 @@
...
+      # end of container
+
+      # sb-ovsdb - v3
+      - name: sb-ovsdb
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "sb-ovsdb"]
+
+        volumeMounts:
+        - mountPath: /etc/sysconfig/origin-node
```
dcbw: Pretty sure we don't need this one.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,343 @@
...
+          name: host-sysconfig-node
+          readOnly: true
+        # Mount the entire run directory for socket access for Docker or CRI-o
+        # TODO: remove
```
dcbw: can remove this commented out one too

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,343 @@
...
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
+        - mountPath: /var/run/kubernetes/
+          name: host-var-run-kubernetes
+#       - mountPath: /etc/origin/node
```
dcbw: Remove this one as well.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,343 @@
...
+      # end of container
+
+      # nb-ovsdb - v3
+      - name: nb-ovsdb
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "nb-ovsdb"]
+
+        volumeMounts:
+        - mountPath: /etc/sysconfig/origin-node
```
dcbw: Pretty sure we don't need this one.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
dcbw: I guess for now we can just have the ovnkube init script do this, since it's hostNetwork?

pecameron:
@dcbw I made the changes that we talked about. PTAL

dcbw:
Image references aren't accessible by CI because they aren't official, so I posted https://github.com/openshift/release/pull/3223 which hopefully gives us something we can use here.

pecameron:
@dcbw PTAL

pecameron:
@danwinship PTAL do I need to make more changes?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,291 @@
...
+          valueFrom:
+            configMapKeyRef:
+              name: ovn-config
+              key: svc_cidr
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_APISERVER
+          value: "{{.K8S_APISERVER}}"
```
danwinship: consider redoing this (everywhere) to match https://github.com/openvswitch/ovn-kubernetes/commit/b375c56 / https://github.com/openvswitch/ovn-kubernetes/commit/026dcbb for consistency with upstream (ie, put the value in the configmap and then read that from here)

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+        volumeMounts:
+        # ovn db is stored in the pod in /etc/openvswitch (default ovs tool location)
+        # and on the host in /var/lib/openvswitch/
+        - mountPath: /etc/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/lib/openvswitch/
+          name: host-var-lib-ovs
+        - mountPath: /var/run/openvswitch/
+          name: host-var-run-ovs
```
danwinship: (note that a bunch of the removed mountpoints here (eg, the CNI stuff in ovnkube-master) can be removed upstream as well)

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,291 @@
...
+
+      - name: ovnkube-master
+        image: {{.OvnImage}}
+
+        command: ["/root/ovnkube.sh", "ovn-master"]
+
+#       securityContext:
+#         runAsUser: 0
+          # Permission could be reduced by selecting an appropriate SELinux policy
+#         privileged: true
```
danwinship: don't leave stuff commented out like this; either it's in or it's out

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,291 @@
...
+        hostPath:
+          path: /var/lib/openvswitch
+      - name: host-var-run-ovs
+        hostPath:
+          path: /var/run/openvswitch
+      - name: host-var-run-kubernetes
+        hostPath:
+          path: /var/run/kubernetes
+      tolerations:
+      - operator: "Exists"
```
danwinship: Copy the values from `openshift-sdn/controller.yaml` here. We don't want to tolerate *all* taints.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -0,0 +1,237 @@
...
+        component: network
+        type: infra
+        openshift.io/component: network
+        beta.kubernetes.io/os: "linux"
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      # Requires fairly broad permissions - ability to read all services and network functions as well
+      # as all pods.
+      serviceAccountName: ovn
```
danwinship: needs to match `002-rbac.yaml`: `ovn-kubernetes-node`

pkg/network/ovn_kubernetes.go
```diff
@@ -0,0 +1,141 @@
...
+		ippools += fmt.Sprintf("%s/%d", net.CIDR, net.HostPrefix)
+	}
+	data.Data["OVN_cidr"] = ippools
+	data.Data["OVN_service_cidr"] = conf.ServiceNetwork
+
+	nodeCfg, err := nodeOVNConfig(conf)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build node config")
+	}
+	data.Data["NodeConfig"] = nodeCfg
```
danwinship: this is not used. drop it

bindata/network/ovn-kubernetes/000-ns.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: ovnkube.sh in the OVN image currently hardcodes this namespace name
+  name: openshift-ovn-kubernetes
```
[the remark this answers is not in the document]
pecameron: Changed the daemonset and deployment. Script change will be in openvswitch/ovn-kubernetes (not done yet)

bindata/network/ovn-kubernetes/002-rbac.yaml
```diff
@@ -0,0 +1,52 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: ovn-kubernetes-node
+  namespace: ovn-kubernetes
```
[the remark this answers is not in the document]
pecameron: Oops! Glad you spotted this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -0,0 +1,291 @@
...
+        component: network
+        type: infra
+        openshift.io/component: network
+        beta.kubernetes.io/os: "linux"
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      # Requires fairly broad permissions - ability to read all services and network functions as well
+      # as all pods.
+      serviceAccountName: ovn
```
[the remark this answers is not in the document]
pecameron: changed

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: OK - feels like rearranging deck chairs (its been both ways a few times over)

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: I need to get the numerous changes from here to ovn-kubernetes and cleanup how community does things. It will happen later. Both community and this need to work and being very similar is a help.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: deleted the comments

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: SND has "node-role.kubernetes.io/master" and "node.kubernetes.io/not-ready"
I changed master as above and node to just have "node.kubernetes.io/not-ready"

bindata/network/ovn-kubernetes/ovnkube-node.yaml
pecameron: Changed

pkg/network/ovn_kubernetes.go
pecameron: deleted line

pecameron:
@danwinship @dcbw PTAL

pecameron:
@danwinship Are there further changes needed in this PR? PTAL

pecameron:
@dcbw @danwinship This passes unit and verify and fails on images. I think we are waiting on other merges to get this to work.

dcbw:
@pecameron yes, it's waiting on @smarterclayton to update the image mirroring in openshift/release.

danwinship:
I pushed a bunch of fixups. Please review and squash.

pkg/network/cluster_config.go
```diff
@@ -59,13 +59,14 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 	}
 
 	if clusterConfig.NetworkType == "" {
-		return errors.Errorf("spec.networkType is required")
+		// the type is required: OpenShiftSDN (default) or OVNKubernetes
+		return errors.Errorf("spec.networkType is required, such as OpenShiftSDN or OVNKubernetes")
```
danwinship: we shouldn't mention OVNKubernetes in the error message if it's not even tech preview

dcbw:
flake: Pulling image "docker-registry.default.svc:5000/ci-op-sg81c58h/pipeline@sha256:dccc5ad1d0f7d37d305a1228d7d68fd94f3c5532c97cf87206f5959987b9a678" ...
error: error pulling image docker-registry.default.svc:5000/ci-op-sg81c58h/pipeline@sha256:dccc5ad1d0f7d37d305a1228d7d68fd94f3c5532c97cf87206f5959987b9a678: Get https://docker-registry.default.svc:5000/v2/ci-op-sg81c58h/pipeline/manifests/sha256:dccc5ad1d0f7d37d305a1228d7d68fd94f3c5532c97cf87206f5959987b9a678: Get https://docker-registry.default.svc:5000/openshift/token?account=serviceaccount&scope=repository%3Aci-op-sg81c58h%2Fpipeline%3Apull: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
2019/03/28 00:25:48 Ran for 5m17s
error: could not run steps: could not wait for build: the build cluster-network-operator failed with reason GenericBuildFailed: Generic Build failure - check logs for details.

dcbw:
Continuous docker registry flakes...

PR #47, opened 2018-12-04
Convert cluster.openshift.io/Network in to the NetworkConfig CRD; update Network status

squeed:
This adds a second controller that watches the cluster-level network configuration. If changes are made, it merges those in to (or creates) the CRD configuration.

When the changes are applied, the "real" controller then updates the Network status object.

### DIFF
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -14,3 +14,22 @@ spec:
     - name: v1
       served: true
       storage: true
+
+---
+
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: networks.config.openshift.io
+spec:
+  group: config.openshift.io
+  names:
+    kind: Network
+    listKind: NetworkList
+    plural: networks
+    singular: network
+  scope: Cluster
+  versions:
+  - name: v1
+    served: true
+    storage: true


danwinship:
But an admin can directly modify the CRD as well, right? (In fact, they have to, to configure some options). What happens if they modify the CRD to contradict the cluster config?

squeed:
@danwinship yeah, I was thinking about that. I can see two possibilities. Either we say "hey, don't do that!", or we remove those fields from the CRD.

Removing overlapping fields from the CRD seems like the right way to me. I don't expect us to be adding much more in the way of fields to the Cluster object (except maybe Isolation).

I'll update the PR (and simplify most of the controller loop) to do this. We'll have to keep the duplicate fields for now, and roll it out in stages.

FWIW, part of this PR makes it possible to get networking up with just the cluster Network object (by defaulting to NetworkPolicy mode for openshift-sdn.

squeed:
Okay, updated this. Now, if there are any fields that were improperly updated in the "downstream" object, revert those changes when we reconcile.

squeed:
Odd, operator logs seem fine.
/retest

squeed:
Hit another AWS quota issue: `Jan 07 21:30:41.565 W persistentvolume=pvc-11be8786-12c3-11e9-ab14-0a9830816660 error deleting EBS volume "vol-03b08c117479245b2": "RequestLimitExceeded: Request limit exceeded.\n\tstatus code: 503, request id: b4ada13a-d246-40a0-9f2e-c95ab2e81a1b"`

/retest

squeed:
Flake city.
/retest

squeed:
All green. @danwinship, would you mind reviewing?

pkg/network/cluster_config.go
```diff
@@ -0,0 +1,109 @@
...
+	// - they do not overlap and do not overlap with the service cidr
+	for _, cnet := range clusterConf.ClusterNetwork {
+		_, cidr, err := net.ParseCIDR(cnet.CIDR)
+		if err != nil {
+			return errors.Errorf("could not parse spec.clusterNetwork %s", cnet.CIDR)
+		}
+		size, _ := cidr.Mask.Size()
+		// The comparison is inverted; smaller number is larger block
+		if cnet.HostPrefix < uint32(size) {
+			return errors.Errorf("hostPrefix %d is larger than its cidr %s",
```
danwinship: To match origin's validation, this should be `<=` not `<`. Also it checks that `HostSubnetLength >= 2` aka `HostPrefix <= 30`

pkg/network/cluster_config.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// StatusFromConfig generates the cluster NetworkStatus from the currently applied configuration
+func StatusFromConfig(operConf *netopv1.NetworkConfigSpec) configv1.NetworkStatus {
+	status := configv1.NetworkStatus{
+		ServiceNetwork: []string{operConf.ServiceNetwork},
+		NetworkType:    string(operConf.DefaultNetwork.Type),
+	}
+
+	// Flip flip flip the bits
```
danwinship: should we fix the operator config to match the cluster config?

pkg/controller/networkconfig/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Merge the cluster config in to the downstream CRD
+	network.MergeClusterConfig(&cfg.Spec, clusterConf.Spec)
+	if reflect.DeepEqual(cfg.Spec, existing.Spec) {
+		return nil
+	}
+
+	// If there are changes to the "downstream" networkconfig, commit it back
+	// to the apiserver
+	log.Println("WARNING: NetworkConfig.networkoperator.openshift.io has fields being overwritten by Network.cluster.openshift.io configuration")
```
danwinship: name of the second api object is wrong; should be `Network.config.openshift.io`

pkg/controller/networkconfig/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	cfg.TypeMeta = metav1.TypeMeta{APIVersion: "networkoperator.openshift.io/v1", Kind: "NetworkConfig"}
```
danwinship: why is this needed?

pkg/controller/networkconfig/cluster.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to transmute operator config")
+	}
+	if err = apply.ApplyObject(context.TODO(), r.client, us); err != nil {
+		return errors.Wrapf(err, "could not apply (%s) %s/%s", cfg.GroupVersionKind(), cfg.GetNamespace(), cfg.GetName())
+	}
+	return nil
+}
+
+// ClusterNetworkStatus generates the cluster.openshift.io/v1/Network.Status
```
danwinship: **config**.openshift.io

pkg/controller/networkconfig/networkconfig_controller.go
```diff
@@ -163,6 +167,22 @@ func (r *ReconcileNetworkConfig) Reconcile(request reconcile.Request) (reconcile
...
 
+	// Update cluster.openshift.io/Network.Status
+	status, err := r.ClusterNetworkStatus(context.TODO(), instance)
+	if err != nil {
+		err = errors.Wrapf(err, "could not generate network status")
+		log.Println(err)
+		return reconcile.Result{}, err
+	}
+	if status != nil {
+		// Don't set the owner reference
```
danwinship: i don't know what this comment means

pkg/util/ip/addr.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type IPPool struct {
+	cidrs []net.IPNet
+}
+
+func (p *IPPool) Add(cidr net.IPNet) error {
+	for _, n := range p.cidrs {
+		if netsOverlap(n, cidr) {
+			return errors.Errorf("cidrs %s and %s overlap",
```
danwinship: CIDRs

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -0,0 +1,113 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// Fetch the NetworkConfig instance
+	instance := &configv1.Network{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
```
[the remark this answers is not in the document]
squeed: This comment is boilerplate automatically generated by the operator-sdk, FWIW. Since we set the ownerReference, the usual kubernetes garbage-collection rules apply.

However, we don't set the owner reference in this case, because I decided that we don't want to delete the CRD. This is only because the network is special, and deleting the network configuration is likely to be a serious error. In most cases, this is not expected behavior, and I could be convinced to change this.

I'll delete the misleading comment, though.

pkg/controller/networkconfig/cluster.go
danwinship: maybe this is one of those things where you're mutating an object you got from a cache rather than DeepCopying it and mutating the copy?

pkg/network/cluster_config.go
squeed: By my reading, it seems to allow a HostPrefix that is the same size as the CIDR? I'm looking [here](https://github.com/openshift/origin/blob/master/pkg/network/apis/network/validation/validation.go#L94)

squeed:
@danwinship Thanks for the excellent review. suggestions incorporated, PTAL.

pkg/controller/networkconfig/cluster.go
squeed: Probably not that, but something similar. The client always copies in to a receiving object, and all the types should be registered with the schema. Just confusing.

pkg/network/cluster_config.go
danwinship: huh... somehow i looked at that same code and got the wrong answer.

squeed:
@danwinship can I get a LGTM?

README.md
```diff
@@ -4,15 +4,44 @@ The Cluster Network Operator installs and upgrades the networking components on
...
-The network operator has a complex configuration, but most parameters have a sensible default.
+The network operator gets its configuration from two objects: the Cluster and the Operator configuration. Most users only need to create the Cluster configuration - the operator will generate its configuration automatically. If you need finer-grained configuration of your network, you will need to create both configurations.
+
+Any changes to the Cluster configuration are propagated down in to the Operator configuration.
+
+
+#### Configuration objects
+*Cluster config*
+- *Type Name*: `Network.config.openshift.io`
+- *Instance Name*: `cluster`
```
danwinship: (I assume this name matches what other people are doing?)

README.md
```diff
@@ -4,15 +4,44 @@ The Cluster Network Operator installs and upgrades the networking components on
...
+- *View Command*: `oc get Network.openshift.io cluster -oyaml`
```
danwinship: Network.**config**.openshift.io

README.md
```diff
@@ -4,15 +4,44 @@ The Cluster Network Operator installs and upgrades the networking components on
...
+  name: cluster
+spec:
+  serviceNetwork: ["172.30.0.0/16"]
+  clusterNetwork: 
+    - cidr: "10.128.0.0/14"
+      hostPrefix: 23 
+  networkType: OpenShiftSDN
+```
+
+*Operator Config*
```
danwinship: (Clarify that the operator config in this example would have been auto-generated. Or alternatively, add a non-auto-generatable line to it and then explain that everything except that one line was auto-generated.)

README.md
```diff
@@ -4,15 +4,44 @@ The Cluster Network Operator installs and upgrades the networking components on
...
 
+Most users will be able to use the top-level OpenShift Config API, which has a [Network type](https://github.com/openshift/api/blob/master/config/v1/types_network.go#L26). The operator will automatically translate the `Network.config.openshift.io` object in to a `NetworkConfig.networkoperator.openshift.io`.
+
 When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
 
 # Configuring
-The network operator has a complex configuration, but most parameters have a sensible default.
+The network operator gets its configuration from two objects: the Cluster and the Operator configuration. Most users only need to create the Cluster configuration - the operator will generate its configuration automatically. If you need finer-grained configuration of your network, you will need to create both configurations.
+
+Any changes to the Cluster configuration are propagated down in to the Operator configuration.
```
danwinship: State explicitly that in case of conflicts, the operator config will be updated to match the cluster config

README.md
```diff
@@ -30,9 +59,14 @@ spec:
 ## Configuring IP address pools
 Users must supply at least two address pools - one for pods, and one for services. These are the ClusterNetworks and ServiceNetwork parameter. Some network plugins, such as OpenShiftSDN, support multiple ClusterNetworks. All address blocks must be non-overlapping. You should select address pools large enough to fit your anticipated workload.
 
+For future expansion, multiple `serviceNetwork` entries are allowed by the configuration but not actually supported by any network plugins. Supplying multiple addresses is invalid.
+
+Each `clusterNetwork` entry has an additional required parameter, `hostPrefix`, that specifies the address size to assign to assign to each individual node.  For example
```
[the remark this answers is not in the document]
danwinship: so either add the example or remove "For example"

README.md
```diff
@@ -30,9 +59,14 @@ spec:
...
+
+IP address pulls are always read from the Cluster configuration and propagated "downwards" in to the Operator configuration. Any changes to the Operator configuration will be ignored.
```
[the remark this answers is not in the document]
danwinship: fix that too

README.md
```diff
@@ -4,15 +4,44 @@ The Cluster Network Operator installs and upgrades the networking components on
...
+#### Example configurations
 
+*Cluster Config*
+```yaml
+apiVersion: config.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+  serviceNetwork: ["172.30.0.0/16"]
```
danwinship: maybe use the expanded syntax for parallel-ness with `clusterNetwork`? (And to match how `oc get` will display it.)

    serviceNetwork:
      - "172.30.0.0/16"
    clusterNetwork:
      - cidr: "10.128.0.0/14"
        hostPrefix: 23

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -0,0 +1,113 @@
...
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile propagates changes from the cluster config to the operator config.
+// In other words, it watches Network.config.openshift.io/v1/cluster and updates
+// NetworkConfig.networkoperator.openshift.io/v1/default.
+func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Printf("Reconciling Network %s/%s\n", request.Namespace, request.Name)
```
danwinship: "Network" here is ambiguous. Either `Network.config.openshift.io` or `cluster config`.

Also, it's non-namespaced so don't print the namespace. (That may apply to ReconcileNetworkConfig too?)

squeed: Good point; fixed (for both)

README.md
squeed: There, updated both to match `oc get` exactly (sans metadata).

README.md
squeed: Yup. Once the installer switches from `NetworkConfig` to `Network`, we can align the names if we like.

squeed:
@danwinship doc nits and log lines fixed, thanks.

PR #48, opened 2018-12-05
 openshift-sdn: Infer MTU from the machine on which we run 

squeed:
When the operator first runs, determine the mtu (by using the interface to the apiserver), then set the vxlan mtu to that less overhead.

Persist this to previous, and use that value for all subsequent runs.

### DIFF
--- a/pkg/network/mtu.go
+++ b/pkg/network/mtu.go
@@ -1,0 +1,42 @@
+package network
+
+import (
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// GetDefaultMTU gets the mtu of the default route.
+func GetDefaultMTU() (int, error) {
+	// Get the interface with the default route
+	// TODO(cdc) handle v6-only nodes
+	routes, err := netlink.RouteList(nil, netlink.FAMILY_V4)
+	if err != nil {
+		return 0, errors.Wrapf(err, "could not list routes")
+	}
+	if len(routes) == 0 {
+		return 0, errors.Errorf("got no routes")
+	}
+
+	const maxMTU = 65536
+	mtu := maxMTU + 1
+	for _, route := range routes {
+		// Skip non-default routes
+		if route.Dst != nil {
+			continue
+		}
+		link, err := netlink.LinkByIndex(route.LinkIndex)
+		if err != nil {
+			return 0, errors.Wrapf(err, "could not retrieve link id %d", route.LinkIndex)
+		}
+
+		newmtu := link.Attrs().MTU
+		if newmtu > 0 && newmtu < mtu {
+			mtu = newmtu
+		}
+	}
+	if mtu > maxMTU {
+		return 0, errors.Errorf("unable to determine MTU")
+	}
+
+	return mtu, nil
+}
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -105,7 +105,9 @@ func isOpenShiftSDNChangeSafe(prev, next *netv1.NetworkConfigSpec) []error {
 	return []error{errors.Errorf("cannot change openshift-sdn configuration")}
 }
 
-func fillOpenShiftSDNDefaults(conf *netv1.NetworkConfigSpec) {
+func fillOpenShiftSDNDefaults(conf, previous *netv1.NetworkConfigSpec, hostMTU int) {
+	// NOTE: If you change any defaults, and it's not a safe chang to roll out
+	// to existing clusters, you MUST use the value from previous instead.
 	if conf.DeployKubeProxy == nil {
 		prox := false
 		conf.DeployKubeProxy = &prox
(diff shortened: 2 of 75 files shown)


jeremyeder:
Awesome - needs a small docs fix for the existing MTU topic https://docs.okd.io/latest/scaling_performance/network_optimization.html#scaling-performance-optimizing-mtu

pkg/network/openshift_sdn_test.go
```diff
@@ -111,7 +112,7 @@ func TestFillOpenShiftSDNDefaults(t *testing.T) {
 	// vars
 	f := false
 	p := uint32(4789)
-	m := uint32(1450)
+	m := uint32(8050)
```
jeremyeder: Is this a typo?

squeed: nope; just testing that a theoretical mtu of 9000 is handled correctly.

squeed:
e2e failure doesn't look like a flake, investigating.

pkg/network/openshift_sdn_test.go
squeed: er, yes, it was a typo. should be 8950

pkg/network/mtu.go
```diff
@@ -0,0 +1,38 @@
...
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// GetDefaultMTU gets the mtu of the default route.
+func GetMTUTo(dest net.IP) (int, error) {
+	// Get the interface with the default route
+	routes, err := netlink.RouteGet(dest)
```
danwinship: If you actually want the interface with the default route, you can find that without needing a `dest`; just get the list of all routes, and find the default one. There is code to do this [in openshift](https://github.com/openshift/origin/blob/master/pkg/network/node/pod.go#L367).

In theory, someone could have routing set up such that the default route is not actually the route to the apiserver, but in theory, they could also have it set up that the route to the apiserver is not the route to the nodes. They can manually fill in the MTU in that case.

pkg/network/openshift_sdn.go
```diff
@@ -123,8 +123,15 @@ func fillOpenShiftSDNDefaults(conf *netv1.NetworkConfigSpec) {
 		var port uint32 = 4789
 		sc.VXLANPort = &port
 	}
+
+	// MTU is special. If it's not supplied, we infer it from  the node on which
```
danwinship: Maybe rather than being "special", there should be separate methods for "defaults" and "inferred values"?

pkg/network/openshift_sdn_test.go
```diff
@@ -47,7 +48,7 @@ func TestRenderOpenShiftSDN(t *testing.T) {
 
 	errs := validateOpenShiftSDN(config)
 	g.Expect(errs).To(HaveLen(0))
-	FillDefaults(config)
+	FillDefaults(config, nil, net.IPv4(4, 4, 4, 4))
```
danwinship: I assume none of the tests that are calling FillDefaults() actually look at the filled-in MTU, and so they aren't going to suddenly fail if the configuration of the CI machines changes?

pkg/network/mtu.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"net"
+
+	"github.com/pkg/errors"
+	"github.com/vishvananda/netlink"
+)
+
+// GetDefaultMTU gets the mtu of the default route.
+func GetMTUTo(dest net.IP) (int, error) {
```
JacobTanenbaum: nit: dockblock function name mismatch

pkg/network/openshift_sdn.go
squeed: I think you're right. We should even just carry filled defaults from `previous` in all cases.

pkg/network/mtu.go
squeed: Yeah, I was indeed thinking of that case, but you'r right - that's overthinking it. I'll switch it to the default route.

pkg/network/openshift_sdn.go
squeed: After a bit more thought, I think it makes sense to leave it more-or-less as-is. In a situation where the default changes, I could imagine keeping previous vs. using the new default needs to be done on a case-by-case basis.

pkg/network/openshift_sdn_test.go
squeed: Good thought, and, yes :-)

squeed:
OK, updated, PTAL.

danwinship:
looks good but the commit message still refers to apiserver

squeed:
Hm. Not sure what's up with those test failures; they're complaining that the certificate for the CI server's registry isn't valid. The cluster is up and running well (with a 9001 mtu!)

pkg/network/mtu.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			return 0, errors.Wrapf(err, "could not retrieve link id %d", route.LinkIndex)
+		}
+
+		newmtu := link.Attrs().MTU
+		if newmtu > 0 && newmtu < mtu {
+			mtu = newmtu
+		}
+	}
+	if mtu == 65537 {
```
knobunc: Should this be a constant both places use?

pkg/network/openshift_sdn.go
```diff
@@ -123,8 +125,15 @@ func fillOpenShiftSDNDefaults(conf *netv1.NetworkConfigSpec) {
 		var port uint32 = 4789
 		sc.VXLANPort = &port
 	}
+
+	// MTU is currently the only field we pull from previous.
+	// If it's not supplied, we infer it from  the node on which we're running.
+	// However, this can never change, so we always prefer previous.
 	if sc.MTU == nil {
-		var mtu uint32 = 1450
+		var mtu uint32 = uint32(hostMTU) - 50 // 50 byte overhead
```
knobunc: Nit: Can you say '50 byte overhead for VxLAN header' or something

danwinship: Nit nit: "VXLAN"

squeed:
Updated, thanks for the feedback.

PR #49, opened 2018-12-17
 openshift-sdn: set dnsip

squeed:
This is the IP address of the kube-dns server. Since we aren't using dnsmasq, we need to explicitly tell openshift-sdn about this.

### DIFF
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -3,10 +3,12 @@ package network
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"reflect"
 
+	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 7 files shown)


squeed:
Test failure is a flake: `level=error msg="\t* aws_vpc_endpoint.s3: Error creating VPC Endpoint: VpcEndpointLimitExceeded: The maximum number of VPC endpoints has been reached."`

pkg/network/openshift_sdn.go
```diff
@@ -231,3 +238,19 @@ func nodeConfig(conf *netv1.NetworkConfigSpec) (string, error) {
 	buf, err := yaml.Marshal(result)
 	return string(buf), err
 }
+
+// clusterDNSIP relies on the fact that the serviceIP for kube-dns is always the
+// 10th IP in the service cidr. This is used by the SDN in a very specific case:
+// the route is hard-coded so that pods with a macvlan interface never lose
+// access to DNS.
```
danwinship: Wait, what?
1. You totally can't just assume that.
2. If you do assume that, then you could just as easily assume it in openshift-sdn directly rather than assuming it here, writing the value out to a config file, and having openshift-sdn read that config file.
3. Except that the macvlan code already special-cases the entire service CIDR anyway, so if the DNS IP is a service CIDR then we don't need to add a special route for it because it's already covered by the existing service CIDR route.

As mentioned in the other bug, I think we can safely get rid of the DNS IP code entirely. If we don't, then it's only useful in the case where DNS IP is *not* an expected value, and so you need to get the actual value from the same place as the code that generates the pod's resolv.conf.

squeed: yeah yeah, you're right.
So, @ironcladlou, is there a way for the network operator to determine the dns server's IP before the cluster-dns-operator has run?

squeed:
Update: this is totally not needed. Thanks for all the feedback.

PR #50, opened 2018-12-17
readme: lots more documentation

squeed:


README.md
```diff
@@ -1,52 +1,202 @@
...
+* all of kubeProxyConfig
+
+### Force-applying an unsafe change
+Administrators may wish to forcefully apply a disruptive change to a cluster that is not serving production traffic. To do this, first they should make the desired configuration change to the CRD. Then, delete the network operator's understanding of the state of the system:
+
+```
+kubectl -n openshift-cluster-network-operator delete configmap applied-default
+```
+
+Be warned: this is an unsafe operation! It may cause the entire cluster to lose connectivity and all pods to be rescheduled.
```
danwinship: It might also not work. Like, if you change the serviceNetwork but don't delete all existing services

README.md
```diff
@@ -1,52 +1,202 @@
...
+```
+
+## Running manually against a test cluster
+If you have a running cluster, you can run the operator manually. Just set the `KUBECONFIG` environment variable.
+
+Note that, like all operators installed by the Cluster Version Operator, it determines all pointers to dependent images via environment variables. You will have to set several environment varables to emulate how the CVO works.
+
+You can determine the current values of these environment variables by inspecting the valid deployed daemonset. Do this before you delete it...
+```sh
+kubectl get -n openshift-cluster-network-operator daemonset cluster-network-operator -ojson | jq -Mrj '.spec.template.spec.containers[0].env[] | select(has("value")) | @text "\(.name)=\(.value) "' | tee images
```
danwinship: you can do that with just kubectl and JSON Template...

also, should we be using `oc` rather than `kubectl` in this doc? Not sure what our style guide would say...

README.md
```diff
@@ -1,52 +1,202 @@
 # Cluster Network Operator
 
-This is an operator that manages the networking components for an openshift cluster.
+The Cluster Network Operator installs and upgrades the networking components on a Kubernetes cluster.
 
+It follows the [Controller pattern](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg#hdr-Controller): it reconciles the state of the cluster against a desired configuration. The configuration specified by a CustomResourceDefinition called `networkoperator.openshift.io/NetworkConfig/v1`, which has a corresponding [type](/openshift/cluster-network-operator/blob/master/pkg/apis/networkoperator/v1/networkconfig_types.go).
 
-## Building
+When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN plugin and a working service network. It normally runs very early in the install process - before the permanent API server has been started.
```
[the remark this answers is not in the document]
squeed: That is a better way of wording it. FYI, the network operator does run using the bootstrap api server (the apiserver running on the bootstrap node). The switch over to the real apiserver happens afterwards, even though the apiserver is host network.

README.md
```diff
@@ -1,52 +1,202 @@
...
-```
-oc create -f manifests/
-oc create -f sample-config.yaml
-```
+Currently, changing the address pools once set is not supported. In the future, some network providers may support expanding the address pools.
 
-Then you can watch the daemonset with `kubectl -n openshift-cluster-network-operator logs <PODID>`.
-
-## Example configuration
+Each ClusterNetwork entry has an additional required parameter, `hostSubnetLength`, that specifies the address size to assign to assign to each individual node. Note that this is currently *reverse* from the usual CIDR notation - a hostSubnetLength of 9 means that the node will be assigned a /23. It is recommended, but not required, to use the same hostSubnetLength for all entries.
```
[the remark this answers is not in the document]
squeed: I thought I'd read that somewhere. Now I can't find it.

README.md
```diff
@@ -1,52 +1,202 @@
 # Cluster Network Operator
 
-This is an operator that manages the networking components for an openshift cluster.
+The Cluster Network Operator installs and upgrades the networking components on a Kubernetes cluster.
 
+It follows the [Controller pattern](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg#hdr-Controller): it reconciles the state of the cluster against a desired configuration. The configuration specified by a CustomResourceDefinition called `networkoperator.openshift.io/NetworkConfig/v1`, which has a corresponding [type](/openshift/cluster-network-operator/blob/master/pkg/apis/networkoperator/v1/networkconfig_types.go).
```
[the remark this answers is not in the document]
squeed: Somewhat, yes.

README.md
```diff
@@ -1,52 +1,202 @@
...
+Users must select a default network provider. This cannot be changed. Different network providers have additional provider-specific settings.
+
+Currently, the only supported value for network Type is `OpenShiftSDN`.
+
+### Configuring OpenShiftSDN
+OpenShiftSDN supports the following configuration options, all of which are optional:
+* `mode`: one of "Subnet" "Multitenant", or "NetworkPolicy". Configures the [isolation mode](https://docs.openshift.com/container-platform/3.11/architecture/networking/sdn.html#overview) for OpenShift SDN. The default is "NetworkPolicy".
+* `vxlanPort`: The port to use for the VXLAN overlay. The default is 4789
+* `MTU`: The MTU to use for the VXLAN overlay. The default is the MTU of the node, minus 50 bytes for overhead. However, this value will never change.
+* `useExternalOpenvswitch`: boolean. If the nodes are already running openvswitch, and OpenShiftSDN should not install its own, set this to true.
```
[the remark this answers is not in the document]
squeed: I see. Do we have any documentation for this?

README.md
```diff
@@ -1,52 +1,202 @@
...
+      mtu: 1450
+      useExternalOpenvswitch: false
+```
+
+## Configuring kube-proxy
+Users may customize the kube-proxy configuration. None of these settings are required:
+
+* `iptablesSyncPeriod`: The interval between periodic iptables refreshes. Default: 30 seconds
+* `bindAddress`: The address to "bind" to - the address for which traffic will be redirected
+* `proxyArguments`: additional command-line flags to pass to kube-proxy - see
```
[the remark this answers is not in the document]
squeed: Bah, we don't actually have a list of all of the possible iptables configuration flags. The only existing one is upstream: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/

README.md
```diff
@@ -1,52 +1,202 @@
...
+```
+
+and any changes will be automatically applied.
+
+## Unsafe changes
+Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested. 
+
+### Safe changes to apply:
+It is safe to edit the following fields in `NetworkConfig.Spec`:
+* deployKubeProxy
```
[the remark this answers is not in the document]
squeed: good catch

README.md
squeed: Even better.

README.md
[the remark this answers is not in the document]
danwinship: I dunno. I hadn't even realized there was such a config option before you added it. Google for matches to whatever ansible config option you based it on?

README.md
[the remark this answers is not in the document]
danwinship: Referencing upstream docs is fine, I think

README.md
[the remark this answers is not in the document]
squeed: There's no documentation for it anywhere. I'll try and chase some down. Flavio added it to openshift-ansible.

README.md
```diff
@@ -1,52 +1,205 @@
...
       mode: NetworkPolicy
+      vxlanPort: 4789
+      mtu: 1450
+      useExternalOpenvswitch: false
+```
+
+## Configuring kube-proxy
+Users may customize the kube-proxy configuration. None of these settings are required:
+
+* `iptablesSyncPeriod`: The interval between periodic iptables refreshes. Default: 30 seconds.
```
pecameron: What is the valid range? How does changing this impact the behavior of the cluster? Any special cases?

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+# Development
+
+## Basic structure
+
+The network operator consists of a controller loop and a rendering system. The basic flow is:
+
+1. The controller loop detects a configuration change
+1. The configuration is preprocessed:
+    - validity is checked.
```
pecameron: What happens if some of the tests pass and some fail?

README.md
```diff
@@ -1,52 +1,205 @@
...
+    - unspecified values are defaulted or carried forward from the previous configuration.
+    - safety of the proposed change is checked.
+1. The configuration is rendered in to a set of kubernetes objects (e.g. DaemonSets).
+1. The desired objects are reconciled against the API server, being created or updated as necessary.
+1. The applied configuration is stored separately, for later comparison.
```
pecameron: The CRD has the desired state (whether or not it makes sense), how do I find the actual state?

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+### Filling defaults
+Because most of the operator's configuration parameters are not changeable, it is important that the applied configuration is stable across upgrades. This has two implications:
+
+**All defaults must be made explicit.**
```
pecameron: Also, the defaults can't change from version to version.

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+The network configuration is transformed internally in to a fully-expressed struct. All optional values must have their default set. For example, if the vxlanPort is not specified, the default of 4789 is chosen and applied to the OpenShiftSDNConfig.
+
+Making all defaults explicit makes it possible to prevent unsafe changes when a newer version of the operator changes a default value.
```
pecameron: This should be at the top of the doc and reflected in the examples. Very important point!

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+**Some values must be carried forward from the previous configuration.**
+
+Some values are computed at run-time but can never be changed. For example, the MTU of the overlay network is determined from the node's interfaces. Changing this is unsafe, so this must always be carried forward.
+
+Note that the fully-expressed configuration is not persisted back to the apiserver. Instead, it is saved only in the stored applied configuration. An alternative would be to inject these values via a mutating webhook. That requires a running service network, which we don't have until after we've run.
```
pecameron: This seems dangerous. Customers will be working with this.

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+### Validation
+Each network provider is responsible for validating their view of the configuration. For example, the OpenShiftSDN provider validates that the vxlanPort is a number between 1 and 65535, that the MTU is sane, etc. No validation is provided via the CRD itself.
```
pecameron: How is a validation failure reported and handled? What does the user have to do to correct it?

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+To do this, just scale the CVO down to 0 replicas and delete the network-operator daemonset.
+
+```sh
+kubectl scale --replicas 0 -n openshift-cluster-version deployments/cluster-version-operator
+kubectl delete -n openshift-cluster-network-operator daemonset cluster-network-operator 
+```
+
+## Building images
+By default, podman is used to build images.
```
pecameron: Are the images pushed to some repo as part of the build? Where do we tell the daemonsets which image to use?

README.md
```diff
@@ -1,52 +1,205 @@
...
+A configuration with minimum parameters set:
 
-Or you could use a docker that supports multi-stage builds
-```
-BUILDCMD="docker build" ./hack/build-image.sh
+```yaml
+apiVersion: "networkoperator.openshift.io/v1"
+kind: "NetworkConfig"
+metadata:
+  name: "default"
```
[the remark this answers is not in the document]
squeed: it must be default.

README.md
```diff
@@ -1,52 +1,205 @@
 # Cluster Network Operator
 
-This is an operator that manages the networking components for an openshift cluster.
+The Cluster Network Operator installs and upgrades the networking components on an OpenShift Kubernetes cluster.
 
+It follows the [Controller pattern](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg#hdr-Controller): it reconciles the state of the cluster against a desired configuration. The configuration specified by a CustomResourceDefinition called `networkoperator.openshift.io/NetworkConfig/v1`, which has a corresponding [type](/openshift/cluster-network-operator/blob/master/pkg/apis/networkoperator/v1/networkconfig_types.go).
 
-## Building
+When the controller has reconciled and all its dependent resources have converged, the cluster should have an installed SDN plugin and a working service network. In OpenShift, the Cluster Network Operator runs very early in the install process -- while the boostrap API server is still running.
```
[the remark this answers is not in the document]
squeed: When we're done, updating will not be disruptive at all.

README.md
```diff
@@ -1,52 +1,205 @@
...
-```
-oc create -f manifests/
-oc create -f sample-config.yaml
-```
+Currently, changing the address pools once set is not supported. In the future, some network providers may support expanding the address pools.
 
-Then you can watch the daemonset with `kubectl -n openshift-cluster-network-operator logs <PODID>`.
-
-## Example configuration
+Each ClusterNetwork entry has an additional required parameter, `hostSubnetLength`, that specifies the address size to assign to assign to each individual node. Note that this is currently *reverse* from the usual CIDR notation - a hostSubnetLength of 9 means that the node will be assigned a /23.
```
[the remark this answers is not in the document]
squeed: Updated.

README.md
```diff
@@ -1,52 +1,202 @@
...
 spec:
   serviceNetwork: "172.30.0.0/16"
   clusterNetworks:
     - cidr: "10.128.0.0/14"
       hostSubnetLength: 9
+    - cidr: "192.168.0.0/18"
+      hostSubnetLength: 9
+```
+
+## Configuring the default network provider
```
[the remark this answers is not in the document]
squeed: Indeed - the multus documentation update will be separate.

README.md
```diff
@@ -1,52 +1,205 @@
...
+Users must select a default network provider. This cannot be changed. Different network providers have additional provider-specific settings.
```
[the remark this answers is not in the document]
squeed: Why? Its more than a plugin; the plugin provides connectivity but not reachability. Provider is the correct terminology here.

README.md
```diff
@@ -1,52 +1,202 @@
...
+
+Currently, the only supported value for network Type is `OpenShiftSDN`.
+
+### Configuring OpenShiftSDN
+OpenShiftSDN supports the following configuration options, all of which are optional:
+* `mode`: one of "Subnet" "Multitenant", or "NetworkPolicy". Configures the [isolation mode](https://docs.openshift.com/container-platform/3.11/architecture/networking/sdn.html#overview) for OpenShift SDN. The default is "NetworkPolicy".
+* `vxlanPort`: The port to use for the VXLAN overlay. The default is 4789
+* `MTU`: The MTU to use for the VXLAN overlay. The default is the MTU of the node, minus 50 bytes for overhead. However, this value will never change.
```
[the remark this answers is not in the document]
squeed: a good way to put it, thanks.

README.md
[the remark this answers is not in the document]
squeed: There is basically nothing. I added a bit of a disclaimer.

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+# Using
+The operator looks for a CRD of type `NetworkConfig` with the name `default`. Whenever this object is created or changed, the operator will apply those changes to the cluster (if they are safe to do so).
+
+You can view and edit that object with
+```
+kubectl edit networkconfig default
+```
+
+and any changes will be automatically applied.
```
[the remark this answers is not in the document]
squeed: We need to update the OperatorStatus object when things are invalid; that will be a later update.

README.md
```diff
@@ -1,52 +1,205 @@
...
+
+## Unsafe changes
+Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested.
```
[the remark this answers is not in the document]
squeed: nothing yet. this will be updated with OperatorStatus.

README.md
squeed: It's not so simple; you need to just inspect the cluster with `kubectl`

README.md
squeed: I'll have an answer when OperatorStatus is done.

README.md
squeed: mentioned in line 140

README.md
squeed: no, it's an implementation detail

README.md
squeed: which is why we're not doing it.

README.md
squeed: this will change with operatorstatus.

README.md
squeed: update: I just needed to actually think for a minute, and the jsonpath stuff worked great. And we should use OC.

README.md
squeed: Good question; I'm not sure that documenting the OpenShift build process is in-scope for this document. If you want to, go for it!

squeed:
@pecameron @danwinship thanks for the thorough review. Updated, PTAL.

README.md
```diff
@@ -1,52 +1,207 @@
...
+      hostSubnetLength: 9
+```
+
+## Configuring the default network provider
+Users must select a default network provider. This cannot be changed. Different network providers have additional provider-specific settings.
+
+Currently, the only supported value for network Type is `OpenShiftSDN`.
+
+### Configuring OpenShiftSDN
+OpenShiftSDN supports the following configuration options, all of which are optional:
```
danwinship: should clarify that these are under `defaultNetwork.openshiftSDNConfig`

README.md
```diff
@@ -1,52 +1,207 @@
...
+      mtu: 1450
+      useExternalOpenvswitch: false
+```
+
+## Configuring kube-proxy
+Users may customize the kube-proxy configuration. None of these settings are required:
+
+* `iptablesSyncPeriod`: The interval between periodic iptables refreshes. Default: 30 seconds. Increasing this can reduce the number of iptables invocations.
+* `bindAddress`: The address to "bind" to - the address for which traffic will be redirected.
+* `proxyArguments`: additional command-line flags to pass to kube-proxy - see the [documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/).
```
danwinship: should clarify how this works; `proxyArguments` is a `map[string][]string`, so the value has to be specified as an array, even though most arguments only accept a single value. (This might make more sense as a footnote to the example below.)

PR #51, opened 2019-01-04
hack: use local GOPATH if GOPATH isn't set

dcbw:
@squeed @danwinship finally got tired of docker container compiles

### DIFF
--- a/hack/build-go.sh
+++ b/hack/build-go.sh
@@ -2,35 +2,26 @@
 
 set -eu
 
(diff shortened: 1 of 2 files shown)


danwinship:
is this copied from somewhree else?

dcbw:
> is this copied from somewhree else?

@danwinship started from ovn-kubernetes and then tightened up and consolidated with bits Casey had in build-go.sh already.

PR #52, opened 2019-01-06
hack: add a test script

dcbw:
@squeed @danwinship mostly cribbed from ovn-kubernetes but significantly cleaned up since we don't need sudo.

### DIFF
--- a/hack/test-go.sh
+++ b/hack/test-go.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE}")/init.sh"
+
+# Check for `go` binary and set ${GOPATH}.
+setup_env
+
+cd ${GOPATH}/src/${CNO_GO_PKG}
+
+if [ -z "$PKGS" ]; then
+  # by default, test everything that's not in vendor
+  PKGS="$(go list ./... | grep -v vendor | xargs echo)"
+fi
+
+CGO_ENABLED=0 go test "${goflags[@]:+${goflags[@]}}" ${PKGS}
+retcode=$?
+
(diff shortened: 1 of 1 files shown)


hack/test-go.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ -z "$PKGS" ]; then
+  # by default, test everything that's not in vendor
+  PKGS="$(go list ./... | grep -v vendor | xargs echo)"
+fi
+
+CGO_ENABLED=0 go test "${goflags[@]:+${goflags[@]}}" ${PKGS}
+retcode=$?
+
+popd
+exit $retcode
```
danwinship: why popd and then exit? You could just change the "pushd" to a "cd"

dcbw: fixed.

dcbw:
@danwinship @squeed PTAL thanks

hack/test-go.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE}")/init.sh"
```
danwinship: /me just noticed for the first time that you can nest `""` inside `$()` inside `""`...

danwinship:
/lgtm
although I wouldn't object if you submitted another PR (or sniped the commitbot on this PR) to consistently use both `""` and `${}` around all variables. (What if I want my `GOPATH` to have a space in it?!)

hack/test-go.sh
dcbw: Yeah I noticed that this weekend, I was also surprised.

PR #54, opened 2019-01-08
multus: add support for the Multus daemonset

dcbw:
Multus will be installed by default but can be turned off by
setting the DisableMultiNetwork property of the Cluster Network Operator
object to 'true'.

[NOTE: we are waiting on https://github.com/openshift/release/pull/2570 to land, which will provide the official images for Multus. Once that lands, I will remove the "hack: no images for now" commit and update the image references.]


bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,58 @@
...
+        image: nfvpe/multus:latest
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "50Mi"
+          limits:
+            cpu: "100m"
+            memory: "50Mi"
+        env:
+        - name: MULTUS_CONF_FILE
```
dougbtv: Dan this might get clobbered by the entrypoint script. What I have been testing with to enable the auto functionality is to add args to the entrypoint, with:

```
        command: ["/entrypoint.sh"]
        args: ["--multus-conf-file=auto"]
```

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,58 @@
...
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      tolerations:
+      - operator: Exists
+      serviceAccountName: multus
+      containers:
+      - name: kube-multus
+        image: nfvpe/multus:latest
```
squeed: You probably know this, but the image url will need to be templated through like the others.

pkg/network/render.go
```diff
@@ -185,9 +213,9 @@ func RenderAdditionalNetworks(conf *netv1.NetworkConfigSpec, manifestDir string)
 		return nil, errors.Errorf("invalid Additional Network Configuration: %v", errs)
 	}
 
-	// render Multus when additional networks is provided
+	// render the CRD when additional networks are provided
```
squeed: As per our previous discussion, we need to render the CRD whenever Multus is enabled so administrators can create additional networks out-of-band. I suspect that Multus also needs the CRD to exist to prevent errors.

pkg/network/render.go
```diff
@@ -185,9 +213,9 @@ func RenderAdditionalNetworks(conf *netv1.NetworkConfigSpec, manifestDir string)
...
 	if len(ans) > 0 {
-		objs, err := renderMultusConfig(manifestDir)
+		objs, err := renderAdditionalNetworksCRD(manifestDir)
```
zshi-redhat: Rendering of CRD and CR may fail in a single run, some details [here](https://github.com/openshift/cluster-network-operator/pull/53#issuecomment-452312576)

pkg/network/render.go
```diff
@@ -41,6 +48,7 @@ func Validate(conf *netv1.NetworkConfigSpec) error {
 	errs := []error{}
 
 	errs = append(errs, ValidateDefaultNetwork(conf)...)
+	errs = append(errs, ValidateMultus(conf)...)
```
zshi-redhat: Shall we also validate additional networks here? If it's not provided, validation of additional network returns immediately.

bindata/network/multus/multus.yaml
dcbw: yep, hence the WIP :)

pkg/network/render.go
zshi-redhat: Shall we also capture config error that when additionalNetworks are provided but DeployMultus is set to false?

dcbw: I don't think Multus cares. It only looks for the pod annotations, and only then does it try to get the networks using the CRD. So it's not going to block pods that don't use multiple networks.

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,58 @@
...
+        component: network
+        type: infra
+        openshift.io/component: network
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      tolerations:
```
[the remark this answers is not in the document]
dcbw: @zshi-redhat multus will be configured on all nodes.

pkg/network/render.go
dcbw: @zshi-redhat no, the validate of additional networks would be handled by the additional network rendering, not multus rendering. Multus and AdditionalNetworks are completely separate codepaths that do not depend on each other.

pkg/network/render.go
dcbw: I was actually intending that to be supported, in case some 3rd party plugin also implements the CRD spec but does not use Multus. Again, I think AdditionalNetworks (eg the NPWG spec CRs) and Multus (an implementation of the NPWG spec) should be separate things.

zshi-redhat:
> > We probably want to make Multus conf file more configurable instead of using auto, reason being that there are some parameters that user may want to configure, for example, the namespaceIsolation, logLevel, logFile etc and in the future when enabling network migration with Multus, user may want to specify default network plugin although we may support configure default network with net-attach-def CR.
> 
> At this point, let's wait until we actually have use-cases for those things before we put them into the operator. I honestly don't think namespaceIsolation is worthwhile here, that's what the Admission Controller and RBAC is for. logLevel and logFile would be controlled by the Network operator when increased debugging is called for, but shouldn't be a user-visible option. Almost _everything_ that multus allows configuration for, should be handled automatically by the operator and not be a passthrough option to the user.

Ok, understood, Operator would be the only user-facing interface,  if there needs to be an user-configurable parameter to multus, it will be exposed in the operator level.

dcbw:
> Ok, understood, Operator would be the only user-facing interface, if there needs to be an user-configurable parameter to multus, it will be exposed in the operator level.

@zshi-redhat yeah, at least for now. Let's keep things simple because once we add something, we can't easily remove it :)

bindata/network/multus/multus.yaml
dcbw: @dougbtv yeah, nfvpe/multus needs a rebuild to pick up the "auto" functionality; it doesn't have it currently.

dcbw:
/retest

flake is during install:
```
time="2019-01-10T07:40:24Z" level=error
time="2019-01-10T07:40:24Z" level=error msg="Error: Error applying plan:"
time="2019-01-10T07:40:24Z" level=error
time="2019-01-10T07:40:24Z" level=error msg="1 error occurred:"
time="2019-01-10T07:40:24Z" level=error msg="\t* module.vpc.aws_route.to_nat_gw[4]: 1 error occurred:"
time="2019-01-10T07:40:24Z" level=error msg="\t* aws_route.to_nat_gw.4: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)"
```

dcbw:
@danwinship @squeed @knobunc @zshi-redhat @dougbtv review on this one would be great, thanks!

bindata/network/sriov/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,57 @@
...
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      tolerations:
+      - operator: Exists
+      serviceAccountName: sriov-device-plugin
+      containers:
+      - name: sriovdp
+        image: {{.SRIOVDevicePluginImage}}
+        args:
+        - --log-dir=sriovdp
```
zshi-redhat: please replace the args with "--log-level=10"

```
args:
- --log-level=10
```

--log-dir will cause failure when creating the log file since there is no log dir volume mounted in this daemonset.

manifests/0000_07_cluster-network-operator_03_daemonset.yaml
```diff
@@ -32,6 +32,18 @@ spec:
...
+        - name: NAD_ADMISSION_CONTROLLER_IMAGE
+          value: "docker.io/nfvpe/net-attach-def-admission-controller:snapshot"
+        - name: CNI_PLUGINS_SUPPORTED_IMAGE
+          value: "docker.io/nfvpe/container-networking-plugins-supported"
+        - name: CNI_PLUGINS_UNSUPPORTED_IMAGE
+          value: "docker.io/nfvpe/container-networking-plugins-unsupported"
+        - name: SRIOV_DEVICE_PLUGIN_IMAGE
+          value: "docker.io/nfvpe/sriov-device-plugin:release-v1"
+        - name: SRIOV_CNI_IMAGE
+          value: "docker.io/nfvpe/sriov-cni:latest"
```
zshi-redhat: let's use 'release-v1' tagged image for sriov-cni
```
value: "docker.io/nfvpe/sriov-cni:release-v1"
```

manifests/image-references
```diff
@@ -14,3 +14,27 @@ spec:
...
+      kind: DockerImage
+      name: docker.io/nfvpe/container-networking-plugins-unsupported
+  - name: sriov-device-plugin
+    from:
+      kind: DockerImage
+      name: docker.io/nfvpe/sriov-device-plugin:release-v1
+  - name: sriov-cni
+    from:
+      kind: DockerImage
+      name: docker.io/nfvpe/sriov-cni:latest
```
zshi-redhat: same here, use ``release-v1`` tagged image for ``sriov-cni``

pkg/network/render.go
```diff
@@ -16,7 +17,15 @@ func Render(conf *netv1.NetworkConfigSpec, manifestDir string) ([]*uns.Unstructu
...
-	o, err := RenderDefaultNetwork(conf, manifestDir)
+	o, pluginName, err := RenderDefaultNetwork(conf, manifestDir)
+	if err != nil {
+		return nil, err
+	}
+	objs = append(objs, o...)
+
+	// render Multus
+	var crdRendered bool
+	o, crdRendered, err = RenderMultus(conf, manifestDir, pluginName)
```
zshi-redhat: what about just passing ``conf`` to ``RenderMultus``, the pluginName can be derived from ``conf.DefaultNetwork.NetworkType``

dcbw: @zshi-redhat I'd rather keep the logic about what the CNI plugin name is alongside the plugin-specific logic

dcbw:
> Shall we also add SR-IOV admission-controller manifests in this PR?

@zshi-redhat yeah, probably

zshi-redhat:
Reconciling of ``(batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller`` failed with below log message, seems the admission-controller Job manifest doesn't specify matched ``selector`` and ``label``.

```
[core@ip-10-0-15-231 cluster-network-operator]$ ./_output/linux/amd64/cluster-network-operator -v 6 --url-only-kubeconfig=/etc/kubernetes/kubeconfig
2019/01/11 06:11:56 Go Version: go1.10.3
2019/01/11 06:11:56 Go OS/Arch: linux/amd64
2019/01/11 06:11:56 operator-sdk Version: v0.1.0
2019/01/11 06:11:56 overriding kubernetes api to https://multus-test-cluster-api.devcluster.openshift.com:6443
INFO[0000] trying to become the leader                  
INFO[0000] Skipping leader election; not running in a cluster 
2019/01/11 06:11:56 Registering Components.
2019/01/11 06:11:56 Starting the Cmd.
2019/01/11 06:11:56 Reconciling NetworkConfig /default
2019/01/11 06:11:56 Starting render phase
2019/01/11 06:11:56 Render phase done, rendered 46 objects
2019/01/11 06:11:56 reconciling (/v1, Kind=ConfigMap) openshift-cluster-network-operator/applied-default
2019/01/11 06:11:56 update was successful
2019/01/11 06:11:56 reconciling (/v1, Kind=Namespace) /openshift-sdn
2019/01/11 06:11:56 update was noop
2019/01/11 06:11:56 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /clusternetworks.network.openshift.io
2019/01/11 06:11:56 update was noop
2019/01/11 06:11:56 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /hostsubnets.network.openshift.io
2019/01/11 06:11:56 update was noop
2019/01/11 06:11:56 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /netnamespaces.network.openshift.io
2019/01/11 06:11:56 update was noop
2019/01/11 06:11:56 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /egressnetworkpolicies.network.openshift.io
2019/01/11 06:11:56 update was noop
2019/01/11 06:11:56 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /openshift-sdn
2019/01/11 06:11:57 update was noop
2019/01/11 06:11:57 reconciling (/v1, Kind=ServiceAccount) openshift-sdn/sdn
2019/01/11 06:11:57 update was successful
2019/01/11 06:11:57 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn
2019/01/11 06:11:58 update was noop
2019/01/11 06:11:58 reconciling (/v1, Kind=Namespace) /openshift-infra
2019/01/11 06:11:58 update was noop
2019/01/11 06:11:58 reconciling (/v1, Kind=ServiceAccount) openshift-infra/sdn-controller
2019/01/11 06:11:59 update was successful
2019/01/11 06:11:59 reconciling (/v1, Kind=ServiceAccount) openshift-sdn/sdn-controller
2019/01/11 06:11:59 update was successful
2019/01/11 06:11:59 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /openshift-sdn-controller
2019/01/11 06:11:59 update was noop
2019/01/11 06:11:59 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn-controller
2019/01/11 06:12:00 update was noop
2019/01/11 06:12:00 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn-controller-infra
2019/01/11 06:12:00 update was noop
2019/01/11 06:12:00 reconciling (rbac.authorization.k8s.io/v1, Kind=Role) openshift-sdn/openshift-sdn-controller-leaderelection
2019/01/11 06:12:01 update was noop
2019/01/11 06:12:01 reconciling (rbac.authorization.k8s.io/v1, Kind=RoleBinding) openshift-sdn/openshift-sdn-controller-leaderelection
2019/01/11 06:12:01 update was noop
2019/01/11 06:12:01 reconciling (rbac.authorization.k8s.io/v1, Kind=Role) openshift-infra/openshift-sdn-controller-account
2019/01/11 06:12:01 update was noop
2019/01/11 06:12:01 reconciling (rbac.authorization.k8s.io/v1, Kind=RoleBinding) openshift-infra/openshift-sdn-controller-account
2019/01/11 06:12:02 update was noop
2019/01/11 06:12:02 reconciling (/v1, Kind=ConfigMap) openshift-sdn/sdn-controller-config
2019/01/11 06:12:02 update was noop
2019/01/11 06:12:02 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/sdn-controller
2019/01/11 06:12:02 update was successful
2019/01/11 06:12:02 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/ovs
2019/01/11 06:12:03 update was successful
2019/01/11 06:12:03 reconciling (/v1, Kind=ConfigMap) openshift-sdn/sdn-config
2019/01/11 06:12:03 update was noop
2019/01/11 06:12:03 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/sdn
2019/01/11 06:12:04 update was successful
2019/01/11 06:12:04 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /network-attachment-definitions.k8s.cni.cncf.io
2019/01/11 06:12:04 does not exist, creating (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /network-attachment-definitions.k8s.cni.cncf.io
2019/01/11 06:12:04 successfully created (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /network-attachment-definitions.k8s.cni.cncf.io
2019/01/11 06:12:04 reconciling (/v1, Kind=Namespace) /multus
2019/01/11 06:12:04 does not exist, creating (/v1, Kind=Namespace) /multus
2019/01/11 06:12:04 successfully created (/v1, Kind=Namespace) /multus
2019/01/11 06:12:04 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /multus
2019/01/11 06:12:05 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /multus
2019/01/11 06:12:05 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /multus
2019/01/11 06:12:05 reconciling (/v1, Kind=ServiceAccount) multus/multus
2019/01/11 06:12:05 does not exist, creating (/v1, Kind=ServiceAccount) multus/multus
2019/01/11 06:12:05 successfully created (/v1, Kind=ServiceAccount) multus/multus
2019/01/11 06:12:05 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /multus
2019/01/11 06:12:05 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /multus
2019/01/11 06:12:06 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /multus
2019/01/11 06:12:06 reconciling (/v1, Kind=ServiceAccount) kube-system/net-attach-def-admission-controller-sa
2019/01/11 06:12:06 does not exist, creating (/v1, Kind=ServiceAccount) kube-system/net-attach-def-admission-controller-sa
2019/01/11 06:12:06 successfully created (/v1, Kind=ServiceAccount) kube-system/net-attach-def-admission-controller-sa
2019/01/11 06:12:06 reconciling (/v1, Kind=Secret) kube-system/net-attach-def-admission-controller-sa-secret
2019/01/11 06:12:06 does not exist, creating (/v1, Kind=Secret) kube-system/net-attach-def-admission-controller-sa-secret
2019/01/11 06:12:06 successfully created (/v1, Kind=Secret) kube-system/net-attach-def-admission-controller-sa-secret
2019/01/11 06:12:06 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller
2019/01/11 06:12:07 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller
2019/01/11 06:12:07 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller
2019/01/11 06:12:07 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-certificates
2019/01/11 06:12:07 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-certificates
2019/01/11 06:12:07 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-certificates
2019/01/11 06:12:07 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-secrets
2019/01/11 06:12:07 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-secrets
2019/01/11 06:12:08 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-secrets
2019/01/11 06:12:08 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-webhook-configs
2019/01/11 06:12:08 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-webhook-configs
2019/01/11 06:12:08 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-webhook-configs
2019/01/11 06:12:08 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-service
2019/01/11 06:12:08 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-service
2019/01/11 06:12:09 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-service
2019/01/11 06:12:09 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-role-binding
2019/01/11 06:12:09 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-role-binding
2019/01/11 06:12:09 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-role-binding
2019/01/11 06:12:09 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-certificates-role-binding
2019/01/11 06:12:09 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-certificates-role-binding
2019/01/11 06:12:09 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-certificates-role-binding
2019/01/11 06:12:09 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-secrets-role-binding
2019/01/11 06:12:09 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-secrets-role-binding
2019/01/11 06:12:10 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-secrets-role-binding
2019/01/11 06:12:10 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-webhook-configs-role-binding
2019/01/11 06:12:10 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-webhook-configs-role-binding
2019/01/11 06:12:10 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-webhook-configs-role-binding
2019/01/11 06:12:10 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-service-role-binding
2019/01/11 06:12:10 does not exist, creating (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-service-role-binding
2019/01/11 06:12:10 successfully created (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-service-role-binding
2019/01/11 06:12:10 reconciling (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller
2019/01/11 06:12:11 does not exist, creating (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller
2019/01/11 06:12:11 successfully created (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller
2019/01/11 06:12:11 reconciling (apps/v1, Kind=Deployment) kube-system/net-attach-def-admission-controller-server
2019/01/11 06:12:11 does not exist, creating (apps/v1, Kind=Deployment) kube-system/net-attach-def-admission-controller-server
2019/01/11 06:12:11 successfully created (apps/v1, Kind=Deployment) kube-system/net-attach-def-admission-controller-server
2019/01/11 06:12:11 reconciling (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-supported
2019/01/11 06:12:11 does not exist, creating (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-supported
2019/01/11 06:12:12 successfully created (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-supported
2019/01/11 06:12:12 reconciling (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-unsupported
2019/01/11 06:12:12 does not exist, creating (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-unsupported
2019/01/11 06:12:12 successfully created (apps/v1, Kind=DaemonSet) multus/containernetworking-cni-plugins-unsupported
2019/01/11 06:12:12 reconciling (/v1, Kind=ConfigMap) multus/multus-cni-config
2019/01/11 06:12:12 does not exist, creating (/v1, Kind=ConfigMap) multus/multus-cni-config
2019/01/11 06:12:12 successfully created (/v1, Kind=ConfigMap) multus/multus-cni-config
2019/01/11 06:12:12 reconciling (apps/v1, Kind=DaemonSet) multus/multus
2019/01/11 06:12:13 does not exist, creating (apps/v1, Kind=DaemonSet) multus/multus
2019/01/11 06:12:13 successfully created (apps/v1, Kind=DaemonSet) multus/multus
2019/01/11 06:12:13 all objects successfully applied







2019/01/11 06:17:13 Reconciling NetworkConfig /default
2019/01/11 06:17:13 Starting render phase
2019/01/11 06:17:13 Render phase done, rendered 46 objects
2019/01/11 06:17:13 reconciling (/v1, Kind=ConfigMap) openshift-cluster-network-operator/applied-default
2019/01/11 06:17:13 update was successful
2019/01/11 06:17:13 reconciling (/v1, Kind=Namespace) /openshift-sdn
2019/01/11 06:17:13 update was noop
2019/01/11 06:17:13 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /clusternetworks.network.openshift.io
2019/01/11 06:17:13 update was noop
2019/01/11 06:17:13 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /hostsubnets.network.openshift.io
2019/01/11 06:17:13 update was noop
2019/01/11 06:17:13 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /netnamespaces.network.openshift.io
2019/01/11 06:17:13 update was noop
2019/01/11 06:17:13 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /egressnetworkpolicies.network.openshift.io
2019/01/11 06:17:13 update was noop
2019/01/11 06:17:13 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /openshift-sdn
2019/01/11 06:17:14 update was noop
2019/01/11 06:17:14 reconciling (/v1, Kind=ServiceAccount) openshift-sdn/sdn
2019/01/11 06:17:14 update was successful
2019/01/11 06:17:14 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn
2019/01/11 06:17:15 update was noop
2019/01/11 06:17:15 reconciling (/v1, Kind=Namespace) /openshift-infra
2019/01/11 06:17:15 update was noop
2019/01/11 06:17:15 reconciling (/v1, Kind=ServiceAccount) openshift-infra/sdn-controller
2019/01/11 06:17:15 update was successful
2019/01/11 06:17:15 reconciling (/v1, Kind=ServiceAccount) openshift-sdn/sdn-controller
2019/01/11 06:17:16 update was successful
2019/01/11 06:17:16 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /openshift-sdn-controller
2019/01/11 06:17:16 update was noop
2019/01/11 06:17:16 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn-controller
2019/01/11 06:17:17 update was noop
2019/01/11 06:17:17 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /openshift-sdn-controller-infra
2019/01/11 06:17:17 update was noop
2019/01/11 06:17:17 reconciling (rbac.authorization.k8s.io/v1, Kind=Role) openshift-sdn/openshift-sdn-controller-leaderelection
2019/01/11 06:17:18 update was noop
2019/01/11 06:17:18 reconciling (rbac.authorization.k8s.io/v1, Kind=RoleBinding) openshift-sdn/openshift-sdn-controller-leaderelection
2019/01/11 06:17:18 update was noop
2019/01/11 06:17:18 reconciling (rbac.authorization.k8s.io/v1, Kind=Role) openshift-infra/openshift-sdn-controller-account
2019/01/11 06:17:18 update was noop
2019/01/11 06:17:18 reconciling (rbac.authorization.k8s.io/v1, Kind=RoleBinding) openshift-infra/openshift-sdn-controller-account
2019/01/11 06:17:19 update was noop
2019/01/11 06:17:19 reconciling (/v1, Kind=ConfigMap) openshift-sdn/sdn-controller-config
2019/01/11 06:17:19 update was noop
2019/01/11 06:17:19 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/sdn-controller
2019/01/11 06:17:19 update was successful
2019/01/11 06:17:19 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/ovs
2019/01/11 06:17:20 update was successful
2019/01/11 06:17:20 reconciling (/v1, Kind=ConfigMap) openshift-sdn/sdn-config
2019/01/11 06:17:20 update was noop
2019/01/11 06:17:20 reconciling (apps/v1, Kind=DaemonSet) openshift-sdn/sdn
2019/01/11 06:17:21 update was successful
2019/01/11 06:17:21 reconciling (apiextensions.k8s.io/v1beta1, Kind=CustomResourceDefinition) /network-attachment-definitions.k8s.cni.cncf.io
2019/01/11 06:17:21 update was noop
2019/01/11 06:17:21 reconciling (/v1, Kind=Namespace) /multus
2019/01/11 06:17:21 update was noop
2019/01/11 06:17:21 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /multus
2019/01/11 06:17:22 update was noop
2019/01/11 06:17:22 reconciling (/v1, Kind=ServiceAccount) multus/multus
2019/01/11 06:17:22 update was successful
2019/01/11 06:17:22 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /multus
2019/01/11 06:17:23 update was noop
2019/01/11 06:17:23 reconciling (/v1, Kind=ServiceAccount) kube-system/net-attach-def-admission-controller-sa
2019/01/11 06:17:23 update was successful
2019/01/11 06:17:23 reconciling (/v1, Kind=Secret) kube-system/net-attach-def-admission-controller-sa-secret
2019/01/11 06:17:23 update was successful
2019/01/11 06:17:23 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller
2019/01/11 06:17:24 update was noop
2019/01/11 06:17:24 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-certificates
2019/01/11 06:17:24 update was noop
2019/01/11 06:17:24 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-secrets
2019/01/11 06:17:25 update was noop
2019/01/11 06:17:25 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-webhook-configs
2019/01/11 06:17:25 update was noop
2019/01/11 06:17:25 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRole) /net-attach-def-admission-controller-service
2019/01/11 06:17:25 update was noop
2019/01/11 06:17:25 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-role-binding
2019/01/11 06:17:26 update was noop
2019/01/11 06:17:26 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-certificates-role-binding
2019/01/11 06:17:26 update was noop
2019/01/11 06:17:26 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-secrets-role-binding
2019/01/11 06:17:27 update was noop
2019/01/11 06:17:27 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-webhook-configs-role-binding
2019/01/11 06:17:27 update was noop
2019/01/11 06:17:27 reconciling (rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding) /net-attach-def-admission-controller-service-role-binding
2019/01/11 06:17:27 update was noop
2019/01/11 06:17:27 reconciling (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller
2019/01/11 06:17:28 could not apply (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller: could not update object (batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller: Job.batch "install-net-attach-def-admission-controller" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`, spec.selector: Invalid value: "null": field is immutable, spec.template: Invalid value: core.PodTemplateSpec{ObjectMeta:v1.ObjectMeta{Name:"", GenerateName:"", Namespace:"", SelfLink:"", UID:"", ResourceVersion:"", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Initializers:(*v1.Initializers)(nil), Finalizers:[]string(nil), ClusterName:""}, Spec:core.PodSpec{Volumes:[]core.Volume(nil), InitContainers:[]core.Container(nil), Containers:[]core.Container{core.Container{Name:"install-net-attach-def-admission-controller-deps", Image:"nfvpe/net-attach-def-admission-controller:snapshot", Command:[]string{"installer"}, Args:[]string{"-namespace=kube-system", "-alsologtostderr"}, WorkingDir:"", Ports:[]core.ContainerPort(nil), EnvFrom:[]core.EnvFromSource(nil), Env:[]core.EnvVar(nil), Resources:core.ResourceRequirements{Limits:core.ResourceList(nil), Requests:core.ResourceList(nil)}, VolumeMounts:[]core.VolumeMount(nil), VolumeDevices:[]core.VolumeDevice(nil), LivenessProbe:(*core.Probe)(nil), ReadinessProbe:(*core.Probe)(nil), Lifecycle:(*core.Lifecycle)(nil), TerminationMessagePath:"/dev/termination-log", TerminationMessagePolicy:"File", ImagePullPolicy:"IfNotPresent", SecurityContext:(*core.SecurityContext)(nil), Stdin:false, StdinOnce:false, TTY:false}}, RestartPolicy:"OnFailure", TerminationGracePeriodSeconds:(*int64)(0xc438273cd0), ActiveDeadlineSeconds:(*int64)(nil), DNSPolicy:"ClusterFirst", NodeSelector:map[string]string(nil), ServiceAccountName:"net-attach-def-admission-controller-sa", AutomountServiceAccountToken:(*bool)(nil), NodeName:"", SecurityContext:(*core.PodSecurityContext)(0xc4331221c0), ImagePullSecrets:[]core.LocalObjectReference(nil), Hostname:"", Subdomain:"", Affinity:(*core.Affinity)(nil), SchedulerName:"default-scheduler", Tolerations:[]core.Toleration(nil), HostAliases:[]core.HostAlias(nil), PriorityClassName:"", Priority:(*int32)(nil), DNSConfig:(*core.PodDNSConfig)(nil), ReadinessGates:[]core.PodReadinessGate(nil)}}: field is immutable]
2019/01/11 06:17:29 Reconciling NetworkConfig /default
2019/01/11 06:17:29 Starting render phase
```

zshi-redhat:
CRD & CR failed to apply in single run, see log in the first file of below github gist:
https://gist.github.com/zshi-redhat/090b3cb3a981852bbdd58ab40da8b493

zshi-redhat:
The additional-networks/cr/cr.yaml [template](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/additional-networks/cr/cr.yaml#L7) also need change as Multus expects a json string instead of an formated object from ``config`` in the template, for example:

```
spec:
  config: |-
{{.AdditionalNetworkConfig | indent 4}}
```

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,78 @@
...
+  namespace: multus
+  labels:
+    app: multus
+data:
+  70-multus.conf: |
+    {
+      "name": "multus",
+      "type": "multus",
+      "delegates": [{
+        "type": "{{.CNIPluginName}}",
```
squeed: Can we use the Multus default CNI configuration directory? Or do we need to hard-code the name?

bindata/network/sriov/002-rbac.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
squeed: if the service account doesn't have any RoleBindings, then you can delete all of this and the namespace's default service account will be used.

bindata/network/multus/cni-plugins-supported.yaml
```diff
@@ -0,0 +1,44 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: containernetworking-cni-plugins-supported
+  namespace: multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches supported CNI plugins on each node.
```
danwinship: can't we just include these containers as initContainers in the multus pod rather than launching additional daemonsets?

bindata/network/multus/cni-plugins-unsupported.yaml
```diff
@@ -0,0 +1,44 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: containernetworking-cni-plugins-unsupported
+  namespace: multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches *unsupported* CNI plugins on each node.
```
danwinship: if they're unsupported then shouldn't the user have to deal with them themsevles?

dcbw: @danwinship yeah, I don't know; Marc and the NFV team decided to do this kind of split.

manifests/0000_07_cluster-network-operator_03_daemonset.yaml
```diff
@@ -32,6 +32,8 @@ spec:
           value: "docker.io/openshift/origin-node:v4.0.0"
         - name: HYPERSHIFT_IMAGE
           value: "docker.io/openshift/origin-hypershift:v4.0.0"
+        - name: MULTUS_IMAGE
+          value: "docker.io/nfvpe/multus:snapshot"
```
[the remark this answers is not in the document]
dougbtv: Good suggestion, quick explanations here... I believe Dan is using this in the short-term as a workaround for the fact that we have downstream `ose-*` images for Multus, but, we don't have `openshift/origin-*` images for Multus (I'm looking into how we can get those) and he needs some public ones.

Currently, there's tags for `:latest` -- latest tagged release, `:vX.Y` for specific tags, but, Dan needed the latest one (last tagged release is fairly old), so the `:snapshot` represents the latest master head -- if need be run the Multus binary directory with `./multus --version` to see where it's built from.

dcbw:
> Reconciling of `(batch/v1, Kind=Job) kube-system/install-net-attach-def-admission-controller` failed with below log message ( `see error message in last three lines` ), seems the admission-controller Job manifest doesn't specify matched `selector` and `label`.

@zshi-redhat going to fix this by making the install a Deployment rather than a Job. At least for now. That also means the AC should be idempotent (or at the very least not re-create a CSR and secret if a secret already exists).

pkg/network/render.go
```diff
@@ -15,8 +15,15 @@ func Render(conf *netv1.NetworkConfigSpec, manifestDir string) ([]*uns.Unstructu
 	log.Printf("Starting render phase")
 	objs := []*uns.Unstructured{}
 
+	// render Multus
+	o, crdRendered, err := RenderMultus(conf, manifestDir)
```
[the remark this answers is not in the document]
dcbw: @danwinship it's actually the network attachment CRD that's getting rendered here.  I'll update.

bindata/network/multus/cni-plugins-supported.yaml
dcbw: done

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -53,11 +53,13 @@ type NetworkConfigSpec struct {
...
 
 	// Additional networks to make available to pods. If they are specified,
-	// pods can request them via annotations.
-	//
-	// Specifying any additionalNetworks will enable Multus across the cluster.
+	// pods can request them via annotations when DeployMultus is true.
 	AdditionalNetworks []AdditionalNetworkDefinition `json:"additionalNetworks"`
 
+	// DeployMultus specifies whether or not Multus should be installed to
+	// provide additional network support for pods.
```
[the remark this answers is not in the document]
dcbw: Fixed.

bindata/network/sriov/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,56 @@
...
+    metadata:
+      labels:
+        app: sriovdp
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/os: linux
```
zshi-redhat: Sriov-device-plugin fails to run on nodes which doesn't contain SR-IOV capable devices or VFs not created.
Perhaps worth a note here to indicating further improvement is needed to run SR-IOV device plugin on selected nodes. This might be solved later with NFD to add labels to SR-IOV capable nodes.

bindata/network/multus/admission-controller.yaml
```diff
@@ -0,0 +1,219 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: net-attach-def-admission-controller-service
+subjects:
+- kind: ServiceAccount
+  name: net-attach-def-admission-controller-sa
+  namespace: multus
+---
+apiVersion: apps/v1
+kind: Deployment
```
zshi-redhat: install-net-attach-def-admission-controller still crashes when running as a deployment even all the resources are created successfully:
```
[zshi@192 openshift-aws]$ oc get pods -n multus
NAME                                                           READY     STATUS             RESTARTS   AGE
install-net-attach-def-admission-controller-7558f8f477-gnfrg   0/1       CrashLoopBackOff   6          23m
multus-4cfnq                                                   1/1       Running            0          3h
multus-5khbs                                                   1/1       Running            0          3h
multus-7vqsz                                                   1/1       Running            0          3h
multus-b42cs                                                   1/1       Running            0          3h
multus-gvqk4                                                   1/1       Running            0          3h
multus-xskd6                                                   1/1       Running            0          3h
net-attach-def-admission-controller-server-5b5d4cbc7d-zmkr2    0/1       CrashLoopBackOff   5          23m
```

```
[zshi@192 openshift-aws]$ oc logs install-net-attach-def-admission-controller-7558f8f477-gnfrg -n multus
I0114 06:16:20.805426       1 main.go:28] starting webhook installation
I0114 06:16:20.807821       1 installer.go:41] generating Certificate Signing Request
2019/01/14 06:16:20 [INFO] received CSR
2019/01/14 06:16:20 [INFO] generating key: rsa-2048
2019/01/14 06:16:21 [INFO] encoded CSR
I0114 06:16:21.603198       1 installer.go:308] raw CSR and private key successfully created
I0114 06:16:21.634796       1 installer.go:58] CSR net-attach-def-admission-controller-csr already exists, trying to reuse it
I0114 06:16:21.634814       1 installer.go:78] using already issued certificate for CSR net-attach-def-admission-controller-csr
I0114 06:16:21.634822       1 installer.go:315] signed certificate successfully obtained
I0114 06:16:21.644833       1 installer.go:280] secret net-attach-def-admission-controller-secret already exists, removing it first
I0114 06:16:21.667687       1 installer.go:285] secret net-attach-def-admission-controller-secret removed
I0114 06:16:21.690156       1 installer.go:322] secret successfully created
I0114 06:16:21.697318       1 installer.go:256] validating webhook net-attach-def-admission-controller-validating-config already exists, removing it first
I0114 06:16:21.710655       1 installer.go:261] validating webhook configuration net-attach-def-admission-controller-validating-config removed
I0114 06:16:21.722235       1 installer.go:329] validating webhook configuration successfully created
I0114 06:16:21.735874       1 installer.go:256] validating webhook net-attach-def-admission-controller-isolating-config already exists, removing it first
I0114 06:16:21.750037       1 installer.go:261] validating webhook configuration net-attach-def-admission-controller-isolating-config removed
I0114 06:16:21.756337       1 installer.go:335] Isolating webhook configuration successfully created
I0114 06:16:21.763667       1 installer.go:244] service net-attach-def-admission-controller-service already exists, removing it first
I0114 06:16:21.844900       1 installer.go:249] service net-attach-def-admission-controller-service removed
I0114 06:16:21.869995       1 installer.go:342] service successfully created
I0114 06:16:21.870012       1 installer.go:344] all resources created successfully
```

```
[zshi@192 openshift-aws]$ oc logs net-attach-def-admission-controller-server-5b5d4cbc7d-zmkr2 -n multus
I0114 06:18:24.810735       1 main.go:34] starting net-attach-def-admission-controller webhook server
F0114 06:18:24.813449       1 main.go:46] error starting web server: tls: private key does not match public key
goroutine 1 [running]:
github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog.stacks(0xc420362800, 0xc420368000, 0x70, 0xbb)
	/go/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/gopath/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog/glog.go:766 +0xcf
github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog.(*loggingT).output(0x19bf320, 0xc400000003, 0xc4200f4840, 0x1932413, 0x7, 0x2e, 0x0)
	/go/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/gopath/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog/glog.go:717 +0x30f
github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog.(*loggingT).printf(0x19bf320, 0xc400000003, 0x1110fd9, 0x1d, 0xc420159f48, 0x1, 0x1)
	/go/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/gopath/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog/glog.go:655 +0x14b
github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog.Fatalf(0x1110fd9, 0x1d, 0xc420159f48, 0x1, 0x1)
	/go/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/gopath/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/vendor/github.com/golang/glog/glog.go:1145 +0x67
main.main()
	/go/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/gopath/src/github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/cmd/webhook/main.go:46 +0x38c
```

dcbw: @zshi-redhat yes, the installer needs to be idempotent and not attempt to re-create resources when it finds them already existing. I believe that's the problem here, though I could be wrong.

squeed: Interesting - the logs should be pretty clear about what is being created / updated. There could be a bug where ServiceAccounts aren't being merged correctly (I fixed one a few months ago already).

dougbtv: I've also seen this happen upstream too, it happens... intermittently though, like, when restarting the admission controller a few times while developing with it. I'm tracking in this issue:

https://github.com/K8sNetworkPlumbingWG/net-attach-def-admission-controller/issues/12

I'll try Dan's suggestion, too.

eparis:
/hold
looks like a feature after freeze

dcbw:
> looks like a feature after freeze

@eparis it is, yes; though it doesn't have an LGTM yet because we are waiting on some images. @squeed is working on the exception request.

bindata/network/multus/admission-controller.yaml
zshi-redhat: I hit the similar [issue](https://github.com/intel/sriov-network-device-plugin/pull/55#issuecomment-450068630) in SR-IOV admission-controller where we have to remove the existing CSR and re-create it.

pkg/network/render.go
zshi-redhat: @dcbw I still see this failure with the latest version of PR, not sure if it's specified to how I run the test (run the binary of cluster-network-operator in one of the master nodes). did you ever hit this issue?

zshi-redhat:
@dcbw  I noticed that the admission controller for Multus was removed from this PR, are we still going to support it in this PR? if yes, I'd also like to add admission controller for SR-IOV, otherwise we can do that later together with Multus admission-controller. Here is the manifest for SRIOV admission controller:
https://gist.github.com/zshi-redhat/03fe4e937e672bcc4636b884e49a939a

squeed:
Exception request submitted.

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -53,11 +53,14 @@ type NetworkConfigSpec struct {
 	DefaultNetwork DefaultNetworkDefinition `json:"defaultNetwork"`
 
 	// Additional networks to make available to pods. If they are specified,
-	// pods can request them via annotations.
-	//
-	// Specifying any additionalNetworks will enable Multus across the cluster.
+	// pods can request them via annotations when multiple networks are enabled.
```
danwinship: Nitpick: this implies that you're allowed to specify `AdditionalNetworks` along with `DisableMultiNetwork: "true"`, and it just won't work at runtime. But in fact it will be rejected at configuration-parsing time.

pkg/network/render.go
```diff
@@ -185,15 +213,17 @@ func RenderAdditionalNetworks(conf *netv1.NetworkConfigSpec, manifestDir string)
...
 
-	// render Multus when additional networks is provided
-	if len(ans) > 0 {
-		objs, err := renderMultusConfig(manifestDir)
+	if len(ans) == 0 {
+		return nil, nil
+	}
+
+	// render the CRD when additional networks are provided
+	if !netAttachDefCRDRendered {
```
danwinship: This can't happen. We only allow additional networks if multus is enabled, and if multus is enabled then either it already rendered the NetworkAttachmentDefinition CRD or else the renderer errored out before reaching this point.

danwinship:
> admission controller for SR-IOV

The handling of SR-IOV in this PR seems kind of sketchy; we pretend in the config that it's just a totally random uninterpreted CNI plugin, but then we deploy a daemonset if it's present and apparently will also have an admission controller? If we're going to have special logic for SR-IOV, maybe it should be more noticeably special in the config as well?

squeed:
> The handling of SR-IOV in this PR seems kind of sketchy;
I agree; @zshi-redhat and I have been discussing the best way to configure SR-IOV and it needs a real configuration type and an object.

Let's pull SR-IOV out of this PR. It's a separate feature.

dcbw:
> I noticed that the admission controller for Multus was removed from this PR, are we still going to support it in this PR?

@zshi-redhat the admission controller was removed from this PR and should be submitted as a follow-up once the issues with it have been ironed out. Doug is exploring the namespaceIsolation option as an interim solution while that happens.

danwinship:
/lgtm
/hold
until the image stuff is resolved. and the PR gets exceptioned.

bmeng:
Maybe we need to have a dhcp daemon running on each node by default.

I was using a separated ds to setup it for testing.
https://github.com/openshift-qe/v3-testfiles/blob/master/networking/multus-cni/DaemonSet-dhcp-daemon.yaml

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,78 @@
...
+        volumeMounts:
+        - mountPath: /host/opt/cni/bin
+          name: cnibin
+      containers:
+      - name: kube-multus
+        image: {{.MultusImage}}
+        command: ["/entrypoint.sh"]
+        args:
+        - "--multus-conf-file=auto"
+        - "--multus-kubeconfig-file-host=/etc/kubernetes/cni/net.d/multus.d/multus.kubeconfig"
```
squeed: Where does this come from? Can we just use the kubelet's kubeconfig?

squeed: Ah, I see, it does it's own magic. Cool.

dcbw:
NOTE: we are waiting on https://github.com/openshift/release/pull/2570 to land, which will provide the official images for Multus. Once that lands, I will remove the "hack: no images for now" commit and update the image references. But this PR demonstrates that Multus passes CI even without official images.

dcbw:
@danwinship real images added now, FWIW

manifests/image-references
```diff
@@ -14,3 +14,15 @@ spec:
     from:
       kind: DockerImage
       name: docker.io/openshift/origin-hypershift:v4.0.0
+  - name: multus
```
knobunc: I think this needs to be multus-cni

bindata/network/multus/000-ns.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: multus
+  labels:
+    name: multus
+    openshift.io/run-level: "0"
```
danwinship: does multus actually need this?

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - "--namespace-isolation=true"
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "50Mi"
+          limits:
+            cpu: "100m"
+            memory: "50Mi"
```
danwinship: Drop the `limits` section unless (a) there's a chance that the pod will attempt to use a lot of memory/CPU in some circumstances, and (b) you actually want to thwart it when that happens. (You probably just copied this from the openshift-sdn daemonset, and the `limits` have been removed there since then.)

bindata/network/multus/multus.yaml
```diff
@@ -0,0 +1,78 @@
...
+        image: {{.MultusImage}}
+        command: ["/entrypoint.sh"]
+        args:
+        - "--multus-conf-file=auto"
+        - "--multus-kubeconfig-file-host=/etc/kubernetes/cni/net.d/multus.d/multus.kubeconfig"
+        - "--namespace-isolation=true"
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "50Mi"
```
danwinship: This says that for scheduling purposes, kube should assume that the pod will use 50M of RSS, and 100 milli-CPUs (aka 10% of a single CPU). The multus pod is basically idle after initial setup, so those numbers are definitely too high, and will prevent the node from being effectively scheduled to capacity.

Probably best to just drop the `resources` section entirely

jeremyeder: I agree it's probably too high, but resources were recently added to every operator and pod by @sjenning.  Example: https://github.com/openshift/cluster-node-tuning-operator/commit/2f6d6179e649e4c15f25df75277a4756eb573465

@sjenning what values would be consistent with the rest of your PRs from last week?

bindata/network/multus/000-ns.yaml
squeed: Yes, definitely - to be except from SecurityContextConstraints so it can create privileged pods.

bindata/network/multus/multus.yaml
dcbw: on discussion with casey, we're just removing this section as the container basically idles after startup

dcbw:
/retest
One master failed to correctly download images for many containers:

kube-system                          etcd-member-ip-10-0-26-81.ec2.internal                    1/1       Running             0          17m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>
multus                               multus-w72wg                                              0/1       CrashLoopBackOff    9          22m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>
openshift-cluster-network-operator   cluster-network-operator-htwh8                            1/1       Running             0          22m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>
openshift-dns                        dns-default-968p5                                         0/2       ContainerCreating   0          21m       <none>       ip-10-0-26-81.ec2.internal   <none>
openshift-kube-apiserver             installer-1-ip-10-0-26-81.ec2.internal                    0/1       ContainerCreating   0          17m       <none>       ip-10-0-26-81.ec2.internal   <none>
openshift-sdn                        ovs-cpvss                                                 0/1       ImageInspectError   0          22m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>
openshift-sdn                        sdn-controller-4tx5t                                      1/1       Running             0          22m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>
openshift-sdn                        sdn-dz8n4                                                 0/1       ImageInspectError   0          22m       10.0.26.81   ip-10-0-26-81.ec2.internal   <none>

dcbw:
@danwinship look better now?

jeremyeder:
+100

PR #55, opened 2019-01-11
openshift-sdn: minor daemonset cleanups:

squeed:
- use url-only-kubeconfig instead of patching the kubeconfig
- pass nodename directly instead of sed'ing

This enables restart-on-config-change.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -4,7 +4,7 @@ metadata:
   namespace: openshift-sdn
   name: sdn-config
 data:
-  sdn-config.yaml.in: |-
+  sdn-config.yaml: |-
 {{.NodeConfig | indent 4}}
 ---
 kind: DaemonSet
@@ -14,7 +14,7 @@ metadata:
   namespace: openshift-sdn
   annotations:
     kubernetes.io/description: |
-      This daemon set launches the OpenShift networking components (kube-proxy, DNS, and openshift-sdn).
+      This daemon set launches the OpenShift networking components (kube-proxy and openshift-sdn).
       It expects that OVS is running on the node.
 spec:
   selector:
@@ -38,8 +38,6 @@ spec:
       hostNetwork: true
       hostPID: true
       containers:
-      # The network container launches the openshift-sdn process, the kube-proxy, and the local DNS service.
-      # It relies on an up to date node-config.yaml being present.
       - name: sdn
         image: {{.NodeImage}}
         command: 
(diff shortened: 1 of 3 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -66,22 +64,18 @@ spec:
             fi
           done
 
-          # Create the node config
-          sed "s/%%NODENAME%%/${NODENAME}/" /config/sdn-config.yaml.in > /tmp/sdn-config.yaml
+          # local environment overrides
+          if [[ -f /etc/sysconfig/openshift-sdn ]]; then
+            set -o allexport
+            source /etc/sysconfig/openshift-sdn
```
danwinship: `set +o allexport` afterward?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -115,14 +109,20 @@ spec:
           name: host-cni-netd
         - mountPath: /var/lib/cni/networks/openshift-sdn
           name: host-var-lib-cni-networks-openshift-sdn
+        - mountPath: /lib/modules
+          name: host-modules
```
danwinship: commit message doesn't explain what this part is about... was it intentional?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -115,14 +109,20 @@ spec:
...
+          readOnly: true
         resources:
           requests:
             cpu: 100m
             memory: 200Mi
         env:
         - name: OPENSHIFT_DNS_DOMAIN
           value: cluster.local
-        - name: NODENAME
+        - name: K8S_NODE_NAME
```
danwinship: how does this work exactly? I don't see anything in origin, kubernetes, or cluster-network-operator that looks at `K8S_NODE_NAME`

squeed: It only merged a few days ago ;-). https://github.com/openshift/origin/blob/master/pkg/cmd/openshift-sdn/cmd.go#L145

bindata/network/openshift-sdn/sdn.yaml
squeed: It's needed for iptables; this fixes a bug that was copied over from openshift-ansible - namely, the volume was added but not the mount.

I'll update the commit

bindata/network/openshift-sdn/sdn.yaml
squeed: Sure.

squeed:
Thanks for the review; updated.

PR #62, opened 2019-01-16
set cpu request and remove limits for pods and operator

sjenning:
@smarterclayton @derekwaynecarr


danwinship:
As I understand it, these were just copied from the ansible config and we have no particular reason for thinking those numbers were right. And the new numbers put us in line with the new recommendations.

/lgtm

PR #68, opened 2019-01-21
Update namespace/pods to current naming conventions

danwinship:
We currently run pods called `cluster-network-operator-xxxxx` in a namespace called `openshift-cluster-network-operator`, but current naming conventions would prefer `network-operator-xxxxx` and `openshift-network-operator`, respectively.

Not sure if there are other things we need to change too?


ironcladlou:
This lgtm, not sure whether you caught all your references, but I agree with the new names.

squeed:
The only other thing we need to change is some documentation in the installer.

/lgtm

PR #72, opened 2019-01-22
set ClusterOperator.Status

danwinship:
OK, this may still need a bit of testing/finishing but it's at least ready for review.

- Currently, at any time exactly one of `Failing`, `Progressing`, and `Available` should be `True`, though we'd talked at least in passing about, eg, making the cluster be both `Failing` and `Available` if the requested configuration is bad but the deployed configuration is good.
- AFAICT if we want to watch the cluster config, operator config, and a bunch of daemonsets, we need to have three separate Controllers and three separate Reconcilers.
- In theory this will not need any changes to support Multus/OVN/Kuryr; the NetworkConfig reconciler will pull the Multus/OVN/Kuryr-related DaemonSet names out of the `network.Render()` result, and pass them to the DaemonSet reconciler, which will then update the operator status based on the status of those DaemonSets rather than / in addition to the openshift-sdn ones. (Although at some point we'll probably need to add Deployment-reconciling support in addition to the DaemonSet support.)
- `StatusManager` exists mostly to deal with the fact that we need three separate Reconcilers, but they have to coordinate the resulting operator status. In particular, if the configuration is invalid, and then a DaemonSet's status changes, we don't want that to result in the operator status becoming Available, so `StatusManager.SetFromDaemonSets()` needs to be able to know not to change the status.
- `StatusManager` is intended to eventually be useful to other operators (which is why it takes the operator name and version as constructor arguments).

### DIFF
--- a/pkg/controller/clusterconfig/clusterconfig_controller.go
+++ b/pkg/controller/clusterconfig/clusterconfig_controller.go
@@ -10,6 +10,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/apply"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/network"
+	"github.com/openshift/cluster-network-operator/pkg/util/clusteroperator"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -22,14 +23,14 @@ import (
 )
 
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager) error {
-	return add(mgr, newReconciler(mgr))
+func Add(mgr manager.Manager, status *clusteroperator.StatusManager) error {
+	return add(mgr, newReconciler(mgr, status))
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, status *clusteroperator.StatusManager) reconcile.Reconciler {
 	configv1.Install(mgr.GetScheme())
-	return &ReconcileClusterConfig{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileClusterConfig{client: mgr.GetClient(), scheme: mgr.GetScheme(), status: status}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -57,6 +58,7 @@ type ReconcileClusterConfig struct {
 	// that reads objects from the cache and writes to the apiserver
 	client client.Client
 	scheme *runtime.Scheme
+	status *clusteroperator.StatusManager
 }
 
 // Reconcile propagates changes from the cluster config to the operator config.
@@ -83,6 +85,7 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
 		}
 		// Error reading the object - requeue the request.
 		log.Println(err)
+		// FIXME: operator status?
 		return reconcile.Result{}, err
 	}
 
@@ -90,6 +93,7 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
 	if err := network.ValidateClusterConfig(clusterConfig.Spec); err != nil {
 		err = errors.Wrapf(err, "failed to validate Network.Spec")
 		log.Println(err)
+		r.status.SetConfigFailing("InvalidClusterConfig", err)
 		return reconcile.Result{}, err
 	}
 
@@ -97,13 +101,18 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
 	if err != nil {
 		err = errors.Wrapf(err, "failed to generate NetworkConfig CRD")
 		log.Println(err)
+		r.status.SetConfigFailing("UpdateOperatorConfig", err)
 		return reconcile.Result{}, err
 	}
+
+	// Clear any cluster config-related errors before applying operator config
+	r.status.SetConfigSuccess()
 
 	if operatorConfig != nil {
 		if err := apply.ApplyObject(context.TODO(), r.client, operatorConfig); err != nil {
 			err = errors.Wrapf(err, "could not apply (%s) %s/%s", operatorConfig.GroupVersionKind(), operatorConfig.GetNamespace(), operatorConfig.GetName())
 			log.Println(err)
+			r.status.SetConfigFailing("ApplyClusterConfig", err)
 			return reconcile.Result{}, err
 		}
 	}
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -83,27 +85,34 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
 		}
 		// Error reading the object - requeue the request.
 		log.Println(err)
+		// FIXME: operator status?
```
squeed: Yeah, tough to know what to do. If it weren't a transient failure, then I suppose we would treat this the same as invalid input. This seems very unlikely though.

pkg/controller/networkconfig/daemonset_controller.go
```diff
@@ -0,0 +1,48 @@
...
+func (r *ReconcileDaemonSets) SetDaemonSets(daemonSets []types.NamespacedName) {
+	r.daemonSets = daemonSets
+}
+
+// Reconcile updates the ClusterOperator.Status to match the current state of the
+// watched DaemonSets
+func (r *ReconcileDaemonSets) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	found := false
+	for _, ds := range r.daemonSets {
+		if ds.Namespace == request.Namespace && ds.Name == request.Name {
```
squeed: Annoying that the operator-sdk doesn't support mixed-scope watches, AFAICT.

pkg/util/clusteroperator/status_manager.go
```diff
@@ -0,0 +1,184 @@
...
+				return status.SetFailing("NoNamespace", fmt.Errorf("Namespace %q does not exist", dsName.Namespace))
+			} else {
+				return status.SetFailing("InternalError", err)
+			}
+		}
+
+		ds := &appsv1.DaemonSet{}
+		if err := status.client.Get(context.TODO(), dsName, ds); err != nil {
+			if errors.IsNotFound(err) {
+				return status.SetFailing("NoDaemonSet", fmt.Errorf("DaemonSet %q does not exist", dsName.String()))
```
squeed: Is it useful to know if more than one daemonset is missing?

pkg/util/clusteroperator/status_manager.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil && !isNotFound {
+		log.Printf("failed to get clusteroperator %s: %v", status.name, err)
+		return fmt.Errorf("failed to get clusteroperator %s: %v", status.name, err)
+	}
+
+	oldConditions := co.Status.Conditions
+	for _, condition := range conditions {
+		co.Status.Conditions = SetStatusCondition(co.Status.Conditions, condition)
+	}
+	co.Status.Version = status.version
```
squeed: It looks like the latest API changes (https://github.com/openshift/api/pull/176) changes how this works.

In the interest of getting this out the door, let's just drop this line, revendor api, and merge the PR, then figure out the operand stuff afterwards.

pkg/util/clusteroperator/status_manager.go
```diff
@@ -0,0 +1,184 @@
...
+func (status *StatusManager) SetFromDaemonSets(daemonSets []types.NamespacedName) error {
+	if status.configFailure {
+		return nil
+	}
+
+	progressing := []string{}
+
+	for _, dsName := range daemonSets {
+		ns := &corev1.Namespace{}
+		if err := status.client.Get(context.TODO(), types.NamespacedName{Name: dsName.Namespace}, ns); err != nil {
```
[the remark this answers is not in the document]
danwinship: according to the comment in networkconfig_controller.go:

	// This client, initialized using mgr.Client() above, is a split client
	// that reads objects from the cache and writes to the apiserver

pkg/util/clusteroperator/status_manager.go
danwinship: yeah, probably

pkg/util/clusteroperator/status_manager.go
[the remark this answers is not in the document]
squeed: Lovely!

squeed:
A few minor things, but this looks awesome. I'll play around with it tomorrow.

Implementing the operand versioning will probably require a bit more sophistication in the daemonset watcher, but I don't think we'll be sad if we merge this before.

pkg/util/clusteroperator/status_manager.go
danwinship: Actually, probably not; this would normally just be a transient failure during the first few moments of the cluster's existence, and no one will ever actually see it. (And if we're going to combine errors then what if we're missing the Namespace for one DaemonSet and the DaemonSet for another, etc?)

pkg/util/clusteroperator/status.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// SetStatusCondition returns the result of setting the specified condition in
+// the given slice of conditions.
+func SetStatusCondition(oldConditions []configv1.ClusterOperatorStatusCondition, condition *configv1.ClusterOperatorStatusCondition) []configv1.ClusterOperatorStatusCondition {
```
deads2k: fyi, I think these exist here: https://github.com/openshift/library-go/blob/master/pkg/config/clusteroperator/v1helpers/status.go

danwinship: That's based on the updated ClusterOperator so would require an openshift/api bump too. We'll grab that in the next update after we get the initial version committed.

danwinship:
repushed with a StatusManager unit test

danwinship:
> @danwinship: The following test **failed**, say `/retest` to rerun them all:

```
        {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-01-25T16:11:35Z",
                "generation": 1,
                "name": "openshift-network-operator",
                "namespace": "",
                "resourceVersion": "8145",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/openshift-network-operator",
                "uid": "e35b0622-20bb-11e9-83a5-121386b5f244"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-01-25T16:11:43Z",
                        "status": "False",
                        "type": "Failing"
                    },
                    {
                        "lastTransitionTime": "2019-01-25T16:20:08Z",
                        "status": "False",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-01-25T16:20:08Z",
                        "status": "True",
                        "type": "Available"
                    }
                ],
                "extension": null,
                "version": ""
            }
        }
```

Somebody Else's Problem! :-)

PR #74, opened 2019-01-24
openshift-sdn, manifests: capture last few log lines as our TerminationMessage

squeed:
This can be useful when we have the Pod but not its logs.

Thanks @deads2k for the suggestion.


danwinship:
yes please
/lgtm

deads2k:
we need more debugging.  merging

mfojtik:
👍

Issue #69, opened 2019-01-22
add validation for cluster and operator config objects


danwinship:
I did `oc edit NetworkConfig.networkoperator.openshift.io default` and changed `hostSubnetLength` from `9` to `nine`.

CNO started logging:

    E0122 09:49:12.552833   10497 reflector.go:205] github.com/openshift/cluster-network-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1.NetworkConfig: v1.NetworkConfigList.Items: []v1.NetworkConfig: v1.NetworkConfig.Spec: v1.NetworkConfigSpec.ClusterNetworks: []v1.ClusterNetwork: v1.ClusterNetwork.HostSubnetLength: readUint32: unexpected character: �, error found in #10 byte of ...|tLength":"nine"}],"d|..., bigger context ...|rks":[{"cidr":"10.128.0.0/14","hostSubnetLength":"nine"}],"defaultNetwork":{"openshiftSDNConfig":{"m|...

once a second.

(Probably a reconciler bug? Just filing this so I don't forget.)

danwinship:
So this is really more client-go than controller-manager, and is basically just the fact that generated clients are not designed to deal with unvalidated objects. We need to take advantage of the new pre-aggregated-apiserver validation hack.

danwinship:
Fixed by #76 and https://github.com/openshift/api/pull/412.

PR #76, opened 2019-01-24
Add openAPIV3Schema to operator config type

danwinship:
The controller-runtime code gets very angry if one of the objects it's watching can't be deserialized into the expected golang type.

For the cluster config, we can fix this using the new CRD validation hack in origin, but for the network config that's not the greatest idea, since then origin would have to vendor cluster-network-operator to get the type, and then we'd have to coordinate changes between the two repos whenever we changed anything.

Instead, we can use OpenAPI validation to ensure that objects will only be accepted by the apiserver if they are at least close-enough-to-valid that we'll be able to deserialize them. (Eg, you don't have a string where an integer is supposed to be.)

This is based on the current definition of the struct but maybe we should remove the unsupported bits (OVN, Kuryr) until the code to deal with them has landed?

Fixes half of #69

### DIFF
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -14,6 +14,87 @@ spec:
     - name: v1
       served: true
       storage: true
+  validation:
+    # Ensure we will be able to deserialize the object into the golang type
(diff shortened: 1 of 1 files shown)


squeed:
Wow, thanks for going through this. This is awesome.

PR #77, opened 2019-01-25
types: remove some unused variables / aspirational comments

squeed:


danwinship:
You need to rebase to after #76 and then remove this stuff from the schema too

squeed:
rebased & updated. I also added `disableMultiNetwork` to the schema while I was at it.

PR #81, opened 2019-01-29
pkg/apis: add missing omitempty

squeed:


danwinship:
and all the other fields look correct
/lgtm

PR #82, opened 2019-01-29
openshift-sdn: join control plane namespaces so multitenant works in 4+

squeed:
This groups control-plane namespaces in to three groups:

- VNID 0 (default, can access and be accessed by all)
  - openshift-kube-apiserver
  - openshift-dns
  - openshift-ingress
  - openshift-monitoring
  - openshift-operator-lifecycle-manager
  - openshift-image-registry

- VNID 1 (control plane)
  - openshift-apiserver
  - kube-system (etcd)

And all others will have their own isolated network.

Fixes SDN-284 / BZ 1659376

### DIFF
--- a/bindata/network/openshift-sdn/004-multitenant.yaml
+++ b/bindata/network/openshift-sdn/004-multitenant.yaml
@@ -1,0 +1,75 @@
+{{if eq .Mode "Multitenant"}}
+# Namespaces which must be reachable to all pods, or can reach all pods
+# - openshift-dns
+# - openshift-ingress
+# - openshift-monitoring
+# - openshift-kube-apiserver
+# - openshift-operator-lifecycle-manager
+# - openshift-image-registry
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-dns
+netid: 0
+netname: openshift-dns
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-ingress
+netid: 0
+netname: openshift-ingress
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-monitoring
+netid: 0
+netname: openshift-monitoring
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-kube-apiserver
+netid: 0
+netname: openshift-kube-apiserver
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-operator-lifecycle-manager
+netid: 0
+netname: openshift-operator-lifecycle-manager
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-image-registry
+netid: 0
+netname: openshift-image-registry
+
+---
+# Namespaces which are part of the "control plane" and need to reach each other:
+# - kube-system (etcd)
+# - openshift-apiserver
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-apiserver
+netid: 1
+netname: openshift-apiserver
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: kube-system
+netid: 1
+netname: kube-system
+
+{{- end}}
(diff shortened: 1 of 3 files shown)


danwinship:
Note that currently the sdn master won't deal correctly with someone else renumbering NetNamespaces out from under it, so we need to fix that before we can support either 3.x upgrades or post-install changes to Mode.

squeed:
No, I haven't dug in to the OLM issue yet. But, if they're making contact, then we should allow it.

Interesting about the renumbering. How does `oc adm pod-network join` work?

danwinship:
> No, I haven't dug in to the OLM issue yet. But, if they're making contact, then we should allow it.

Yeah, but I think the right way to do that might be to put OLM in VNID 0. With a name like "operator lifecycle manager", it seems likely that it's going to want to talk to lots of operators...

> Interesting about the renumbering. How does `oc adm pod-network join` work?

It adds an annotation to the NetNamespace which the master then acts on. I don't remember why we did it that way. I believe it involved amusingly quaint ideas about backward-compatibility constraints.

squeed:
> It adds an annotation to the NetNamespace which the master then acts on. I don't remember why we did it that way. I believe it involved amusingly quaint ideas about backward-compatibility constraints.

I just looked in to the code that does this. Annoyingly, you can't request that the vnid be changed to a specific value, you can only request a join to an existing namespace or the global network. So instead of just creating the NetNamespaces, I suppose we have to request that they be joined.

squeed:
Oh, but it's even more annoying than that. The operator shouldn't request networks be joined over-and-over, that's a waste, but it can't actually create networks with a requested netid. I think we need an origin change.

danwinship:
> I think we need an origin change.

Yes, sorry, I meant to imply that.

danwinship:
However, we don't need it *now*; we only need it when there's the possibility of the CNO trying to force NetIDs onto NetNamespaces while the sdn-controller is running. aka, when we support upgrades or plugin switches

danwinship:
> I think the right way to do that might be to put OLM in VNID 0.

Pretty sure this is correct. Can you update that and repush? (Marc wants multitenant working in beta2.)

squeed:
Yup, OLM allows for arbitrary registries to be created, so it needs to be VNID 0. Updated.

danwinship:
/hold
(rerunning tests with a hack to run e2e-aws multitenant)

squeed:
Added a DNM commit that sets default to Multitenant so we can get a CI run.

danwinship:
```
"status": "True",
"type": "Failing"
"reason": "ApplyOperatorConfig",
"message": "could not apply (network.openshift.io/v1, Kind=NetNamespace) /openshift-dns: could not retrieve existing (network.openshift.io/v1, Kind=NetNamespace) /openshift-dns: no matches for kind \"NetNamespace\" in version \"network.openshift.io/v1\"",
```

uh... ?

danwinship:
Looks like the same RESTMapper problem as in #89. But the CRD *does* exist.

In fact, even if I manually create all the NetNamespaces, CNO still complains that it can't fetch the existing object because the Kind doesn't exist...

squeed:
Yeah, I saw that error too when I was working on the local-hack.

cmd/cluster-network-operator/main.go
```diff
@@ -82,7 +83,10 @@ func main() {
 	}
 
 	// Create a new Cmd to provide shared dependencies and start components
-	mgr, err := manager.New(cfg, manager.Options{Namespace: namespace})
+	mgr, err := manager.New(cfg, manager.Options{
+		Namespace:      namespace,
+		MapperProvider: k8sutil.NewDynamicRESTMapper,
```
squeed: This option must have been fun to trace back and find.

squeed:
Ran a quick local test, seemed to work without complaint.

squeed:
If this PR doesn't merge shortly, I'd like to pull that RestMapper out separately; @JacobTanenbaum needs it for #89.

danwinship:
Failures look mostly like the `openshift-image-registry` stuff I noted in SDN-284. The question is, is this just because the tests are doing weird things (pushing images to the registry from a pod?) or do we actually need to put `openshift-image-registry` into the global network for normal cluster operation (even though we apparently don't need it during the course of bringing up the cluster).

squeed:
Yes, reading the documentation, it seems the registry should be globally reachable.

squeed:
Ah, the tests that are failing are all NetworkPolicy tests.

danwinship:
We're running the upstream NetworkPolicy tests now? We should get rid of our duplicate copy of them then...
and yes, `Services should preserve source pod IP for traffic thru service cluster IP` is also expected to fail with multitenant. So yeah, it looks like we're good now. Uncommit the hack and I'll lgtm it.

squeed:
Hack uncommitted - Networkpolicy remains the default for now.
/hold cancel

squeed:
Looks like a flake.
/retest

PR #84, opened 2019-01-31
add additional NetworkType openshiftSRIOV

zshi-redhat:


danwinship:
I assume this is post-4.0, and not covered by the multus exception?
/hold

zshi-redhat:
> I assume this is post-4.0, and not covered by the multus exception?
> /hold

This is not covered by Multus exception.
We'd like to have it supported as Tech Preview in 4.0, this is not enabled by default.

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -179,6 +182,9 @@ const (
 
 	// NetworkTypeRaw
 	NetworkTypeRaw NetworkType = "Raw"
+
+	// NetworkTypeOpenShiftSRIOV
+	NetworkTypeOpenshiftSRIOV NetworkType = "OpenshiftSRIOV"
```
squeed: OpenShift b/c Dan Winship is watching :-)

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -110,6 +110,9 @@ type AdditionalNetworkDefinition struct {
 	// The name of the network. This will be populated in the resulting CRD
 	Name string `json:"name"`
 
+	// The resourceName of SR-IOV device. This will be populated in the resulting CRD
+	ResourceName string `json:"resourceName"`
```
squeed: We need to document the AdditionalNetwork block in the README - could you please add a basic section that matches the other configuration variables?

bindata/network/additional-networks/openshift-sriov/sriov-cni.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
squeed: Does it make sense to put this as another container in the multus daemonset?

squeed: That would be my preference.

danwinship: (also, not really related to this PR, but would it make sense to copy the multus initcontainers into the openshift-sdn daemonset, so that if you're running openshift-sdn, we don't actually need to deploy the multus daemonset too?)

pkg/apis/networkoperator/v1/networkconfig_types.go
zshi-redhat: Just realize that we might not want to expose this ResourceName in current release.
Reason is that:
ResourceName specified in network CRD shall match with resourceName exposed by device plugin, but before device plugin is launched, there is no way for network-admin to know which resourceName shall be used. In current release, all devices found by SR-IOV device plugin will be exposed as the same resourceName (for example, openshift.com/sriov), so perhaps we could set the resourceName in the code when openshiftSRIOV network type is specified and not to bother network-admin to set a fixed value of resourceName.

bindata/network/additional-networks/openshift-sriov/sriov-cni.yaml
zshi-redhat: If Multus is not enabled, do we support using openshift-sdn together with CNI plugins from those initcontainers?  If not,  then there might be less willingness to copy those CNI plugins in openshift-sdn.

bindata/network/additional-networks/cr/cr.yaml
```diff
@@ -3,6 +3,10 @@ kind: NetworkAttachmentDefinition
 metadata:
   name: {{.AdditionalNetworkName}}
   namespace: default
+{{if .ConfigAnnotations}}
+  annotations:
+    k8s.v1.cni.cncf.io/resourceName: openshift.com/sriov
```
squeed: Shouldn't this be `openshift.io/sriov`?

zshi-redhat: Functionally it doesn't matter whether the resourceName suffix is ``com`` or ``io``,  but I agree that ``openshift.io`` would be more appropriate name used in operator, going to make the change.

bindata/network/additional-networks/openshift-sriov/sriov-cni.yaml
zshi-redhat: I agree that this could be achieved easily by passing a env var to enable/disable multus deployment. Considering later when adding ovn-kubernetes support, it may require adding multus initcontainers there as well because ovn could be one of the default network plugins.

bindata/network/additional-networks/openshift-sriov/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,53 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: sriov-device-plugin
+        component: network
+        type: infra
+        openshift.io/component: network
+      annotations:
+        scheduler.alpha.kubernetes.io/critical-pod: ''
```
squeed: You can get rid of this annotation; it doesn't have any effect anymore.

pkg/apis/networkoperator/v1/networkconfig_types.go
squeed: Just need to fix the capitalization.

squeed:
It looks like when you use a SRIOV network, you still need to create your own CNI configuration in the RawCNIConfiguration. That's not how I'd intended for this to work (though I could be convinced)

I would rather that when you do NetworkTypeSRIOV, it generates the NetworkAttachmentDefinition itself, including the CNI configuration json.

zshi-redhat:
> It looks like when you use a SRIOV network, you still need to create your own CNI configuration in the RawCNIConfiguration. That's not how I'd intended for this to work (though I could be convinced)
> 
> I would rather that when you do NetworkTypeSRIOV, it generates the NetworkAttachmentDefinition itself, including the CNI configuration json.

Ok, I'll create separate templates in bindata for openshift-sriov, including both net-attach-def CRD and SR-IOV CR.

pkg/apis/networkoperator/v1/networkconfig_types.go
zshi-redhat: Done

bindata/network/additional-networks/openshift-sriov/sriov-device-plugin.yaml
zshi-redhat: Removed

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -179,6 +179,9 @@ const (
 
 	// NetworkTypeRaw
 	NetworkTypeRaw NetworkType = "Raw"
+
+	// NetworkTypeOpenShiftSRIOV
+	NetworkTypeOpenShiftSRIOV NetworkType = "OpenShiftSRIOV"
```
booxter: why is it specifically OpenShift? What's the difference from a regular SR-IOV network that is managed by upstream CNI/DP? Is it the hardcoded openshift.io prefix?

zshi-redhat: There are two slight differences between openshift SRIOV and upstream ones:
1) The resource name advertised via openshift SRIOV DP is called 'openshift.io/sriov'.
2) Limited set of NICs supported in openshift SRIOV DP.
Besides above, there is no major difference on the design and architecture. Openshift 4.0 and 4.1 releases are based on upstream release-v1 branch (both DP and CNI).

squeed:
Good news: The network operator API has moved to openshift/api. This means that proposing API changes is kind of annoying now.

The way you have to do it is to (sigh) actually edit `vendor/github.com/openshift/api/network/v1/types_network.go` in a PR with your changes.

squeed:
In light of the fact that we don't know how to set up SRIOV without the user specifying a CNI configuration blob, let's just make that part of the configuration.

So the SRIOVConfig would have a CNIConfig member, to make it obvious that it was required.

zshi-redhat:
> In light of the fact that we don't know how to set up SRIOV without the user specifying a CNI configuration blob, let's just make that part of the configuration.
> 
> So the SRIOVConfig would have a CNIConfig member, to make it obvious that it was required.

Added ``SRIOVCNIConfig`` as a member of ``AdditionalNetworkDefinition`` to avoid using of ``rawCNIConfig`` for SR-IOV related network configuration;  With this change, SR-IOV configuration in network operator will be separated from Raw CNI configuration in terms of NetworkType, CNI Config, Validation, Rendering etc. There will be two supported types of additional Networks, one is ``Raw``, the other is ``OpenShiftSRIOV``.

zshi-redhat:
> Good news: The network operator API has moved to openshift/api. This means that proposing API changes is kind of annoying now.
> 
> The way you have to do it is to (sigh) actually edit `vendor/github.com/openshift/api/network/v1/types_network.go` in a PR with your changes.

Thanks for the pointer @squeed !
Do I need to send a separate PR in ``openshift/api`` for the API change in ``vendor/github.com/openshift/api/network/v1/types_network.go``?

squeed:
> Do I need to send a separate PR in `openshift/api` for the API change in `vendor/github.com/openshift/api/network/v1/types_network.go`?

The process is this (which I'll document in the README):
1. File your PR against the operator, but manually edit the vendored API type. This is so we can see the proposed change as a whole. 
2. Once the API changes look good, you file the PR against `openshift/api` and get that merged
3. You update your original to vendor in the "real" `openshift/api` types.

Yeah, a bit complicated, but so bad.

pkg/network/additional_networks.go
```diff
@@ -57,3 +58,65 @@ func validateRaw(conf *operv1.AdditionalNetworkDefinition) []error {
 
 	return out
 }
+
+// renderOpenShiftSRIOVNetworkCRD returns the manifests of OpenShiftSRIOV NetworkAttachmentDefinition.
+func renderOpenShiftSRIOVNetworkCRD(manifestDir string) ([]*uns.Unstructured, error) {
```
squeed: can you just combine these in to a single function `renderOpenShiftSRIOV`?

pkg/network/additional_networks.go
```diff
@@ -57,3 +58,65 @@ func validateRaw(conf *operv1.AdditionalNetworkDefinition) []error {
...
+func validateOpenShiftSRIOV(conf *operv1.AdditionalNetworkDefinition) []error {
+	out := []error{}
+	var SRIOVConfig map[string]interface{}
+	var err error
+
+	if conf.Name == "" {
+		out = append(out, errors.Errorf("Additional Network Name cannot be nil"))
+	}
+
+	confBytes := []byte(conf.SRIOVCNIConfig)
```
squeed: Can you add a specific check for an empty CNI config?

pkg/network/multus.go
```diff
@@ -17,6 +17,7 @@ func renderMultusConfig(manifestDir string) ([]*uns.Unstructured, error) {
 	data := render.MakeRenderData()
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
 	data.Data["MultusImage"] = os.Getenv("MULTUS_IMAGE")
+	data.Data["SRIOVCNIImage"] = os.Getenv("SRIOV_CNI_IMAGE")
```
squeed: Is this still necessary?

vendor/github.com/openshift/api/operator/v1/types_network.go
```diff
@@ -111,6 +111,10 @@ type AdditionalNetworkDefinition struct {
 	// rawCNIConfig is the raw CNI configuration json to create in the
 	// NetworkAttachmentDefinition CRD
 	RawCNIConfig string `json:"rawCNIConfig"`
+
+	// SRIOVCNIConfig is the OpenShiftSRIOV CNI configuration json to create in the
+	// NetworkAttachmentDefinition CRD
+	SRIOVCNIConfig string `json:"SRIOVCNIConfig"`
```
squeed: You should create a `*struct SRIOVConfig` with a single member ` CNIConfig`

pkg/network/additional_networks.go
```diff
@@ -10,12 +11,12 @@ import (
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// renderAdditionalNetworksCRD returns the manifests of the NetworkAttachmentDefinition.
-func renderAdditionalNetworksCRD(manifestDir string) ([]*uns.Unstructured, error) {
+// renderRawNetworkCRD returns the manifests of the NetworkAttachmentDefinition.
+func renderRawNetworkCRD(manifestDir string) ([]*uns.Unstructured, error) {
```
squeed: I don't think this should change. Instead, we should always render the CRD whenever multus is enabled. There should only be one instance of the CRD in `bindata/`. So, I would set up the directory structure like this:

```
/bindata/additional-networks/crd/000-network-attachment-crd.yaml
/bindata/additional-networks/raw/network-attachment.yaml
/bindata/additional-networks/sriov/network-attachment.yaml (plus the other sr-iov files)
```

zshi-redhat: Done

vendor/github.com/openshift/api/operator/v1/types_network.go
zshi-redhat: Done

pkg/network/multus.go
zshi-redhat: Removed from ``renderMultusConfig`` and be rendered in ``renderOpenShiftSRIOV``.

pkg/network/additional_networks.go
zshi-redhat: Done

pkg/network/additional_networks.go
zshi-redhat: Done

bindata/network/additional-networks/sriov/sriov-cni.yaml
```diff
@@ -0,0 +1,42 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: sriov-cni
+  namespace: openshift-sriov
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the SR-IOV CNI plugin on each node.
```
squeed: "This daemonset copies the SR-IOV CNI plugin on to each node"

pkg/network/additional_networks_test.go
```diff
@@ -60,3 +67,49 @@ func TestValidateRaw(t *testing.T) {
...
+	g := NewGomegaWithT(t)
+
+	for _, cfg := range NetworkAttachmentConfig.Spec.AdditionalNetworks {
+		switch cfg.Type {
+		case operv1.NetworkTypeOpenShiftSRIOV:
+			err := validateOpenShiftSRIOV(&cfg)
+			g.Expect(err).To(BeEmpty())
+		}
+	}
+
```
squeed: You should do a `NetworkAttachmentConfig.DeepCopy()` here, and then mutate it locally for the test.

bindata/network/additional-networks/sriov/sriov-cni.yaml
zshi-redhat: Done

pkg/network/additional_networks_test.go
zshi-redhat: Done

zshi-redhat:
@dcbw @knobunc please review this SR-IOV integration in network operator, Thanks!

bindata/network/additional-networks/sriov/sriov-cni.yaml
```diff
@@ -0,0 +1,42 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: sriov-cni
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
```
booxter: why do we need hostNetwork for a container that merely copies a file to host directory?

bindata/network/additional-networks/sriov/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,53 @@
...
+  template:
+    metadata:
+      labels:
+        app: sriov-device-plugin
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
+      hostPID: true
```
booxter: why do we need hostPID for the plugin container?

bindata/network/additional-networks/sriov/sriov-device-plugin.yaml
```diff
@@ -0,0 +1,53 @@
...
+        operator: Exists
+        effect: NoSchedule
+      serviceAccountName: sriov-device-plugin
+      containers:
+      - name: sriov-device-plugin
+        image: {{.SRIOVDevicePluginImage}}
+        args:
+        - --log-level=10
+        securityContext:
+          privileged: true
```
booxter: why is device plugin privileged? it doesn't seem to require any meddling with the host, it only discovers and advertises devices so I would think it should work without the privilege escalated? (Perhaps it's because of the bug in openshift installer where we don't apply proper SELinux labels to `/var/lib/kubelet/device-plugins` directory? In which case, maybe the right course is fixing the installer / SELinux packages?)

manifests/0000_07_cluster-network-operator_03_daemonset.yaml
```diff
@@ -37,6 +37,10 @@ spec:
           value: "quay.io/openshift/origin-container-networking-plugins-supported:v4.0.0"
         - name: CNI_PLUGINS_UNSUPPORTED_IMAGE
           value: "quay.io/openshift/origin-container-networking-plugins-unsupported:v4.0.0"
+        - name: SRIOV_CNI_IMAGE
```
booxter: These images are too old for KubeVirt / CNV. Is it possible to configure the operator to deploy a different version of the images? Or perhaps disable SR-IOV deployment by this operator so that we can handle deployment of the right components in our own network cluster operator?

This wouldn't be an issue if we would treat OpenShift and upstream versions of the same plugin as different plugins with their own names. Then you would have one CRD for OpenShift with e.g. `type: openshift-sriov` and another with more modern version of the upstream plugin with `type: sriov`. But then we would hit the problem with the old version of the plugin that consumes all VFs found on the node, regardless of wishes of other components (modern plugin can use just part of available VFs). Regardless, it's just not elegant to manually split VFs between two versions of the same plugin, prone to errors and under-utilization.

So the only solution I see is to be able to tell OpenShift operator to not deploy any SR-IOV bits (no CRD, no CNI, no DP), and handle deployment ourselves. Is it something we can consider? If not, it's probably a blocker for CNV SR-IOV integration into OpenShift. Thoughts?

zshi-redhat: I think the cluster operator doesn't control which versions of SR-IOV components are deployed, it's managed by upper layer operator to replace the image version of CNI, DP for upgrading. However, it is possible to disable SR-IOV deployment in network operator, in fact, it's not enabled by default, it will only be deployed when additional network type ``OpenShiftSRIOV`` is specified and corresponding ``SRIOVConfig`` is provided. Re CRD creation, current implementation uses a separate template for SR-IOV CR rendering, it's also triggered when ``OpenShiftSRIOV`` network type is used, but these don't block user creating their own CR with different templates.
With above, I think it's possible for CNV deployment to use a different set of SR-IOV manifests or templates, just don't try to have two sets of SR-IOV components deployed in the same cluster.

This may improve later with OpenShift 4.2, see SR-IOV roadmap [here] (https://docs.google.com/document/d/1lKJjv7ufuMHL_31jPiVEVu6LnlRz8y6Aw-0gV8yY8hc/edit#) when SR-IOV support updates to use latest master.

bindata/network/additional-networks/sriov/sriov-cni.yaml
zshi-redhat: this is not necessary, will remove

bindata/network/additional-networks/sriov/sriov-device-plugin.yaml
zshi-redhat: will remove

bindata/network/additional-networks/sriov/sriov-device-plugin.yaml
zshi-redhat: Yes, I think the PR upstream that fixing selinux problem of device-plugins dir is not merged yet.

booxter: You refer to https://github.com/kubernetes/kubernetes/pull/73241 right? Yes it's not merged; if this is the reason why this manifest carries `privileged: true` then maybe we should tag it with a comment as something that we should revisit later when the fix is merged upstream and synced into OpenShift.

zshi-redhat:
@danwinship please help to review if this PR looks better now, Thanks!

danwinship:
@dcbw PTAL; I've updated it to no longer have a separate `NetworkTypeSRIOV`

I would like to merge the daemonsets into the existing multus daemonsets like I suggested in the DHCP PR, but that can happen later

PR #85, opened 2019-02-01
do not merge: extra commit to test ci

squeed:
Just want to make sure we still run e2e correctly.


squeed:
Good, e2e-aws was triggered.

PR #86, opened 2019-02-01
Allow DefaultNetwork.Type to have arbitrary values.

squeed:
The intention is for users of third-party network plugins to still be able to set their network type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,10 @@ Users must select a default network provider. This cannot be changed. Different 
 
 The network type is always read from the Cluster configuration.
 
-Currently, the only supported value for network Type is `OpenShiftSDN`.
+Currently, the only understood value for network Type is `OpenShiftSDN`.
+
+Other values are ignored. If you wish to use use a third-party network provider not managed by the operator, set the network type to something meaningful to you. The operator will not install or upgrade a network provider, but all other Network Operator functionality remains.
+
 
 ### Configuring OpenShiftSDN
 OpenShiftSDN supports the following configuration options, all of which are optional:
--- a/pkg/network/render_test.go
+++ b/pkg/network/render_test.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+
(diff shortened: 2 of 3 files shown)


pkg/network/render_test.go
```diff
@@ -33,3 +35,48 @@ func TestIsChangeSafe(t *testing.T) {
...
+					CIDR:             "10.128.0.0/15",
+					HostSubnetLength: 9,
+				},
+				{
+					CIDR:             "10.0.0.0/14",
+					HostSubnetLength: 8,
+				},
+			},
+			DefaultNetwork: netv1.DefaultNetworkDefinition{
+				Type: "MyAwesomeThirtPartyPlugin",
```
danwinship: Yes, it doesn't really matter, but I know the typo there ("`Thirt`") is eventually going to annoy me...

README.md
```diff
@@ -94,7 +94,7 @@ Users must select a default network provider. This cannot be changed. Different
 
 The network type is always read from the Cluster configuration.
 
-Currently, the only supported value for network Type is `OpenShiftSDN`.
+Currently, the only supported value for network Type is `OpenShiftSDN`. *All other network types will be ignored.* If you use a third-party network provider that is not managed by the operator, you should set your network type to something meaningful to you.
```
danwinship: Clarify "ignored". Is it copied from the cluster config to the operator config? Is the rest of the config still validated?

pkg/network/render_test.go
squeed: We can't have that.

README.md
squeed: Good point; clarified.

squeed:
OK, updated. @danwinship, PTAL.

danwinship:
Should we try to do something about the case where they specify `OpenShiftSND` or `OpenshiftSDN` or (everyone's favorite typo) `OpenShitSDN`? Maybe add a `Message` to the `Available` operator status clarifying that the operator doesn't recognize the plugin name and so hasn't deployed a plugin?

danwinship:
/lgtm
Casey is going to add another card about having a clear status here

PR #89, opened 2019-02-06
configure endpoints to expose openshift-sdn metrics to Prometheus

JacobTanenbaum:
Add the required service, serviceMonitor, role, and rolebinding to allow Prometheus
to scrape the metrics that the sdn provides

### DIFF
--- a/bindata/network/openshift-sdn/monitor.yaml
+++ b/bindata/network/openshift-sdn/monitor.yaml
@@ -1,0 +1,70 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: sdn
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: monitor-sdn
+  namespace: openshift-sdn
+spec:
+  endpoints:
+  - interval: 30s
+    port: metrics
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+    - openshift-sdn
+  selector:
+    matchLabels:
+      app: sdn
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: sdn
+  name: sdn
+  namespace: openshift-sdn
+spec:
+  selector:
+    app: sdn
+  clusterIP: None
+  ports:
+  - name: metrics
+    port: 9101
+    protocol: TCP
+    targetPort: 9101
+  sessionAffinity: None
+  type: ClusterIP
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-sdn
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - services
+  - endpoints
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-sdn
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: openshift-monitoring
--- a/bindata/network/openshift-sdn/000-ns.yaml
+++ b/bindata/network/openshift-sdn/000-ns.yaml
@@ -5,6 +5,7 @@ metadata:
   labels:
     name: openshift-sdn
     openshift.io/run-level: "0"
+    openshift.io/cluster-monitoring: "true"
   annotations:
     openshift.io/node-selector: "" #override default node selector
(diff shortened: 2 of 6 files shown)


bindata/network/openshift-sdn/monitor.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: sdn
+  name: monitor-sdn
+  namespace: openshift-sdn
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
squeed: Given that the metrics aren't behind TLS, we don't need the `bearerTokenFile` or the `tlsConfig`

squeed:
I'm pretty sure we can't use the trick that @abhinavdahiya suggested in chat, because we need to run to completion before the monitoring operator will even run. So I would really like to see us ignore CRD-doesn't-exist failures. 

You can test this by spinning up a test cluster, stopping the CVO, then deleting the openshift-monitoring namespace and the ServiceMonitor CRD. When you run your local version of the network operator, I would expect the Operator.Status field to be successful, but I fear it would be failing.

squeed:
@brancz would you (or someone on your team) please the SeriviceMonitor a quick look?

bindata/network/openshift-sdn/monitor.yaml
```diff
@@ -0,0 +1,74 @@
...
+    honorLabels: true
```
brancz: Do you actually need this? Typically this is only required if you have things like the `namespace`, `pod`, `service`, `job` or `instance` label in the metrics you're exposing. And if you do, we should probably talk about that as well :slightly_smiling_face: .

bindata/network/openshift-sdn/monitor.yaml
```diff
@@ -0,0 +1,74 @@
...
+    interval: 30s
+    path: /metrics
```
brancz: this is the default, no need to specify explicitly

bindata/network/openshift-sdn/monitor.yaml
```diff
@@ -0,0 +1,74 @@
...
+    port: metrics
+    scheme: http
```
brancz: this is the default, no need to specify explicitly

squeed:
Looks like you'll have to update one of the unit tests too, you can run those locally with `go test ./pkg/...`

JacobTanenbaum:
screenshot_2019-02-07_14-34-47

squeed:
Checked the logs, everything is as expected:
```
2019/02/07 19:35:56 reconciling (monitoring.coreos.com/v1, Kind=ServiceMonitor) openshift-sdn/monitor-sdn
2019/02/07 19:35:56 could not apply (monitoring.coreos.com/v1, Kind=ServiceMonitor) openshift-sdn/monitor-sdn: could not retrieve existing (monitoring.coreos.com/v1, Kind=ServiceMonitor) openshift-sdn/monitor-sdn: no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
2019/02/07 19:35:56 Object has ignore-errors annotation set, continuing
```

squeed:
Darn, it looks like the ServiceMonitor object is never applied. Looks like we'll have to do whatever the CVO does. It looks like it resets the RestMapper, which is the mapping from Group-Version-Kind to API objects.

brancz:
The screenshot looks good, I'll let you take care of the rest :)

squeed:
Had a bit of a think, I think we're going to have to do something cleverer. My idea: add an annotation that tells `apply.go` to retrieve the CRD first. Something like `networkoperator.openshift.io/check-for-crd: skip`. Then, before it does a `.Get()`, it checks that the CRD is on the server and skips if not.

JacobTanenbaum:
rebased with PR https://github.com/openshift/cluster-network-operator/pull/95 
as the RESTMapper is now dynamic the CRD creation time is not an issue

squeed:
Now that everything should be working, can you check in the e2e artifacts that everything as expected:
- Check the network operator logs for weird errors
- Download the saved prometheus metrics and see that the kube-proxy ones are showing up

Once you've done that, I'll lgtm.

squeed:
woohoo:
```
2019/02/11 17:03:45 reconciling (monitoring.coreos.com/v1, Kind=ServiceMonitor) openshift-sdn
/monitor-sdn
2019/02/11 17:03:45 does not exist, creating (monitoring.coreos.com/v1, Kind=ServiceMonitor) 
openshift-sdn/monitor-sdn
2019/02/11 17:03:45 successfully created (monitoring.coreos.com/v1, Kind=ServiceMonitor) open
shift-sdn/monitor-sdn
```

JacobTanenbaum:
\test e2e-aws

squeed:
er,
/approve

squeed:
/approve
one last try.

danwinship:
/approve
?

PR #90, opened 2019-02-06
bindata: use system-node-critical priority

sjenning:
This in combination with https://github.com/openshift/origin/pull/21978 should help avoid OOM of sdn and ovs pods

xref https://github.com/openshift/cluster-network-operator/pull/80

@derekwaynecarr @danwinship @dcbw @squeed


squeed:
Is there any harm in merging this now without the corresponding origin PR? I don't think so, just checking.

PR #91, opened 2019-02-07
controller: allow objects to signify errors should be ignored.

squeed:
This is most useful when we want to create an instance of a CRD, but the definition itself may not yet exist. We shouldn't necessarily ignore that in all cases, but it's useful for non-critical objects.


squeed:
cc @JacobTanenbaum

squeed:
Nevermind, Jacob is just going to pull this in to #89. Jacob, feel free to close this PR.

PR #94, opened 2019-02-08
hack: add a development way to run the operator as part of the installer

squeed:
This is a stupid way to tell the installer not to run the network operator, and let us run it manually as part of the install process.

### DIFF
--- a/INSTALLER-HACKING.md
+++ b/INSTALLER-HACKING.md
@@ -1,0 +1,61 @@
+# Running local builds with the installer
+
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator.
+
+This requires `oc` and `kubectl` in your $PATH.
+
+## Steps
+
+You will need to execute all steps for every installation, since the installer deletes the intermediate files we customize.
+
+1. In one window, prepare the cluster:
+
+```
+export CLUSTER_DIR=<your cluster directory>
+openshift-install --dir=$CLUSTER_DIR create manifests
+```
+
+2. In another window, prepare the operator. This will apply some overrides that disable the default cluster-network-operator. It will also extract the release image.
+
+```
+export CLUSTER_DIR=<your cluster directory>
+hack/run-locally.sh prepare
+hack/build-go.sh
+```
+
+3. Optionally, override image references. If you are doing downstream image development, e.g. working on openshift-sdn, you should build, publish, and edit `$CLUSTER_DIR/env.sh` to point to your development image.
+
+4. In the installer window, create the cluster *and move on to the next step*
+
+```
+openshift-install --dir=$CLUSTER_DIR create cluster
+```
+
+5. In the operator window, *while the installer is running*, launch the operator locally
+```
+hack/run-locally.sh start
+```
+
+
+## Tips, Tricks, & Limitations
+
+### Cleaning up
+After destroying your cluster, you should delete your `$CLUSTER_DIR`.
+
+If you're using libvirt, you can use `./scripts/maintenance/virsh-cleanup.sh` in the installer repository to quickly delete the virtual machines instead of waiting for Terraform.
+
+### RBAC
+This runs the operator with the `admin` role, so it will have full permissions. When run by the real installer, this will not be the case. Created objects, e.g. DaemonSets, will still have RBAC, though.
+
+### Saving install-config between runs
+If you're sick of answering all the questions over and over, you can shortcut the installer. Just do
+```
+openshift-install --dir=$CLUSTER_DIR create install-config
+cp $CLUSTER_DIR/install-config.yaml ~/.cache/openshift-install/install-config.yaml
+```
+
+Then, for subsequent installer runs, you can do
+```
+rm -r $CLUSTER_DIR; mkdir -p $CLUSTER_DIR
+cp ~/.cache/openshift-install/install-config.yaml $CLUSTER_DIR/
+```
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -1,0 +1,107 @@
+#!/bin/bash 
+
+# run-locally.sh is some scaffolding around running a local instance of the
+# network operator with the installer.
+# See INSTALLER-HACKING.md
+
+set -o errexit
+set -o nounset
+
+# Install our overrides so the cluster doesn't run the network operator.
+function override() {
+    if [[ ! -e ${CLUSTER_DIR}/manifests/cvo-overrides.yaml ]]; then
+        echo "cannot find cvo-overrides.yaml; please run"
+        echo "openshift-install --dir=${CLUSTER_DIR} create manifests"
+        exit 1
+    fi
+
+    # Patch the CVO to not create the network operator
+    echo "Applying overrides to ${CLUSTER_DIR}/manifests/cvo-overrides.yaml"
+
+    kubectl --kubeconfig=hack/null-kubeconfig patch --type=json --local=true -f=${CLUSTER_DIR}/manifests/cvo-overrides.yaml -p "$(cat hack/overrides-patch.yaml)" -o yaml > ${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new
+
+    mv ${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new ${CLUSTER_DIR}/manifests/cvo-overrides.yaml
+}
+
+# Extract the image references from the release image.
+function build_env() {
+    echo "Writing image references to ${CLUSTER_DIR}/env.sh"
+    oc --kubeconfig=hack/null-kubeconfig convert --local=true -f manifests/0000_07_cluster-network-operator_03_daemonset.yaml -ojsonpath='{range .spec.template.spec.containers[0].env[?(@.value)]}{.name}{"="}{.value}{"\n"}' > ${CLUSTER_DIR}/env.sh
+}
+
+# start_operator waits for the cluster to come up, then launches
+# the operator locally.
+function start_operator() {
+    export KUBECONFIG=${CLUSTER_DIR}/auth/kubeconfig
+
+    echo "Waiting for $KUBECONFIG"
+    while true; do 
+        if [[ ! -e ${KUBECONFIG} ]]; then
+            sleep 10
+        else
+            break
+        fi
+    done
+
+    # A few environment variables we set in the Daemonset
+    export POD_NAME=LOCAL
+    export KUBERNETES_SERVICE_PORT=6443
+    export KUBERNETES_SERVICE_HOST=$(oc config view -o jsonpath='{.clusters[0].cluster.server}' | awk -F'[/:]' '{print $4}')
+
+    echo "Waiting for the apiserver to come up and for the network configuration to be rceated."
+    echo "You can ignore the error message 'error: the server doesn't have a resource type \"Network\"'"
+
+    while true; do 
+        if oc get Network.config.openshift.io cluster; then
+            echo "Network config exists, continuing"
+            break
+        else
+            echo "No network config or apiserver not up yet, retrying"
+            sleep 15
+            continue
+        fi
+    done
+    echo "Starting operator"
+    env $(cat ${CLUSTER_DIR}/env.sh) _output/linux/amd64/cluster-network-operator
+}
+
+function usage() {
+    >&2 echo "Usage: $0 (prepare|start)
+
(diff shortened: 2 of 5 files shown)


INSTALLER-HACKING.md
```diff
@@ -0,0 +1,63 @@
+# Running local builds with the installer
+
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you.
+
+## Step 0: Prepare your credentials
+
+You will need to do this once, and every time your token resets.
```
danwinship: This sentence is very mysterious on the first read-through...

And are you sure this is needed? I just tried `podman pull registry.svc.ci.openshift.org/openshift/origin-release:v4.0` and it worked without me logging in.

Anyway, if it is sometimes necessary, I was going to suggest that this could just be moved into the script; if the `podman pull` fails, ask the user to go to the CI server console, copy the login command, and paste it into the terminal. Then the script can extract the token from that and run `podman login` and then try again.

INSTALLER-HACKING.md
```diff
@@ -0,0 +1,63 @@
...
+```
+podman login -u <YOUR GITHUB ID> -p <YOUR TOKEN> registry.svc.ci.openshift.org
+```
+
+Tokens expire regularly - about every week.
+
+
+## Step 1: Prepare the cluster & environment
+
+You will need to do this for every new cluster
```
danwinship: After I destroy the cluster, can I run again from the same `$CLUSTER_DIR`? If so, which parts of this can I skip the second time around?

INSTALLER-HACKING.md
```diff
@@ -0,0 +1,63 @@
...
+3. Optional: Override images
+If you are doing downstream image development, e.g. working on openshift-sdn or the like, you should build and publish a custom image now. Then, edit `env.sh` to point to your development image
+
+## Step 2: Create the cluster
+
+1. In the installer window, 
+
+```
+openshift-install --dir=$CLUSTER_DIR create cluster
+```
```
danwinship: ... and don't wait for it to finish; the install can't complete until you bring up the network operator, right?

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash 
+
+if [ -z "$CLUSTER_DIR" ]; then
```
danwinship: I'm not sure this ever all got written out anywhere, but preferred OpenShift shell-scripting style uses `[[`/`]]` and `${CLUSTER_DIR}`. (Likewise throughout.)

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
...
+    echo CLUSTER_DIR is required
+    exit 1
+fi
+
+set -o errexit
+set -o nounset
+
+# Install our overrides so the cluster doesn't run the network operator.
+function override() {
+    local cluster_dir=$1
```
danwinship: You use this variable for the first half of the function but then fall back to `$CLUSTER_DIR` in the second half. Probably no need to pass this in as an argument, just use the global everywhere.

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
...
+    echo "Pulling the release image $RELEASE_IMAGE"
+    echo "If this fails, you'll need to log in."
+    podman pull $RELEASE_IMAGE
+
+    echo > env.sh
+
+    echo
+    echo "Extracting image references to env.sh"
+
+    # If I were cleverer, I'd pull this from the release playload.
```
danwinship: typo "playload"

(also, while we're wishing we were cleverer, it would also be nice to dynamically figure out which image references we need, like by sucking the names out of `manifests/0000_07_cluster-network-operator_03_daemonset.yaml` or whatever)

pecameron:
@squeed Is this the way to develop the ovnkube type? I install and by the time I get to log in networking is setup and fixed. Can't change type.

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
...
+
+
+case "${1:-""}" in
+    prepare)
+        override $CLUSTER_DIR;
+        build_env;
+        ;;
+    start)
+
+
```
danwinship: (gratuitous blank lines)

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
...
+    if [ ! -e $cluster_dir/manifests/cvo-overrides.yaml ]; then
+        echo "cannot find cvo-overrides.yaml; please run"
+        echo "openshift-install --dir=$cluster_dir create manifests"
+        exit 1
+    fi
+
+    # Patch the CVO to not create the network operator
+    echo "Applying overrides to $CLUSTER_DIR/manifests/cvo-overrides.yaml"
+
+    kubectl patch --type=json --local=true -f=$CLUSTER_DIR/manifests/cvo-overrides.yaml -p "$(cat hack/overrides-patch.yaml)" -o yaml > $CLUSTER_DIR/manifests/.cvo-overrides.yaml.new
```
danwinship: Even though it doesn't need it, this will fail if you don't have `$KUBECONFIG` pointing to a valid kubeconfig file. Apparently a minimal kubeconfig, which you can point to with `--kubeconfig` is:

    clusters:
    - cluster:
        server: http://localhost

hack/run-locally.sh
```diff
@@ -0,0 +1,121 @@
...
+    echo "If this fails, you'll need to log in."
+    podman pull $RELEASE_IMAGE
+
+    echo > env.sh
+
+    echo
+    echo "Extracting image references to env.sh"
+
+    # If I were cleverer, I'd pull this from the release playload.
+    echo "export NODE_IMAGE=$(podman run --rm $RELEASE_IMAGE image node)" >> env.sh
```
danwinship: I get an error printed for each of these commands:

    ERRO[0001] could not find slirp4netns, the network namespace won't be configured: exec: "slirp4netns": executable file not found in $PATH 

The script works, but the errors make it seem like something's wrong. Adding `--net none` to the `podman run` command suppresses the error

INSTALLER-HACKING.md
squeed: Looks like it was only needed for me, because I logged in to push to my repository, and the token expired. Neat.

hack/run-locally.sh
squeed: Looks like you can even do `KUBECONFIG=/dev/null`

danwinship: I get `error: no configuration has been provided` with that; the validation function appears to want a config containing at least one cluster. (This may depend on your kubectl version?)

squeed:
Updated, now with 100% more cleverness. Now we automatically extract the image references.

hack/run-locally.sh
squeed: Weirdly enough, `/dev/null` works for `kubectl patch` and `oc adm release extract`, but not for `oc convert`.  Buhhh.

.gitignore
```diff
@@ -3,6 +3,7 @@
 tmp/_output
 tmp/_test
 _output
+env.sh
```
danwinship: is there any reason this can't be dumped into `$CLUSTER_DIR`?

INSTALLER-HACKING.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+## Tips, Tricks, & Limitations
+
+### Cleaning up
+After destroying your cluster, you should delete your `$CLUSTER_DIR`. If you're using libvirt, you can use `./scripts/maintenance/virsh-cleanup.sh` in the installer repository.
+
+### RBAC
+This runs the operator with the `admin` role, so it will have full permissions. When run as the real installer, this will not be the case. So you can't debug RBAC issues here.
```
danwinship: "when run **by** the real installer"?

FWIW we're going to have to figure out how to fix this fairly quickly, since part of the OVN work will be making sure we have RBAC set up correctly.

INSTALLER-HACKING.md
```diff
@@ -0,0 +1,66 @@
...
+rm -r $CLUSTER_DIR; mkdir -p $CLUSTER_DIR
+cp ~/.cache/openshift-install/install-config.yaml $CLUSTER_DIR/
+```
+
+### Waiting for the apiserver
+If you see the error message
+```
+error: the server doesn't have a resource type "Network"
+```
+don't fret - it's expected as part of the bootstrap process.
```
danwinship: maybe move this up to the `hack/run-locally.sh start` step?

hack/run-locally.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash 
+
+
+
```
danwinship: (gratuitous blank lines)

hack/run-locally.sh
```diff
@@ -0,0 +1,109 @@
...
+    if [[ ! -e ${CLUSTER_DIR}/manifests/cvo-overrides.yaml ]]; then
+        echo "cannot find cvo-overrides.yaml; please run"
+        echo "openshift-install --dir=${CLUSTER_DIR} create manifests"
+        exit 1
+    fi
+
+    # Patch the CVO to not create the network operator
+    echo "Applying overrides to ${CLUSTER_DIR}/manifests/cvo-overrides.yaml"
+
+    kubectl --kubeconfig=hack/null-kubeconfig patch --type=json --local=true -f=${CLUSTER_DIR}/manifests/cvo-overrides.yaml -p "$(cat hack/overrides-patch.yaml)" -o yaml > ${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new
```
danwinship: Since we're requiring `oc` elsewhere, change this to `oc` for consistency?

squeed: For some reason this doesn't work with `oc`. I don't remember why.

INSTALLER-HACKING.md
squeed: This only applies to the operator process. Any ovn processes will still go through rbac.

.gitignore
squeed: good point, done.

INSTALLER-HACKING.md
squeed: Also, "run as" -- German prepositions are leaking out again.

INSTALLER-HACKING.md
squeed: good idea.

squeed:
OK, updated.

danwinship:
> Updated, now with 100% more cleverness. Now we automatically extract the image references.

Ah, I had meant using the in-tree daemonset.yaml, not the one from the image, which is what you're using I guess? In particular, I was thinking about the fact that for OVN, the operator will also need to have `$OVN_IMAGE` set.

Maybe just add a comment about this to step 3?

squeed:
Mmm, I was thinking of a different use case. That's starting to sound like actual code ;-).

squeed:
Wait nevermind, that's really easy and obviously the right thing to do. Hold please.

squeed:
That makes it infinitely simpler, and you don't need a new `oc`. Updated.

hack/run-locally.sh
```diff
@@ -0,0 +1,107 @@
...
+
+    mv ${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new ${CLUSTER_DIR}/manifests/cvo-overrides.yaml
+}
+
+# Extract the image references from the release image.
+function build_env() {
+    echo "Writing image references to ${CLUSTER_DIR}/env.sh"
+    oc --kubeconfig=hack/null-kubeconfig convert --local=true -f manifests/0000_07_cluster-network-operator_03_daemonset.yaml -ojsonpath='{range .spec.template.spec.containers[0].env[?(@.value)]}{.name}{"="}{.value}{"\n"}' > ${CLUSTER_DIR}/env.sh
```
danwinship: oh. heh. I had assumed you had some reason for extracting the exact version numbers of out of the built image :-)

hack/run-locally.sh
```diff
@@ -0,0 +1,107 @@
...
+        fi
+    done
+
+    # A few environment variables we set in the Daemonset
+    export POD_NAME=LOCAL
+    export KUBERNETES_SERVICE_PORT=6443
+    export KUBERNETES_SERVICE_HOST=$(oc config view -o jsonpath='{.clusters[0].cluster.server}' | awk -F'[/:]' '{print $4}')
+
+    echo "Waiting for the apiserver to come up and for the network configuration to be rceated."
+    echo "You can ignore the error message 'error: the server doesn't have a resource type \"Network\"'"
```
danwinship: You can ignore the error message 'error: the server doesn't have a resource type "Network"'
    The connection to the server danwinship-api.devcluster.openshift.com:6443 was refused - did you specify the right host or port?

Anyway, we can make further tweaks to the instructions later

PR #95, opened 2019-02-11
Replace the controller manager's RESTMapper with a dynamic one

danwinship:
controller-manager assumes that all resource Kinds will have been created before it starts up, so it can't deal with the case of creating a CRD and then creating an instance of that CRD (or waiting for another operator to create a CRD, and then creating an instance of that CRD). This is maybe not the bestest fix, but it works and we can discuss it further upstream.

Note that this code would be especially bad/inefficient if we thought that the controller was going to be repeatedly asked about objects of non-existent types, but there's no reason that would happen in CNO, so the reload path should only get hit in the creating-an-instance-of-a-previously-created-CRD case.

### DIFF
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/cluster-network-operator/pkg/apis"
 	"github.com/openshift/cluster-network-operator/pkg/controller"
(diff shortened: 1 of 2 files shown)


squeed:
/lgtm
Thanks for tracking that one down.
FYI: @shawn-hurley, you might be interested in this. Perhaps this should be the default for the operator-sdk

PR #100, opened 2019-02-13
hack: add HACK_MINIMIZE command to make running on libvirt possible.

squeed:
This is an awful hack, but it is fast(er).

Also, wait for the network operator namespace to come in to existence before continuing.

### DIFF
--- a/hack/overrides-minimize-patch.yaml
+++ b/hack/overrides-minimize-patch.yaml
@@ -1,0 +1,31 @@
+# When doing rapid local development on libvirt, it's nice to remove
+# some of the more resource-heavy components.
+# To use this, set the environment variable HACK_MINIMIZE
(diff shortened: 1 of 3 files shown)


squeed:
I'm not particularly proud of this.

hack/overrides-minimize-patch.yaml
```diff
@@ -0,0 +1,31 @@
+# When doing rapid local development on libvirt, it's nice to remove
+# some of the more resource-heavy components.
+# To use this, set the environment variable HACK_MINIMIZE
```
danwinship: document in INSTALLER-HACKING.md?

squeed: I was a bit embarrassed, but done.

PR #102, opened 2019-02-14
Update to be able to track both DaemonSets and Deployments

danwinship:
OVN uses a deployment, so we'll need to be able to track a mix of Deployments and DaemonSets now.

(In fact, failing to track the Deployment doesn't actually affect anything with OVN, because the node DaemonSet can't become ready until after the master Deployment is ready anyway. But whatever.)

Not tested yet because we appear to be having AWS account issue, but you can still review the concept...


pkg/controller/networkconfig/networkconfig_controller.go
```diff
@@ -185,15 +189,24 @@ func (r *ReconcileNetworkConfig) Reconcile(request reconcile.Request) (reconcile
...
 			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
+			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		}
 	}
-	r.daemonSetReconciler.SetDaemonSets(daemonSets)
+	r.status.SetDaemonSets(daemonSets)
+	r.status.SetDeployments(deployments)
+
+	allResources := []types.NamespacedName{}
```
squeed: I realize this is a bit pathological, but you could totally have a NamespacedName collision between daemonsets and deployments.

squeed: Especially if you moved from one Kind to the other.

squeed: Ahh, now I see why that doesn't really matter.

squeed:
Seems reasonable to me.

In theory you could move the namespace checks to the Watch predicate, but that seems more error-prone - and certainly doesn't get any extra efficiency.

danwinship:
/hold
not sure why the bot removed my WIP label... but this is still not actually tested

pkg/controller/networkconfig/networkconfig_controller.go
danwinship: I started working on a controller-runtime PR to make it include a `GroupVersionKind` in `reconcile.Request`. But as you figured out, it doesn't actually matter here.

danwinship:
/hold
not needed for 4.0

danwinship:
/hold cancel
the rewrites here make fixing some of the other status stuff easier

PR #103, opened 2019-02-20
Config: rename some things to match Network.config.openshift.io

squeed:
- Rename ClusterNetworks to ClusterNetwork
- Make ServiceNetwork an array
- move HostSubnetLength to HostPrefix
- Rename the object name to "cluster"

These breaking changes are allowable as long as we get them merged by Friday.


README.md
```diff
@@ -81,8 +82,9 @@ Currently, changing the address pools once set is not supported. In the future,
 Example
 ```yaml
 spec:
-  serviceNetwork: "172.30.0.0/16"
-  clusterNetworks:
+  serviceNetwork:
+  - "172.30.0.0/16"
+  clusterNetwork:
     - cidr: "10.128.0.0/14"
       hostSubnetLength: 9
```
danwinship: need to update `hostSubnetLength` in this example

pkg/apis/networkoperator/v1/networkconfig_types.go
```diff
@@ -79,8 +81,8 @@ type NetworkConfigSpec struct {
 // 2^HostSubnetLength will be allocated when nodes join the cluster.
```
danwinship: comment here is wrong... you should probably grep for other stray references...

pkg/network/openshift_sdn_test.go
```diff
@@ -296,3 +296,136 @@ func TestOpenShiftSDNMultitenant(t *testing.T) {
 		g.Expect(objs).To(ContainElement(HaveKubernetesID("NetNamespace", "", ns)))
 	}
 }
+
+func TestOpenshiftControllerConfig(t *testing.T) {
```
danwinship: this and `TestOpenshiftNodeConfig` seem like an unrelated patch?

danwinship:
> These breaking changes are allowable as long as we get them merged by Friday.

This will break people using older versions of the installer though, right? Are we sure that's not a problem?
And if we are allowed to do that, then can you remove the deprecated improperly-camelcased versions of the config strings?

squeed:
Hm, I removed those deprecated strings, and it looks like a `go rename` reverted that somehow. Oops.

pkg/network/openshift_sdn_test.go
squeed: I just added it for parity with the node config, which I definitely wanted to test.

squeed:
OK, re-grepped and updated, thanks.

squeed:
And, yes, we don't need to worry about breaking old clusters.

danwinship:
> And, yes, we don't need to worry about breaking old clusters.

(I was asking about people using the old installer to install new clusters, but we've confirmed that we don't need to worry about that.)

pkg/network/openshift_sdn_test.go
```diff
@@ -101,7 +101,6 @@ func TestRenderOpenShiftSDN(t *testing.T) {
 		g.Expect(err).NotTo(HaveOccurred())
 
 		tweakMetaForCompare(cur)
-		g.Expect(cur).To(Equal(upd))
```
danwinship: This looks wrong? You change `cur` but then don't look at it?

squeed: Ah yeah, good catch.

PR #104, opened 2019-02-21
Revert "Config: rename some things to match Network.config.openshift.io"

squeed:
This reverts commit 100b1b75c3e66fb0008170861cd71f0279901620.

There is probably nothing wrong with this code, but it snuck through without CI because of a weird bug.


squeed:
cc @smarterclayton

PR #107, opened 2019-02-26
HACKING: add steps for how to build the node image.

squeed:


squeed:
cc @JacobTanenbaum

pecameron:
@squeed origin/images has both node and sdn. Which are we using? (Also, ovn-kubernetes is missing)

squeed:
@pecameron the "sdn" image in origin/images becomes $REGISTRY/openshift-node. Confusing, yes. Since ovn should have a much easier custom image building process, hopefully we never need to add that here!

squeed:
cc @LalatenduMohanty

INSTALLER-HACKING.md
```diff
@@ -67,3 +67,42 @@ To reduce memory usage, you can remove some non-essential components from the cl
...
+```
+
+2. Build the sdn process
+```
+make WHAT=./cmd/openshift-sdn
+```
+
+3. Build and push the node image
+```
+podman build -t ${REGISTRY}/origin-node:latest -f Dockerfile.node-hacking .
```
danwinship: maybe tag with a "weird" version number here (and request that version below) to ensure that if you did something wrong, it will just fail rather than giving you an upstream image?

INSTALLER-HACKING.md
```diff
@@ -67,3 +67,42 @@ To reduce memory usage, you can remove some non-essential components from the cl
...
+cat <<EOF > Dockerfile.node-hacking
+FROM docker.io/openshift/origin-node:v4.0.0
+COPY _output/local/bin/linux/amd64/openshift-sdn /usr/bin/openshift-sdn
+EOF
+```
+
+#### Building a development sdn image
+1.  Pick a registry. You can use the Openshift CI one if you are a member of the organization.
+```
+export REGISTRY=registry.svc.ci.openshift.org/<YOUR-GH-UID>
```
danwinship: expand `GH` there. it took me a second...

INSTALLER-HACKING.md
squeed: Only advantage of latest is that it has an implicit ImagePullPolicy of Always. Otherwise you have to bump the tag, change the daemonset's image pull policy, or manually do a `crictl image rm`

You are right to be worried about somehow accidentally pushing to the wrong place.

pecameron:
@squeed In what way is ovn easier than sdn? I have been using sdn as the model.

danwinship:
> In what way is ovn easier than sdn?

Because we know how to build OVN images that are equivalent to the "real" ones. The reason this hack is needed is because apparently literally no one knows how to build one-off SDN images correctly.

pecameron:
@danwinship The issue with building the images is ocp builds using the public (customer visible) released fast data path ovs rpms (currently 2.9.0). We (I) don't know how OKD images are built, we don't now and have never built an OKD image for OVN. We cannot add a non-released (e.g., 2.11) ovs to the public facing fast data path since that effectively releases it. There is an internal facing fast data path repo from openvswitch. ART has now created a new internal facing fast data path repo that can have non-released versions and have populated it with 2.11 rpms. The problem that we are working on has to do with building a image from the internal fast datapath repo with the non-released ovs image. This process is not fully worked out.

Since ovn also uses ovs/ovn from fast datapath, ovn builds will have the same issue. And hopefully same solution.

For the record, we have built an ovn v4.0 image from ose-ovn-kubernetes (I don't think we are shipping it in 4.0 since ovn is not in 4.0). I don't know if anyone has used this image. I am in the process if getting openshift/ovn-kubernetes up and building OKD images. This is a work in progress. As part of this I will get it to build with non-releases ovs/ovn. All of the existing test images are built using centos:7 and a private ovs repo with ovs 2.9.2 (which is not available in fast datapath).

danwinship:
@pecameron this PR isn't about the OVS problem, it's about how to test changes to origin

INSTALLER-HACKING.md
```diff
@@ -67,3 +67,42 @@ To reduce memory usage, you can remove some non-essential components from the cl
...
+2. Create the following dockerfile in your origin repo:
+```
+cat <<EOF > Dockerfile.node-hacking
+FROM docker.io/openshift/origin-node:v4.0.0
+COPY _output/local/bin/linux/amd64/openshift-sdn /usr/bin/openshift-sdn
+EOF
+```
+
+#### Building a development sdn image
+1.  Pick a registry. You can use the Openshift CI one if you are a member of the organization.
```
LalatenduMohanty: > if you are a member of the organization.

Do you mean member of OpenShift organization in github?

PR #108, opened 2019-02-26
openshift-sdn: more service-catalog netnamespace fixes

squeed:
This adds 3 more namespaces to the control-plane namespace:
- openshift-template-service-broker
- openshift-ansible-service-broker
- kube-service-catalog-controller-manager

Fixes BZ1680214

/cc @jboyd01


jboyd01:
Thanks Casey
/lgtm

PR #109, opened 2019-02-27
Add priorityClass to sdn-controller

ravisantoshgudimetla:
Add `system-node-critical` priorityClass to ensure that these DS pods won't be evicted by default scheduler.

### DIFF
--- a/bindata/network/openshift-sdn/controller.yaml
+++ b/bindata/network/openshift-sdn/controller.yaml
@@ -28,6 +28,7 @@ spec:
       labels:
         app: sdn-controller
     spec:
+      priorityClassName: system-node-critical
       containers:
       - name: sdn-controller
         image: {{.HypershiftImage}}


danwinship:
What is the criterion for being considered "critical"? Note that sdn-controller doesn't manage the actual network connectivity, but merely does higher-level bookkeeping (eg, assigning IP subnets to newly-created nodes) for the "sdn" and "sdn-ovs" pods (which are already marked critical). And it already runs in a HA configuration, so if the sdn-controller on one master is evicted then one of the others will just take over and most likely the users won't even notice.

So, if we care about limiting the number of pods that we mark critical, then we can probably get away with *not* marking sdn-controller critical. But if we'd rather just go with "better safe than sorry", then ok, sure, let's mark it critical.

ravisantoshgudimetla:
> Note that sdn-controller doesn't manage the actual network connectivity, but merely does higher-level bookkeeping (eg, assigning IP subnets to newly-created nodes) for the "sdn" and "sdn-ovs" pods (which are already marked critical). And it already runs in a HA configuration, so if the sdn-controller on one master is evicted then one of the others will just take over and most likely the users won't even notice.

@danwinship - Even in HA configuration, we are not evicting certain control plane components even though the operator is capable of spinning up new pods and there are other pods that can serve when a pod is evicted. But, if you think this is not needed for sdn-controller, I will close the PR. I created this PR as part of  https://bugzilla.redhat.com/show_bug.cgi?id=1683100, where QE reported we are not having priority class for sdn-controller. Please feel free to re-open the PR, in case you think, this is worth while.

danwinship:
> But, if you think this is not needed for sdn-controller

Well, I don't know. That's why I was asking. I don't know what our criteria are for marking pods critical, and I was just pointing out that *maybe* sdn-controller shouldn't count (unlike sdn and sdn-ovs, which clearly should).

ravisantoshgudimetla:
>  Well, I don't know. That's why I was asking. I don't know what our criteria are for marking pods critical, and I was just pointing out that _maybe_ sdn-controller shouldn't count (unlike sdn and sdn-ovs, which clearly should).

The criticality is dependent on availability of that service in cluster. For example, if all the 3 pods are evicted from the 3 nodes at the same time(this is extremely hypothetical but a possibility), are we fine with it? If the answer to the above question is no, then we should add priorityClass, if not, we can keep this PR closed. Also, it's mostly the operator developer's call.

PR #110, opened 2019-02-28
Report version number from payload "when we hit level"

smarterclayton:
As part of closing out cluster operators they must report when they
reach "level" - have rolled out the code in the update. For the network
operator that involves ensuring all the daemonsets and deployments
have the latest code deployed, and then reporting the current
controller release version via cluster operator status.

Related to https://github.com/openshift/cluster-image-registry-operator/pull/219

Needs a bump to vendor and updates to the cluster operator conditions
coming from @danwinship

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"os"
 	"reflect"
 	"strings"
 
@@ -47,7 +48,7 @@ func New(client client.Client, name, version string) *StatusManager {
 }
 
 // Set updates the ClusterOperator.Status with the provided conditions
-func (status *StatusManager) Set(conditions ...configv1.ClusterOperatorStatusCondition) {
+func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...configv1.ClusterOperatorStatusCondition) {
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
 	err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
 	isNotFound := errors.IsNotFound(err)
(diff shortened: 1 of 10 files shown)


smarterclayton:
@derekwaynecarr

squeed:
I know that @danwinship is also working on implementing the latest Operator.Status type; can this be rolled in to that PR?

danwinship:
rebased and pushed. @squeed can you review?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -57,6 +58,16 @@ func (status *StatusManager) Set(conditions ...configv1.ClusterOperatorStatusCon
 	}
 
 	oldStatus := co.Status.DeepCopy()
+
+	if reachedAvailableLevel {
+		if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
+			co.Status.Versions = []configv1.OperandVersion{
```
squeed: @smarterclayton is this right? We don't update the operator's version until all the other operands are updated? That's not what I'd expected.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -47,7 +48,7 @@ func New(client client.Client, name, version string) *StatusManager {
 }
 
 // Set updates the ClusterOperator.Status with the provided conditions
-func (status *StatusManager) Set(conditions ...configv1.ClusterOperatorStatusCondition) {
+func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...configv1.ClusterOperatorStatusCondition) {
```
squeed: I'm surprised that we can't compute `reachedAvailableLevel` directly from the conditions - is that really true?

pkg/controller/statusmanager/status_manager.go
squeed: Ah, nevermind - this is correct. Not what I expected, but correct.

squeed:
I wish we could determine reaching level directly from conditions, but I'm not about to block this PR. So if everyone else thinks it's OK, then I'm OK with it.

danwinship:
@squeed I had started to rework the figuring-out-if-we're-available code (as part of also improving the Available/Progressing Messages to better match the doc recommendations), but I figured I could do that separately later

/lgtm

PR #112, opened 2019-03-02
manifests: ensure system priorities

derekwaynecarr:
blocks passing smoke tests for ensuring control plane pods always schedule.

see: openshift/origin#22217


danwinship:
ok, so this answers the question from #109 about "how critical does it have to be to be critical".
/lgtm

PR #115, opened 2019-03-05
Bug 1680562: ClusterOperator.Status fixes

danwinship:
This is mostly a rewrite/reorg in preparation for getting versioning working, but it makes sense independently as well, and fixes the "It may take about 5 minutes to update the operator status back to normal after fix the problem in network config" problem.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1680562


squeed:
This looks awesome. I've got nothing.
/lgtm.

PR #116, opened 2019-03-06
openshift-sdn: mount full sysconfig directory

squeed:
When we try and mount `/etc/sysconfig/openshift-sdn`, but it doesn't exist, runc creates it as a directory. This is annoying when we want users to be able to enable logging - it needs to be a file.

So, just mount /etc/sysconfig.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -70,6 +70,11 @@ spec:
             source /etc/sysconfig/openshift-sdn
             set +o allexport
           fi
+          #BUG: cdc accidentally mounted /etc/sysconfig/openshift-sdn as DirectoryOrCreate; clean it up so we can ultimately mount /etc/sysconfig/openshift-sdn as FileOrCreate
+          # Once this is released, then we can mount it properly
+          if [[ -d /etc/sysconfig/openshift-sdn ]]; then
+            rmdir /etc/sysconfig/openshift-sdn || true
+          fi
 
           # Take over network functions on the node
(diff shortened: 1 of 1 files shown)


danwinship:
ok, but... wouldn't it be awesomer to just use a configmap?

dcbw:
Or maybe FileOrCreate? Though permissions might be an issue there.

squeed:
Wow, somehow missed that gh email.
@danwinship, this is the current solution for "I want verbose logs only on this node." Maybe we can just enable debug globally for the SDN, but traditionally we've tried to avoid that.

@dcbw, you're totally right, FileOrCreate is the right answer here. I hadn't seen that :-)

danwinship:
ah, yes, per node makes it trickier. though it still seems like configmap rather than file-on-local-disk is the right answer

squeed:
Updated to use FileOrCreate.

This PR will break upgrades, badly, since the path will already exist as a directory. I'll ask if that's a problem.

squeed:
Nevermind, changing the mount point now will result in an unrecoverable death during upgrade. So back to the directory.

@dcbw PTAL

danwinship:
maybe add an `rmdir /etc/sysconfig/openshift-sdn >& /dev/null || true` to the startup script to prepare us for fixing this in the future?

squeed:
Very good idea. Updated.

squeed:
Test flake seems odd. The SDN came up, so this PR didn't have any effect.

/retest

squeed:
Looking at the test failure, etcd never seemed to come up (boo).
/retest

squeed:
all green!

PR #117, opened 2019-03-06
Remove crd manifest which is now in cluster-config-operator

sttts:
Depends on https://github.com/openshift/cluster-config-operator/pull/23.

### DIFF
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -94,22 +94,3 @@ spec:
                       type: string
         status:
           type: object
-
----
-
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  name: networks.config.openshift.io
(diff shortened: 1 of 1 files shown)


manifests/0000_07_cluster-network-operator_01_crd.yaml
```diff
@@ -1,115 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  name: networkconfigs.networkoperator.openshift.io
```
danwinship: This file installs both `networks.config.openshift.io` and `networkconfigs.networkoperator.openshift.io` (the operator-specific config). cluster-config-operator only installs the former, so you can't get rid of this file completely.

sttts: fixed

danwinship:
sorry, forgot to re-review this, and now it seems soltysh submitted a duplicate PR (#124) which is already lgtm'ed

PR #119, opened 2019-03-07
Adds parameter to set multus log level to verbose in multus.yaml

dougbtv:
This requires that this pull request is merged first: https://github.com/openshift/multus-cni/pull/7/files

If this gets merged before that pull request -- Multus may fail to come up due to the fact that previous versions did not support unknown parameters (that will be mitigated in the future by changes within the referenced PR)


danwinship:
/hold
(that will keep the bot from merging the PR; use "/hold cancel" after the upstream PR merges)

PR #120, opened 2019-03-12
Re-group the operator configuration to Network.operator.openshift.io

squeed:
A gigantic PR that does nothing interesting :-)


danwinship:
Does anything else need to be updated for this? Or does no one else use the operator config at this point?

README.md
```diff
@@ -47,8 +47,8 @@ spec:
 *Corresponding Operator Config*
 This configuration is the auto-generated translation of the above Cluster configuration.
 ```yaml
-apiVersion: networkoperator.openshift.io/v1
-kind: NetworkConfig
+apiVersion: operator.openshift.io/v1
```
pecameron: Could you provide the details for the example you mention above of changing the VXLANPORT. A specific question, do the repeated fields between the cluster config and operator config have to be stated and the same? How is this input to the installer?

README.md
```diff
@@ -144,20 +144,13 @@ spec:
...
-
-and any changes will be automatically applied.
+The operator is expected to run as a pod (via a Deployment) inside a kubernetes cluster. It will retrieve the configuration above and reconcile the desired configuration. A suitable manifest for running the operator is located in `manifests/`.
 
 ## Unsafe changes
 Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested. 
 
 ### Safe changes to apply:
-It is safe to edit the following fields in `NetworkConfig.Spec`:
+It is safe to edit the following fields in the Operator configuration:
```
pecameron: In the safe and unsafe configurations, I think you make changes by editing the cluster and operator config objects and save them. Is that so?  If yes, how do you back out errors in an incorrect config? How does the problem show in the status?

danwinship:
> Outside of a request to expand the discussion in the README, this
> /lgtm

> openshift-merge-robot merged commit 643a622 into openshift:master 

@pecameron don't say "/lgtm" if you don't want it to merge yet...

squeed:
Well, that was easy.

@danwinship, to answer your question, yes people will be using the operator config (e.g. to enable Multitenant). That's one of the reasons it was moved to openshift/api.

pecameron:
@danwinship forgot the merge was automatic. The README changes can be done, or not, at another time.

PR #121, opened 2019-03-13
Bug 1687973: Fix operator status during upgrade

danwinship:
Don't explicitly mark the operator as Available=False when marking it Progressing=True... However, if we mark it Progressing=True and there is no Available status at all, then mark it Available=False in that case.

The net effect of this is that we should still be Available=False at startup, but we will be Available=True Progressing=True during upgrades.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -71,6 +71,20 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...confi
 	for _, condition := range conditions {
 		v1helpers.SetStatusCondition(&co.Status.Conditions, condition)
 	}
+
+	progressingCondition := v1helpers.FindStatusCondition(co.Status.Conditions, configv1.OperatorProgressing)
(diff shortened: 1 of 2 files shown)


squeed:
Seems reasonable.

One question to think about for the future: It doesn't really seem like the network can ever go Available = false, right?

danwinship:
> One question to think about for the future: It doesn't really seem like the network can ever go Available = false, right?

With this PR or in general? There is no longer any Available: True -> False transition in the code after this PR, but it could happen in theory in the future given a slightly improved definition of Available.

Although honestly, anything bad enough for us to say that the network was not available would probably involve the CNO not being able to talk to the apiserver, so I guess we don't actually have to write the code to set Available False and then back to True again later. :smile:

squeed:
That's kind of what I was thinking. Though, theoretically, the cno is host-nework, along with the apiserver, so it could report non-available status. Ideally, if all the sdn pods were in CrashLoopBackoff, we would report as not available, but I don't think it would be a useful signal.

PR #123, opened 2019-03-14
network: don't allow changing DisableMultiNetwork

squeed:
We don't currently support it, and it's not tested.

"Fixes" 1688212.


danwinship:
yup
/lgtm

PR #124, opened 2019-03-14
Remove networks.config.openshift.io which is already instantiated by cluster-config-operator

soltysh:
Namely this is the place https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_05_config-operator_01_network.crd.yaml and that is guaranteed to be instantiated before you run your operator.

### DIFF
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -95,21 +95,3 @@ spec:
         status:
           type: object
 
----
-
-apiVersion: apiextensions.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


deads2k:
> Namely this is the place https://github.com/openshift/cluster-config-operator/blob/master/manifests/0000_05_config-operator_01_network.crd.yaml and that is guaranteed to be instantiated before you run your operator.

and that cluster-config-operator  image provides a render command used by the installer to produce manifests for config.openshift.io CRDs because operators rely upon those resources for determining how to behavior across components.

deads2k:
sorry, wrong button.

squeed:
aws igw creation timeout
/retest

PR #125, opened 2019-03-15
Add macVlanConfig into operator config

s1061123:
This change introduce macVlanConfig object into operator config.

### DIFF
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -67,7 +67,7 @@ spec:
               type: array
               items:
                 type: object
-                required: ["type", "name", "rawCNIConfig"]
+                required: ["type", "name"]
                 properties:
                   type:
                     type: string
@@ -75,6 +75,19 @@ spec:
                     type: string
                   rawCNIConfig:
                     type: string
+                  macVlanConfig:
+                    type: object
+                    required: ["master", "ipam"]
+                    properties:
+                      master:
+                        type: string
+                      ipam:
+                        type: string
+                      mode:
+                        type: string
+                      mtu:
+                        type: integer
+                        minimum: 0
             disableMultiNetwork:
               type: boolean
             deployKubeProxy:
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -77,7 +77,7 @@ spec:
                     type: string
                   macVlanConfig:
                     type: object
-                    required: ["master", "ipam"]
+                    required: ["master" ]
                     properties:
                       master:
                         type: string
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -79,7 +79,6 @@ spec:
                     type: string
                   macVlanConfig:
                     type: object
-                    required: ["master"]
                     properties:
                       master:
                         type: string
@@ -87,9 +86,6 @@ spec:
                         type: string
                       mode:
                         type: string
-                      mtu:
-                        type: integer
-                        minimum: 0
             disableMultiNetwork:
               type: boolean
             deployKubeProxy:
--- a/manifests/0000_07_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_07_cluster-network-operator_01_crd.yaml
@@ -68,6 +68,8 @@ spec:
               items:
                 type: object
                 required: ["type", "name"]
+                minProperties: 3
+                maxProperties: 3
                 properties:
                   type:
                     type: string
@@ -77,7 +79,7 @@ spec:
                     type: string
                   macVlanConfig:
                     type: object
-                    required: ["master" ]
+                    required: ["master"]
                     properties:
                       master:
                         type: string


s1061123:
@dcbw Could you please take a look into it?

squeed:
/hold
Some background: I suggested to Tomo that he file a PR with just the proposed API changes early.

manifests/0000_07_cluster-network-operator_01_crd.yaml
```diff
@@ -67,14 +67,27 @@ spec:
...
                 properties:
                   type:
                     type: string
                   name:
                     type: string
                   rawCNIConfig:
                     type: string
+                  macVlanConfig:
+                    type: object
+                    required: ["master", "ipam"]
```
squeed: I would say that ipam should not be required - if omitted, it defaults to "dhcp"

manifests/0000_07_cluster-network-operator_01_crd.yaml
```diff
@@ -67,14 +67,27 @@ spec:
...
+                    properties:
+                      master:
+                        type: string
```
squeed: Is there any way we can auto-detect this? 

Given that we'll have to write a small daemonset that installs the macvlan CNI configuration, it can include a small binary that detects the uplink interface.

manifests/0000_07_cluster-network-operator_01_crd.yaml
s1061123: Sure. Changed.

manifests/0000_07_cluster-network-operator_01_crd.yaml
s1061123: In case of baremetal deployment, it is very hard I guess because user may wan to have several isolated networks over some hardware interfaces.

In addition, user may put hardware NICs asymmetry, such as:

- workerA: eth0(openshift-sdn), eth1(isolate netA)
- workerB: eth0(openshift-sdn),
- workerC: eth0(openshift-sdn), eth1(isolate netB), eth2(isolate netA)

And put the pod with pod affinity with their policy.

This case, we also need to discover isolated network as well as NIC somehow, so we could put it in future feature.

s1061123:
@dcbw please take a look into it?

s1061123:
@danwinship PTAL?

s1061123:
@dcbw PTAL?

manifests/0000_07_cluster-network-operator_01_crd.yaml
danwinship: While we definitely want people to be able to specify `master`, it seems like autodetecting and using the interface with the default route would be the right answer for a lot of users. It would be good if `master` wasn't required

manifests/0000_07_cluster-network-operator_01_crd.yaml
```diff
@@ -67,14 +67,27 @@ spec:
               type: array
               items:
                 type: object
-                required: ["type", "name", "rawCNIConfig"]
+                required: ["type", "name"]
```
danwinship: Make this:

    required: ["type", "name"]
    minProperties: 3
    maxProperties: 3

which effectively says "you must specify `type`, `name`, and exactly one other property" (ie, either `rawCNIConfig` or `macVlanConfig` but not both)

manifests/0000_07_cluster-network-operator_01_crd.yaml
```diff
@@ -67,14 +67,27 @@ spec:
...
+                    type: object
+                    required: ["master" ]
+                    properties:
+                      master:
+                        type: string
+                      ipam:
+                        type: string
+                      mode:
+                        type: string
+                      mtu:
```
danwinship: when would this not just be the mtu of the master?

s1061123: To optimize TCP traffic (to avoid segmentation) for specific target, but it is of course not usual stuff.

manifests/0000_07_cluster-network-operator_01_crd.yaml
s1061123: fixed.

manifests/0000_07_cluster-network-operator_01_crd.yaml
s1061123: If it add macvlan to default route interface, this is the same network as openshift-sdn so it is not reason why they add additional interface. User want to add additional interface to connect other networks other than openshift-sdn, I believe. So I suppose we should have master parameter as required variable.

s1061123: In addition, macvlan to default route interface, the macvlan interface connects to host network (aka underlay network) and the interface will skip OvS flow rules. This seems to make the openshift network vulnerable. I am not clear on the use-case of such situation. What do you think about it?

danwinship: The use case is wanting a public IP distinct from the node's public IP. For instance, the OpenShift egress router uses a macvlan, but it is currently implemented via a special pod annotation that makes the openshift-sdn network plugin create the macvlan in addition to attaching the pod to the OVS bridge. In the future we want to implement it using multus.

dcbw: Also, putting a macvlan onto the same interface as openshift-sdn *does* mean increased performance because you're not running things through the VXLAN tunnel. So you can theoretically get closer to wirespeed on the macvlans. So it's still pretty useful for higher bandwidth pods even on the default interface.

manifests/0000_07_cluster-network-operator_01_crd.yaml
dcbw: @s1061123 do we have a concrete use-case for having the MTU at this point? eg do we know of specific situations or scenarios that the MTU of the master wouldn't work?

s1061123: removed mtu.

s1061123: I don't have clear use cases with detailed situation, so removed.

manifests/0000_07_cluster-network-operator_01_crd.yaml
s1061123: Okey, so 'master' goes to optional parameter.

s1061123:
@danwinship @dcbw PTAL?

s1061123:
Let me close it because it is too old.

PR #128, opened 2019-03-21
specify the same location for the openshift-sdn binary

JacobTanenbaum:
RHEL7 unifies /usr/bin and /bin but we should refer to the location of the openshift-sdn binary as the same place even though the two are functionally equivalent


dcbw:
/retest
AWS route creation timeouts.

PR #130, opened 2019-03-26
DHCP Daemon

dougbtv:
Adds rendering, bindata & unit tests for the containernetworking/plugins DHCP CNI plugin running as a daemon.

Currently has no dependency on the macvlan functionality, however, it anticipates it with a commented-out `TODO` block.

### DIFF
--- a/bindata/network/multus/003-dhcp-daemon.yaml
+++ b/bindata/network/multus/003-dhcp-daemon.yaml
@@ -1,0 +1,62 @@
+{{if .RenderDHCP}}
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: dhcp-daemon
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the containernetworking plugins DHCP daemon on each node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: dhcp-daemon
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: dhcp-daemon
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        beta.kubernetes.io/os: linux
+      tolerations:
+      - operator: Exists
+      initContainers:
+      - name: dhcp-daemon-initialization
+        image: {{.CNIPluginsSupportedImage}}
+        command: ["/bin/sh"]
+        args: ["-c", "rm -f /var/run/cni/dhcp.sock"]
+        volumeMounts:
+        - name: socketpath
+          mountPath: /var/run/cni
+      containers:
+      - name: dhcp-daemon
+        # Based on: https://github.com/s1061123/cni-dhcp-daemon/blob/master/Dockerfile
+        image: {{.CNIPluginsSupportedImage}}
+        imagePullPolicy: Always
+        command: ["/usr/src/plugins/bin/dhcp"]
+        args:
+        - "daemon"
+        - "-hostprefix"
+        - "/host"
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - name: socketpath
+          mountPath: /var/run/cni
+        - name: procpath
+          mountPath: /host/proc
+      volumes:
+        - name: socketpath
+          hostPath:
+            path: /var/run/cni
+        - name: procpath
+          hostPath:
+            path: /proc
+{{- end}}
--- a/pkg/network/dhcp_daemon.go
+++ b/pkg/network/dhcp_daemon.go
@@ -1,0 +1,63 @@
+package network
+
+import (
+	"encoding/json"
+	operv1 "github.com/openshift/api/operator/v1"
+	"log"
+)
+
+// UseDHCP determines if the the DHCP CNI plugin running as a daemon should be rendered.
+func UseDHCP(conf *operv1.NetworkSpec) bool {
+	renderdhcp := false
+
+	// This isn't useful without Multinetwork.
+	if *conf.DisableMultiNetwork {
+		return renderdhcp
+	}
+
+	// Look and see if we have an AdditionalNetworks
+	if conf.AdditionalNetworks != nil {
+		for _, addnet := range conf.AdditionalNetworks {
+			// Parse the RawCNIConfig
+			var rawConfig map[string]interface{}
+			var err error
+
+			confBytes := []byte(addnet.RawCNIConfig)
+			err = json.Unmarshal(confBytes, &rawConfig)
+			if err != nil {
+				log.Printf("WARNING: Not rendering DHCP daemonset, failed to Unmarshal RawCNIConfig: %v", confBytes)
+				return renderdhcp
+			}
+
+			// Cycle through the IPAM keys, and determine if the type is dhcp
+			if rawConfig["ipam"] != nil {
+				ipam, okipamcast := rawConfig["ipam"].(map[string]interface{})
+				if !okipamcast {
+					log.Printf("WARNING: IPAM element has data of type %T but wanted map[string]interface{}", rawConfig["ipam"])
+					continue
+				}
+
+				for key, value := range ipam {
+					if key == "type" {
+						typeval, oktypecast := value.(string)
+						if !oktypecast {
+							log.Printf("WARNING: IPAM type element has data of type %T but wanted string", value)
+							break
+						}
+
+						if typeval == "dhcp" {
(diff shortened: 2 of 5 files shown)


bindata/network/dhcp/dhcp-daemon.yaml
```diff
@@ -0,0 +1,63 @@
...
+        # Based on: https://github.com/s1061123/cni-dhcp-daemon/blob/master/Dockerfile
+        image: {{.CNIPluginsSupportedImage}}
+        imagePullPolicy: Always
+        command: ["/usr/src/plugins/bin/dhcp"]
+        args:
+        - "daemon"
+        - "-hostprefix"
+        - "/host"
+        securityContext:
+          privileged: true
```
danwinship: why does this need to be privileged?

bindata/network/dhcp/dhcp-daemon.yaml
```diff
@@ -0,0 +1,63 @@
...
+        beta.kubernetes.io/os: linux
+      tolerations:
+      - operator: Exists
+      initContainers:
+      - name: cni-plugins-supported
+        image: {{.CNIPluginsSupportedImage}}
+        command: ["/bin/sh"]
+        args: ["-c", "rm -f /var/run/cni/dhcp.sock"]
+        securityContext:
+          privileged: true
```
danwinship: even assuming the `dhcp-daemon` container needs to be privileged, why does the `initContainer` need to be privileged? (also, needs a better `name`)

pkg/network/dhcp_daemon_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Type: operv1.NetworkTypeOpenShiftSDN,
+			OpenShiftSDNConfig: &operv1.OpenShiftSDNConfig{
+				Mode: operv1.SDNModeNetworkPolicy,
+			},
+		},
+	},
+}
+
+// TestRenderNoDHCP tests a rendering with the DHCP daemonset.
+func TestRenderWithDHCP(t *testing.T) {
```
danwinship: comment should say `TestRenderWithDHCP`

bindata/network/dhcp/dhcp-daemon.yaml
```diff
@@ -0,0 +1,63 @@
+---
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: dhcp-daemon
```
danwinship: could this just be added to the existing multus DaemonSet as an optional additional container? (using `{{if ...}}` to optionally include/exclude part of the template based on a variable set by the renderer, as in `bindata/network/openshift-sdn/004-multitenant.yaml`)

pkg/network/render.go
```diff
@@ -318,3 +326,76 @@ func RenderMultus(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstruct
 	out = append(out, objs...)
 	return out, nil
 }
+
+// RenderDHCP generates the manifests for the reference DHCP CNI plugin running as a daemon
+func RenderDHCP(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstructured, error) {
+
+	// This isn't useful with Multinetwork.
+	if *conf.DisableMultiNetwork {
```
danwinship: isn't useful *without* MultiNetwork

pkg/network/render.go
```diff
@@ -318,3 +326,76 @@ func RenderMultus(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstruct
...
+		return nil, nil
+	}
+
+	renderdhcp := false
+
+	// Look and see if we have an AdditionalNetworks
+	if conf.AdditionalNetworks != nil {
+
+		for _, addnet := range conf.AdditionalNetworks {
+
```
danwinship: (all these blank lines at the starts of blocks are non-idiomatic...)

pkg/network/render.go
```diff
@@ -318,3 +326,76 @@ func RenderMultus(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstruct
...
+			confBytes := []byte(addnet.RawCNIConfig)
+			err = json.Unmarshal(confBytes, &rawConfig)
+			if err != nil {
+				log.Printf("WARNING: Not rendering DHCP daemonset, failed to Unmarshal RawCNIConfig: %v", confBytes)
+				return nil, nil
+			}
+
+			// Cycle through the IPAM keys, and determine if the type is dhcp
+			if rawConfig["ipam"] != nil {
+				ipam := rawConfig["ipam"].(map[string]interface{})
```
danwinship: This will panic if the CNI is invalid (`"ipam": "yes"`). Need to do a checked typecast.

Likewise below. Alternatively, demarshal it into a real type and make `json.Unmarshal` do the checking for you:

    type CNINetConfIPAMOnly struct {
    	IPAM struct {
    		Type string `json:"type"`
    	} `json:"ipam"`
    }

dougbtv:
@danwinship -- thanks a bunch for the thorough review, appreciate it.

I pushed a new commit, I believe I covered what you've suggested, including:

* Removed privileged security context (it's not necessary)
* Updated name of init container
* Updated comments per suggestions.
* Moved DHCP yaml to Multus dir and have it handled by Multus renderer (nice suggestion)
* Removed blank lines at beginning of functions.
* Checked for errors during type conversion, added test with bogus IPAM definition.

Thanks again.

bindata/network/multus/003-dhcp-daemon.yaml
```diff
@@ -0,0 +1,60 @@
+{{if .RenderDHCP}}
```
danwinship: So I actually meant to put this inside `multus.yaml`; rather than deploying an additional DaemonSet, with an additional Pod on each node, just stick the containers here into the existing multus daemonset, conditionally on `RenderDHCP` being set.

(This could be done as an optimization after the initial PR merges maybe though.)

bindata/network/multus/003-dhcp-daemon.yaml
```diff
@@ -0,0 +1,60 @@
...
+          mountPath: /var/run/cni
+        - name: procpath
+          mountPath: /host/proc
+      volumes:
+        - name: socketpath
+          hostPath:
+            path: /var/run/cni
+        - name: procpath
+          hostPath:
+            path: /proc
```
danwinship: hm... so I was wondering why we need this, and the answer is, so that the daemon can nsenter into pod netnamespaces and fiddle with their network interfaces, except that in that case, probably it did actually need to be privileged?

pkg/network/dhcp_daemon.go
```diff
@@ -0,0 +1,57 @@
...
+			// Cycle through the IPAM keys, and determine if the type is dhcp
+			if rawConfig["ipam"] != nil {
+				ipam, ok := rawConfig["ipam"].(map[string]interface{})
+				if !ok {
+					return renderdhcp, errors.Errorf("IPAM element has data of type %T but wanted map[string]interface{}", rawConfig["ipam"])
+				}
+
+				for key, value := range ipam {
+					if key == "type" {
+						if value.(string) == "dhcp" {
```
danwinship: The `for key, value` doesn't help here; you're still doing an unsafe cast. (eg,  it would crash on `{"ipam": { "type": 5 }}`

bindata/network/multus/003-dhcp-daemon.yaml
dougbtv: Ahhh yes, I had started to wonder why I had the privileged clause in my original sample daemonset from PoCs

bindata/network/multus/003-dhcp-daemon.yaml
dougbtv: Hey Dan -- what's the best way to test for the presence of a certain named container in the unit tests? I saw there's a custom matcher that does it for specific `kind:` but, was having trouble testing for the presence of a specific container within a daemonset.

pkg/network/dhcp_daemon.go
dougbtv: Great catch, missed that one, thanks.

dougbtv:
@danwinship -- got another suite of changes in.

* Fixed the missed check on the type casting. 
* Returned privileged to the dhcp-daemon container
* Also tweaked the UseDHCP() method so that it doesn't error but just warn (otherwise it'd prevent Multus from starting at all just due to some bad data, for example)
* Did not address moving from daemonset for DHCP to containers within Multus -- that should be fine, but, I was having trouble validating it in my tests.

Thanks again for the reviews.

danwinship:
> Did not address moving from daemonset for DHCP to containers within Multus -- that should be fine, but, I was having trouble validating it in my tests.

Yeah, probably needs a new helper function. We can do this later.

PR #136, opened 2019-04-03
Call syncFailing on SetNotFailing method

rcarrillocruz:
We should also need to call syncFailing in case level is nil.

Signed-off-by: Ricardo Carrillo Cruz <ricardo.carrillo.cruz@gmail.com>

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -144,8 +144,8 @@ func (status *StatusManager) SetFailing(level StatusLevel, reason, message strin
 func (status *StatusManager) SetNotFailing(level StatusLevel) {
 	if status.failing[level] != nil {
 		status.failing[level] = nil
-		status.syncFailing()
 	}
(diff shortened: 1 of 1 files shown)


danwinship:
> We should also need to call syncFailing in case level is nil.

Hm. Why? If `status.failing[level]` was already `nil` then the ClusterOperator should already have a status of Failing False...

> Signed-off-by: Ricardo Carrillo Cruz

(we don't bother with signed-off-by)

squeed:
This is as a result of https://bugzilla.redhat.com/show_bug.cgi?id=1690747

danwinship:
ah... ok, yes, if the ClusterOperator gets deleted then yes, I guess we have to explicitly re-set the status. And `Set()` will already turn that into a no-op if it hasn't changed. OK

/lgtm

danwinship:
@rcarrillocruz if a PR is to fix a bugzilla bug you should link to the bug from the PR, for context

PR #137, opened 2019-04-03
Remove /etc/cni/net.d/ file when plugin is down

pecameron:
The node is ready when the cni plugin creates its .conf file
in /etc/cni/net.d/ This change deletes thefile when the cni
plugin terminates.

bug: 1654044
https://bugzilla.redhat.com/show_bug.cgi?id=1654044

SDN 352 - 4.0 Keep node NotReady until sdn is up.
https://jira.coreos.com/browse/SDN-352

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -48,7 +48,7 @@ spec:
           set -euo pipefail
 
           # if another process is listening on the cni-server socket, wait until it exits
-          trap 'kill $(jobs -p); exit 0' TERM
+          trap 'kill $(jobs -p); rm -f /etc/cni/net.d/80-openshift-network.conf ; exit 0' TERM
           retries=0
           while true; do
             if echo 'test' | socat - UNIX-CONNECT:/var/run/openshift-sdn/cni-server.sock >/dev/null; then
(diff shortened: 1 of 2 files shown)


pecameron:
@dcbw Here is the cno part of the 1654044 fix for 4.0 PTAL

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -48,7 +48,7 @@ spec:
           set -euo pipefail
 
           # if another process is listening on the cni-server socket, wait until it exits
-          trap 'kill $(jobs -p); exit 0' TERM
+          trap 'kill $(jobs -p); rm -Rf /etc/cni/net.d/80-openshift-network.conf ; exit 0' TERM
```
danwinship: This seems redundant with the preStop hook... does it buy us anything?

(Also, in all three cases, the `R` flag to `rm` is unnecessary.)

pecameron:
@danwinship @dcbw removed the "R" PTAL

bindata/network/openshift-sdn/sdn.yaml
dcbw: @danwinship good question; though perhaps there are cases where the termination hook doesn't fire as we expect. Not sure.

pecameron: @dcbw There is a race form when the processes dies to when the file is deleted. It is good to limit the exposure.

PR #143, opened 2019-04-09
status: catch a few more Progressing cases

squeed:
The easiest way to tell that a daemonset is progressing is that UpdatedNumberScheduled is less than Desired.

Also, catch the case where the daemonset update hasn't yet been observed by the controller.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -193,8 +193,12 @@ func (status *StatusManager) SetFromPods() {
 
 		if ds.Status.NumberUnavailable > 0 {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
-		} else if ds.Status.NumberAvailable == 0 {
+		} else if ds.Status.NumberAvailable == 0 { // NOTE: update this if we ever expect empty (unscheduled) daemonsets ~cdc
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not yet scheduled on any nodes", dsName.String()))
+		} else if ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
+		} else if ds.Generation > ds.Status.ObservedGeneration {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is being processed (generation %d, observed generation %d)", dsName.String(), ds.Generation, ds.Status.ObservedGeneration))
 		}
 
 		if !(ds.Generation <= ds.Status.ObservedGeneration && ds.Status.UpdatedNumberScheduled == ds.Status.DesiredNumberScheduled && ds.Status.NumberUnavailable == 0 && ds.Annotations["release.openshift.io/version"] == targetLevel) {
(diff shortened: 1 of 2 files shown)


squeed:
/hold
Wanted to submit this for review now, but ran out of time to completely test it.

squeed:
context: Clayton noticed in https://bugzilla.redhat.com/show_bug.cgi?id=1695200 that LastTransitionTime for Progressing wasn't bumped after an update.

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -307,6 +308,16 @@ func TestStatusManagerSetFromDaemonSets(t *testing.T) {
 		t.Fatalf("unexpected Status.Conditions: %#v", co.Status.Conditions)
 	}
 
+	progressingTS := metav1.Now()
+	for _, cond := range co.Status.Conditions {
+		if cond.Type == configv1.OperatorProgressing {
```
danwinship: cond := v1helpers.FindStatusCondition(co.Status.Conditions, configv1.OperatorProgressing)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -195,6 +195,10 @@ func (status *StatusManager) SetFromPods() {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
 		} else if ds.Status.NumberAvailable == 0 {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not yet scheduled on any nodes", dsName.String()))
+		} else if ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
+		} else if ds.Generation > ds.Status.ObservedGeneration {
```
danwinship: The DaemonSet documentation is egregiously ambiguous, but it seems like it should not be possible to hit this case, given the previous case?

Also, we probably need a corresponding change to the Deployment-tracking code?

squeed: I played through an upgrade, and saw this case (generation > observedGeneration) after the operator (well, me at the CLI) had updated the daemonset and before the daemonset controller had reacted to it. 

And, yes, the deployment controller also needs some tweaks.

squeed:
Still not sure if / how to catch the (admittedly unlikely) case where we update a DS then go offline until the controller has completely rolled out the change. The only thing I can think of is persisting observedgeneration somewhere, which feels yucky.

squeed:
Updated, tested with a running cluster and verified that progressing had a useful status. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1695200

/hold cancel
@danwinship PTAL, thanks!

pkg/controller/statusmanager/status_manager.go
danwinship: > after the operator had updated the daemonset and before the daemonset controller had reacted to it.

right, but that's very transient; the daemonset controller should then react to it, at which point `ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled` will be true

danwinship:
> Still not sure if / how to catch the (admittedly unlikely) case where we update a DS then go offline until the controller has completely rolled out the change. The only thing I can think of is persisting observedgeneration somewhere, which feels yucky.

Hm... compare `ds.Status.Conditions[something].LastTransitionTime` with operator.Status.Conditions["Progressing"].LastTransitionTime?

squeed:
> Hm... compare `ds.Status.Conditions[something].LastTransitionTime` with operator.Status.Conditions["Progressing"].LastTransitionTime?

It does't look like the DaemonSet controller is actually setting any `Status.Condition[]` entries yet.

PR #146, opened 2019-04-17
hack/run-locally: add mode to attach to existing cluster

squeed:
I find myself sometimes wanting to run the cno on an existing cluster. Add some scaffolding to make that easier.

### DIFF
--- a/INSTALLER-HACKING.md
+++ b/INSTALLER-HACKING.md
@@ -38,6 +38,17 @@ It will print status messages as it waits for the installer to make progress.
 openshift-install --dir=$CLUSTER_DIR create cluster
 ```
 
+## Attaching to an existing cluster
+
+If you have an already-up cluster you would like to take over, then you can instead run
+
+```
+ATTACH_RUNNING=1 hack/run-locally.sh
+```
+
+It will stop the deployed operator, set up the environment variables, and execute
+your local build of the operator.
+
 ## Tips, Tricks, & Limitations
 
 ### Cleaning up
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -31,11 +31,28 @@ function override_install_manifests() {
     fi
 }
 
+# Install our overrides in to an existing cluster
+function stop_deployed_operator() {
+    echo "Telling the CVO to ignore the network operator"
+    kubectl patch --type=json -p "$(cat hack/overrides-patch.yaml)" clusterversion version
+
+    echo "Scaling the deployed network operator to 0"
+    kubectl scale deployment -n openshift-network-operator network-operator --replicas 0
+}
+
 # Extract environment variables for the CNO DaemonSet
 function setup_operator_env() {
     if [[ manifests/0000_07_cluster-network-operator_03_daemonset.yaml -nt "${CLUSTER_DIR}/env.sh" ]]; then
         echo "Copying environment variables from manifest to ${CLUSTER_DIR}/env.sh"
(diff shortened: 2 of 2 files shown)


danwinship:
I had started work on this too: https://github.com/danwinship/cluster-network-operator/commit/85d6d165

So let's see. Differences...
- I deleted a bunch more from `README.md`, which is good because some of that stuff is out of date anyway (eg, the list of environment variables you need to set). I think going forward we shouldn't try to document how to run the operator manually in the README, and just tell people if they want to do something that `run-locally.sh` doesn't quite do, then they should read `run-locally.sh` to figure out how to do it.
- You used `ATTACH_RUNNING=1`, while I tried to autodetect whether we were creating or attaching based on whether `CLUSTER_DIR` or `KUBECONFIG` was set. I realized that this wasn't ideal because you might have `CLUSTER_DIR` set from a previous run when you actually wanted to attach to a running cluster instead (or vice versa). I was thinking that maybe that if both variables were set it should check whether `KUBECONFIG` was inside or outside of `CLUSTER_DIR`?
  - Additionally, you require `CLUSTER_DIR` to be set even in the attach case, to provide a place to put `env.sh`, whereas I just dumped it into the same dir as `KUBECONFIG`.
  - If we want to keep an explicit "attach to running cluster" indication, I feel like `run-locally.sh --attach` would be better than `ATTACH_RUNNING=1 run-locally.sh`
- You included code to stop the running CNO whereas I just left the documentation about how to stop it by hand. Your way is clearly more in the spirit of `run-locally.sh`.
- You set up the local CNO's environment based on the previously-running CNO's environment, whereas I reused the existing `setup_operator_env` to set it up based on the working dir. I think my version is correct here; the locally-built CNO may require environment variables that the deployed CNO didn't (eg if you're testing out support for a new multus plugin type, it might need to deploy some additional image).
- You always run `stop_deployed_operator` and print its messages, whereas post- #140 I was trying to make it so if you `^C` and run it again, it won't re-print any messages that aren't relevant this time. (Related: should probably mention in the docs that if you `^C` it, the old CNO won't be restarted.)

squeed:
(whoops, github webapp state; missed your comment)

I decided against auto-detecting whether or not to attach mostly because I want to make the most complicated case, i.e. running the operator with the installer, simpler.

As for extracting the environment variables from the running cluster, my main concern was getting the VERSION variable so that the CVO isn't angry. But now that I write this, I'm not sure that it really matters. I guess my other motivation is that, when I attach to an already-running cluster, I'm likely not testing any new container images and just want to change some configuration, so I'd like to avoid the churn.

danwinship:
> As for extracting the environment variables from the running cluster, my main concern was getting the VERSION variable so that the CVO isn't angry. But now that I write this, I'm not sure that it really matters.

Yeah, we're telling the CVO to ignore us anyway, right?

> I guess my other motivation is that, when I attach to an already-running cluster, I'm likely not testing any new container images and just want to change some configuration, so I'd like to avoid the churn.

What churn exactly? You mean the fact that if you run a 4.2 CNO against a 4.1 cluster with 4.2 environment variables, that it might try to install 4.2 openshift-sdn, etc? Well, yes, but if you ran the 4.2 CNO with the 4.1 environment variables, and it had to restart something, then it would be using the 4.2 DaemonSet yaml to run the 4.1 image, which presumably wouldn't work.

If you want to avoid churn, the way to do that is to run a local CNO which is very similar to the previously-running CNO.

There might, in the future, be environment variables that we need to make sure that the new CNO gets the same value as the previously-running CNO had. But the `IMAGE` variables are basically part of the CNO's internal API, and can't be mixed and matched.

danwinship:
@squeed needs-rebase. With #164 we should be getting the correct `RELEASE_VERSION` (though test that; I only tested the new cluster case, not the existing cluster case), and I believe the existing code for getting the image names is correct (and, in particular, if we had been getting them from the cluster's daemonset then I wouldn't have been able to test ovn-kubernetes correctly).

dcbw:
Obsoleted by https://github.com/openshift/cluster-network-operator/pull/186

PR #147, opened 2019-04-18
Change "Failing" to "Degraded"

danwinship:
The new docs imply that, eg, we should not consider "the configuration is invalid and unreconcilable" to be "Degraded", but Eric says to just treat this as a rename for now (https://github.com/openshift/cluster-version-operator/pull/161#issuecomment-484200502).


squeed:
"All conditions are degraded, but some are more degraded than others."

PR #151, opened 2019-04-23
openshift-sdn: configure sdn readiness probe

squeed:
By checking to see that the sdn configuration file gets written out, we ensure the daemonset doesn't roll out a completely broken change.

It also slows down sdn rollouts slightly.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -138,6 +138,12 @@ spec:
           preStop:
             exec:
               command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf", "/host/opt/cni/bin/openshift-sdn"]
+        readinessProbe:
+          exec:
+            # openshift-sdn writes this file when it is ready to handle pod requests.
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -138,6 +138,12 @@ spec:
           preStop:
             exec:
               command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf", "/host/opt/cni/bin/openshift-sdn"]
+        readinessProbe:
+          exec:
+            # openshift-sdn writes this file when it is ready to handle pod requests.
+            command: ["ls", "/etc/cni/net.d/80-openshift-network.conf"]
```
danwinship: does kubelet print out stderr of exec probes? Anyway, wouldn't `test -f` be more idiomatic anyway?

squeed: good point, `test -f` is better. No clue about exec probe logging.

squeed:
Test failures seem like a flake... the cluster came up.
/retest

PR #152, opened 2019-04-23
additional-networks: add Namespace support

dcbw:
Additional networks created at install time should be able to be placed
in specific namespaces as this is how authorization to attach these
networks to pods is performed.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1679946

Depends on https://github.com/openshift/api/pull/294

@squeed @danwinship @s1061123 @dougbtv


pkg/network/additional_networks.go
```diff
@@ -28,9 +30,15 @@ func renderRawCNIConfig(conf *operv1.AdditionalNetworkDefinition, manifestDir st
 	var err error
 	objs := []*uns.Unstructured{}
 
+	namespace := conf.Namespace
```
squeed: I added a "getOr" function to the templating language, You can do something like this
```
namespace: {{getOr . "AdditionalNetworkNamespace" "default" }}
```

Up to you which you'd rather do.

dcbw: Updated.

zshi-redhat:
Thanks for taking care of the sriov changes!
It might be worth a few sentences in the [openshift-docs](https://github.com/openshift/openshift-docs/blob/enterprise-4.1/modules/nw-multinetwork-creating-first-attachments.adoc) about the ``namespace`` config and how the CR is configured via operator.

squeed:
looks good to me - not giving it the real flag until I confirm that this is the right way to merge api changes.

dcbw:
> looks good to me - not giving it the real flag until I confirm that this is the right way to merge api changes.

@squeed can you explain a bit more? The hold on this PR was to wait until the openshift/api changes merge, at which point I'd re-vendor that into CNO and update this PR.

squeed:
Er, yeah, we're in agreement. Just mentioning that it's almost-lgtm.

squeed:
The API pr has merged - let's get this in.

dcbw:
@squeed updated

PR #153, opened 2019-04-23
Fix handling of Deployment Progressing status

danwinship:
Apparently the "Progressing" condition on a Deployment does not get cleared when the Deployment becomes "Available".

(Casey added this as part of #143 but it wasn't tested there since openshift-sdn doesn't use any Deployments, just DaemonSets.)


squeed:
Wow
/lgtm

PR #154, opened 2019-04-23
Add helper script to open OVN ports on AWS

danwinship:
While waiting to figure out https://github.com/openshift/installer/issues/1218, here's a helper to open the ports for you in a cluster. It can be run almost immediately after `openshift-install create cluster` starts (and if you run it too soon, it will error out and you can try again later).

I didn't bother making `run-locally.sh` run it automatically yet, although we could... but it's hopefully temporary?

Requires the `awscli` and `jq` packages (and requires that `awscli` has been set up with the correct defaults, which I believe is explained somewhere in the openshift docs...)

### DIFF
--- a/hack/open-ovn-ports.sh
+++ b/hack/open-ovn-ports.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+function get_group_id() {
+    name="$1"
+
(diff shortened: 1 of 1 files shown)


dcbw:
Is this just for local testing?

danwinship:
Yes, this is so you can bring up a cluster running OVN while we figure out the eventual solution. It's explicitly intended to go away later

PR #155, opened 2019-04-24
openshift-sdn: daemonset improvements (exec, proxy healthz, journal)

squeed:
Some daemonset improvements:
- Log sdn and ovs-vswitchd to the system journal, so logs persist long-term
- Exec ovs-vswitchd directly, so it's restarted quickly
- Add kube-proxy healthz as liveness probe, so any kube-proxy hangs will cause it to be restarted

The ovs-vswitchd command line is taken directly from what ovs-vsctl created.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -60,6 +60,7 @@ spec:
           }
           trap quit SIGTERM
           /usr/share/openvswitch/scripts/ovs-ctl start --no-ovs-vswitchd --system-id=random
+          tail --follow=name /var/log/openvswitch/ovsdb-server.log &
 
           # Restrict the number of pthreads ovs-vswitchd creates to reduce the
           # amount of RSS it uses on hosts with many cores
@@ -69,12 +70,9 @@ spec:
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-revalidator-threads=4
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
           fi
-          /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl load-kmod
 
-          tail --follow=name /var/log/openvswitch/ovs-vswitchd.log /var/log/openvswitch/ovsdb-server.log &
-          while true; do
-            sleep 10000
-          done
+          exec ovs-vswitchd unix:/var/run/openvswitch/db.sock -vconsole:info -vsyslog:info --mlockall --no-chdir --pidfile=/var/run/openvswitch/ovs-vswitchd.pid --syslog-method=unix:/dev/log
         securityContext:
           privileged: true
         volumeMounts:
@@ -90,6 +88,8 @@ spec:
           readOnly: true
         - mountPath: /etc/openvswitch
           name: host-config-openvswitch
+        - mountPath: /dev/log
+          name: host-dev-log
         resources:
           requests:
             cpu: 200m
@@ -117,6 +117,10 @@ spec:
       - name: host-config-openvswitch
         hostPath:
           path: /var/lib/openvswitch
+      - name: host-dev-log
+        hostPath:
+          path: /dev/log
+          type: Socket
       tolerations:
       - operator: "Exists"
 {{- end}}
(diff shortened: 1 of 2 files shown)


squeed:
oh, hang on, I bet this isn't going to work because it will leave orphaned copies of ovsdb

squeed:
Whew, that's not the case. Neat.

squeed:
@pecameron eventually, yes. One thing at a time.

danwinship:
> Exec ovs-vswitchd directly, so it's restarted quickly

Explain?

(Also, if we're running ovs-vswitchd directly, should we just run ovsdb directly too for consistency?)

squeed:
> > Exec ovs-vswitchd directly, so it's restarted quickly
> 
> Explain?
Oh, just that ovs-vswitchd is daemonized by `ovs-ctl start`. Then we run `ovs-ctl status` every 5 seconds as a liveness probe. By exec'ing ovs-vswitchd directly, kubelet detects any exits / crashes immediately and restarts it, rather than needing to wait for the probe to fail.

> (Also, if we're running ovs-vswitchd directly, should we just run ovsdb directly too for consistency?)
`ovs-ctl` does some db initialization as part of its startup routine that I can't separate out.

squeed:
Also, there's no way to specify the syslog path with `ovs-ctl` - you *have* to call `ovs-vswitchd` manually.

danwinship:
So we'll detect ovs-vswitchd exits/crashes directly, but will still need to wait 5 seconds to catch ovsdb exits/crashes. Maybe we should try to get some ovs-ctl improvements made upstream.

danwinship:
and we're only getting ovs-vswitchd logs logged to syslog, not ovsdb

squeed:
> So we'll detect ovs-vswitchd exits/crashes directly, but will still need to wait 5 seconds to catch ovsdb exits/crashes. Maybe we should try to get some ovs-ctl improvements made upstream.

Yup.

Realistically, ovsdb is essentially unused in this case, and quite stable, so I'm OK with that.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -47,6 +47,9 @@ spec:
           #!/bin/bash
           set -euo pipefail
 
+          # tee all output to the host's journal
+          exec 1> >(logger -p user.info --size 4096 -s -t openshift-sdn/${K8S_POD_NAME}) 2>&1
```
danwinship: This just seems terrible, but there doesn't seem to be any more obvious way of either (a) getting glog to log to syslog instead of stderr or (b) getting kubernetes to output pod logs to syslog.

squeed: years ago, as I was on the rkt / rktnetes project, we were trying to get structured logging + journald shipping in to k8s. Everyone said "no, docker log is good enough, honestly" and the project went nowhere.

Yup.

squeed:
Test is a flake:
`level=error msg="\t* module.vpc.aws_route.to_nat_gw[4]: 1 error occurred:"`

/retest

squeed:
flakes are terraform / aws issues
/retest

squeed:
All green - can I get an lgtm?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -69,12 +70,9 @@ spec:
...
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
           fi
-          /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl load-kmod
 
-          tail --follow=name /var/log/openvswitch/ovs-vswitchd.log /var/log/openvswitch/ovsdb-server.log &
-          while true; do
-            sleep 10000
-          done
+          exec ovs-vswitchd unix:/var/run/openvswitch/db.sock -vconsole:info -vsyslog:warn --mlockall --no-chdir --pidfile=/var/run/openvswitch/ovs-vswitchd.pid --syslog-method=unix:/dev/log
```
danwinship: shouldn't we be passing the same value for `-vconsole` and `-vsyslog`?

squeed: I was looking in to that, and ovs-vswitchd is spammy and it's info messages are basically useless.

danwinship: ok, so shouldn't we do `-vconsole:warn' then?

squeed: On second thought, we might as well just set them both to info.

squeed:
OK, syslog and console are now consistent.

squeed:
After further thoughts (and conversations with dan) I'm not sure this is a great idea. Closing in favor of a few smaller PRs.

PR #156, opened 2019-04-25
Use non-deprecated crio URL format

danwinship:
`W0418 15:38:05.687451    2233 util_unix.go:77] Using "/var/run/crio/crio.sock" as endpoint is deprecated, please consider using full url format "unix:///var/run/crio/crio.sock".`


squeed:
flake was terraform / aws issue
/retest

PR #157, opened 2019-04-26
Add node-selector annotation to namespace

ravisantoshgudimetla:
**Why this change?**
When https://github.com/openshift/cluster-kube-apiserver-operator/pull/394 merges, all the pods running openshift cluster will have a [defaultNodeSelector](https://docs.openshift.com/container-platform/3.11/admin_guide/managing_projects.html#setting-the-cluster-wide-default-node-selector) if it has been set by cluster-admin including pods running in `openshift-*` namespace. The main advantage of having that feature is in a multi-tenant environment, any new project created can be steered towards compute nodes(non-master nodes).

**What should I do?**
If you think, the pods in your namespace shouldn't have defaultNodeSelector set, please review this PR. If not feel free to close this. The annotation added as part of this PR lets all the pods created within your project to not have the defaultNodeSelector set.

/cc @deads2k @sjenning

### DIFF
--- a/bindata/network/openshift-sdn/003-rbac-controller.yaml
+++ b/bindata/network/openshift-sdn/003-rbac-controller.yaml
@@ -5,6 +5,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-infra
+  annotations:
+    openshift.io/node-selector: ""
 
 ---
 # The controller also requires a service account in openshift-infra
--- a/manifests/0000_07_cluster-network-operator_00_namespace.yaml
+++ b/manifests/0000_07_cluster-network-operator_00_namespace.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-network-operator
+  annotations:
+    openshift.io/node-selector: ""
   labels:
     name: openshift-network-operator
     openshift.io/run-level: "0"


bindata/network/openshift-sdn/003-rbac-controller.yaml
```diff
@@ -5,6 +5,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-infra
```
danwinship: So this isn't actually our Namespace; sdn-controller ends up referring to it because of code it shares with openshift-controller-manager, and no one else has created it yet when we bring up the network, so we have to create it ourselves. We don't create any pods in it though. I don't know if anyone else does. Anyway, whatever change gets made here needs to be coordinate with whatever other module would have been responsible for creating this Namespace if we didn't. (Origin? Installer?)

manifests/0000_07_cluster-network-operator_00_namespace.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: openshift-network-operator
+  annotations:
+    openshift.io/node-selector: ""
```
danwinship: CNO gets scheduled onto masters, so yes, this is correct

deads2k:
spoke on slack.  casey and dan winship were ok with this.  install worked, annotation took, merging

PR #159, opened 2019-04-29
openshift-sdn: add kube-proxy liveness probe

squeed:
The kube-proxy code exposes a liveness endpoint. Configure the pod to use that.


squeed:
vpc error
/retest

PR #161, opened 2019-04-30
openshift-sdn: log ovs and sdn to the system journal

squeed:
For SDN, use a bash hack to redirect all stderr and stdout to the system journal, so that logs are preserved via journald. For OVS, use the native syslog support.

Philosophically, this is because the sdn processes, although distributed via a daemonset, really are "node" components.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -69,9 +69,11 @@ spec:
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-revalidator-threads=4
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
           fi
-          /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
 
-          tail --follow=name /var/log/openvswitch/ovs-vswitchd.log /var/log/openvswitch/ovsdb-server.log &
+          /usr/share/openvswitch/scripts/ovs-ctl load-kmod
+          ovs-vswitchd unix:/var/run/openvswitch/db.sock -vconsole:info -vsyslog:info -vfile:info --mlockall --no-chdir --log-file=/var/log/openvswitch/ovs-vswitchd.log --pidfile=/var/run/openvswitch/ovs-vswitchd.pid --detach --monitor --syslog-method=unix:/dev/log
+
+          tail --follow=name /var/log/openvswitch/ovsdb-server.log &
           while true; do
             sleep 10000
           done
@@ -90,6 +92,8 @@ spec:
           readOnly: true
         - mountPath: /etc/openvswitch
           name: host-config-openvswitch
+        - mountPath: /dev/log
+          name: host-dev-log
         resources:
           requests:
             cpu: 200m
@@ -117,6 +121,10 @@ spec:
       - name: host-config-openvswitch
         hostPath:
           path: /var/lib/openvswitch
+      - name: host-dev-log
+        hostPath:
+          path: /dev/log
+          type: Socket
       tolerations:
       - operator: "Exists"
 {{- end}}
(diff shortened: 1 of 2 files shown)


squeed:
fyi @smarterclayton, in case you have some Feelings about pods logging to the journal.

squeed:
@pecameron hence the reason for sending logs to journald.

danwinship:
> Philosophically, this is because the sdn processes, although distributed via a daemonset, really are "node" components.

Which I think means we should push for real upstream support for this (eventually). This is a use case that was not anticipated when pod logging was designed (hence the upstream pushback against your earlier attempts to implement this), but it's a key part of how we're using kubernetes now.

smarterclayton:
Hold up.  Reverting this while we discuss.

I need to understand the implications of this, including whether this can cause availability issues on the cluster.  Is there a write up of those challenges?

smarterclayton:
Please reopen and do a couple of runs so we can gather measurements about how much this impacts the cluster, whether it has a change to node workload, what IO alterations we have, whether this could cause the journal to starve, etc.

This is a big change to land this close to release - while I think it is the right thing, it needs more discussion and double checking.

smarterclayton:
```
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414147    2171 roundrobin.go:240] Delete endpoint 10.131.0.29:5353 for service "openshift-dns/dns-default:dns"                                                                                                                               │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414163    2171 roundrobin.go:310] LoadBalancerRR: Setting endpoints for openshift-dns/dns-default:dns-tcp to [10.128.0.65:5353 10.128.2.25:5353 10.129.0.80:5353 10.129.2.32:5353 10.130.0.62:5353 10.131.0.29:5353]                         │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414176    2171 roundrobin.go:240] Delete endpoint 10.131.0.29:5353 for service "openshift-dns/dns-default:dns-tcp"                                                                                                                           │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414192    2171 roundrobin.go:310] LoadBalancerRR: Setting endpoints for openshift-dns/dns-default:metrics to [10.128.0.65:9153 10.128.2.25:9153 10.129.0.80:9153 10.129.2.32:9153 10.130.0.62:9153 10.131.0.29:9153]                         │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414204    2171 roundrobin.go:240] Delete endpoint 10.131.0.29:9153 for service "openshift-dns/dns-default:metrics"                                                                                                                           │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414276    2171 roundrobin.go:310] LoadBalancerRR: Setting endpoints for openshift-marketplace/redhat-operators:grpc to [10.129.2.46:50051]                                                                                                   │
│Apr 30 13:52:03 ip-10-0-170-87 openshift-sdn/sdn-684zz[2272]: I0430 13:52:03.414289    2171 roundrobin.go:240] Delete endpoint 10.129.2.46:50051 for service "openshift-marketplace/redhat-operators:grpc"
```

This doesn't belong in the system journal, for instance.

squeed:
@smarterclayton this basically brings back the logging we had in 3.9

smarterclayton:
Right, which caused problems on nodes because of the volume of output.

I am not opposed to the journal.  But I want to see that you guys have laid out the risks, the costs, whether this is introducing a new failure mode, and whether we should be adding this 4 days before code freeze.  If that discussion has happened, let me know where it is.  If it's still on going, let's add it here so we can make a full discussion.

At v(4) levels kube-proxy output can run starter level cluster nodes and cause log drops.  What's our mitigation for that here?  That was fixed in 3.10 by moving to daemonsets, but with a cost.  If we're fixing the issue, we need to consider the right fix.

smarterclayton:
Can you guys gather some of that info - other teams may want to do this and
I want to have a clear set of rules and guidelines they can crib off of.

PR #163, opened 2019-04-30
Revert "Revert "openshift-sdn: log ovs and sdn to the system journal""

squeed:
Brings #161 back

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -69,9 +69,11 @@ spec:
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-revalidator-threads=4
               ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
           fi
-          /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
 
-          tail --follow=name /var/log/openvswitch/ovs-vswitchd.log /var/log/openvswitch/ovsdb-server.log &
+          /usr/share/openvswitch/scripts/ovs-ctl load-kmod
+          ovs-vswitchd unix:/var/run/openvswitch/db.sock -vconsole:info -vsyslog:info -vfile:info --mlockall --no-chdir --log-file=/var/log/openvswitch/ovs-vswitchd.log --pidfile=/var/run/openvswitch/ovs-vswitchd.pid --detach --monitor --syslog-method=unix:/dev/log
+
+          tail --follow=name /var/log/openvswitch/ovsdb-server.log &
           while true; do
             sleep 10000
           done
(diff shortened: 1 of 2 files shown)


squeed:
for soak time:
/hold

squeed:
This basically brings our logging back to what we had in 3.9 / uncontainerized 3.10.

squeed:
@smarterclayton observes that -V=4 kube-proxy logging has caused availability / io starvation in the past, so perhaps this isn't actually safe.

My instinct would be that I'd rather have preserved logs at a lower verbosity over container-contstrained high verbosity. We gave ourselves the ability to enable per-node debugging with a sysconfig.

smarterclayton:
I liked most of the logs that were written - especially the SDN health check failure.  That was extremely valuable.  On a 20k service cluster, I think per service or per endpoint stuff can be risky.  Do we have any other "per service" output in ovs that we'd have to mask?

danwinship:
> Do we have any other "per service" output in ovs that we'd have to mask?

No, kube-proxy output makes up the bulk of sdn logging these days. (It's the sdn container, not the ovs container; the ovs container only logs stuff per-pod.)

When you say "mask", what do you mean exactly? Carrying patches to change the log level in the vendored kube-proxy code, or just grepping out high-frequency messages before piping to `logger`?


The kube-proxy logging is sometimes useful for debugging issues (eg, figuring out retroactively where 172.30.0.1 was pointing at a given time) although it's also incomplete in some ways and redundant in others. We should probably revisit exactly what gets logged anyway.

Although too late for 4.1, it wouldn't be too hard to make it possible to run the proxy and the sdn core as separate containers again (with separate logging), although that might be a lose in terms of memory usage.

squeed:
FWIW, essentially all of the loud messages come from the hybrid proxy wrapper, so we wouldn't have to carry any patches to change them.

squeed:
So, I'd like to know how much of the concern is raw disk IO vs. specific journald concerns. Is logging to journald that much more expensive than crio logs? (It could be - I've not heard great things, but I've never done any sort of benchmarking). 

As @danwinship mentioned, we've found it quite useful to indicate exactly the state of the service network at a point in time - these sorts of point-in-time RCA explorations are not uncommon.

danwinship:
> FWIW, essentially all of the loud messages come from the hybrid proxy wrapper, so we wouldn't have to carry any patches to change them.

The actual messages are logged by the upstream userspace proxy code (so they're *because of* the hybrid proxy wrapper, but not *from* the hybrid proxy wrapper).

smarterclayton:
We should be able to set log level per filename grep pattern with `-vmodule` to mask the worst offenders.

smarterclayton:
journal still has bottlenecks deep that can cause one process to block / limit another.  so hygiene for now is those who must log to journal must only do so with known bounded quantities

PR #165, opened 2019-05-01
Add toleration

ravisantoshgudimetla:
As of now, we are tolerating only `node-role.kubernetes.io/master` taint etc during scheduling stage. If we don't have the blanket toleration, there is a very good chance that these pods won't tolerate other taints that got added to the node, for example `disk-pressure` etc. The side-effect is that we will tolerate `NoExecute` taints as well

Please feel free to close this PR, if you think, you just need to tolerate the current taint(s) you have.

/cc @sjenning @derekwaynecarr

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -295,9 +295,4 @@ spec:
         hostPath:
           path: /var/run/kubernetes
       tolerations:
-      - key: "node-role.kubernetes.io/master"
-        operator: "Exists"
-        effect: "NoSchedule"
-      - key: "node.kubernetes.io/not-ready"
-        operator: "Exists"
-        effect: "NoSchedule"
+      - operator: "Exists"
--- a/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
@@ -65,9 +65,4 @@ spec:
         runAsNonRoot: true
         runAsUser: 65534
       tolerations:
-      - key: "node-role.kubernetes.io/master"
-        operator: Exists
-        effect: NoSchedule
-      - key: "node.kubernetes.io/not-ready"
-        operator: Exists
-        effect: NoSchedule
+      - operator: Exists


danwinship:
I believe the current tolerations are correct (but let's discuss):
- The CNI-plugin daemonsets (`sdn`, `ovnkube-node`, `multus`, and `sriov-cni`) and the daemonsets that provide support services for them (`ovs` and `dhcp-daemon`) tolerate all taints, because they are required on all nodes (or at least *potentially* all nodes in the case of the multus-y stuff) to bring up the network
  - actually, `sriov-device-plugin` currently only tolerates `node-role.kubernetes.io/master` and that's probably a mistake. @zshi-redhat `sriov-cni` requires `sriov-device-plugin` in order to run doesn't it? So they should be deployed to the same nodes. (In fact, we should probably deploy a single `DaemonSet` with 2 containers.)
- The "master"/"controller"-type daemonsets/deployments (`network-operator`, `sdn-controller`, `ovnkube-master`) have both a `nodeSelector` and a `toleration` for `node-role.kubernetes.io/master`, and a toleration for `node.kubernetes.io/not-ready`, so they can be brought up on masters before the network is ready. But they don't need to run on all masters, so if one of the masters has an issue like `disk-pressure` then we should run it somewhere else. And if *all* of the masters have taints then I think we've kind of already lost?

ravisantoshgudimetla:
> But they don't need to run on all masters, so if one of the masters has an issue like disk-pressure then we should run it somewhere else.

It depends for example if storage or some team has a controller that adds a particular to evict storage related pods but not networking related pods, you want to add that toleration to your pod. So, it's upto to decide if you want to add it on a case-by-case basis or have a blanket toleration to any type of taint.

zshi-redhat:
> I believe the current tolerations are correct (but let's discuss):
> 
> * The CNI-plugin daemonsets (`sdn`, `ovnkube-node`, `multus`, and `sriov-cni`) and the daemonsets that provide support services for them (`ovs` and `dhcp-daemon`) tolerate all taints, because they are required on all nodes (or at least _potentially_ all nodes in the case of the multus-y stuff) to bring up the network
>   
>   * actually, `sriov-device-plugin` currently only tolerates `node-role.kubernetes.io/master` and that's probably a mistake. @zshi-redhat `sriov-cni` requires `sriov-device-plugin` in order to run doesn't it? So they should be deployed to the same nodes. (In fact, we should probably deploy a single `DaemonSet` with 2 containers.)

Yes, `sriov-cni` requires `sriov-device-plugin` in order to run, otherwise it cannot tell which sriov device to use for a pod, so it's reasonable to schedule them on the same node. If `sriov-device-plugin` is scheduled on master node, it will report zero num of resource when no sriov-configured device found. In the future, the scheduling of `sriov-device-plugin` will likely rely on `sriov-capable` and `sriov-configured` labels specified by Node Feature Discovery(NFD). 



> * The "master"/"controller"-type daemonsets/deployments (`network-operator`, `sdn-controller`, `ovnkube-master`) have both a `nodeSelector` and a `toleration` for `node-role.kubernetes.io/master`, and a toleration for `node.kubernetes.io/not-ready`, so they can be brought up on masters before the network is ready. But they don't need to run on all masters, so if one of the masters has an issue like `disk-pressure` then we should run it somewhere else. And if _all_ of the masters have taints then I think we've kind of already lost?

squeed:
> It depends for example if storage or some team has a controller that adds a particular to evict storage related pods but not networking related pods, you want to add that toleration to your pod. So, it's upto to decide if you want to add it on a case-by-case basis or have a blanket toleration to any type of taint.

Presumably they would do that with node anti-affinity, rather than tainting the masters...

danwinship:
Filed #168 about the sriov tolerations. I think we're otherwise correct.

PR #166, opened 2019-05-01
OVN Kubernetes fixes

danwinship:
This does not get CNO+OVN into a fully-working state. But all of these fixes are correct as-is and don't depend on any as-yet-unmerged upstream stuff.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -90,13 +90,6 @@ spec:
           name: host-var-lib-ovs
         - mountPath: /var/run/openvswitch/
           name: host-var-run-ovs
-        # CNI related mounts which we take over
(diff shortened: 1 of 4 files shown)


bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -226,14 +221,17 @@ spec:
...
-      - name: host-opt-cni-bin
+      - name: host-cni-bin
         hostPath:
-          path: /opt/cni/bin
-      - name: host-etc-cni-netd
+          path: /var/lib/cni/bin
+      - name: host-cni-netd
         hostPath:
           path: /etc/kubernetes/cni/net.d
       - name: host-config-openvswitch
```
squeed: While you're at it, can you move this to `/var/lib/openvswitch` like we do for openshift-sdn? We had selinux errors with trying to write to `/etc` before, and `/etc/origin` is no more.

PR #172, opened 2019-05-08
More ovn-kubernetes fixes

danwinship:
1. At some point the `ovnkube-master` service got renamed to `ovnkube-db`, and `ovnkube.sh` changed to manually create its Endpoints after the sbdb comes up, rather than having them be created automatically by kubernetes. So update the Service and our RBAC permissions for that.
2. `ovnkube.sh` no longer hardcodes the namespace, so update to pass in the correct namespace in the pods

(Note that these changes only touch files that are ovn-kubernetes-only, so this PR has no effect on 4.1.)

### DIFF
--- a/bindata/network/ovn-kubernetes/003-rbac-controller.yaml
+++ b/bindata/network/ovn-kubernetes/003-rbac-controller.yaml
@@ -57,3 +57,33 @@ subjects:
 - kind: ServiceAccount
   name: ovn-kubernetes-controller
   namespace: openshift-ovn-kubernetes
+
+---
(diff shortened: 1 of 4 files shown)


smarterclayton:
As we get closer to release, please ensure code changes have a bug and the bug is associated in the PR title - follow the conventions described in previous emails about how to associate bugs with PRs.  The PR title must be `Bug XXXX: <description>`.

danwinship:
I assume that was a semi-automated comment, but note from the initial description:

> (Note that these changes only touch files that are ovn-kubernetes-only, so this PR has no effect on 4.1.)

PR #174, opened 2019-05-17
bindata: add resource requests to multus DS pods

sjenning:
in service to https://github.com/openshift/origin/pull/22787


dcbw:
@dougbtv @squeed 

/lgtm
/approve

PR #175, opened 2019-05-22
Give the ability to patch nodes to openshift-sdn

rcarrillocruz:
This is needed for SDN-260, which is about tainting
nodes with a too small MTU, thus the need for update permission.

### DIFF
--- a/bindata/network/openshift-sdn/002-rbac.yaml
+++ b/bindata/network/openshift-sdn/002-rbac.yaml
@@ -17,7 +17,6 @@ rules:
 - apiGroups: [""]
   resources:
   - namespaces
-  - nodes
   - endpoints
   - services
   - pods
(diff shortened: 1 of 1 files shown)


danwinship:
/hold
is there no more-targeted permission we can give, rather than the ability to update all fields of all nodes?

squeed:
> /hold
> is there no more-targeted permission we can give, rather than the ability to update all fields of all nodes?

Not that I could see. Taints are part of the Spec, not the Status.

danwinship:
It's not possible to give a node permission to update its own node but not other nodes?

rcarrillocruz:
Not really finding how to achieve that from RBAC docs.
The closest seems 'resourceName', however I don't think we can set 'update on Nodes with resourceName=foo", since the resourceName of the node is not static.

squeed:
/hold
We decided this should be patch.

PR #176, opened 2019-05-22
manifests: change image references to 4.2

squeed:
This has no real effect - since the release tooling *always* replaces these with references to builds, but it helps for attach.sh and for documentation.

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: network-operator
-        image: docker.io/openshift/origin-cluster-network-operator:v4.0.0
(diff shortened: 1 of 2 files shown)


manifests/0000_70_cluster-network-operator_03_daemonset.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: network-operator
-        image: docker.io/openshift/origin-cluster-network-operator:v4.0.0
+        image: quay.io/openshift/origin-cluster-network-operator:4.2
```
pecameron: Should this be ":latest" until the the 4.2 branch is made?

squeed: Good question. The latest tag is unstable, and moves around between releases (probably a bug, but not going to worry about that right now). The mirroring destination is independent of the branch name. So this is what we got.

PR #179, opened 2019-05-24
[WIP] Add support for chained plugins.

squeed:
This adds an additional API configuration field for chained plugins to install their desired CNI config. Then, the operator plumbs this through to openshift-sdn.

This needs openshift/origin#22898 to merge first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,19 @@ Currently, the only understood values for network Type are `OpenShiftSDN` and `O
 
 Other values are ignored. If you wish to use use a third-party network provider not managed by the operator, set the network type to something meaningful to you. The operator will not install or upgrade a network provider, but all other Network Operator functionality remains.
 
+### Adding chained plugins
+You can add raw CNI configuration snippets to be installed as CNI chained plugins. These will be included in the CNI configuration file used by the default network provider. Note that these are *not* multus multiple networks. Rather, they are for plugins that manipulate the container's existing network. For example, if you want to adjust your container's sysctls, you can configure
+
+```yaml
+spec:
+  defaultNetwork:
+    type: OpenShiftSDN
+    openshiftSDNConfig: {}
+    chainedPlugins:
+      - '{"name": "tuning", "sysctl": { "net.core.somaxconn": "500"}}`
+```
+
+Not all default networks support chained plugins. Currently, only OpenShiftSDN does.
 
 ### Configuring OpenShiftSDN
 OpenShiftSDN supports the following configuration options, all of which are optional:
@@ -116,14 +129,7 @@ These configuration flags are only in the Operator configuration object.
 
 Example:
 ```yaml
-spec:
-  defaultNetwork:
-    type: OpenShiftSDN
-    openshiftSDNConfig:
-      mode: NetworkPolicy
-      vxlanPort: 4789
-      mtu: 1450
-      useExternalOpenvswitch: false
+
 ```
 
 ### Configuring OVNKubernetes
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -6,6 +6,9 @@ metadata:
 data:
   sdn-config.yaml: |-
 {{.NodeConfig | indent 4}}
+  80-openshift.conflist: |-
+{{.CNIConfig | indent 4}}
+
 ---
 kind: DaemonSet
 apiVersion: apps/v1
@@ -48,7 +51,6 @@ spec:
           set -euo pipefail
 
           # if another process is listening on the cni-server socket, wait until it exits
-          trap 'kill $(jobs -p); rm -f /etc/cni/net.d/80-openshift-network.conf ; exit 0' TERM
           retries=0
           while true; do
             if echo 'test' | socat - UNIX-CONNECT:/var/run/openshift-sdn/cni-server.sock &>/dev/null; then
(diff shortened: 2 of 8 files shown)


squeed:
TODO: document the final conclusions in README.

vendor/github.com/openshift/api/operator/v1/types_network.go
```diff
@@ -94,6 +94,10 @@ type DefaultNetworkDefinition struct {
 	// not implemented.
 	// +optional
 	OVNKubernetesConfig *OVNKubernetesConfig `json:"ovnKubernetesConfig,omitempty"`
+
+	// ChainedPlugins is an array of entries to insert as chained plugins for execution
+	// after the default network.
+	ChainedPlugins []string `json:"chainedPlugins,omitempty"`
```
squeed: I'm not super happy about this API. We should probably give ourselves room to expand. For example, we could also include an explicit tuning configuration, rather than requiring raw CNI.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -77,11 +79,11 @@ spec:
           fi
 
           # Take over network functions on the node
-          rm -f /etc/cni/net.d/80-openshift-network.conf
+          rm -f /etc/cni/net.d/80-openshift-network.conf {{.CNIPath}}
```
dcbw: Maybe add a comment to keep the `rm -f 80-openshift-network.conf` for upgrade reasons?

pkg/network/cni.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	for i, entry := range conf.DefaultNetwork.ChainedPlugins {
+		var m map[string]interface{}
+		if err := json.Unmarshal([]byte(entry), &m); err != nil {
+			out = append(out, errors.Wrapf(err, "invalid json in Spec.DefaultNetwork.ChainedPlugin[%d]", i))
+			continue
+		}
+
+		_, ok := m["type"]
```
dcbw: `if _, ok := m["type"]; !ok {` ? :)

pkg/network/cni.go
```diff
@@ -0,0 +1,61 @@
...
+			out = append(out, errors.Errorf("invalid CNI plugin entry in Spec.DefaultNetwork.ChainedPlugin[%d]: must have 'type' key", i))
+		}
+	}
+
+	return out
+}
+
+// makeCNIConfig merges the "primary" plugin config with any desired
+func makeCNIConfig(conf *operv1.NetworkSpec, netName, cniVersion, primaryPluginConfig string) string {
+
```
dcbw: lame empty newline

pkg/network/cni.go
```diff
@@ -0,0 +1,61 @@
...
+		_, ok := m["type"]
+		if !ok {
+			out = append(out, errors.Errorf("invalid CNI plugin entry in Spec.DefaultNetwork.ChainedPlugin[%d]: must have 'type' key", i))
+		}
+	}
+
+	return out
+}
+
+// makeCNIConfig merges the "primary" plugin config with any desired
```
dcbw: perhaps say it merges the primary "plugin config JSON" with any chained plugin JSON config? eg that these random strings are actually JSON.

vendor/github.com/openshift/api/operator/v1/types_network.go
dcbw: This should say what the strings are supposed to be; eg CNI JSON configuration in string form.

pkg/network/openshift_sdn.go
```diff
@@ -52,6 +53,9 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	}
 	data.Data["NodeConfig"] = nodeCfg
 
+	data.Data["CNIConfig"] = makeCNIConfig(conf, CNINetworkName, "0.3.1",
+		`{"type": "openshift-sdn"}`)
```
dcbw: Not super happy about this, can we just have the SDN itself do the combination? eg the CNO makes the template from the ChainedPlugins and the SDN inserts itself at position 0 in the list and then writes the result conflist to the final CNI configdir that CRIO looks for?

danwinship: I was going to say something similar before, but there is already a crazy amount of coupling between the operator and the SDN plugin itself anyway. This doesn't really make it worse. (And besides, what happens if one of the chained plugins wants to use some feature that requires using version 0.4.0 of the spec?)

squeed: Yeah, I thought about that for a long time, and couldn't find a solution I like.

First of all, the cniVersion is a tricky coordination point no matter where you place it. It needs to be the same for all plugins, and all plugins need to support the selected versions. If we bump openshift-sdn to 0.4.0, then all chained plugins instantly *must* support 0.4.0. I suppose this is a problem with CNI, and we may wish to revisit that decision. Wherever we do the config munging doesn't fix this problem.

As for having openshift-sdn insert itself, I decided against that for a few reasons:

1. In general, manipulating untyped json is a pain in go :-)
2. This is easier add to other plugins, e.g. ovn. We can just write a sidecar in bash that waits for healthz and copies a file.
3. Generally, CNI configuration is expected to be administrator-defined. openshift-sdn writes it's own configuration file for availability signalling and historical reasons. And because it has no knobs.

I'm happy to revisit this. It's not great either way you look at it.

vendor/github.com/openshift/api/operator/v1/types_network.go
squeed: OK, changed the API slightly, and added better documentation.

bindata/network/openshift-sdn/sdn.yaml
squeed: done

README.md
```diff
@@ -104,6 +104,19 @@ Currently, the only understood values for network Type are `OpenShiftSDN` and `O
 
 Other values are ignored. If you wish to use use a third-party network provider not managed by the operator, set the network type to something meaningful to you. The operator will not install or upgrade a network provider, but all other Network Operator functionality remains.
 
+### Adding chained plugins
+You can add raw CNI configuration snippets to be installed as CNI chained plugins. These will be included in the CNI configuration file used by the default network provider. Note that these are *not* multus multiple networks. Rather, they are for plugins that manipulate the container's existing network. For example, if you want to adjust your container's sysctls, you can configure
```
dcbw: Should note that this applies to *all containers* in the cluster.

danwinship:
So I keep having this thought that we should fork multus into "CNO CNI" which would be the top-level CNI plugin that would run for all configurations, and would include the existing multus functionality, plus anything else we needed to do either sometimes (chained Istio) or always (anti-MCS iptables hacks) or to some-but-not-all-pods regardless of network plugin (assign-macvlan), ...

dcbw:
Alternatively, we can have Istio run via Multus. Casey verified that Istio's CNI plugin doesn't actually use PrevResult which means that it's not using anything that chained plugins actually provide. And since Multus looks in a specific CNO-defined directory for openshift-sdn's CNI config file before indicating node readiness, we don't have an ordering issue where Istio could come before openshift-sdn, because istio's CNI config will be in a different directory.

We can avoid adding stuff to the CNO API to support Istio for 4.2 and figure out what the API should actually be for Istio and other chained plugins next round.

squeed:
Yeah, closing this for now. We can revisit this as needed,

PR #181, opened 2019-05-27
openshift-sdn: ovs shutdown ordering.

squeed:
When looking in to other bugs, I saw a lot of errors from kubelet:

Pod "ovs-XXX" is terminated, but some containers are still running.

It looks like the SIGTERM isn't being respected. Let's fix this two ways: Stop running tail in an infinite loop (which might cause the signal to be ignored), and tell the kubelet to stop OVS explicitly.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -71,10 +71,9 @@ spec:
           fi
           /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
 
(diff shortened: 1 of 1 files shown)


squeed:
/hold
It still takes 30 seconds or more for ovs to stop.

squeed:
/lgtm cancel
something funky is going on.

squeed:
Turns out the funkiness is a larger issue; this change is fine. Made a few other improvements.

/hold cancel

squeed:
Okay, tested this new version and it correctly terminates the process. Added a readiness probe while I was at it.
PTAL

PR #182, opened 2019-05-29
Render the default ClusterNetwork

danwinship:
Currently CNO writes out a controller-config.yaml for sdn-controller, which the sdn-controller then turns into the default ClusterNetwork object. But this is just historical baggage, and there's no reason the CNO couldn't just create the ClusterNetwork itself.

Currently sdn-controller requires that it receives a controller-config on startup, and it will get annoyed if it finds a pre-existing ClusterNetwork object that doesn't match the controller-config. So we still have to create the controller-config as well. But once this has merged, we can change sdn-controller to read its config from the ClusterNetwork object instead and not need a config file.

### DIFF
--- a/bindata/network/openshift-sdn/005-clusternetwork.yaml
+++ b/bindata/network/openshift-sdn/005-clusternetwork.yaml
@@ -1,0 +1,1 @@
+{{.ClusterNetwork}}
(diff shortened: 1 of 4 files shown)


squeed:
Are these test flakes, or are we somehow stomping on the ClusterNetwork created in the upgrade case?

danwinship:
e2e-aws-upgrade is failing everywhere. Presumably the kube-controller-manager thing that they thought might have been caused by #181.

PR #183, opened 2019-05-30
Revert "openshift-sdn: ovs shutdown ordering."

wking:
This reverts commit 6a3ff5cf3d1130318b640c806a958bf736db6fc7, #181.

I'm not clear on the connection, but this change seems to have [increased the likelihood of fatal upgrade timouts][1], and that's blocking org-wide master progress.  Ideally we'd fix [the underlying issue][2], but this reversion should help unblock master CI while we work on that.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1715515
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1714699

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -71,9 +71,10 @@ spec:
           fi
           /usr/share/openvswitch/scripts/ovs-ctl start --no-ovsdb-server --system-id=random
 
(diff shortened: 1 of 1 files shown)


danwinship:
/lgtm
It seems like basically our infrastructure was accidentally depending on the fact that we are not shutting OVS down correctly...

squeed:
So, did this fix anything? I would be surprised if it did. Either way, kubelet killed OVS.... it just took an extra 60 seconds of thumb-twiddling to do so.

PR #186, opened 2019-06-04
run-locally.sh: add from-scratch cluster up capability

dcbw:
```
hack/run-locally.sh new -n ovn -i /path/to/openshift/installer -c /tmp/ovntest -m "quay.io/openshift/origin-ovn-kubernetes:latest"
hack/run-locally.sh override -c /tmp/ovntest
```

@squeed @rcarrillocruz @danwinship @pecameron @JacobTanenbaum

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -64,43 +64,43 @@ BUILDCMD="docker build" ./hack/build-image.sh
 
 # Running local builds with the installer
 
-If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator.
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator. This script will either create a new cluster and run openshift-install for you, or attach to an existing cluster and restart the network operator with overrides.
 
 This requires `oc` and `kubectl` in your $PATH.
 
-## Steps
+The OpenShift installer is available from https://github.com/openshift/installer and may be built from git.
 
-You will need to execute all steps for every installation, since the installer deletes the intermediate files we customize.
+## Steps for creating a new cluster
 
-1. If you haven't already, build the operator:
+This mode also requires `openshift-install` in your $PATH or explicitly passed to `hack/run-locally.sh` with the `-i <installer binary>` option.
+
+1. If you haven't already, build the cluster-network-operator:
 
 ```
 hack/build-go.sh
 ```
 
-2. In one window, prepare the cluster:
+2. Run `hack/run-locally.sh`, optionally passing an existing install-config.yaml (`-f` option), network plugin to use (`-n` option), a network plugin container development image (`-m` option), and a request to wait after creating manifests (`-w` option) for additional manual overrides. Available network plugins are `sdn`/`OpenShiftSDN` and `ovn`/`OVNKubernetes`.
 
 ```
 export CLUSTER_DIR=<your cluster directory>
-openshift-install --dir=$CLUSTER_DIR create manifests
+hack/run-locally.sh -n OpenShiftSDN -m "docker.io/myawesome/origin-node:latest"
 ```
 
-3. In another window, start the operator:
+It will print status messages as it waits for the installer to make progress.
+
+Passing the `-w` option will wait after creating manifests to allow for additional manual overrides before the cluster is started.
+
+## Steps for "attach" mode
+
+If you have an already-up cluster you would like to take over, then you can instead run:
 
 ```
 export CLUSTER_DIR=<your cluster directory>
 hack/run-locally.sh
 ```
 
-It will print status messages as it waits for the installer to make progress.
-
-4. Optionally, override image references. If you are doing downstream image development, e.g. working on openshift-sdn, you should build, publish, and edit `$CLUSTER_DIR/env.sh` to point to your development image.
-
-5. In the installer window, start the installer:
-
-```
-openshift-install --dir=$CLUSTER_DIR create cluster
-```
+It will stop the deployed operator, set up the environment variables, and execute your local build of the operator. A network plugin containter development image (`-m` option) may also be given.
 
 ## Tips, Tricks, & Limitations
 
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -2,7 +2,7 @@
 
 # run-locally.sh is some scaffolding around running a local instance of the
 # network operator with the installer.
-# See INSTALLER-HACKING.md
+# See HACKING.md
 
 set -o errexit
 set -o nounset
@@ -31,13 +31,43 @@ function override_install_manifests() {
     fi
 }
 
-# Extract environment variables for the CNO DaemonSet
-function setup_operator_env() {
+# Extract environment from a running cluster
+function extract_environment_from_running_cluster() {
+    if [[ ! -e "${CLUSTER_DIR}/env.sh" ]]; then
+        echo "Copying environment variables from manifest to ${CLUSTER_DIR}/env.sh"
+        kubectl get deployment -n openshift-network-operator network-operator -ojsonpath='{range .spec.template.spec.containers[0].env[?(@.value)]}{.name}{"="}{.value}{"\n"}' > "${CLUSTER_DIR}/env.sh"
+    fi
+}
+
+# Extract environment from the installer's cluster-network-operator manifests
+function extract_environment_from_manifests() {
     if [[ manifests/0000_70_cluster-network-operator_03_daemonset.yaml -nt "${CLUSTER_DIR}/env.sh" ]]; then
         echo "Copying environment variables from manifest to ${CLUSTER_DIR}/env.sh"
         oc --kubeconfig=hack/null-kubeconfig patch --local=true -f manifests/0000_70_cluster-network-operator_03_daemonset.yaml -p '{}' -ojsonpath='{range .spec.template.spec.containers[0].env[?(@.value)]}{.name}{"="}{.value}{"\n"}' > "${CLUSTER_DIR}/env.sh"
     fi
+}
+
+# Update cluster-network-operator environment variables
+function setup_operator_env() {
+    local image_env_key="$1"
+    local plugin_image="$2"
+
     sed -i -e "s/^RELEASE_VERSION=.*/RELEASE_VERSION=${RELEASE_VERSION}/" "${CLUSTER_DIR}/env.sh"
+
+    if [[ -n "${plugin_image}" ]]; then
+        sed -i -e "s#^${image_env_key}=.*#${image_env_key}=${plugin_image}#" "${CLUSTER_DIR}/env.sh"
+    fi
+}
+
+# Patch the CNO out of the cluster-version-operator and scale it down to zero so
+# we can run our local CNO instead
+function stop_deployed_operator() {
+    kubectl patch --type=json -p "$(cat hack/overrides-patch.yaml)" clusterversion version
+
+    if [[ -n "$(kubectl get deployments -n openshift-network-operator 2> /dev/null | grep network-operator)" ]]; then
+        echo "Scaling the deployed network operator to 0"
+        kubectl scale deployment -n openshift-network-operator network-operator --replicas 0
+    fi
 }
 
 # wait_for_cluster waits for the cluster to come up and sets some variables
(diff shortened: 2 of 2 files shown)


hack/run-locally.sh
```diff
@@ -33,11 +33,18 @@ function override_install_manifests() {
 
 # Extract environment variables for the CNO DaemonSet
 function setup_operator_env() {
+    local image_env_key=${1}
```
danwinship: local image_env_key="$1"

etc

hack/run-locally.sh
```diff
@@ -33,11 +33,18 @@ function override_install_manifests() {
...
+    local plugin_image=${2}
+
     if [[ manifests/0000_70_cluster-network-operator_03_daemonset.yaml -nt "${CLUSTER_DIR}/env.sh" ]]; then
         echo "Copying environment variables from manifest to ${CLUSTER_DIR}/env.sh"
         oc --kubeconfig=hack/null-kubeconfig patch --local=true -f manifests/0000_70_cluster-network-operator_03_daemonset.yaml -p '{}' -ojsonpath='{range .spec.template.spec.containers[0].env[?(@.value)]}{.name}{"="}{.value}{"\n"}' > "${CLUSTER_DIR}/env.sh"
     fi
     sed -i -e "s/^RELEASE_VERSION=.*/RELEASE_VERSION=${RELEASE_VERSION}/" "${CLUSTER_DIR}/env.sh"
+
+    if [[ -n "${image_env_key}" ]]; then
+        sed -i -e "s#^${image_env_key}=.*#${image_env_key}=${plugin_image}#" "${CLUSTER_DIR}/env.sh"
```
danwinship: This should be `if [[ -n "${plugin_image}" ]];` I think; `${IMAGE_ENV_KEY}` always gets set

hack/run-locally.sh
```diff
@@ -94,11 +101,64 @@ function wait_for_cluster() {
...
+            echo "Option -${OPTARG} requires an argument." >&2
+            exit 1
+            ;;
+    esac
+done
+
 if [[ -z "${CLUSTER_DIR:-}" ]]; then
-    echo "error: CLUSTER_DIR must be set"
+    echo "error: CLUSTER_DIR must be set or '-c <cluster-dir>' must be given"
     echo "For more info, see INSTALLER-HACKING.md"
```
danwinship: remember to update `INSTALLER-HACKING.md` once this is all finalized

hack/run-locally.sh
```diff
@@ -94,11 +101,64 @@ function wait_for_cluster() {
     fi
 }
 
+function from_scratch() {
+    if [[ -n "${INSTALLER_DIR:-}" ]]; then
```
danwinship: It seems to me that you should be able to decide if this is "from scratch" based on the contents of `CLUSTER_DIR`...

hack/run-locally.sh
```diff
@@ -94,11 +101,64 @@ function wait_for_cluster() {
...
-    echo "error: CLUSTER_DIR must be set"
+    echo "error: CLUSTER_DIR must be set or '-c <cluster-dir>' must be given"
     echo "For more info, see INSTALLER-HACKING.md"
     exit 1
 fi
+mkdir -p ${CLUSTER_DIR} >& /dev/null
+
+if from_scratch ; then
+    echo "Creating install config..."
+    ${INSTALLER_DIR}/bin/openshift-install --dir=${CLUSTER_DIR} create install-config
```
danwinship: That kind of looks like you're expecting the caller to pass the path to the installer source tree... People are not expected to build the installer from source any more unless they are hacking on the installer. You're supposed to download a pre-built binary.

You could take a path to the `openshift-install` binary itself instead. And make it optional and just run it from `$PATH` if it's not specified.

hack/run-locally.sh
```diff
@@ -94,11 +101,64 @@ function wait_for_cluster() {
...
 }
 
+function from_scratch() {
+    if [[ -n "${INSTALLER_DIR:-}" ]]; then
+        return 0
+    fi
+    return 1
+}
+
+PLUGIN_IMAGE=${PLUGIN_IMAGE:-}
```
danwinship: Always put quotes around variable expansions. (And for `${FOO}/bar/baz`, put quotes around the whole path.)

hack/run-locally.sh
```diff
@@ -111,7 +171,22 @@ if [[ -e "${CLUSTER_DIR}/manifests/cvo-overrides.yaml" ]]; then
...
+    ID=$!
+    echo "Started cluster installer PID ${ID}"
+fi
+
 wait_for_cluster
-setup_operator_env
+
+if [[ "${NETWORK_PLUGIN}" == "OVNKubernetes" ]]; then
+    echo "Attempting to open OVN ports for AWS..."
+    while ! hack/open-ovn-ports.sh >& /dev/null; do
```
danwinship: You don't need to retry this. The security groups have definitely been created after `wait_for_cluster` returns.

hack/run-locally.sh
dcbw: I picked installer_dir because (a) we need it to do from_scratch, and (b) you don't need it to do the not-from-scratch. You need cluster-dir either way.

hack/run-locally.sh
```diff
@@ -94,11 +101,64 @@ function wait_for_cluster() {
...
-    echo "error: CLUSTER_DIR must be set"
+    echo "error: CLUSTER_DIR must be set or '-c <cluster-dir>' must be given"
     echo "For more info, see INSTALLER-HACKING.md"
     exit 1
 fi
+mkdir -p "${CLUSTER_DIR}" >& /dev/null
+
+if from_scratch ; then
+    echo "Creating install config..."
+    "${INSTALLER_PATH}/openshift-install" --dir="${CLUSTER_DIR}" create install-config
```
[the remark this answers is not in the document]
dcbw: @squeed I've added an option to pass an existing install-config.yaml which will then be copied to the install dir and  have its network plugin fixed up.

hack/run-locally.sh
dcbw: @danwinship your suggestion is better, fixing that up.

hack/run-locally.sh
dcbw: done.

dcbw:
> Could use a "usage" or "help"

@pecameron done

danwinship:
/hold
@pecameron be careful about `/lgtm`-ing PRs when other people are still reviewing them.

pecameron:
@danwinship how are we supposed to handle this? After reviewing do we get together on slack to discuss and agree for one of us to set the l g t m tag? I thought the requires changes requested tag prevented the merge?
What do you think?

dcbw:
@pecameron no, once a PR has lgtm and approve it is ready to merge. I wouldn't do an lgtm until you're OK with the PR merging.

dcbw:
@pecameron updated HACKING.md for your comments, PTAL thanks!

HACKING.md
```diff
@@ -102,6 +104,37 @@ It will print status messages as it waits for the installer to make progress.
...
+
+```
+hack/build-go.sh
+```
+
+2. Run `hack/run-locally.sh`, optionally passing an existing install-config.yaml (`-f` option), network plugin to use (`-n` option), and network plugin container development image (`-m` option). Available network plugins are `sdn` (OpenShift SDN) and `ovn` (ovn-kubernetes).
+
+```
+export CLUSTER_DIR=<your cluster directory>
+hack/run-locally.sh new -n OpenShiftSDN -m "docker.io/myawesome/origin-node:latest"
```
pecameron: "-n OpenShiftSDN" is not among the "Available network plugins" above.

hack/run-locally.sh
```diff
@@ -94,24 +121,184 @@ function wait_for_cluster() {
...
+
+    # Update the requested network plugin
+    sed -i -e "s/networkType: [a-zA-Z]*$/networkType: $NETWORK_PLUGIN/" "${CLUSTER_DIR}/install-config.yaml"
+
+    echo "Creating install manifests..."
+    "${INSTALLER_PATH}" --dir="${CLUSTER_DIR}" create manifests
+fi
+
+if [[ "${MODE}" != "attach" ]]; then
+    if [[ ! -e "${CLUSTER_DIR}/manifests" && ! -e "${CLUSTER_DIR}/auth" ]]; then
```
squeed: Probably only want to check for `/auth` in this case.

hack/run-locally.sh
```diff
@@ -94,24 +121,184 @@ function wait_for_cluster() {
...
+        echo "error: cannot find installer state; please run"
+        echo "  openshift-install --dir=${CLUSTER_DIR} create manifests"
```
squeed: `create cluster` would be better, right?

dcbw: @squeed actually no, because we create-cluster later in the 'new' case below, and at this point all we really need are the manifests.

dcbw:
@danwinship @pecameron @squeed updated, PTAL thanks!

hack/run-locally.sh
```diff
@@ -94,24 +121,186 @@ function wait_for_cluster() {
...
+    OPTIND=2
+    while getopts "c:f:i:m:n:" opt; do
+        case $opt in
+            c) CLUSTER_DIR="${OPTARG}";;
+            f) INSTALL_CONFIG="${OPTARG}";;
+            i) INSTALLER_PATH="${OPTARG}";;
+            m) PLUGIN_IMAGE="${OPTARG}";;
+            n)
+                case ${OPTARG} in
+                    ovn)
```
rcarrillocruz: This fallthrough is not working for me, I get "hack/run-locally.sh: line 193: syntax error near unexpected token `)'".
Doing something like "ovn" | "OVNKubernetes" works and would catch both.

Same for sdn/OpenShiftSDN

dcbw: @rcarrillocruz fixed, PTAL thanks!

hack/run-locally.sh
```diff
@@ -31,13 +31,40 @@ function override_install_manifests() {
...
+
     sed -i -e "s/^RELEASE_VERSION=.*/RELEASE_VERSION=${RELEASE_VERSION}/" "${CLUSTER_DIR}/env.sh"
+
+    if [[ -n "${plugin_image}" ]]; then
+        sed -i -e "s#^${image_env_key}=.*#${image_env_key}=${plugin_image}#" "${CLUSTER_DIR}/env.sh"
+    fi
+}
+
+# Install our overrides in to an existing cluster
+function stop_deployed_operator() {
```
danwinship: I initially wrote "comment does not match function", but I guess it does, except that the comment and the function name describe what the function is doing from completely different perspectives, so they seem mismatched

hack/run-locally.sh
```diff
@@ -94,24 +121,184 @@ function wait_for_cluster() {
...
+
+The following options are accepted for 'new' mode:
+
+ -f               the path to an openshift-install-created install-config.yaml file; if not given one will be created
+ -i               path to the openshift-install binary; if not given PATH will be searched
+ -m               custom network plugin container image name
+ -n [net plugin]  the name of the network plugin to deploy; one of [sdn|OpenShiftSDN|ovn|OVNKubernetes]
+
+The following environment variables are honored:
+ - CLUSTER_DIR: the cluster installation temp directory
```
danwinship: Above you say that `-c` is required... I don't think it's really that useful to allow both forms. (Also, the examples in HACKING.md are inconsistent.)

dcbw: @danwinship I think it's still useful, eg if you're going to use the same cluster temp dir a lot to re-test the CNO in a running cluster. But I'd probably just use -c for most stuff becuase the command-line looks better that way (to me) especially with CTRL+R. I updated the help text. PTAL.

dcbw:
@danwinship updated, PTAL thanks!

HACKING.md
```diff
@@ -64,15 +64,17 @@ BUILDCMD="docker build" ./hack/build-image.sh
 
 # Running local builds with the installer
 
-If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator.
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator. This script will either override existing rendered cluster manifests ("override" mode), create a new cluster ("new" mode) and run openshift-install for you, or attach to an existing cluster ("attach" mode) and restart the network operator with overrides.
 
 This requires `oc` and `kubectl` in your $PATH.
 
-## Steps
+The OpenShift installer is available from https://github.com/openshift/installer and may be built from git.
```
pecameron: Might mention the go version.

dcbw: @pecameron updated PTAL thanks.

dcbw:
@rcarrillocruz @danwinship any more comments on this one?

HACKING.md
```diff
@@ -64,15 +64,17 @@ BUILDCMD="docker build" ./hack/build-image.sh
 
 # Running local builds with the installer
 
-If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator.
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator. This script will either override existing rendered cluster manifests ("override" mode), create a new cluster ("new" mode) and run openshift-install for you, or attach to an existing cluster ("attach" mode) and restart the network operator with overrides.
 
 This requires `oc` and `kubectl` in your $PATH.
 
-## Steps
+The OpenShift installer is available from https://github.com/openshift/installer and may be built from git using Go 1.12 or later.
```
danwinship: Don't say what version of go the installer needs. We don't want to have to update our README every time they change something.

HACKING.md
```diff
@@ -64,15 +64,17 @@ BUILDCMD="docker build" ./hack/build-image.sh
 
 # Running local builds with the installer
 
-If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator.
+If you want to run a local build with the installer, you need to do a bit of hacking. There is a script that will do most of the dirty work for you. It will modify the installer so the network operator doesn't run, and spin up a local copy of the network operator. This script will either override existing rendered cluster manifests ("override" mode), create a new cluster ("new" mode) and run openshift-install for you, or attach to an existing cluster ("attach" mode) and restart the network operator with overrides.
```
danwinship: So thinking about this some more...

One nice thing about the current system is that you can run `./hack/run-locally.sh` to bring up a cluster, do some stuff, figure out the bug you were investigating, hit `^C` to stop the CNO, edit some files, `./hack/build-go.sh`, and then `./hack/run-locally.sh` again to run the newly-rebuilt CNO against the already-running cluster.

I guess the new version still supports that, but you have to use different arguments the second time you run `./hack/run-locally.sh` to make it do `attach` rather than `new`. It would be cool if you didn't have to do that.

hack/run-locally.sh
```diff
@@ -31,13 +31,41 @@ function override_install_manifests() {
...
+    if [[ -n "${plugin_image}" ]]; then
+        sed -i -e "s#^${image_env_key}=.*#${image_env_key}=${plugin_image}#" "${CLUSTER_DIR}/env.sh"
+    fi
+}
+
+# Patch the CNO out of the cluster-version-operator and scale it down to zero so
+# we can run our local CNO instead
+function stop_deployed_operator() {
+    echo "Telling the CVO to ignore the network operator"
+    kubectl patch --type=json -p "$(cat hack/overrides-patch.yaml)" clusterversion version
```
danwinship: So, tying into my earlier comment about running the script, killing it, rebuilding, and running again, note that the existing code goes out of its way to not print messages claiming it's doing something if that thing has already been done. (ie, ideally, if you attach to a cluster that is already not running a CNO, it shouldn't tell you that it's stopping the CNO.)

danwinship:
/lgtm
let's get this in. if it has problems, we'll discover them

PR #189, opened 2019-06-07
adding alerts for sdn in prometheus

JacobTanenbaum:
adding 2 initial alerts to openshift-sdn
  1. Adding alert to check if any containers in the openshift-sdn are in crashloopbackoff
  2. Adding alert if there is a mismatch between the number of nodes and the number of ovs containers

### DIFF
--- a/bindata/network/openshift-sdn/alert-rules.yaml
+++ b/bindata/network/openshift-sdn/alert-rules.yaml
@@ -1,0 +1,41 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: networking-rules
+  namespace: openshift-sdn
+spec:
+  groups:
+  - name: general.rules
+    rules:
+    - alert: NodeWithoutOVSPod
+      annotations:
+        message: |
+          All nodes should be running an ovs pod, {{"{{"}} $labels.node {{"}}"}} is not.
+      expr: |
+        (kube_node_info unless on(node) kube_pod_info{namespace="openshift-sdn",  pod=~"ovs.*"}) > 0
+      for: 20m
+      labels:
+        severity: warning
+    - alert: NodeWithoutSDNPod
+      annotations:
+        message: |
+          All nodes should be running an sdn pod, {{"{{"}} $labels.node {{"}}"}} is not.
+      expr: |
+        (kube_node_info unless on(node) kube_pod_info{namespace="openshift-sdn",  pod=~"sdn.*"}) > 0
+      for: 20m
+      labels:
+        severity: warning
+    - alert: NetworkPodsCrashLooping
+      annotations:
+        message: Pod {{"{{"}} $labels.namespace{{"}}"}}/{{"{{"}} $labels.pod{{"}}"}} ({{"{{"}} $labels.container
+          {{"}}"}}) is restarting {{"{{"}} printf "%.2f" $value {{"}}"}} times / 5 minutes.
+      expr: |
+        rate(kube_pod_container_status_restarts_total{namespace="openshift-sdn"}[15m]) * 60 * 5 > 0
+      for: 1h
+      labels:
+        severity: critical


JacobTanenbaum:
@squeed initial PR for alertmanager alerts in cluster-network-operator

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,29 @@
...
+spec:
+  groups:
+  - name: general.rules
+    rules: 
+    - alert: NodeWithoutOVSPod
+      annotations:
+        message: |
+          All nodes should be running an ovs pod. One or more are not 
+      expr: |
+        :kube_pod_info_node_count: != sum(kube_pod_container_info{namespace="openshift-sdn", pod=~"ovs.*"})
```
danwinship: OK, so I don't know a lot about prometheus. Will this trigger immediately upon that condition changing? We don't want to have it generating "alerts" during an upgrade (when each node is expected to briefly not be running OVS while it restarts) or when a new node is added to the cluster (in which case the "node count" will presumably go up before the OVS pod gets a chance to start).

JacobTanenbaum: It would trigger immediately, what kind of a delay would be appropriate?

danwinship: Hm... I'm not sure this check makes sense at all actually. If the pods are *crashing* then that's alert-worthy, but that should be caught by the other check. If they are not running for some reason other than "because they crashed" then presumably it's intentional.

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+      for: 1h
+      labels:
+        severity: warning
+    - alert: NetworkPodsCrashLooping
+      annotations:
+        message: Pod {{"{{"}} $labels.namespace{{"}}"}}/{{"{{"}} $labels.pod{{"}}"}} ({{"{{"}} $labels.container
+          {{"}}"}}) is restarting {{"{{"}} printf "%.2f" $value {{"}}"}} times / 5 minutes.
+      expr: |
+        rate(kube_pod_container_status_restarts_total{namespace="openshift-sdn"}[15m]) * 60 * 5 > 0
+      for: 1h
```
squeed: This can probably reduced to 20m

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  groups:
+  - name: general.rules
+    rules:
+    - alert: NodeWithoutOVSPod
+      annotations:
+        message: |
+          All nodes should be running an ovs pod. One or more are not
+      expr: |
+        :kube_pod_info_node_count: != sum(kube_pod_container_info{namespace="openshift-sdn", pod=~"ovs.*"})
```
squeed: Is there a way to alert on the specific node(s) missing the pod?

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: monitoring.coreos.com/v1
```
squeed: nit - can you call this `alert-rules.yaml` or something like that?

squeed:
@brancz has offered to give this a proper review, thanks!

squeed:
As I think about this, I think it would be good if these alerts were combined in to one. What I want is one alert per node, indicating that the node does not have a running OVS. It shouldn't matter if ovs is crashed, or somehow hasn't been scheduled.

It's important for this alert to be per-node:

- It makes the `for` directive actually useful
- It lets administrators debug the issue
- In large clusters, a unitary alert might always be firing.

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: PrometheusRule
+metadata:
+  labels:
+  prometheus: k8s
```
brancz: this looks incorrectly indented

bindata/network/openshift-sdn/rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+        :kube_pod_info_node_count: != sum(kube_pod_container_info{namespace="openshift-sdn", pod=~"ovs.*"})
+      for: 1h
+      labels:
+        severity: warning
+    - alert: NetworkPodsCrashLooping
+      annotations:
+        message: Pod {{"{{"}} $labels.namespace{{"}}"}}/{{"{{"}} $labels.pod{{"}}"}} ({{"{{"}} $labels.container
+          {{"}}"}}) is restarting {{"{{"}} printf "%.2f" $value {{"}}"}} times / 5 minutes.
+      expr: |
+        rate(kube_pod_container_status_restarts_total{namespace="openshift-sdn"}[15m]) * 60 * 5 > 0
```
brancz: this is already covered by the general alerts in the cluster monitoring stack: https://github.com/openshift/cluster-monitoring-operator/blob/5092b55844dfec03d61f0c767f01da887e95f40c/assets/prometheus-k8s/rules.yaml#L527-L535

bindata/network/openshift-sdn/rules.yaml
brancz: Yes this is possible, you can use the `kube_pod_info` metric (with the appropriate selector) and join them onto the `kube_node_info` metric. That way we'll be able to figure out which nodes have such a pod and which doesn't.

squeed:
So, I chatted with @brancz, and he told me that we *definitely* want to write our own alert for an SDN or OVS pod is crashlooping. This is because the generic "openshift pods are crashlooping" alert specifically *does not* include a `node` label.

We want the `node` label. Pretty soon, if you view a node object in the web console, alerts firing for that node will be explicitly surfaced, and we want to be included there. So, yes, we should continue with this approach, and write the following alerts

* Node has crashlooping {SDN, OVS} pod
* Node has no scheduled {SDN, OVS} pod (for too long)

bindata/network/openshift-sdn/alert-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+        message: |
+          All nodes should be running an ovs pod. {{ $labels.node }} is not.
+      expr: |
+        (kube_node_info unless on(node) kube_pod_info{namespace="openshift-sdn",  pod=~"ovs.*"}) > 0
+      for: 20m
+      labels:
+        severity: warning
+    - alert: NetworkPodsCrashLooping
+      annotations:
+        message: Pod {{"{{"}} $labels.namespace{{"}}"}}/{{"{{"}} $labels.pod{{"}}"}} ({{"{{"}} $labels.container
```
squeed: ugh, that double escape. nothing else we can do, though :-/

bindata/network/openshift-sdn/alert-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  groups:
+  - name: general.rules
+    rules:
+    - alert: NodeWithoutOVSPod
+      annotations:
+        message: |
+          All nodes should be running an ovs pod. {{ $labels.node }} is not.
+      expr: |
+        (kube_node_info unless on(node) kube_pod_info{namespace="openshift-sdn",  pod=~"ovs.*"}) > 0
```
squeed: Might as well add the sdn pod to this alert too.

bindata/network/openshift-sdn/alert-rules.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: monitoring.coreos.com/v1
```
squeed: This needs the magic annotation that makes the network operator not care if creation fails.

squeed:
This is an awesome start.
/lgtm

bindata/network/openshift-sdn/alert-rules.yaml
```diff
@@ -0,0 +1,41 @@
...
+        severity: warning
+    - alert: NetworkPodsCrashLooping
+      annotations:
+        message: Pod {{"{{"}} $labels.namespace{{"}}"}}/{{"{{"}} $labels.pod{{"}}"}} ({{"{{"}} $labels.container
+          {{"}}"}}) is restarting {{"{{"}} printf "%.2f" $value {{"}}"}} times / 5 minutes.
+      expr: |
+        rate(kube_pod_container_status_restarts_total{namespace="openshift-sdn"}[15m]) * 60 * 5 > 0
+      for: 1h
+      labels:
+        severity: critical
```
brancz: I have a hard time believing that this should be paged about. There's no guarantee that this is actually having user impact. Paging/critical should be when there is proven user impact, like percentage network bytes dropped (this could be the reason), but the philosophy is that paging/critical alerts are symptom based, not cause based. Warning or info level alerts are also important but shouldn't wake up someone.

PR #191, opened 2019-06-11
Further SDN config updates

danwinship:
* https://github.com/openshift/origin/pull/23022 makes the sdn-controller ignore the passed-in config file and just use the `ClusterNetwork` instead, so now we can drop the creation of the config file. (Aside: this means we'll end up leaving a stale `openshift-sdn/sdn-controller-config` `ConfigMap` around in old clusters. Should we do something to ensure that gets deleted, so as not to confuse people noticing it in the future?)
* https://github.com/openshift/api/pull/342 added MTU to `ClusterNetwork`, so pull in that update and set the new field


bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -47,6 +47,10 @@ spec:
           type: integer
           minimum: 1
           maximum: 65535
+        mtu:
+          type: integer
+          minimum: 576
+          maximum: 65536
```
pecameron: 65536 is 50 or 100 too large.

squeed:
> Should we do something to ensure that gets deleted, so as not to confuse people noticing it in the future?

For sure. I was initially thinking we should persist created object somewhere, then compare against that and do GC at the end of reconcile.

However, I think it's reasonable to, for now, have a hard-coded list of resources that should not exist.

bindata/network/openshift-sdn/001-crd.yaml
danwinship: This is openshift-sdn specific, so it would be 50, not 100. Except also isn't it 51 too large? It seems like the max MTU should be 65535, not 65536. But anyway, "576" and "65536" match the check in `validateOpenShiftSDN()`; we accept any value here that the code will accept there.

danwinship:
/hold
until I add the code to delete the stale object

danwinship:
/hold cancel
filed https://jira.coreos.com/browse/SDN-455 about cleaning things up

PR #192, opened 2019-06-11
update kubelet label to use upstream GA label

rphillips:
ref: https://jira.coreos.com/browse/POD-114


danwinship:
yay stable features!
/lgtm

PR #195, opened 2019-06-12
Render a kube-proxy config YAML file

danwinship:
Soon the "node" part of openshift-sdn will be getting all of its configuration from the `ClusterNetwork` object and `node-config.yaml` will be used only for the kube-proxy configuration bits. This makes CNO write out a kube-proxy config YAML for openshift-sdn to use for that instead (as well as laying the groundwork for writing out kube-proxy configs for other network plugins).

In https://jira.coreos.com/browse/SDN-383 we had talked about rewriting this to be more in terms of the `KubeProxyConfiguration` type and less in terms of command-line arguments, but it turns out that the upstream code isn't really set up well for that at the moment. (See the comment on `GenerateKubeProxyConfiguration` below.) So for the moment this works in terms of translation from command-line arguments, though without depending on any of the actual kube-proxy command-line argument code, so it won't break when they finally remove support for that.


squeed:
FYI, https://docs.google.com/document/d/1q0oIXUIEnFm6dSnAtZfnsvJCJL39x8ozVE6N4iGtaLw/edit?ts=5cf7c281#heading=h.d7jh5gr66xtr is a proposal for a pretty significant refactor of the kube-proxy configuration.

danwinship:
I'd say the fact of an impending rewrite is another argument in favor of *not* exposing `KubeProxyConfiguration` directly in our own API (ie, not implementing the "some kind of patch mechanism" suggested in the jira card). Plugins (both openshift-sdn and third-party plugins that want to tie in to CNO) should expose specific options for the things they want to make configurable, and then we can deal with assembling that into a proper configuration file in the background.

pkg/network/openshift_sdn.go
```diff
@@ -54,12 +53,25 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	}
 	data.Data["NodeConfig"] = nodeCfg
 
+	kpcDefaults := map[string][]string{
+		"metrics-bind-address":    {"0.0.0.0"},
+		"metrics-port":            {"9101"},
+		"proxy-mode":              {"iptables"},
+		"iptables-masquerade-bit": {"0"},
```
squeed: Is the default we've been using? The upstream default is 14.

squeed:
Per our offline conversation, I do think using the actual type is better, even if it bloats the yaml a bit.

This looks good to me.

pkg/network/openshift_sdn.go
danwinship: Yes. I'm not sure why, but we've always been overriding it. (Note that in addition to the defaults set in CNO, there's *another* set of defaults [applied by the openshift-sdn binary](https://github.com/openshift/origin/blob/v4.1.0/pkg/cmd/openshift-sdn/proxy.go#L45) at startup. The new code here combines the two sets of defaults so that after https://github.com/openshift/origin/pull/23156 when we pass `--proxy-config`, all of the configuration (except `HostnameOverride`) will be on the CNO side and openshift-sdn won't change any of it.

squeed: Gotcha. Can't wait to get these all cleaned up.

danwinship:
Next test run should result in sdn actually using the generated kube-proxy-config.yaml rather than node-config.yaml...

PR #196, opened 2019-06-13
verify-style: run dep check

squeed:
This will catch if we leave the project in a funny state.


squeed:
Ah, looks like we need to install dep in our `release` tooling first. huh.

hack/verify-style.sh
```diff
@@ -20,6 +20,10 @@ if [[ ! $(which operator-sdk) ]]; then
   echo "operator-sdk not found on PATH. To install:"
   echo "go get -u github.com/operator-framework/operator-sdk/commands/operator-sdk"
 fi
+if [[ ! $(which dep) ]]; then
```
danwinship: while you're adding this, I'm pretty sure you can remove the check for operator-sdk just above, can't you?

hack/verify-style.sh
```diff
@@ -41,6 +45,9 @@ fi
 echo "Running go vet..."
 go vet $GOPKGS
 
+echo "Running dep check..."
+dep check
+
 # Temporariy disabled because it breaks openshift-ci
```
danwinship: we should fix this too while we're dealing with things...

hack/verify-style.sh
squeed: definitely - especially since we don't `exit 1` anyways.

hack/verify-style.sh
squeed: hah, it should just be removed, now that our types are in openshift/api

squeed:
Uh, what!? It works on my machine!?

```
Running dep check... 
# vendor is out of sync:
github.com/gogo/protobuf: hash of vendored tree not equal to digest in Gopkg.lock
```

squeed:
Wow, turns out that the .gitignore we got from Github excluded `tags`, not `/tags`. Fixed...

PR #198, opened 2019-06-13
[WIP] Cluster Network Operator Should Result in No-Op for Unknown Network Types

vareti:
OpenShift Cluster Network Operator should result in a No-Op and not do anything if `NetworkType` is specified as anything other than `OpenShiftSDN`, `OVNKubernetes`, `Kuryr` and `Raw`. This will help third party network operators like Nuage to manage network in an OpenShift Cluster.

Like OpenShift Cluster Network Operator, Nuage SDN Operator also uses `Networks.config.openshift.io` and `Networks.operator.openshift.io` objects to manage networking in an OpenShift cluster. The operator that manages the network for an OpenShift cluster will depend on the `NetworkType` value specified during installation.

### DIFF
--- a/pkg/controller/clusterconfig/clusterconfig_controller.go
+++ b/pkg/controller/clusterconfig/clusterconfig_controller.go
@@ -89,7 +89,7 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
 	}
 
 	if err := network.ValidateNetworkType(clusterConfig.Spec.NetworkType); err != nil {
-		log.Printf("Unsupported NetworkType specified: %v", err)
+		log.Printf("Failed to validate NetworkType: %v", err)
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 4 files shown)


squeed:
HI there, thanks for the PR.

I don't think we want the network operator to always short-circuit. It manages other aspects of networking, even when it's not installing a default network plugin. For example, it will (soon) install kube-proxy for third-party network providers, and it will always install Multus.

Instead, I think there are two things that need to happen.
1- You should stop using `Network.operator.openshift.io` - that is *only* for the cluster-network-operator. You should define your own type
2 - We need to make sure that our configuration object lets you select the features, such that the cluster is configured in a way for your network plugin to succeed.

If you need to disable kube-proxy or multus, you can do so already. If there's something else, then file an issue and we can get you the knobs you need.

vareti:
Hi @squeed 

Thanks for your comment.

It would be really great if you can provide the knobs for a 3rd party network provider to manage OpenShift cluster networking . Nuage provides a [CNI plugin](github.com/nuagenetworks/nuage-cni), a [control plane](github.com/nuagenetworks/nuage-kubernetes) and custom flavor of OVS to forward the traffic. It is still expected that kube-proxy to be already running though. I could not find a way of doing this without CNO and Nuage SDN operator stepping on each other. I will file an issue for this feature enhancement. 

I will be glad to help in making any changes to support this feature. Please let me now.

PR #199, opened 2019-06-14
Add namespaces managed by CNO to related objects field

rcarrillocruz:
This allows must-gather tool to give information about CNO objects
for free.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -17,6 +17,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -50,10 +51,10 @@ func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager) *Re
 	configv1.Install(mgr.GetScheme())
 	operv1.Install(mgr.GetScheme())
 	return &ReconcileOperConfig{
-		client: mgr.GetClient(),
-		scheme: mgr.GetScheme(),
-		status: status,
-
+		client:        mgr.GetClient(),
+		scheme:        mgr.GetScheme(),
+		status:        status,
+		mapper:        mgr.GetRESTMapper(),
 		podReconciler: newPodReconciler(status),
 	}
 }
@@ -95,10 +96,10 @@ var _ reconcile.Reconciler = &ReconcileOperConfig{}
 type ReconcileOperConfig struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
-	status *statusmanager.StatusManager
-
+	client        client.Client
+	scheme        *runtime.Scheme
+	status        *statusmanager.StatusManager
+	mapper        meta.RESTMapper
 	podReconciler *ReconcilePods
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -58,6 +58,22 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...confi
 
 	oldStatus := co.Status.DeepCopy()
 
+	related := []configv1.ObjectReference{
+		{
+			Resource: "namespaces",
+			Name:     "openshift-sdn",
```
danwinship: (Aside: is "namespace**s**" correct? I would have expected it to want the singular.)

This should be computed dynamically based on the actual Namespaces we created. The same way that ReconcileOperConfig figures out the set of active DaemonSets and Deployments and then tells the StatusManager about them, it should do the same with Namespaces. (Except it should add the controller's namespace to the list too.)

rcarrillocruz: Plural seems fine, I found it in other spots of the code:

https://github.com/openshift/machine-api-operator/blob/6066765d8ff090fd081c7a20e59955d00004b0ad/pkg/operator/status.go#L165
https://github.com/openshift/cluster-samples-operator/blob/df3dbf4d8a119058f80067eca568c6f78ff8d725/pkg/operatorstatus/operatorstatus.go#L169
https://github.com/openshift/cluster-autoscaler-operator/blob/62768a6ba48010d12c4d86b8b06689a6970a7841/pkg/operator/operator.go#L110

squeed: Plural is correct, because this is a Resource, not a Kind. However, dan is right, this should be computed automatically.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -171,6 +155,17 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName)
...
 
+func (status *StatusManager) SetRelatedObjects(objects []configv1.ObjectReference) {
+	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
+	err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
+	isNotFound := errors.IsNotFound(err)
+	if err != nil && !isNotFound {
+		log.Printf("Failed to get ClusterOperator %q: %v", status.name, err)
+		return
+	}
+	co.Status.RelatedObjects = objects
```
danwinship: This is just modifying the copy that you just fetched, which will be GC'ed as soon as this function returns.

You can just store the objects in a field on `status` here, and then set them on the object in `Set`

danwinship:
@rcarrillocruz good except now you're never copying `status.relatedObjects` over to `co.Status.RelatedObjects`.

Also, please squash your commits rather than doing fixups. It's generally easier to deal with if the PR is always in a ready-to-be-merged form.

danwinship:
Doh. Sorry, I guess I wasn't clear. `SetRelatedObjects` should just set a field on the `StatusManager` object, like in [the previous version](https://github.com/openshift/cluster-network-operator/blob/6d0e4bb5007372881b6c7ba2927f6a0ec36396bd/pkg/controller/statusmanager/status_manager.go#L159), and then `Set` should copy that into the object being updated, [like in the original version](https://github.com/openshift/cluster-network-operator/blob/70dfa26d90047fd4ede5541f0f82404325682b18/pkg/controller/statusmanager/status_manager.go#L75) except with the dynamically-generated list (`co.Status.RelatedObjects = status.relatedObjects`)

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -208,15 +208,23 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+	relatedObjects := []configv1.ObjectReference{}
 	for _, obj := range objs {
 		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
 			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
 			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		}
+		relatedObjects = append(relatedObjects, configv1.ObjectReference{
+			Group:     obj.GetObjectKind().GroupVersionKind().Group,
+			Resource:  obj.GetKind(),
```
squeed: This isn't right - the object kind is "Namespace", whereas the API resource is "namespaces". You need to actually look up the conversion between the Kind and the Resource using the the RESTMapper.

It's a bit awkward, yes. It looks like a few other operators do it. See if they hard-code the mapping or do something cleverer.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -205,18 +207,38 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 	objs = append([]*uns.Unstructured{app}, objs...)
 
+	config, err := rest.InClusterConfig()
```
squeed: No need to create your own RestMapper - just get it from the manager.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -205,18 +207,38 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+	relatedObjects := []configv1.ObjectReference{}
 	for _, obj := range objs {
 		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
 			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
 			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		}
+		restMapping, err := mapper.RESTMapping(obj.GroupVersionKind().GroupKind())
+		if err != nil {
+			log.Printf("Failed to get REST mapping:%v", err)
```
squeed: needs a `continue` here, and maybe a bit more useful error message.

squeed:
Two small comments, then this should be good to go.

squeed:
/approve

This adds all objects to the Related field. I think that's fine - thoughts, @danwinship?

danwinship:
> This adds all objects to the Related field. I think that's fine - thoughts, @danwinship?

That was the plan, wasn't it? Then we can use that to figure out what we need to clean up?

The API documentation is not detailed enough to either confirm or deny that this is OK.

PR #201, opened 2019-06-14
Install standalone kube-proxy.

squeed:
This installs a standalone kube-proxy when needed. This is whenever we're not installing a network that is known to include it itself. This means that kube-proxy will be installed for third-party network plugins.

### DIFF
--- a/bindata/kube-proxy/kube-proxy.yaml
+++ b/bindata/kube-proxy/kube-proxy.yaml
@@ -1,0 +1,91 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: openshift-kube-proxy
+  name: proxy-config
+data:
+  kube-proxy-config.yaml: |-
+{{.KubeProxyConfig | indent 4}}
+
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: openshift-kube-proxy
+  namespace: openshift-kube-proxy
+  annotations:
+    kubernetes.io/description: |
+      This daemonset is the kubernetes service proxy (kube-proxy).
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: kube-proxy
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: kube-proxy
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      serviceAccountName: openshift-kube-proxy
+      hostNetwork: true
+      priorityClassName: system-node-critical
+      containers:
+      - name: kube-proxy
+        image: {{.KubeProxyImage}}
+        command:
+        - /bin/bash
+        - -c
+        - exec /usr/bin/hyperkube kube-proxy --config=/config/kube-proxy-config.yaml --hostname-override "${K8S_NODE_NAME}"
+        securityContext:
+          privileged: true
+        env:
+        # Tell kube-proxy to talk to the apiserver directly.
+        - name: KUBERNETES_SERVICE_PORT
+          value: "{{.KUBERNETES_SERVICE_PORT}}"
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{.KUBERNETES_SERVICE_HOST}}"
+        - name: K8S_NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        volumeMounts:
+        # Because we don't actually run iptables locally, but chroot in to the host
+        - mountPath: /host
+          name: host-slash
+          readOnly: true
+        - mountPath: /config
+          name: config
+          readOnly: true
+        terminationMessagePolicy: FallbackToLogsOnError
+        ports:
+        - name: healthz
+          containerPort: 10256
+        - name: metrics
+          containerPort: 9101
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: healthz
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: healthz
+          initialDelaySeconds: 5
+          periodSeconds: 5
+      restartPolicy: Always
+      tolerations:
+      - operator: Exists
+      nodeSelector:
+        kubernetes.io/os: linux
+      volumes:
+      - name: host-slash
+        hostPath:
+          path: /
+      - name: config
+        configMap:
+          name: proxy-config
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -54,6 +54,7 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	kpcDefaults := map[string][]string{
 		"metrics-bind-address":    {"0.0.0.0"},
 		"metrics-port":            {"9101"},
+		"healthz-port":            {"10256"},
 		"proxy-mode":              {"iptables"},
 		"iptables-masquerade-bit": {"0"},
 	}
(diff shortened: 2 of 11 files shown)


squeed:
TODO: merge kube-proxy image fixes in https://github.com/openshift/origin/pull/23163, https://github.com/openshift/release/pull/4064.

TODO: figure out the best way to delete kube-proxy if DeployKubeProxy is changed.

bindata/kube-proxy/kube-proxy.yaml
```diff
@@ -0,0 +1,71 @@
...
+        # Because we don't actually run iptables locally, but chroot in to the host
+        - mountPath: /host
+          name: host-slash
+          readOnly: true
+        terminationMessagePolicy: FallbackToLogsOnError
+        ports:
+        - name: healthz
+          containerPort: 10256
+        - name: metrics
+          containerPort: 9101
```
danwinship: If we allow these to be overridden then these should be template fields (and likewise in other places).
If we don't allow these to be overridden, then the code should be updated accordingly.

pkg/network/kube-proxy.go
```diff
@@ -0,0 +1,175 @@
...
+	if conf.KubeProxyConfig.ProxyArguments == nil {
+		conf.KubeProxyConfig.ProxyArguments = map[string][]string{}
+	}
+
+	if conf.KubeProxyConfig.ProxyArguments["metrics-bind-address"] == nil {
+		conf.KubeProxyConfig.ProxyArguments["metrics-bind-address"] = []string{"0.0.0.0"}
+	}
+
+	if conf.KubeProxyConfig.ProxyArguments["metrics-port"] == nil {
+		conf.KubeProxyConfig.ProxyArguments["metrics-port"] = []string{"9101"}
```
danwinship: Note that this isn't the upstream default. I'm not sure why we override it in openshift-sdn.

pkg/network/kube-proxy.go
```diff
@@ -0,0 +1,175 @@
...
+
+			existingArgs = append(existingArgs, args...)
+			err = uns.SetNestedStringSlice(container, existingArgs, "command")
+			if err != nil {
+				return nil, errors.Wrapf(err, "cannot inject kube-proxy args")
+			}
+
+			containers[0] = container
+
+			err = uns.SetNestedSlice(manifest.Object, containers, "spec", "template", "spec", "containers")
```
danwinship: it wouldn't be easier to serialize the args to proper yaml and then template them in?

bindata/kube-proxy/kube-proxy.yaml
squeed: good point.

pkg/network/kube-proxy.go
squeed: It's tough to control exactly how yaml will serialize stuff. I wasn't comfortable it so that the templated yaml could always safely be inserted.

This was ugly to figure out but is otherwise correct.

pkg/network/kube-proxy.go
squeed: Because this is accessed remotely (from Prometheus). So, it needs to be in the open port range (9000-9999)

squeed: Because this is accessed remotely (from Prometheus). So, it needs to be in the open port range (9000-9999)

squeed: Because this is accessed remotely (from Prometheus). So, it needs to be in the open port range (9000-9999)

danwinship:
> TODO: figure out the best way to delete kube-proxy if DeployKubeProxy is changed.

https://jira.coreos.com/browse/SDN-455 "Track/clean-up CNO-created resources" (which I'd created re: `controller-config.yaml`) should cover this: if we record that we have created the kube proxy DaemonSet when we create it, then when you change `DeployKubeProxy` we'll see that the DaemonSet is no longer in the set of objects-to-reconcile, so we can delete it.

squeed:
OK, rebased on your kube-proxy config changes. PTAL.

/hold cancel

pkg/network/proxy.go
```diff
@@ -1,15 +1,36 @@
...
+// openshift-sdn and OVN-Kubernetes deploy their own kube-proxy. All other
+// network providers are assumed to require kube-proxy
+func ShouldDeployKubeProxy(conf *operv1.NetworkSpec) bool {
+	switch conf.DefaultNetwork.Type {
+	case operv1.NetworkTypeOpenShiftSDN:
+		return false
+	case operv1.NetworkTypeOVNKubernetes:
+		return false
+	default:
+		log.Printf("NOTICE: Unknown network type %s, installing kube-proxy", conf.DefaultNetwork.Type)
```
danwinship: this will end up being logged repeatedly, since this function is called from multiple places

pkg/network/proxy.go
```diff
@@ -51,3 +82,65 @@ func validateKubeProxy(conf *operv1.NetworkSpec) []error {
...
+		"metrics-port":         {"9101"},
+		"proxy-mode":           {"iptables"},
+	}
+
+	kpc, err := kubeProxyConfiguration(conf, kpcDefaults)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to generate kube-proxy configuration file")
+	}
+
+	// render the manifests on disk
```
danwinship: this comment keeps getting copied into various files and is wrong everywhere

bindata/kube-proxy/kube-proxy.yaml
danwinship: This is still not addressed

pkg/network/proxy.go
squeed: it's not wrong, it's just... misleading. which makes it wrong.

pkg/network/proxy.go
squeed: yeah, it's probably a bit too verbose. removed.

bindata/kube-proxy/kube-proxy.yaml
squeed: Also a good point. fixed.

squeed:
OK, updated. Brought a bit of sanity to openshift-sdn's treat of kube-proxy args, which was very silly indeed.

pkg/network/openshift_sdn.go
```diff
@@ -141,14 +142,6 @@ func fillOpenShiftSDNDefaults(conf, previous *operv1.NetworkSpec, hostMTU int) {
 		conf.KubeProxyConfig.ProxyArguments = map[string][]string{}
 	}
 
-	if conf.KubeProxyConfig.ProxyArguments["metrics-bind-address"] == nil {
-		conf.KubeProxyConfig.ProxyArguments["metrics-bind-address"] = []string{"0.0.0.0"}
```
danwinship: Double-checking: because this used to be in the defaults, we previously would have serialized a last-applied-configuration that contained these values explicitly. When the user upgrades to 4.2, we will no longer be adding those values to the conf, and so CNO will perceive their configuration as having changed. However, this won't matter because we allow all kube-proxy config changes, so we're good.

Right?

pkg/network/proxy.go
```diff
@@ -49,5 +78,76 @@ func validateKubeProxy(conf *operv1.NetworkSpec) []error {
 		}
 	}
 
+	// Don't allow ports to be overridden
+	if _, ok := p.ProxyArguments["metrics-port"]; ok {
+		out = append(out, errors.Errorf("kube-proxy --metrics-port cannot be overridden"))
```
danwinship: should allow it to be "overridden" to the default value, right?

pkg/network/openshift_sdn.go
squeed: That was my thinking. I actually have a branch with some tests to make sure we don't violate this for a theoretical existing cluster.

pkg/network/proxy.go
squeed: Do you think that's necessary? I'm not against it, but it's not obviously needed to me.

squeed:
test failure *seems* like a flake.

/retest

pkg/network/proxy.go
danwinship: I'm worried some existing cluster might be doing that. That might be paranoid...

squeed: If they're doing that, then they're going to see many, many failures in the system Prometheus, since it will be failing to scrape. So I'm not concerned.

danwinship: sorry, when I said "default", I meant *our* default, not the default default. In other words, if someone is explicitly setting it to 9101, we should allow that

squeed: *shrug* ok.

squeed:
updated, plta.

PR #203, opened 2019-06-15
Slightly improve Progressing statuses

danwinship:
Before, during an upgrade, the status would go:

    DaemonSet %q update is rolling out (0 out of 3 updated)
    ...
    DaemonSet %q is not available (awaiting 1 nodes)
    ...
    DaemonSet %q update is rolling out (1 out of 3 updated)
    ...
    DaemonSet %q is not available (awaiting 1 nodes)
    ...
    DaemonSet %q update is rolling out (2 out of 3 updated)
    ...
    DaemonSet %q is not available (awaiting 1 nodes)
    ...

Fix it to make the "update is rolling out" status take precedence over
the "not available" status.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -174,12 +174,12 @@ func (status *StatusManager) SetFromPods() {
 			continue
 		}
 
-		if ds.Status.NumberUnavailable > 0 {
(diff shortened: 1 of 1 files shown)


squeed:
this always bugged me too. I'd filed SDN-396 to fix it.

/lgtm

PR #205, opened 2019-06-19
Actually pass --proxy-config to openshift-sdn

danwinship:
https://github.com/openshift/origin/pull/23205 is failing because while we generate a proxy config now, we forgot to actually tell openshift-sdn to use it...

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -4,8 +4,6 @@ metadata:
   namespace: openshift-sdn
   name: sdn-config
 data:
-  sdn-config.yaml: |-
-{{.NodeConfig | indent 4}}
   kube-proxy-config.yaml: |-
 {{.KubeProxyConfig | indent 4}}
 ---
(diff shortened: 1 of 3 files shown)


danwinship:
/hold
i want to examine the test run artifacts before merging

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -83,7 +83,7 @@ spec:
           cp -f /opt/cni/bin/* /host/opt/cni/bin/
 
           # Launch the network process
-          exec /usr/bin/openshift-sdn --config=/config/sdn-config.yaml  --url-only-kubeconfig=/etc/kubernetes/kubeconfig --v=${DEBUG_LOGLEVEL:-2}
+          exec /usr/bin/openshift-sdn --config=/config/sdn-config.yaml --proxy-config=/config/kube-proxy-config.yaml --url-only-kubeconfig=/etc/kubernetes/kubeconfig --v=${DEBUG_LOGLEVEL:-2}
```
squeed: don't you need to remove `--config` too?

danwinship: Sigh, too much back-and-forth. Was thinking I couldn't remove that because openshift-sdn still required it, but of course it doesn't at this point.

squeed:
bah, looks like the type wasn't installed!? 

`error reading config: no kind "KubeProxyConfiguration" is registered for version "kubeproxy.config.k8s.io/v1alpha1" in scheme "github.com/openshift/library-go/pkg/config/helpers/readresource.go:74"`

danwinship:
Meh. openshift-sdn is trying to read it in directly as a `k8s.io/kubernetes/pkg/proxy/apis/config.KubeProxyConfiguration`, but it needs to read it in as a `k8s.io/kube-proxy/config/v1alpha1.KubeProxyConfiguration` and then convert it.

danwinship:
that was still using the old openshift-sdn for some reason...

PR #206, opened 2019-06-20
remove all sriov bits from CNO

zshi-redhat:
SR-IOV components will be deployed and configured
via a dedicated operator 'SR-IOV Network Operator'.
This commit removes all SR-IOV related config and
deployment files.


squeed:
Do we think this is being used? What is the level of support of sr-iov? If we remove this, than anyone using this will suffer disruption.

If this is released at a tech-preview level, then that's OK. If not, then we need to determine how to migrate existing customers (cc @fepan @dougbtv)

zshi-redhat:
> Do we think this is being used? What is the level of support of sr-iov? If we remove this, than anyone using this will suffer disruption.
> 
> If this is released at a tech-preview level, then that's OK. If not, then we need to determine how to migrate existing customers (cc @fepan @dougbtv)

SR-IOV support is Tech Preview in 4.1 and will still be Tech Preview in 4.2.

phoracek:
@zshi-redhat is the SR-IOV operator already available somewhere?

fepan:
@phoracek yes, it's at https://github.com/openshift/sriov-network-operator

phoracek:
Thanks @fepan. Would it be possible to run it on vanilla Kubernetes or is it coupled with openshift/api?

fepan:
I believe it is tied to Openshift currently, at least we are using OpenShift's drain node utility. I'm not sure if there are other hard dependencies on OpenShift at this point, @pliurh would be the better person to answer this.

pliurh:
@phoracek  Currently, although I'm only developing on a Openshift cluster, I think the sriov-network-operator can work with vanilla Kubernetes as well. Since it doesn't require any openshift API. The drain node function is from a Openshift lib, but underneath it invokes the kubernetes API. And also Mellanox team which are working on a plugin of the sriov-network-operator, told me that it can work with vanilla Kubernetes in their environment.

phoracek:
That sounds great @pliurh, thanks!

PR #207, opened 2019-06-20
hack/run-locally.sh: don't open AWS ports for libvirt clusters

dcbw:
@danwinship @pecameron @JacobTanenbaum @rcarrillocruz @squeed

### DIFF
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -122,6 +122,8 @@ function wait_for_cluster() {
         echo ""
         echo "Ready to start operator. (If this fails, try running '$0' again.)"
     fi
+
+    detect_platform
 }
 
 # Copy an existing install-config into CLUSTER_DIR, or create a new one there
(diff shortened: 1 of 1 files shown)


hack/run-locally.sh
```diff
@@ -152,9 +154,20 @@ function create_cluster() {
...
+    else
+        echo "could not detect install platform from terraform.tfstate" >&2
+        exit 1
+    fi
+}
+
 # Fix up the AWS firewall so network plugin components can talk to each other
 function open_aws_ports() {
-    if [[ "${NETWORK_PLUGIN}" == "OVNKubernetes" ]]; then
+    if [[ "${NETWORK_PLUGIN}" == "OVNKubernetes" && "${PLATFORM}" == "aws" ]]; then
```
danwinship: I was assuming that if multi-platform run-locally.sh support arrived before open-ovn-ports.sh went away that we'd add support for other platforms to open-ovn-ports.sh. Is there no firewalling in libvirt?

hack/run-locally.sh
```diff
@@ -249,6 +263,7 @@ if [[ -z "$(which oc 2> /dev/null || exit 0)" ]]; then
     exit 1
 fi
 
+PLATFORM="aws"
```
danwinship: hm?

hack/run-locally.sh
```diff
@@ -152,9 +154,20 @@ function create_cluster() {
     trap "trap - TERM && kill -- -$$" INT TERM EXIT
 }
 
+function detect_platform() {
```
danwinship: fwiw if you're only doing this when the cluster is up, you can do `oc get infrastructure cluster -o jsonpath='{.status.platform}'`

hack/run-locally.sh
dcbw: @danwinship there's whatever firewalling RHCOS does, which appears to be literally nothing. Like, all it has are nat+filter tables with default chain policies of ACCEPT. So I guess if that ever changes or it ever starts running firewalld we'd need to update this.

dcbw:
Closing in favor of https://github.com/openshift/cluster-network-operator/pull/216

PR #208, opened 2019-06-20
Adds RBAC aggregation for the net-attach-def

dougbtv:
Aggregates the viewing and editing for net-attach-defs to users with the view & edit roles.

reported in this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1721444

### DIFF
--- a/bindata/network/multus/002-rbac.yaml
+++ b/bindata/network/multus/002-rbac.yaml
@@ -96,3 +96,17 @@ rules:
   - update
   - patch
   - delete
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: multus:view
+rules:
+- apiGroups: ["k8s.cni.cncf.io"]
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - watch
--- a/bindata/network/additional-networks/crd/002-rbac.yaml
+++ b/bindata/network/additional-networks/crd/002-rbac.yaml
@@ -1,0 +1,15 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: multus:view
+rules:
+- apiGroups: ["k8s.cni.cncf.io"]
+  resources: ["*"]
+  verbs:
+  - get
+  - list
+  - watch
+
--- a/pkg/network/multus_test.go
+++ b/pkg/network/multus_test.go
@@ -50,10 +50,11 @@ func TestRenderMultus(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-multus", "multus")))
 
 	// It's important that the namespace is first
-	g.Expect(len(objs)).To(Equal(10))
+	g.Expect(len(objs)).To(Equal(11))
 	g.Expect(objs[0]).To(HaveKubernetesID("CustomResourceDefinition", "", "network-attachment-definitions.k8s.cni.cncf.io"))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("Namespace", "", "openshift-multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRole", "", "multus")))
+	g.Expect(objs).To(ContainElement(HaveKubernetesID("Role", "", "multus:view")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ServiceAccount", "openshift-multus", "multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-multus", "multus")))


bindata/network/multus/002-rbac.yaml
```diff
@@ -63,3 +63,48 @@ subjects:
...
+kind: ClusterRole
+metadata:
+  labels: 
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: multus:view
+rules:
+- apiGroups: ["apiextensions.k8s.io"]
+  resources:
+  - customresourcedefinitions
+  - customresourcedefinitions/status
```
danwinship: A multus-specific policy should not be giving people permission to view all CustomResourceDefinition objects. If that's actually needed, it should be added to the base role bindings in origin.
(Also, including the `/status` subresource explicitly should not be needed if you're not specifying different permissions from the parent resource.)

But I think this isn't needed anyway; you should be able to give people permission to view net-attach-defs without them also needing permission to view CRDs. I think.

bindata/network/multus/002-rbac.yaml
```diff
@@ -63,3 +63,48 @@ subjects:
...
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels: 
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: multus:edit
```
danwinship: This authorization is wrong. We only want cluster administrators to be able to create CRDs and network-attachment-definitions, and they already have that permission.

bindata/network/multus/002-rbac.yaml
```diff
@@ -63,3 +63,48 @@ subjects:
...
+- apiGroups: ["apiextensions.k8s.io"]
+  resources:
+  - customresourcedefinitions
+  - customresourcedefinitions/status
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: ["k8s.cni.cncf.io"]
+  resources: ["*"]
```
danwinship: I don't remember all the details of how roles and bindings work, but we need to make sure that this doesn't give project admins the ability to view net-attach-defs in *other* namespaces.

squeed:
Now that I sit and think for a bit, I feel like ordinary users shouldn't be able to edit or create NetworkAttachmentDefinitions. It is a pretty wide security hole.

dougbtv:
Thanks for the input guys, good catches. I've got a new commit pushed up that limits to aggregating to only view permission (that is, it removes the aggregation to the edit permission). Additionally, it removes the overall access to custom resource definitions.

dougbtv:
Also fixed up the unit tests to look specifically for the multus:view ClusterRole and also updated the count of objects therein. I made these as incremental commits, please lmk if you'd like me to squash them and rebase, too.

squeed:
@dougbtv can you squash? Then I'll lgtm.

dougbtv:
Sure thing / No prob @squeed -- have a new-and-rebased single commit up

bindata/network/multus/002-rbac.yaml
squeed: I needs to be a Role, not a ClusterRole.

bindata/network/multus/002-rbac.yaml
```diff
@@ -63,3 +63,17 @@ subjects:
 - kind: ServiceAccount
   name: multus
   namespace: openshift-multus
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
squeed: As mentioned below, I believe this needs to be a Role.

dougbtv:
Thanks guys for the attention to detail on the Role vs. ClusterRole, edited per input.

bindata/network/multus/002-rbac.yaml
```diff
@@ -52,14 +52,28 @@ metadata:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
```
squeed: er, I'm confused. This is the role assigned to the service account used by the multus binary itself, right? Then I doubt you can change this. I believe that adding the new Role below you've done what you need.

dougbtv: Nice catch, this indeed should not have been changed, you're right -- all it needs is the additional Role.

dougbtv:
@squeed -- thanks for the poke, new rebased commit up with the changes to not alter the ClusterRole as directly used by Multus.

squeed:
@dougbtv all the retests in the world aren't going to work:
```
                "conditions": [
                    {
                        "lastTransitionTime": "2019-08-01T21:12:21Z",
                        "message": "Error while updating operator configuration: could not apply (rbac.authorization.k8s.io/v1, Kind=Role) /multus:view: could not create (rbac.authorization.k8s.io/v1, Kind=Role) /multus:view: the server does not allow this method on the requested resource",
                        "reason": "ApplyOperatorConfig",
                        "status": "True",
                        "type": "Degraded"
                    },
```

squeed:
@dougbtv you need to fix the rbac issues. The problem is that you're crating the Role before you're creating the CRD, I think.

dougbtv:
@squeed -- thanks for the heads up there, I'm... not sure how I missed that log entry. Thanks for picking that up for me, and for the tip, appreciate it.

dougbtv:
Pushed a rebase (let's see if the permission are any different for the CNO, now?)

PR #210, opened 2019-06-24
Make Kuryr pods labels unambiguous

dulek:
Commit changes the labels of Kuryr pods to make sure we distinguish
kuryr-cni and kuryr-controller. Also selector field is added as it seems
to be required.


squeed:
Oops.
/lgtm
/approve

PR #212, opened 2019-06-25
sdn-ovs: allow ovs daemons to run as non-root

orgcandman:
The two chowns are to set the ownership from the host mounted
volumes, since they may be incorrectly mapped.

sdn-33
ref: https://jira.coreos.com/browse/SDN-33

Signed-off-by: Aaron Conole <aconole@redhat.com>

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -35,6 +35,8 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
+          chown -R openvswitch:openvswitch /var/run/openvswitch
+          chown -R openvswitch:openvswitch /etc/openvswitch
 
           # if another process is listening on the cni-server socket, wait until it exits
           trap 'kill $(jobs -p); exit 0' TERM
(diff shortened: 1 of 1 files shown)


orgcandman:
@squeed PTAL

squeed:
Wow, that works!? Cooll! I always assumed you needed to be root to actually talk to the kernel vswtch.

danwinship:
So does this make OVS actually start up as a non-root user or does it still start up as root but then setuid itself? If the former, then is it possible to make the OVS pod non-`privileged`? If we need to keep `privileged` for some of the initial setup steps (like the `chown`) then is it possible to make it use a privileged `initContainer` but an unprivileged main container?

orgcandman:
Dan Winship <notifications@github.com> writes:

> So does this make OVS actually start up as a non-root user or does it
> still start up as root but then setuid itself?

It starts as root and downgrades.  It still needs to preserve some
capabilities.

> If the former, then is it possible to make the OVS pod non-privileged? If we need to keep privileged for some of
> the initial setup steps (like the chown) then is it possible to make it use a privileged initContainer but an
> unprivileged main container?

I think it's possible if the unprivileged container can preserve some of
the capabilities bits.

danwinship:
> I think it's possible if the unprivileged container can preserve some of the capabilities bits.

Nope, it doesn't work that way. They'd be totally separate containers.

PR #214, opened 2019-06-26
Add EnableUnidling flag to openshift-sdn

squeed:
This PR has 2 changes:
1. Improved handling of the Applied configuration for upgrades
  - Call ApplyDefaults on the previously applied configuration
  - A test case that makes sure we can adopt 4.1 clusters
2. The EnableUnidling flag, and its corresponding API change

If this all looks good, I'll split off an api PR.


danwinship:
> Call ApplyDefaults on the previously applied configuration

So how does that help?

Also, there's a typo in the commit message: "preivously"

danwinship:
> So how does that help?

Ah, duh, never mind, it helps by filling in the default value of `EnableUnidling`.

pkg/network/openshift_sdn.go
```diff
@@ -48,9 +48,13 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 		"metrics-bind-address":    {"0.0.0.0"},
 		"metrics-port":            {"9101"},
 		"healthz-port":            {"10256"},
-		"proxy-mode":              {"iptables"},
+		"proxy-mode":              {"unidling+iptables"},
 		"iptables-masquerade-bit": {"0"},
 	}
+	if *c.EnableUnidling == false {
```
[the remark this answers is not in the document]
squeed: somehow that seemed hard to read to me.

pkg/network/openshift_sdn.go
```diff
@@ -100,16 +113,33 @@ func validateOpenShiftSDN(conf *operv1.NetworkSpec) []error {
 	return out
 }
 
-// isOpenShiftSDNChangeSafe currently returns an error if any changes are made.
+// isOpenShiftSDNChangeSafe allows only EnableUnidling changes.
 // In the future, we may support rolling out MTU or external openvswitch alterations.
```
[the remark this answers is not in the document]
squeed: Which was intentional, but the comment needs to be updated.

vendor/github.com/openshift/api/operator/v1/types_network.go
```diff
@@ -139,6 +139,11 @@ type OpenShiftSDNConfig struct {
 	// it will be provided separately. If set, you must provide it yourself.
 	// +optional
 	UseExternalOpenvswitch *bool `json:"useExternalOpenvswitch,omitempty"`
+
+	// EnableUnidling controls whether or not the hybrid unidling proxy will be set up
+	// by default, unidling is enabled. However, it can cause iptables lock contention
+	// issues on large clusters when many services are unidled.
```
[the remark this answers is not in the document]
squeed: Good point, updated.

pkg/network/openshift_sdn.go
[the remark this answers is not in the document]
danwinship: "`// In the future, we may support rolling out MTU or external openvswitch alterations`"
(ie, you need to fix one more place)

squeed:
API changes merged, rebased, revendored, etc. etc. @danwinship, PTAL

pkg/network/openshift_sdn.go
```diff
@@ -48,9 +48,13 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 		"metrics-bind-address":    {"0.0.0.0"},
 		"metrics-port":            {"9101"},
 		"healthz-port":            {"10256"},
-		"proxy-mode":              {"iptables"},
+		"proxy-mode":              {"unidling+iptables"},
```
danwinship: This is wrong: we have to apply the value of `c.EnableUnidling` after merging `conf` and `kpcDefaults`. If a 4.1 user redundantly specified `proxy-mode: iptables`, then in 4.1 they would have gotten unidling, but the way this code is currently written here, in 4.2 they would get no unidling. (There should be a test case for this.)

(The simplest way to do this might be to pass `c.EnableUnidling` to `kubeProxyConfiguration()`.)

squeed: Argh, good point. The "unidling+" is an implementation detail.

squeed:
OK, fixed. Thanks a ton for the review. Solved the proxy-mode thing a slightly different way. PTAL.

pkg/network/openshift_sdn.go
```diff
@@ -48,10 +49,28 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
...
+	if !*c.EnableUnidling {
+		kpcDefaults["proxy-mode"][0] = "iptables"
+	}
+
+	// If the user specified a proxy-mode, append "unidling+" to it, since
+	// that's an internal detail that was added in 4.2.
+	kpcOverrides := map[string]operv1.ProxyArgumentList{}
+	if *c.EnableUnidling && len(conf.KubeProxyConfig.ProxyArguments["proxy-mode"]) > 0 {
+		pma := conf.KubeProxyConfig.ProxyArguments["proxy-mode"]
+		if !strings.HasPrefix(pma[0], "unidling+") {
```
danwinship: FWIW, validation has already confirmed that if `*c.EnableUnidling` is set (after defaulting), then `conf.KubeProxyConfig.ProxyArguments["proxy-mode"]` is either unset or `["iptables"]`. So you can simplify this to just:

    kpcOverrides := map[string]operv1.ProxyArgumentList{}
    if *c.EnableUnidling {
            kpcOverrides["proxy-mode"] = operv1.ProxyArgumentList{"unidling+iptables"}`
    }

pkg/network/openshift_sdn.go
```diff
@@ -48,10 +49,28 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
...
+			// duplicate so we don't mutate conf
+			kpcOverrides["proxy-mode"] = make(operv1.ProxyArgumentList, len(pma))
+			copy(kpcOverrides["proxy-mode"], pma)
+			kpcOverrides["proxy-mode"][0] = "unidling+" + kpcOverrides["proxy-mode"][0]
+		}
+
+	}
+
+	kpc, err := kubeProxyConfiguration(conf, kpcDefaults, kpcOverrides)
```
danwinship: maybe rearrange to `defaults, config, overrides` to match the order they're merged in? or not

pkg/network/openshift_sdn.go
squeed: Ah yeah, good point.

pkg/network/openshift_sdn.go
squeed: yeah, thought about that. might as well be nice to future maintainers.

squeed:
Updated, PTAL.

PR #216, opened 2019-06-26
hack/run-locally.sh: misc fixes (AWS ovn-kube ports, help text, etc)

dcbw:
1) only open AWS ports when using AWS
2) print help text on errors rather than referring to HACKING.md
3) back up the install-config.yaml so it can be cached by the user

@danwinship @squeed @pecameron @JacobTanenbaum @rcarrillocruz

### DIFF
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -135,6 +135,9 @@ function ensure_install_config() {
     else
         cp "${INSTALL_CONFIG}" "${CLUSTER_DIR}/install-config.yaml"
     fi
+    # since the install-config is consumed by the installer, back it up for
+    # subsequent re-use
+    cp "${CLUSTER_DIR}/install-config.yaml" "${CLUSTER_DIR}/install-config.yaml.bak.$(date '+%s')"
 }
 
 # Create installer manifests and substitute the requested network plugin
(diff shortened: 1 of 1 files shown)


hack/run-locally.sh
```diff
@@ -135,6 +135,9 @@ function ensure_install_config() {
     else
         cp "${INSTALL_CONFIG}" "${CLUSTER_DIR}/install-config.yaml"
     fi
+    # since the install-config is consumed by the installer, back it up for
+    # subsequent re-use
+    cp "${CLUSTER_DIR}/install-config.yaml" "${CLUSTER_DIR}/install-config.yaml.bak"
```
pecameron: you might want to put a sequence number to save more than the last config.

hack/run-locally.sh
```diff
@@ -196,6 +200,7 @@ CLUSTER_DIR="${CLUSTER_DIR:-}"
 INSTALLER_PATH="${INSTALLER_PATH:-}"
 INSTALL_CONFIG="${INSTALL_CONFIG:-}"
 WAIT_FOR_MANIFEST_UPDATES="${WAIT_FOR_MANIFEST_UPDATES:-}"
+PLATFORM=
```
danwinship: cruft

dcbw:
@danwinship @pecameron PTAL thanks

hack/run-locally.sh
```diff
@@ -180,6 +191,7 @@ The following options are always accepted but only used for new clusters:
  -f               the path to an openshift-install-created install-config.yaml file; if not given one will be created
  -i               path to the openshift-install binary; if not given PATH will be searched
  -n [net plugin]  the name of the network plugin to deploy; one of [sdn|OpenShiftSDN|ovn|OVNKubernetes]
+ -s [secret]      additional pull secret to add to install-config.yaml
```
danwinship: What's the use case here? Wouldn't it be better to just add that pull secret to your template install-config permanently?

dcbw:
@danwinship removed the "-s" bits and updated the Wiki instructions to combine the secrets outside the installer and just paste them in. PTAL thanks.

PR #217, opened 2019-06-28
OVN: use config file via ConfigMap rather than environment variables

pecameron:
Create a new daemonset version that passes the ovn-config configMap
into the container as a file. Modify master and node daemonsets.

Create a ovnkube config file using a configMap.

This must merge after https://github.com/ovn-org/ovn-kubernetes/pull/748
which sets up daemonset version 4

SDN-456
https://jira.coreos.com/browse/SDN-456

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
@@ -1,0 +1,25 @@
+---
+# The config file used by ovnkube is passed as a configmap.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ovnkube-config
+  namespace: openshift-ovn-kubernetes
+data:
+  ovn_k8s.conf:   |-
+    [default]
+    mtu="{{.MTU}}"
+    cluster-subnets="{{.K8S_cluster_subnets}}"
+
+    [kubernetes]
+    service-cidr="{{.OVN_service_cidr}}"
+    ovn-config-namespace="openshift-ovn-kubernetes"
+    apiserver="{{.K8S_APISERVER}}"
+
+    [logging]
+    loglevel="{{.OVNKUBE_loglevel}}"
+
+    [gateway]
+    mode=local
+    nodeport=true
+
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -1,5 +1,5 @@
 # ovnkube-master
-# daemonset version 3
+# daemonset version 4
 # starts master daemons, each in a separate container
 # it is run on the master node(s)
 kind: Deployment
@@ -43,7 +43,7 @@ spec:
 
       # run-ovn-northd - v3
       - name: run-ovn-northd
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
 
         command: ["/root/ovnkube.sh", "run-ovn-northd"]
 
@@ -56,6 +56,8 @@ spec:
           name: host-var-lib-ovs
         - mountPath: /var/run/openvswitch/
           name: host-var-run-ovs
+        - mountPath: /var/run/ovn-config
+          name: ovn-config-mount
 
         resources:
           requests:
@@ -63,37 +65,16 @@ spec:
             memory: 300Mi
         env:
         - name: OVN_DAEMONSET_VERSION
-          value: "3"
-        - name: OVN_LOG_NORTHD
-          value: "-vconsole:info"
-        - name: OVN_NET_CIDR
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: net_cidr
-        - name: OVN_SVC_CIDR
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: svc_cidr
+          value: "4"
         - name: K8S_NODE
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-        - name: K8S_APISERVER
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: k8s_apiserver
-        - name: OVN_KUBERNETES_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
       # end of container
 
       # nb-ovsdb - v3
       - name: nb-ovsdb
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
 
         command: ["/root/ovnkube.sh", "nb-ovsdb"]
 
@@ -106,6 +87,8 @@ spec:
           name: host-var-lib-ovs
         - mountPath: /var/run/openvswitch/
           name: host-var-run-ovs
+        - mountPath: /var/run/ovn-config
+          name: ovn-config-mount
 
         resources:
           requests:
@@ -113,37 +96,16 @@ spec:
             memory: 300Mi
         env:
         - name: OVN_DAEMONSET_VERSION
-          value: "3"
-        - name: OVN_LOG_NB
-          value: "-vconsole:info -vfile:info"
-        - name: OVN_NET_CIDR
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: net_cidr
-        - name: OVN_SVC_CIDR
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: svc_cidr
+          value: "4"
         - name: K8S_NODE
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-        - name: K8S_APISERVER
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: k8s_apiserver
-        - name: OVN_KUBERNETES_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
       # end of container
 
       # sb-ovsdb - v3
       - name: sb-ovsdb
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
 
         command: ["/root/ovnkube.sh", "sb-ovsdb"]
 
@@ -158,6 +120,8 @@ spec:
           name: host-var-run-ovs
         - mountPath: /var/run/kubernetes/
           name: host-var-run-kubernetes
+        - mountPath: /var/run/ovn-config
+          name: ovn-config-mount
 
         resources:
           requests:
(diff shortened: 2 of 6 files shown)


pecameron:
@dcbw This has the changes we talked about. PTAL

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -56,35 +56,20 @@ spec:
...
         - name: K8S_NODE
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-        - name: K8S_APISERVER
-          valueFrom:
-            configMapKeyRef:
-              name: ovn-config
-              key: k8s_apiserver
         - name: OVN_KUBERNETES_NAMESPACE
```
dcbw: @pecameron I think OVN_KUBERNETES_NAMESPACE can be removed too since that is now passed in the config file.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -106,35 +91,20 @@ spec:
...
```
dcbw: Same as above, now passed in config file.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -158,35 +128,20 @@ spec:
...
```
dcbw: Now passed in config file.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -209,37 +164,22 @@ spec:
...
```
dcbw: Now passed in config file.

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: openshift-ovn-kubernetes
+data:
+  config:   |-
+    [default]
+    mtu="{{.MTU}}"
+
+    [kubernetes]
+    service-cidr="{{.OVN_service_cidr}}"
+    cluster-subnets={{.K8S_cluster_subnets}}
+    ovn-config-namespace="{{.OVNKUBE_namespace}}"
```
dcbw: @pecameron for OpenShift this value will always be "openshift-ovn-kubernetes" so I think we can just hardcode it here for the CNO. Upstream of course it needs to be configurable but will default to "ovn-kubernetes".

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: OVN_KUBERNETES_NAMESPACE is used elsewhere in the ovnkube.sh script. Also upstream doesn't use operators and we need to be careful to keep this correct.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: Same comment as above.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: ovnkube.sh ovn-master calls:
ready_to_start_node () {
# See if ep is available ...
  ovn_db_host=$(kubectl --server=${k8s_apiserver} --token=${k8s_token} --certificate-authority=${K8S_CACERT} \
    get ep -n ${ovn_kubernetes_namespace} ovnkube-db 2>/dev/null | grep 6642 | sed 's/:/ /' | awk '/ovnkube-db/{ print $2 }')
  if [[ ${ovn_db_host} == "" ]] ; then

pecameron:
@dcbw made the config file change. The yaml file change will need a change to ovn-config configmap to pass it namespace in. I'll get to that next.

pecameron:
@dcbw added namespace to the ovn-config configmap and and removed OVN_KUBERNETES_NAMESPACE from the daemonsets. PTAL

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: openshift-ovn-kubernetes
+data:
+  config:   |-
+    [default]
+    mtu="{{.MTU}}"
+
+    [kubernetes]
+    service-cidr="{{.OVN_service_cidr}}"
+    cluster-subnets={{.K8S_cluster_subnets}}
+    ovn-config-namespace="{{.OVN_namespace}}"
```
dcbw: @pecameron For the namespace comment, I actually meant, can't we just do:

`ovn-config-namespace=openshift-ovn-kubernetes`

here and we're done? No need to template that if we know it's always going to be the same thing for OpenShift.

pkg/network/ovn_kubernetes.go
```diff
@@ -52,6 +52,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.U
 		svcpools += fmt.Sprintf("%s", net)
 	}
 	data.Data["OVN_service_cidr"] = svcpools
+	data.Data["OVN_namespace"] = "openshift-ovn-kubernetes"
```
dcbw: Per above, do we really need to template this variable if it's going to be hardcoded?

pkg/network/ovn_kubernetes.go
```diff
@@ -52,6 +52,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.U
...
+
+	// ovnkube config file
+	data.Data["K8S_cluster_subnets"] = ippools
+	data.Data["OVNKUBE_loglevel"] = 2
+	data.Data["OVNKUBE_nb_privkey"] = ""
```
dcbw: I think for now for all these certs, let's just ignore them completely. When we do get SSL support we can add the template variables in.

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+    [logging]
+    loglevel=2
+
+    [ovnnorth]
+    client-privkey="{{.OVNKUBE_nb_privkey}}"
```
dcbw: For all the cert variables, let's just remove them from the config for now, since the default is empty anyway. When we figure out what needs to be done for SSL we can add them back.

pkg/network/ovn_kubernetes.go
```diff
@@ -52,6 +52,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.U
 		svcpools += fmt.Sprintf("%s", net)
 	}
 	data.Data["OVN_service_cidr"] = svcpools
+	data.Data["OVN_namespace"] = "openshift-ovn-kubernetes"
+
+	// ovnkube config file
+	data.Data["K8S_cluster_subnets"] = ippools
+	data.Data["OVNKUBE_loglevel"] = 2
```
dcbw: The config file hardcodes the loglevel, so we should be consistent. Either hardcode it in the config and don't bother templating, or template it all the way through and not hardcode it in the config file.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -58,13 +66,19 @@ spec:
...
             cpu: 100m
             memory: 300Mi
         env:
         - name: OVN_DAEMONSET_VERSION
-          value: "3"
+          value: "4"
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/root/ovnkube.sh", "cleanup-ovs-server"]
```
dcbw: @pecameron I looked at the upstream cleanup stuff for a while today RE our question Monday about cleanup. And I think we just don't follow upstream here at all, we just leave everything be. Reason is because we'll pick things up and reconcile on restart, and we may restart often for upgrades.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -152,44 +145,27 @@ spec:
...
-        - name: OVN_KUBERNETES_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
 
         lifecycle:
           preStop:
             exec:
-              command: ["rm","-f","/etc/cni/net.d/10-ovn-kubernetes.conf", "/host/opt/cni/bin/ovn-k8s-cni-overlay"]
+              command: ["/root/ovnkube.sh", "cleanup-ovn-node"]
```
dcbw: @pecameron same here, I think we should stick with what we currently have for the preStop hook. Otherwise we'll blow away a bunch of our logical network stuff and that'll disrupt networking during upgrades. We should just ignore the upstream cleanup-ovn-node stuff.

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
pecameron: It come up in both configmaps. Having it once here keeps things in sync.

pkg/network/ovn_kubernetes.go
pecameron: As above it comes up in both configmaps

pkg/network/ovn_kubernetes.go
pecameron: ok

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
dcbw: @pecameron Ok, that's fine then. When Numan's HA stuff lands we'll remove anything to do with endpoints from ovnkube.sh and then we can consolidate.

pkg/network/ovn_kubernetes.go
pecameron: Changed

bindata/network/ovn-kubernetes/ovnkube-node.yaml
pecameron: OK

bindata/network/ovn-kubernetes/ovnkube-node.yaml
pecameron: OK

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
squeed: Indeed, we should not be putting secrets in configmaps.

The typical solution is to write them on disk via Secrets, and just configure the path to them via config files.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -233,5 +209,24 @@ spec:
       - name: host-var-lib-cni-networks-ovn-kubernetes
         hostPath:
           path: /var/lib/cni/networks/ovn-k8s-cni-overlay
+      - configMap:
```
squeed: If these are already included in the ovn configuration files, why do they *also* need to be mounted to disk?

pecameron: ovn-config has a bunch of (ultimately) script variables. "cat" of the file is a simple (quick) way to get a value.
ovnkube-config is mounted and used as the config file for the ovnkube program.
I likely don't understand your question.

pecameron:
@dcbw @squeed PTAL

pecameron:
@dcbw @squeed This is now working. PTAL

bindata/network/ovn-kubernetes/006-ovnkube-config.yaml
pecameron: @squeed client-privkey is a defined part of the config file. 
@dcbw @squeed, eliminated [ovnnorth] and [ovnsouth] sections. When ssl is supported they can be added back.

pkg/network/ovn_kubernetes.go
pecameron: @dcbw Its now templated.

danwinship:
/hold
depends on https://github.com/ovn-org/ovn-kubernetes/pull/748 being merged and then openshift/ovn-kubernetes being rebased

pecameron:
@squeed @dcbw  replaced this with https://github.com/openshift/cluster-network-operator/pull/333
Refactored after PR 329 merged.

PR #218, opened 2019-07-01
Multus should look for CNI configuration in a different directory than kubelet

squeed:
This changes Multus to look for "default" CNI configurations in a different directory than where it writes it's file for CRIO. It also updates all plugins to write to that directory (when multus is enabled).

### DIFF
--- a/bindata/network/kuryr/004-daemon.yaml
+++ b/bindata/network/kuryr/004-daemon.yaml
@@ -82,10 +82,10 @@ spec:
       volumes:
       - name: bin
         hostPath:
-          path: /var/lib/cni/bin
+          path: {{.CNIBinDir}}
       - name: net-conf
         hostPath:
-          path: /etc/kubernetes/cni/net.d
+          path: {{.CNIConfDir}}
       - name: config-volume
         configMap:
           name: kuryr-config
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -55,6 +55,7 @@ spec:
         command: ["/entrypoint.sh"]
         args:
         - "--multus-conf-file=auto"
+        - "--multus-autoconfig-dir=/host/var/run/multus/cni/net.d"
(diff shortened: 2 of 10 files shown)


squeed:
Hmm, all the nodes came up Ready, so I don't think this caused any issues.
/retest

squeed:
Found the problem:
`Multus: Err in loading K8s Delegates k8s args: GetK8sClient: failed to get context for the kubeconfig /etc/cni/net.d/multus.d/multus.kubeconfig, refer Multus README.md for the usage guide: stat /etc/cni/net.d/multus.d/multus.kubeconfig: no such file or directory`

danwinship:
So can you clarify what this is changing? I thought we basically already did this...

squeed:
Multus implemented this option, but now we actually *use* it.

danwinship:
But I could have sworn we already did this... or was that just for the CNI binaries and now this includes the conf files as well?

squeed:
> But I could have sworn we already did this... or was that just for the CNI binaries and now this includes the conf files as well?

CNI binaries have always been the same directory - and that doesn't change. All I recall is talking about implementing this in multus - and now we're using that feature.

squeed:
Interesting, looks like it's actually failing upgrade from 4.1 - taking a look.

squeed:
This continuously fails upgrade. It seems that crio is somehow marking the node as NotReady, even though the CNI configuration file (from multus) is definitely on disk. Of course, restarting crio fixes it.

/hold

squeed:
Yeah, confirmed - there's a weird crio situation where if `/etc/kubernetes/cni/net.d/80-openshift.conf` is deleted, crio reports the node as NotReady, even though `00-multus.conf` exists. Restarting crio fixes this, but that's not acceptable.

Of course, we're in to this situation exactly because we didn't do the right thing and relied on usually winning a race in the first place.

squeed:
Filed https://github.com/cri-o/ocicni/issues/46 - let's see how that goes.

squeed:
closing for now (so it doesn't get accidentally merged).

PR #220, opened 2019-07-02
[WIP] Adding only IPTableSyncCluster

JacobTanenbaum:
for the purposes of testing, one or more rules breaks monitoring


JacobTanenbaum:
fulfilled the purposes of running testing

PR #224, opened 2019-07-03
Put openshift-etcd netnamespace on netid 1

rcarrillocruz:
Currently multitenant is broken since apiserver cannot connect
to etcd

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1719653

### DIFF
--- a/bindata/network/openshift-sdn/004-multitenant.yaml
+++ b/bindata/network/openshift-sdn/004-multitenant.yaml
@@ -55,7 +55,8 @@ netname: openshift-image-registry
 
 ---
 # Namespaces which are part of the "control plane" and need to reach each other:
-# - kube-system (etcd)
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/004-multitenant.yaml
```diff
@@ -55,7 +55,7 @@ netname: openshift-image-registry
 
 ---
 # Namespaces which are part of the "control plane" and need to reach each other:
-# - kube-system (etcd)
```
squeed: Can you add openshift-etcd to the comment?

squeed: I meant, add openshift-etcd to the list of namespaces, along with kube-system

PR #226, opened 2019-07-04
removing experimental annotation on multus ds config. 

alexanderConstantinescu:
This PR is created with respect to Bugzilla ID: [1726069](https://bugzilla.redhat.com/show_bug.cgi?id=1726069)

I'm paraphrasing the bug report: "Now that multus pods already use .spec.priorityClassName, they should not keep the experimental 'critical-pod' annotation"

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -20,8 +20,6 @@ spec:
         component: network
         type: infra
         openshift.io/component: network
-      annotations:
-        scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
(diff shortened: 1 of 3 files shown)


squeed:
Looks fine. While you're at it, can you also update the manifests in `bindata/network/ovn-kubernetes` to use the correct style (use priorityClassName, remove critical-pod).

alexanderConstantinescu:
> Looks fine. While you're at it, can you also update the manifests in `bindata/network/ovn-kubernetes` to use the correct style (use priorityClassName, remove critical-pod).

Done, please have a look

PR #227, opened 2019-07-04
Improve OpenStack resource naming and tagging

dulek:
This PR adds ClusterID to OpenStack resources names (in <cluster-id>-<name> form, exactly like openshift/installer is naming terraform resources) and openshiftClusterID=<cluster-id> tag to Octavia loadbalancer so it may be easily identified when installer is destroying the cluster.

Upgrade of gophercloud was required for the latter.


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -447,12 +449,40 @@ func waitForOpenStackLb(client *gophercloud.ServiceClient, lbId string) error {
...
+
+	apiVersions, err := apiversions.ExtractAPIVersions(allPages)
+	if err != nil {
+		return "", err
+	}
+
+	var octaviaTagSupport bool
+	octaviaTagSupport = false
+	for _, apiVersion := range apiVersions {
+		if apiVersion.ID >= MinOctaviaVersionWithTagSupport {
```
squeed: I think you need to use a real semver parser here...

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -447,12 +449,40 @@ func waitForOpenStackLb(client *gophercloud.ServiceClient, lbId string) error {
...
+
+	opts := loadbalancers.ListOpts{
 		Name:        name,
 		VipAddress:  vipAddress,
 		VipSubnetID: vipSubnetId,
-	}).AllPages()
+	}
+	if octaviaTagSupport {
+		opts.Tags = []string{tag}
+	} else {
```
squeed: Not knowing anything about OpenStack, does it make sense to set both the Tags and Description in the older API case?

squeed:
A few small questions, otherwise seems fine. cc @ricardocar who knows OpenStack a bit better than I.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: Yeah, this probably deserves a comment, it took me a while to understand @MaysaMacedo code, on which I based this.

Basically Octavia returns all the API versions it supports, so once we hit any of v2.5, v2.6, v2.7, v2.8, v2.9, we're assuming everything's fine.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: Not really, the code corresponding to this is here, it simply finds all the LB's having the cluster ID tag in either description or tags: https://github.com/openshift/installer/blob/63c63a0015f8f7cd894a93c0a6c0d6565ee8d05a/pkg/destroy/openstack/openstack.go#L606

Our goal is to support OpenShift's installation on OpenStack back to Queens release. The description is only a workaround that we use to support Queens and Rocky. Once support for those is deprecated we can remove this part. I guess this also needs a comment explaining it.

dulek:
I just updated the code to include explanations of how Octavia API versions are handled and why that way.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: Okay, so we discovered we'll be doing more Octavia version detection, so I use proper semver parser now.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -47,7 +47,8 @@ const (
 	CloudName             = "openstack"
 	CloudsSecretKey       = "clouds.yaml"
 	// NOTE(dulek): This one is hardcoded in openshift/installer.
-	InfrastructureCRDName = "cluster"
+	InfrastructureCRDName           = "cluster"
```
luis5tb: this needs gofmt I think

dulek: Actually gofmt put it like that - seems like the comment stops the alignment section.

PR #228, opened 2019-07-04
Introduce gosec for security checks

JAORMX:
gosec does static code analysis and checks for common security issues in golang codebases.

This PR introduces the tool, and sets it up as a test target in the Makefile; it also fixes a couple of warnings that were found.

### DIFF
--- a/cmd/cluster-network-renderer/main.go
+++ b/cmd/cluster-network-renderer/main.go
@@ -86,6 +86,7 @@ func readConfigObject(path string) (*operv1.Network, error) {
 
 // writeObjects serializes the list of objects as a single yaml file
 func writeObjects(path string, objs []*uns.Unstructured) error {
+	// #nosec
 	fp, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
 	if err != nil {
 		return errors.Wrapf(err, "could not open output file %s", path)
--- a/hack/test-sec.sh
+++ b/hack/test-sec.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+source "$(dirname "${BASH_SOURCE}")/init.sh"
+
+# Check for `go` binary and set ${GOPATH}.
+setup_env
+
+cd ${GOPATH}/src/${CNO_GO_PKG}
+
+if [[ ! $(which go) ]]; then
+  echo "go not found on PATH. To install:"
+  echo "https://golang.org/dl/"
+  exit 1
+fi
+if [[ ! $(which gosec) ]]; then
+  echo "gosec not found on PATH. To install:"
+  echo "go get -u github.com/securego/gosec/cmd/gosec"
+  exit 1
+fi
+
+gosec -severity medium --confidence medium \
+    -exclude G304 \
+    -quiet ./...
+retcode=$?
+
+exit $retcode


cmd/cluster-network-renderer/main.go
```diff
@@ -69,7 +70,7 @@ func render() error {
 
 // readConfigObject reads a NetworkConfig object from disk
 func readConfigObject(path string) (*operv1.Network, error) {
-	f, err := os.Open(path)
+	f, err := os.Open(filepath.Clean(path))
```
danwinship: What is the gosec warning? Who says that this is good practice?

cmd/cluster-network-renderer/main.go
```diff
@@ -87,7 +87,7 @@ func readConfigObject(path string) (*operv1.Network, error) {
 
 // writeObjects serializes the list of objects as a single yaml file
 func writeObjects(path string, objs []*uns.Unstructured) error {
-	fp, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+	fp, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
```
danwinship: This is just a test program, and the output is not security-sensitive anyway. So this seems like it's just making things less convenient.

cmd/cluster-network-renderer/main.go
JAORMX: The warning is:

[/home/josorior/go/src/github.com/openshift/cluster-network-operator/cmd/cluster-network-renderer/main.go:72] - G304: Potential file inclusion via variable (Confidence: HIGH, Severity: MEDIUM)

And the idea behind the warning is that, if a user has provided the path, they could potentially traverse the directory structure in unintended ways by inserting "../" in the given path. So, any path that's opened in the codebase that might come from a user (or an untrusted source) should be cleaned up: https://golang.org/pkg/path/filepath/#Clean

If you think this is not a problem in this codebase, we could easily just ignore this warning when invoking gosec.

cmd/cluster-network-renderer/main.go
JAORMX: As someone that's not acquainted with the codebase, there was no indication that this would be a test program. If you think these more restrictive permissions are not appropriate, we can skip this statement by adding the `// #nosec` comment above it.

cmd/cluster-network-renderer/main.go
squeed: For sure - adding this line would *create* a bug. This is a CLI tool, so I would 100% expect the user to be able to traverse any path they want.

cmd/cluster-network-renderer/main.go
squeed: If you want to fix this, can you find a way to write the file with the UMASK?

pkg/render/render.go
```diff
@@ -71,7 +71,7 @@ func RenderTemplate(path string, d *RenderData) ([]*unstructured.Unstructured, e
 	tmpl.Funcs(template.FuncMap{"getOr": getOr, "isSet": isSet})
 	tmpl.Funcs(sprig.TxtFuncMap())
 
-	source, err := ioutil.ReadFile(path)
+	source, err := ioutil.ReadFile(filepath.Clean(path))
```
squeed: This also introduces a bug. If the path contains parent-directory marks, then we absolutely should follow them.

If we wanted to reject them (which we don't), then as a general rule, silently removing them is the wrong way to go about it. Instead, you should report a failure to the user and halt the program.

cmd/cluster-network-renderer/main.go
danwinship: That warning seems pretty specific to the case of code that responds to HTTP requests by combining a trusted path prefix with a user-provided suffix, and doesn't really apply to anything that we're doing in OpenShift. Additionally, the fact that adding `filepath.Clean()` like this makes the warning go away means that the analyzer isn't even doing a good job, because that doesn't actually fix the HTTP-server-combining-path security hole. (You'd need to clean the user-provided suffix separately *before* combining it with the trusted prefix.)

So I think we should be disabling this warning in *all* OpenShift codebases, because it's mostly not relevant, and it seems to encourage cargo-cult fixes that don't actually address the problem.

cmd/cluster-network-renderer/main.go
danwinship: This is also kind of more a problem for web-server-type programs than it is for us, but permissions are still a problem for us, so I'm ok with the check in general. But not in this program, so yes, please `// #nosec` it

JAORMX: done

cmd/cluster-network-renderer/main.go
JAORMX: alright, disabled this check.

PR #230, opened 2019-07-08
Kuryr: Use ens3 as link_iface in kuryr.conf

dulek:
Seems like RHCOS VM's use ens3 as the name of the Neutron interface they
get. In order for Kuryr to correctly attach subports to that interface
this commit sets [binding]link_iface=ens3 in kuryr.conf.

### DIFF
--- a/bindata/network/kuryr/003-config.yaml
+++ b/bindata/network/kuryr/003-config.yaml
@@ -10,7 +10,7 @@ data:
 
     [binding]
     default_driver = kuryr.lib.binding.drivers.vlan
-    link_iface = eth0
+    link_iface = ens3
 
     [cni_daemon]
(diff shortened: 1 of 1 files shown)


danwinship:
This seems pretty fragile... there's no way to discover the right interface automatically?

dulek:
@danwinship: Agreed. We initially looked at autodetection, but it seems it isn't really trivial in Neutron. Anyway that would be a change in Kuryr itself as I don't see how it could be done from the CNO. We're tracking that idea here and hope to give it some more thinking in the next sprint: https://blueprints.launchpad.net/kuryr-kubernetes/+spec/autodetect-nested-iface. 

Meanwhile I believe changing the setting to match what we see in RHCOS isn't really controversial.

PR #231, opened 2019-07-08
Kuryr: Disable kube-proxy

dulek:
Kuryr handles services on its own, so it doesn't need kube-proxy. This
commit disables its deployment when Kuryr is configured.

### DIFF
--- a/pkg/network/proxy.go
+++ b/pkg/network/proxy.go
@@ -16,13 +16,16 @@ import (
 
 // ShouldDeployKubeProxy determines if the desired network type should
 // install kube-proxy.
-// openshift-sdn and OVN-Kubernetes deploy their own kube-proxy. All other
+// openshift-sdn deploys its own kube-proxy. ovn-kubernetes and
+// Kuryr-Kubernetes handle services on their own. All other
 // network providers are assumed to require kube-proxy
 func ShouldDeployKubeProxy(conf *operv1.NetworkSpec) bool {
 	switch conf.DefaultNetwork.Type {
 	case operv1.NetworkTypeOpenShiftSDN:
 		return false
 	case operv1.NetworkTypeOVNKubernetes:
+		return false
+	case operv1.NetworkTypeKuryr:
 		return false
 	default:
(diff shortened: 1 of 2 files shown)


pkg/network/proxy.go
```diff
@@ -16,14 +16,17 @@ import (
 
 // ShouldDeployKubeProxy determines if the desired network type should
 // install kube-proxy.
-// openshift-sdn and OVN-Kubernetes deploy their own kube-proxy. All other
+// openshift-sdn and OVN-Kubernetes deploy their own kube-proxy.
+// Kuryr-Kubernetes handles services on its own. All other
```
danwinship: Actually, the existing text is wrong about ovn-kubernetes. It should say "OpenShift SDN deploys its own kube-proxy. OVN Kubernetes and Kuryr-Kubernetes handle services on their own".

pkg/network/proxy.go
```diff
@@ -16,14 +16,17 @@ import (
...
 // network providers are assumed to require kube-proxy
 func ShouldDeployKubeProxy(conf *operv1.NetworkSpec) bool {
 	switch conf.DefaultNetwork.Type {
 	case operv1.NetworkTypeOpenShiftSDN:
 		return false
 	case operv1.NetworkTypeOVNKubernetes:
 		return false
+	case operv1.NetworkTypeKuryr:
+		return false
```
danwinship: You ought to have unit tests that need to be updated because of this change...

dulek:
@danwinship: I updated the comment and added unit test case.

danwinship:
/lgtm

I was assuming that since the original change didn't break anything that that must mean you didn't have unit tests of the kuryr rendering at all, and that's what I was commenting on. But I see that `pkg/network/kuryr_test.go` does exist, it's just that (like the openshift-sdn tests), it doesn't check *every* object that gets generated, so it didn't notice one more or fewer. Which is probably for the best, in terms of not having to update the test results for every test any time something tiny changes...

PR #232, opened 2019-07-08
Adding upgradeable status and setting it to always "true" for now.

alexanderConstantinescu:
This PR sets the "Upgradeable" condition to always be true. Fixes [Bugzilla report 1727026](https://bugzilla.redhat.com/show_bug.cgi?id=1727026)

Some requested feedback on this PR: I am not sure if the best way of doing this is by using the setter (as I have done). This will mean that we just reset the upgradeable status to true everytime any condition status is set of course. I didn't find any more straightforward way of doing this however, i.e: clearer way and without code duplication. Please let me know if anyone has any ideas.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -267,6 +267,9 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// Run a pod status check just to clear any initial inconsitencies at startup of the CNO
+	r.status.SetFromPods()
+
 	// Update Network.config.openshift.io.Status
 	status, err := r.ClusterNetworkStatus(context.TODO(), operConfig)
 	if err != nil {
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -86,6 +86,13 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...confi
 		)
 	}
 
+	v1helpers.SetStatusCondition(&co.Status.Conditions,
+		configv1.ClusterOperatorStatusCondition{
+			Type:   configv1.OperatorUpgradeable,
+			Status: configv1.ConditionTrue,
+		},
+	)
+
 	if reflect.DeepEqual(oldStatus, co.Status) {
 		return
 	}
--- a/pkg/controller/statusmanager/status_manager_test.go
+++ b/pkg/controller/statusmanager/status_manager_test.go
@@ -70,6 +70,11 @@ func TestStatusManager_set(t *testing.T) {
 		t.Fatalf("unexpected error (expected Not Found): %v", err)
 	}
 
+	condUpdate := configv1.ClusterOperatorStatusCondition{
+		Type:   configv1.OperatorUpgradeable,
+		Status: configv1.ConditionTrue,
+	}
+
 	condFail := configv1.ClusterOperatorStatusCondition{
 		Type:    configv1.OperatorDegraded,
 		Status:  configv1.ConditionTrue,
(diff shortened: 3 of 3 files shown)


danwinship:
test-verify failure is legit:

    gofmt failed. fix with
    gofmt -w -s ./pkg/controller/statusmanager/status_manager.go ./pkg/controller/statusmanager/status_manager_test.go

> I am not sure if the best way of doing this is by using the setter (as I have done).

I think that's probably the best place to put it for now... we're going to need to refactor `StatusManager` a bit soon to fix https://bugzilla.redhat.com/show_bug.cgi?id=1725220 anyway

alexanderConstantinescu:
> test-verify failure is legit:

Sure, saw that too in the test results. Fixing it tomorrow 

Thanks for the feedback!

alexanderConstantinescu:
**Update**

@danwinship @squeed : I have updated this PR to fix [1725220](https://bugzilla.redhat.com/show_bug.cgi?id=1725220) as well. We now consolidate the state of all the related daemonset/deployment pods once the CNO boots [here,](https://github.com/openshift/cluster-network-operator/pull/232/commits/9b559d0e66bbd41855f6d647572ba8f501928f8e#diff-56dd8b1c4d128ccb1f60958db5d9e6c4R220) thus eliminating conditions such as the ones mentioned in that bug report. 

I think this deserves some thought and discussion though. As mentioned in the leading comment to this PR, we will reset the "Upgradeable" status over and over, for nothing really. 

I have tried to refactor the StatusManager a bit to have some form of initialization mechanism in it's constructor ("New"). The problem is that for that to work we'd need to communicate with the k8s API server and retrieve the CNO from the resource list and update it back with the correct initialized status. Retrieving it is a problem however as the k8s uses something called a DelegatingClient (consisting of a cache client and API client), and always defaulting to the cache client. This cache is however only populated once the controller manager starts (which is after the initialization of the StatusManager).....

...long story short: do you have any idea of how to refactor the StatusManager to a smarter StatusManager? Keeping in mind that the StatusManager is a shared object coordinating state between all CNO controllers (pod_controller.go, operconfig_controller.go, clusterconfig_controller.go) and needs to be initialized before all others.

I am up for suggestions

danwinship:
> ...long story short: do you have any idea of how to refactor the StatusManager to a smarter StatusManager? Keeping in mind that the StatusManager is a shared object coordinating state between all CNO controllers (pod_controller.go, operconfig_controller.go, clusterconfig_controller.go) and needs to be initialized before all others.

Usually in cases like this we split out a `Start` method for the stuff that has to happen a little bit later after everything else has initialized. So the `StatusManager` could still be *created* very early, but wouldn't try to synchronize with the apiserver until `Start` is called.

squeed:
>  As mentioned in the leading comment to this PR, we will reset the "Upgradeable" status over and over, for nothing really.

That's not really a problem here - we short-circuit if the status is the same anyways.

alexanderConstantinescu:
PR updated according to the discussion yesterday, that being:

- We will not initialize the StatusManager with default states prior to starting the controller manager, we will just construct the object. Basically it remains as is. 

This PR thus addresses bugs: [1727026](https://bugzilla.redhat.com/show_bug.cgi?id=1727026) and [ 1725220](https://bugzilla.redhat.com/show_bug.cgi?id=1725220)

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -231,6 +231,7 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	r.status.SetDaemonSets(daemonSets)
 	r.status.SetDeployments(deployments)
 	r.status.SetRelatedObjects(relatedObjects)
+	r.status.SetFromPods()
```
danwinship: Calling `SetFromPods()` before reconciling the objects will mean that you'll get a `Progressing` status with a bunch of warnings like "`Error getting DaemonSet openshift-sdn/sdn: Not Found`". That will get cleared eventually, but we prefer to not temporarily set incorrect statuses.

Moving this to *after* the `Apply the object to the cluster` loop should fix that.

alexanderConstantinescu: Thanks for the comment, just updated my PR

PR #233, opened 2019-07-08
Bug 1751756: Kuryr: Add health monitor to Octavia load balancer

dulek:
Kuryr bootstrap process creates an Octavia loadbalancer that serves as
OpenShift API endpoint for the pods (i.e. kubernetes.default service on
the first IP of service subnet). It turns out during beginning of the
installation phase, the master nodes will not yet provide API access,
only bootstrap node will.

In order to support such situation this commit adds the bootstrap node
to the loadbalancer pool. Also to make sure the LB will point to correct
addresses, a health monitor is added to the pool. This ensures that the
master nodes will only get added to the LB when they're up.

Besides that it sets weight for load balancer members to prioritize genuine master nodes in favor of bootstrap node.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"regexp"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
@@ -619,7 +620,7 @@ func ensureOpenStackLbMonitor(client *gophercloud.ServiceClient, name, poolId st
 // it does not exist creates it. Will fail if multiple LB pool members are
 // matching all criteria.
 func ensureOpenStackLbPoolMember(client *gophercloud.ServiceClient, name, lbId, poolId,
-	address, subnetId string, port int) (string, error) {
+	address, subnetId string, port, weight int) (string, error) {
 	page, err := pools.ListMembers(client, poolId, pools.ListMembersOpts{
 		Name:         name,
 		Address:      address,
@@ -642,6 +643,7 @@ func ensureOpenStackLbPoolMember(client *gophercloud.ServiceClient, name, lbId, 
 			Address:      address,
 			ProtocolPort: port,
 			SubnetID:     subnetId,
+			Weight:       &weight,
 		}
 		poolsObj, err := pools.CreateMember(client, poolId, opts).Extract()
 		if err != nil {
@@ -1016,8 +1018,16 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		if len(port.FixedIPs) > 0 {
 			portIp := port.FixedIPs[0].IPAddress
 			log.Printf("Found port %s with IP %s", port.ID, portIp)
+
+			// We want bootstrap to stop being used as soon as possible, as it will serve
+			// outdated data during the bootstrap transition period.
+			weight := 100
+			if strings.HasSuffix(port.Name, "bootstrap-port") {
+				weight = 1
+			}
+
 			memberId, err := ensureOpenStackLbPoolMember(lbClient, port.Name, lbId,
-				poolId, portIp, svcSubnetId, 6443)
+				poolId, portIp, svcSubnetId, 6443, weight)
 			if err != nil {
 				log.Printf("Failed to add port %s (%s) to LB pool %s: %s", port.ID, port.Name, poolId, err)
 				continue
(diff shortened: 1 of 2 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -410,13 +411,19 @@ func ensureOpenStackSg(client *gophercloud.ServiceClient, name, tag string) (str
...
-func ensureOpenStackSgRule(client *gophercloud.ServiceClient, sgId, remotePrefix string) error {
+func ensureOpenStackSgRule(client *gophercloud.ServiceClient, sgId, remotePrefix string, portMin, portMax int) error {
 	opts := rules.CreateOpts{
 		SecGroupID:     sgId,
 		EtherType:      rules.EtherType4,
 		Direction:      rules.DirIngress,
 		RemoteIPPrefix: remotePrefix,
 	}
+	// Let's just assume that we're getting passed -1 when we aren't supposed to set those
+	if portMin >= 0 && portMax >= 0 {
```
squeed: is it possible for one of them to be -1?

dulek: -1 is just a mnemonic for "not set it", if my golang understanding isn't completely wrong I would need to create 2 versions of this method otherwise.
I can also add a boolean parameter marking if portMin and portMax should be ignored, if you think that's cleaner?

squeed: Oh nevermind, I see those are only set from the code. This is fine.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -525,6 +532,52 @@ func ensureOpenStackLbPool(client *gophercloud.ServiceClient, name, lbId string)
...
+	} else {
+		opts := monitors.CreateOpts{
+			Name:       name,
+			PoolID:     poolId,
+			Type:       monitors.TypeTCP,
+			MaxRetries: 3,
+			Delay:      30,
+			Timeout:    10,
+
+			// FIXME(dulek): Code fragment below won't work in OpenStack Rocky as Octavia won't allow url_path on
```
squeed: This is too bad. We configure the apiserver to report unhealthy for 30 seconds before shutting down, so that connections aren't disrupted. Definitely file a bug / issue to properly configure server health checks.

dulek: All right, I can probably try to make this work like that at least on on Stein if I have version detection. In that case the only issue I have is access to the healthz endpoint - does API allow unauthenticated access to it?

danwinship:
> Kuryr bootstrap process creates an Octavia loadbalancer that serves as
> OpenShift API endpoint for the pods (i.e. kubernetes.default service on
> the first IP of service subnet). It turns out during beginning of the
> installation phase, the master nodes will not yet provide API access,
> only bootstrap node will.

Isn't this just reinventing the wheel? How do non-Kuryr OpenShift-on-OpenStack installs work?

pkg/platform/openstack/kuryr_bootstrap.go
squeed: I believe it does - AWS does a similar thing.

dulek: Yup, we confirmed that. :)

dulek:
Okay, this isn't necessary immediately and I'll need to figure out Octavia API compatibility for different versions as each of supported versions treats HTTPS health monitors differently.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: I added the required version detection for Stein.

dulek:
Alright, we believe this will become necessary to fix 1751756. We're now in the process of testing all about it, so I'm going to hold this for now:

/hold

MaysaMacedo:
This was tested and it's working fine.

/lgtm

dulek:
> > Kuryr bootstrap process creates an Octavia loadbalancer that serves as
> > OpenShift API endpoint for the pods (i.e. kubernetes.default service on
> > the first IP of service subnet). It turns out during beginning of the
> > installation phase, the master nodes will not yet provide API access,
> > only bootstrap node will.
> 
> Isn't this just reinventing the wheel? How do non-Kuryr OpenShift-on-OpenStack installs work?

In case of OpenShift-SDN + ShiftStack (how a nice name, isn't it?) the API LB creation is handled by kube-proxy. Kuryr runs without it (no way to connect kube-proxy to how pods are networked by Kuryr, believe me - we tried), so while it would make sense for Kuryr itself to handle it, we encountered some issues in that approach that would require more work and testing. That's why we're sticking to LB creation on bootstrap.

I'm not saying we're not going to get rid of this in 4.3 or 4.4, but definitely not in 4.2 timeframe.

danwinship:
> In case of OpenShift-SDN + ShiftStack (how a nice name, isn't it?) the API LB creation is handled by kube-proxy.

Kube-proxy does not create load balancers.

But anyway, you've already gotten review on the code part. I'm just approving the Gopkg.lock part.
/approve

PR #235, opened 2019-07-09
Multus admission controller support

s1061123:
Talk with @dougbtv we decide not to bring isolation feature at this time. So the PR only contains /validation for net-attach-def.

### DIFF
--- a/bindata/network/multus-admission-controller/001-service.yaml
+++ b/bindata/network/multus-admission-controller/001-service.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: multus-admission-controller
+  namespace: openshift-multus
+  labels:
+    app: multus-admission-controller
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: multus-admission-controller-secret
+spec:
+  ports:
+  - port: 443
+    targetPort: 6443
+  selector:
+    app: multus-admission-controller
--- a/bindata/network/multus-admission-controller/admission-controller.yaml
+++ b/bindata/network/multus-admission-controller/admission-controller.yaml
@@ -1,0 +1,63 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: multus-admission-controller
+  namespace: openshift-multus
+  labels:
+    app: multus-admission-controller
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the Multus admisson controller component on each node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: multus-admission-controller
+      namespace: openshift-multus
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: multus-admission-controller
+        namespace: openshift-multus
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      containers:
+      - name: multus-admission-controller
+        image: {{.MultusAdmissionControllerImage}}
+        command:
+        - /usr/bin/webhook
+        args:
+        - -bind-address=0.0.0.0
+        - -port=6443
+        - -tls-private-key-file=/etc/webhook/tls.key
+        - -tls-cert-file=/etc/webhook/tls.crt
+        - -alsologtostderr=true
+        volumeMounts:
+        - name: webhook-certs
+          mountPath: /etc/webhook
+          readOnly: True
+        imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            cpu: 10m
+      serviceAccountName: multus
+      priorityClassName: "system-cluster-critical"
+      restartPolicy: Always
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      volumes:
+      - name: webhook-certs
+        secret:
+          secretName: multus-admission-controller-secret
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
+        effect: NoSchedule
+      - key: "node.kubernetes.io/not-ready"
+        operator: Exists
+        effect: NoSchedule
--- a/pkg/controller/operconfig/configmap_controller.go
+++ b/pkg/controller/operconfig/configmap_controller.go
@@ -1,0 +1,108 @@
+package operconfig
+
+import (
+	"bytes"
+	"context"
+	"log"
+
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"github.com/openshift/cluster-network-operator/pkg/names"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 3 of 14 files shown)


bindata/network/multus-admission-controller/001-service.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: multus-admission-controller
+  namespace: openshift-multus
+  labels:
+    app: multus-admission-controller
+  annotations:
+    service.alpha.openshift.io/serving-cert-secret-name: multus-admission-controller-secret
+spec:
+  ports:
+  - port: 443
+    targetPort: 443
```
squeed: I'd recommend adding this as port 6443, and running the admission controller as non-root.

bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -0,0 +1,63 @@
...
+        node-role.kubernetes.io/master: ""
+      volumes:
+      - name: webhook-certs
+        secret:
+          secretName: multus-admission-controller-secret
+      tolerations:
+      - key: "node-role.kubernetes.io/master"
+        operator: Exists
+        effect: NoSchedule
+      - key: "node.kubernetes.io/not-ready"
```
squeed: I don't think you can tolerate this, since your pod isn't HostNetwork.

bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -0,0 +1,63 @@
+---
+kind: DaemonSet
```
squeed: Does this need to run on all nodes? I would rather either be a Deployment with a reasonable number of replicas (3?), or a DaemonSet only on the control plane.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -86,6 +88,16 @@ func add(mgr manager.Manager, r *ReconcileOperConfig) error {
 		return err
 	}
 
+	// Likewise for the ConfigMap reconciler
```
squeed: Rather than putting this here, please create an `Add()` function similar to `clusterconfig_controller.go` and add it to `add_networkconfig.go()`

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -86,6 +88,16 @@ func add(mgr manager.Manager, r *ReconcileOperConfig) error {
...
+	c, err = controller.New("configmap-controller", mgr, controller.Options{Reconciler: r.configMapReconciler})
+	if err != nil {
+		return err
+	}
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{})
```
squeed: I'm not certain if it will work, but I'd like to see this use `hander.EnqueueRequestForOwner{}`. That way we don't try and reconcile *every* configmap update.

pkg/controller/operconfig/configmap_controller.go
```diff
@@ -0,0 +1,78 @@
...
+			modified = true
+			webhookConfig.Webhooks[idx].ClientConfig.CABundle = []byte(caBundleData)
+		}
+	}
+	if !modified {
+		return reconcile.Result{}, err
+	}
+
+	// Update webhookConfig
+	err = r.client.Update(context.TODO(), webhookConfig)
```
squeed: If the webhook configuration is created by the render phase, then won't it be overridden by the next render phase? You may need to update the reconciliation logic, or create the object here.

There is probably a best practice for this which I am not aware of.

squeed:
Hey @deads2k, is there a best-practice for how to update the webhook's configuration with the CA provided by the service-serving-cert?

bindata/network/multus-admission-controller/admission-controller.yaml
s1061123: As the conclusion of OpenShift plumbing meeting, we decide to run admission controller for all master nodes (for HA). So DaemonSet + master toleration could be possible.

Could you please let me know how to run Pod for all master node (that will be configured so we cannot use '3' as fixed number) with ReplicaSet, if you know about it? I will follow you.

bindata/network/multus-admission-controller/admission-controller.yaml
s1061123: As the conclusion of OpenShift plumbing meeting, we decide to run admission controller for all master nodes (for HA). So DaemonSet + master toleration could be possible.

Could you please let me know how to run Pod for all master node (that will be configured so we cannot use '3' as fixed number) with ReplicaSet, if you know about it? I will follow you.

pkg/controller/operconfig/configmap_controller.go
s1061123: In this case, we need to bring only BundleCA value ConfigMap value to webhook. So ConfigMap value is not changed and webhook is already created initially, so it should make sense.

Could you please let me know the reason why need to create WebhookConfig here?
Do you mean we need to update ConfigMap at that time? If so what should be changed in ConfigMap?

bindata/network/multus-admission-controller/001-service.yaml
s1061123: Changed.

pkg/controller/operconfig/operconfig_controller.go
s1061123: Changed.

pkg/controller/operconfig/configmap_controller.go
s1061123: I've check WebhookConfigration with `oc get ValidatingWebhookConfiguration multus.openshift.io -o jsonpath='{.webhooks[0].clientConfig.caBundle}' --watch=true`. I can see same output (same caBundle) periodically (about 5min) without any changes. So CNO touch the webhook but caBundle is not changed.

pkg/controller/operconfig/configmap_controller.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func addConfigMapReconciler(mgr manager.Manager, r reconcile.Reconciler) error {
+	// ConfigMap reconciler
+	c, err := controller.New("configmap-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{})
```
squeed: Can you configure this to be 
`c.Watch(...), &handler.EnqueueRequestForOwner{OwnerType: &operv1.Network})`

pkg/controller/operconfig/configmap_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Reconcile updates ValidatingWebhookConfiguration CABundle, given from SERVICE_CA_CONFIGMAP
+func (r *ReconcileConfigMaps) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if request.Namespace != names.APPLIED_NAMESPACE || request.Name != names.SERVICE_CA_CONFIGMAP {
```
squeed: Since this is called the configmap_controller, I would rather either
1 - rename this entire controller to the MultusWebhookController, or
2 - based on namespace + name, call separate functions. Something like

```go
if request.Namespace == ... {
  return reconcileMultusWebhook()
}
```

squeed:
Two small comments, then it looks good.

pkg/controller/operconfig/configmap_controller.go
s1061123: Changed.

pkg/controller/operconfig/configmap_controller.go
s1061123: Changed 2).

PR #236, opened 2019-07-10
Kuryr: Fix setting status on Network CRD

dulek:
Looks like due to an overlook we haven't enabled setting status on
Network CRD when running with Kuryr. This made the certificates
presented by openshfit-apiserver lack the IP as Common Name, causing SSL
validation to fail when trying to call API using the IP.

This commit fixes that by making sure StatusFromOperatorConfig() is not
returning nil when Kuryr is configured.

### DIFF
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -91,6 +91,8 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec) *configv1.NetworkSta
 		// continue
 	case operv1.NetworkTypeOVNKubernetes:
 		// continue
+	case operv1.NetworkTypeKuryr:
+		// continue
 	default:
 		return nil
 	}


pkg/network/cluster_config.go
```diff
@@ -116,6 +118,15 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec) *configv1.NetworkSta
 		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.OpenShiftSDNConfig.MTU)
 	case operv1.NetworkTypeOVNKubernetes:
 		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.OVNKubernetesConfig.MTU)
+	case operv1.NetworkTypeKuryr:
+		// NOTE(dulek): For Kuryr this doesn't really matter too much as MTU is managed by Neutron and we'll
```
danwinship: If it doesn't matter, you should be able to just leave it as 0

dulek: Great, I simply wasn't sure why it gets written there so wanted to give a value closest to the truth.

danwinship: In the openshift-sdn/ovn-kubernetes case, this is the MTU of the tunnel (VXLAN/GENEVE), and it's possible that the value that CNO autodetects will be wrong (if you have a heterogeneous cluster), so we wanted it to be easy to see what that autodetected value was. If you don't have a tunnel, or it's guaranteed that the autodetected value is correct, then there's no reason to worry about it here.

PR #238, opened 2019-07-11
Add OWNERS files for openstack components.

squeed:


squeed:
cc @dulek

danwinship:
FTR, although it's ugly, it's possible to have file-level reviewers (eg, for `pkg/network/kuryr.go`). eg, see [`k/k/OWNERS`](https://github.com/kubernetes/kubernetes/blob/master/OWNERS).

Also, we could make use of `OWNERS_ALIASES` to avoid duplicating the list of kuryr hackers in multiple places

squeed:
Updated to use aliases, good point. That regex business looks like more trouble than it's worth :-).

squeed:
aws grumpiness
/retest

PR #239, opened 2019-07-12
Make Kuryr connect to API through api-int LB

dulek:
Turns out we should use api-int loadbalancer to make Kuryr connect to
the OpenShift API. This commit implements that.


dulek:
/lgtm

I doubt I can lgtm my own stuff?

PR #241, opened 2019-07-16
Adding egress IP addresses to the CRD output

alexanderConstantinescu:
This addresses bug [1707749](https://bugzilla.redhat.com/show_bug.cgi?id=1707749)

Please let me know if the description should be modified.

/Alex

### DIFF
--- a/bindata/network/openshift-sdn/001-crd.yaml
+++ b/bindata/network/openshift-sdn/001-crd.yaml
@@ -117,6 +117,14 @@ spec:
     type: string
     description: The CIDR range of the overlay network assigned to the node for its pods
     JSONPath: .subnet
+  - name: Egress CIDRs
+    type: string
+    description: The network egress CIDRs
+    JSONPath: .egressCIDRs
+  - name: Egress IPs
+    type: string
+    description: The network egress IP addresses
+    JSONPath: .egressIPs
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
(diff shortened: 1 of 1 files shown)


danwinship:
What output does this result in? I thought I had tried this before and determined that the CRD printer didn't support array-valued columns.

alexanderConstantinescu:
> What output does this result in? I thought I had tried this before and determined that the CRD printer didn't support array-valued columns.

Please find an attached screenshot of a local edit with bogus IPs. 

`NAME                                                    NETID      EGRESSIPS
default                                                 0          
kube-node-lease                                         14579022   
kube-public                                             9731826    
openshift-sdn                                           15140382   [89.16.153.196 98.65.165.22]
`

bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -117,6 +117,10 @@ spec:
     type: string
     description: The CIDR range of the overlay network assigned to the node for its pods
     JSONPath: .subnet
+  - name: EgressIPs
```
danwinship: In HostSubnet, you should also print the `EgressCIDRs`, before the `EgressIPs`. (https://github.com/openshift/origin/blob/release-3.11/pkg/printers/internalversion/printer.go#L66)

alexanderConstantinescu: PR updated

PR #242, opened 2019-07-16
Kuryr: Expand svc net to fix VRRP ports conflicts

dulek:

To fix this the service network created in OpenStack will be expanded to
be twice bigger than the ServiceNetwork value from NetworkSpec. Then the
other half of it will be used for Amphora VRRP ports, while OpenShift
will only choose IP's from the half indicated by
NetworkSpec.ServiceNetwork.

Obviously this may lead to conflicts if expanded network overlaps with
ClusterNetwork, so validation is expanded to check that.

### DIFF
--- a/pkg/network/kuryr.go
+++ b/pkg/network/kuryr.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"net"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -13,6 +14,7 @@ import (
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"github.com/openshift/cluster-network-operator/pkg/render"
+	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
 )
 
 // renderKuryr returns manifests for Kuryr SDN.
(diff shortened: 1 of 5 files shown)


dulek:
We still need to verify this.

danwinship:
> Obviously this may lead to conflicts if expanded network overlaps with
> ClusterNetwork, so validation is expanded to check that.

It might also overlap with a network the user is using for something unrelated to OpenShift. And it contradicts the documentation of the installer configuration.

One way that you could hack around this would be to cut the value in half rather than doubling it: if the `config.openshift.io/v1/Network` has a `spec.ServiceNetwork` of `["172.30.0.0/16"]`, then set `status.ServiceNetwork` to `["172.30.0.0/17"]` and use `172.30.128.0/17` for the VRRP IPs. The kube-controller-manager-operator reads the value from `status`, not `spec`, so this would cause it to only assign Service IPs from the bottom half of the originally configured range.

Alternatively, you could have a separate kuryr-specific config option for specifying the VRRP IP range.

dulek:
> > Obviously this may lead to conflicts if expanded network overlaps with
> > ClusterNetwork, so validation is expanded to check that.
> 
> It might also overlap with a network the user is using for something unrelated to OpenShift. And it contradicts the documentation of the installer configuration.
> 
> One way that you could hack around this would be to cut the value in half rather than doubling it: if the `config.openshift.io/v1/Network` has a `spec.ServiceNetwork` of `["172.30.0.0/16"]`, then set `status.ServiceNetwork` to `["172.30.0.0/17"]` and use `172.30.128.0/17` for the VRRP IPs. The kube-controller-manager-operator reads the value from `status`, not `spec`, so this would cause it to only assign Service IPs from the bottom half of the originally configured range.
> 
> Alternatively, you could have a separate kuryr-specific config option for specifying the VRRP IP range.

Makes sense, I'll look into that.

squeed:
Oh boy, this raises an interesting question: what CIDR do you install in the Status field? That value is consumed by the control plane for validating Service objects, so it should probably see the /17.

Were you using kube-proxy, you would need to tell it the "real" ServiceCIDR so it could set up the masquerading bits correctly.

luis5tb:
> Oh boy, this raises an interesting question: what CIDR do you install in the Status field? That value is consumed by the control plane for validating Service objects, so it should probably see the /17.
> 
> Were you using kube-proxy, you would need to tell it the "real" ServiceCIDR so it could set up the masquerading bits correctly.

Good thing is that when kuryr is enabled kube-proxy is not used/deployed

dulek:
> > Oh boy, this raises an interesting question: what CIDR do you install in the Status field? That value is consumed by the control plane for validating Service objects, so it should probably see the /17.
> > Were you using kube-proxy, you would need to tell it the "real" ServiceCIDR so it could set up the masquerading bits correctly.
> 
> Good thing is that when kuryr is enabled kube-proxy is not used/deployed

Yep, my understanding is that Status field in the current implementation should have /16, the subnet expansion should be transparent for anyone besides Octavia.

If I use the @danwinship idea of splitting /16 into two /17, then Status gets /17.

dulek:
I think this gets superseded by #244, this version deliberately goes over ranges specified in the install-config and might conflict with whatever the user has in his environment.

PR #243, opened 2019-07-16
Drop open-ovn-ports.sh

danwinship:
no longer needed with current versions of the installer


caseydavenport:
@danwinship @squeed was this logic moved somewhere else or replaced with something else?

I'm wondering because we use some of this scripting logic to open ports for Calico as well, and noticed that it has recently stopped working on OpenShift v4.2.

danwinship:
This script was only ever used by `./hack/run-locally.sh`, not in real clusters.

openshift-install should be opening ports for VXLAN and Geneve on all clusters, regardless of network configuration. If you need any ports beyond that, then you need to either open them yourself, or switch to using something in the 9000-9999 range (as in #313), which is open between all nodes, but not exposed outside the cluster.

PR #244, opened 2019-07-16
Kuryr: Split svc net to avoid VRRP ports conflicts

dulek:

To fix this the service network created in OpenStack will be whole 
ServiceNetwork value from NetworkSpec. Then the other half of it
will be used for Amphora VRRP ports, while OpenShift will only choose
IP's from the half that we override in NetworkSpec.ServiceNetwork and
will be set in the Status.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -207,14 +207,14 @@ func findOpenStackSubnet(client *gophercloud.ServiceClient, name, tag string) (s
 // Looks for a Neutron subnet by name, tag, network ID, CIDR and optionally
 // gateway IP, if it does not exist creates it. Will fail if two subnets match
 // all the criteria.
-func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId, cidr string, gatewayIp *string) (string, error) {
+func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId, cidr, gatewayIp, rangeStart, rangeEnd string) (string, error) {
 	dhcp := false
 	page, err := subnets.List(client, subnets.ListOpts{
 		Name:       name,
 		Tags:       tag,
 		NetworkID:  netId,
 		CIDR:       cidr,
-		GatewayIP:  *gatewayIp,
+		GatewayIP:  gatewayIp,
 		IPVersion:  4,
 		EnableDHCP: &dhcp,
 	}).AllPages()
@@ -234,9 +234,15 @@ func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId, 
 			Name:       name,
 			NetworkID:  netId,
 			CIDR:       cidr,
-			GatewayIP:  gatewayIp,
+			GatewayIP:  &gatewayIp,
 			IPVersion:  gophercloud.IPv4,
 			EnableDHCP: &dhcp,
+			AllocationPools: []subnets.AllocationPool{
+				{
+					Start: rangeStart,
+					End:   rangeEnd,
+				},
+			},
 		}
 		subnetObj, err := subnets.Create(client, opts).Extract()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -728,9 +734,19 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
 	ipStr := ip.String()
-	log.Printf("Ensuring services subnet with %s CIDR and %s gateway", conf.ServiceNetwork[0], ipStr)
+	splitSvcNet, vrrpNet := iputil.SplitNet(*svcNet)
+	svcPrefixLen, _ := splitSvcNet.Mask.Size()
+	splitSvcCIDR := fmt.Sprintf("%s/%d", splitSvcNet.IP.String(), svcPrefixLen)
+	vrrpNetStart := iputil.FirstUsableIP(vrrpNet).String()
+	vrrpNetEnd := iputil.LastUsableIP(vrrpNet).String()
+
+	log.Printf("Overriding ServiceNetwork[0] to %s", splitSvcCIDR)
+	conf.ServiceNetwork[0] = splitSvcCIDR
```
dulek: I know, ideally we would only change that when saving the Status. Let's just treat this as PoC, we still need to see if that works fine.

pkg/util/ip/addr_test.go
```diff
@@ -169,3 +170,80 @@ func TestFirstUsableIP(t *testing.T) {
...
+	}
+
+	for _, tc := range testcases {
+		_, cidr, err := net.ParseCIDR(tc.cidr)
+		g.Expect(err).NotTo(HaveOccurred())
+		lower, upper := SplitNet(*cidr)
+		o, _ := lower.Mask.Size()
+		g.Expect(fmt.Sprintf("%s/%d", lower.IP.String(), o)).To(Equal(tc.expectedDouble))
+		o, _ = upper.Mask.Size()
+		g.Expect(fmt.Sprintf("%s/%d", upper.IP.String(), o)).To(Equal(tc.expectedOther))
```
danwinship: maybe check that `IterateIP4(lastIP(lower), 1) == upper.IP` and that `lastIP(cidr) == lastIP(upper)` ?

pkg/util/ip/addr_test.go
```diff
@@ -169,3 +170,80 @@ func TestFirstUsableIP(t *testing.T) {
...
+	}
+}
+
+func TestIterateIP4(t *testing.T) {
```
danwinship: you should really have at least one test that crosses a byte boundary. eg, `{ "10.0.0.0", -1, "9.255.255.255" }`

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -728,9 +734,17 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+	svcPrefixLen, _ := splitSvcNet.Mask.Size()
+	splitSvcCIDR := fmt.Sprintf("%s/%d", splitSvcNet.IP.String(), svcPrefixLen)
+	vrrpNetStart := iputil.FirstUsableIP(vrrpNet).String()
+	vrrpNetEnd := iputil.IterateIP4(ip, -1).String() // End pool before the gateway on last usable IP.
+
+	log.Printf("Ensuring services subnet with %s CIDR and %s gateway with allocation pools <%s; %s>",
+		conf.ServiceNetwork[0], ipStr, vrrpNetStart, vrrpNetEnd)
 	svcSubnetId, err := ensureOpenStackSubnet(client, generateName("kuryr-service-subnet", clusterID), tag,
-		svcNetId, conf.ServiceNetwork[0], &ipStr)
+		svcNetId, conf.ServiceNetwork[0], ipStr, vrrpNetStart, vrrpNetEnd)
```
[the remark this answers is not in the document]
dulek: No, but it's worth to explain it in details, I guess. I'll add the comment below here:

```
// Service subnet. This is a bit problematic as OpenStack Octavia will reserve 2 IP's from the service subnet
// for each load balancer - the requested one and a "random" one for VRRP port. Obviously OpenShift will not know
// about the latter and it'll lead to conflicts. To work this around we will create a Neutron subnet with whole
// conf.ServiceNetwork[0] CIDR, but tell OpenShift to use only the lower part and request OpenStack (through
// Neutron's allocation ranges) to only reserve the upper part for "random" port allocations that Octavia will use
// for its VRRP ports.
```

Bottom line here is that yes, the subnet we create in OpenStack needs to be conf.ServiceNetwork[0] and we tell OpenStack Neutron to only use the upper part for random IP allocation, while OpenShift will be told to use the lower part.

pkg/util/ip/addr.go
```diff
@@ -63,3 +77,22 @@ func FirstUsableIP(subnet net.IPNet) net.IP {
 	ip[len(ip)-1] += 1
 	return ip
 }
+
+// This function splits given subnet into two even halves
+func SplitNet(subnet net.IPNet) (net.IPNet, net.IPNet) {
+	ones, _ := subnet.Mask.Size()
+	posByte := uint(ones / 8) // byte in which prefix ends
+	posBit := uint(ones % 8)  // bit in which prefix ends
```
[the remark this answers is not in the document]
dulek: Fair point!

pkg/util/ip/addr_test.go
dulek: Done.

pkg/util/ip/addr_test.go
dulek: Done.

pkg/platform/openstack/kuryr_bootstrap.go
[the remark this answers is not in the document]
dulek: Right, I just see that the initial PR comment had a mistake that you relied on. Fixed.

dulek:
Approach was agreed to be wrong.

PR #245, opened 2019-07-16
Adds Proxy Controller

danehans:
Required by https://github.com/openshift/cluster-network-operator/pull/271

- [x] Add controller to reconcile `Proxy` object.
- [x] Validate `Proxy.spec` fields.
- [x] Write `http|https|no Proxy` spec to status upon successful validation. 

__Note:__ `Proxy` CA reconciliation requirements are being addressed in https://github.com/openshift/cluster-network-operator/pull/271.

Jira: [SDN-492](https://jira.coreos.com/browse/SDN-492) 

PTAL @squeed @openshift/networking


danwinship:
> Should proxy reconciliation result be reflected in network operator status?

Yes, and adding a new `StatusLevel` as you did is correct.

pkg/controller/proxyconfig/proxyconfig_controller.go
```diff
@@ -0,0 +1,99 @@
...
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("proxyconfig-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource config.openshift.io/v1/Proxy
+	err = c.Watch(&source.Kind{Type: &configv1.Proxy{}}, &handler.EnqueueRequestForObject{})
```
squeed: It probably makes sense here to watch for changes to configv1.Network here too, so we can get the `Network.Status` field.

danwinship:
Given the possibility of us having to deal with other abstractly-proxy-related things in the future, I think the naming should be more specific than just "proxy" (eg, in `pkg/proxy/`). Maybe "cluster-proxy" or "operator-proxy"? Or one of those plus "egress". Or cluster-operator-egress-proxy...

And if this is likely to be maintained by the edge team going forward then the directories should get their own `OWNERS` files. (And we should start reorganizing `pkg/network/` too since this would now make two files there that aren't owned by the SDN team.)

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,100 @@
...
+	set := sets.NewString(
+		"127.0.0.1",
+		"localhost",
+		network.Status.ServiceNetwork[0],
+		apiServerURL.Hostname(),
+		internalAPIServer.Hostname(),
+	)
+	platform := infra.Status.PlatformStatus.Type
+
+	// TODO: Does a better way exist to get machineCIDR and controlplane replicas?
```
danehans: @squeed @danwinship PTAL and let me know if you have a better approach for parsing `replicas` and `machineCIDR` from the installer config or using some other resource to get the data.

cmd/cluster-network-operator/main.go
```diff
@@ -108,8 +113,62 @@ func main() {
...
-	// Start the Cmd
-	if err := mgr.Start(signals.SetupSignalHandler()); err != nil {
+	// Create the stop channel and start the operator
+	stop := signals.SetupSignalHandler()
+	if err := start(mgr, stop); err != nil {
 		log.Fatal(err)
 	}
 }
+
+// start creates the default Proxy if it does not exist and then
```
squeed: If the default proxy is null, does it make sense to just add it to `manifests` and have the CVO create it?

squeed: or the cluster-config-operator.

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource config.openshift.io/v1/Proxy
+	err = c.Watch(&source.Kind{Type: &configv1.Proxy{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
```
squeed: I think you'll *also* have to watch for the `network.config.openshift.io` object here to get the CIDRs out of Status. Then, short-circuit reconciling until you've seen both.

squeed: That being said, given that network changes so infrequently, this may not really be necessary.

danehans:
> Given the possibility of us having to deal with other abstractly-proxy-related things in the future, I think the naming should be more specific than just "proxy" (eg, in pkg/proxy/)

@danwinship good call. I'll update the naming based on one of your suggestions.

> And if this is likely to be maintained by the edge team going forward then the directories should get their own OWNERS files. (And we should start reorganizing pkg/network/ too since this would now make two files there that aren't owned by the SDN team.)

@knobunc can you confirm who will own the proxy controller? @danwinship it's my understanding that I'm helping with the proxy controller implementation, but the network edge team is not responsible for maintaining.

cmd/cluster-network-operator/main.go
danehans: > If the default proxy is null, does it make sense to just add it to manifests and have the CVO create it?

It's my understanding that https://github.com/openshift/cluster-config-operator/pull/72 does this. This chunk of code tries to solve the problem of "what if someone deletes the cluster proxy". It's my understanding that this object should always exist, but nothing prevents a user from deleting the resource.

danehans: @squeed I'm going to remove the code from `main.go` since I discovered that the proxy "cluster can not be deleted from the apiserver:
`Error from server (Forbidden): proxies.config.openshift.io "cluster" is forbidden: deleting required proxies.config.openshift.io resource, named cluster, is not allowed`

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,191 @@
...
+func isSpecHTTPProxySet(proxyConfig *configv1.ProxySpec) bool {
+	if len(proxyConfig.HTTPProxy) == 0 {
+		return false
+	}
+
+	return true
+}
+
+// isSpecHTTPSProxySet checks whether spec.httpsProxy of proxy is set.
+func isSpecHTTPSProxySet(proxyConfig *configv1.ProxySpec) bool {
```
ricardomaraschini: Maybe I should not even being looking at this but could we make it `return len(proxyConfig.HTTPSProxy) > 0`

ricardomaraschini: The same applies to other functions below. Feel free to just ignore my comments.

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,336 @@
...
+func validateHTTPReadinessEndpoint(endpoint string) error {
+	if err := validateHTTPReadinessEndpointWithRetries(endpoint, proxyProbeMaxRetries); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateHTTPReadinessEndpointWithRetries tries to validate an http
+// endpoint in a finite loop and returns the last result if it never succeeds.
```
ricardomaraschini: I would say that it returns immediately if it fails, no? If it finds a `404`, for example, it returns immediately. Maybe I am not understanding how this is supposed to work.

danehans: `retries` allows the probe to be run multiple times in case of transient failures. This is a similar approach to pod readiness checks.

ricardomaraschini: Hello @danehans . I would say that this function will return on the first failure, even if transient. This function retries only in case of successes.

ricardomaraschini: I would say that this function will return on the first failure, even if transient. This function retries only in case of successes.

danehans: I see the bug in the logic now, thanks for catching this. I'll push an update.

pkg/controller/proxyconfig/controller.go
danehans: `return len(proxyConfig.HTTPSProxy) > 0` did not provide the behavior I was looking for, so I reverted back in my latest commit.

ricardomaraschini: Weird. I have done this sample below and it seems like the behavior is the same. Could you point me in what cases it does not provide the behavior you expect? https://play.golang.org/p/opLt8dNrOKv

ricardomaraschini: I just came across the commit below, the patch is wrong as it uses `return len(proxyConfig.HTTPProxy) == 0` instead of `return len(proxyConfig.HTTPProxy) > 0`.



https://github.com/openshift/cluster-network-operator/commit/abf6019614e993a26ca20f56dec8b27d8e5157bb#diff-6cc3b6cf53aa0e5d6919e9943b4b174aR139

pkg/controller/proxyconfig/validation.go
ricardomaraschini: It seems like you reverted this one together when you reverted the `return len(proxyConfig.HTTPProxy) > 0` thingie :-)

pkg/controller/proxyconfig/controller.go
danehans: Thanks for catching that. I've updated the PR.

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,127 @@
...
+	var ic installConfig
+	data, ok := cluster.Data["install-config"]
+	if !ok {
+		return "", fmt.Errorf("missing install-config in configmap")
+	}
+	if err := yaml.Unmarshal([]byte(data), &ic); err != nil {
+		return "", fmt.Errorf("invalid install-config: %v\njson:\n%s", err, data)
+	}
+
+	if platform != configv1.VSpherePlatformType &&
```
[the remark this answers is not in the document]
derekwaynecarr: Stupid phone commented on wrong line.  This was meant for block below

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,346 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// validateHTTPSReadinessEndpointWithRetries tries to validate endpoint
+// by using certBundle as trusted CAs to create a TLS connection using a
+// finite loop based on retries, returning an error if it never succeeds.
+func validateHTTPSReadinessEndpointWithRetries(certBundle []*x509.Certificate, endpoint string, retries int) error {
```
ricardomaraschini: The old buggy behavior of `validateHTTPReadinessEndpointWithRetries()` is present here.

pkg/names/names.go
```diff
@@ -31,3 +36,21 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
 // MULTUS_VALIDATING_WEBHOOK is the name of the ValidatingWebhookConfiguration for multus-admission-controller
 // that is used in multus admission controller deployment
 const MULTUS_VALIDATING_WEBHOOK = "multus.openshift.io"
+
+// PROXY_TRUSTED_CA_CONFIGMAP is the name of the proxy.spec.trustedCA
+// ConfigMap that contains the trusted CA certificate bundle used for
+// proxying HTTPS connections.
+const PROXY_TRUSTED_CA_CONFIGMAP = "proxy-ca-bundle"
```
bparees: we want to change this name, right?  trusted-ca-bundle or something?

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,127 @@
...
+		return "", fmt.Errorf("failed to parse install config replicas: %v", err)
+	}
+
+	for i := int64(0); i < int64(replicas); i++ {
+		etcdHost := fmt.Sprintf("etcd-%d.%s", i, infra.Status.EtcdDiscoveryDomain)
+		set.Insert(etcdHost)
+	}
+
+	for _, clusterNetwork := range network.Status.ClusterNetwork {
+		set.Insert(clusterNetwork.CIDR)
```
bparees: I thought someone had indicated we needed every host listed because there wasn't a definite host CIDR?

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,346 @@
...
+				return fmt.Errorf("invalid URI for endpoint %s: %v", endpoint, err)
+			}
+			switch {
+			case scheme == proxyHTTPScheme:
+				if err := validateHTTPReadinessEndpoint(endpoint); err != nil {
+					return fmt.Errorf("readinessEndpoint probe failed for endpoint %s", endpoint)
+				}
+			case scheme == proxyHTTPSScheme:
+				if !specTrustedCASet(proxyConfig) {
+					return fmt.Errorf("readinessEndpoint with an %q scheme requires trustedCA to be set", proxyHTTPSScheme)
```
bparees: i don't think using an https proxy necessitates providing a trustedCA.  it's possible (perhaps unlikely) that your proxy is using a cert that's signed by a system-trust chain.

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,346 @@
...
+				}
+				if err := validateHTTPSReadinessEndpoint(readinessCerts, endpoint); err != nil {
```
bparees: does readinessCerts include the system certs we're going to be augmenting with the user-provided certs?  i think it should.

(and if the user provided no certs, then it would just be the base system certs)

pkg/controller/proxyconfig/status.go
danehans: @bparees I recall the initial discussion around dynamically managing based on host ip, but then we settled on using the machineCIDR, @squeed please confirm.

pkg/controller/proxyconfig/validation.go
danehans: Good point. I'll move the trusted ca check into the `isSpecTrustedCASet` block.

pkg/controller/proxyconfig/validation.go
danehans: I am renaming `readinessCerts` to `trustedCAcerts`. This var contains the list of validated certs from `trustedCA`. The validation ensures the configmap reference is the merged user/system bundle name and the list contains pem encoded certs.

danehans: > and if the user provided no certs, then it would just be the base system certs

I need to update the cert publishing controller to meet this requirement.

pkg/controller/proxyconfig/validation.go
danehans: ack

pkg/names/names.go
danehans: I'll update, I thought operators were already writing code based on the "proxy-ca-bundle" name.

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,127 @@
...
+	type installConfig struct {
+		ControlPlane struct {
+			Replicas string `json:"replicas"`
+		} `json:"controlPlane"`
+		Networking struct {
+			MachineCIDR string `json:"machineCIDR"`
+		} `json:"networking"`
+	}
+	var ic installConfig
+	data, ok := cluster.Data["install-config"]
```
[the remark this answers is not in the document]
danehans: Done

pkg/controller/proxyconfig/status.go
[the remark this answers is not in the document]
danehans: Done

pkg/controller/proxyconfig/validation.go
danehans: @bparees I'm thinking through the implementation for your above comment. What would be the reason for providing a trustedCA to a proxy if not used for proxying https connections?

bparees: well what i was stating was the opposite (one might use https proxy and not provide a CA).

but as for why one might provide a CA but not use https..   for proxy, no reason i can think of, but since we're discussing extending this to be a generic "provide CAs to components" feature, it's easy to envision that someone might have CAs and no https proxy, in that case.

pkg/controller/proxyconfig/validation.go
danehans: https://github.com/openshift/cluster-network-operator/pull/271 has been updated to a) build off this PR and b) to support "if the user provided no certs, then it would just be the base system certs" requirement.

pkg/controller/proxyconfig/validation.go
danehans: > one might use https proxy and not provide a CA

@bparees got it. I added the support for this requirement here: https://github.com/openshift/cluster-network-operator/pull/271/files#diff-1502a5b3db6bd911880e1d0d9346b561R47-R59

pkg/names/names.go
danehans: I updated the name to `trusted-ca-bundle` here and the proxy workflow gdoc.

pkg/controller/proxyconfig/status.go
squeed: Yeah, I also prefer using MachineCIDR. The alternative - updating for every host - adds a lot of churn, and misses some other potential IPs such as AWS internal load balancers.

pkg/controller/proxyconfig/validation.go
danehans: > does readinessCerts include the system certs we're going to be augmenting with the user-provided certs?

@bparees yes, if add'l certs are provided from the installer or creating configmap `openshift-config-managed/user-ca-bundle`. Otherwise it's a list of certs from the system trust bundle. __Note:__ `readinessCerts ` has been renamed to `certBundle` here: https://github.com/openshift/cluster-network-operator/pull/271/files#diff-1502a5b3db6bd911880e1d0d9346b561R74

pkg/controller/proxyconfig/status.go
bparees: sure, i prefer CIDR as well, as long as we are confident it is sufficient.

pkg/controller/proxyconfig/validation.go
bparees: sounds good.

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,126 @@
...
+	if !ok {
+		return "", fmt.Errorf("missing install-config in configmap")
+	}
+	if err := yaml.Unmarshal([]byte(data), &ic); err != nil {
+		return "", fmt.Errorf("invalid install-config: %v\njson:\n%s", err, data)
+	}
+
+	switch infra.Status.PlatformStatus.Type {
+	case configv1.AWSPlatformType:
```
squeed: This is the case for gcp and azure, too.

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,126 @@
...
+		set.Insert("169.254.169.254", ic.Networking.MachineCIDR)
+	default:
+		return "", fmt.Errorf("unsupported infrastructure provider: %s", infra.Status.PlatformStatus.Type)
```
squeed: am I misreading this, or does this line mean we *only* support aws? If so, that seems... wrong?

pkg/controller/proxyconfig/status.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+
+	return nil
+}
+
+// mergeUserSystemNoProxy merges user-supplied noProxy settings from proxy
+// with cluster-wide noProxy settings, returning a merged, comma-separated
+// string of noProxy settings.
+func mergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network, cluster *corev1.ConfigMap) (string, error) {
```
squeed: This is a lot of code. I'd like to see some tests added for this. (In a follow-up PR is fine).

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,225 @@
...
+				log.Println("proxy not found; reconciliation will be skipped", "request", request)
+				return reconcile.Result{}, nil
+			}
+			// Error reading the object - requeue the request.
+			return reconcile.Result{}, fmt.Errorf("failed to get proxy %q: %v", request, err)
+		}
+
+		// A nil proxy is generated by upgrades and installs not requiring a proxy.
+		validate := true
+		if !isSpecHTTPProxySet(&proxyConfig.Spec) && !isSpecHTTPSProxySet(&proxyConfig.Spec) {
```
bparees: should also validate if noproxy is set (even if nothing else is set)

pkg/controller/proxyconfig/status.go
danehans: @squeed I paired back the PR and I'm working on tests. Since quite a few groups are waiting on the PR and I successfully tested the code against a live cluster, I was hoping I can submit a follow-on PR to add test.

squeed: That seems fine.

pkg/controller/proxyconfig/controller.go
squeed: Yeah, @danehans, you probably want to add a Watch for all of the dependent objects you retrieve starting in line 101

danehans: I'm marking this resolved. We can address this issue in the future if needed.

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,53 @@
...
+const (
+	proxyHTTPScheme = "http"
+	// noProxyWildcard is the string used to as a wildcard attached to a
+	// domain suffix in proxy.spec.noProxy to bypass proxy for httpProxy
+	// and httpsProxy.
+	noProxyWildcard = "*"
+)
+
+// ValidateProxyConfig ensures that proxyConfig is valid.
+func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySpec) error {
```
squeed: (possible follow-up todo): Do we want to check if the specified proxy is reachable?

pkg/controller/proxyconfig/status.go
danehans: I was unsure, so I just took a conservative approach. I'm adding GCP and Azure platform types.

pkg/controller/proxyconfig/status.go
danehans: Yes, see https://github.com/openshift/cluster-network-operator/pull/245#discussion_r311647398

pkg/controller/proxyconfig/validation.go
danehans: @squeed prior to my last commit (`efca2437f3b756f2d92613bc15221db2755c5306`), this PR included support for `readinesCheck` which accomplished your ask. However, it was decided to have this PR simply validate the core `http/https/no proxy` proxy fields and write to status. This PR will provide a functional proxy environment so other groups are not blocked by this work. I'll provide a follow-on PR shortly that includes `readinesCheck` support.

squeed: :+1:

pkg/controller/proxyconfig/status.go
squeed: I'm still a bit confused, then - won't this cause the network operator to be Degraded for all platforms except AWS?

danehans: Yes, but I have a PR lined up to add GCP and Azure.

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, netConfig); err != nil {
+		log.Printf("failed to get network config 'cluster': %v", err)
+		return reconcile.Result{}, err
+	}
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: "cluster-config-v1", Namespace: "kube-system"},
+		clusterConfig); err != nil {
+		log.Printf("failed to get configmap 'kube-system/cluster-config-v1': %v", err)
+		return reconcile.Result{}, err
+	}
```
bparees: should these 3 failure cases set the degraded status also?  they all mean we effectively ignored the proxy spec

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,52 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid httpProxy URI: %v", err)
+		}
+		if scheme != proxyHTTPScheme {
+			return fmt.Errorf("httpProxy requires an %q URI scheme", proxyHTTPScheme)
+		}
+	}
+
+	if isSpecHTTPSProxySet(proxyConfig) {
+		_, err := validation.URI(proxyConfig.HTTPSProxy)
```
bparees: should still confirm the scheme is at least `http` or `https`?

pkg/controller/proxyconfig/controller.go
danehans: @bparees yes, let me update the PR. As with `r.ValidateProxyConfig()` returning `err != nil`, I'm also going to update each call to `return reconcile.Result{}, nil` upon `err != nil`.

pkg/controller/proxyconfig/validation.go
danehans: ack

danehans:
@squeed @bparees I addressed booth of your comments in my latest commit, PTAL.

pkg/controller/proxyconfig/validation.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+	}
+
+	if isSpecHTTPSProxySet(proxyConfig) {
+		scheme, err := validation.URI(proxyConfig.HTTPSProxy)
+		if err != nil {
+			return fmt.Errorf("invalid httpsProxy URI: %v", err)
+		}
+		if scheme != proxyHTTPScheme && scheme != proxyHTTPSScheme {
+			return fmt.Errorf("httpsProxy requires a %q or %s URI scheme", proxyHTTPScheme, proxyHTTPSScheme)
```
bparees: why is one of these %q and the other %s?

squeed:
/approve
/lgtm

Please take a note of the follow-up issues.

PR #246, opened 2019-07-16
Kuryr: Sync LB members with Kubernetes endpoints

MaysaMacedo:
The API LB members should match the default/kubernetes endpoints.
This commit adds support to sync the members with the endpoints
when a reconciliation occurs.


MaysaMacedo:
We still need to verify this.

MaysaMacedo:
> We still need to verify this.

verified.

PR #249, opened 2019-07-17
Kuryr: Set large data timeouts on the API LB

dulek:
Since Rocky Octavia supports setting timeouts for periods without data
passing through connections. This patch makes use of that we use that to
increase the timeouts to 1 hour for the OpenShift API LB. The default of
50 seconds is unacceptable for an LB used for watching the OpenShift
API.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -48,6 +48,7 @@ const (
 	// NOTE(dulek): This one is hardcoded in openshift/installer.
 	InfrastructureCRDName           = "cluster"
 	MinOctaviaVersionWithTagSupport = "v2.5"
+	MinOctaviaVersionWithTimeouts   = "v2.1"
 	KubernetesEndpointsName         = "kubernetes"
 	KubernetesEndpointsNamespace    = "default"
 )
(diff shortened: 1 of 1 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -623,6 +624,19 @@ func ensureOpenStackLbListener(client *gophercloud.ServiceClient, name, lbId, po
...
 			LoadbalancerID: lbId,
 		}
+
+		// NOTE(dulek): If Octavia supports setting data timeouts in listeners (Rocky+) we set them to 1 hour as this
+		//              LB will be used for watching the Kubernetes API, that shouldn't time out after the default 50 seconds.
+		timeoutSupport, err := IsOctaviaVersionSupported(client, MinOctaviaVersionWithTimeouts)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to determine if Octavia supports listener timeouts API")
+		}
+		timeout := 3600000
```
squeed: Probably unnecessarily high, as the k8s client scaffolding always enables tcp keepalives with an interval of (I believe) 30 seconds. However, if tcp keepalives aren't enough to hold the connection open, then disregard this.

dulek: @squeed: Thanks, I'm really glad you pay attention to what I'm adding as PR's, it's really helpful as I don't know all the OpenShift/SDN quirks.

So this refers to [this HAProxy option](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4-timeout%20server) and it seems it's only for data activity, not tcp keepalives. The default it 50 seconds and we're sure we experienced being disconnected with that value in place.

squeed: Gotcha, makes sense. We spent a bunch of time chasing down timeouts and keepalives in the aws load balancers, so this is somewhat fresh in my memory (and surprising how quickly a cluster goes south when the details are missed).

I agree - 50 seconds is much too long. I think this value is fine.

luis5tb:
just a simple rebase

PR #250, opened 2019-07-17
dep ensure -update github.com/openshift/api

s1061123:
Also changes "operv1.ProxyArgumentList" to fix build error


s1061123:
thank you for the review, @danwinship!

squeed:
Hmm, not sure what's up with CI.
/refresh

s1061123:
We get the result for e2e-aws and upgrade may take 2h to complete (from my last result).
Let's see the result...

PR #251, opened 2019-07-17
openshift-sdn: give sdn-controller the ability to patch nodes.

squeed:
We need this to remove the NetworkUnavailable status that the GCP cloud-provider always sets. This is because it requires routes to be programmed, but we have it configured to not actually program routes.

### DIFF
--- a/bindata/network/openshift-sdn/003-rbac-controller.yaml
+++ b/bindata/network/openshift-sdn/003-rbac-controller.yaml
@@ -39,6 +39,12 @@ rules:
   - list
   - get
   - watch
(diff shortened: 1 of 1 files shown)


squeed:
cc @abhinavdahiya

danwinship:
We will need similar gratuitous-GCP-unbreaking code in ovn-kubernetes won't we? Do we have a card for that?

squeed:
> We will need similar gratuitous-GCP-unbreaking code in ovn-kubernetes won't we? Do we have a card for that?

Yup, good point. Created SDN-490.

PR #254, opened 2019-07-18
Adding build support for archs other than amd64

Pensu:
This modifies the Dockerfile to add the build support for archs other than amd64, based on the commit: https://github.com/openshift/cluster-network-operator/commit/1c9828fdcacec9240aff3a402429df4c3e1dd2ef

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,14 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-network-operator
 COPY . .
-RUN hack/build-go.sh
+RUN hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/cluster-network-operator/_output/linux/$(go env GOARCH)/cluster-network-operator /tmp/build/; \
+    cp /go/src/github.com/openshift/cluster-network-operator/_output/linux/$(go env GOARCH)/cluster-network-renderer /tmp/build/
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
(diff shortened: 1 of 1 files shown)


danwinship:
/assign @squeed 
So I guess the original Dockerfiles were just copied from something random and didn't work at all, and the RHEL one got fixed at the point when we needed to actually start building it (#131) but the non-RHEL one never did?

Dockerfile
```diff
@@ -1,11 +1,14 @@
...
 WORKDIR /go/src/github.com/openshift/cluster-network-operator
 COPY . .
-RUN hack/build-go.sh
+RUN hack/build-go.sh; \
+    mkdir -p /tmp/build; \
+    cp /go/src/github.com/openshift/cluster-network-operator/_output/linux/$(go env GOARCH)/cluster-network-operator /tmp/build/; \
+    cp /go/src/github.com/openshift/cluster-network-operator/_output/linux/$(go env GOARCH)/cluster-network-renderer /tmp/build/
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
-COPY --from=builder /go/src/github.com/openshift/cluster-network-operator/_output/linux/amd64/cluster-network-operator /usr/bin/
```
dcbw: Why not just use $(go env GOARCH) here? Does /tmp/build add anything?

Pensu: @dcbw Yeah, but that would require `golang` to be present in the base i.e. registry.svc.ci.openshift.org/openshift/origin-v4.0:base image.

Pensu:
> /assign @squeed
> So I guess the original Dockerfiles were just copied from something random and didn't work at all, and the RHEL one got fixed at the point when we needed to actually start building it (#131) but the non-RHEL one never did?

Yeah, looks like it. The RHEL one works fine, only the non-rhel one has amd64 hardcoded into it.

Pensu:
@danwinship Hi, looks like I don't have the permission to rerun e2e-aws test. It looks like a flake, can you please rerun this? Thanks.

PR #177, opened 2019-05-24
Macvlan interface support in network-operator

s1061123:
This change introduces macvlan interface type in NetworkType and allows network-operator
to configure macvlan without to use RawCNIConfig.

Depends on https://github.com/openshift/api/pull/321

@dcbw @squeed @dougbtv @danwinship

PR #255, opened 2019-07-18
SimpleMacvlan support

s1061123:


danwinship:
does this obsolete #177 ?

fepan:
> does this obsolete #177 ?

I think we still might pursue generic maclvan later, which is what #177 intended to be. But for now we just want to call this "simple macvlan" to indicate a basic level of config.

pkg/network/additional_networks.go
```diff
@@ -58,3 +62,140 @@ func validateRaw(conf *operv1.AdditionalNetworkDefinition) []error {
...
+	return objs, nil
+}
+
+// validateIPAMConfig checks its IPAMConfig.
+func validateIPAMConfig(conf *operv1.IPAMConfig) []error {
+	out := []error{}
+
+	switch conf.Type {
+	case operv1.IPAMTypeDHCP:
+	case operv1.IPAMTypeStatic:
```
squeed: Surely there is some validation you can do here? Check to see if supplied routes are valid IP addresses?

pkg/network/additional_networks_test.go
```diff
@@ -65,3 +115,71 @@ func TestValidateRaw(t *testing.T) {
...
 }
+
+func TestRenderSimpleMacvlanConfig(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	for _, cfg := range NetworkAttachmentConfigSimpleMacvlan.Spec.AdditionalNetworks {
+		objs, err := renderSimpleMacvlanConfig(&cfg, manifestDir)
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(objs).To(HaveLen(1))
+		g.Expect(objs).To(
```
squeed: I'd like to see a test that validates the NetworkAttachmentDefinition content. You can use the "MatchYAML" and MatchJSON functions of gomega, if you like.

pkg/network/dhcp_daemon.go
```diff
@@ -6,7 +6,59 @@ import (
 	"log"
 )
 
-// UseDHCP determines if the the DHCP CNI plugin running as a daemon should be rendered.
+// UseDHCPRaw determines if the the DHCP CNI plugin running as a daemon should be rendered.
+func UseDHCPRaw(addnet *operv1.AdditionalNetworkDefinition) bool {
```
squeed: Hmm. Do we need to look in to RawCNI like this? Is this to support 4.1 clusters?

squeed: Nevermind, I see this is pulled from another location.

squeed:
A few small things, otherwise this looks basically ready to go. Thanks!

pkg/network/additional_networks_test.go
s1061123: added.

pkg/network/additional_networks.go
s1061123: added.

squeed:
/lgtm
/approve
A few small things that can be fixed in a followup PR

bindata/network/additional-networks/simplemacvlan/simplemacvlan.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: {{.AdditionalNetworkName}}
+  namespace: default
+spec:
+  config: '{
+    "cniVersion": "0.3.0",
```
squeed: This should be v0.3.1, but they're actually identical... long story. This can be fixed later.

squeed:
Please make sure you file a PR updating README.md with this new configuration.

PR #260, opened 2019-07-23
Fix Watch arguments in configmap operator

s1061123:
EnqueueRequestForOwner{} will enqueue the owner of the target,
hence it enqueue configmap owner, network CR. This change fixes
reconciler function to match it.

### DIFF
--- a/pkg/controller/operconfig/configmap_controller.go
+++ b/pkg/controller/operconfig/configmap_controller.go
@@ -46,7 +46,7 @@ func addConfigMapReconciler(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


dougbtv:
Related to this BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1731392

squeed:
Interesting! Apparently EnqueueRequestForOwner doesn't do what I thought it does. I'm chatting quickly now with the operator-sdk team to see if there's a way to avoid watching all ConfigMaps across all namespaces.

squeed:
OK, one small change needed: You also need to set `IsController: true` on the `handler.EnqueueRequestForOwner{}` parameter. Otherwise, this is fine.

dougbtv:
@squeed thanks for the double check with the operator-sdk team! appreciate it. I've got a new fix up with the additional parameter `IsController: true` on the `handler.EnqueueRequestForOwner{}`

PR #262, opened 2019-07-24
Followup #255 comments

s1061123:
This change contains the following fix:
 + updating README.md with this new configuration.
 + fix macvlan config "cniVersion" to 0.3.1 and its namespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,151 @@ spec:
      iptables-min-sync-period: ["30s"]
 ```
 
+## Configuring Additional Networks
+Users can configure additional networks, based on [Kubernetes Network Plumbing Working Group's Kubernetes Network Custom Resource Definition De-facto Standard Version 1](https://github.com/K8sNetworkPlumbingWG/multi-net-spec/blob/master/%5Bv1%5D%20Kubernetes%20Network%20Custom%20Resource%20Definition%20De-facto%20Standard.md).
+
+* `name`: name of network attachment definition, required
+* `namespace`: namespace for the network attachment definition. The default is default namespace
+* `type`: specify network attachment definition type, required
+
+Currently, the understood values for type are:
+* `Raw`
+* `SimpleMacvlan`
+
+```yaml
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -193,6 +193,115 @@ spec:
...
+  name: test-network-1
+  namespace: namespace-test-1
+  ownerReferences:
+  - apiVersion: operator.openshift.io/v1
+    # (snip)
+spec:
+  config: '{ "cniVersion": "0.3.1", "type": "macvlan", "master": "eth1", "mode": "bridge", "ipam": { "type": "dhcp" } }'
+```
+
+### Configuring SimpleMacvlan
```
squeed: It would be good to explain briefly what macvlan does, how it works, and why someone might use it. It's also important to point out the security considerations.

README.md
```diff
@@ -193,6 +193,115 @@ spec:
...
+    simpleMacvlanConfig:
+      master: eth0
+      mode: bridge
+      mtu: 1515
+      ipamConfig:
+        type: dhcp
+```
+
+#### Configuring Static IPAM
+Users can configure static IPAM with following parameters:
```
squeed: I would also mention that the IP address can come from an annotation on the pod, and how to configure that.

README.md
s1061123: Added.

README.md
s1061123: Added.

README.md
```diff
@@ -193,6 +193,151 @@ spec:
...
+
+* `addresses`:
+  * `address`: Address is the IP address in CIDR format, required
```
squeed: Is the address required? I thought you could set that via an annotation?

s1061123: fixed.

squeed:
/approve
/lgtm

I wouldn't mind a bit more detail in the README, but this looks good. If you file a PR with just README changes, it skips CI :-).

PR #263, opened 2019-07-25
Add tolerations permissions

ravisantoshgudimetla:
As of now, tolerations to pod spec can be added by any user and the tolerations sometimes helps
in landing on master nodes. Using https://github.com/openshift/origin/pull/23427, we're moving to
a RBAC model to ensure that tolerations can be applied to a pod by service account with privileges to access them

### DIFF
--- a/bindata/network/openshift-sdn/002-rbac.yaml
+++ b/bindata/network/openshift-sdn/002-rbac.yaml
@@ -54,6 +54,19 @@ rules:
   - patch
   - update
 
+- apiGroups:
+  - "toleration.scheduling.openshift.io"
+  resources:
+  - node-role.kubernetes.io/master
+  - node.kubernetes.io/unreachable
+  - node.kubernetes.io/not-ready
+  resourceNames:
+  - "NoSchedule:"
+  - "NoExecute:"  
+  verbs:
+  - exists
+  - equal
+    
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/bindata/network/openshift-sdn/003-rbac-controller.yaml
+++ b/bindata/network/openshift-sdn/003-rbac-controller.yaml
@@ -60,6 +60,19 @@ subjects:
   name: sdn-controller
   namespace: openshift-sdn
 
+- apiGroups:
+  - "toleration.scheduling.openshift.io"
+  resources:
+  - node-role.kubernetes.io/master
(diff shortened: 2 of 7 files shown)


bindata/network/openshift-sdn/003-rbac-controller.yaml
```diff
@@ -60,6 +60,19 @@ subjects:
   name: sdn-controller
   namespace: openshift-sdn
 
+- apiGroups:
```
squeed: Mis-paste - this needs to be in a ClusterRole, not a ClusterRoleBinding.

bindata/network/openshift-sdn/002-rbac.yaml
```diff
@@ -54,6 +54,19 @@ rules:
   - patch
   - update
 
+- apiGroups:
+  - "toleration.scheduling.openshift.io"
+  resources:
+  - node-role.kubernetes.io/master
+  - node.kubernetes.io/unreachable
+  - node.kubernetes.io/not-ready
```
danwinship: So does this only add permission for those specific tolerations? Network plugins need to tolerate all possible taints (you don't want the network to go down because the node is low on disk space, etc), so the roles in `bindata/network/kuryr/001-rbac.yaml`, `bindata/network/multus/002-rbac.yaml`, `bindata/network/openshift-sdn/002-rbac.yaml`, and `bindata/network/ovn-kubernetes/002-rbac.yaml` need to reflect that.

Also, you need to add permissions (for all tolerations) to `bindata/kube-proxy/001-rbac.yaml`, which was probably added to the repo after you started working on this patch.

ravisantoshgudimetla: Ok, let me look into that. Thanks :)

bindata/network/openshift-sdn/003-rbac-controller.yaml
ravisantoshgudimetla: Oops. It is.

bindata/network/openshift-sdn/002-rbac.yaml
ravisantoshgudimetla: @danwinship - Yes, this add permissions for those tolerations only. This PR was just to give everyone an idea. I'll make more changes depending on the existing tolerations for any DS or deployment.yaml.

PR #264, opened 2019-07-25
Add sg rule allowing traffic between pods

MaysaMacedo:
By default pods under the same subnet should be able to
talk with each other, this commit adds a new security group
rule allowing this traffic.


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -854,6 +854,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to add rule opening traffic from workers and masters")
 	}
+	log.Print("Allowing traffic from pods in same subnet")
```
dulek: The comment is misleading - this is actually allowing all pod-pod traffic, which is fine. Note that all the subnets are connected through a router.

PR #266, opened 2019-07-25
update kube-proxy cmd

sallyom:
After this merges: https://github.com/openshift/sdn/pull/14 the kube-proxy cmd will be standalone rather than embedded in hyperkube, so need to update the DaemonSet.  

/cc @danwinship

### DIFF
--- a/bindata/kube-proxy/kube-proxy.yaml
+++ b/bindata/kube-proxy/kube-proxy.yaml
@@ -40,7 +40,7 @@ spec:
         command:
         - /bin/bash
         - -c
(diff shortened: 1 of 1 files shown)


danwinship:
@squeed nothing in CI is currently actually using the kube-proxy image, right, so it's fine if we possibly break it for 24 hours or so? (To do this "right" we'd have to first update the DaemonSet to try both `hyperkube kube-proxy` and `kube-proxy`, then update the image, then update the DaemonSet again to remove the `hyperkube kube-proxy` support. But if no one will notice if we cheat, then let's just cheat.)

squeed:
> nothing in CI is currently actually using the kube-proxy image, right, so it's fine if we possibly break it for 24 hours or so?

That's right. Cheat away.

PR #237, opened 2019-07-10
Condense CNI reference plugins images into a single image

dougbtv:
This condenses the images for the container networking CNI reference plugins into a single image instead of having them in two images (with -supported & -unsupported suffixes).

I had ART update their build instructions, and their changes are available here @ https://gitlab.cee.redhat.com/openshift-art/ocp-build-data/merge_requests/172/diffs

Additionally, I have a pending PR for updating image mirroring in openshift/release here @ https://github.com/openshift/release/pull/4334

I believe this PR should be in a hold state until the openshift/release PR gets merged.

Also, quite happy to take any input on other changes should I have missed any, I tried to grep through and find any references to the -supported/-unsupported images to have them updated, but, any eyeballs are appreciated for sure.

PR #267, opened 2019-07-30
Adapt to merger of CNI Plugins images

yselkowitz:
In 4.2, the split between container-networking-plugins-{,un}supported
was removed, with a single container-networking-plugins image taking
their place.


danwinship:
This is a duplicate of #237, but last we knew the build config was broken so the change wouldn't work anyway?

PR #271, opened 2019-08-02
Adds Trust Bundle Publishing to Proxy Controller

danehans:
Depends on https://github.com/openshift/cluster-network-operator/pull/245

- [x] Reconcile "openshift-config-managed/trusted-ca-bundle" `ConfigMap` object.
- [x] Validate `Data` of "openshift-config-managed/user-ca-bundle" `ConfigMap` object.
- [x] Validate cluster-network-operator's system trust bundle as valid certificate(s).
- [x] Merge user/system trust bundles.
- [x] Create/update "openshift-config-managed/trusted-ca-bundle" `ConfigMap` object.
- [x] Godocs

Jira: [SDN-501](https://jira.coreos.com/browse/SDN-501) 

PTAL @squeed @danwinship @dcbw @bparees @JacobTanenbaum


pkg/controller/trustbundle-publisher/controller.go
```diff
@@ -0,0 +1,206 @@
+// The trustbundle-publisher controller is responsible for merging a user
+// provided trust bundle with the system trust bundle and publishing the
+// merged bundle to a configmap in the "openshift-config-managed" namespace.
```
bparees: ultimately this is going to be folded into the proxy controller right?  (so that we only perform this merge/publication after confirming we can pass proxy validation with these CAs)

pkg/controller/trustbundle-publisher/controller.go
```diff
@@ -0,0 +1,206 @@
...
+	log.Printf("Completed reconciling %s/%s\n", request.Namespace, request.Name)
+	return reconcile.Result{}, nil
+}
+
+// ensureMergedConfigMap merges the additionalData with systemData into
+// a single byte slice, ensures the merged byte slice contains valid PEM
+// encoded certificates, embeds the merged byte slice into a configmap
+// named "proxy-ca-bundle" in namespace "openshift-config-managed" and
+// returns the configmap.
+func (r *ConfigMapReconciler) ensureMergedConfigMap(additionalData, systemData []byte) (*corev1.ConfigMap, error) {
```
squeed: Is it a problem if the additional CA is already present in the system trust? Will that break expectations?

bparees: it's just going to result in 2 copies of that cert in the bundle, right?

pkg/controller/trustbundle-publisher/controller.go
danehans: @bparees this controller will work with the proxy controller by watching/validating the user-provided add'l trust bundle configmap, merging the bundle with the system bundle and creating the combined bundle configmap that is consumed by the proxy controller. The proxy controller will use the combined bundle in the configmap to validate https readinessEndpoints and for https proxying by validating the bundle referenced in trustedCA of a proxy object.

bparees: that still makes it sound like the combined bundle is published before it's been validated as working/usable.

meaning an admin could update the bundle to contain invalid content, which would be propagated to the downstream consumers and break proxy usage.

pkg/controller/trustbundle-publisher/controller.go
squeed: Yup. The question is: does that break anything? Does the bundle need to be unique? (I have no idea)

bparees: i'd be surprised but it would definitely be a good thing to test.

pkg/controller/trustbundle-publisher/controller.go
danehans: @bparees https://github.com/openshift/cluster-network-operator/pull/271 validates the user provided trust bundle, merges with the system bundle to form the `trusted-ca-bundle` configmap. If a user-provided bundle does not exist, https://github.com/openshift/cluster-network-operator/pull/271/files#diff-6cc3b6cf53aa0e5d6919e9943b4b174aR151 creates a system bundle for the `trusted-ca-bundle` configmap to use when `proxy.spec.httpsProxy` is defined.

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,361 @@
...
+				e.Meta.GetNamespace() == names.TRUST_BUNDLE_CONFIGMAP_NS
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return e.Meta.GetName() == names.ADDL_TRUST_BUNDLE_CONFIGMAP &&
+				e.Meta.GetNamespace() == names.TRUST_BUNDLE_CONFIGMAP_NS
+		},
+	}
+
+	// Watch for changes to the trust bundle configmap.
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, pred)
```
bparees: you also need to watch the configmap pointed to by the proxyconfig resource, is that being done somewhere?

pkg/controller/proxyconfig/controller.go
```diff
@@ -0,0 +1,361 @@
...
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver.
+	client client.Client
+	scheme *runtime.Scheme
+	status *statusmanager.StatusManager
+}
+
+// Reconcile expects request to refer to a proxy object named "cluster"
+// in the default namespace or to a configmap object named
+// "user-ca-bundle" in namespace "openshift-config-managed", and will
```
bparees: this comment is wrong...   i'd expect this controller to be watching 3 things:

1) the trust-bundle configmap in openshift-config-managed  (this CM has a fixed name), because you need to stomp it back to the correct value if it gets updated by something else
2) the additional CA bundle configmap in the openshift-config namespace (this configmap name can change, so you need to detect if it changes, by watching the proxy config resource), because if the bundle changes, you need to re-merge the values w/ the system values and publish it to the CM in (1).
3) the proxyconfig resource (so you can determine the name of the configmap in (2) and perform the merge action on it)

manifests/0000_70_cluster-network-operator_04_deployment.yaml
```diff
@@ -62,6 +65,12 @@ spec:
         - name: host-kubeconfig
           hostPath:
             path: /etc/kubernetes/kubeconfig
+        - name: trusted-ca
+          configMap:
+            name: trusted-ca
+            optional: true
+            items:
+            - key: ca-bundle.crt
```
bparees: i don't think this belongs in this PR.  This PR should be constrained to the job of taking the user provided CAs, merging them with the system CAs, and publishing that merged content into the openshift-config-managed configmap.

it should not delve into unrelated aspects like ensuring the network operator is mounting/consuming the proxy CA bundle

pkg/controller/proxyconfig/controller.go
```diff
@@ -57,6 +75,12 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	return nil
 }
 
+// handleConfigMap returns true if meta name is "user-ca-bundle" and
+// meta namespace is "openshift-config-managed".
```
bparees: comment looks wrong.  the configmap name in "openshift-config-managed" is `trusted-ca-bundle`

pkg/controller/proxyconfig/controller.go
```diff
@@ -67,72 +91,140 @@ type ReconcileProxyConfig struct {
 }
 
 // Reconcile expects request to refer to a proxy object named "cluster"
-// in the default namespace and will ensure the proxy object is in the
-// desired state.
+// in the default namespace or a configmap object named "user-ca-bundle"
+// in namespace "openshift-config-managed", and will ensure either
+// object is in the desired state.
```
bparees: this comment is wrong... i'd expect this controller to be watching 3 things:

    the trust-bundle configmap in openshift-config-managed (this CM has a fixed name), because you need to stomp it back to the correct value if it gets updated by something else
    the additional CA bundle configmap in the openshift-config namespace (this configmap name can change, so you need to detect if it changes, by watching the proxy config resource), because if the bundle changes, you need to re-merge the values w/ the system values and publish it to the CM in (1).
    the proxyconfig resource (so you can determine the name of the configmap in (2) and perform the merge action on it)

pkg/controller/proxyconfig/controller.go
```diff
@@ -154,3 +246,116 @@ func isSpecHTTPSProxySet(proxyConfig *configv1.ProxySpec) bool {
...
+	}
+	if _, err := r.validateTrustBundleConfigMap(mergedCfgMap); err != nil {
+		return nil, fmt.Errorf("failed to validate merged configmap %s/%s: %v", mergedCfgMap.Namespace,
+			mergedCfgMap.Name, err)
+	}
+
+	return mergedCfgMap, nil
+}
+
+// syncTrustedCABundleConfigMap checks if mergedConfigMap named "proxy-ca-bundle"
```
bparees: `trusted-ca-bundle` ?

manifests/0000_70_cluster-network-operator_04_deployment.yaml
danehans: ack, I'll move the operator ca mounting to a separate PR.

pkg/controller/proxyconfig/controller.go
danehans: @bparees I'm confused now. I thought `proxy.spec.trustedCA` points to the user-provided configmap which resides in ns `openshift-config`.The controller merges the user-provided ca bundle with the system  contrbundle and spits out the merged bundle into configmap 'trusted-ca'bundle' in ns 'openshift-config-managed'.

bparees: > spits out the merged bundle into configmap 'trusted-ca'bundle' in ns 'openshift-config-managed'.

right.  Hence my statement of:

> the configmap name in "openshift-config-managed" is trusted-ca-bundle

vs your doc which says:

> handleConfigMap returns true if meta name is "user-ca-bundle" and meta namespace is "openshift-config-managed".

pkg/controller/proxyconfig/controller.go
danehans: @bparees `pred` along with `handleConfigMap()` is watching for user-provided configmaps from `openshift-config` ns. The controller merges any a user-provided trust bundle with the system bundle and publishes the merged bundle to `trusted-ca-bundle` configmap in ns `openshift-config-managed`. If `proxy.spec.trustedCA` does not refer to a configmap but `proxy.spec.httpsProxy` is set, then the controller will generate `trsuted-ca-bundle` configmap using the system bundle.

bparees: >  If proxy.spec.trustedCA does not refer to a configmap but proxy.spec.httpsProxy is set, then the controller will generate trsuted-ca-bundle configmap using the system bundle.

i think it should always generated the bundle regardless of whether httpsProxy is set.  Otherwise the bundle risks going stale if someone unsets httpsProxy.

pkg/controller/proxyconfig/validation.go
```diff
@@ -25,7 +33,7 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
 			return fmt.Errorf("invalid httpProxy URI: %v", err)
 		}
 		if scheme != proxyHTTPScheme {
-			return fmt.Errorf("httpProxy requires an %q URI scheme", proxyHTTPScheme)
+			return fmt.Errorf("httpProxy requires an '%s' URI scheme", proxyHTTPScheme)
```
wking: just leaning in here on a brief skim, but why change `%q` to the `'%s'`.  Seems like more characters for the same effect?

pkg/controller/proxyconfig/controller.go
```diff
@@ -67,72 +90,145 @@ type ReconcileProxyConfig struct {
...
+
+		// Only proceed if the required config objects can be collected.
+		if validate {
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Name: names.PROXY_CONFIG}, infraConfig); err != nil {
+				log.Printf("failed to get infrastructure config '%s': %v", names.PROXY_CONFIG, err)
+				r.status.SetDegraded(statusmanager.ProxyConfig, "InfraConfigError",
+					fmt.Sprintf("Error getting infrastructure config %s: %v.", names.PROXY_CONFIG, err))
+				return reconcile.Result{}, nil
+			}
+			if err := r.client.Get(context.TODO(), types.NamespacedName{Name: names.PROXY_CONFIG}, netConfig); err != nil {
```
squeed: wrong `names.`

pkg/controller/proxyconfig/controller.go
```diff
@@ -67,72 +90,145 @@ type ReconcileProxyConfig struct {
...
+				log.Printf("Failed to validate proxy '%s': %v", proxyConfig.Name, err)
+				r.status.SetDegraded(statusmanager.ProxyConfig, "InvalidProxyConfig",
+					fmt.Sprintf("The configuration is invalid for proxy '%s' (%v). "+
+						"Use 'oc edit proxy.config.openshift.io %s' to fix.", proxyConfig.Name, err, proxyConfig.Name))
+				return reconcile.Result{}, nil
+			}
+		}
+
+		// Update proxy status.
+		if err := r.syncProxyStatus(proxyConfig, infraConfig, netConfig, clusterCfgMap); err != nil {
```
squeed: if `validate == false`, then  `infraConfig`, `netConfig`, and `clusterCfgMap` will be `nil` - is that OK?

pkg/controller/proxyconfig/validation.go
danehans: @wking consistency with formatting of other error messages. I can go back and update the other errors to use `%q`, but I'd prefer addressing that in a follow-on PR due to time constraints.

wking: I have no stake in this; don't hold things up on my account

pkg/controller/proxyconfig/controller.go
danehans: ack

pkg/controller/proxyconfig/controller.go
danehans: @squeed thanks for pointing this out. I'm going to pull the calls out of the `verify` block since the values are only being used for spec>status sync.

pkg/controller/proxyconfig/controller.go
danehans: @bparees I'm having problems implementing 1. 2 and 3 are working as expected. Reconciliation of 1 keeps cycling.  Since https://github.com/openshift/cluster-network-operator/pull/274 and operators are waiting on this PR, I would like to implement 1 in a follow-on PR. Additionally, this PR is getting quite large. Lastly, I'm not sure a watch is the best approach. For example, what if someone deletes the resource? I'm thinking a go routine that periodically ensures the configmap exists may be a better direction. Thoughts?

pkg/controller/proxyconfig/controller.go
danehans: Updated to `trusted-ca-bundle`.

pkg/controller/proxyconfig/controller.go
danehans: godoc for `handleConfigMap()` fixed to reference 'openshift-config' namespace. See https://github.com/openshift/cluster-network-operator/pull/271#discussion_r313666739 for details why a 'openshift-config-managed/trusted-ca-bundle'  watch is not included in this PR.

pkg/controller/proxyconfig/controller.go
danehans: @bparees see https://github.com/openshift/cluster-network-operator/pull/271#discussion_r313666739

pkg/controller/proxyconfig/controller.go
danehans: a bundle is being generated regardless of httpsProxy being set. If `trustedCA` is provided, a combined `trustedCA/system` bundle is generated. Otherwise, a bundle is generated containing the system bundle.

pkg/controller/proxyconfig/controller.go
bparees: >  I'm having problems implementing 1. 2 and 3 are working as expected. Reconciliation of 1 keeps cycling.

Can you elaborate?

>  I would like to implement 1 in a follow-on PR.

so currently if a user edits or deletes the openshift-config-managed CM, nothing will restore/fix it?  But otherwise the CM is getting created/updated appropriately when the user provides CAs?  Yes i think we can consider that sufficient for a first pass for now.


> I'm thinking a go routine that periodically ensures the configmap exists may be a better direction. 

that is part of what a controller is supposed to do (for resources it cares about).  A controller consists of two aspects:

1) watch for created/updated/delete events, take action when the event happens (which can include recreating a resource that was deleted)
2) periodically running a sync/relist process which checks that the state of the system is as desired, in case the controller missed an event (such as because it was down).

danehans: @bparees 
> Can you elaborate?

I'll post a link to the code and log messages shortly.

> so currently if a user edits or deletes the openshift-config-managed CM, nothing will restore/fix it?

It will be restored when 1) a proxy reconciliation event is triggered 2) any changes to the user-provided configmap will trigger a sync causing `openshift-config-managed/trusted-ca-bundle` to be restored to the desired state or 3) as you mention, on periodic resync.

pkg/controller/proxyconfig/controller.go
```diff
@@ -48,6 +51,21 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	// so filter events before they are provided to the controller event handlers.
+	pred := predicate.Funcs{
+		UpdateFunc:  func(e event.UpdateEvent) bool { return handleConfigMap(e.MetaNew) },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return handleConfigMap(e.Meta) },
+		CreateFunc:  func(e event.CreateEvent) bool { return handleConfigMap(e.Meta) },
+		GenericFunc: func(e event.GenericEvent) bool { return handleConfigMap(e.Meta) },
+	}
+
+	// Watch for changes to the additional trust bundle configmap.
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, pred)
```
bparees: operator-fwk doesn't let you setup the watch on a specific namespace?

pkg/names/names.go
```diff
@@ -37,10 +37,51 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+// TRUST_BUNDLE_CONFIGMAP_KEY is the name of the data key containing
+// the PEM encoded trust bundle.
+const TRUST_BUNDLE_CONFIGMAP_KEY = "ca-bundle.crt"
+
+// SYSTEM_TRUST_BUNDLE is the full path to the file containing
+// the system trust bundle.
+//const SYSTEM_TRUST_BUNDLE = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+// TODO: remove following line used for testing.
+//const SYSTEM_TRUST_BUNDLE = "/usr/local/etc/openssl/cert.pem"
+const SYSTEM_TRUST_BUNDLE = "/Users/daneyonhansen/tmp/test.pem"
```
bparees: needs to be set back to the real value?

pkg/names/names.go
```diff
@@ -37,10 +37,51 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+func TrustedCABundleConfigMap() types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: TRUSTED_CA_BUNDLE_CONFIGMAP_NS,
+		Name:      TRUSTED_CA_BUNDLE_CONFIGMAP_NAME,
+	}
+}
+
+// AddlTrustBundleConfigMapNS returns the namespaced name of the
+// namespace containing the user-provided trust bundle ConfigMap.
+func AddlTrustBundleConfigMapNS() types.NamespacedName {
```
bparees: seems slightly odd that we're returning a "NamespacedName" type when the value we are returning is just the Namespace?

danehans: @bparees thanks for catching this. It's an unused func so I'll remove it.

pkg/names/names.go
danehans: ack

pkg/controller/proxyconfig/controller.go
danehans: @bparees CNO uses the controller-runtime pkg. Cluster Ingress Operator also uses this pkg and follows the same [pattern](https://github.com/openshift/cluster-ingress-operator/blob/master/pkg/operator/controller/certificate-publisher/controller.go#L79-L86) of using [predicates](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/event-filtering.md) to filter events from being transformed into requests that get reconciled. The predicate [implemented](https://github.com/openshift/cluster-network-operator/pull/271/files#diff-6cc3b6cf53aa0e5d6919e9943b4b174aR80) in this PR will filter the `ConfigMap` source to namespace `openshift-config`.

bparees: yeah, just thought we could be more efficient by scoping to the namespace.  but this is fine.

danehans:
aws-e2e-upgrade job failure due to:
`Aug 15 20:45:11.316 E clusterversion/version changed Failing to True: ClusterOperatorDegraded: Cluster operator network is reporting a failure: failed to generate system trust bundle configmap ’openshift-config-managed/trusted-ca-bundle (failed to validate trust bundle /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem: certificate is not a CA certificate: <nil>).`

I found that ` /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` contains a certificate with serial # `9b:7e:06:49:a3:3e:62:b9:d5:ee:90:48:71:29:ef:57` that does not contain `CA: True`. I am removing the following from `CertificateData` in `validation/trustbundle.go`:
```
		if !cert.IsCA {
			return nil, nil, fmt.Errorf("certificate is not a CA certificate: %v", err)
		}
```

PR #274, opened 2019-08-06
Add CA bundle injector 

JacobTanenbaum:
This adds the ability to inject a combined CA bundle into configmaps that are setup to want them

JIRA [SDN-496](https://jira.coreos.com/browse/SDN-496)

PTAL @danehans


pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	proxyAnnotation := map[string]string{
+		names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION: "true",
+	}
+	cfpMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: proxyAnnotation,
+		},
+	}
+	// The events fire for changes/createion of the trusted-ca-bundle and any configmaps with the
```
ricardomaraschini: small typo here

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+
+var _ reconcile.Reconciler = &ReconcileConfigMapInjector{}
+
+type ReconcileConfigMapInjector struct {
+	client client.Client
+	scheme *runtime.Scheme
+	status *statusmanager.StatusManager
+}
+
+func (r *ReconcileConfigMapInjector) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
squeed: A bit of docstring explaining what this does would be helpful

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		configMapsToChange = append(configMapsToChange, *requestedCAbundleConfigMap)
+	}
+
+	for _, configMap := range configMapsToChange {
+		if configMap.Data == nil {
+			configMap.Data = map[string]string{names.TRUST_BUNDLE_CONFIGMAP_KEY: string(combinedCAbundleData)}
+		} else {
+			configMap.Data[names.TRUST_BUNDLE_CONFIGMAP_KEY] = string(combinedCAbundleData)
+		}
+		err = r.client.Update(context.TODO(), &configMap)
```
squeed: Because you're touching so many configmaps, this might result in a lot of conflicts. I'd recommend changing this to a patch, as well as using the retryOnConflict helpers.

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+
+// TRUST_BUNDLE_CONFIGMAP_ANNOTATION is the name of the annotation that
+// determines whether or not to inject the combined ca certificate
+const TRUST_BUNDLE_CONFIGMAP_ANNOTATION = "config.openshift.io/inject-trusted-cabundle"
+
+// Proxy returns the namespaced name of the proxy
+// object named "cluster" in namespace "openshift-config-managed".
+func Proxy() types.NamespacedName {
+	return types.NamespacedName{
+		Name: PROXY_CONFIG,
```
squeed: Missing the namespace?

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
+package configmapcainjector
```
danehans: I suggest renaming the file to `controller.go` and the directory to `configmap_ca_injector` (idiomatic go). pkg will still be `configmapcainjector`.

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+var _ reconcile.Reconciler = &ReconcileConfigMapInjector{}
+
+type ReconcileConfigMapInjector struct {
+	client client.Client
+	scheme *runtime.Scheme
+	status *statusmanager.StatusManager
+}
+
+func (r *ReconcileConfigMapInjector) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Printf("Reconciling update for ca-certs from %s/%s\n", request.Name, request.Namespace)
```
danehans: s/Reconciling update for ca-certs from/Reconciling configmap from/

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+
+	combinedCAbundleConfigMap := &corev1.ConfigMap{}
```
danehans: I would like to refrain from referencing the trusted ca bundle by different names such as combined, merged, etc. to avoid confusion. I would like to standardize with `trustedCABundle` + `whatever` representing the trusted ca bundle (i.e. combined user/system bundle) and `additionalTrustBundle` + `whatever` representing the user-provided bundle.

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+	//build a list of configMaps
+	configMapsToChange := []corev1.ConfigMap{}
+
+	//The trusted-ca-bundle changed
+	if request.Name == names.TRUST_BUNDLE_CONFIGMAP && request.Namespace == names.TRUST_BUNDLE_CONFIGMAP_NS {
+
+		listOfConfigMaps := &corev1.ConfigMapList{}
+		err = r.client.List(context.TODO(), &client.ListOptions{}, listOfConfigMaps)
+		if err != nil {
+			return reconcile.Result{}, nil
```
danehans: Include error context, i.e. "Failed to list..."

danehans: Use a list options [field selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) to filter results by `names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION == "true"`.

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		return reconcile.Result{}, err
+	}
+
+	//build a list of configMaps
+	configMapsToChange := []corev1.ConfigMap{}
+
+	//The trusted-ca-bundle changed
+	if request.Name == names.TRUST_BUNDLE_CONFIGMAP && request.Namespace == names.TRUST_BUNDLE_CONFIGMAP_NS {
+
+		listOfConfigMaps := &corev1.ConfigMapList{}
```
danehans: s/listOfConfigMaps/configMapList/

pkg/controller/configmapcainjector/configmapcainjector_controller.go
```diff
@@ -0,0 +1,165 @@
...
+		err = r.client.List(context.TODO(), &client.ListOptions{}, listOfConfigMaps)
+		if err != nil {
+			return reconcile.Result{}, nil
+		}
+		for _, configMap := range listOfConfigMaps.Items {
+			if configMap.ObjectMeta.Annotations[names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION] == "true" {
```
danehans: You should be able to remove this `if` after implementing https://github.com/openshift/cluster-network-operator/pull/274/files#r311695478

pkg/controller/configmapcainjector/configmapcainjector_controller.go
danehans: @squeed take a look at my comment here: https://github.com/openshift/cluster-network-operator/pull/274/files#r311696339 and let me know your thoughts on the approach I suggested.

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+// ADDL_TRUST_BUNDLE_CONFIGMAP and TRUST_BUNDLE_CONFIGMAP
+// ConfigMaps.
+const TRUST_BUNDLE_CONFIGMAP_NS = "openshift-config-managed"
+
+// TRUST_BUNDLE_CONFIGMAP_ANNOTATION is the name of the annotation that
+// determines whether or not to inject the combined ca certificate
+const TRUST_BUNDLE_CONFIGMAP_ANNOTATION = "config.openshift.io/inject-trusted-cabundle"
+
+// Proxy returns the namespaced name of the proxy
+// object named "cluster" in namespace "openshift-config-managed".
```
danehans: proxy "cluster' resided in the default ns, not the "openshift-config-managed" ns. cc @squeed

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+func Proxy() types.NamespacedName {
+	return types.NamespacedName{
+		Name: PROXY_CONFIG,
+	}
+}
+
+// TrustBundleConfigMap returns the namespaced name of the ConfigMap
+// containing the merged user/system trust bundle.
+func TrustBundleConfigMap() types.NamespacedName {
```
danehans: s/TrustBundleConfigMap/TrustedCABundleConfigMap/

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
 // that is used in multus admission controller deployment
 const MULTUS_VALIDATING_WEBHOOK = "multus.openshift.io"
+
+// TRUST_BUNDLE_CONFIGMAP_KEY is the name of the data key containing
+// the PEM encoded trust bundle.
+const TRUST_BUNDLE_CONFIGMAP_KEY = "ca-bundle.crt"
+
+// TRUST_BUNDLE_CONFIGMAP is the name of the ConfigMap
+// containing the combined user/system trust bundle.
+const TRUST_BUNDLE_CONFIGMAP = "trusted-ca-bundle"
```
danehans: s/TRUST_BUNDLE_CONFIGMAP/TRUSTED_CA_BUNDLE_CONFIGMAP/

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+
+// TRUST_BUNDLE_CONFIGMAP_NS is the namespace that hosts the
+// ADDL_TRUST_BUNDLE_CONFIGMAP and TRUST_BUNDLE_CONFIGMAP
+// ConfigMaps.
+const TRUST_BUNDLE_CONFIGMAP_NS = "openshift-config-managed"
```
danehans: s/TRUST_BUNDLE_CONFIGMAP_NS/TRUSTED_CA_BUNDLE_CONFIGMAP_NS/

pkg/names/names.go
```diff
@@ -31,3 +36,37 @@ const SERVICE_CA_CONFIGMAP = "openshift-service-ca"
...
+
+// TRUST_BUNDLE_CONFIGMAP_ANNOTATION is the name of the annotation that
+// determines whether or not to inject the combined ca certificate
+const TRUST_BUNDLE_CONFIGMAP_ANNOTATION = "config.openshift.io/inject-trusted-cabundle"
```
danehans: s/TRUST_BUNDLE_CONFIGMAP_ANNOTATION/TRUSTED_CA_BUNDLE_CONFIGMAP_ANNOTATION/

pkg/controller/configmapcainjector/configmapcainjector_controller.go
JacobTanenbaum: @danalanerh for configmaps the only valid field selectors are meta.name and meta.namespace

pkg/controller/configmapcainjector/configmapcainjector_controller.go
JacobTanenbaum: This is also not a full kubeClient patch is not implemented. I did put in the RetryOnConflict

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	//build a list of configMaps
+	configMapsToChange := []corev1.ConfigMap{}
+
+	//The trusted-ca-bundle changed
+	if request.Name == names.TRUST_BUNDLE_CONFIGMAP && request.Namespace == names.TRUST_BUNDLE_CONFIGMAP_NS {
+
+		configMapList := &corev1.ConfigMapList{}
+		err = r.client.List(context.TODO(), &client.ListOptions{}, configMapList)
```
bparees: this is going to list every configmap in the cluster, right?  i really think this needs to move to a listwatcher so we have a cache, otherwise this is going to put a lot of load on the apiserver w/ every reconcile loop.

pkg/names/names.go
bparees: it doesn't reside in any namespace, it's a cluster scoped resource.

pkg/names/names.go
bparees: proxy config is a cluster scoped resource, it is not namespaced.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	cfpMap := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations: proxyAnnotation,
+		},
+	}
+	// The events fire for changes/creation of the trusted-ca-bundle and any configmaps with the
+	//"config.openshift.io/inject-trusted-cabundle" annotation
+	pred := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.MetaNew.GetAnnotations()[names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION] == "true" ||
```
dcbw: @JacobTanenbaum this is basically the same code for all 4 events, so just make a function for it and call it from each one.

pkg/controller/configmap_ca_injector/controller.go
dcbw: @bparees AFAICT there's no way to get an actual KubeClient in this code (controller-runtime doesn't seem to provide it). But @JacobTanenbaum I think what you can do is to use the Predicate create/delete/update functions for your cache handling. So for example you'd have a `configMapCache map[types.NamespacedName]ConfigMap` as your cache and whenever your Predicate's Create is called if the event.Object passes your test, add it to the cache. Likewise whenever your Predicate's Delete is called remove that object from the cache. Update is a bit tricker because if the object doesn't  pass the test, you need to delete it from the cache, but if it passes the test you add it to the cache.

Then in your Reconcile you can just range over the cache map instead of re-grabbing stuff.

@bparees that said, doesn't all this come from a cache inside the shared indexer anyway? If true, why would we double-cache things?

pkg/controller/configmapcainjector/configmapcainjector_controller.go
dcbw: @JacobTanenbaum @squeed so Patch was apparently added upstream in https://github.com/kubernetes-sigs/controller-runtime/pull/235 which merged 2019-03-19. And server-side patch was added in https://github.com/kubernetes-sigs/controller-runtime/pull/435.

And those are only in controller-runtime 0.2.0-alpha and later. CNO currently vendors operator-sdk 0.4.1 which doesn't have that. And even upstream operator-sdk only locks controller-runtime 0.1.10 https://github.com/operator-framework/operator-sdk/blob/master/go.mod

So until upstream operator-sdk updates, and then CNO revendors from that, we can't get Patch.

pkg/controller/configmap_ca_injector/controller.go
bparees: yeah i didn't realize you guys are using the operator-sdk which gives you a caching client.  I thought it was just a normal client-go client.  This is fine, if indeed the client is backed by a local cache that's being updated via watch events.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,177 @@
...
+	if reconciler == nil {
+		return fmt.Errorf("failed to create reconciler")
+	}
+
+	return add(mgr, reconciler)
+}
+
+func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager) reconcile.Reconciler {
+	if err := configv1.Install(mgr.GetScheme()); err != nil {
+		return &ReconcileConfigMapInjector{}
```
ricardomaraschini: Shouldn't this be `return nil` ?

ricardomaraschini: Maybe even "bubble-up" the error returned by `Install()`

squeed:
So, a question - by making this reliant on an annotation, it means that we always need to watch *all* configmaps forever. This could easily result in scaling issues, since our client will have every configmap in cache.

I would suggest making the API be that a specific label needs to be set. While the operator-framework doesn't currently support label-selected watches, the underlying client machinery certainly does. That will help us down-the-line.

@enj as the maintainer of the service-ca-controller (which does a similar thing), what do you think?

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,177 @@
...
+			return e.Meta.GetAnnotations()[names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION] == "true" ||
+				(e.Meta.GetName() == names.TRUST_BUNDLE_CONFIGMAP && e.Meta.GetNamespace() == names.TRUST_BUNDLE_CONFIGMAP_NS)
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return e.Meta.GetAnnotations()[names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION] == "true" ||
+				(e.Meta.GetName() == names.TRUST_BUNDLE_CONFIGMAP && e.Meta.GetNamespace() == names.TRUST_BUNDLE_CONFIGMAP_NS)
+		},
+	}
+
+	err = c.Watch(&source.Kind{Type: cfpMap}, &handler.EnqueueRequestForObject{}, pred)
```
ricardomaraschini: We should not ignore this error.

JacobTanenbaum: I added the check for the error

pkg/names/names.go
danehans: Understood, this is a doc bug that was fixed in https://github.com/openshift/cluster-network-operator/pull/245/files#diff-dc24d687a587f957b553ce29a8dfaff8R40-R46. It will be fixed in this PR when I rebase.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,182 @@
...
+type ReconcileConfigMapInjector struct {
+	client client.Client
+	scheme *runtime.Scheme
+	status *statusmanager.StatusManager
+}
+
+// Reconcile expectes requests to refers to configmaps of two different types.
+// 1. a configmap named trusted-ca-bundle in namespace openshift-config-managed and will ensure that all configmaps with the annotation
+// config.openshift.io/inject-trusted-cabundle = true have the certificate information stored in trusted-ca-bundle's ca-bundle.crt entry.
+// 2. a configmap with the annotation config.openshift.io/inject-trusted-cabundle = true and will insure that it contains the ca-bundle.crt
```
danehans: Update 2 to indicate all configmaps in all namespaces with the annotation.

pkg/names/names.go
danehans: @JacobTanenbaum this but should be fixed when you rebase this PR off #245 that just merged.

pkg/controller/configmapcainjector/configmapcainjector_controller.go
danehans: @JacobTanenbaum ack.

pkg/names/names.go
danehans: @bparees what is the difference? I thought default, cluster-scoped and non-namespaced are the same.

bparees: `default` is an actual namespace you can go look at on a cluster

```
 oc get sa -n default
NAME       SECRETS   AGE
builder    2         124m
default    2         130m
deployer   2         124m

```

squeed:
hey @bparees do you have a thought about the label-vs-annotation question?

bparees:
> hey @bparees do you have a thought about the label-vs-annotation question?

sorry, I do not..  we'll have to wait for the service-ca-controller owners to weigh in.

squeed:
Everyone I've spoken to on the operator-sdk side has suggested very strongly to use a label, rather than an annotation, as the API.

squeed:
Confirmed with some other teams who have done this (monitoring): Unless there's an extremely good reason, we should use a label. So, I'm going to block this until we switch to label-triggered injection.

/hold

pkg/names/names.go
deads2k: I think we're likely to have multiple different trust bundles.  Let's add proxy to this name so that we can distinguish.  We can always provide combinations at some point, but we cannot resplit.

bparees: We ended up on this name because the thinking (as driven by @smarterclayton) was that ultimately we would want this to be the mechanism for providing the canonical "CAs to trust" to things on the cluster that wanted to be told what to trust.  e.g. if an admin wanted to entirely replace the system trust bundle, this would eventually be that admin-provided bundle.  (today it aggregates in system, but at some point we might make it optional to not do that).

So given that target use, i'm reluctant to put proxy back in the name of this thing.   But i'm not sure how to reconcile that w/ your goal of "there might be many CA bundles".  

I suppose if we ever reach the world I described above, we could introduce "trusted-ca-bundle" as a thing at that point in time, so I guess i'm ok w/ the proposed rename.

tldr: we started out including proxy in the name, i'm ok going back to it I guess.

But this probably also means you're not going to like the label name that we've socialized for people to put on their configmaps (that gets the CAs injected into the CM).  So we better settle on that asap as we're causing downstream teams churn every time we change direction on it (we already moved them from annotations to labels).

pkg/names/names.go
bparees: I think @deads2k is going to suggest this name include proxy for the same reasons he mentioned in another PR (that we might have other bundles to inject in the future).  but i'll let him comment.

bparees: and @smarterclayton since we went down this path somewhat at your behest, if you do *not* think we should be anticipating multiple bundles, would also like to hear from you.

pkg/names/names.go
bparees: @smarterclayton is going to weigh in on this, please don't rename anything yet.

pkg/names/names.go
bparees: @smarterclayton is going to weigh in on this, please don't rename anything yet.

pkg/controller/configmap_ca_injector/controller.go
squeed: Regardless of how this is currently implemented, the operator-sdk team is receptive to the idea of a label-scoped watch - https://github.com/operator-framework/operator-sdk/issues/1808

bparees: i've gotten confirmation that proceeding w/ using a label is fine.

pkg/controller/configmapcainjector/configmapcainjector_controller.go
squeed: Bummer. Thanks for looking in to that.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	if request.Name == names.TRUST_BUNDLE_CONFIGMAP && request.Namespace == names.TRUST_BUNDLE_CONFIGMAP_NS {
+
+		configMapList := &corev1.ConfigMapList{}
+		selector := labels.Set(map[string]string{names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION: "true"}).AsSelector()
+		err = r.client.List(context.TODO(), &client.ListOptions{LabelSelector: selector}, configMapList)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		configMapsToChange = configMapList.Items
+	} else {
```
squeed: nit: add a comment like "a single labeled config map changed."

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,172 @@
...
+	configMapsToChange := []corev1.ConfigMap{}
+
+	//The trusted-ca-bundle changed
+	if request.Name == names.TRUST_BUNDLE_CONFIGMAP && request.Namespace == names.TRUST_BUNDLE_CONFIGMAP_NS {
+
+		configMapList := &corev1.ConfigMapList{}
+		selector := labels.Set(map[string]string{names.TRUST_BUNDLE_CONFIGMAP_ANNOTATION: "true"}).AsSelector()
+		err = r.client.List(context.TODO(), &client.ListOptions{LabelSelector: selector}, configMapList)
+		if err != nil {
+			return reconcile.Result{}, err
```
squeed: Should log the error here too.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,172 @@
...
+		configMapsToChange = append(configMapsToChange, *requestedCAbundleConfigMap)
+	}
+
+	for _, configMap := range configMapsToChange {
+		if configMap.Data == nil {
+			configMap.Data = map[string]string{names.TRUST_BUNDLE_CONFIGMAP_KEY: string(trustedCAbundleData)}
+		} else {
+			configMap.Data[names.TRUST_BUNDLE_CONFIGMAP_KEY] = string(trustedCAbundleData)
+		}
+		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
squeed: Suggestion: gather errors in an array, try to update all configmaps, then handle a non-zero `errors` array. That way a single failure won't cause all others to block.

I suggest also adding an error threshold - if `len(errors) > 5`, then just give up.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,172 @@
...
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileConfigMapInjector) validateTrustBundleConfigMap(cfgMap *corev1.ConfigMap) ([]byte, error) {
```
squeed: nit: does this function doesn't really seem to do anything besides discard the first return value?

squeed:
A few small code-cleanups, and it looks like there are some outstanding naming questions (though these are internal, so maybe we don't care). Otherwise getting close.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			errArray = append(errArray, err)
+			if len(errArray) > 5 {
+				return reconcile.Result{}, fmt.Errorf("Too many errors attempting to update configmaps with CA cert. data")
+			}
+		}
+
+	}
+
+	if errArray != nil {
```
squeed: `if len() > 0`

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmapcainjector/configmapcainjector_controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,176 @@
...
+
+	errArray := []error{}
+
+	for _, configMap := range configMapsToChange {
+		if configMap.Data == nil {
+			configMap.Data = map[string]string{names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY: string(trustedCAbundleData)}
+		} else {
+			configMap.Data[names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY] = string(trustedCAbundleData)
+		}
+		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
squeed: Sorry, brain fart, I should have missed this earlier. In order for RetryOnConflict to work, you have to Get() the config map again in the retry function.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,176 @@
...
+			return reconcile.Result{}, err
+		}
+		configMapsToChange = append(configMapsToChange, *requestedCAbundleConfigMap)
+	}
+
+	errArray := []error{}
+
+	for _, configMap := range configMapsToChange {
+		if configMap.Data == nil {
+			configMap.Data = map[string]string{names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY: string(trustedCAbundleData)}
```
squeed: You always need to DeepCopy() an object returned from Get(), or else you pollute the cache.

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,183 @@
...
+
+		//get the requested object
+		requestedCAbundleConfigMap := &corev1.ConfigMap{}
+		requestedCAbundleConfigMapName := types.NamespacedName{
+			Namespace: request.Namespace,
+			Name:      request.Name,
+		}
+		err = r.client.Get(context.TODO(), requestedCAbundleConfigMapName, requestedCAbundleConfigMap)
+		if err != nil {
+			log.Println(err)
```
squeed: Subtle case: you might have a bit of a stale notification, and the ConfigMap has since been deleted. So, you need to add
```
if apierrors.IsNotFound(err){
}
```

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,183 @@
...
+				log.Println(err)
+				return err
+			}
+			configMapToUpdate := retrievedConfigMap.DeepCopy()
+			if configMapToUpdate.Data == nil {
+				configMapToUpdate.Data = map[string]string{names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY: string(trustedCAbundleData)}
+			} else {
+				configMapToUpdate.Data[names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY] = string(trustedCAbundleData)
+			}
+			err = r.client.Update(context.TODO(), configMapToUpdate)
```
squeed: I recommend adding a DeepEquals check here, and skipping the `Update()` if already equals.

JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,192 @@
...
+	if request.Name == names.TRUSTED_CA_BUNDLE_CONFIGMAP && request.Namespace == names.TRUSTED_CA_BUNDLE_CONFIGMAP_NS {
+
+		configMapList := &corev1.ConfigMapList{}
+		selector := labels.Set(map[string]string{names.TRUSTED_CA_BUNDLE_CONFIGMAP_LABEL: "true"}).AsSelector()
+		err = r.client.List(context.TODO(), &client.ListOptions{LabelSelector: selector}, configMapList)
+		if err != nil {
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		configMapsToChange = configMapList.Items
```
squeed: Add a log line, something like `log.Printf("trusted-ca-bundle changed, updating %d configMaps", len(configMapsToChange)`

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,192 @@
...
+		return reconcile.Result{}, err
+	}
+	_, trustedCAbundleData, err := validation.TrustBundleConfigMap(trustedCAbundleConfigMap)
+
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	//build a list of configMaps
+	configMapsToChange := []corev1.ConfigMap{}
```
squeed: This should be an array of pointers.

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/controller/configmap_ca_injector/controller.go
squeed: Result: nevermind. Somehow we're not using the shared lister.

squeed:
I think this is ready to merge - with one question: is the validation cherry-picked from #271 still correct?

squeed:
I'll approve this - @danehans, can you give it the final lgtm?
/approve

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+	trustedCAbundleConfigMap := &corev1.ConfigMap{}
+	trustedCAbundleConfigMapName := types.NamespacedName{
+		Namespace: names.TRUSTED_CA_BUNDLE_CONFIGMAP_NS,
+		Name:      names.TRUSTED_CA_BUNDLE_CONFIGMAP,
+	}
+	err := r.client.Get(context.TODO(), trustedCAbundleConfigMapName, trustedCAbundleConfigMap)
+	if err != nil {
+		log.Println(err)
+		return reconcile.Result{}, err
```
squeed: If IsNotExist, then this should return nil, not err - @danehans can you confirm?

squeed: Or perhaps it should succeed and delete all the CAs?

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -0,0 +1,189 @@
...
+
+	errs := []error{}
+
+	for _, configMap := range configMapsToChange {
+		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			retrievedConfigMap := &corev1.ConfigMap{}
+			err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: configMap.Namespace, Name: configMap.Name}, retrievedConfigMap)
+			if err != nil {
+				log.Println(err)
+				return err
```
squeed: again, if IsNotExist, return nil

danehans: Agreed.

pkg/controller/configmap_ca_injector/controller.go
danehans: Lets return nil.

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

pkg/names/names.go
danehans: @bparees thanks for the clarification.

pkg/controller/configmap_ca_injector/controller.go
JacobTanenbaum: fixed

danehans:
@JacobTanenbaum this PR makes an api call to get the trust bundle configmap for every configmap injection request. See if you can improve in a future PR by adding `[]byte` to `ReconcileConfigMapInjector`, storing the trust bundle configmap data in the slice, updating the `[]byte` whenever a change to the trust bundle configmap is observed and the configmap data and local `[]byte` differ, using the `[]byte` for injection requests when it's not empty or getting the trust bundle data and populating the `[]byte` when it's empty. Thoughts @squeed?

bparees:
holding in case squashing is in order or other housekeeping.

squeed:
> this PR makes an api call to get the trust bundle configmap for every configmap injection request.

That should be cached, so I'm not so concerned about that.

squeed:
Looks fine to me, commit-wise
/hold cancel
woo

akostadinov:
Is there any documentation about how this should be used from user point of view?

PR #275, opened 2019-08-07
Bug 1743302: Kuryr: Expand svc net to fix VRRP ports conflicts

dulek:

To fix this the service network created in OpenStack will be expanded to
be twice bigger than the ServiceNetwork value from NetworkSpec. Then the
other half of it will be used for Amphora VRRP ports, while OpenShift
will only choose IP's from the half indicated by
NetworkSpec.ServiceNetwork.

Obviously this may lead to conflicts if expanded network overlaps with
ClusterNetwork, so validation is expanded to check that.


dulek:
This will fail `dep` check until openshift/api#401 is merged.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -207,14 +207,14 @@ func findOpenStackSubnet(client *gophercloud.ServiceClient, name, tag string) (s
 // Looks for a Neutron subnet by name, tag, network ID, CIDR and optionally
 // gateway IP, if it does not exist creates it. Will fail if two subnets match
 // all the criteria.
-func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId, cidr string, gatewayIp *string) (string, error) {
+func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId, cidr, gatewayIp string, allocationRanges []iputil.IPRange) (string, error) {
```
MaysaMacedo: It might be good to update the comment above to remove the optional for the gateway IP and state that a predefined allocationPools will be used in creation.

dulek: Done.

PR #277, opened 2019-08-07
Adds readinessEndpoint validation for Proxy API type.

danehans:
Requires https://github.com/openshift/cluster-network-operator/pull/245.

- [x] Updates `Proxy` validation to support `readinessCheck` (when present).
- [x] Write `http|https|no Proxy` spec to status upon successful validation and a successful connection to `readinessEndpoints`.
- [ ] Unit tests.

__Note:__ `readinessEndpoints` with a https URI scheme requires https://github.com/openshift/cluster-network-operator/pull/271.

Jira: [SDN-492](https://jira.coreos.com/browse/SDN-492) 

PTAL @squeed @bparees @JacobTanenbaum

### DIFF
--- a/pkg/controller/proxyconfig/controller.go
+++ b/pkg/controller/proxyconfig/controller.go
@@ -154,3 +154,9 @@ func isSpecHTTPSProxySet(proxyConfig *configv1.ProxySpec) bool {
 func isSpecNoProxySet(proxyConfig *configv1.ProxySpec) bool {
 	return len(proxyConfig.NoProxy) > 0
 }
+
+// isSpecReadinessEndpointsSet returns true if spec.readinessEndpoints of
+// proxyConfig is set.
+func isSpecReadinessEndpointsSet(proxyConfig *configv1.ProxySpec) bool {
+	return len(proxyConfig.ReadinessEndpoints) > 0
+}
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -3,7 +3,10 @@ package proxyconfig
 import (
 	"fmt"
 	"net"
+	"net/http"
+	"net/url"
 	"strings"
+	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/util/validation"
@@ -12,6 +15,11 @@ import (
 const (
 	proxyHTTPScheme  = "http"
 	proxyHTTPSScheme = "https"
+	// proxyProbeMaxRetries is the number of times to attempt an http GET
+	// to a readinessEndpoints endpoint.
+	proxyProbeMaxRetries = 3
+	// proxyProbeWaitTime is the time to wait before retrying a failed proxy probe.
+	proxyProbeWaitTime = 1 * time.Second
 	// noProxyWildcard is the string used to as a wildcard attached to a
 	// domain suffix in proxy.spec.noProxy to bypass proxying.
 	noProxyWildcard = "*"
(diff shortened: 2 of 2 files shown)


pkg/controller/proxyconfig/validation.go
```diff
@@ -48,5 +52,63 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
+		}
+	}
+
+	return err
+}
+
+// runHTTPReadinessProbe issues an http GET request to endpoint and returns
+// an error if a 2XX or 3XX http status code is not returned.
+func runHTTPReadinessProbe(endpoint string) error {
+	resp, err := http.Get(endpoint)
```
bparees: how do we know this get is actually going through the spec proxy value that's being configured?  i'm pretty sure it's not.

danehans: Thanks for catching this. Yes, I need to have CNO request proxy injection. Let me submit a PR for that.

danehans: @bparees https://github.com/openshift/cluster-network-operator/pull/277/files#diff-1502a5b3db6bd911880e1d0d9346b561R107-R142 added proxy support to the http readiness probe.

pkg/controller/proxyconfig/validation.go
```diff
@@ -52,5 +56,87 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
+	/*proxyURL, err := url.Parse(httpProxy)
+	if err != nil {
+		return fmt.Errorf("failed to parse httpProxy url '%s': %v", httpProxy, err)
+	}*/
+
+	transport := &http.Transport{
+		// TODO: Uncomment when proxy test environment is available to
+		//  force transport to use a proxy.
+		//Proxy: http.ProxyURL(proxyURL),
+		Proxy: http.ProxyFromEnvironment,
```
bparees: you shouldn't be getting the proxy from the environment, you should be using the proxy from proxyconfig.spec since that's the proxy you want to make sure actually works.  We want to validate the spec value because that is the one the administrator is asserting the cluster should start using, and we want to confirm it works before we propagate it into status (at which point all the consumers, including the network operator itself, will pick it up and start using it).

danehans: @bparees Understood, see the lines that I commented out. `proxyURL` is the parsed URL from proxy.spec.httpProxy that gets passed in as `httpProxy`. My plan was to change this after we have a working CI. I don't have a proxy setup in my dev env. Maybe I should run a local proxy since I'm running CNO locally for testing.

pkg/controller/proxyconfig/controller.go
```diff
@@ -154,3 +154,9 @@ func isSpecHTTPSProxySet(proxyConfig *configv1.ProxySpec) bool {
 func isSpecNoProxySet(proxyConfig *configv1.ProxySpec) bool {
 	return len(proxyConfig.NoProxy) > 0
 }
+
+// isSpecReadinessEndpoints returns true if spec.readinessEndpoints of
+// proxyConfig is set.
+func isSpecReadinessEndpoints(proxyConfig *configv1.ProxySpec) bool {
```
bparees: i might be stepping on a golang idiom, but i would have thought to call this "has" not "is"

pkg/controller/proxyconfig/validation.go
```diff
@@ -52,5 +57,82 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
+
+	return nil
+}
+
+// validateHTTPReadinessEndpointWithRetries tries to validate an http
+// endpoint in a finite loop and returns the last result if it never succeeds.
+func validateHTTPReadinessEndpointWithRetries(httpProxy, endpoint string, retries int) error {
+	var err error
+	for i := 0; i < retries; i++ {
+		err = runHTTPReadinessProbe(httpProxy, endpoint)
```
bparees: might make sense to include a sleep between retries, for a second.

pkg/controller/proxyconfig/controller.go
danehans: Thanks for catching this. The func name should have `Set` suffix.

pkg/controller/proxyconfig/validation.go
danehans: @bparees PR updated to include wait time between probes.

pkg/controller/proxyconfig/validation.go
```diff
@@ -52,5 +60,85 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
 }
+
+// validateHTTPReadinessEndpointWithRetries tries to validate an http
+// endpoint in a finite loop and returns the last result if it never succeeds.
+func validateHTTPReadinessEndpointWithRetries(httpProxy, endpoint string, retries int) error {
+	var err error
+	for i := 0; i < retries; i++ {
+		if i != 0 {
+			time.Sleep(proxyProbeWaitTime)
+		}
```
bparees: you could just move this after the "if err=nil return" check and not have to look at `i`.

pkg/controller/proxyconfig/validation.go
```diff
@@ -52,5 +60,85 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
+		err = runHTTPReadinessProbe(httpProxy, endpoint)
```
bparees:
```suggestion
		if err := runHTTPReadinessProbe(httpProxy, endpoint); err==nil {
		  return nil
		}
		time.Sleep(proxyProbeWaitTime)
```

pkg/controller/proxyconfig/validation.go
danehans: ack

pkg/controller/proxyconfig/validation.go
danehans: @bparees I just pushed a fix.

pkg/controller/proxyconfig/validation.go
danehans: I updated the http probe transport to use proxy from `proxy.spec.httpProxy` instead of from env var.

pkg/controller/proxyconfig/validation.go
danehans: CNO does not need proxy injection. Instead the http probe transport has been updated to use a proxy from `proxy.spec.httpProxy`.

PR #278, opened 2019-08-08
Add Kuryr DNS modifications webhook

MaysaMacedo:
Due to Octavia not supporting UDP listeners in Queens version,
coredns deployments that uses UDP are not supported as well.
This PR creates a mutating webhook to inject a dnsConfig option,
"use-vc", to pods deployments, so TCP is used instead of UDP
for dns resolution.

### DIFF
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -12,6 +12,10 @@ type KuryrBootstrapResult struct {
 	PodSecurityGroups []string
 	ClusterID         string
 	OpenStackCloud    clientconfig.Cloud
+	WebhookCA         string
+	WebhookCAKey      string
+	WebhookCert       string
+	WebhookKey        string
 }
 
 type BootstrapResult struct {
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -2,6 +2,7 @@ package openstack
 
 import (
 	"context"
+	b64 "encoding/base64"
 	"fmt"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"k8s.io/api/core/v1"
@@ -27,6 +28,8 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/utils/openstack/clientconfig"
+	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/cluster-network-operator/pkg/platform/openstack/util/cert"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -51,6 +54,7 @@ const (
 	MinOctaviaVersionWithTimeouts   = "v2.1"
 	KubernetesEndpointsName         = "kubernetes"
 	KubernetesEndpointsNamespace    = "default"
+	KuryrNamespace                  = "openshift-kuryr"
 )
 
 func GetClusterID(kubeClient client.Client) (string, error) {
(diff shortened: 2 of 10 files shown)


MaysaMacedo:
We still need to verify this.

danwinship:
This seems like a very heavy-weight solution. It seems like this would make more sense as a kubelet (or cri-o or whatever) configuration option to change the default behavior, rather than modifying every pod to request non-default behavior.

MaysaMacedo:
@danwinship We tried injecting options to the `resolv.conf` of masters and workers, similar to how it's done here [1]. Unfortunately, kubelet does not merge options from the host's `resolv.conf` [2].

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/master/00-master/openstack/files/dhcp-dhclient-conf.yaml
[2] https://github.com/kubernetes/kubernetes/issues/59031

danwinship:
Sure, but... well, it's hard to explain specifically what I mean since I don't know what the `kuryr-dns-webhook` is doing exactly, but presumably it's setting some field on every `Pod` that causes kubelet to generate the pod resolv.conf differently, and I'm saying it doesn't make sense to have *every* pod have a "non-default" resolv.conf; we should just change kubelet's defaults somehow instead, to do the right thing automatically.

dulek:
@danwinship: Well, the issue is that cluster default pod DNS options are not configurable at all. See kubelet code [1] - ndots:5 is simply hardcoded. You can see that a try to implement that was rejected by the community [2] and admission controller was suggested as a workaround here [3].

I totally agree that a webhook is quite an overkill, but I don't really see any other way at the moment. And obviously it's fault of Octavia missing some functionality. We can totally only install the workaround for Octavia's older than Train, I think, but it's not yet released.

[1] https://github.com/kubernetes/kubernetes/blob/48388fec7eaad4ac8d84fbe20673ffacf41964a1/pkg/kubelet/network/dns/dns.go#L352
[2] https://github.com/kubernetes/kubernetes/pull/68932
[3] https://github.com/kubernetes/kubernetes/issues/59031

pkg/bootstrap/types.go
```diff
@@ -12,6 +12,10 @@ type KuryrBootstrapResult struct {
 	PodSecurityGroups []string
 	ClusterID         string
 	OpenStackCloud    clientconfig.Cloud
+	CA                string
```
dulek: How about renaming it to WebhookCA and WebhookCAKey to make sure it's not confused with CA Kuryr uses to connect to OpenStack API?

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -51,6 +53,9 @@ const (
 	MinOctaviaVersionWithTimeouts   = "v2.1"
 	KubernetesEndpointsName         = "kubernetes"
 	KubernetesEndpointsNamespace    = "default"
+	AdmissionControllerCA           = "kuryr-dns-admission-controller-secret"
```
dulek: This and `WebHookSecret` should probably go into names/names.go and be passed to templates and not hardcoded there.

pkg/util/cert/cert.go
```diff
@@ -0,0 +1,104 @@
...
+
+	certKeyPEM := bytes.Buffer{}
+	if err := pem.Encode(&certKeyPEM, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(certPrivKey)}); err != nil {
+		return "", "", errors.Wrap(err, "failed to PEM encode cert private key")
+	}
+	certPEM := bytes.Buffer{}
+	if err := pem.Encode(&certPEM, &pem.Block{Type: "CERTIFICATE", Bytes: certBytes}); err != nil {
+		return "", "", errors.Wrap(err, "failed to PEM encode cert")
+	}
+	return b64.StdEncoding.EncodeToString(certPEM.Bytes()), b64.StdEncoding.EncodeToString(certKeyPEM.Bytes()), nil
```
dulek: Please move encrypting it to `ensureCertificate()` for consistency how CA cert is generated.

pkg/util/cert/cert.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+func GenerateCA(networkName string) ([]byte, []byte, error) {
+	caTemplate := x509.Certificate{
+		SerialNumber: big.NewInt(1653),
```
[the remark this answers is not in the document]
MaysaMacedo: It makes sense. If someone else needs it in the future they can just move the code. 
Regarding the 1653, is just a random number that unequily identifies the cert, I can change that  to a more meaningful value.

PR #279, opened 2019-08-08
Bug 1737026: Making set internal to the statusManager and adding mutex for thread safety

alexanderConstantinescu:
This addresses bug: [1737026](https://bugzilla.redhat.com/show_bug.cgi?id=1737026)

I am not sure if this addresses the concerns you mentioned in the bug @danwinship , but in this case I made set internal and I added a mutex inside `set`. 

I inspired myself on `runningPodsLock` in pod.go from openshift/sdn.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/ghodss/yaml"
 
@@ -24,15 +25,17 @@ import (
 type StatusLevel int
 
 const (
-	ClusterConfig  StatusLevel = iota
-	OperatorConfig StatusLevel = iota
-	ProxyConfig    StatusLevel = iota
-	PodDeployment  StatusLevel = iota
-	maxStatusLevel StatusLevel = iota
+	ClusterConfig StatusLevel = iota
+	OperatorConfig
+	ProxyConfig
+	PodDeployment
+	maxStatusLevel
 )
 
 // StatusManager coordinates changes to ClusterOperator.Status
 type StatusManager struct {
+	sync.Mutex
+
 	client  client.Client
 	name    string
 	version string
@@ -49,7 +52,8 @@ func New(client client.Client, name, version string) *StatusManager {
 }
 
 // Set updates the ClusterOperator.Status with the provided conditions
-func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...configv1.ClusterOperatorStatusCondition) {
+func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...configv1.ClusterOperatorStatusCondition) {
+
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
 	err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
 	isNotFound := errors.IsNotFound(err)
@@ -122,11 +126,11 @@ func (status *StatusManager) Set(reachedAvailableLevel bool, conditions ...confi
 func (status *StatusManager) syncDegraded() {
 	for _, c := range status.failing {
 		if c != nil {
-			status.Set(false, *c)
+			status.set(false, *c)
 			return
 		}
 	}
-	status.Set(
+	status.set(
 		false,
 		configv1.ClusterOperatorStatusCondition{
 			Type:   configv1.OperatorDegraded,
@@ -135,10 +139,8 @@ func (status *StatusManager) syncDegraded() {
 	)
 }
 
-// SetDegraded marks the operator as Degraded with the given reason and message. If it
-// is not already failing for a lower-level reason, the operator's status will be updated.
-func (status *StatusManager) SetDegraded(level StatusLevel, reason, message string) {
-	status.failing[level] = &configv1.ClusterOperatorStatusCondition{
+func (status *StatusManager) setDegraded(statusLevel StatusLevel, reason, message string) {
+	status.failing[statusLevel] = &configv1.ClusterOperatorStatusCondition{
 		Type:    configv1.OperatorDegraded,
 		Status:  configv1.ConditionTrue,
 		Reason:  reason,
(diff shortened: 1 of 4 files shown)


alexanderConstantinescu:
FYI: just created an alternative solution using a queue instead. I think `status_manager_test.go`  definitely becomes more logically contrived and complicated using that structure, and thus prefer this (hence the PR of this branch). But let me know your opinions.

squeed:
The mutex is nice and simple - I prefer it.

However, we could still run in to conflicts from the apiserver, so we should probably wrap the logic of `set()` in a `RetryOnConflict()` loop. See https://github.com/kubernetes/client-go/blob/master/examples/create-update-delete-deployment/main.go#L125 for how its' generally shaped.

danwinship:
> However, we could still run in to conflicts from the apiserver

is anyone else allowed to update our ClusterOperator status?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -36,6 +37,8 @@ type StatusManager struct {
 	name    string
 	version string
 
+	statusUpdateLock sync.Mutex
```
danwinship: The problem isn't just with status updates. Eg, it's possible that one thread is calling `SetFromPods` while another thread is calling `SetDaemonSets`, but we're not synchronizing the access to `status.daemonSets`. We need to mutex the whole `StatusManager` and all of its methods.

squeed:
> is anyone else allowed to update our ClusterOperator status?

Ah, brain fart - I thought this was a different object. No, that would be very, very unlikely.

pkg/controller/statusmanager/status_manager.go
alexanderConstantinescu: > The problem isn't just with status updates. Eg, it's possible that one thread is calling `SetFromPods` while another thread is calling `SetDaemonSets`, but we're not synchronizing the access to `status.daemonSets`. We need to mutex the whole `StatusManager` and all of its methods.

Correct me if I'm wrong, but `SetDaemonSets`, `SetDeployments` and `SetRelatedObjects` are only called from here. Which means that we can rely on them and `SetFromPods` always being called sequentially from only one thread. 

I guess the concern could be that multiple threads step on each others toes by calling `SetFromPods` individually (and one of them not having the most recent state and thus getting it wrong). But, seeing as how `set` now is synchronized it shouldn't lead to any fatal "errors" (i.e one calling `Create` instead of `Update`)? Whatever status is set at any given moment in time by any of these X threads (and might be incorrect) should be re-updated any moment later once the new thread acquires the lock and performs his update with the most up-to-date state? 

The only thing I can come to think of, as I am writing this, is `SetDegraded` and `SetNotDegraded` modifying `status.failing` unsynchronized. I will thus update my PR with a fix for this.

danwinship: > But, seeing as how `set` now is synchronized it shouldn't lead to any fatal "errors" (i.e one calling `Create` instead of `Update`)?

Note that golang does not give you any more automatic thread-safety than C does. If the `operconfig_controller` thread calls `SetDaemonSets` while the `pod_controller` thread calls `SetFromPods`, without using any synchronization primitives, then the `SetFromPods` thread could see literally any value in `status.daemonSets` (including in particular an invalid pointer).

alexanderConstantinescu: Indeed, I overlooked that. I have a (maybe not so pretty) solution to that which I will push and let you review.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -37,6 +36,9 @@ type StatusManager struct {
 	name    string
 	version string
 
+	degradedMutex sync.Mutex
```
squeed: As a general rule, using multiple mutexes increases the chance of a deadlock. Unless you really need that fine-grained access control, I'd recommend just using a single one.

alexanderConstantinescu: The reason I did that was because `setDegraded` is called from "all over the place" (`controller.go`, `operconfig_controller.go`, `clusterconfig_controller.go`), which - with only one mutex - risks generating a high thread contention. `setDegraded` never invokes a lock on `resourceMutex`, so a deadlock should thus be avoided.

danwinship: > risks generating a high thread contention

It's called from lots of places, but it's not called especially *frequently*. I don't think there will be a problem with only one mutex

alexanderConstantinescu: Fair enough, I will update my PR

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -91,28 +91,28 @@ func (r *ReconcileClusterConfig) Reconcile(request reconcile.Request) (reconcile
...
 			log.Printf("Could not apply operator config: %v", err)
-			r.status.SetDegraded(statusmanager.ClusterConfig, "ApplyOperatorConfig",
+			r.status.SetDegraded(statusmanager.ClusterConfig, true, "ApplyOperatorConfig",
 				fmt.Sprintf("Error while trying to update operator configuration: %v", err))
 			return reconcile.Result{}, err
 		}
 	}
 
-	r.status.SetNotDegraded(statusmanager.ClusterConfig)
+	r.status.SetDegraded(statusmanager.ClusterConfig, false, "", "")
```
danwinship: I don't like having the unnecessary `"", ""` there. That's why I had separate `SetDegraded` and `SetNotDegraded` to begin with.

You can keep the internal `setDegraded` with a bool but still have separate public `SetDegraded` and `SetNotDegraded` methods

pkg/controller/statusmanager/status_manager.go
```diff
@@ -21,14 +22,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-type StatusLevel int
```
danwinship: why did you get rid of the `StatusLevel` type?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -37,6 +36,8 @@ type StatusManager struct {
 	name    string
 	version string
 
+	managerMutex sync.Mutex
```
danwinship: The golang convention when you have a single mutex that covers the whole object is to use an "embedded field", normally at the top of the struct. So:

    type StatusManager struct {
            sync.Mutex

            client client.Client
            ...

and then you just use `status.Lock()` / `defer status.Unlock()`

pkg/controller/statusmanager/status_manager.go
alexanderConstantinescu: I thought it's more explicit when reading the code and seeing the that real type of `StatusLevel` is an `int`, directly.

danwinship: but it's not an arbitrary `int`. You can't do `SetNotDegraded(42)`.

Doing aliases for `int` like this is golang's equivalent of enums

alexanderConstantinescu: Ahh, I didn't know that!

alexanderConstantinescu:
/retitle Bug 1737026: Making set internal to the statusManager and adding mutex for thread safety

PR #280, opened 2019-08-08
Fixes proxy validation error formatting

danehans:
This PR fixes a formatting error in the proxy controller that was added in https://github.com/openshift/cluster-network-operator/pull/245.

/assign @squeed @bparees

### DIFF
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -25,7 +25,7 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
 			return fmt.Errorf("invalid httpProxy URI: %v", err)
 		}
 		if scheme != proxyHTTPScheme {
-			return fmt.Errorf("httpProxy requires an %q URI scheme", proxyHTTPScheme)
+			return fmt.Errorf("httpProxy requires an %s URI scheme", proxyHTTPScheme)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/proxyconfig/validation.go
```diff
@@ -34,8 +34,9 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
 		if err != nil {
 			return fmt.Errorf("invalid httpsProxy URI: %v", err)
 		}
-		if scheme != proxyHTTPScheme && scheme != proxyHTTPSScheme {
-			return fmt.Errorf("httpsProxy requires a %q or %s URI scheme", proxyHTTPScheme, proxyHTTPSScheme)
+		// TODO: Add proxyHTTPSScheme when CA support is added.
+		if scheme != proxyHTTPScheme {
+			return fmt.Errorf("httpsProxy requires a %s URI scheme", proxyHTTPScheme)
```
bparees: i don't see a reason not to allow https today.  it will work for passthrough proxies.

bparees: @danehans bump

danehans: @bparees I addressed your comments in my latest commit.

bparees:
/retest

@squeed i think this is an easy approve, unless you want to pick on the use of single quotes vs %q.

PR #281, opened 2019-08-08
Adds cvo opt-in annotation for proxy

danehans:
Required by https://github.com/openshift/cluster-network-operator/pull/277

Adds cvo proxy opt-in annotation to inject proxy env var's from https://github.com/openshift/cluster-version-operator/pull/224.

/assign @squeed @bparees

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_daemonset.yaml
@@ -5,6 +5,8 @@ metadata:
   namespace: openshift-network-operator
   labels:
     name: network-operator
+  annotations:
+    config.openshift.io/inject-proxy: network-operator
 spec:
   selector:
     matchLabels:


danehans:
@squeed I can use some guidance on how this should be tested. I was thinking an e2e, but I don't see any exist e2e for CNO.

squeed:
@danehans yeah, we don't have plans for this kind of e2e (though we have designs for other e2e bits..). Honestly the best way to test this, if you're not sure that it does what it should, is to jump on the CI cluster and check everything out.

squeed:
Hmm - if we're the ones generating this, can we just update our own env ;-) ?

manifests/0000_70_cluster-network-operator_03_daemonset.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   namespace: openshift-network-operator
   labels:
     name: network-operator
+  annotations:
+    config.openshift.io/inject-proxy: operator
```
bparees: looks like the container name is network-operator, not operator, in this deployment.

bparees:
> Hmm - if we're the ones generating this, can we just update our own env ;-) ?

?

the operator deployment is managed by the CVO.  So you need this annotation so the CVO will add the proxy env vars to your deployment.  If you tried to add them yourself, the CVO would stomp your deployment back to what it thinks it is supposed to look like (no env vars)

squeed:
> the operator deployment is managed by the CVO. So you need this annotation so the CVO will add the proxy env vars to your deployment. If you tried to add them yourself, the CVO would stomp your deployment back to what it thinks it is supposed to look like (no env vars)

I just meant with `os.Setenv()`. But perhaps that's racy.

bparees:
> I just meant with os.Setenv(). But perhaps that's racy.

we were told that golang's transport libs initilize on load, so setting the env var after is not going to have a useful effect.

and @danehans that is something to consider when you are doing your readiness endpoint checking as well... ensuring that the transport connection you are opening is actually going through the proxy (in addition to using the generated CA bundle)

bparees:
/lgtm
(I don't have approver on this repo)

manifests/0000_70_cluster-network-operator_03_daemonset.yaml
danehans: Updated to container name `network-operator` in latest commit.

danehans:
@bparees @squat lgtm was removed since I pushed a new commit to fix the annotation value.

squat:
cc the othe squ.*: @squeed

danehans:
This PR may not be needed since #277 will use static proxy settings instead of from env vars.

bparees:
> This PR may not be needed since #277 will use static proxy settings instead of from env vars.

it's needed if and only if the network operator itself (not its operands) needs to make off-cluster requests.  (not including the off-cluster requests that will be made to the proxy readiness endpoint)

danehans:
Closing b/c we don't believe CNO needs to consume proxy.

PR #283, opened 2019-08-09
Adds noProxy generation when http/https proxy is specified

danehans:
Previously, `status.noProxy` would only being generated if `spec.noProxy` was specified. `status.noProxy` should also be generated if `spec.httpProxy` or `spec.httpsProxy` are specified. If `spec.httpProxy` or `spec.httpsProxy` is specified without `spec.noProxy`, `status.noProxy` should be generated with only the cluster-wide default noProxy string.

### DIFF
--- a/pkg/controller/proxyconfig/status.go
+++ b/pkg/controller/proxyconfig/status.go
@@ -22,7 +22,7 @@ func (r *ReconcileProxyConfig) syncProxyStatus(proxy *configv1.Proxy, infra *con
 	var noProxy string
 	updated := proxy.DeepCopy()
 
-	if isSpecNoProxySet(&proxy.Spec) {
+	if isSpecNoProxySet(&proxy.Spec) || isSpecHTTPProxySet(&proxy.Spec) || isSpecHTTPSProxySet(&proxy.Spec) {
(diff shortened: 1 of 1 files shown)


pkg/controller/proxyconfig/status.go
```diff
@@ -105,8 +107,10 @@ func mergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 		set.Insert(clusterNetwork.CIDR)
 	}
 
-	for _, userValue := range strings.Split(proxy.Spec.NoProxy, ",") {
-		set.Insert(userValue)
+	if len(proxy.Spec.NoProxy) != 0 {
+		for _, userValue := range strings.Split(proxy.Spec.NoProxy, ",") {
```
wking: nit, but I like [this approach][1] better than your wrapping length guard because it also protects against injecting empty strings for values like `,i.start.with.a.comma.com`.

[1]: https://github.com/openshift/installer/pull/2198/files#diff-ff1d8a3b05383eef08ceaf043a235067R145

danehans:
Superseded by https://github.com/openshift/cluster-network-operator/pull/284

PR #285, opened 2019-08-13
Enable Kuryr HA

dulek:
This commit aims to make kuryr-controller instances to be running in A/P
HA mode. This means kuryr-controller will now run as DaemonSet on every
master node.

### DIFF
--- a/bindata/network/kuryr/003-config.yaml
+++ b/bindata/network/kuryr/003-config.yaml
@@ -26,7 +26,7 @@ data:
     ssl_ca_crt_file = /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
     token_file = /var/run/secrets/kubernetes.io/serviceaccount/token
     ssl_verify_server_crt = true
-    controller_ha = false
+    controller_ha = true
     controller_ha_elector_port = 16401
     watch_retry_timeout = 3600
     pod_vif_driver = nested-vlan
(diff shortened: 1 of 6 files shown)


dulek:
We need to start discussion here if we want it enabled as QE had not yet verified the feature, but let's keep the PR up in any case.

squeed:
Do you have a release PR to build the new image?

dulek:
@squeed: Yeah, they're there for a while now: https://github.com/openshift/release/tree/master/ci-operator/config/openshift/leader-elector. That's why I'm pretty surprised this happens.

squeed:
/approve
I'd like a member of the kuryr / openstack team to give the final lgtm.

dulek:
@squeed: Alright, we had kuryr-contoller HA tested and interestingly it seems to fail after few hours. This is most likely out of 4.2.

/hold

squeed:
hey @dulek, do you want to re-work this or should it be closed?

dulek:
@squeed: Right, thanks for pointing this out. We fixed the issue in kuryr-controller, so this solely depends if our QE team is able to commit to testing this. Most of the guys are on PTO now, so I'll need to wait for answer.

Meanwhile I can definitely rebase this.

dulek:
Okay, I'm going to close this actually. We'll just revive it once we think it's feasible.

PR #286, opened 2019-08-13
Fixes Bug 1740027: Checks length of values in noProxy merge

danehans:
Previously, values of dependent resources (i.e. network) were assumed to be present. This PR checks the presence of the value before using it in the noProxy merge function.

/assign @squeed @bparees


pkg/util/proxyconfig/merge.go
```diff
@@ -20,24 +20,37 @@ import (
...
-
 	set := sets.NewString(
 		"127.0.0.1",
 		"localhost",
-		network.Status.ServiceNetwork[0],
-		apiServerURL.Hostname(),
-		internalAPIServer.Hostname(),
 	)
 
+	if len(infra.Status.APIServerURL) > 0 {
```
abhinavdahiya: return early on error seems like more Go-idomatic rather than the if-else blocks.

PR #289, opened 2019-08-14
Adds proxy https readinessEndpoints support

danehans:
Adds support for for TLS-based probes for validating a `httpsProxy` using one or more `readinessEndpoints` that contain a `https` scheme.

Requires #271 

/assign @squeed @bparees 

PTAL at [this](https://github.com/openshift/cluster-network-operator/compare/master...danehans:ca_bundle_controller_https_readiness?expand=1#diff-1502a5b3db6bd911880e1d0d9346b561R72-R103) block and [this](https://github.com/openshift/cluster-network-operator/compare/master...danehans:ca_bundle_controller_https_readiness?expand=1#diff-1502a5b3db6bd911880e1d0d9346b561R180-R263
) block. All other changes are due to the #271 dependency.


bparees:
last 2 commits (the ones that add the probes) lgtm

pkg/controller/proxyconfig/validation.go
```diff
@@ -159,21 +177,41 @@ func (r *ReconcileProxyConfig) validateSystemTrustBundle(trustBundle string) ([]
...
+		return fmt.Errorf("failed to parse proxy url '%s': %v", proxy, err)
+	}
+
+	endpointURL, err := url.Parse(endpoint)
+	if err != nil {
+		return fmt.Errorf("failed to parse endpoint url '%s': %v", endpoint, err)
+	}
+
+	if (endpointURL.Scheme == proxyHTTPScheme && proxyURL.Scheme == proxyHTTPSScheme) ||
+		(endpointURL.Scheme == proxyHTTPSScheme && proxyURL.Scheme == proxyHTTPScheme) {
```
bparees: was one of these pairings supposed to be http?

pkg/controller/proxyconfig/validation.go
```diff
@@ -93,27 +108,30 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
 // validateTrustedCA validates that trustedCA is a valid ConfigMap
 // reference and that the ConfigMap contains a valid trust bundle,
 // returning the byte slices of the certificate data from the
-// validated trustedCA and system trust bundles.
-func (r *ReconcileProxyConfig) validateTrustedCA(trustedCA string) ([]byte, []byte, error) {
+// validated trustedCA and system trust bundles. If trustedCA is nil,
+// "trusted-ca-bundle" is used.
```
bparees: hm.  If trustedCA is nil, shouldn't we be using the system-bundle (and then updating trusted-ca-bundle to be the system bundle, if validation passes)?

pkg/controller/proxyconfig/validation.go
danehans: I tried to combined the 2 checks into 1. I'm going to separate.

pkg/controller/proxyconfig/validation.go
danehans: @bparees yes, if `trustedCA` in proxy spec is nil, then the system bundle is [generated](https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/proxyconfig/controller.go#L138-L148) and will be used. `validateTrustedCA()` requires a string which represents the value of the `openshift-config-managed/trusted-ca-bundle` configmap. I think I can drop the string parameter from this func, but I would like to do that in a follow-on PR b/c I am currently rebuilding my cluster and will be unable to test. I will update the comment to provide a better explanation.

pkg/controller/proxyconfig/validation.go
```diff
@@ -93,27 +108,31 @@ func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySp
...
 // returning the byte slices of the certificate data from the
-// validated trustedCA and system trust bundles.
-func (r *ReconcileProxyConfig) validateTrustedCA(trustedCA string) ([]byte, []byte, error) {
+// validated trustedCA and system trust bundles. If trustedCA is nil,
+// "trusted-ca-bundle" is used, consisting of only the system
+// trust bundle.
+func (r *ReconcileProxyConfig) validateTrustedCA(trustedCA string) ([]*x509.Certificate, []byte, []byte, error) {
+	if len(trustedCA) == 0 {
+		trustedCA = names.TRUSTED_CA_BUNDLE_CONFIGMAP
+	}
```
bparees: i think this logic (and the comment above) are still incorrect, per my earlier comment.

If "trustedCA" is empty that means "proxyconfig.trustedca.name" is empty (in other words, the admin has not pointed us to a configmap of user provided CAs).

If that is the case:
1) there's no point in validating anything
2) we should be returning the system trust bundle (ie the trust bundle included in our image).  Not the openshift-config-managed CM bundle.


There is never a reason to validate the openshift-config-managed CM bundle, as this is controller is itself responsible for producing that CM, so it better be valid for us to have produced it.  Furthermore the purpose of the validation we are doing here is to ensure that the CA bundle can be used to talk to the proxy.  That needs to happen *before* we copy any bundles into openshift-config-managed, because once it's in openshift-config-managed, it's going to get injected into the various consumer CMs.  So validating openshift-config-managed isn't useful (it's too late).

danehans: > If "trustedCA" is empty that means "proxyconfig.trustedca.name" is empty (in other words, the admin has not pointed us to a configmap of user provided CAs).

@bparees understood. I'm calling `validateTrustedCA()` instead of making multiple func calls to get the `trusted-ca-bundle` configmap, parse the data key into a `[]byte` and then parse the `[]byte` into a `[]*x509.Certificate` that is used as a parameter for the `validateReadinessEndpoint()` func. Some of the funcs can be improved and renamed to remove the validation prefix, but I would prefer doing that in a follow-on PR at this point.

> we should be returning the system trust bundle (ie the trust bundle included in our image). Not the openshift-config-managed CM bundle.

`trusted-ca-bundle` configmap contains the system trust bundle whether or not trustedCA is set. Even if I used the local system trust bundle, I would have to make multiple calls to read the local trust bundle file, parse the file into a `[]byte` and then parse the `[]byte` into a `[]*x509.Certificate` that is used as a parameter for the `validateReadinessEndpoint()` func. `validateTrustedCA()` already does this, so I prefer to use the func for the time being and I can improve how the funcs are composed in a future PR.

bparees: this logic needs to validate that the set of CAs we are going to publish into trusted-ca-bundle, are sufficient to talk to the proxy in question.

So we cannot use the current set off CAs in trusted-ca-bundle to do that, we need to produce the new bundle (ie "just the system CAs" or "system CAs plus the new user provided CAs") and use that bundle to perform the validation(readiness checks)

danehans: > we need to produce the new bundle (ie "just the system CAs" or "system CAs plus the new user provided CAs") and use that bundle to perform the validation(readiness checks)

@bparees I pushed a [new commit](https://github.com/openshift/cluster-network-operator/pull/289/commits/3a154407548acf78cc7a878e21d9bca2fd1ef776) that accomplishes this workflow. See this [testing doc](https://gist.github.com/danehans/4d43adf773166b41470783e4ca4f6fb6#file-test-md) for details. When `proxy.spec.trustedCA` is updated, `openshift-config-managed/trusted-ca-bundle` configmap is updated before proxy config validation (i.e. running `readinessEndpoints` probes). `openshift-config-managed/trusted-ca-bundle` will contain the combined user/system bundle if `proxy.spec.trustedCA` is updated to a configmap with a different user-proivded bundle. `openshift-config-managed/trusted-ca-bundle` will contain only the system bundle if `proxy.spec.trustedCA` is empty. The controller will use this bundle for validating the `httpsProxy` using `readinessEndpoints` that contain a `https` scheme. If the https proxy's identity cert has been signed by a ca outside the system trust bundle and `proxy.spec.trustedCA` is empty, the readiness probe will fail.

bparees:
/lgtm

future refactor should:

1) only generate the proposed bundle a single time (and then pass it to Validate, instead of having Validate construct its own bundle) so that we are actually validating the same bundle we ultimately publish
2) combine/refactor the two codepaths for "did the proxyconfig change" and "did the configmap change" since both codepaths basically do the same thing, it's just duplicated code.

PR #292, opened 2019-08-16
manifests: the openshift-sdn type isn't actually required

squeed:
This makes it impossible to change the network configuration from what's in the cluster.

@weliang1 this should fix the issue you saw.


danwinship:
a quick scan suggests that everything else marked as `required` actually is required.
/lgtm

PR #293, opened 2019-08-20
Bug 1743587: Adding cni-version to multus daemonset yaml

alexanderConstantinescu:
A backport of PR [291](https://github.com/openshift/cluster-network-operator/pull/291)

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -58,6 +58,7 @@ spec:
         - "--multus-kubeconfig-file-host=/etc/kubernetes/cni/net.d/multus.d/multus.kubeconfig"
         - "--namespace-isolation=true"
         - "--multus-log-level=verbose"
+        - "--cni-version=0.3.1"
         securityContext:
           privileged: true
         volumeMounts:


alexanderConstantinescu:
/retitle Backport 1743587: Adding cni-version to multus daemonset yaml

alexanderConstantinescu:
/retitle Bug 1743587: Adding cni-version to multus daemonset yaml

eparis:
This just requires lgtm/approval from the engineering team.  Does this need to merge with https://github.com/openshift/multus-cni/pull/20 ? If one merged without the other will something break?  (and if yes, how did they both pass ci?)

danwinship:
> Does this need to merge with [openshift/multus-cni#20](https://github.com/openshift/multus-cni/pull/20) ? If one merged without the other will something break?

Both are needed, but they don't need to merge in sync; if the multus PR merges first, it will just add a command-line option that isn't getting used. If this merges first then the multus setup container will log a warning about an unrecognized command-line option but will otherwise behave exactly like it did before this PR.

alexanderConstantinescu:
FYI: [multus PR](https://github.com/openshift/multus-cni/pull/20) has been approved, so this should be "good to go".

PR #295, opened 2019-08-20
Adds .svc and .svc.cluster.local to default noProxy

danehans:
Fixes Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1743507

Previously, only the service cidr was added to the default noProxy list. This does not help when pods communicate with a service using a DNS name. This PR adds `.svc` and `svc.cluster.local` to the default noProxt list.

/assign @knobunc @bparees

### DIFF
--- a/pkg/util/proxyconfig/merge.go
+++ b/pkg/util/proxyconfig/merge.go
@@ -23,6 +23,8 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 	set := sets.NewString(
 		"127.0.0.1",
 		"localhost",
+		".svc",
+		".cluster.local",
 	)
 
 	if len(infra.Status.APIServerURL) > 0 {


pkg/util/proxyconfig/merge.go
```diff
@@ -23,6 +23,8 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 	set := sets.NewString(
 		"127.0.0.1",
 		"localhost",
+		".svc",
+		".svc.cluster.local",
```
bparees: did you confirm these with @sdodson ?

sdodson:
```suggestion
		".cluster.local",
```
We had `.cluster.local` rather than `.svc.cluster.local`
https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_facts/library/openshift_facts.py#L852-L863

danehans: @bparees @sdodson I updated the PR to use `.cluster.local` instead of  `.svc.cluster.local`.

danehans:
An example `proxy.status.noProxy` with this PR:
```
noProxy: .cluster.local,.svc,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.dhansen.devcluster.openshift.com,api.dhansen.devcluster.openshift.com,etcd-0.dhansen.devcluster.openshift.com,etcd-1.dhansen.devcluster.openshift.com,etcd-2.dhansen.devcluster.openshift.com,example.com,localhost
```

knobunc:
/approve
Thanks @danehans

pkg/util/proxyconfig/merge.go
```diff
@@ -23,6 +23,8 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 	set := sets.NewString(
 		"127.0.0.1",
 		"localhost",
+		".svc",
+		".cluster.local",
```
runcom: This must be done in the installer as well

runcom: cc @abhinavdahiya related to https://bugzilla.redhat.com/show_bug.cgi?id=1744532#c13

runcom: attempted to fix this here https://github.com/openshift/installer/pull/2257

PR #296, opened 2019-08-20
Refactors noProxy merge func

danehans:
Previously `MergeUserSystemNoProxy()` was including specific api-server and etcd dns names when creating the default noProxy list. This PR uses the cluster's `baseDomain` from the openshift `config.DNS` API as a noProxy entry. For example:
```
$ oc get proxy/cluster -o yaml
<SNIP>
status:
  httpProxy: http://admin:admin@35.196.128.173:3128
  noProxy: .dhansen.devcluster.openshift.com,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,example.com,localhost
```

This approach will cover etcd and api-server names and any additional names added to the cluster's base domain. For example, the cluster-ingress-operator creates a wildcard alias record for routes. For example:
```
*.apps.dhansen.devcluster.openshift.com.
```
These routes should not be proxied. 

/assign @bparees @knobunc

### DIFF
--- a/pkg/util/proxyconfig/merge.go
+++ b/pkg/util/proxyconfig/merge.go
@@ -2,8 +2,6 @@ package proxyconfig
 
 import (
 	"fmt"
-	"net/url"
-	"strconv"
 	"strings"
 
 	"github.com/ghodss/yaml"
@@ -19,31 +17,12 @@ import (
 // string of noProxy settings. If no user supplied noProxy settings are
 // provided, a comma-separated string of cluster-wide noProxy settings
 // are returned.
-func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network, cluster *corev1.ConfigMap) (string, error) {
+func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network,
+	dns *configv1.DNS, cluster *corev1.ConfigMap) (string, error) {
 	set := sets.NewString(
 		"127.0.0.1",
 		"localhost",
 	)
-
-	if len(infra.Status.APIServerURL) > 0 {
-		apiServerURL, err := url.Parse(infra.Status.APIServerURL)
-		if err != nil {
-			return "", fmt.Errorf("failed to parse api server url")
-		}
-		set.Insert(apiServerURL.Hostname())
-	} else {
-		return "", fmt.Errorf("api server url missing from infrastructure config '%s'", infra.Name)
-	}
-
-	if len(infra.Status.APIServerInternalURL) > 0 {
-		internalAPIServer, err := url.Parse(infra.Status.APIServerInternalURL)
-		if err != nil {
-			return "", fmt.Errorf("failed to parse internal api server internal url")
-		}
-		set.Insert(internalAPIServer.Hostname())
-	} else {
-		return "", fmt.Errorf("internal api server url missing from infrastructure config '%s'", infra.Name)
-	}
 
 	if len(network.Status.ServiceNetwork) > 0 {
 		set.Insert(network.Status.ServiceNetwork[0])
@@ -51,11 +30,14 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 		return "", fmt.Errorf("serviceNetwork missing from network '%s' status", network.Name)
 	}
 
+	if len(dns.Spec.BaseDomain) > 0 {
+		set.Insert("." + dns.Spec.BaseDomain)
+	} else {
+		return "", fmt.Errorf("dns '%s' is missing baseDomain", dns.Name)
+	}
+
 	// TODO: This will be flexible when master machine management is more dynamic.
 	type installConfig struct {
-		ControlPlane struct {
-			Replicas string `json:"replicas"`
-		} `json:"controlPlane"`
 		Networking struct {
 			MachineCIDR string `json:"machineCIDR"`
 		} `json:"networking"`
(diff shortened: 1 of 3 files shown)


pkg/util/proxyconfig/merge.go
```diff
@@ -19,43 +17,27 @@ import (
...
-	}
-
 	if len(network.Status.ServiceNetwork) > 0 {
 		set.Insert(network.Status.ServiceNetwork[0])
 	} else {
 		return "", fmt.Errorf("serviceNetwork missing from network '%s' status", network.Name)
 	}
 
+	if len(dns.Spec.BaseDomain) > 0 {
+		set.Insert("." + dns.Spec.BaseDomain)
```
bparees: @deads2k here's where we're adding the cluster domain to noproxy. I believe we made this choice based on what @sdodson found we were doing in 3.x.

bparees: cc @spadgett

sdodson: This would be `.cluster.local` or `.dhansen.devcluster.openshift.com`? The former is what we had added in 3.x but in 4.x since we're provisioning hosts into `.dhansen.devcluster.openshift` that zone it probably makes sense to add that as well.

spadgett:
cc @enj

pkg/util/proxyconfig/merge.go
bparees: `.dhansen.devcluster.openshift.com`, i believe.  `.cluster.local` is being added via hardcode elsewhere.

bparees: @spadgett i would expect that this routing/proxy question would have been an issue in 3.x as well, so either:

1) we were noproxying the domain in 3.x
2) customers were manually adding it
3) console traffic was going through the customer proxies without issue (which presumably means they were doing so without any additional CAs)
4) ??

sdodson: It would've been either 2 or 3.

spadgett: For 3.x, the browser talked directly to the OAuth server. We didn't make requests to the OAuth server from our pod.

The exception is the 3.11 admin console, which is the same code base as the 4.x console. We let users customize the CA for talking to the OAuth server in the admin console, though. I'm not familiar with how proxy worked in 3.x, but presumably it was possible to add HTTP proxy env vars directly to the admin console deployment.

deads2k:
/hold

This is being discussed.  It's not clear that the ingress subdomain can be assumed to be in-cluster.  Just don't want to merge in the interim.

pkg/util/proxyconfig/merge.go
danehans: @sdodson @bparees that is correct, this PR uses `.${CLUSTER_NAME}.${BASE_DOMAIN}` as a default for no proxy. In my cluster, this is `.dhansen.devcluster.openshift.com`. https://github.com/openshift/cluster-network-operator/pull/295 added `.svc` and `.cluster.local` as a no proxy default.

danehans:
`"Cluster did not complete upgrade: timed out waiting for the condition: Working towards registry.svc.ci.openshift.org/ci-op-v419r356/release@sha256:3387479f940a92d4b2ef992fc34d7bbef8f54ebfbb15c312b327e1857a77aa0d: downloading update",`

I believe a PR landed to fix this issue.
/test e2e-aws-upgrade

pkg/util/proxyconfig/merge.go
bparees: @danehans per the email thread discussion, let's drop the cluster domain from the noproxy autogenerated list and move forward w/ this PR.

we'll resolve the fallout (e.g. console being broken) on a case by case basis.

danehans:
```
level=error msg="  on ../tmp/openshift-install-065943236/vpc/master-elb.tf line 1, in resource \"aws_lb\" \"api_internal\":"
level=error msg="   1: resource \"aws_lb\" \"api_internal\" {"
```

/test e2e-aws-upgrade

pkg/util/proxyconfig/merge.go
```diff
@@ -19,43 +17,27 @@ import (
 // string of noProxy settings. If no user supplied noProxy settings are
 // provided, a comma-separated string of cluster-wide noProxy settings
 // are returned.
-func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network, cluster *corev1.ConfigMap) (string, error) {
+func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network,
```
abhinavdahiya: not usable by installer as you have the configmap to get install-config here https://github.com/openshift/cluster-network-operator/pull/296/files#diff-b9dde8483f6b2bbdb5b0a7695c57743aR40

danehans:
@deads2k I am seeing the following auth operator error trying to create a proxy-enabled cluster using IPI:

```
E0916 21:24:49.420417       1 controller.go:129] {AuthenticationOperator2 AuthenticationOperator2} failed with: error checking current version: unable to check route health: failed to GET route: proxyconnect tcp: x509: certificate signed by unknown authority
```

https://github.com/openshift/cluster-authentication-operator/pull/194 attempts to fix this error. However, I don't understand why this call should be proxied. Any updates on https://github.com/openshift/cluster-network-operator/pull/296#issuecomment-523875516. Otherwise, should this PR be refactored to only include `noProxy` for specific routes instead of the entire cluster domain?

danehans:
Per https://github.com/openshift/cluster-network-operator/pull/296#issuecomment-523875516 ingress domain will not be added to default no proxy.

PR #299, opened 2019-08-22
Alternate configuration directory + 00-multus.conf regeneration

dougbtv:
This is a continuation of #218 and includes @squeed 's commit therein.

In order to address the race condition where openshift-sdn CNI config is loaded by crio/ocicni first and then does not read the `00-multus.conf` file as generated by the entrypoint.

This implements a parameters to enter a loop to detect the presence/absence of the openshift-sdn CNI configuration and a flag to restart CRIO on generation of the `00-multus.conf`. This does **not** override the `"name":` field (e.g. using `--override-network-name=true` which would copy that from source CNI config to generated `00-multus.conf`)

*NOTE*: This may not be necessary if we have the intended changes to ocicni. So I intend to put this on a hold.

In reference to [this BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1732598).

### DIFF
--- a/bindata/network/kuryr/004-daemon.yaml
+++ b/bindata/network/kuryr/004-daemon.yaml
@@ -89,10 +89,10 @@ spec:
       volumes:
       - name: bin
         hostPath:
-          path: /var/lib/cni/bin
+          path: {{.CNIBinDir}}
       - name: net-conf
         hostPath:
-          path: /etc/kubernetes/cni/net.d
+          path: {{.CNIConfDir}}
       - name: config-volume
         configMap:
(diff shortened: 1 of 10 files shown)


dougbtv:
/retest

New changes up to Multus entrypoint to make for a one-shot restart of CRIO in this PR: https://github.com/openshift/multus-cni/pull/23

dougbtv:
Also rebased.

dougbtv:
/hold cancel

@squeed & @dcbw when you're ready, see if this looks alright. Thanks guys.

PR #301, opened 2019-08-27
Set version on network cluster operator with kuryr SDN

luis5tb:
This PR ensures the VERSION field is set on the network cluster
operator config object by adding it to the different kuryr
components, i.e., the kuryr-controller and admission-controller
deployments, and kuryr-cni daemon set.


dulek:
/lgtm

It's worth saying that this fixes setting the network operator to "Available". This is really important change.

PR #302, opened 2019-08-28
Bug 1746360: openshift-sdn: allow empty mode

squeed:
We can fill a sensible default anyways. So this is just confusing and user-unfriendly.


squeed:
No, we always validate before filling defaults. This may be worth a rethink, but that's not the current shape of the code.

PR #303, opened 2019-08-28
Bug 1737026: Use retry mechanism for fixing updates on stale objects in cache

alexanderConstantinescu:
Hi

Following PR [279](https://github.com/openshift/cluster-network-operator/pull/279) the status manager is now thread safe. However, we still face issues concerning stale objects being retrieved from the API server when performing `status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)`. This means that in between two updates the API server does not have enough time to update its cache, so the second update is done on a stale object leading to logs such as: `Failed to update ClusterOperator "network": Operation cannot be fulfilled on clusteroperators.config.openshift.io "network": the object has been modified; please apply your changes to the latest version and try again` which can throw us off. 

This PR introduces the retry mechanism from `client-go`, which should fix this issue.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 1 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -53,72 +55,73 @@ func New(client client.Client, name, version string) *StatusManager {
...
-			log.Printf("Failed to create ClusterOperator %q: %v", co.Name, err)
-		} else {
-			log.Printf("Created ClusterOperator with conditions:\n%s", string(buf))
+		buf, err := yaml.Marshal(co.Status.Conditions)
+		if err != nil {
+			buf = []byte(fmt.Sprintf("(failed to convert to YAML: %s)", err))
+		}
+		if isNotFound {
+			if err := status.client.Create(context.TODO(), co); err != nil {
+				return fmt.Errorf("Failed to create ClusterOperator %q: %v", co.Name, err)
```
danwinship: `RetryOnConflict` only works if you return the actual client error object. You should do the wrapping outside the `RetryOnConflict` call. (It's ok if the wrapped errors become slightly-less precise.)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -53,72 +55,73 @@ func New(client client.Client, name, version string) *StatusManager {
...
+			}
+			return fmt.Errorf("Created ClusterOperator with conditions:\n%s", string(buf))
```
danwinship: er, no, don't return an error on success. Figure out another way to handle the logging

pkg/controller/statusmanager/status_manager.go
```diff
@@ -113,9 +114,11 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
 			log.Printf("Created ClusterOperator with conditions:\n%s", string(buf))
 		}
 	} else {
-		err = status.client.Status().Update(context.TODO(), co)
-		if err != nil {
-			log.Printf("Failed to update ClusterOperator %q: %v", co.Name, err)
+		resultErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
squeed: Is this going to work? I don't think so. You need to re-fetch `co` and re-update it.

alexanderConstantinescu: > Is this going to work? I don't think so.

Why is that?

danwinship: https://github.com/kubernetes/kubernetes/pull/82284/files

danwinship: If that doesn't answer your question then let me know so I can make further improvements there...

alexanderConstantinescu: That makes it more clear. Thanks for that PR

pkg/controller/statusmanager/status_manager.go
```diff
@@ -53,72 +54,76 @@ func New(client client.Client, name, version string) *StatusManager {
 
 // Set updates the ClusterOperator.Status with the provided conditions
 func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...configv1.ClusterOperatorStatusCondition) {
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
+		err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
+		isNotFound := errors.IsNotFound(err)
+		if err != nil && !isNotFound {
+			log.Printf("Failed to get ClusterOperator %q: %v", status.name, err)
+			return err
```
danwinship: you're logging `err` outside of the `RetryOnConflict`, so if you log here too we're going to get double error messages. I think you can just remove all the `log.Printf` calls inside the loop. The single error message outside the loop is fine.

alexanderConstantinescu: I thought it would be good to get a lower layer log in case something fails - i.e if it's the `Get`,`Create`,`Update`. Besides, today we log `Updated ClusterOperator with conditions` for every successful update, is removing that acceptable?

danwinship: Definitely keep the non-error logging ("Updated/Created ClusterOperator", etc).

For the errors, they shouldn't really happen, and I think there should be enough information to figure them out without needing further clarification anyway. eg, if it's "Not Found" then that's obviously an error on the "Get". And if it's "no route to host" or "timed out", it doesn't really matter which operation it occurred on...

alexanderConstantinescu: Ok, just updated the PR taking into account your input.

PR #305, opened 2019-08-29
Bug 1744532: Adds machineCIDR noProxy support for all platforms

danehans:
Previously, machineCIDR was only being added to the default noProxy list for clusters installed in cloud provider environments. According to https://bugzilla.redhat.com/show_bug.cgi?id=1744532, machineCIDR is used for vSphere and bare metal installations too. This PR adds machineCIDR to the default noProxy list for all installation types, provides consistency with PR https://github.com/openshift/installer/pull/2286

### DIFF
--- a/pkg/util/proxyconfig/no_proxy.go
+++ b/pkg/util/proxyconfig/no_proxy.go
@@ -20,11 +20,30 @@ import (
 // provided, a comma-separated string of cluster-wide noProxy settings
 // are returned.
 func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network, cluster *corev1.ConfigMap) (string, error) {
+	// TODO: This will be flexible when master machine management is more dynamic.
+	type installConfig struct {
+		ControlPlane struct {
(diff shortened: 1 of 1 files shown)


pkg/util/proxyconfig/merge.go
```diff
@@ -20,11 +20,30 @@ import (
...
+	type installConfig struct {
+		ControlPlane struct {
+			Replicas string `json:"replicas"`
+		} `json:"controlPlane"`
+		Networking struct {
+			MachineCIDR string `json:"machineCIDR"`
+		} `json:"networking"`
+	}
+	var ic installConfig
+	data, ok := cluster.Data["install-config"]
```
squeed: Do all clusters have an install-config? Including UPI clusters?

danehans: @squeed yes, all install types include an install-config with machineCIDR. See the [user docs ](https://github.com/openshift/installer/blob/master/docs/user) as a reference. A similar installer [PR](https://github.com/openshift/installer/pull/2286) was just merged.

PR #306, opened 2019-08-30
Delete current related objects not been rendered by controller

rcarrillocruz:
As that means they are from an older version, thus no longer needed.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -16,7 +16,10 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 
@@ -39,6 +42,7 @@ type StatusManager struct {
 	sync.Mutex
 
 	client  client.Client
+	mapper  meta.RESTMapper
 	name    string
 	version string
 
@@ -49,8 +53,53 @@ type StatusManager struct {
 	relatedObjects []configv1.ObjectReference
 }
 
-func New(client client.Client, name, version string) *StatusManager {
-	return &StatusManager{client: client, name: name, version: version}
+func New(client client.Client, mapper meta.RESTMapper, name, version string) *StatusManager {
+	return &StatusManager{client: client, mapper: mapper, name: name, version: version}
+}
+
+// deleteRelatedObjects checks for related objects attached to ClusterOperator and deletes
+// whatever is not been rendered from manifests. This is a mechanism to cleanup objects
+// that are no longer needed and are probably present from a previous version
+func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.ClusterOperator) {
+	if status.relatedObjects == nil {
+		return
+	}
+
+	for _, currentObj := range co.Status.RelatedObjects {
+		var found bool = false
+		for _, renderedObj := range status.relatedObjects {
+			found = reflect.DeepEqual(currentObj, renderedObj)
+
+			if found {
+				break
+			}
+		}
+		if !found {
+			gvr := schema.GroupVersionResource{
+				Group:    currentObj.Group,
+				Resource: currentObj.Resource,
+			}
+			gvk, err := status.mapper.KindFor(gvr)
+			if err != nil {
+				log.Printf("Error getting GVK of object for deletion: %v", err)
+				status.relatedObjects = append(status.relatedObjects, currentObj)
+				continue
+			}
+			log.Printf("Detected related object with GVK %+v, namespace %v and name %v not rendered by manifests, deleting...", gvk, currentObj.Namespace, currentObj.Name)
+			objToDelete := &uns.Unstructured{}
+			objToDelete.SetName(currentObj.Name)
+			objToDelete.SetNamespace(currentObj.Namespace)
+			objToDelete.SetGroupVersionKind(gvk)
+			err = status.client.Delete(context.TODO(), objToDelete, client.PropagationPolicy("Background"))
+			if err != nil {
+				log.Printf("Error deleting related object: %v", err)
+				if !errors.IsNotFound(err) {
+					status.relatedObjects = append(status.relatedObjects, currentObj)
+				}
+				continue
+			}
+		}
+	}
 }
 
 // Set updates the ClusterOperator.Status with the provided conditions
@@ -64,6 +113,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
 		}
 
 		oldStatus := co.Status.DeepCopy()
+		status.deleteRelatedObjectsNotRendered(co)
 		co.Status.RelatedObjects = status.relatedObjects
 
 		if reachedAvailableLevel {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -11,7 +11,7 @@ var AddToManagerFuncs []func(manager.Manager, *statusmanager.StatusManager) erro
 
 // AddToManager adds all Controllers to the Manager
 func AddToManager(m manager.Manager) error {
-	s := statusmanager.New(m.GetClient(), "network", operatorversion.Version)
+	s := statusmanager.New(m.GetClient(), m.GetRESTMapper(), "network", operatorversion.Version)
 
 	for _, f := range AddToManagerFuncs {
 		if err := f(m, s); err != nil {
--- a/pkg/controller/statusmanager/status_manager_test.go
+++ b/pkg/controller/statusmanager/status_manager_test.go
@@ -10,7 +10,10 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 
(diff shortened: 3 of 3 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -184,6 +185,29 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName)
...
+				break
+			}
+		}
+		if !found {
+			obj_to_delete := &uns.Unstructured{}
+			key := types.NamespacedName{
+				Name:      status_obj.Name,
+				Namespace: status_obj.Namespace,
+			}
+			err := status.client.Get(context.TODO(), key, obj_to_delete)
```
danwinship: This won't work; the object you pass in has to be a proper API object, so the client can introspect its type and figure out what kind of object to get. (Just specifying the name isn't enough, because eg, it's totally valid to have a Service and Pod with the same name in the same namespace.) You should be able to find the right golang type with the `RESTMapper`... it's basically the opposite of the operation you perform in `operconfig_controller.go` to fill in `status_obj.Group` and `.Resource`. You might need to use something in `reflect` to actually construct an empty object of that type. (It might make sense at the same time to change `StatusManager.SetRelatedObjects` to take a `[]interface{}` of API objects and construct the array of `ObjectReference`s itself. That way all of the `RESTMapper` code is in `StatusManager` rather than being split between there and the operator config controller.)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -186,6 +190,33 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName)
 func (status *StatusManager) SetRelatedObjects(relatedObjects []configv1.ObjectReference) {
 	status.Lock()
 	defer status.Unlock()
+
+	var found bool = false
+	for _, status_obj := range status.relatedObjects {
```
danwinship: You don't want to compare against the current value of `status.relatedObjects`, you want to compare against the `RelatedObjects` field of the actual operator status object. So you need to move all of this from `SetRelatedObjects` to `set`

pkg/controller/statusmanager/status_manager.go
```diff
@@ -186,6 +190,33 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName)
...
+			gvk, err := status.mapper.KindFor(gvr)
+			if err != nil {
+				log.Printf("Error getting GVK of object for deletion: %v", err)
+				continue
+			}
+			obj_to_delete := &uns.Unstructured{}
+			obj_to_delete.SetName(status_obj.Name)
+			obj_to_delete.SetNamespace(status_obj.Namespace)
+			obj_to_delete.SetGroupVersionKind(gvk)
+			status.client.Delete(context.TODO(), obj_to_delete)
```
danwinship: log an error on error

squeed: If possible, we want to do a Background delete here. That means that we won't wait for descendant objects (e.g. pods of deployments) to delete - so we don't block the operator.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -186,6 +190,33 @@ func (status *StatusManager) SetDeployments(deployments []types.NamespacedName)
...
+	var found bool = false
+	for _, status_obj := range status.relatedObjects {
+		for _, obj := range relatedObjects {
+			found = reflect.DeepEqual(status_obj, obj)
+
+			if found {
+				break
+			}
+		}
+		if !found {
```
squeed: Definitely log something before deleting. Something like "deleting stale object XXX"

pkg/controller/statusmanager/status_manager.go
```diff
@@ -63,6 +67,39 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
...
+				gvk, err := status.mapper.KindFor(gvr)
+				if err != nil {
+					log.Printf("Error getting GVK of object for deletion: %v", err)
+					continue
+				}
+				obj_to_delete := &uns.Unstructured{}
+				obj_to_delete.SetName(current_obj.Name)
+				obj_to_delete.SetNamespace(current_obj.Namespace)
+				obj_to_delete.SetGroupVersionKind(gvk)
+				status.client.Delete(context.Background(), obj_to_delete)
```
squeed: 2 things.
- you still need to `err = Delete()`
- This isn't going to trigger a background delete. We might not have the ability to request a background delete via our client, so let's just ignore that.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -63,6 +67,39 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
...
+				if err != nil {
+					log.Printf("Error deleting related object: %v", err)
```
squeed: Put the object back in Related so we don't lose track of it. As long as the error isn't something like "already deleted"

pkg/controller/statusmanager/status_manager.go
squeed: Also, a quick check shows that we'll have the ability to issue a Background delete when 336 merges.

rcarrillocruz: 336 merged, and added missing err var, marking as resolved

pkg/controller/statusmanager/status_manager.go
```diff
@@ -63,6 +67,39 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
...
+				gvk, err := status.mapper.KindFor(gvr)
+				if err != nil {
+					log.Printf("Error getting GVK of object for deletion: %v", err)
+					continue
+				}
+				obj_to_delete := &uns.Unstructured{}
+				obj_to_delete.SetName(current_obj.Name)
+				obj_to_delete.SetNamespace(current_obj.Namespace)
+				obj_to_delete.SetGroupVersionKind(gvk)
+				err = status.client.Delete(context.Background(), obj_to_delete)
```
squeed: This still doesn't do background deletes. They are a Delete option - check the signature of status.client.Delete. The `context.Background()` here is incorrect - it should be `context.TODO`.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,43 @@ type StatusManager struct {
 	relatedObjects []configv1.ObjectReference
 }
 
-func New(client client.Client, name, version string) *StatusManager {
-	return &StatusManager{client: client, name: name, version: version}
+func New(client client.Client, mapper meta.RESTMapper, name, version string) *StatusManager {
+	return &StatusManager{client: client, mapper: mapper, name: name, version: version}
+}
+
+func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.ClusterOperator) {
```
squeed: A bit of docstring would be nice.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,43 @@ type StatusManager struct {
...
+	var found bool = false
```
squeed: minor nit: I'd rather see this variable defined inside the while loop, since it's not meant to carry value "between" executions of the outer-most loop. Defining it outside is confusing to me.

I understand that it's still correct. Just a bit confusing.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,43 @@ type StatusManager struct {
...
+	for _, current_obj := range co.Status.RelatedObjects {
+		for _, rendered_obj := range status.relatedObjects {
+			found = reflect.DeepEqual(current_obj, rendered_obj)
+
+			if found {
+				break
+			}
+		}
+		if !found {
+			log.Printf("Detected related object not rendered by manifests, deleting...")
```
squeed: Print the GVK + namespace/name, so we know what's being deleted.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -63,6 +102,7 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
 			return err
 		}
 
+		status.deleteRelatedObjectsNotRendered(co)
```
squeed: You need to call this after the DeepCopy since you're mutating RelatedObjects.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,48 @@ type StatusManager struct {
...
+					break
+				}
+			}
+			if !found {
+				gvr := schema.GroupVersionResource{
+					Group:    current_obj.Group,
+					Resource: current_obj.Resource,
+				}
+				gvk, err := status.mapper.KindFor(gvr)
+				log.Printf("Detected related object with GVK %+v, namespace %v and name %v not rendered by manifests, deleting...", gvk, current_obj.Namespace, current_obj.Name)
```
squeed: whoops, silly issue: this log line should be below the `if err != nil ...`

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,48 @@ type StatusManager struct {
...
+				if err != nil {
+					log.Printf("Error getting GVK of object for deletion: %v", err)
```
squeed: In this case, should we be re-adding this to relatedObjects? Probably.

rcarrillocruz: Not sure I follow.
If there's an error, the status.relatedObjects never changes, if we re-add the object wouldn't we have duplicates?

squeed: Right - my read of the code is that this would be equivalent to a failure to delete. This means that we've now lost the reference to this object, right?

To be fair, this is very unlikely, so I don't think we really need to worry about it.

rcarrillocruz: Yeah, but what relatedObjects hold are not really references to objects. They are instances of configv1 ObjectReference, which are merely a struct containing Group/Resource/Name/Namespace.
So even if we had a failure on delete on the real object, we would still have a valid relatedObject for it, which is just a representation, not a reference perse.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,49 @@ type StatusManager struct {
...
-	return &StatusManager{client: client, name: name, version: version}
+func New(client client.Client, mapper meta.RESTMapper, name, version string) *StatusManager {
+	return &StatusManager{client: client, mapper: mapper, name: name, version: version}
+}
+
+// deleteRelatedObjects checks for related objects attached to ClusterOperator and deletes
+// whatever is not been rendered from manifests. This is a mechanism to cleanup objects
+// that are no longer needed and are probably present from a previous version
+func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.ClusterOperator) {
+	if status.relatedObjects != nil {
```
danwinship: don't indent the whole function; return early on the exceptional case:

    if status.relatedObjects == nil {
            return
    }
    ...

(or move the check out to `set()` and only call this function when `relatedObjects` is non-empty)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,49 @@ type StatusManager struct {
...
+		for _, current_obj := range co.Status.RelatedObjects {
```
danwinship: We don't usually use underscores in names in go. `currentObj` (likewise `renderedObj`, `objToDelete`, etc)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,49 @@ type StatusManager struct {
...
+			}
+			if !found {
+				gvr := schema.GroupVersionResource{
+					Group:    current_obj.Group,
+					Resource: current_obj.Resource,
+				}
+				gvk, err := status.mapper.KindFor(gvr)
+				if err != nil {
+					log.Printf("Error getting GVK of object for deletion: %v", err)
+					co.Status.RelatedObjects = append(co.Status.RelatedObjects, current_obj)
```
danwinship: `co.Status.RelatedObjects` gets overwritten as soon as you return from this function; you need to append it to `status.relatedObjects`.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,51 @@ type StatusManager struct {
...
+		}
+		if !found {
+			gvr := schema.GroupVersionResource{
+				Group:    currentObj.Group,
+				Resource: currentObj.Resource,
+			}
+			gvk, err := status.mapper.KindFor(gvr)
+			if err != nil {
+				log.Printf("Error getting GVK of object for deletion: %v", err)
+				co.Status.RelatedObjects = append(status.relatedObjects, currentObj)
```
danwinship: `status.relatedObjects = append(status.relatedObjects, currentObj)`
(and likewise below)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,8 +53,51 @@ type StatusManager struct {
...
+				continue
+			}
+			log.Printf("Detected related object with GVK %+v, namespace %v and name %v not rendered by manifests, deleting...", gvk, currentObj.Namespace, currentObj.Name)
+			objToDelete := &uns.Unstructured{}
+			objToDelete.SetName(currentObj.Name)
+			objToDelete.SetNamespace(currentObj.Namespace)
+			objToDelete.SetGroupVersionKind(gvk)
+			err = status.client.Delete(context.TODO(), objToDelete, client.PropagationPolicy("Background"))
+			if err != nil {
+				log.Printf("Error deleting related object: %v", err)
```
squeed: if `apierrors.IsNotFound`, then continue without re-adding the object.

PR #309, opened 2019-09-09
Kuryr: Set region_name in kuryr.conf

dulek:
With recent change to kuryr-lib it is possible to specify OpenStack
cloud region name that should be used by Kuryr. This commit implements
that by setting that option in kuryr.conf.


luis5tb:
Tested it and it works! Thanks!

PR #311, opened 2019-09-09
Bug 1732598: Alternate CNI config dir

dougbtv:
#299 has been used to work around the CRIO + Multus CNI configuration issues. However, this caused collateral damage to Kuryr and OpenShift upgrades. #299 was reverted with #310 -- this PR adds the commit for the alternate CNI configuration direct from #299 back, as well as a commit for updating the Multus entrypoint parameters.

**UPDATE**: Notably different from #299 is that this *no longer performs a CRIO restart*.

----

It's been observed that with #299 in place, openshift-sdn and Kuryr were having trouble with write operations to the directories within `/run` -- notably #299 uses `/var/run/multus/cni/net.d` as an alternate configuration directory.

Additionally, some selinux logs were gathered both with the denials, and also with an experiment where selinux permissive mode was enabled, those logs are available @ https://pastebin.com/iGvBec6u

Luis Tomas has also noted that it "seems the process being denied has container_t label, while the directory has var_run_t" with:

```
scontext=system_u:system_r:container_t:s0:c0,c912 tcontext=system_u:object_r:var_run_t:s0
```

As an initial attempt: This moves the alternate configuration directory inside the `/etc/.../net.d` into a new directory named `sourceconfig.d`.

Further from that, I'm also looking into how I might address the selinux options, however, I need input from those that wield greater selinux powers than I.

### DIFF
--- a/bindata/network/kuryr/004-daemon.yaml
+++ b/bindata/network/kuryr/004-daemon.yaml
@@ -89,10 +89,10 @@ spec:
       volumes:
       - name: bin
         hostPath:
-          path: /var/lib/cni/bin
+          path: {{.CNIBinDir}}
       - name: net-conf
         hostPath:
-          path: /etc/kubernetes/cni/net.d
+          path: {{.CNIConfDir}}
       - name: config-volume
         configMap:
           name: kuryr-config
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -44,6 +44,7 @@ spec:
         command: ["/entrypoint.sh"]
         args:
         - "--multus-conf-file=auto"
+        - "--multus-autoconfig-dir=/host/var/run/multus/cni/net.d"
         - "--multus-kubeconfig-file-host=/etc/kubernetes/cni/net.d/multus.d/multus.kubeconfig"
         - "--namespace-isolation=true"
         - "--multus-log-level=verbose"
@@ -55,8 +56,10 @@ spec:
         securityContext:
           privileged: true
         volumeMounts:
-        - name: cni
+        - name: system-cni-dir
           mountPath: /host/etc/cni/net.d
+        - name: multus-cni-dir
+          mountPath: /host/var/run/multus/cni/net.d
         - name: cnibin
           mountPath: /host/opt/cni/bin
         env:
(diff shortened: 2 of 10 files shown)


dougbtv:
/retest 

Error looked flakey for e2e-aws-upgrade with:

```
level=error msg="When applying changes to module.bootstrap.aws_s3_bucket.ignition, provider"
level=error msg="\"aws\" produced an unexpected new value for was present, but now absent."
```

e2e-aws results were for previous commit.

dougbtv:
On CI results @ https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/311/pull-ci-openshift-cluster-network-operator-master-e2e-aws-upgrade/721

I'm seeing an error like:

```
 healthcheck.go:92] SDN healthcheck unable to reconnect to OVS server: dial unix /var/run/openvswitch/db.sock: connect: no such file or directory
```

Now, here's the interesting part... I moved the alternate CNI configuration dir to /etc, right...
But! There's still a mount for /run for Multus. It's used for the CRIO restart. I wonder if that's the culprit, and not the CNI configuration directory itself, maybe been chasing up the wrong tree.

Thinking on this one, but, maybe there's a more surgical way to restart CRIO without having to mount all of `/run`.

dulek:
@MaysaMacedo will check if this PR is still problematic for Kuryr.

MaysaMacedo:
> @MaysaMacedo will check if this PR is still problematic for Kuryr.

We are still having permission problems when using:
```
net-conf:
    Path:          /etc/kubernetes/cni/net.d/sourceconfig.d
```

dulek:
> > @MaysaMacedo will check if this PR is still problematic for Kuryr.
> 
> We are still having permission problems when using:
> 
> ```
> net-conf:
>     Path:          /etc/kubernetes/cni/net.d/sourceconfig.d
> ```

I'm confirming that, my installation suffered that as well. Please note that only one kuryr-cni pod was affected and restarting it seemed to get rid of the issue.

dougbtv:
Updated with a commit to remove the CRIO restart, and not wholesale mount `/run`.

dougbtv:
/retitle Bug 1732598: Alternate CNI config dir

danwinship:
Commit message on the second commit seems wrong now.

So, the expected behavior of this PR now is that it should 100% fix the multus-vs-openshift-sdn-vs-CRI-O problem on *new installs*, but it does not necessarily fix the problem with upgrades, which will be fixed by a later ocicni fix?

dougbtv:
@danwinship -- that is correct, this should work during installation, however, it's anticipated to fail on upgrade until we have the ocicni changes, you got it.

Thanks for the feedback on the commit message, I've given it an update.

dougbtv:
Unsure how to tell when the CRIO changes that this also requires make it to the build, but... in the meanwhile, I'll at least give it a:

/retest

PR #312, opened 2019-09-10
Bug 1750991: Fixes http readinessEndpoint validation for httpsProxy

danehans:
Previously, a readiness endpoint with a https url scheme would be considered invalid if `httpsProxy` used a http scheme. It's common for clients to be configured with https proxies that contain an http url scheme. This PR updates `readinessEndpoint` validation to allow for this use case and also simplified the name of the http/https scheme constants.

### DIFF
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	proxyHTTPScheme  = "http"
-	proxyHTTPSScheme = "https"
+	schemeHTTP  = "http"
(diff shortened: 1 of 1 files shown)


squeed:
As a matter of taste, disambiguating between `httpScheme` and `httpsScheme` is pretty hard for me. Given that these constants are, well, constant, I personally prefer to skip the variables and just spell them every time.

I'll leave that up to you. But if you keep the variables, I'd prefer `schemeHTTP` and `schemeHTTPS`, which are much easier to read.

danehans:
@squeed I updated the constants.

PR #319, opened 2019-09-17
Bug 1753014: Kuryr: Fix kuryr bootstrap when lbaas API has ERROR status

MaysaMacedo:
In case the API lbaas gets to a provisiong status with ERROR, the
reconciliation is triggered and the same lbaas with ERROR status
is used on the creation of the next lbaas resources, e.g. pools,
listeners and members. The operator will never finish Kuryr
bootstraping phase.

This commit fixes the issue by deleting the lbaas in case it's
with ERROR status and creating a new one.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -441,6 +441,34 @@ func ensureOpenStackSgRule(client *gophercloud.ServiceClient, sgId, remotePrefix
 	return nil
 }
 
+// Delete a LoadBalancer and all associated resources
+func deleteOpenStackLb(client *gophercloud.ServiceClient, lbId string) error {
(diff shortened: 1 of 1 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -469,9 +469,10 @@ func ensureOpenStackLb(client *gophercloud.ServiceClient, name, vipAddress, vipS
 	}
 
 	opts := loadbalancers.ListOpts{
-		Name:        name,
-		VipAddress:  vipAddress,
-		VipSubnetID: vipSubnetId,
+		Name:               name,
+		VipAddress:         vipAddress,
+		VipSubnetID:        vipSubnetId,
+		ProvisioningStatus: "ACTIVE",
```
dulek: This will trigger IP conflict when a new LB will get created.

PR #321, opened 2019-09-20
Bug 1751246: openshift-sdn: add better ovs liveness/readiness probe

squeed:
This changes the liveness/readiness probe for OVS to (hopefully) a more rigorous end-to-end check. The challenge is to find one that doesn't rely on br0 being present, which won't happen until openshift-sdn is up and running.

So, we do ovs-ctl status, which checks that the pidfiles have pids that are up and running. It also adds an ovs-appctl command to verify that the vswitchd is running, and a ovs-vsctl check to see that the DB is running.

/cc @orgcandman

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -99,15 +99,34 @@ spec:
         livenessProbe:
           exec:
             command:
-            - /usr/share/openvswitch/scripts/ovs-ctl
-            - status
+            # Validate that ovs-vswitchd and ovsdb are running. These three commands check
+            # 1. that the pids in the pid file are up
+            # 2. that ovs-vswitchd is responding to queries
+            # 3. that ovsdb is responding to queries
+            # 4. if br0 is configured, that its management process is responding
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
+              /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null &&
+              /usr/bin/ovs-vsctl -t 5 show > /dev/null &&
+              if /usr/bin/ovs-vsctl -t 5 br-exists br0; then /usr/bin/ovs-ofctl -t 5 -O OpenFlow13 probe br0; else true; fi
           initialDelaySeconds: 15
           periodSeconds: 5
         readinessProbe:
           exec:
             command:
-            - /usr/share/openvswitch/scripts/ovs-ctl
-            - status
+            # The same validation as above, but without checking for br0
+            # This is because it's created by openshift-sdn, so looking for it
+            # won't be very meaningful.
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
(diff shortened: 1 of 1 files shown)


squeed:
This can wait until v4.3 cuts.
/hold

danwinship:
> This changes the liveness/readiness probe for OVS to (hopefully) a more rigorous end-to-end check.

do you have any examples of situations where the current probes are failing us?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -99,15 +99,26 @@ spec:
...
             command:
-            - /usr/share/openvswitch/scripts/ovs-ctl
-            - status
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
+              /usr/bin/ovs-appctl ofproto/list > /dev/null &&
+              /usr/bin/ovs-vsctl show > /dev/null &&
```
danwinship: maybe add comments explaining what each of these tests is testing?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -99,15 +99,26 @@ spec:
...
+              if ovs-vsctl br-exists br0; then ovs-ofctl -O OpenFlow13 probe br0; else true; fi
```
danwinship: ovs-ofctl(1) says: "`With the -t or --timeout option, [probe] can test whether an OpenFlow switch or controller is up and running.`" which kind of suggests we should probably use that? (Maybe in the `ovs-vsctl` and `ovs-appctl` commands too?)

orgcandman: Yes - best to use the '-t ' option because if some error happens during the command execution and ovs-vswitchd or the controller doesn't respond, the command will hang.

squeed:
> do you have any examples of situations where the current probes are failing us?

Nothing conclusive, but the linked bug (1751246) hints at this being a possibility. So it would be good to get some more confidence.

bindata/network/openshift-sdn/sdn-ovs.yaml
squeed: done

bindata/network/openshift-sdn/sdn-ovs.yaml
squeed: very good suggestion. added timeouts.

orgcandman:
the appctl command should cover it for ovs-vswitchd.  You can make an appctl command to ovsdb also (test something like: ovs-appctl -t ovsdb-server list-commands).  But if ovs-vsctl returns then the db is up.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -99,15 +99,34 @@ spec:
         livenessProbe:
           exec:
             command:
-            - /usr/share/openvswitch/scripts/ovs-ctl
-            - status
+            # Validate that ovs-vswitchd and ovsdb are running. These three commands check
+            # 1. that the pids in the pid file are up
+            # 2. that ovs-vswitchd is responding to queries
+            # 3. that ovsdb is responding to queries
+            # 4. if br0 is configured, that it's management process is responding
```
danwinship: "its"

squeed: argh! I *never* get this wrong!

PR #322, opened 2019-09-23
Correct the way nodes are computed for alert ClusterIPTablesStale

JacobTanenbaum:
Change `kube_pod_info_node_count` to `sum(kube_pod_info{namespace="openshift-sdn",  pod=~"ovs.*"})` 

this more accuratly computes the alert by returning the number of nodes that have an ovs pod running

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1754592

### DIFF
--- a/bindata/network/openshift-sdn/alert-rules.yaml
+++ b/bindata/network/openshift-sdn/alert-rules.yaml
@@ -67,7 +67,10 @@ spec:
       annotations:
         message: The average time between iptables resyncs is too high. NOTE - There is some scrape delay and other offsets, 90s isn't exact but it is still too high.
       expr: |
-        time() - (sum(kubeproxy_sync_proxy_rules_last_timestamp_seconds) / :kube_pod_info_node_count:) > 90
+        quantile(0.95,
+            timestamp(kubeproxy_sync_proxy_rules_last_timestamp_seconds)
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/alert-rules.yaml
```diff
@@ -67,7 +67,7 @@ spec:
       annotations:
         message: The average time between iptables resyncs is too high. NOTE - There is some scrape delay and other offsets, 90s isn't exact but it is still too high.
       expr: |
-        time() - (sum(kubeproxy_sync_proxy_rules_last_timestamp_seconds) / :kube_pod_info_node_count:) > 90
+        (time() - (sum(kubeproxy_sync_proxy_rules_last_timestamp_seconds)) / sum(kube_pod_info{namespace="openshift-sdn",  pod=~"ovs.*"})) > 90
```
danwinship: The new parenthesization is weird; before it was `TIME - (SUM / COUNT) > 90`, now it's `(TIME - (SUM) / COUNT) > 90` which I guess is still the same thing because of order of operations but it's weirder.

JacobTanenbaum: fixed

PR #323, opened 2019-09-23
Bug 1754606: [release-4.2] Correct the way nodes are computed for alert ClusterIPTablesStale

JacobTanenbaum:
Change `kube_pod_info_node_count` to `sum(kube_pod_info{namespace="openshift-sdn", pod=~"ovs.*"})`

this more accuratly computes the alert by returning the number of nodes that have an ovs pod running.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1754592


smarterclayton:
Approved for 4.2 because this is a top noise maker in the fleet

PR #324, opened 2019-09-23
Bug 1754682: The DHCP daemon should mount the directory for the socketfile

dougbtv:
Was incorrectly using a file outside of the host bind mount.


dougbtv:
/retest

looks like a bunch of CSI failures and unrelated to this change

PR #326, opened 2019-09-26
Adds aws service endpoints to default noProxy

danehans:
Previously, calls to AWS API endpoints were being proxied, when they shouldn't be. This PR adds the AWS service endpoint subdomain to the default `noProxy` list for AWS platform types.


danehans:
/hold

See https://github.com/openshift/installer/pull/2417#issuecomment-535588509 for details.

danehans:
Closing until we finalize an approach on managing a proxy whitelist.

PR #328, opened 2019-09-27
Fixes Bug 1757973: Removes external api server from default noProxy

danehans:
External cluster resources such as routes should not be automatically no-proxied. We should provide cluster admins the choice on whether or not to proxy these connections. This PR removes the external api-server hostname from the default noProxy list.

/assign @bparees @deads2k @knobunc

### DIFF
--- a/pkg/util/proxyconfig/no_proxy.go
+++ b/pkg/util/proxyconfig/no_proxy.go
@@ -46,16 +46,6 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 		ic.Networking.MachineCIDR,
 	)
 
-	if len(infra.Status.APIServerURL) > 0 {
-		apiServerURL, err := url.Parse(infra.Status.APIServerURL)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


bparees:
definitely would like @deads2k to confirm, but i believe this is correct per the slack discussion.

squeed:
This makes sense. Normally, services of type LoadBalancer still have ClusterIPs, but since this LB is managed by the installer (Terraform), we don't get that kind of behavior.

danehans:
Associated installer PR: https://github.com/openshift/installer/pull/2425

danehans:
/hold

until https://github.com/openshift/installer/pull/2425#issuecomment-536135085 is confirmed whether a bug should be filed and verification from @deads2k

danehans:
@squeed based on https://github.com/openshift/installer/pull/2425#issuecomment-536730235, I created [Bugzilla bug 1757973](https://bugzilla.redhat.com/show_bug.cgi?id=1757973) but the bot is unhappy with something in the bug definition.

bparees:
/bugzilla refresh

@danehans i made the same update here.

PR #329, opened 2019-09-30
ovn-kubernetes: run binaries directly

squeed:
While we're in the process of adding HA and TLS, let's just execute the binaries directly.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -1,12 +1,7 @@
-# ovnkube-master
-# daemonset version 3
-# starts master daemons, each in a separate container
-# it is run on the master node(s)
 kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: ovnkube-master
-  # namespace set up by install
   namespace: openshift-ovn-kubernetes
   annotations:
     kubernetes.io/description: |
@@ -19,6 +14,11 @@ spec:
       name: ovnkube-master
   strategy:
     type: RollingUpdate
+    rollingUpdate:
+      # by default, Deployments spin up the new pod before terminating the old one
+      # but we don't want that - because ovsdb holds the lock.
+      maxSurge: 0
+      maxUnavailable: 1
   template:
     metadata:
       labels:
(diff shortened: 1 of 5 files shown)


squeed:
Filed https://bugzilla.redhat.com/show_bug.cgi?id=1757094 to make it so we can run `ovn-controller` via `ovn-ctl`.

squeed:
Filed https://bugzilla.redhat.com/show_bug.cgi?id=1757097 to make it so we can use ovs-ctl for northd.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,25 +26,52 @@ spec:
...
+            sleep 5
+            (( retries += 1))
+            if [[ "${retries}" -gt 10 ]]; then
+              echo "timed out waiting for local databases"
+              exit 1
+            fi
+          done
+          ovn-nbctl set-connection ptcp:${OVN_NB_PORT} -- set connection . inactivity_probe=0
+          ovn-sbctl set-connection ptcp:${OVN_SB_PORT} -- set connection . inactivity_probe=0
+          exec ovn-northd --no-chdir -vconsole:info -vfile:off --pidfile=/var/run/openvswitch/ovn-northd.pid
```
dcbw: @squeed would be useful to have env vars for log levels for all these components so we can change that more easily when we need to.

squeed: agreed - I'll add the sysconfig overrides for now, and we can think about a configmap that allows per-host overrides.

squeed:
@dcbw updated. Added per-node environment variable overriding, which is a "cute hack" but works very well.

This PR is blocking some other improvements (notably TLS and the upstream rebase), so I'd like to get this merged soon.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,203 @@ spec:
...
-            cpu: 100m
-            memory: 300Mi
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -e
+          if [[ -f "/env/${K8S_NODE}" ]]; then
+            set -o allexport
+            source "/env/${K8S_NODE}"
```
danwinship: It seems wrong that if you want to override the master environment, you have to guess what node it's going to run on and then add node-specific overrides for that. ovnkube-node can look in `/env/${K8S_NODE}` but ovnkube-master should get its overrides from `/env/master`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,203 @@ spec:
...
+            set +o allexport
+          fi
+          retries=0
+          while true; do
+            if [[ ! -f /var/run/openvswitch/ovn-northd.pid ]]; then
```
danwinship: I don't think this is reliable; there may be a stale ovn-northd.pid file

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -28,77 +22,166 @@ spec:
...
+          chown -R openvswitch:openvswitch /etc/openvswitch
+          function quit {
+              /usr/share/openvswitch/scripts/ovs-ctl stop
+              exit 0
+          }
+          trap quit SIGTERM
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random
 
-        command: ["/root/ovnkube.sh", "ovs-server"]
+          if [[ `nproc` -gt 12 ]]; then
```
danwinship: This is unnecessary with openvswitch greater than mumble mumble and we should eventually remove it

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -28,77 +22,166 @@ spec:
...
+              ovs-vsctl --no-wait set Open_vSwitch . other_config:n-revalidator-threads=4
+              ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
+          fi
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random
+          ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
 
+          tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
+          tail -F --pid=$(cat /var/run/openvswitch/ovsdb-server.pid) /var/log/openvswitch/ovsdb-server.log &
```
danwinship: it would be simpler to just run separate vswitchd and dbserver containers wouldn't it?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -28,77 +22,166 @@ spec:
...
+        - -c
+        - |
+          set -e
+          if [[ -f "/env/${K8S_NODE}" ]]; then
+            set -o allexport
+            source "/env/${K8S_NODE}"
+            set +o allexport
+          fi
+          cp -f /usr/libexec/cni/ovn-k8s-cni-overlay /cni-bin-dir/
+          db_ip=$(kubectl get ep ovnkube-db -o jsonpath='{.subsets[0].addresses[0].ip}')
```
danwinship: You're assuming this already exists, which is not at all guaranteed

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -9,3 +9,23 @@ data:
   net_cidr:   {{.OVN_cidr}}
   svc_cidr:   {{.OVN_service_cidr}}
   k8s_apiserver: "{{.K8S_APISERVER}}"
+
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: ovn-env-overrides
```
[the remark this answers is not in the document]
squeed: Or maybe just `env-overrides`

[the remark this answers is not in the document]
squeed: Yup :-/

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,203 @@ spec:
...
-      # Requires fairly broad permissions - ability to read all services and network functions as well
-      # as all pods.
       serviceAccountName: ovn-kubernetes-controller
       hostNetwork: true
-      hostPID: true
       priorityClassName: "system-cluster-critical"
+      # volumes in all containers:
+      # (container) -> (host)
+      # /etc/openvswitch -> /var/lib/ovn - ovsdb schema (selinux blocks /etc on host)
+      # /var/lib/openvswitch -> /var/lib/ovn - ovsdb data
```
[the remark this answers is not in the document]
squeed: As for `/etc/openvswitch`, it's just used for system-id.conf.. for now. You're right, that's sketchy and I'll change it.

As for the other renaming (/run/openvswitch -> /run/ovn), I did this because I wanted to be 100% certain there was no interaction between the master and node processes.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: Yeah, that makes sense. I'll update it.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,203 @@ spec:
...
+            set +o allexport
+          fi
+          retries=0
+          nb_ok=0
+          sb_ok=0
+          while true; do
+            if ovn-nbctl show; then
+              nb_ok=1
+            else
+              echo "nbdb not yet up"
```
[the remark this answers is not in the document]
squeed: I'm not convinced we need to wait. Let me see if I can take it out.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,203 @@ spec:
...
+          name: host-var-lib-ovn
+        - mountPath: /run/openvswitch/
+          name: host-run-ovn
+        - mountPath: /env
+          name: env-overrides
         resources:
           requests:
             cpu: 100m
             memory: 300Mi
+      - name: nbdb
```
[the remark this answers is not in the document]
squeed: Pepperidge Farm Remembers.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: There's no easy way to tell if ovn-northd is running. We no longer run the host pid namespace, so we can't watch the pid.

I'm also not sure we need to wait for northd.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: as in, in this PR?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: I tried to do this for openshift-sdn and it didn't work. I don't remember why. ovsdb on the node side is very, very boring though.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: if not, it will fail, and the process will crashloop until it does.

bindata/network/ovn-kubernetes/004-config.yaml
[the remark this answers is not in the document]
squeed: also, it's in the ovn-kube namespace. I'm going to call it `env-overrides`

bindata/network/ovn-kubernetes/ovnkube-node.yaml
danwinship: if we already depend on ovs > mumble then yes, otherwise we should leave a comment with the correct value of "mumble" in it

which, ftr, is "`openvswitch2.11-2.11.0-11.el7fdn`"

bindata/network/ovn-kubernetes/ovnkube-master.yaml
danwinship: `ovnkube.sh` does

    ovs-appctl -t /var/run/openvswitch/ovn-northd.${northd_pid}.ctl version &> /dev/null

bindata/network/ovn-kubernetes/ovnkube-node.yaml
danwinship: I feel like crashlooping should not be an expected part of startup.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: I chose _master because the underscore can't appear in a domain name.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: True enough, though this will all go away when we have HA, and endpoints are just watched automatically.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: I don't see any northd dependency in ovnkube. More precisely, the only code that talks to sbdb is the node process.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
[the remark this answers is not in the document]
squeed: Doesn't seem to be necessary. Removed and everything came up fine.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: Removing the ready check and everything comes up fine, even with a decent amount of chaos-monkeying

squeed:
@danwinship updated, thanks for the review. Take a look at the directory updates and tell me what you think.

squeed:
/retest
(aws flake)

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: Trying to figure out which version we have. None of the `--version` commands seem to give me a commit hash.

squeed: Aha, looks like we have openvswitch2.11.x86_64 0:2.11.0-18.el7fdp. That particular bit of cargo-culting can die.

squeed: Wait, I'm not familiar with the suffixes. Is there any chance that `el7fdp` is different from `el7fdn`?

squeed:
Some failed tests:

*ignorable:* 
- [sig-network] Networking Granular Checks: Services should function for client IP based session affinity: http [Suite:openshift/conformance/parallel] [Suite:k8s] -- filed https://github.com/openshift/origin/pull/23896
- [sig-network] Networking Granular Checks: Services should function for client IP based session affinity: udp [Suite:openshift/conformance/parallel] [Suite:k8s] -- same
- [Feature:Platform][Smoke] Managed cluster should ensure control plane operators do not make themselves unevictable [Suite:openshift/conformance/parallel] --same

*broken*:
- [sig-network] Services should allow pods to hairpin back to themselves through services [Suite:openshift/conformance/parallel] [Suite:k8s] -- *broken*, filed SDN-587
- [sig-network] Networking Granular Checks: Services should function for endpoint-Service: http [Suite:openshift/conformance/parallel] [Suite:k8s]
- [sig-network] Networking Granular Checks: Services should function for endpoint-Service: udp [Suite:openshift/conformance/parallel] [Suite:k8s]

I do not think this PR caused any of these issues. Investigating further.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: Added a retry loop.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
danwinship: > Wait, I'm not familiar with the suffixes. Is there any chance that el7fdp is different from el7fdn?

um... I think one is like testing and the other is release or somethign like that... @knobunc should know

pecameron: @squeed @danwinship ovn/ovs is from fast datapath.

dcbw: @squeed 
"fdn" = Fast Datapath Next (eg, stuff still under testing and not yet released to customers)
"fdp" = Fast Data Path (eg stuff that has been officially QA-ed and released)

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -9,3 +9,27 @@ data:
...
+# to set a single node in to debug mode, create an env-file with the node name
+# in this configmap. The next time the ovn-kubernetes processes start on that node,
+# they will pick up the changes
+# e.g.
+# data:
+#   ip-10-0-135-96.us-east-2.compute.internal: |
+#     OVN_KUBE_LOG_LEVEL=5
+#     OVN_LOG_LEVEL=dbg
+#     OVS_LOG_LEVEL=dbg
+# To adjust master log levels, the use _master
```
danwinship: drop "the"

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,150 @@ spec:
...
-      # as all pods.
       serviceAccountName: ovn-kubernetes-controller
       hostNetwork: true
-      hostPID: true
       priorityClassName: "system-cluster-critical"
+      # volumes in all containers:
+      # (container) -> (host)
+      # /etc/openvswitch -> /var/lib/ovn/etc - ovsdb data
+      # /var/lib/openvswitch -> /var/lib/ovn/data - ovsdb pki state
+      # /run/openvswitch -> tmpfs - sockets & pids
```
danwinship: You've confirmed it's just sockets and pids? `/run` would normally be persistent across an OVN restart, so this is a semantic change...

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -9,3 +9,27 @@ data:
...
+
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: env-overrides
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    # This is for debugging purposes and will be updated external to the operator
+    networkoperator.openshift.io/create-only: ""
```
danwinship: oh, so hey, it turns out that it's accidentally undocumented, but you can actually do

    volumes:
    - name: env-overrides
      configMap:
        name: env-overrides
        optional: true

and then it won't complain if the configmap doesn't exist. So we don't actually need to create the default empty `env-overrides` configmap...

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -28,215 +28,150 @@ spec:
...
+            source /env/_master
+            set +o allexport
+          fi
+          exec /usr/share/openvswitch/scripts/ovn-ctl run_nb_ovsdb \
+            --no-monitor \
+            --ovn-nb-log="-vconsole:${OVN_LOG_LEVEL} -vfile:off"
+        lifecycle:
+          postStart:
+            exec:
+              command: ["ovn-nbctl", "set-connection", "ptcp:{{.OVN_NB_PORT}}", "--", "set", "connection", ".", "inactivity_probe=0"]
```
danwinship: Is it guaranteed that nbdb will be ready by the earliest time kubelet could decide to run this command?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
danwinship: so it sounds like we *can* get rid of this

bindata/network/ovn-kubernetes/004-config.yaml
squeed: Cool, I'll move this to the README.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: Nope, and even worse: it's not guaranteed that postStart hooks will be run after the container's entrypoint is executed! I'll add a loop.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: done.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: Yup, but thanks for checking.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: gotten rid of.

squeed:
All test failures are known issues (and will be ignored soon).

squeed:
Update on those two bugs: the OVN team says we don't need to use ovn-ctl for northd and ovn-controller. Just for the DBs.

PR #333, opened 2019-10-04
OVN: use config file via ConfigMap rather than environment variables

pecameron:
Configure ovnkube using a config file passed by the ovnkube-config
configMap and mountedinto the container.

SDN-456
https://jira.coreos.com/browse/SDN-456

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -1,11 +1,24 @@
 ---
-# The network cidr and service cidr are set in the ovn-config configmap
+# The ovnconfig config file. Used by both node and master processes.
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: ovn-config
+  name: ovnkube-config
   namespace: openshift-ovn-kubernetes
 data:
-  net_cidr:   {{.OVN_cidr}}
-  svc_cidr:   {{.OVN_service_cidr}}
-  k8s_apiserver: "{{.K8S_APISERVER}}"
+  ovnkube.conf:   |-
+    [default]
+    mtu="{{.MTU}}"
+    cluster-subnets="{{.OVN_cidr}}"
+
+    [kubernetes]
+    service-cidr="{{.OVN_service_cidr}}"
+    ovn-config-namespace="openshift-ovn-kubernetes"
+    apiserver="{{.K8S_APISERVER}}"
+
+    [logging]
+    logfile="/dev/stdout"
+
+    [gateway]
+    mode=local
+    nodeport=true
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -40,7 +40,7 @@ spec:
       containers:
       # ovn-northd: convert network objects in nbdb to flows in sbdb
       - name: northd
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
         command:
         - /bin/bash
         - -c
@@ -66,6 +66,12 @@ spec:
             nb_addr_list="${nb_addr_list}ssl:${host}:{{.OVN_NB_PORT}}"
             sb_addr_list="${sb_addr_list}ssl:${host}:{{.OVN_SB_PORT}}"
           done
+          echo /ovn-cert/tls.key
+          cat /ovn-cert/tls.key
+          echo /ovn-cert/tls.crt
+          cat /ovn-cert/tls.crt
+          echo /ovn-ca/ca-bundle.crt
+          cat /ovn-ca/ca-bundle.crt
           
           exec ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off \
@@ -98,7 +104,7 @@ spec:
 
       # nbdb: the northbound, or logical network object DB. In raft mode 
       - name: nbdb
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
         command:
         - /bin/bash
         - -c
@@ -113,6 +119,12 @@ spec:
           OVN_NODES_ARRAY=({{.OVN_NODES}})
           MASTER_NODE=$(getent ahostsv4 "${OVN_NODES_ARRAY[0]}" | grep RAW | awk '{print $1}')
           LOCALHOST=$(getent ahostsv4 "${K8S_NODE}" | grep RAW | awk '{print $1}')
+          echo /ovn-cert/tls.key
+          cat /ovn-cert/tls.key
+          echo /ovn-cert/tls.crt
+          cat /ovn-cert/tls.crt
+          echo /ovn-ca/ca-bundle.crt
+          cat /ovn-ca/ca-bundle.crt
 
           if [[ "$LOCALHOST" == "$MASTER_NODE" ]]; then
             exec /usr/share/openvswitch/scripts/ovn-ctl \
@@ -191,7 +203,7 @@ spec:
       
       # sbdb: The southbound, or flow DB. In raft mode 
       - name: sbdb
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
         command:
         - /bin/bash
         - -c
@@ -206,6 +218,12 @@ spec:
           OVN_NODES_ARRAY=({{.OVN_NODES}})
           MASTER_NODE=$(getent ahostsv4 "${OVN_NODES_ARRAY[0]}" | grep RAW | awk '{print $1}')
           LOCALHOST=$(getent ahostsv4 "${K8S_NODE}" | grep RAW | awk '{print $1}')
+          echo /ovn-cert/tls.key
+          cat /ovn-cert/tls.key
+          echo /ovn-cert/tls.crt
+          cat /ovn-cert/tls.crt
+          echo /ovn-ca/ca-bundle.crt
+          cat /ovn-ca/ca-bundle.crt
 
           if [[ "$LOCALHOST" == "$MASTER_NODE" ]]; then
             exec /usr/share/openvswitch/scripts/ovn-ctl \
@@ -281,7 +299,7 @@ spec:
 
       # ovnkube master: convert kubernetes objects in to nbdb logical network components
       - name: ovnkube-master
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
         command:
         - /bin/bash
         - -c
@@ -313,6 +331,12 @@ spec:
             nb_addr_list="${nb_addr_list}ssl:${host}:{{.OVN_NB_PORT}}"
             sb_addr_list="${sb_addr_list}ssl://${host}:{{.OVN_SB_PORT}}"
           done
+          echo /ovn-cert/tls.key
+          cat /ovn-cert/tls.key
+          echo /ovn-cert/tls.crt
+          cat /ovn-cert/tls.crt
+          echo /ovn-ca/ca-bundle.crt
+          cat /ovn-ca/ca-bundle.crt
           
           # start nbctl daemon for caching
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/run/openvswitch/ovnk-nbctl.pid \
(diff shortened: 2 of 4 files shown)


bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -1,11 +1,26 @@
...
+  name: ovnkube-config
   namespace: openshift-ovn-kubernetes
 data:
-  net_cidr:   {{.OVN_cidr}}
-  svc_cidr:   {{.OVN_service_cidr}}
-  k8s_apiserver: "{{.K8S_APISERVER}}"
+  ovn_k8s.conf:   |-
+    [default]
+    mtu="{{.MTU}}"
+    cluster-subnets="{{.K8S_cluster_subnets}}"
```
danwinship: The changes more-or-less look good, except that we're already pretty terrible about having a consistent naming style for template variables, and the new ones aren't helping. We should either consistently use CamelCase (as openshift-sdn mostly does) or consistently use lowercase_with_underscores. And let's get rid of the "OVN"/"K8S"/"OVNKUBE" prefixes where they aren't helping.

pecameron: went with mostly camel case

danwinship:
You should squash #335 into this

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -1,11 +1,26 @@
 ---
-# The network cidr and service cidr are set in the ovn-config configmap
+# The config file used by ovnkube is passed as a configmap.
```
squeed: "The ovnkube config file. Used by both node and master processes."

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -1,11 +1,26 @@
...
+    mtu="{{.MTU}}"
+    cluster-subnets="{{.ClusterSubnets}}"
+
+    [kubernetes]
+    service-cidr="{{.ServiceCidr}}"
+    ovn-config-namespace="openshift-ovn-kubernetes"
+    apiserver="{{.K8S_APISERVER}}"
+
+    [logging]
+    loglevel="{{.OvnkubeLogLevel}}"
```
squeed: This should be overridable via the environment config map trick. So we need to pass loglevel via environment args.

squeed:
This is a better way to do things, for sure. However, we need to add something to ovnkube that causes it to exit on configuration changes.

bindata/network/ovn-kubernetes/004-config.yaml
pecameron: changed

danwinship:
> You should squash #335 into this

oops, too late. needs-rebase now

bindata/network/ovn-kubernetes/004-config.yaml
pecameron: OK.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -291,6 +278,9 @@ spec:
       - name: host-var-lib-cni-networks-ovn-kubernetes
         hostPath:
           path: /var/lib/cni/networks/ovn-k8s-cni-overlay
+      - name: ovnkube-config-mount
```
squeed: just call this `ovnkube-config`

pecameron: OK

bindata/network/ovn-kubernetes/004-config.yaml
squeed: Should this still be here?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -178,20 +178,18 @@ spec:
...
             --init-master "${K8S_NODE}" \
-            --cluster-subnets "${OVN_NET_CIDR}" \
-            --k8s-service-cidr "${OVN_SVC_CIDR}" \
-            --k8s-apiserver "{{.K8S_APISERVER}}" \
-            --ovn-config-namespace openshift-ovn-kubernetes \
-            --nodeport \
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
-            --logfile /dev/stdout
+            --config-file=/run/ovnkube-config/ovnkube.conf
+
```
squeed: remove this newline

bindata/network/ovn-kubernetes/004-config.yaml
squeed: Yeah, I think you should delete this, since we pass `--loglevel` to both processes.

squeed:
Two small changes, otherwise looks fine.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: Done

pecameron:
@squeed PTAL

squeed:
Those test failures are not the expected set of failures for ovn-kube. Something is wrong. You'll have to dig in a bit further.

pecameron:
@squeed didn't spot anything (still not sure what to look for or where) so I ran the test again.

squeed:
Huh... something is weird with the ingress operator:
```
                    {
                        "message": "pods \"ingress-operator-695f4b555-\" is forbidden: no SecurityContextConstraints found in cluster",
                        "reason": "FailedCreate",
                        "status": "True",
                        "type": "ReplicaFailure"
                    },
```

I don't think this is an ovn problem.

/test e2e-aws-ovn-kubernetes

squeed:
Digging a bit further, the openshift-apiserver is having network connectivity issues:

https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/333/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-kubernetes/45/artifacts/e2e-aws-ovn-kubernetes/pods/openshift-apiserver_apiserver-5s6kh_openshift-apiserver.log

I wonder if this will be fixed by HA.

pecameron:
@squeed, not sure HA will help with not being able to access the api server.

pecameron:
@squeed changes to exit ovnkube on configmap change is in https://github.com/ovn-org/ovn-kubernetes/pull/860

squeed:
/test e2e-aws-ovn-kubernetes
--

pecameron:
Failed:
ft-controller-manager-operator/openshift-controller-manager-operator\" (448 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-image-registry/image-registry\" (409 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-kube-apiserver-operator/kube-apiserver-operator\" (424 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-kube-controller-manager-operator/kube-controller-manager-operator\" (429 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-kube-scheduler-operator/kube-scheduler-operator\" (433 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-machine-api/cluster-autoscaler-operator\" (154 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-machine-api/machine-api-operator\" (435 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-machine-config-operator/machine-config-daemon\" (437 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-operator-lifecycle-manager/olm-operator\" (438 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-service-catalog-apiserver-operator/openshift-service-catalog-apiserver-operator\" (413 of 449): the server does not recognize this resource, check extension API servers\n* Could not update servicemonitor \"openshift-service-catalog-controller-manager-operator/openshift-service-catalog-controller-manager-operator\" (416 of 449): the server does not recognize this resource, check extension API servers"

/test e2e-aws-ovn-kubernetes

pecameron:
@dcbw I am at a total loss what is going wrong with this PR in the ovn e2e test. The code changes are simple and I don't think we get far enough to run cno. Could you PTAL? Thanks, phil

pecameron:
@squeed added ssl config to the ovnkube-config configmap PTAL

pecameron:
@squeed I did the rebase and pushed. PTAL

squeed:
Test failure is because ovn-kube isn't talking to the apiserver correctly. Presumably the service-cidr isn't working either.

```
E1203 16:06:52.457450       1 reflector.go:126] k8s.io/client-go/informers/factory.go:133: Failed to list *v1.Pod: Get http://localhost:8080/api/v1/pods?limit=500&resourceVersion=0: dial tcp [::1]:8080: connect: connection refused
```

pecameron:
@squeed how is localhost:8080 set up? Can we tell if apiserver is dead or we just can't reach it?
Looked in kubelet.log and found CrashLoopBackoff  and Error failed to sync messages.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -322,11 +322,6 @@ spec:
 
           exec /usr/bin/ovnkube \
             --init-master "${K8S_NODE}" \
-            --cluster-subnets "${OVN_NET_CIDR}" \
-            --k8s-service-cidr "${OVN_SVC_CIDR}" \
-            --k8s-apiserver "{{.K8S_APISERVER}}" \
-            --ovn-config-namespace openshift-ovn-kubernetes \
-            --nodeport \
```
dcbw: @pecameron don't we need a `--config-file=/run/ovnkube-config/ovnkube.conf` here too?

dcbw: You can also remove the `--logfile /dev/stdout` argument from this ovnkube command since it's now in the config file.

pecameron: @dcbw I added the --config. Looking at the code it appears to be the default, if it exists. I noticed and pushed similar changes before reading your comment above. --logfile will get removed in the next push.

dcbw: @pecameron there is a default config file, but that default path sure isn't /run/ovnkube-config/

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -336,6 +331,7 @@ spec:
             --sb-client-privkey /ovn-cert/tls.key \
             --sb-client-cert /ovn-cert/tls.crt \
             --sb-client-cacert /ovn-ca/ca-bundle.crt
+            --config-file=/run/ovnkube-config/ovnkube.conf \
```
dcbw: @pecameron stray `\` at the end of the last line here, that needs to move up to the end of the previous line?

pecameron: Yup

squeed:
Well, OVN came up, so this PR is probably okay. Weird that CI failed though

squeed:
It does seem like the MTU is problematic. Can you check to see if the MTU is correctly respected by all of the interfaces? Perhaps the gateway code hard-codes it when it shouldn't.

squeed:
Yes, scanning the gateway code shows that it doesn't seem to consume mtu @dcbw

pecameron:
@squeed, @dcbw, whatever is happening here is not happening in other PRs. The mtu is generally handled properly. This might be a corner case. In the spirit of race conditions as we have seen in not getting the correct number of masters, is it possible that the sll certs are not set by the time they are used? The problem discussed in the f2f was in host networking ssl between ovs/ovn components.

squeed:
My evidence for this is that the control plane seems to come up fine, but the openshift-apiserver (a non-hostnetwork component) times out when trying to write a response back to the kube-apsierver (which *is* hostnetwork).

Looking at the gateway code, I see that the interface created does not have the MTU passed.

Get what I'm saying? Would you mind checking to see if I'm right? If not, can you please spin up a cluster for me and I can check?

pecameron:
@squeed Yes I get what you are saying. Why does that happen in PR 333 and not happen in other PRs? What you are asserting should have wide spread impact.

The analysis from the f2f showed an ssl problem accessing sdb:9642 using host networking. The connection was established and ssl failed.

I am spinning up a cluster, I'll let you know when it is ready.

pecameron:
@squeed cluster didn't come up, bad token. Trying again.

PR #335, opened 2019-10-04
ovn-kubernetes: fix endpoint namespace

dcbw:
Upstream defaults to just 'ovn-kubernetes' but we need to use
openshift-ovn-kubernetes.

Fixes: 0ac2cdf16adffca72e385337df95d54d9e97e249

@squeed @JacobTanenbaum @danwinship

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -181,6 +181,7 @@ spec:
             --cluster-subnets "${OVN_NET_CIDR}" \
             --k8s-service-cidr "${OVN_SVC_CIDR}" \
             --k8s-apiserver "{{.K8S_APISERVER}}" \
+            --ovn-config-namespace openshift-ovn-kubernetes \
             --nodeport \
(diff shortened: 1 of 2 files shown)


JacobTanenbaum:
LGTM

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -181,6 +181,7 @@ spec:
             --cluster-subnets "${OVN_NET_CIDR}" \
             --k8s-service-cidr "${OVN_SVC_CIDR}" \
             --k8s-apiserver "{{.K8S_APISERVER}}" \
+            --ovn-config-namespace openshift-ovn-kubernetes \
```
danwinship: `--ovn-config-namespace` is only relevant if we're using a configmap, which we aren't yet. It seems like this part belongs in #333 and all you need here is the kubectl fix

dcbw: @danwinship it will be used for the master creating endpoints in https://github.com/ovn-org/ovn-kubernetes/pull/763 at some point here too

PR #336, opened 2019-10-07
Bump versions to 1.16

rcarrillocruz:


Gopkg.toml
```diff
@@ -10,38 +10,44 @@ required = [
...
+[[override]]
+  name = "gopkg.in/fsnotify.v1"
+  source = "https://github.com/fsnotify/fsnotify.git"
+
 [[override]]
   name = "k8s.io/code-generator"
-  # revision for tag "kubernetes-1.12.3"
-  revision = "3dcf91f64f638563e5106f21f50c31fa361c918d"
+  # revision for tag "kubernetes-1.16.0"
+  revision = "cd179ad6a2693011d6f2fa5cd64c6680ee99379f"
```
danwinship: @squeed why did we do it this way rather than just doing `version = "kubernetes-1.16.0"` ?

pkg/util/k8s/kubeproxy.go
```diff
@@ -68,7 +68,6 @@ func GenerateKubeProxyConfiguration(args map[string]operv1.ProxyArgumentList) (s
 
 	kpc.UDPIdleTimeout.Duration = ka.getDuration("udp-timeout")
 
-	kpc.Conntrack.Max = ka.getOptInt32("conntrack-max")
```
danwinship: I assume the unit tests are going to need to be updated to reflect this field going away. It might be good to add an explicit test that we get an error now if the user sets `conntrack-max` in the config.

rcarrillocruz: yah, i'm actually cleaning up text matrixes setting up max.
Should've really put WIP on the BZ, will amend

Gopkg.toml
squeed: Cargo cullted from a long time ago. As I recall, client-go didn't used to tag k8s versions.

We can almost certainly switch to more meaningful tags now.

Gopkg.toml
```diff
@@ -10,38 +10,44 @@ required = [
   "k8s.io/gengo/args",
 ]
 
+[[override]]
```
squeed: Want to add a comment as to why you added this? (Some weird dep thing, right?)

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -124,7 +124,7 @@ func (r *ReconcileConfigMapInjector) Reconcile(request reconcile.Request) (recon
 
 		configMapList := &corev1.ConfigMapList{}
 		selector := labels.Set(map[string]string{names.TRUSTED_CA_BUNDLE_CONFIGMAP_LABEL: "true"}).AsSelector()
-		err = r.client.List(context.TODO(), &client.ListOptions{LabelSelector: selector}, configMapList)
+		err = r.client.List(context.TODO(), configMapList, &client.ListOptions{LabelSelector: selector})
```
squeed: what!?!? did they seriously change the Client interface? Or was this a bug?

Gopkg.toml
rcarrillocruz: yep, unsure if it's a local issue with my dep version, was hitting this: https://github.com/golang/dep/issues/1799

Gopkg.toml
rcarrillocruz: i support the idea of switching to tags, way more human readable. Can amend later once I sort out CI issues

pkg/controller/configmap_ca_injector/controller.go
rcarrillocruz: unsure when, but yeah, interface has the target obj first, filter second. From what I can see it's been like that a couple releases

Gopkg.toml
```diff
@@ -10,38 +10,45 @@ required = [
...
+  # Had to add this since I was getting golang panics: https://github.com/golang/dep/issues/1799
+  name = "gopkg.in/fsnotify.v1"
+  source = "https://github.com/fsnotify/fsnotify.git"
+
 [[override]]
   name = "k8s.io/code-generator"
-  # revision for tag "kubernetes-1.12.3"
-  revision = "3dcf91f64f638563e5106f21f50c31fa361c918d"
+  # revision for tag "kubernetes-1.16.0"
```
squeed: I would say we can get rid of these comments :-)

squeed:
@rcarrillocruz I'll approve this to unblock others, but can you file a follow-up PR that addresses @danwinship's comment around `--conntrack-max`? We should set the network to Degraded if that option has been passed (or rewrite it to the new destination).

PR #337, opened 2019-10-07
SDN-394: Making the sdn controller a deployment

alexanderConstantinescu:
This PR makes the SDN controller a deployment. 

* We will not use a `nodeSelector: master` anymore as that is not required (and should not be) for running the SDN controller. 
* We will run two replicas, as to mimic an active-active recovery.
* We will however introduce `antiAffinity` with itself, as to make sure that the leader election is done properly (today it's done by using to the `hostName` as key, as seen [here](https://github.com/openshift/sdn/blob/master/pkg/openshift-network-controller/network_controller.go#L68). This is required as the `hostName` equals the node's name for the SDN components (`hostNetwork: true`).

/assign @squeed

### DIFF
--- a/pkg/network/openshift_sdn_test.go
+++ b/pkg/network/openshift_sdn_test.go
@@ -70,14 +70,14 @@ func TestRenderOpenShiftSDN(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ServiceAccount", "openshift-sdn", "sdn-controller")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "openshift-sdn")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-sdn", "sdn")))
-	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-sdn", "sdn-controller")))
+	g.Expect(objs).To(ContainElement(HaveKubernetesID("Deployment", "openshift-sdn", "sdn-controller")))
 
 	// No netnamespaces by default
 	g.Expect(objs).NotTo(ContainElement(HaveKubernetesID("NetNamespace", "", "openshift-ingress")))
 
-	// make sure all deployments are in the master
+	// make sure all deployments, except the sdn-controller, are on the master,
 	for _, obj := range objs {
-		if obj.GetKind() != "Deployment" {
+		if obj.GetKind() != "Deployment" || obj.GetName() == "sdn-controller" {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/network/openshift_sdn_test.go
```diff
@@ -70,14 +70,14 @@ func TestRenderOpenShiftSDN(t *testing.T) {
...
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "openshift-sdn")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-sdn", "sdn")))
-	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-sdn", "sdn-controller")))
+	g.Expect(objs).To(ContainElement(HaveKubernetesID("Deployment", "openshift-sdn", "sdn-controller")))
 
 	// No netnamespaces by default
 	g.Expect(objs).NotTo(ContainElement(HaveKubernetesID("NetNamespace", "", "openshift-ingress")))
 
-	// make sure all deployments are in the master
+	// make sure all deployments, except the sdn-controller, are on the master,
```
squeed: hah, this particular assertion isn't really meaningful - let's get rid of it.

danwinship:
>     * We will not use a `nodeSelector: master` anymore as that is not required (and should not be) for running the SDN controller.

It's true that it's not *required* but isn't it desired anyway?

danwinship:
um, also, SDN-297 is about OVN masters not SDN masters... typo? or expanding the scope of the card?

danwinship:
ah
/retitle SDN-394: Making the sdn controller a deployment

squeed:
> > * ... `nodeSelector: master` ...
> 
> It's true that it's not _required_ but isn't it desired anyway?

It was my idea to get rid of it. I can't think of any reason why we have to run on control-plane nodes. We don't need guaranteed resources to keep the cluster alive. We don't have any particularly scary ClusterRoles (yah, okay, patching ClusterNetwork could break network policy).

My thinking was basically
1. We don't need to run the controller on every control-plane node. We just want 2 of them, so one takes over quickly
2. By default, libvirt sets up a single master node. Therefore
3. We should tolerate, but not require, master nodes.

I'll ask around today to see if this is a controversial position.

danwinship:
what do other redundant controllers do on libvirt? fail?

squeed:
Surveying the state of OpenShift, it seems that the other control loops are either master node DaemonSets* or Deployments with replica 1.

So either we're behind or pioneers. Or possibly both.

* or are effectively DaemonSets but are actually static Pods.

danwinship:
Theoretically, we shouldn't really need to be running two sdn-controllers at the same time; if it's unavailable for a brief period, it should catch up when the new one starts. But I think we're not perfect about recovering from missed events in all cases... (eg, as a minor example, if we miss a Namespace deletion then we'll leave a stray NetNamespace around forever).

PR #341, opened 2019-10-08
ovn custom geneve port number

pecameron:
SDN-494 - ovn custom geneve port number.
https://jira.coreos.com/browse/SDN-494

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -46,7 +46,7 @@ _output/linux/amd64/cluster-network-renderer --config sample-config.yaml --out o
 ```
 
 ### Building images
-By default, podman is used to build images. 
+By default, podman is used to build images.
 
 ```
 ./hack/build-image.sh
--- a/README.md
+++ b/README.md
@@ -132,8 +132,9 @@ spec:
 ```
 
 ### Configuring OVNKubernetes
-OVNKubernetes supports the following configuration options, all of which are optional:
+OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed:
 * `MTU`: The MTU to use for the geneve overlay. The default is the MTU of the node that the cluster-network-operator is first run on, minus 100 bytes for geneve overhead. If the nodes in your cluster don't all have the same MTU then you may need to set this explicitly.
+* `genevePort`: The UDP port to use for the Geneve overlay. The default is 6081.
 
 These configuration flags are only in the Operator configuration object.
 
@@ -144,6 +145,7 @@ spec:
     type: OVNKubernetes
     ovnKubernetesConfig:
       mtu: 1400
+      genevePort: 6081
 ```
 
 Additionally, you can configure per-node verbosity for ovn-kubernetes. This is useful
@@ -366,7 +368,7 @@ spec:
 The operator is expected to run as a pod (via a Deployment) inside a kubernetes cluster. It will retrieve the configuration above and reconcile the desired configuration. A suitable manifest for running the operator is located in `manifests/`.
 
 ## Unsafe changes
-Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested. 
+Most network changes are unsafe to roll out to a production cluster. Therefore, the network operator will stop reconciling if it detects that an unsafe change has been requested.
 
 ### Safe changes to apply:
 It is safe to edit the following fields in the Operator configuration:
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -44,6 +44,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["KUBERNETES_SERVICE_PORT"] = os.Getenv("KUBERNETES_SERVICE_PORT")
 	data.Data["K8S_APISERVER"] = fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT"))
 	data.Data["MTU"] = c.MTU
+	data.Data["GenevePort"] = c.GenevePort
 	data.Data["CNIConfDir"] = pluginCNIConfDir(conf)
 	data.Data["CNIBinDir"] = CNIBinDir
 	data.Data["OVN_NB_PORT"] = OVN_NB_PORT
@@ -108,6 +109,9 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 		if oc.MTU != nil && (*oc.MTU < 576 || *oc.MTU > 65536) {
 			out = append(out, errors.Errorf("invalid MTU %d", *oc.MTU))
 		}
+		if oc.GenevePort != nil && (*oc.GenevePort < 1 || *oc.GenevePort > 65535) {
+			out = append(out, errors.Errorf("invalid GenevePort %d", *oc.GenevePort))
+		}
 	}
 
 	return out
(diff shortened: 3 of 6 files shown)


pecameron:
@dcbw @danwinship @squeed PTAL

HACKING.md
```diff
@@ -171,3 +171,27 @@ podman push ${REGISTRY}/origin-node:latest
 ```
 echo "NODE_IMAGE=${REGISTRY}/origin-node:latest" >> ${CLUSTER_DIR}/env.sh
 ```
+
+# OVN - Change Geneve Port
```
danwinship: We don't need to document this here; it will eventually be in the full OpenShift documentation

README.md
```diff
@@ -132,8 +132,9 @@ spec:
 ```
 
 ### Configuring OVNKubernetes
-OVNKubernetes supports the following configuration options, all of which are optional:
+OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed:
 * `MTU`: The MTU to use for the geneve overlay. The default is the MTU of the node that the cluster-network-operator is first run on, minus 100 bytes for geneve overhead. If the nodes in your cluster don't all have the same MTU then you may need to set this explicitly.
+* `genevePort`: The UDP port to use for the Geneve overlay. The default is 6081, on OpenShift the default is 9081.
```
danwinship: I thought we were changing the nbdb and sbdb ports but leaving the Geneve port unchanged, for each of tcpdump debugging.

pkg/network/ovn_kubernetes.go
```diff
@@ -110,16 +125,27 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 	}
 
 	sc := conf.DefaultNetwork.OVNKubernetesConfig
-	// MTU is currently the only field we pull from previous.
-	// If it's not supplied, we infer it from  the node on which we're running.
+	// MTU  and GenevePort are currently the only fields we pull from previous.
```
danwinship: Don't pull GenevePort from previous; we only do it with MTU because if MTU isn't set in previous then we autodetect it. For GenevePort, just set the default value if it's not set in `conf`; see `fillOpenShiftSDNDefaults`.

pkg/network/ovn_kubernetes.go
```diff
@@ -110,16 +125,27 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
...
+	// If MTU is not supplied, we infer it from  the node on which we're running.
 	// However, this can never change, so we always prefer previous.
 	if sc.MTU == nil {
 		var mtu uint32 = uint32(hostMTU) - 100 // 100 byte geneve header
-		if previous != nil && previous.DefaultNetwork.OVNKubernetesConfig != nil {
+		if previous != nil && previous.DefaultNetwork.OVNKubernetesConfig != nil &&
+			previous.DefaultNetwork.Type == operv1.NetworkTypeOVNKubernetes &&
+			previous.DefaultNetwork.OVNKubernetesConfig.MTU != nil {
```
danwinship: don't need this; we will have already validated that `DefaultNetwork.Type` didn't change

pkg/network/ovn_kubernetes_test.go
```diff
@@ -167,8 +178,12 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
 	// change the mtu
 	mtu := uint32(70000)
 	next.DefaultNetwork.OVNKubernetesConfig.MTU = &mtu
-	errs = isOVNKubernetesChangeSafe(prev, next)
-	g.Expect(errs).To(HaveLen(1))
 
-	g.Expect(errs[0]).To(MatchError("cannot change ovn-kubernetes configuration"))
+	// change the geneve port
+	geneve := uint32(70001)
```
danwinship: you should use a valid port value here

sample-ovn-config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: "cluster"
+spec:
+  serviceNetwork:
+    - "172.30.0.0/16"
+  clusterNetwork:
+    - cidr: "10.128.0.0/14"
+      hostPrefix: 23
+  defaultNetwork:
+    type: OVNKubernetes
+    RovnKubernetesConfig:
```
danwinship: typo. also, we still don't need this file

HACKING.md
pecameron: We can document it here as a convenience to developers and information for the doc team. Eventually can be a long time.

README.md
pecameron: The sdb and ndb ports changed in PR 313. This also shifts the geneve port to 9000-9999.

pkg/network/ovn_kubernetes_test.go
pecameron: The test is for an invalid port.

pkg/network/ovn_kubernetes.go
pecameron: We can't change the port. If its in prev we use it. Where is fillOpenShiftSDNDefaults.

pecameron: OK. Where is fillOpenShiftSDNDefaults?

pkg/network/ovn_kubernetes.go
pecameron: ok

sample-ovn-config.yaml
pecameron: Fixed typo. It was not obvious how to code this.

README.md
danwinship: My understanding is that we had explicitly decided that we didn't want to change the default geneve port.

pkg/network/ovn_kubernetes.go
danwinship: `pkg/network/openshift_sdn.go`

pkg/network/ovn_kubernetes_test.go
danwinship: The test for an invalid port is earlier; this line is in the test that changes to the config aren't allowed, but you used the same invalid port. It's better to test this with a valid port so that we're confirming that you can't change it even if you try to change it to a different valid port.

README.md
pecameron: @squeed What is your desire in changing the geneve port. See @danwinship comment above.

pkg/network/ovn_kubernetes_test.go
pecameron: Changed

pecameron:
@danwinship @squeed PTAL

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -56,13 +56,16 @@ spec:
           trap quit SIGTERM
           /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
+          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=${GENEVE_PORT} enable-protocol
```
dcbw: @pecameron good catch, though we also need to pass GENEVE_PORT into ovnkube --node in the ovnkube node container as well.

pecameron:
/hold
waiting for PR 333 to land and configmap to be updated.

README.md
pecameron: Will use 6081 and allow customer to change port.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
pecameron: @dcbw, created cno PR 345 for above. ovnkube encap-port should wait for sdn-456 (cno pr 333) to land.

pkg/network/ovn_kubernetes.go
pecameron: Changed

pecameron: Made the change

pecameron:
@knobunc @JacobTanenbaum PTAL 
I addressed danwinship's comments.

PR #343, opened 2019-10-08
Bug 1762921: [4.2] The Multus admission controller should not have the not-ready toleration

dougbtv:
This ends up causing lots of errors to be logged during cluster start.
Additionally, it does not open any security holes as the Multus binary itself
is capable of performing namespace isolation (if/when that is added to the
admission controller in the future)


squeed:
@dougbtv this still neeeds a bz.

danwinship:
/retitle Bug 1762921: [4.2] The Multus admission controller should not have the not-ready toleration

PR #344, opened 2019-10-09
Add test to assert we fail if conntrack-max kubeproxy config option is passed

rcarrillocruz:
Since rebase to 1.16, conntrack-max is no longer a valid option
of KubeProxyConfig, thus degrade the operator if used.

### DIFF
--- a/pkg/util/k8s/kubeproxy_test.go
+++ b/pkg/util/k8s/kubeproxy_test.go
@@ -329,6 +329,13 @@ winkernel:
 			},
 			err: "unused arguments: blah-blah-blah",
 		},
+		{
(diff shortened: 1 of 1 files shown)


rcarrillocruz:
This will need to add tests, but awaiting reviews before to know if the location of fix is right.

danwinship:
I'm pretty sure you don't need to add any new code; `GenerateKubeProxyConfiguration` will already return an error if you pass in unknown flags, so when you removed the `kpc.Conntrack.Max = ka.getOptInt32("conntrack-max")` you made it an error to pass `"conntrack-max"` in `ProxyArguments`. We just need a unit test to confirm that.

PR #345, opened 2019-10-10
Add enable-protocol (geneve port) to ovs setup

pecameron:
This is needed on bare metal and clouds that rely on iptables.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -56,6 +56,7 @@ spec:
           trap quit SIGTERM
           /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
+          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
 
           tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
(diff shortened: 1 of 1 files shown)


pecameron:
@dcbw @squeed @danwinship PTAL

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -56,6 +56,7 @@ spec:
           trap quit SIGTERM
           /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
+          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
```
squeed: A bit confused - shouldn't we template in the value from the network operator config?

pecameron: @casey, this PR is not related to the genevePort configuration, PR 341 (which is blocked on PR 333), it is related to PR 329.
I will add the template in PR 341 when the genevePort is configurable. Adding this directly to PR 341 is fixing a problem that is not part of PR 341.

squeed: Gotcha. For my understanding, what does it fix?

pecameron: @squeed it opens the port in systems that use iptables (such as bare metal RHEL). In this case its the geneve port. Elsewhere in ovnkube.sh it opens the sdb(6642)/ndb(6641) ports.

pecameron:
@squeed @dcbw @JacobTanenbaum Is this ready to go in? PTAL

squeed:
@dcbw @pecameron do we need to set up the iptables chroot scripts for this?

PR #347, opened 2019-10-10
Simple TLS controller

squeed:
This adds a PKI CRD and a controller to implement it. Given a simple request for a PKI, it creates a CA and a single certificate. It delegates this to library-go's CertificateRotationController.

A bit of ugliness is needed to duct-tape a typed and untyped controller library, but it works well enough.

This will be used for implementing TLS for OVN.


danwinship:
> This will be used for implementing TLS for OVN.

Why can't we use the existing certificate-y stuff?

Also, any chance this could be used for IPsec?

squeed:
hey @deads2k, this uses the CertRotationController from library-go. Would you mind giving it a look-over to see that we're using it in an expected manner?

squeed:
> > This will be used for implementing TLS for OVN.
> 
> Why can't we use the existing certificate-y stuff?

Because the existing certificate managers all rely on the network to be up. The exception to this is the kube CA / CSR endpoint itself, but I don't want to use that for two reasons. First of all, it doesn't really save us a ton of effort, because it doesn't handle rotation. Second of all, I'd rather not issue certificates that have any sort of kube API identity.

> Also, any chance this could be used for IPsec?

Definitely. We may need to get a change in to library-go that allows us to set the ciphers.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: pkis.network.operator.openshift.io
+spec:
+  group: network.operator.openshift.io
+  names:
+    kind: PKI
```
squeed: I'm going to rename this to something more obscure. I don't want people discovering `oc create pki`

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
```
danwinship: since we're generating some other stuff now anyway, should we kubebuild this rather than writing it by hand?

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -0,0 +1,41 @@
...
+metadata:
+  name: pkis.network.operator.openshift.io
+spec:
+  group: network.operator.openshift.io
```
danwinship: Are we allowed to define "subdomains" of `operator.openshift.io` here? That's not reserved for future use by `openshift/api`?

pkg/controller/pki/pki_controller.go
```diff
@@ -0,0 +1,265 @@
...
+		clientset,
+		24*time.Hour,
+		informers.WithNamespace(config.Namespace))
+
+	cont, err := certrotation.NewCertRotationController(
+		fmt.Sprintf("%s/%s", config.Namespace, config.Name),
+		certrotation.SigningRotation{
+			Namespace:     config.Namespace,
+			Name:          spec.CACert.Name,
+			Validity:      spec.CACert.ValidFor.Duration,
```
deads2k: I would not expose duration values to end users.  You'll get people at the crazy ends of the spectrum and this certrotationcontroller is likely to misbehave at the ends.

/cc @sttts

pkg/apis/network/v1/pki_types.go
```diff
@@ -0,0 +1,60 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
```
danwinship: The documentation in this file completely fails to document the type...

pkg/apis/network/v1/pki_types.go
```diff
@@ -0,0 +1,60 @@
...
+// PKISpec defines the desired state of PKI.
+// +k8s:openapi-gen=true
+type PKISpec struct {
+	// caCert is information about the CA certificate and key.
+	CACert CertSpec `json:"caCert"`
```
danwinship: Is this information that the user provides or information that the controller fills in? If the former, what gets done with that information?

pkg/apis/network/v1/pki_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// caBundleName is the name of the config-map containing the CA bundle.
+	CABundleName string `json:"caBundleName"`
```
danwinship: What format is the config-map in? Why is this needed? If you're creating a CA then why do we need to provide a CA bundle? Or do you mean the newly-created CA gets published to the object named here? Why not just require the `ConfigMap` to have the same name as the `PKI`?

pkg/apis/network/v1/pki_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// serverCert is the server certificate to create.
+	// It will be created with both ServerAuth and ClientAuth
+	ServerCert CertSpec `json:"serverCert"`
```
danwinship: Maybe include the words "key usage" or something to make that clearer.
Also, why both? That seems weird.
Also, where does the certificate get created to?

pkg/apis/network/v1/pki_types.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PKI is a simple certificate authority. It creates a CA and generates
+// certificates signed by that CA.
+// +k8s:openapi-gen=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=pkis,scope=Namespaced
+type PKI struct {
```
danwinship: This type should come first in the file. Also, the docs make it sound like the API object itself is taking action, rather than just describing the actions that the CNO will take.

pkg/apis/network/v1/register.go
```diff
@@ -0,0 +1,38 @@
...
+	GroupVersion  = schema.GroupVersion{Group: GroupName, Version: "v1"}
+	schemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	// Install is a function which adds this version to a scheme
+	Install = schemeBuilder.AddToScheme
+
+	// SchemeGroupVersion generated code relies on this name
+	// Deprecated
+	SchemeGroupVersion = GroupVersion
+	// AddToScheme exists solely to keep the old generators creating valid code
+	// DEPRECATED
```
danwinship: are we actually using these "old generators"?

pkg/controller/pki/pki_controller.go
squeed: Fair enough.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
squeed: We can't use kube-builder without converting the whole codebase to a kubebuilder-managed project. We could switch to controller-gen. I'll look in to how dreadful that is.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
squeed: Good question. @deads2k, do you have any thoughts about this?

squeed: Good question. @deads2k, how do you feel about us using `network.operator.openshift.io` as a Group?

pkg/apis/network/v1/register.go
squeed: Hah, no. This is just boilerplate copied directly from openshift/api.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
squeed: Ah, it's not so bad.

pkg/apis/network/v1/pki_types.go
squeed: so it does.

pkg/apis/network/v1/pki_types.go
squeed: good point. hard-coding the names is easier.

squeed:
Thanks, as always, for the excellent review, updated.

I renamed PKI to OperatorPKI. Updated the documentation extensively. I also added in kubebuilder struct -> CRD conversion.

pkg/apis/network/v1/operatorpki_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// both ClientAuth and ServerAuth enabled
+	TargetCert CertSpec `json:"targetCert"`
+}
+
+// CertSpec defines common certificate configuration.
+type CertSpec struct {
+	// commonName is the value in the certificate's CN
+	//
+	// +kubebuilder:validation:MinLength=1
+	CommonName string `json:"commonName"`
```
danwinship: Is the `CertSpec` type still useful? Could we just put `TargetCertCommonName` in `OperatorPKISpec`?

pkg/controller/pki/pki_controller.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// sync causes the underlying cert controller to try and reconcile
+func (p *pki) sync() error {
+	return p.controller.RunOnce()
+}
+
+// toClientCert is a certificate "decorator" that adds ClientAuth to the
+// list of ExtendedKeyUsages. This allows the generated certificate to be
+// used for both client and server auth.
```
danwinship: Oh, yeah, you didn't answer why you were doing this. Are you going to have the client and server using the same cert? Why not just make two certificates? That seems way more sane...

pecameron:
@squeed Your comment at the top says it creates 1 cert. Are we using the same cert/ca for both sdb and ndb, or are there two of these?

pkg/controller/pki/pki_controller.go
squeed: for OVN, we have to use the same cert, since the databases will be either servers or clients, based on the replication state.

Also, the internal library-go code currently only allows creating a single cert. Eventually we could improve this, but it's not necessary now.

pkg/apis/network/v1/operatorpki_types.go
squeed: I suspect we'll be adding more properties (e.g. validity), so I decided to keep it.

squeed:
> @squeed Your comment at the top says it creates 1 cert. Are we using the same cert/ca for both sdb and ndb, or are there two of these?

Yup. The intended use case for this is OVN. We're creating a single cert for all users.

squeed:
/retest

(aws flake)

danwinship:
I'm vaguely worried that there may be some cryptographic reason why it would be bad to have the client and server using the same keys, but the only discussion of that I can find is https://crypto.stackexchange.com/questions/7874/client-and-server-using-same-ssl-certificate-any-issues which suggests it's fine other than the obvious issues we don't care about.

squeed:
Establishing a mTLS channel consists of the client sending `Encrypt(Server, nonce)` followed by `Sign(Client, challenge)`. If there was a way to recover the key from one Encrypt and one Sign operation, then we're in big trouble.

Note: the `challenge` is not actually chosen, it's a hash of the TLS protocol state. So you can't use the client cert as an oracle. Clever.

danwinship:
Oh, right, the client certificate doesn't actually play any role in establishing the encryption. (I was worrying that if both sides started with the same secrets then some of the math might "cancel out" and result in a predictable symmetric encryption key being generated.)

squeed:
/hold
to stop the retesting

squeed:
hurray, kube-builder doesn't create creatable CRDs:

`The CustomResourceDefinition "operatorpkis.network.operator.openshift.io" is invalid: spec.versions: Invalid value: []apiextensions.CustomResourceDefinitionVersion{apiextensions.CustomResourceDefinitionVersion{Name:"v1", Served:true, Storage:true, Schema:(*apiextensions.CustomResourceValidation)(0xc01ffd42a8), Subresources:(*apiextensions.CustomResourceSubresources)(nil), AdditionalPrinterColumns:[]apiextensions.CustomResourceColumnDefinition(nil)}}: per-version schemas may not all be set to identical values (top-level validation should be used instead)`

squeed:
Needed to pick up the latest (unreleased) controller-tools version.

squeed:
aws flake
/retest

squeed:
registry flake
/retest

squeed:
hey @danwinship looks like the event got lost - would you mind re-tagging?

PR #354, opened 2019-10-17
Bug 1755586: Removing simpleMacvlanConfig from CNO not delete old network-attachment-definition

s1061123:
This change adds to remove net-attach-def object if corresponding additionalNetwork is removed in CNO CRD object.


pkg/network/render.go
```diff
@@ -389,3 +393,48 @@ func RenderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
 	out = append(out, objs...)
 	return out, nil
 }
+
+func ClearNetAttachDef(client cli.Client, prev *operv1.NetworkSpec, conf *operv1.NetworkSpec) error {
```
squeed: nit: This should be in `pkg/controller/operconfig_controller`, not `pkg/network`.

squeed: also, can you add a docblock describing this?

pkg/network/render.go
```diff
@@ -389,3 +393,48 @@ func RenderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
...
+
+	for _, netAttach := range deleteNetAttachDefs {
+		log.Printf("delete net-attach-def/%s namespace: %s", netAttach.Name, netAttach.Namespace)
+
+		if netAttach.Namespace == "" {
+			netAttach.Namespace = "default"
+		}
+
+		obj := &netattachv1.NetworkAttachmentDefinition{}
+		err := client.Get(context.TODO(), cli.ObjectKey{
```
squeed: Why bother getting it? I don't think you need to.

pkg/network/render.go
```diff
@@ -389,3 +393,48 @@ func RenderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
...
+			Namespace: netAttach.Namespace,
+			Name:      netAttach.Name,
+		}, obj)
+		if err != nil {
+			log.Printf("failed to get net-attach-def %s in namespace %s: %v", netAttach.Name, netAttach.Namespace, err)
+			break
+		}
+
```
squeed: You should only delete the NetAttachDef if the OwnerReference is set to ours. That way we won't accidentally step on anyone who re-creates it.

pkg/network/render.go
```diff
@@ -389,3 +393,48 @@ func RenderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
...
+		if netAttach.Namespace == "" {
+			netAttach.Namespace = "default"
+		}
+
+		obj := &netattachv1.NetworkAttachmentDefinition{}
+		err := client.Get(context.TODO(), cli.ObjectKey{
+			Namespace: netAttach.Namespace,
+			Name:      netAttach.Name,
+		}, obj)
+		if err != nil {
```
squeed: Don't print an error if `apierrors.IsNotFound()`

pkg/network/render.go
```diff
@@ -389,3 +393,48 @@ func RenderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
...
+			log.Printf("failed to get net-attach-def %s in namespace %s: %v", netAttach.Name, netAttach.Namespace, err)
+			break
+		}
+
+		err = client.Delete(context.TODO(), obj)
+		if err != nil {
+			log.Printf("failed to delete net-attach-def %s in namespace %s: %v", netAttach.Name, netAttach.Namespace, err)
```
squeed: Don't print an error if apierrors.IsNotFound()

squeed:
overall I think this is OK. We have a more robust mechanism for deleting stale objects in #306, but that won't be backported to previous versions, so I would be OK with this also being merged.

pkg/network/render.go
s1061123: changed.

pkg/network/render.go
s1061123: fixed.

pkg/network/render.go
s1061123: Done.

s1061123:
@dcbw PTAL?

s1061123:
@dcbw PTAL?

s1061123:
@dcbw PTAL?

PR #355, opened 2019-10-18
Switch back to upstream controller-runtime 1.16 dep

rcarrillocruz:


squeed:
/lgtm
/approve

If there's a tagged release before code freeze, let's make sure we switch to it.

squeed:
I don't think we should be bumping gophercloud when we pick this up. I would just remove the controller-runtime, then do `dep ensure -add sigs.k8s.io/controller-runtime@COMMIT`

PR #357, opened 2019-10-21
OVNKubernetes: introduce OVS anti-selector

mmirecki:
OVNKubernetes is shipped with its own openvswitch service running
as a separate daemonset. However, in some cases we may want to use
an external openvswitch running on the host (e.g. to use it for
host networking). In order to support a mixed environment where some
hosts have this external service and some don't, we introduce an
openvswitch anti-selector:

network.operator.openshift.io/external-openvswitch

Note that this selector has to be set before the operator is started.
Otherwise, the label would not take effect on already scheduled
pods.

Related to:
https://github.com/openshift/cluster-network-operator/pull/346


bindata/network/ovn-kubernetes/ovs-node.yaml
```diff
@@ -0,0 +1,155 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: ovnkube-node
```
phoracek: Most of the names and labels here need to be changed.

bindata/network/ovn-kubernetes/ovs-node.yaml
```diff
@@ -0,0 +1,155 @@
...
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the ovn-kubernetes per node networking components.
```
phoracek: This is not correct anymore

bindata/network/ovn-kubernetes/ovs-node.yaml
```diff
@@ -0,0 +1,143 @@
...
+        hostPath:
+          path: /lib/modules
+      - name: var-lib-openvswitch
+        hostPath:
+          path: /var/lib/openvswitch/data
+      - name: etc-openvswitch
+        hostPath:
+          path: /var/lib/openvswitch/etc
+      - name: run-openvswitch
+        emptyDir: {}
```
phoracek: Commenting here since I can't on the OVN volume:

This code will not work with OVS installed on host, will it? It would create OVS socket in an empty tmp directory. We need to make sure that we mount OVS socket running on host if external-ovs is used.

squeed: Yes, you'll have to change this to a `hostPath` mount. That's fine.

bindata/network/ovn-kubernetes/ovs-node.yaml
```diff
@@ -0,0 +1,143 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: ovs-node
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches ovs per node networking components.
```
dcbw: ovs -> Open vSwitch
per node -> per-node

bindata/network/ovn-kubernetes/ovs-node.yaml
dcbw: @phoracek @squeed I don't quite understand. Wouldn't this daemonset only run on nodes that don't have the external OVS? If so, why would it care that it "won't work with OVS installed on the host"?

dcbw:
@phoracek @mmirecki Whats the case where some nodes will have system OVS and others will have containerized OVS? Why wouldn't this be consistent across the entire cluster?

phoracek:
@dcbw in case you have RHCOS8 masters and RHEL7 workers, where you want to use OVS SLB bonding for workload. I know, very specific. It is easy to install external OVS on RHEL, very hard to have it on RHCOS.

bindata/network/ovn-kubernetes/ovs-node.yaml
phoracek: @dcbw I commented here, but it was meant for the OVN DS.

squeed: @dcbw correct - we now have separated ovs-vswitchd from the ovnkube-node pod. That means that we need to share `/run/openvswitch' between the two pods (or from the host's ovs)

That means that, for all pods, `/run/openvswitch` needs to be a hostPath volume.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -275,14 +196,12 @@ spec:
         hostPath:
           path: /var/lib/openvswitch/etc
       - name: run-openvswitch
-        emptyDir: {}
+        hostPath:
+          path: /run/openvswitch
```
phoracek: Could you verify that this is indeed the path when the OVS socket lives when OVS is installed on RHEL7 from an RPM?

bindata/network/ovn-kubernetes/ovs-node.yaml
```diff
@@ -0,0 +1,144 @@
...
+        beta.kubernetes.io/os: "linux"
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: network.operator.openshift.io/external-openvswitch
+                operator: DoesNotExist        
+      volumes:
+      # used for iptables wrapper scripts
```
phoracek: Do we need iptables wrapper for OVS?

danwinship: no

bindata/network/ovn-kubernetes/ovnkube-node.yaml
mmirecki: Yes
Note also that /var/run/openvswitch is a link to /run/openvswitch.

phoracek:
@squeed similar patch landed for OpenShiftSDN, could we move this one through the finish line?

mmirecki:
The OpenShiftSDN equivalent [PR 346](https://github.com/openshift/cluster-network-operator/pull/346) is merged. 
Can we proceed with this one?

phoracek:
@pecameron we'd appreciate if this got in, so we can start initial testing of OVN with external OVS before your patches land.

squeed:
Okay, now that 4.4 has cut, we can try and merge this. @JacobTanenbaum , can you review please?

pecameron:
@danwinship @dcbw @JacobTanenbaum PTAL after rebase.

mmirecki:
Rebased, please review

PR #358, opened 2019-10-21
controller/status: set Degraded if rollout is hung.

squeed:
If we find ourselves with a DaemonSet or Deployment that hasn't made any progress for the past 10 minutes, then highlight this to the operator as a Degraded state.

This doesn't affect any of the other states (Progressing, Available, etc.).

/cc @danwinship @rcarrillocruz

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -1,0 +1,278 @@
+package statusmanager
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"reflect"
+	"strings"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	client "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// if a rollout has not made any progress by this time,
+	// mark ourselves as Degraded
+	ProgressTimeout = 10 * time.Minute
+
+	// lastSeenAnnotation - the annotation where we stash our state
+	lastSeenAnnotation = "network.operator.openshift.io/last-seen-state"
+)
+
+// podState is a snapshot of the last-seen-state and last-changed-times
+// for pod-creating entities, as marshalled to json in an annotation
+type podState struct {
+	// "public" for marshalling to json, since we can't have complex keys
+	DaemonsetStates  []daemonsetState
+	DeploymentStates []deploymentState
+}
+
+// daemonsetState is the internal state we use to check if a rollout has
+// stalled.
+type daemonsetState struct {
+	types.NamespacedName
+
+	LastSeenStatus appsv1.DaemonSetStatus
+	LastChangeTime time.Time
+}
+
+// deploymentState is the same as daemonsetState.. but for deployments!
+type deploymentState struct {
+	types.NamespacedName
+
+	LastSeenStatus appsv1.DeploymentStatus
+	LastChangeTime time.Time
+}
+
+// SetFromPods sets the operator Degraded/Progressing/Available status, based on
+// the current status of the manager's DaemonSets and Deployments.
+func (status *StatusManager) SetFromPods() {
+	status.Lock()
+	defer status.Unlock()
+
+	targetLevel := os.Getenv("RELEASE_VERSION")
+	reachedAvailableLevel := (len(status.daemonSets) + len(status.deployments)) > 0
+
+	progressing := []string{}
+	hung := []string{}
+
+	daemonsetStates, deploymentStates := status.getLastPodState()
+
+	for _, dsName := range status.daemonSets {
+		ds := &appsv1.DaemonSet{}
+		if err := status.client.Get(context.TODO(), dsName, ds); err != nil {
+			log.Printf("Error getting DaemonSet %q: %v", dsName.String(), err)
+			progressing = append(progressing, fmt.Sprintf("Waiting for DaemonSet %q to be created", dsName.String()))
+			// Assume the OperConfig Controller is in the process of reconciling
+			// things; it will set a Degraded status if it fails.
+			continue
+		}
+
+		dsProgressing := false
+
+		if ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
+			dsProgressing = true
+		} else if ds.Status.NumberUnavailable > 0 {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
+			dsProgressing = true
+		} else if ds.Status.NumberAvailable == 0 { // NOTE: update this if we ever expect empty (unscheduled) daemonsets ~cdc
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not yet scheduled on any nodes", dsName.String()))
+			dsProgressing = true
+		} else if ds.Generation > ds.Status.ObservedGeneration {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is being processed (generation %d, observed generation %d)", dsName.String(), ds.Generation, ds.Status.ObservedGeneration))
+			dsProgressing = true
+		}
+
+		if dsProgressing || ds.Annotations["release.openshift.io/version"] != targetLevel {
+			reachedAvailableLevel = false
+		}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -49,11 +57,26 @@ type StatusManager struct {
...
-	deployments    []types.NamespacedName
+	daemonSets     map[types.NamespacedName]*daemonsetState
+	deployments    map[types.NamespacedName]*deploymentState
 	relatedObjects []configv1.ObjectReference
 }
 
+// daemonsetState is the internal state we use to check if a rollout has
+// stalled. Since we don't get a last-changed time from the apiserver, we
+// need to synthesize it ourselves. This means that we may flap between
+// Degraded if the operator restarts.
```
danwinship: Clayton hates it when operator statuses flap on startup... we could put the last change time into the Progressing message and then parse it back out on each update...

pkg/controller/statusmanager/status_manager.go
```diff
@@ -225,13 +248,29 @@ func (status *StatusManager) SetNotDegraded(statusLevel StatusLevel) {
...
 	status.Lock()
 	defer status.Unlock()
-	status.daemonSets = daemonSets
+	out := map[types.NamespacedName]*daemonsetState{}
+	for _, name := range daemonSets {
+		if existing, ok := status.daemonSets[name]; ok {
+			out[name] = existing
+		} else {
+			out[name] = nil
+		}
```
danwinship: That basically boils down to `out[name] = status.daemonSets[name]`

pkg/controller/statusmanager/status_manager.go
```diff
@@ -261,22 +301,45 @@ func (status *StatusManager) SetFromPods() {
 			continue
 		}
 
+		// Update our last-seen-state if new or changed
+		if dsState == nil {
```
danwinship: Does using a `*daemonState` rather than a `daemonState` buy you anything here? (ie, if you have a zero-initialized `dsState` then is `reflect.DeepEqual(ds.Status, dsState.lastSeenStatus)` guaranteed to be `false` and so you'd hit the next clause anyway?)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -261,22 +301,45 @@ func (status *StatusManager) SetFromPods() {
...
+			dsProgressing = true
 		}
 
-		if !(ds.Generation <= ds.Status.ObservedGeneration && ds.Status.UpdatedNumberScheduled == ds.Status.DesiredNumberScheduled && ds.Status.NumberUnavailable == 0 && ds.Annotations["release.openshift.io/version"] == targetLevel) {
+		if dsProgressing || ds.Annotations["release.openshift.io/version"] != targetLevel {
 			reachedAvailableLevel = false
 		}
+
+		// Catch hung rollouts
+		if dsProgressing && (time.Now().Sub(dsState.lastChangeTime)) > ProgressTimeout {
```
danwinship: `time.Since(dsState.lastChangeTime)`

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -494,9 +496,238 @@ func TestStatusManagerSetFromDaemonSets(t *testing.T) {
...
+		},
+		{
+			Type:   configv1.OperatorAvailable,
+			Status: configv1.ConditionTrue,
+		},
+	}) {
+		t.Fatalf("unexpected Status.Conditions: %#v", co.Status.Conditions)
+	}
+
+	// intermission: set the last-update-time to last year, make sure we
```
danwinship: an hour ago, not last year

pkg/controller/statusmanager/status_manager.go
squeed: Argh. Fine. We need to know the last change time of each daemonset/deployment individually. I suppose we could marshall it as yaml so it's reasonably readable.

pkg/controller/statusmanager/status_manager.go
squeed: hah, so it does.

pkg/controller/statusmanager/status_manager.go
squeed: I thought about that for a while; I decided I hate zero values (especially for time) and didn't want to think about if a zero-value Status is possible.

pkg/controller/statusmanager/status_manager.go
squeed: much neater, thanks!

pkg/controller/statusmanager/status_manager.go
danwinship: You can just add it to the individual lines (`"Deployment openshift-sdn/sdn is not available (awaiting 1 nodes since 2019-10-22T14:22:00)"`.
Or use an annotation if you don't like putting it in the status.

squeed:
@danwinship rewrote this to store status in a type and marshal it to json. Thanks

squeed:
Uh... I was kind of tired when I wrote that comment. I meant to say: I rewrote this to store state in an annotation on the ClusterOperator in json.

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+	ProgressTimeout = 10 * time.Minute
+
+	// lastSeenAnnotation - the annotation where we stash our state
+	lastSeenAnnotation = "network.operator.openshift.io/last-seen-state"
+)
+
+// podState is a snapshot of the last-seen-state and last-changed-times
+// for pod-creating entities
+type podState struct {
+	// "public" for marshalling to json, since we can't have complex keys
```
danwinship: It seems like it would be make more sense to just have a separate type for what gets json-marshaled since these fields are only used inside setLastPodState/getLastPodState?

Actually you don't even need a struct type for the internal type; just have the helper funcs take/return a pair of maps, and convert to/from a struct-of-arrays internally.

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// maps for internal use
+	daemonsetState  map[types.NamespacedName]daemonsetState
+	deploymentState map[types.NamespacedName]deploymentState
+}
+
+// daemonsetState is the internal state we use to check if a rollout has
+// stalled. Since we don't get a last-changed time from the apiserver, we
+// need to synthesize it ourselves. This means that we may flap between
+// Degraded if the operator restarts.
```
danwinship: this comment is now wrong?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+type daemonsetState struct {
+	LastSeenStatus appsv1.DaemonSetStatus
+	LastChangeTime time.Time
+	Name           types.NamespacedName
```
danwinship: maybe inline the `types.NamespacedName` field?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		if !ok {
+			dsState.Name = dsName
+			dsState.LastChangeTime = time.Now()
+			ds.Status.DeepCopyInto(&dsState.LastSeenStatus)
+			lastState.daemonsetState[dsName] = dsState
+		}
+
+		// Catch hung rollouts
+		if dsProgressing && ok && (time.Since(dsState.LastChangeTime)) > ProgressTimeout {
```
danwinship: You could move the `if dsProgressing` check out around the entire `lastState` thing.
Also, you should remove no-longer-`Progressing` DaemonSets from the state.

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+
+		if dsProgressing || ds.Annotations["release.openshift.io/version"] != targetLevel {
+			reachedAvailableLevel = false
+		}
+
+		dsState, ok := lastState.daemonsetState[dsName]
+		if ok && !reflect.DeepEqual(dsState.LastSeenStatus, ds.Status) {
+			ok = false // invalidate
+		}
+		if !ok {
```
danwinship: you could simplify the two checks to `if !ok || !reflect.DeepEqual(dsState.LastSeenStatus, ds.Status)`. And maybe not use `ok` as a name? It's so vague. (`exists`? `set`?)

pkg/controller/statusmanager/pod_status.go
```diff
@@ -0,0 +1,278 @@
...
+			depState.Name = depName
+			depState.LastChangeTime = time.Now()
+			dep.Status.DeepCopyInto(&depState.LastSeenStatus)
+			lastState.deploymentState[depName] = depState
+		}
+
+		// Catch hung rollouts
+		if depProgressing && ok && (time.Since(depState.LastChangeTime)) > ProgressTimeout {
```
danwinship: as above you could move the `if depProgressing` earlier, and you need to do cleanup

pkg/controller/statusmanager/pod_status.go
squeed: yeah, good point.

pkg/controller/statusmanager/pod_status.go
squeed: Sure

squeed:
Test failure is due to https://bugzilla.redhat.com/show_bug.cgi?id=1765280
/retest

squeed:
@danwinship updated, PTAL

PR #359, opened 2019-10-22
Update the api to make way for windows/linux hybrid networking

JacobTanenbaum:
dep ensure -update github.com/openshift/api

needed to bring in changes to the OVNKubernetesConfigStruct which added new config options


danwinship:
/lgtm cancel
doesn't build

JacobTanenbaum:
@danwinship fixed the commit

PR #360, opened 2019-10-22
Bug 1764256: Handle kubeconfigs with an empty context.

squeed:
Fixes a nil panic for kubeconfigs with no context.


danwinship:
Have you tested this against one of their weird kubeconfigs?

squeed:
What in the heck is going on here!?

squeed:
Ah, this got b0rken by the switch from master to 4.4. I'll have to close this and re-open against 4.3.

PR #361, opened 2019-10-22
Bump image versions to 4.3 now that 4.2 have been GA'd

rcarrillocruz:


squeed:
for the record, this is officially a no-op (since these values are just placeholders), but we have some development scripts that rely on them.

/lgtm
/approve

PR #364, opened 2019-10-24
The reference CNI plugins should be copied based on the node's host OS

dougbtv:
Depending on the host OS the CNI plugins may need to be build specifically,
so these are stored in alternate directories and copied according to the host OS.

...Putting on hold until I can properly test (and potentially) the newly added bash script used to copy the CNI binaries.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -31,13 +31,42 @@ spec:
       initContainers:
       - name: cni-plugins
         image: {{.CNIPluginsImage}}
-        command: ["/bin/sh"]
-        args: ["-c", "cp -rf /usr/src/plugins/bin/* /host/opt/cni/bin"]
+        command: 
+          - /bin/bash
+          - -c
+          - |
+            #!/bin/bash
+            set -x
+            . /host/etc/os-release
+            rhelmajor=
+            # detect which version we're using in order to copy the proper binaries
+            case "${ID}" in
+              rhcos) rhelmajor=8
+              ;;
+              rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .) 
+              ;;
+              *) echo "FATAL ERROR: Unsupported OS ID=${ID}"; exit 1
+              ;;
+            esac
+            # Set which directory we'll copy from, detect if it exists
+            # When it doesn't exist, fall back to the original directory.
+            SOURCEDIR=/usr/src/plugins/rhel${rhelmajor}/bin/
+            if [ -d "${SOURCEDIR}" ]; then
+              echo "Detected OS version ${rhelmajor}, ${SOURCEDIR} exists"
+            else
+              echo "Source directory unavailable for OS version: ${rhelmajor}"
+              SOURCEDIR=/usr/src/plugins/bin/
+            fi
+            cp -rf ${SOURCEDIR}* /host/opt/cni/bin
+
         securityContext:
           privileged: true
         volumeMounts:
         - mountPath: /host/opt/cni/bin
           name: cnibin
+        - mountPath: /host/etc/os-release
+          name: os-release
+          readOnly: true
       containers:
       - name: kube-multus
         image: {{.MultusImage}}
@@ -79,3 +108,7 @@ spec:
         - name: cnibin
           hostPath:
             path: /var/lib/cni/bin
+        - name: os-release
+          hostPath:
+            path: /etc/os-release
+            type: File


lsm5:
[@mrunalp, @squeed please correct me if I'm wrong below..]

We need this so that plugins can be built without for rhel7 and rhel8 without static linking in glibc of a particular rhel version and then the right version of the binaries can be installed by detecting the host's rhel version.

dougbtv:
My initial tests of this are passing. Notably, this can only be somewhat tested before the CNI plugins image is updated with the new paths. Otherwise, it just always falls back to the original path.

...however, it should be backwards compatible with the current CNI plugins image.

@lsm5 -- do you know the reliable NON-symbolic-link path to the `os-release` file across both rhel7/rhel8? That's my only major concern remaining (aside from noted limitation of testing above) -- currently /etc/os-release is a symlink to /usr/lib/os-release -- I decided to account for having both and preferring /etc/os-release if available. I can't refer to a symlink, though.

dougbtv:
/retitle The reference CNI plugins should be copied based on the node's host OS
/hold cancel

bindata/network/multus/004-configmap.yaml
```diff
@@ -0,0 +1,40 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cni-copy-bash-script
```
danwinship: In general we just inline shell scripts into the pod specs. eg, look at [`bindata/network/openshift-sdn/sdn.yaml`](https://github.com/dougbtv/cluster-network-operator/blob/copy-cni-plugins-rhel7-rhel8/bindata/network/openshift-sdn/sdn.yaml#L56)

bindata/network/multus/004-configmap.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: openshift-multus
+  labels:
+    app: openshift-multus
+data:
+  copycni.sh: |-
+    #!/bin/bash
+    set -x
+    # source our os-release variables.
+    if [ -f "/host/etc-os-release/os-release" ]; then
```
danwinship: So rather than all of this, what about just `uname -r`? If it contains `.el7` you're on RHEL 7 and if it contains `.el8` you're on RHEL 8.

bindata/network/multus/004-configmap.yaml
```diff
@@ -0,0 +1,40 @@
...
+      . /host/usr/lib/os-release
+    fi
+    rhelmajor=
+    # detect which version we're using in order to copy the proper binaries
+    case "${ID}" in
+      rhcos) rhelmajor=8
+      ;;
+      rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .) 
+      ;;
+      *) echo "WARNING: Unsupported OS ID=${ID}"
```
danwinship: This should probably be a fatal error

bindata/network/multus/004-configmap.yaml
lsm5: @dougbtv fwiw, on fedora `/etc/os-release` is a symlink to `/usr/lib/os-release` but on rhel8 and rhel7, they don't seem to be symlinks, atleast on my envs, but anyway `uname -r` looks great.

bindata/network/multus/multus.yaml
```diff
@@ -31,13 +31,22 @@ spec:
...
         image: {{.CNIPluginsImage}}
-        command: ["/bin/sh"]
-        args: ["-c", "cp -rf /usr/src/plugins/bin/* /host/opt/cni/bin"]
+        command: ["/bin/copycni.sh"]
         securityContext:
           privileged: true
         volumeMounts:
         - mountPath: /host/opt/cni/bin
           name: cnibin
+        - mountPath: /host/etc-os-release
```
squeed: I wouldn't do it this way. If you mount `/etc/os-release` in to the container, then crio will follow the symlinks and it will Just Work.

Just make sure the hostPath volume is `type: File`

bindata/network/multus/004-configmap.yaml
cgwalters: > So rather than all of this, what about just uname -r? If it contains .el7 you're on RHEL 7 and if it contains .el8 you're on RHEL 8.

Mrm...yes but `/etc/os-release` is *designed* for this use case...parsing kernel version strings isn't.

danwinship: Hard-coding knowledge of where RHEL points its symlinks to is gross. But given Casey's comment below about using a `type: File` mount, we can just parse `/etc/os-release`.

bindata/network/multus/multus.yaml
dougbtv: Thanks for the tip on the `type: File` -- updated (and removed the alternate method to look at both locations)

bindata/network/multus/004-configmap.yaml
dougbtv: Good call updated the echo output and added `exit 1`

bindata/network/multus/004-configmap.yaml
dougbtv: I kept the methodology to use the `/etc/os-release`, and have now implemented Casey's suggesiton on the `type: File` so I could remove the double dip for the file + symlink.

bindata/network/multus/004-configmap.yaml
dougbtv: Thanks for the example -- nice to not have it floating out in another file, got this updated to have it inline'd per the example.

knobunc:
/lgtm

Thanks Doug

PR #366, opened 2019-10-28
verify if an older version of the ovs database exists

JacobTanenbaum:
commit 0ac2cdf1... changes the location of the ovs database. The database was originally in
/etc/origin/openvswitch on the host and after Casey's change got moved to /var/lib/openvswitch. In the upgrade case where a database exists in the old location move the database to the new location so the cluster's networking remains stable.


JacobTanenbaum:
@dcbw PTAL

PR #379, opened 2019-10-31
multus: copy RHEL8 binaries on FCOS

vrutkovs:
This ensures OKD-on-FCOS doesn't crash when FCOS is used as a base. Currently RHEL8 binaries are used, proper Fedora support would be added later.

See https://github.com/openshift/installer/pull/2548#issuecomment-548474221

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -45,6 +45,14 @@ spec:
               ;;
               rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .) 
               ;;
+              fedora)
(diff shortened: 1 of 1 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -45,6 +45,14 @@ spec:
               ;;
               rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .) 
               ;;
+              fedora)
+                if [ "${VARIANT_ID}" == "coreos" ]; then
+                  rhelmajor=8
```
cgwalters: This will eventually become problematic when FCOS moves on a lot farther from RHEL8 - say a new openssl or whatever.  But for now seems sane.

PR #380, opened 2019-11-01
adding initial alert rules to ovnkubernetes

JacobTanenbaum:
- check that every node has a running ovnkube-node pod
 - check that none of the pods in openshift-ovn-kubernetes namespace are in CrashLoopBackoff

### DIFF
--- a/bindata/network/ovn-kubernetes/alert-rules.yaml
+++ b/bindata/network/ovn-kubernetes/alert-rules.yaml
@@ -1,0 +1,32 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  annotations:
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/alert-rules.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: networking-rules
+  namespace: openshift-sdn
```
squeed: openshift-ovn-kubernetes

JacobTanenbaum: done

PR #381, opened 2019-11-01
ovn/master: enable unidling

dcbw:
@danwinship @knobunc @JacobTanenbaum @pecameron @squeed


dcbw:
> does this go into the ovnkube-config configmap as well?

@pecameron it could, but since it's something we'll enable everywhere, I think it's OK to keep it as a CLI flag. At least for now, it's not going to depend on any CNO/cluster configuration which is where the configmap helps us.

PR #382, opened 2019-11-04
[wip] raft so I can see CI

squeed:


squeed:
Bah, it gave Alexander access to the CI env, not me.

PR #385, opened 2019-11-06
Bug 1761609: Remove SDN ds liveness check

rcarrillocruz:
Fixes BZ 1761609

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -159,11 +159,6 @@ spec:
           preStop:
             exec:
               command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf", "/host/opt/cni/bin/openshift-sdn"]
-        # this comes from the kube-proxy code
-        livenessProbe:
-          httpGet:
-            path: /healthz
-            port: healthz
         readinessProbe:
           exec:
             # openshift-sdn writes this file when it is ready to handle pod requests.


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -159,11 +159,6 @@ spec:
           preStop:
             exec:
               command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf", "/host/opt/cni/bin/openshift-sdn"]
-        # this comes from the kube-proxy code
```
abhat: @squeed do you think this should be removed altogether or we should just tweak around the probe parameters at scale to account for slower response times?

danwinship: The problem (IIRC) is that the kube-proxy liveness server doesn't get started until after the sdn "node" stuff is fully started up, so in large clusters where startup takes longer, the kubelet may time out on the liveness check before it's even running. So the fix, if we want to keep the liveness check, is to rearrange sdn startup. But I think there was some feeling that the liveness check isn't really super helpful anyway...

squeed:
Right.
At this point, we've not seen the liveness check detect any sort of liveness failure.. Especially troublesome is that if we take a long time to fill our informer caches (because of slow apiserver or big cluster), the liveness check will shoot us in the head. I know that I had assumed the liveness would hold off until the pod entered Ready state. Nope.


/approve
/lgtm

PR #334, opened 2019-10-04
Bug 1758656: Removes external api server from default noProxy

danehans:
4.2.z backport of https://bugzilla.redhat.com/show_bug.cgi?id=1757973

PR #388, opened 2019-11-07
Bug 1766066: Revert "Merge pull request #334 from danehans/bz_1758656"

eparis:
This reverts commit a85860904ef2c95cd92e50ca979dc4563ee71b8e, reversing
changes made to 1eed8132527076b9ffca3dc244001a39855f1545.

We found in BZ https://bugzilla.redhat.com/show_bug.cgi?id=1762618 that at least
some of our code, running on cluster, was accessing the external instead of internal
api.

While it is possible to make it work (by allowing your proxy to talk from internal
to the apiserver) this is a change/regression from GA. Since we hit this with at
least 1 part of our code we have decided to revert and try again with more testing.


eparis:
manually overriding BZ check since this is a revert and the BZ in question is the long term fix...

PR #389, opened 2019-11-07
Bug 1725832: The Multus binary should be copied based on the host operating system

dougbtv:
This adds a new flag to `--skip-multus-binary-copy` in the Multus entrypoint
and instead uses an init container to make the copy based on the host OS.
Notably, this retains the atomic opeation of copying the Multus binary
to a temporary filename and then renaming the file, as is used in the Multus
entrypoint.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -75,6 +75,55 @@ spec:
         - mountPath: /host/etc/os-release
           name: os-release
           readOnly: true
+      - name: multus-binary-copy
+        image: {{.MultusImage}}
+        command:
+          - /bin/bash
+          - -c
+          - |
+            #!/bin/bash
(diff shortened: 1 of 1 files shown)


cgwalters:
@dougbtv 

In https://github.com/coreos/fedora-coreos-tracker/issues/354 I referenced this PR because I thought  it was doing the *build* part but now I look it seems actually it's just the "copy bits out" and I'm not seeing the Dockerfile that builds for both in this repository.

In https://github.com/openshift/machine-config-operator/pull/1766 we're looking at doing the same trick with the MCO and I'd like to reference/inspect the existing network code doing this, do you have a reference to it?

danwinship:
https://github.com/openshift/multus-cni/blob/master/Dockerfile.openshift

PR #390, opened 2019-11-08
more accuratly alert on a single nodes iptable stale

JacobTanenbaum:
change from a query that relies on time() to one that uses the more accurate timestamp()

### DIFF
--- a/bindata/network/openshift-sdn/alert-rules.yaml
+++ b/bindata/network/openshift-sdn/alert-rules.yaml
@@ -56,9 +56,11 @@ spec:
     - alert: NodeIPTablesStale
     # there is some scrape delay and some other offset 120 is not really 120s but it is still too long
       annotations:
-        message: SDN pod {{"{{"}} $labels.pod {{"}}"}} on node {{"{{"}} $labels.node {{"}}"}} has gone too long without syncing iptables rules. NOTE - There is some scrape delay and other offsets, 120s isn't exact but it is still too high.
+        message: SDN pod {{"{{"}} $labels.pod {{"}}"}} on node {{"{{"}} $labels.node {{"}}"}} has gone too long without syncing iptables rules.
       expr: |
-        (time() - kubeproxy_sync_proxy_rules_last_timestamp_seconds) * on(pod) group_right kube_pod_info{namespace="openshift-sdn",  pod=~"sdn-[^-]*"} > 120
+            (timestamp(kubeproxy_sync_proxy_rules_last_timestamp_seconds)
+            - on(pod) kubeproxy_sync_proxy_rules_last_timestamp_seconds)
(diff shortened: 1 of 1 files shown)


danwinship:
huh... so can you translate exactly what the old query does and what the new version does?

(Also, as mentioned in https://jira.coreos.com/browse/SDN-584, `kubeproxy_sync_proxy_rules_last_timestamp_seconds` will stop being useful in 4.4 so we need to eventually figure out something new.)

JacobTanenbaum:
the old query:
`(time() - kubeproxy_sync_proxy_rules_last_timestamp_seconds) * on(pod) group_right kube_pod_info{namespace="openshift-sdn",  pod=~"sdn-[^-]*"}`
measures the difference between the time since the current time and the last time that kubeproxy was sync'ed 

the new query:
`timestamp(kubeproxy_sync_proxy_rules_last_timestamp_seconds)
            - on(pod) kubeproxy_sync_proxy_rules_last_timestamp_seconds
            * on(pod) group_right kube_pod_info{namespace="openshift-sdn",  pod=~"sdn-[^-]*"}`

measures the difference between the last time prometheus "read" the value of `kubeproxy_sync_proxy_rules_last_timestamp_seconds` and what the value of `kubeproxy_sync_proxy_rules_last_timestamp_seconds` is.

the old query is much more susceptible to scrape delay and clock skew  
@danwinship, @squeed did I explain this correctly?

danwinship:
That's a good explanation of the big picture but I don't understand the details either... Like, why did it change from "`time() - METRICNAME`" to "`timestamp() - on(pod) METRICNAME`"? What does the "`on(pod)`" do? And why was that whole clause before the `*` in parentheses before but it's not now?

squeed:
where `time()` is a scalar value, `timestamp(metricname)` is a synthetic instant vector. That means that it needs to be joined to it's corresponding entry in the `metricname` instant vector.

I know this is kind of crazy, but one of the cool things of promql is that you don't need to join most of the time. This means that, when you do, it becomes a bit surprising. The implicit joins are super useful.

danwinship:
and the change in parenthesization is ok? I assume the `*` isn't really multiplication and so maybe there isn't standard order of operations?
@squeed it sounds like you understand this better anyway, so I'll leave it to you to lgtm

squeed:
The `*` is totally multiplication. It's a cute hack. If the node is up, then the value is 1, else 0. So that way we don't alert on down nodes.

I checked with some Prometheus developers, and, yes, we need to worry about order of operations. Very good point. This is a bug - @JacobTanenbaum, can you fix?

JacobTanenbaum:
@danwinship changed the parenthesis

danwinship:
/lgtm
? (the bot didn't seem to notice Casey's comment)

PR #391, opened 2019-11-08
Autodetect OVN octavia provider

luis5tb:
This makes the CNO to detect if ovn provider is configured in
octavia, and if that so configures kuryr to use it. Thanks to this
the need for a VM per svc is removed and loadbalancing is done by
using the ovn distributed loadbalancer (aka ovn flows)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -1077,6 +1081,31 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+		page, err := providers.List(lbClient).AllPages()
+		if err != nil {
+			log.Print("failed to get lbs provider list, using default octavia provider")
+		} else {
+			providers, err := providers.ExtractProviders(page)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to extract providers list")
+			}
+			for _, provider := range providers {
+				if provider.Name == OVNProvider {
```
dulek: Is this a standardized name or can admin name providers however he likes? I mean - is it like flavors?

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -624,8 +628,8 @@ func ensureOpenStackLbMonitor(client *gophercloud.ServiceClient, name, poolId st
 			PoolID:     poolId,
 			Type:       monitors.TypeTCP,
 			MaxRetries: 3,
-			Delay:      10,
-			Timeout:    10,
+			Delay:      5,
+			Timeout:    4,
```
dulek: Why those changes?

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/loadbalancers"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/monitors"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
+	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/providers"
```
dulek: I think you need to regenerate `dep` after that.

pkg/platform/openstack/kuryr_bootstrap.go
luis5tb: That is static, and seems it must be 'ovn' for the ovn-octavia driver

pkg/platform/openstack/kuryr_bootstrap.go
luis5tb: yep! I'll update now that gophercloud side got merged

PR #392, opened 2019-11-08
[wip] Test raft

dcbw:
so I can see the CI


dcbw:
Latest failure is due to a bug which is fixed by https://github.com/ovn-org/ovn-kubernetes/pull/908

PR #300, opened 2019-08-26
Adding status to configmap ca injector controller

JacobTanenbaum:
To help debug potential issues with the configmap ca injector add statusmanager and more clearly log some errors

PR #395, opened 2019-11-12
pkg/controller/configmap_ca_injector: "proxy not found" -> "ConfigMap ... not found"

wking:
Adjusting one of the log lines from 482ce00a0c (#300) to match another log line from that same commit.  There's nothing inherently proxy-ish about the trusted CA bundle ConfigMap, so we don't need a proxy-specific log line.  And giving the namespace/name is a more actionable error message (admins will know which resource they need to chase down).


wking:
[GCP failed to install, with the console stuck in SyncLoopRefreshProgressing][1].

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/395/pull-ci-openshift-cluster-network-operator-master-e2e-gcp/157

PR #398, opened 2019-11-14
Bug 1769015: Bump library go

abhat:


abhat:
I will uncomment the crdTypes in https://github.com/openshift/origin/pull/24058/files#diff-b212b5fb9c78264b938f2bc554a80376R140 via a separate PR on origin. 

@alexanderConstantinescu, @JacobTanenbaum - the egressnetworkpolicies, hostsubnets etc. are not CRDs right?

/cc @soltysh, @squeed PTAL.

bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -1,207 +1,326 @@
+---
```
abhat: looks like the generated CRDs are missing range validations and `additionalPrinterColumns` as well as comment text around the `validation` stanza. If we need to add those back, I can merge the changes from the old CRD yaml into this.

danwinship: We shouldn't have two copies; if there's a generated copy in library-go we should update the code to pull from that directly.

As for the differences:
- We can live without the comments
- The loss of validation on `ClusterNetwork` isn't that big of a deal since it's not user-generated anyway. (Also, if we end up merging any of the openshift-sdn IPv6 work, we'd have to drop the IP/CIDR validation anyway.)
- The loss of validation on the "egress" fields in `NetNamespace` and `HostSubnet` is a little more of a problem, since those are user-modifiable. Likewise all of `EgressNetworkPolicy`. If we lose validation there we need to make sure that the code in openshift-sdn validates the fields when it reads them in and won't crash if they're bad.
  - Alternately, maybe we should extend multus-admission-controller to be cno-admission-controller and validate SDN types too?
  - Alternately you may be able to add back some of the validation by adding `kubebuilder` annotations to the API objects in `openshift/api`.
- Likewise I think you can get the `additionalPrinterColumns` back via `kubebuilder` annotations and if so we should definitely do that because `oc get` becomes useless for these types without that.

abhat: Fair enough. I will try to add `kubebuilder` annotations/markers for `additionalPrinterColumns` as well as field validations where possible. For the other ones, I will merge the existing ones in to the generated CRDs. There are no two copies of it - just the one checked in under `bindata/network`. There is none under `openshift/api` repo. I was asking @alexanderConstantinescu how we generated the `001-crd.yaml` to begin with. Looking at the file history @squeed checked in that yaml, but looks like it was a manual process most likely.

squeed: Is it possible for this to just be a symlink in the `manifests` directory? That way we more easily grasp what is installed.

Yes, we should be adding kubebuilder annotations to `openshift/api` as much as possible.

abhat: @squeed Not sure I understood the first part of your comment. The `001-crd.yaml` doesn't exist at all in the manifests directory today. Do you want me to add a sym-link to this file in `manifests`? Or are you referring to the `0000_70_cluster-network-operator_01_crd.yaml` which is currently checked into `vendor/` as well as has a copy in the `manifests` dir?

squeed:
So, are we using the generated CRDs from library-go or not? Did we have to customize them?

abhat:
This is the interim manual fix for 4.3; Once the annotations changes are
merged in openshift/api we can pull in the vendored CRDs for 4.4.

squeed:
ack
/approve
@danwinship can you give this a quick look-over and lgtm?

Dockerfile
```diff
@@ -10,6 +10,7 @@ FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /tmp/build/cluster-network-operator /usr/bin/
 COPY --from=builder /tmp/build/cluster-network-renderer /usr/bin/
 COPY manifests /manifests
+COPY vendor/github.com/openshift/api/operator/v1/0000_70_cluster-network-operator_01_crd.yaml /manifests
```
danwinship: Diffing the old and new, it looks like we are losing the `required` tags (which should be set on `spec.defaultNetwork`, `spec.clusterNetwork[].cidr`, `spec.clusterNetwork[].hostPrefix`, `spec.defaultNetwork.type`, `spec.additionalNetworks[].type`, and `spec.additionalNetworks[].name`)

bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -64,46 +122,91 @@ spec:
...
-            pattern: '^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$'
-        egressCIDRs:
-          type: array
-          items:
-            type: string
-            pattern: '^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])/([0-9]|[12][0-9]|3[0-2])$'
+      required:
+      - host
+      - hostIP
+      - subnet
```
danwinship: And this has the opposite problem; `HostSubnet`'s `subnet` field should not be required

bindata/network/openshift-sdn/001-crd.yaml
```diff
@@ -165,43 +309,102 @@ spec:
...
   names:
     kind: EgressNetworkPolicy
-    singular: egressnetworkpolicy
+    listKind: EgressNetworkPolicyList
     plural: egressnetworkpolicies
+    singular: egressnetworkpolicy
+  scope: Namespaced
   validation:
-    # This should be mostly equivalent to ValidateEgressNetworkPolicy
+    # This should be mostly equivalent to ValidateEgressNetworkPolicy
```
danwinship: (gratuitous extra spaces at end of line)

Dockerfile
abhat: thanks @danwinship. I have addressed the review comments. Looks like we need to add more annotations for these required fields in https://github.com/openshift/api/pull/527

abhat:
@squeed/@danwinship can you lgtm please?

danwinship:
lgtm

PR #400, opened 2019-11-15
Make ovn-kubernetes work with IPv6

russellb:
These changes represent the cluster-network-operator portion of making ovn-kubernetes work with single-stack IPv6.

Some of the most invasive changes are with how IP addresses are handled for OVN databases.  Instead of using DNS, we only use IP addresses taken straight from Nodes.  In our AWS IPv6 environment, the node names only resolve to their IPv4 addresses.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -156,9 +156,14 @@ spec:
               - -c
               - |
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
+                OVN_NODE_IPS_ARRAY=({{.OVN_NODE_IPS}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
+                  USEIPV6=
+                  if [[ ${OVN_NODE_IPS_ARRAY[0]} == *":"* ]]; then
+                    USEIPV6=":[::]"
+                  fi
                   retries=0
-                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}}:[::] -- set connection . inactivity_probe=0; do
+                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}}${USEIPV6} -- set connection . inactivity_probe=0; do
                     (( retries += 1 ))
                   if [[ "${retries}" -gt 40 ]]; then
                     echo "too many failed ovn-nbctl attempts, giving up"
@@ -258,9 +263,14 @@ spec:
               - -c
               - |
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
+                OVN_NODE_IPS_ARRAY=({{.OVN_NODE_IPS}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
+                  USEIPV6=
+                  if [[ ${OVN_NODE_IPS_ARRAY[0]} == *":"* ]]; then
+                    USEIPV6=":[::]"
+                  fi
                   retries=0
-                  while ! ovn-sbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_SB_PORT}}:[::] -- set connection . inactivity_probe=0; do
+                  while ! ovn-sbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_SB_PORT}}${USEIPV6} -- set connection . inactivity_probe=0; do
                     (( retries += 1 ))
                   if [[ "${retries}" -gt 40 ]]; then
                     echo "too many failed ovn-sbctl attempts, giving up"
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -150,7 +150,7 @@ spec:
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
                   retries=0
-                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}} -- set connection . inactivity_probe=0; do
+                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}}:[::] -- set connection . inactivity_probe=0; do
                     (( retries += 1 ))
                   if [[ "${retries}" -gt 40 ]]; then
                     echo "too many failed ovn-nbctl attempts, giving up"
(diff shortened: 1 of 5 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -150,7 +158,7 @@ spec:
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
                   retries=0
-                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}} -- set connection . inactivity_probe=0; do
+                  while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}}:[::] -- set connection . inactivity_probe=0; do
```
russellb: This piece still needs to be fixed to be conditional on IPv6

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -244,7 +260,7 @@ spec:
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
                   retries=0
-                  while ! ovn-sbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_SB_PORT}} -- set connection . inactivity_probe=0; do
+                  while ! ovn-sbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_SB_PORT}}:[::] -- set connection . inactivity_probe=0; do
```
russellb: This needs to be fixed to be conditional on IPv6

bindata/network/ovn-kubernetes/ovnkube-master.yaml
russellb: fixed

bindata/network/ovn-kubernetes/ovnkube-master.yaml
russellb: fixed

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -293,6 +293,9 @@ spec:
           name: ovn-cert
         - mountPath: /ovn-ca
           name: ovn-ca
+        - mountPath: /lib/modules
```
danwinship: This commit (`Allow ovnkube-node to load kernel modules `) should not be needed any more; kubelet will always have created iptables rules before any pods run, so now that we've updated kubelet to realize it should be doing IPv6 iptables rules, ovn-kubernetes won't need to load the module itself.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -190,6 +190,11 @@ spec:
             sleep 5
           done
 
+          if [[ ${db_ip} == *":"* ]]; then
+            # IPv6
+            db_ip="[${db_ip}]"
```
danwinship: it appears that the `db_ip` variable isn't actually used any more (other than being checked whether it's the empty string), so we don't need this commit (`Add square brackets around OVN DB IPv6 addresses`) either

pkg/network/ovn_kubernetes.go
```diff
@@ -160,16 +174,42 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
...
+		if ip == "" {
+			return nil, fmt.Errorf("No InternalIP found on master node '%s'", masterNode.Name)
+		}
+		parsedIP := net.ParseIP(ip)
+		if parsedIP == nil {
+			return nil, fmt.Errorf("Failed to parse InternalIP '%s' found on master node '%s'", ip, masterNode.Name)
+		}
+		if parsedIP.To4() == nil {
+			// IPv6, wrap address in brackets
+			ip = fmt.Sprintf("[%s]", ip)
```
danwinship: It's a little weird to call them "IPs" when they're "IPs with brackets around them", although I guess the alternative would end up being a lot messier.

pkg/network/ovn_kubernetes.go
```diff
@@ -160,16 +174,42 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
...
 	}
 
-	sort.Strings(ovnMasterNodes)
+	sort.Sort(ovnMasters)
+	ovnMasterNodes := []string{}
+	ovnMasterNodeIPs := []string{}
+	for _, m := range ovnMasters {
+		ovnMasterNodes = append(ovnMasterNodes, m.Name)
+		ovnMasterNodeIPs = append(ovnMasterNodeIPs, m.IP)
+	}
```
danwinship: you could just change `OVNBootstrapResult` to have the array of structs so you don't need to do this

bindata/network/ovn-kubernetes/ovnkube-node.yaml
russellb: Thanks, this went through a few rebases.  I'll remove that commit.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
russellb: OK - we can remove this and find out pretty quickly if it works!

pkg/network/ovn_kubernetes.go
russellb: bracketedIP?  :-)

pkg/network/ovn_kubernetes.go
russellb: True - but the consumer of OVNBootstrapResult needs 2 arrays of strings, anyway, so something similar would be done there instead.

dcbw:
So one issue is this locks the IPs of the masters, and Casey was concerned about what happens if a master reboots and gets a different DHCP address. On some cloud providers the hostname follows the IP, but that's not necessarily the case. eg the hostname can be stable but the IP might not be, which is the advantage of DNS-ing the hostname in the daemonset bash script...

danwinship:
> So one issue is this locks the IPs of the masters

But the existing code already locks the *names* of the masters, which is also no good. So we're going to need a total rewrite at some point anyway.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -109,15 +109,23 @@ spec:
...
+          OVN_NODE_IPS_ARRAY=({{.OVN_NODE_IPS}})
+          MASTER_NODE_IP="${OVN_NODE_IPS_ARRAY[0]}"
+          NODE_INDEX=0
+          for i in "${!OVN_NODES_ARRAY[@]}"; do
+            if [[ "${OVN_NODES_ARRAY[i]}" == "${K8S_NODE}" ]]; then
+              NODE_INDEX=${i}
+              break
+            fi
+          done
+          THIS_NODE_IP="${OVN_NODE_IPS_ARRAY[${NODE_INDEX}]}"
```
squeed: You don't need to do this; you can pass in the PodIP via the downward API.

squeed: (see https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api )

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -148,9 +156,14 @@ spec:
               - -c
               - |
                 OVN_NODES_ARRAY=({{.OVN_NODES}})
+                OVN_NODE_IPS_ARRAY=({{.OVN_NODE_IPS}})
                 if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
+                  USEIPV6=
```
squeed: Oh hey, this is neat! Can you just tell it to bind on `0.0.0.0` for the v4 case?

pkg/network/ovn_kubernetes.go
squeed: Actually, if you use the downward API, I think you can get rid of `{{.OVN_NODES}}` entirely anyways.

russellb:
This PR is deprecated in favor of @danwinship 's updates in #417 and #424

PR #401, opened 2019-11-15
Bug 1779183: Don't mark the operator un-Available or Degraded because of multus-admission-controller

danwinship:
Changes the status reporting so that `multus-admission-controller` is treated as "non-critical", meaning that we will transition to `Available` if everything else has deployed even if `multus-admission-controller` hasn't (though in that case we'd be both `Available` and `Progressing`) and we will never go to `Degraded` just because `multus-admission-controller` took too long to deploy.

You could argue that we should *eventually* mark the operator `Degraded` if `multus-admission-controller` doesn't deploy, but if so, then we should improve the status tracking to only degrade if `multus-admission-controller` fails to deploy *after* the credentials operator is `Available`, since it is not expected to deploy before that.

@squeed I know we were talking about this somewhere so if there's a related issue/bugzilla bug/jira card, can you link it?

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-network-operator/pkg/names"
 
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/controller/statusmanager/pod_status.go
```diff
@@ -276,3 +286,7 @@ func (status *StatusManager) setLastPodState(
 		return status.client.Patch(context.TODO(), newStatus, client.MergeFrom(oldStatus))
 	})
 }
+
+func isNonCritical(obj metav1.Object) bool {
+	return obj.GetAnnotations()[names.NonCriticalAnnotation] == "true"
```
squeed: Since this condition is a boolean, the convention would be to set an empty string for true.

squeed:
Seems fine, one minor nit.

squeed:
(The card is https://jira.coreos.com/browse/SDN-668 )

danwinship:
> (The card is https://jira.coreos.com/browse/SDN-668 )

That's the card for marking hung rollouts, but I couldn't remember if there was a card / bug / issue about *not* doing it for multus-admission-controller.

danwinship:
(updated to use `""` for annotation)

squeed:
No, I couldn't find a bug for this - I'll open one.

danwinship:
/retest
and review ping

squeed:
/retest
/bugzilla refresh
(because this will need to be cherrypicked)

danwinship:
/retitle Bug 1779183: Don't mark the operator un-Available or Degraded because of multus-admission-controller
/retest

PR #405, opened 2019-11-26
Bug 1770223: Reverts PR #388

danehans:
This PR reverts the changes introduced in PR #388 by removing the external api server from the default `noProxy` generated list.

/cc @wking @mfojtik 
/assign @squeed

### DIFF
--- a/pkg/util/proxyconfig/no_proxy.go
+++ b/pkg/util/proxyconfig/no_proxy.go
@@ -46,16 +46,6 @@ func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructur
 		ic.Networking.MachineCIDR,
 	)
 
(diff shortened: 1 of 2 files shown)


wking:
The last touch in master was #328.  The 4.2 reversion in #388 doesn't seem to have ever had a master analog in this repo, and the bug links were [manually overridden][1].  I think the combination of those was the process gap that this issue fell through; good to get it fixed now :).

[1]: https://github.com/openshift/cluster-network-operator/pull/388#issuecomment-551156453

wking:
Also wanted to drop a link to https://github.com/openshift/installer/pull/2668/commits/6a102cf4b4e5ee8d755c62a60f28a0128abbc646 , which is the install-side revert reversion matching this PR.

wking:
/lgtm

But I'm not an approver.

PR #408, opened 2019-12-02
Bug 1764256: Handle kubeconfigs with an empty context.

squeed:
Fixes a nil panic for kubeconfigs with no context (1764256).


squeed:
This doesn't seem to happen often enough to matter.

PR #410, opened 2019-12-02
SDN-662: OVN raft followups

alexanderConstantinescu:
This PR adds a `PodDisruptionBudget` for maintaining concensus for the raft cluster + a `readinessProbe` for the `sbdb` and `nbdb` checking that all members have joined the cluster, and if they have: waits 30 seconds according to what has been mentioned here: https://jira.coreos.com/browse/SDN-662

/assign @squeed

### DIFF
--- a/bindata/network/ovn-kubernetes/007-pdb.yaml
+++ b/bindata/network/ovn-kubernetes/007-pdb.yaml
@@ -1,0 +1,10 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: ovn-raft-quorum-guard
+  namespace: openshift-ovn-kubernetes
+spec:
+  minAvailable: {{.OVN_MIN_AVAILABLE}} 
+  selector:
+    matchLabels:
+      name: ovnkube-master
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -188,6 +188,25 @@ spec:
           containerPort: {{.OVN_NB_PORT}}
         - name: nb-db-raft-port
           containerPort: {{.OVN_NB_RAFT_PORT}}
+        readinessProbe:
+          initalDelaySeconds: 30
+          exec:
+            command:
+            - /bin/bash
+            - -c
+            - |
+              set -e
+              OVN_NODES_ARRAY=({{.OVN_NODES}})
+              AVAILABLE_NODES=0 
+              for node in "${OVN_NODES_ARRAY[@]}"; do
+                node_ip=$(getent ahostsv4 "${node}" | grep RAW | awk '{print $1}')
+                if ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound | grep -q $node_ip; then
+                  (( AVAILABLE_NODES += 1 ))
+                fi
+              done
+              if [[ "${AVAILABLE_NODES}" -lt {{.OVN_MIN_AVAILABLE}} ]]; then
+                exit 1
+              fi
       
       # sbdb: The southbound, or flow DB. In raft mode 
       - name: sbdb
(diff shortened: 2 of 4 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -188,6 +188,21 @@ spec:
...
+        readinessProbe:
+          exec:
+            command:
+            - /bin/bash
+            - -c
+            - |
+              set -xe
+              OVN_NODES_ARRAY=({{.OVN_NODES}}) 
+              for node in "${OVN_NODES_ARRAY[@]}"; do
+                node_ip=$(getent ahostsv4 "${node}" | grep RAW | awk '{print $1}')
```
squeed: If I'm reading this correctly, it means that if any node goes down, all of the OVN databases will go not Ready. That strikes me as... not good. Maybe we should just make sure there is at least one other node?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -188,6 +188,21 @@ spec:
...
+                if ! ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound | grep -q $node_ip; then
+                  exit 1
+                fi
+              done
+              sleep 30
```
squeed: I don't think you need this sleep. If you want to tweak the timing of readiness checks, there are a few specific knobs, see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#probe-v1-core

bindata/network/ovn-kubernetes/007-pdb.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: ovn-raft-quorum-guard
+  namespace: openshift-ovn-kubernetes
+spec:
+  maxUnavailable: {{.OVN_MAX_UNAVAILABLE}}
```
squeed: Probably better to specify MinAvaiable instead.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: Indeed, my bad. I got confused with the lifecycle hook and thought it only runs once. Proposition: I'll change it to detecting if the amount of cluster members is higher than the MaxUnavailable, that should make it more precise.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: This was only added because the US specified "Either way, a readiness probe that waits 30 seconds so rollouts are slow.". I can remove it though. I am not convinced this adds a lot of benefits. Why do we need to be slow when rolling out? And aren't there possible problems with ovnkube-node rolling out before ovnkube-master and having to wait for 30 seconds?

bindata/network/ovn-kubernetes/007-pdb.yaml
alexanderConstantinescu: How come? Shouldn't change anything in terms of how K8s schedules things, or am I missing something?

alexanderConstantinescu:
@squeed: I added a livenessProbe. This one checks that the node itself is up and able to respond, as supposed to the readinessProbe which checks the general cluster member status.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -188,6 +188,43 @@ spec:
           containerPort: {{.OVN_NB_PORT}}
         - name: nb-db-raft-port
           containerPort: {{.OVN_NB_RAFT_PORT}}
+        livenessProbe:
```
squeed: No liveness probes for now - they are really hard to get right and usually break things.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -188,6 +188,25 @@ spec:
...
+            command:
+            - /bin/bash
+            - -c
+            - |
+              set -e
+              OVN_NODES_ARRAY=({{.OVN_NODES}})
+              AVAILABLE_NODES=0 
+              for node in "${OVN_NODES_ARRAY[@]}"; do
+                node_ip=$(getent ahostsv4 "${node}" | grep RAW | awk '{print $1}')
+                if ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound | grep -q $node_ip; then
```
squeed: I just tested this, and it turns out we don't ever remove servers if we're not available. So we need to parse the "connections" line. I'll give you a sample in a bit.

squeed: Yeah, all we need to do is check that the DB port is open. it turns out ovsdb takes care of this for us and only opens the port when consensus is achieved.

squeed:
Alexander is looking in to something else, he asked me to get this over the line.
Continued in #414

PR #414, opened 2019-12-04
ovn-kube: add pod disruption budget, readiness check

squeed:
Add a PodDisruptionBudget to protect the raft quorum.

Configure a readines probe for the DBs: ovsdb raft only opens its port once it has a raft consensus. Utilize that.

Also, add a TerminationMessagePolicy.

(Closes SDN-662)

/cc @alexanderConstantinescu

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -1,3 +1,19 @@
+# The ovnkube control-plane components
+
+# The pod disruption budget ensures that we keep a raft quorum
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: ovn-raft-quorum-guard
+  namespace: openshift-ovn-kubernetes
+spec:
+  minAvailable: {{.OVN_MIN_AVAILABLE}}
+  selector:
+    matchLabels:
+      name: ovnkube-master
+
+---
+
 kind: DaemonSet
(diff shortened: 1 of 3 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -253,6 +275,10 @@ spec:
                   sleep 2
                   done
                 fi
+        readinessProbe:
+          initialDelaySeconds: 30
```
alexanderConstantinescu: Question, I am not sure how readinessProbes impact hostNetwork pods. But will ovnkube-node be able to communicate with its depedencies in ovnkube-master during the first 30 seconds while ovnkube-master is not ready? Otherwise I suspect we might have a lot of errors logged in those pods (if they don't even CrashLoop?)

squeed: Very good question. In this case, that won't be an issue, because we're not using a Service / Endpoints. Basically, only Ready pods can show up in a Service. However, we talk directly to the pods, so that's not an issue here.

squeed:
Looks like the test failure is due to a pod not being created in time, but the logs indicate it only waits about 5 seconds. That seems... wrong. Looking further.

squeed:
Aha, seems to be a bit of a flake
/test e2e-aws-ovn

squeed:
Filed https://bugzilla.redhat.com/show_bug.cgi?id=1780143 for test flake.

PR #419, opened 2019-12-10
Use app label instead of name on ovnkube-master ds

rcarrillocruz:


danwinship:
/lgtm
/hold
until e2e-aws-ovn passes, feel free to cancel hold when it does

PR #420, opened 2019-12-10
Bug 1781707: openshift-sdn, CNO: handle new kubeconfig path

squeed:
We pull the apiserver url from the kubelet's kubeconfig. This has changed, and can be found in one of two places. Check them both.

Ultimately we need to generate this in a more sustainable way, but this is a good workaround (and easily backportable).

### DIFF
--- a/bindata/network/openshift-sdn/controller.yaml
+++ b/bindata/network/openshift-sdn/controller.yaml
@@ -27,11 +27,10 @@ spec:
             cpu: 10m
             memory: 50Mi
         env:
-        # TODO: bind-mount the kubelet's kubeconfig like the openshift-sdn process
-        # need to get the `oc` binary in to the hypershift image first
-        - name: KUBERNETES_SERVICE_PORT # allows the controller to communicate with apiserver directly on same host.
+        # point sdn-controller to the internal apiserver load balancer
+        - name: KUBERNETES_SERVICE_PORT
           value: "{{.KUBERNETES_SERVICE_PORT}}"
-        - name: KUBERNETES_SERVICE_HOST # allows the controller to communicate with apiserver directly on same host.
+        - name: KUBERNETES_SERVICE_HOST
           value: "{{.KUBERNETES_SERVICE_HOST}}"
         terminationMessagePolicy: FallbackToLogsOnError
       hostNetwork: true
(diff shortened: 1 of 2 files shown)


juanluisvaladas:
/retest
/lgtm as a long as the location of the apiserver's kubeconfig doesn't change very often.
In the long run we probably want to have a standard for every component managed by the cluster operator.

danwinship:
Have you discussed this with whatever team moved the file so they know we're depending on it?

danwinship:
also maybe a good excuse to switch sdn to using `KUBERNETES_SERVICE_HOST`/`KUBERNETES_SERVICE_PORT` overrides? We ought to be able to just set those variables and then stop passing `--url-only-kubeconfig`, right?

juanluisvaladas:
GCP seems to have an IAM problem

squeed:
/hold
to stop the retests

squeed:
> also maybe a good excuse to switch sdn to using `KUBERNETES_SERVICE_HOST`/`KUBERNETES_SERVICE_PORT` overrides? We ought to be able to just set those variables and then stop passing `--url-only-kubeconfig`, right?

I can switch the SDN to use this, but we still have a bootstrapping problem for the CNO that nobody knows how to solve. Since this is a customer-affecting bug, we need to work around it for now.

danwinship:
> I can switch the SDN to use this, but we still have a bootstrapping problem for the CNO

right, I realized we'd still need the fix for CNO, but we can at least only have the ugliness in one place rather than two

squeed:
OK, updated to contain the ugliness.

squeed:
CI currently stuck... nothing to do right now.

squeed:
Bah:
`open /var/lib/kubelet/kubeconfig: permission denied`

squeed:
weird, failed with no artifacts. either fluke or we really nuked the cluster.
/test e2e-gcp-upgrade

squeed:
OK, now I'm really confused: the CNO is privileged and yet it's still getting permission denied.

danwinship:
selinux

squeed:
> selinux

Yup, but privileged: true *should* be sufficient. Bah.

squeed:
Well, that's never going to pass. Looking at the manifests, the kubeconfig we expect should continue to exist on the masters. So, we can just get it from the CNO to fix the immediate in-the-wild bug.

So, undo the changes to the CNO, but still stop the SDN from reading kubeconfig.

squeed:
gcp took a nap
/retest

squeed:
filed SDN-740 to stop reading the kubeconfig.

squeed:
still more gcp oddities
/retest

squeed:
Let's pick this to 4.2 too.

juanluisvaladas:
Before merging? OK
/cherrypick release-4.2

PR #421, opened 2019-12-10
Read install-config and wait for the requested amount of control plane nodes

alexanderConstantinescu:
As seen by @pecameron, the CNO might proceed to bootstrapping OVN whilst some control plane nodes are unknown to him/her. 

This PR while make sure we wait until all control plane nodes that should exist, do exist. It does so by reading the `install-config.yaml` which is written as the `cluster-config-v1` configmap in `kube-system`, and proceeds to decoding the user defined amount of control plane nodes. It only proceeds with the OVN bootstrap once the kubeclient return the expected amount of nodes.  

I didn't have any better way of ascertaining this information then doing this, so if anyone has a better idea / knows how to retrieve this information more easily, please raise a hand. 

/assign @pecameron @danwinship

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1224,6 +1224,7 @@
     "k8s.io/apimachinery/pkg/util/runtime",
     "k8s.io/apimachinery/pkg/util/sets",
     "k8s.io/apimachinery/pkg/util/validation",
+    "k8s.io/apimachinery/pkg/util/wait",
     "k8s.io/apimachinery/pkg/util/yaml",
     "k8s.io/client-go/discovery",
     "k8s.io/client-go/informers",
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -7,22 +7,28 @@ import (
 	"path/filepath"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
+	"time"
 
-	"github.com/pkg/errors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
+	yaml "github.com/ghodss/yaml"
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"github.com/openshift/cluster-network-operator/pkg/render"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const OVN_NB_PORT = "9641"
 const OVN_SB_PORT = "9642"
 const OVN_NB_RAFT_PORT = "9643"
 const OVN_SB_RAFT_PORT = "9644"
+const CLUSTER_CONFIG_NAME = "cluster-config-v1"
+const CLUSTER_CONFIG_NAMESPACE = "kube-system"
 
 // renderOVNKubernetes returns the manifests for the ovn-kubernetes.
 // This creates
@@ -130,6 +136,7 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 }
 
 func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int) {
+
 	if conf.DefaultNetwork.OVNKubernetesConfig == nil {
 		conf.DefaultNetwork.OVNKubernetesConfig = &operv1.OVNKubernetesConfig{}
 	}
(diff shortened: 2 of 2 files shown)


squeed:
Hmm, I think I have another idea for how we can go about this

We just need to pick a leader once, for cluster bootstrapping. After that point, nodes that have joined the cluster ignore it. So let's just do that once and persist it. We should add a bootstrap persistence layer the same as we currently do for applied configuration. This is probably better than sorting the node names, since nodes can change.

That's also helpful because, eventually, we're going to want to be able to manage cluster membership.

We should also probably add the number of masters too so we can compute the PodDisruptionBudget correctly.

alexanderConstantinescu:
> Hmm, I think I have another idea for how we can go about this
> 
> We just need to pick a leader once, for cluster bootstrapping. After that point, nodes that have joined the cluster ignore it. So let's just do that once and persist it. We should add a bootstrap persistence layer the same as we currently do for applied configuration. This is probably better than sorting the node names, since nodes can change.
> 
> That's also helpful because, eventually, we're going to want to be able to manage cluster membership.
> 
> We should also probably add the number of masters too so we can compute the PodDisruptionBudget correctly.

Not sure I follow fully. I guess you are mentioning using a watcher of sorts, which would react to node changes? Should we do this here in the CNO? Any change we'd react to here would end up modifying the Daemonset template and would thus impact all running pods, not only the new one.

pecameron:
@squeed, @alexanderConstantinescu @dcbw Where do we stand with this. It is blocking SDN-456, SDN-494? It is a general flake that can impact any ovn cluster bring up.

danwinship:
> I didn't have any better way of ascertaining this information then doing this, so if anyone has a better idea / knows how to retrieve this information more easily, please raise a hand.

There *are* better objects for getting this information from but I don't know what they are. Try looking around in openshift/api ?

danwinship:
```
danw@p50:~> oc get machineconfigpools
NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT
master   rendered-master-a511557f13d8d6600f3c1659c8e80389   True      False      False      3              3                   3                     0
worker   rendered-worker-90b599b3f6e2ca5f180932e99c9e85a5   True      False      False      3              3                   3                     0
```
so I think you want to wait until the number of ovn-kube masters matches the `master` MachineConfigPool's `machineCount`

alexanderConstantinescu:
> ```
> danw@p50:~> oc get machineconfigpools
> NAME     CONFIG                                             UPDATED   UPDATING   DEGRADED   MACHINECOUNT   READYMACHINECOUNT   UPDATEDMACHINECOUNT   DEGRADEDMACHINECOUNT
> master   rendered-master-a511557f13d8d6600f3c1659c8e80389   True      False      False      3              3                   3                     0
> worker   rendered-worker-90b599b3f6e2ca5f180932e99c9e85a5   True      False      False      3              3                   3                     0
> ```
> 
> so I think you want to wait until the number of ovn-kube masters matches the `master` MachineConfigPool's `machineCount`

Thanks for the update, I had seen that already.  I was wondering though and could need some help regarding how to retrieve CRD information through client-go (I guess openshift/client-go)?

dcbw:
@alexanderConstantinescu is this PR obsoleted by #447?

alexanderConstantinescu:
> @alexanderConstantinescu is this PR obsoleted by #447?

Yes, but I kept this one opened in case any reviewer wanted to compare the different solutions and get a better understanding. But, if PR #447 works and get an approval, then it's definitely preferred to this. 

/hold

alexanderConstantinescu:
/hold cancel

Since #447 is closed due to "impossibility", this is re-opened. 

FYI: we cannot use the `machineconfigpool`, that CRD object is created _after_ networking. We thus have no other information except the configmap to ascertain that all control plane nodes are up when the CNO starts. 

The other option is: don't change anything. Worst thing that happens is that we re-boot ovnkube-master/node when a new control plane node pops up. CNO re-bootstraps and should do so when reconciling in `operconfig_controller.go`, which should happen - at the very least - every 5 min. If a new node is found, the ovnkube template would change during the bootstrapping phase, and thus launch a new rollout of ovnkube-master/node, with the correct cluster info. 

What do the reviewers think?

danwinship:
We do need to be able to handle the case of adding/changing a master in general, but I think it makes sense to optimize the install-time case to avoid having a gratuitous extra network restart

alexanderConstantinescu:
> of adding/changing a master in general,

I am pretty sure we do allow that, it's just that it takes a bit of time to reconcile, and, as mentioned, we re-deploy OVN when that happens (but there does not seem to be an option to avoid that, from my analysis).

/retest

danwinship:
Right, that's what I meant; it's ok to take a bit of time/redeploy *in the general case*, but we want to avoid doing that during the initial install. Which is what your patch implements.

(Aside: I know that raft clusters have to have an odd number of masters for technical reasons, but does ovn-kubernetes implement that as a strict requirement? ie, if one of the masters goes away and we try to redeploy the masters as a 2-node cluster, will OVN fail to start? And if it does start, what does that imply? Will the raft-ness fail somehow?)

alexanderConstantinescu:
> Right, that's what I meant; it's ok to take a bit of time/redeploy _in the general case_, but we want to avoid doing that during the initial install. Which is what your patch implements.
> 
> (Aside: I know that raft clusters have to have an odd number of masters for technical reasons, but does ovn-kubernetes implement that as a strict requirement? ie, if one of the masters goes away and we try to redeploy the masters as a 2-node cluster, will OVN fail to start? And if it does start, what does that imply? Will the raft-ness fail somehow?)

That's a good remark. I am not sure OVN will fail to start, but it won't be able to establish concensus for any leader election - I am not sure whatever practical implications this has. I'll try to do a test run of this and check. 

We protect ourselves from this during roll-outs, by using the `PodDisruptionBudget`. But it is true that if my test run shows that an OVN deployments - running less than the minimum required amount of nodes for concensus - does not work, that we should then enforce this during the deployment of OVN using the CNO.

alexanderConstantinescu:
> Right, that's what I meant; it's ok to take a bit of time/redeploy _in the general case_, but we want to avoid doing that during the initial install. Which is what your patch implements.
> 
> (Aside: I know that raft clusters have to have an odd number of masters for technical reasons, but does ovn-kubernetes implement that as a strict requirement? ie, if one of the masters goes away and we try to redeploy the masters as a 2-node cluster, will OVN fail to start? And if it does start, what does that imply? Will the raft-ness fail somehow?)

So the results are in: 

OVN Raft never fails! It's indestructible!

I have tried several scenarios: you can create an OVN raft cluster with any number of nodes you want, it never fails (both odd and even). You can delete any number of members you want, the cluster never indicated problems (I have actually opened an issue concerning this to OVN: https://bugzilla.redhat.com/show_bug.cgi?id=1798158). I have notified QE of a couple of test cases I thought they would need to run (https://issues.redhat.com/browse/SDN-551) and I will wait for them to collect the official results and assign bugs as we go. 

Having read through the documentation of OVN raft: https://github.com/openvswitch/ovs/blob/master/Documentation/ref/ovsdb.7.rst#clustered-database-service-model it seems that the cluster should go down once enough members have left (N-1) / 2. This is not the case today from what I saw, hence the bug above. 

Having said that: I am not sure if there's anything from what I've mentioned here that concerns the patch delivered through this PR.

PR #422, opened 2019-12-12
Allow changing LogLevel online in Kuryr

dulek:
This commit makes sure that Kuryr pods are restarted when LogLevel
option is changed in NetworkSpec. This is possible by injecting each
Kuryr pod with a hash representing set of all the options passed to the
templates. If the options change, hash will change and
controller-manager will rollout new DaemonSet/Deployment.


dulek:
Trying to wake this up.

/hold

PR #424, opened 2019-12-16
ipv6 stuff part 2

danwinship:
This is a reworking of the rest of #400 that wasn't in #417 

In addition to updating for some of the review comments there, I also reworked it to do more work in the CNO itself and less in the daemonset shell scripts.

@russellb

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -68,26 +68,11 @@ spec:
             set +o allexport
           fi
 
-          OVN_NODES_ARRAY=({{.OVN_NODES}})
-          
-          nb_addr_list=""
-          sb_addr_list=""
-          
-          for i in "${!OVN_NODES_ARRAY[@]}"; do
-            if [[ $i != 0 ]]; then
-              nb_addr_list="${nb_addr_list},"
-              sb_addr_list="${sb_addr_list},"
-            fi
-            host=$(getent ahostsv4 "${OVN_NODES_ARRAY[$i]}" | grep RAW | awk '{print $1}')
-            nb_addr_list="${nb_addr_list}ssl:${host}:{{.OVN_NB_PORT}}"
-            sb_addr_list="${sb_addr_list}ssl:${host}:{{.OVN_SB_PORT}}"
-          done
-          
           exec ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off \
             --pidfile=/var/run/openvswitch/ovn-northd.pid \
-            --ovnnb-db "${nb_addr_list}" \
-            --ovnsb-db "${sb_addr_list}" \
+            --ovnnb-db "{{.OVN_NB_DB_LIST}}" \
+            --ovnsb-db "{{.OVN_SB_DB_LIST}}" \
             -p /ovn-cert/tls.key \
             -c /ovn-cert/tls.crt \
             -C /ovn-ca/ca-bundle.crt 
@@ -126,10 +111,6 @@ spec:
             source /env/_master
             set +o allexport
           fi
-          
-          OVN_NODES_ARRAY=({{.OVN_NODES}})
-          MASTER_NODE=$(getent ahostsv4 "${OVN_NODES_ARRAY[0]}" | grep RAW | awk '{print $1}')
-          LOCALHOST=$(getent ahostsv4 "${K8S_NODE}" | grep RAW | awk '{print $1}')
 
           # Determine the ovn rundir.
           if [[ -f /usr/bin/ovn-appctl ]] ; then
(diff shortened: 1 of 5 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -164,10 +146,14 @@ spec:
               - /bin/bash
               - -c
               - |
-                OVN_NODES_ARRAY=({{.OVN_NODES}})
-                if [[ "$K8S_NODE" == "${OVN_NODES_ARRAY[0]}" ]]; then
+                MASTER_IP="{{.OVN_MASTER_IP}}"
+                if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
+                  USEIPV6=
```
russellb: A comment from the old review was to use "0.0.0.0" here, which I guess would be a little more explicit about v4 vs v6

russellb:
Nice work @danwinship , thanks for picking this up

danwinship:
reworked the ipv4/ipv6 listening stuff... I don't think it really makes sense to explicitly specify `:0.0.0.0` in the IPv4-only case, but I did change it to note that specifying `:[::]` actually makes it dual-stack, not IPv6. And added a unit test!

squeed:
> but I did change it to note that specifying :[::] actually makes it dual-stack, not IPv6

Can we just hard-code it to dual-stack then? And remove any of the conditionals.

pkg/network/ovn_kubernetes.go
```diff
@@ -161,17 +167,53 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 		return nil, fmt.Errorf("unable to bootstrap OVN, no master nodes found")
 	}
 
-	ovnMasterNodes := []string{}
-	for _, masterNode := range masterNodeList.Items {
-		ovnMasterNodes = append(ovnMasterNodes, masterNode.Name)
+	ovnMasterIPs := make([]string, len(masterNodeList.Items))
+	for i, masterNode := range masterNodeList.Items {
+		var ip string
+		for _, address := range masterNode.Status.Addresses {
```
squeed: I think this logic for Node -> IP is correct, but is there any chance that this won't match the `Pod.Status.HostIP`? That's what me match on to decide who is the raft-bootstrap node.

danwinship:
> Can we just hard-code it to dual-stack then?

No, I've been flipping pod-network servers to dual stack, but host network servers that are going to be single-stack ipv4 in 4.3.0 should remain single-stack ipv4 in 4.3.z (and maybe even beyond that), to avoid any possible security issues around changing their level of exposure to the external network.

pkg/network/ovn_kubernetes.go
danwinship: It matches kubelet's current logic, but there's no guarantee kubelet's logic won't change. I'm not sure we can do better without actually deploying the pods first and seeing what `hostIP` they get though.

squeed: This is somewhat beyond this PR, but we don't actually need to match on IP address for the pod to decide if its the master. We can match on NodeName. I guess we needed to do IP resolution because some of the OVN components don't resolve AAAA?

danwinship: We were matching on node name before. In Russell's original PR we were writing both node names and IPs into the template, and you pointed out that we could get rid of the names and do everything with IPs. But I guess then I got rid of most of the other reason for passing in the raw IPs so maybe I should switch `MASTER_IP` back to `MASTER_NODE`...

squeed: > you pointed out that we could get rid of the names and do everything with IPs

What can I say. I'm eventually consistent...

danwinship: > maybe I should switch `MASTER_IP` back to `MASTER_NODE`

except we'd still need `MASTER_IP` as well anyway. So let's just stick with this code for now.

> I guess we needed to do IP resolution because some of the OVN components don't resolve AAAA?

OVN/OVS only does DNS resolution if you link it with libunbound. (Not sure why... maybe it wants async DNS or something?) Anyway, the RHEL packages aren't built with that, so we have to pass IP addresses. And we had problems before (in master) with trying to resolve IPs address-family-independently in the shell scripts (getting v6 when we wanted v4 etc) so we have to get the IPs from the actual node objects.

squeed: Huh. Good to know.

squeed:
whoops, needs a `gofmt`.

squeed:
/hold
to stop the retests

danwinship:
@dcbw needs re-lgtm

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -150,10 +137,6 @@ spec:
             --pidfile /var/run/openvswitch/ovnkube-node.pid \
             --metrics-bind-address "0.0.0.0:9101"
         env:
-        - name: OVN_HYBRID_OVERLAY_ENABLE
-          value: "{{ .OVNHybridOverlayEnable }}"
-        - name: OVN_HYBRID_OVERLAY_NET_CIDR
-          value: "{{ .OVNHybridOverlayNetCIDR }}"
```
dcbw: @danwinship ?

dcbw: Ah, nevermind, it's templated in elsewhere.

PR #426, opened 2019-12-18
run-locally: rebuild the CNO

squeed:
This is a common mistake; let's just always re-build.

### DIFF
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -234,6 +234,10 @@ if [[ -z "$(which oc 2> /dev/null || exit 0)" ]]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


danwinship:
I think I have intentionally taken advantage of the fact that it doesn't do this before, but I think I've more frequently accidentally screwed up and run an old binary.
/lgtm

(I'd like this even better if you also fixed the bug where CNO/run-locally.sh suddenly crashes for no reason with an error about `OperatorPKI`, forcing you to restart it halfway through. Because now that restart is going to be slower...)

squeed:
/override ci/prow/e2e-gcp

this is just a dev script; not tested by CI

PR #427, opened 2019-12-19
osdn, ovnk: add publishNotReadyAddresses to metrics services

squeed:
Since these services are just used to configure Prometheus scrape targets, rather than load-balance traffic, we should not remove not-ready addresses from them.

### DIFF
--- a/bindata/network/openshift-sdn/monitor.yaml
+++ b/bindata/network/openshift-sdn/monitor.yaml
@@ -30,6 +30,7 @@ spec:
   selector:
     app: sdn
   clusterIP: None
+  publishNotReadyAddresses: true
   ports:
   - name: metrics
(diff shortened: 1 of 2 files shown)


JacobTanenbaum:
\lgtm

danwinship:
makes sense
/lgtm
did we send any prometheus patches to ovn-kubernetes upstream? if so we should fix this in the upstream daemonsets too

squeed:
>  did we send any prometheus patches to ovn-kubernetes upstream? if so we should fix this in the upstream daemonsets too

More or less, yeah. Good point, I'll update.

squeed:
fled https://github.com/ovn-org/ovn-kubernetes/pull/974

PR #435, opened 2019-12-28
ovn: use app label instead of name for master metrics and quorum guard

dcbw:
Finish what https://github.com/openshift/cluster-network-operator/pull/419 was trying to do. We need all the master things to use 'app' not 'name' for metrics scraping to work correctly.

@JacobTanenbaum @danwinship @juanluisvaladas @alexanderConstantinescu @rcarrillocruz @pecameron


vrutkovs:
This breaks OVN upgrades, since CVO uses `apply` and  label selectors are immutable

PR #437, opened 2020-01-06
Remove some unwanted logging. Fix an error message

danwinship:
#333 added logging of the ovn-kube TLS keys to the startup of all of the daemons. I assume @pecameron did this for debugging reasons at one point and then forgot to remove it? If the admin wants to see the key they can just look at the Secret directly... we should not be logging private keys to pod logs.

I'm assuming that for the other fix here (removing "PHIL" from an error message) we actually do want to keep that check.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -178,10 +178,6 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 		return nil, fmt.Errorf("unable to bootstrap OVN, no master nodes found")
 	}
 
-	if len(masterNodeList.Items) < 3 {
-		return nil, fmt.Errorf("PHIL Need at least 3 nodes, have %d", len(masterNodeList.Items))
-	}
-
 	ovnMasterNodes := []string{}
 	for _, masterNode := range masterNodeList.Items {
 		ovnMasterNodes = append(ovnMasterNodes, masterNode.Name)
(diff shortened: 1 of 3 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -174,12 +174,8 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 		return nil, err
 	}
 
-	if len(masterNodeList.Items) == 0 {
-		return nil, fmt.Errorf("unable to bootstrap OVN, no master nodes found")
-	}
-
 	if len(masterNodeList.Items) < 3 {
```
alexanderConstantinescu: Yeah, this should not have been merged. Could you remove this check altogether? Nothing says that we should ever have X master nodes.

alexanderConstantinescu: Edit: just saw that you did. Sorry about that

danwinship: yeah, I realized afterward, this is an attempt to deal with the problem #421 is trying to fix

alexanderConstantinescu:
@pecameron You seemed to have hard-coded the ovnkube-master/node log-level to 4, independently of what might be specified in any configfile/configmap. See L75 in `pkg/network/ovn_kubernetes.go`. 

```
data.Data["OvnkubeLogLevel"] = 4
```

@danwinship : could you change that in this PR? I am only asking as to have one atomic change taking care of all of those issues.

danwinship:
That's not really a change. Previously the daemonset specified

    env:
    - name: OVN_KUBE_LOG_LEVEL
      value: 4

now it specifies

    env:
    - name: OVN_KUBE_LOG_LEVEL
      value: "{{.OvnkubeLogLevel}}"

and templates in the `4`. But either way you can override it via the env-overrides configmap.

so... I guess there's not really any reason to be templating the default value in...

dcbw:
@trozet https://github.com/openshift/cluster-network-operator/pull/440

PR #438, opened 2020-01-07
Allow to change the cluster default network type

pliurh:
Allow users to change the cluster default network type by updating the 'networkType' field of the network.config.openshift.io CR. This not only support the use case of migrating the cluster network from Openshift SDN to OVN-Kubernetes, but also the use case vise Versa.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -171,14 +171,17 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	// Compare against previous applied configuration to see if this change
 	// is safe.
 	if prev != nil {
-		// We may need to fill defaults here -- sort of as a poor-man's
-		// upconversion scheme -- if we add additional fields to the config.
-		err = network.IsChangeSafe(prev, &operConfig.Spec)
-		if err != nil {
-			log.Printf("Not applying unsafe change: %v", err)
-			r.status.SetDegraded(statusmanager.OperatorConfig, "InvalidOperatorConfig",
-				fmt.Sprintf("Not applying unsafe configuration change: %v. Use 'oc edit network.operator.openshift.io cluster' to undo the change.", err))
-			return reconcile.Result{}, err
+		// Check if the operator is put in the 'Network Migration' mode.
+		if _, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok {
+			// We may need to fill defaults here -- sort of as a poor-man's
+			// upconversion scheme -- if we add additional fields to the config.
+			err = network.IsChangeSafe(prev, &operConfig.Spec)
+			if err != nil {
+				log.Printf("Not applying unsafe change: %v", err)
+				r.status.SetDegraded(statusmanager.OperatorConfig, "InvalidOperatorConfig",
+					fmt.Sprintf("Not applying unsafe configuration change: %v. Use 'oc edit network.operator.openshift.io cluster' to undo the change.", err))
+				return reconcile.Result{}, err
+			}
 		}
 	}
 
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -33,6 +33,10 @@ const IgnoreObjectErrorAnnotation = "networkoperator.openshift.io/ignore-errors"
 // that they are not critical to the functioning of the pod network
 const NonCriticalAnnotation = "networkoperator.openshift.io/non-critical"
 
+// NetworkMigrationAnnotation is an annotation on the networks.operator.openshift.io CR to indicate
+// that executing network migration (switching the default network type of the cluster) is allowed.
+const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migration"
+
 // SERVICE_CA_CONFIGMAP is the name of the ConfigMap that contains service CA bundle
 // that is used in multus admission controller deployment
 const SERVICE_CA_CONFIGMAP = "openshift-service-ca"


danwinship:
/hold
We can't allow this in general; it would completely break the cluster. We need to have some way of allowing it as part of the supported migration procedure, but blocking it otherwise.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -132,7 +132,14 @@ spec:
...
             exec:
-              command: ["/usr/share/openvswitch/scripts/ovs-ctl", "stop"]
+              command: 
+              - /bin/bash
+              - -c
+              - |
+                #!/bin/bash
+                /usr/share/openvswitch/scripts/ovs-ctl stop
+                rm -rf /run/openvswitch/*
+                rm -rf /var/lib/openvswitch/*
```
danwinship: We don't want to do this on ordinary termination; it would mean that every upgrade would tear down the network. (Likewise for the ovn-kube change.)

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -55,6 +55,10 @@ spec:
       containers:
       - name: sdn
         image: {{.SDNImage}}
+        lifecycle:
+          preStop:
+            exec:
+              command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf"]
```
danwinship: there is already a `preStop` handler at the end of the container spec that does this

bindata/network/ovn-kubernetes/monitor.yaml
```diff
@@ -76,9 +76,9 @@ spec:
   publishNotReadyAddresses: true
   ports:
   - name: metrics
-    port: 9101
+    port: 9103
```
danwinship: (this could be done as a separate PR...)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -90,6 +90,99 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 			objToDelete.SetName(currentObj.Name)
 			objToDelete.SetNamespace(currentObj.Namespace)
 			objToDelete.SetGroupVersionKind(gvk)
+			if objToDelete.GetKind() == "Namespace" {
+				continue
```
danwinship: Why was this added? We want to delete unused namespaces...
(Though... I guess probably we need to fix it to do that *last*.)

pkg/controller/statusmanager/status_manager.go
```diff
@@ -114,6 +207,8 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...confi
 
 		oldStatus := co.Status.DeepCopy()
 		status.deleteRelatedObjectsNotRendered(co)
+		status.deleteDaemonSetNotRendered(co)
+		status.deleteDeploymentNotRendered(co)
```
danwinship: why is this needed? The DaemonSets and Deployments should also be in RelatedObjects

pliurh:
> /hold
> We can't allow this in general; it would completely break the cluster. We need to have some way of allowing it as part of the supported migration procedure, but blocking it otherwise.

Agree. What's your suggestion on how to do it? How about adding an annotation like `maintenance: "migration"` in the `cluster` CR to indicate if this kind of operation can be allowed or not?

bindata/network/ovn-kubernetes/monitor.yaml
pliurh: Will do.

pkg/controller/statusmanager/status_manager.go
pliurh: Due to some reason, the 'namespace' deletion is always stuck at the "Terminating" state forever, even when all the related objects in that namespace have been deleted. And it will block the same Namespace creation if we what to perform a rollback from ovn-kubernetes to openshift-sdn. 

I agree that we need to remove the unused namespaces anyway. But I think maybe we can fix it later.

danwinship: > Due to some reason, the 'namespace' deletion is always stuck at the "Terminating" state forever, even when all the related objects in that namespace have been deleted.

Maybe there is something else in that Namespace that isn't in the related objects then?

danwinship:
An annotation sounds good. It should be more namespaced than that though. Maybe something like

    "networkoperator.openshift.io/network-migration": ""

(Standard annotation semantics are that if there's nothing to say about the annotation beyond the fact that it exists, then you just give it an empty-string value rather than `"true"` or something like that.)

You still need to figure out exactly what that entails though; you don't want CNO to start changing DaemonSets right away because that would interfere with the controlled node-by-node migration plan... do you stop the CNO during the migration? Is the rest of the code for dealing with migrations posted somewhere?

bindata/network/openshift-sdn/sdn-ovs.yaml
pliurh: fixed.

bindata/network/openshift-sdn/sdn.yaml
pliurh: fixed.

pkg/controller/statusmanager/status_manager.go
pliurh: The origin intention is to remove the deployment and daemonset before other objects. But I found it is unnecessary after more tests. It shall be removed.

pliurh:
> An annotation sounds good. It should be more namespaced than that though. Maybe something like
> 
> ```
> "networkoperator.openshift.io/network-migration": ""
> ```
> 
> (Standard annotation semantics are that if there's nothing to say about the annotation beyond the fact that it exists, then you just give it an empty-string value rather than `"true"` or something like that.)
> 
> You still need to figure out exactly what that entails though; you don't want CNO to start changing DaemonSets right away because that would interfere with the controlled node-by-node migration plan... do you stop the CNO during the migration? Is the rest of the code for dealing with migrations posted somewhere?

In 4.4, we plan to take the cluster reboot approach, which means all the nodes will be rebooted during the migration. Therefore, the whole cluster shall be considered as out of service during the migration. So all the pod will be connected to the new network backend after reboot. I gave a live demo at the Group C demo on Jan 8th, however, the demo record has not been added to the agenda doc yet. But here is a previous [demo record](https://drive.google.com/file/d/1AuyTzuJ0xjI3vhyvOQD72vwPtDxp2zUT/view), which shows the whole procedure of the 'cluster reboot' migration.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -31,6 +31,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	// The Network Migration Annotation which indicates if the operator is allowed to change the cluster network type.
+	MigrationAnno = "networkoperator.openshift.io/network-migration"
```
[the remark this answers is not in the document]
pliurh: fixed.

pkg/controller/statusmanager/status_manager.go
pliurh: Yes, there are some objects which are not managed by the CNO. I am not sure what is the proper way to delete all of them, then delete the namespace itself.

danwinship: hm. do you remember what objects?

pliurh: I did some further tests. It turns out that the namespace will be stuck in 'Terminating' before the cluster reboot, but can be deleted eventually after that. Therefore I remove the lines above.

danwinship:
Thanks. Can you squash the commits?

pliurh:
> Thanks. Can you squash the commits?

Done.

pliurh:
> So this is looking pretty simple now. We still need to figure out what's going wrong with deleting the namespace

Before rebooting, the cluster is in an unstable status. The stuck namespace has a finalizer 'kubernetes', which blocks the deletion. I think it should be fine, as long as, the namespace can be deleted successfully after the cluster back to normal.

PR #441, opened 2020-01-08
Adding run-remotely.sh script 

alexanderConstantinescu:
This PR introduces a `run-remotely.sh` script which allows us to run the CNO against a remotely running cluster. It's based off of `run-locally.sh` and takes only a kubeconfig argument. It strips off some of the logic surrounding manifests, install-config, etc that the `run-locally.sh` script has. 

I am not familiar with the reasons behind having the `PLUGIN_IMAGE` variable. But I kept it, in case someone was using it to modify the `env.sh` using that mechanism.

### DIFF
--- a/hack/run-locally.sh
+++ b/hack/run-locally.sh
@@ -13,6 +13,14 @@ if [[ -n "${HACK_MINIMIZE:-}" ]]; then
     echo ""
 fi
 
+function run_vs_existing_cluster {
+    echo "Attaching to already running cluster..."
+
+    wait_for_cluster
+    stop_deployed_operator
+    extract_environment_from_running_cluster
+}
+
 # Install our overrides so the cluster doesn't run the network operator.
 function override_install_manifests() {
     # Patch the CVO to not create the network operator
@@ -72,8 +80,6 @@ function stop_deployed_operator() {
 
 # wait_for_cluster waits for the cluster to come up and sets some variables
 function wait_for_cluster() {
-    export KUBECONFIG="${CLUSTER_DIR}/auth/kubeconfig"
-
     if [[ ! -e "${KUBECONFIG}" ]]; then
         echo "Waiting for installer to create a kubeconfig..."
         while [[ ! -e "${KUBECONFIG}" ]]; do
@@ -158,7 +164,8 @@ function create_cluster() {
 function print_usage() {
     >&2 echo "Usage: $0 [-c <cluster-dir>] [options]
 
-$0 detects whether the cluster given by the '-c' option does not yet exist or
+$0 runs against an existing cluster using the specific '-k' option or 
+detects whether the cluster given by the '-c' option does not yet exist or
 is already running. If it does not yet exist the cluster will be created using
 the options supplied below. If it is already running, any running
 cluster-network-operator will be stopped, plugin image overrides applied, and
@@ -166,7 +173,11 @@ a local cluster-network-operator started.
 
 The following options are accepted for both new and existing clusters:
 
- -c DIR           the cluster installation temporary directory (can also be given via CLUSTER_DIR); required
+ -c DIR           the cluster installation temporary directory (can also be given via CLUSTER_DIR); 
+ -k KUBECONFIG    the kubeconfig pointing to the existing cluster (can also be given via KUBECONFIG); 
+
+Either -c or -k (or their corresponding environment variable) is required
+
  -m IMAGE         custom network plugin container image name
 
 The following options are always accepted but only used for new clusters:
@@ -178,6 +189,7 @@ The following options are always accepted but only used for new clusters:
 
 The following environment variables are honored:
  - CLUSTER_DIR: the cluster installation temp directory
+ - KUBECONFIG: the kubeconfig pointing to the existing cluster 
  - INSTALLER_PATH: the path to the openshift-install binary for 'new' mode
  - INSTALL_CONFIG: the path to the install-config.yaml file for 'new' mode
 "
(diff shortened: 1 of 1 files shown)


danwinship:
> This PR introduces a `run-remotely.sh` script which allows us to run the CNO against a remotely running cluster.

That's what `run-locally.sh` is for... (The CNO binary is local, but the cluster is remote.) Rather than having a separate script, it would make more sense to just modify run-locally's "attach" mode to be able to work with a kubeconfig directly rather than only a cluster dir.

dcbw:
FWIW, PLUGIN_IMAGE is for if you have a custom ovnkube or SDN docker image that you want to run in the cluster to test changes. It's less useful now that we have infrastructure for running CI with ovnkube (which we didn't have at the time). But perhaps still useful.

alexanderConstantinescu:
/test e2e-gcp

flake, flake, flake....man, what is it snowing?!

hack/run-locally.sh
```diff
@@ -13,13 +13,21 @@ if [[ -n "${HACK_MINIMIZE:-}" ]]; then
...
+    wait_for_cluster
+    stop_deployed_operator
+    extract_environment_from_running_cluster
+}
+
 # Install our overrides so the cluster doesn't run the network operator.
 function override_install_manifests() {
     # Patch the CVO to not create the network operator
-    kubectl --kubeconfig=hack/null-kubeconfig patch --type=json --local=true -f="${CLUSTER_DIR}/manifests/cvo-overrides.yaml" -p "$(cat hack/overrides-patch.yaml)" -o yaml > "${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new"
+    oc --kubeconfig=hack/null-kubeconfig patch --type=json --local=true -f="${CLUSTER_DIR}/manifests/cvo-overrides.yaml" -p "$(cat hack/overrides-patch.yaml)" -o yaml > "${CLUSTER_DIR}/manifests/.cvo-overrides.yaml.new"
```
danwinship: Maybe split this stuff out into a separate commit

hack/run-locally.sh
```diff
@@ -158,15 +164,20 @@ function create_cluster() {
...
 cluster-network-operator will be stopped, plugin image overrides applied, and
 a local cluster-network-operator started.
 
 The following options are accepted for both new and existing clusters:
 
- -c DIR           the cluster installation temporary directory (can also be given via CLUSTER_DIR); required
+ -c DIR           the cluster installation temporary directory (can also be given via CLUSTER_DIR); 
+ -k KUBECONFIG    the kubeconfig pointing to the existing cluster (can also be given via KUBECONFIG); 
+
+-c or -k are required
```
danwinship: "Either -c or -k is required"
or really
"Either -c or -k (or their corresponding environment variable) is required"

hack/run-locally.sh
```diff
@@ -190,13 +202,17 @@ CLUSTER_DIR="${CLUSTER_DIR:-}"
...
+while getopts "c:f:i:m:k:n:w" opt; do
     case $opt in
-        c) CLUSTER_DIR="${OPTARG}";;
+        c) CLUSTER_DIR="${OPTARG}"
+	   mkdir -p "${CLUSTER_DIR}" >& /dev/null
+	   ;;
         f) INSTALL_CONFIG="${OPTARG}";;
         i) INSTALLER_PATH="${OPTARG}";;
         m) PLUGIN_IMAGE="${OPTARG}";;
+	k) KUBECONFIG="${OPTARG}";;
```
danwinship: change tabs to spaces everywhere

hack/run-locally.sh
```diff
@@ -239,7 +254,13 @@ echo "rebuilding the CNO"
 hack/build-go.sh
 
 # Autodetect the state of the cluster to determine which mode to run in
-if [[ -z "$(ls -A ${CLUSTER_DIR} 2> /dev/null | grep -v install-config.yaml | grep -v .openshift_install | grep -v env.sh)" ]]; then
+if [[ ! -z "${KUBECONFIG}" ]]; then
```
danwinship: I think if `CLUSTER_DIR` is set then we should use that by default. I often have a stray `KUBECONFIG` env var pointing to some no-longer-running cluster:

    danw@p50:~> echo $KUBECONFIG 
    /home/danw/Desktop/kubeconfig-1790652
    danw@p50:~> oc get nodes
    Unable to connect to the server: dial tcp: lookup api.ci-op-pdnwmn0m-endurance-a5b11.llc.devcluster.openshift.com on 10.11.5.19:53: no such host

hack/run-locally.sh
```diff
@@ -239,7 +254,13 @@ echo "rebuilding the CNO"
...
+    export CLUSTER_DIR=$PWD
```
danwinship: hm... what does this end up getting used for? Would it be better to use a temporary directory?

alexanderConstantinescu: This is used for the `${CLUSTER_DIR}/env.sh` which contains the image overrides. I could use a temporary directory. Would `/tmp/env.sh` be fine?

danwinship: Use `mktemp -d`.
Alternatively, write the environment to `${KUBECONFIG}.cno-env.sh` ?

alexanderConstantinescu: Used `mktemp -d`

alexanderConstantinescu: Acutally, I've been re-thinking this. Usually what we want is also to just create a `env.sh` and modify the image targeted, then re-deploy. 

I thus think it would be easier if we just use `CLUSTER_DIR=/tmp` if we only have a `$KUBECONFIG`. Then modify the `/tmp/env.sh` and re-deploy. I think it's easier as such.

alexanderConstantinescu:
@danwinship Could you have a final look at this?Thanks in advance!

PR #443, opened 2020-01-13
Change the metrics port for OVN-Kubernetes

pliurh:
Change the port number from 9101 to 9103 to avoid the port conflict during SDN migration.

Per https://github.com/openshift/cluster-network-operator/pull/438#discussion_r364287748


danwinship:
I think this is right and everything else just gets the metrics port from the Service definition rather than hardcoding it. Right, Jacob?
/approve
/assign @JacobTanenbaum 
/test e2e-aws-ovn

JacobTanenbaum:
@danwinship Yes, I believe that is how the metrics ports work now 
/lgtm

PR #444, opened 2020-01-13
Bug 1791277: Stop deploying kuryr-admission-controller if double listeners supported

luis5tb:
If octavia amphora provider supports double listeners, i.e.,
exposing the same port but on different protocol (e.g., udp and tcp)
then there is no need for deploying the kuryr-admission-controller
to enforce dns resolution over TCP, as we can create both udp and
tcp listeners on port 53 for the DNS service

### DIFF
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -5,19 +5,20 @@ import (
 )
 
 type KuryrBootstrapResult struct {
-	ServiceSubnet     string
-	PodSubnetpool     string
-	WorkerNodesRouter string
-	WorkerNodesSubnet string
-	PodSecurityGroups []string
-	ExternalNetwork   string
-	ClusterID         string
-	OctaviaProvider   string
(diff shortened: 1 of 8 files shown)


pkg/bootstrap/types.go
```diff
@@ -13,6 +13,7 @@ type KuryrBootstrapResult struct {
 	ExternalNetwork   string
 	ClusterID         string
 	OctaviaProvider   string
+	OctaviaDL         bool
```
dulek: Even if it's effectively double listeners at the moment, I'd rather use multiple listeners. And probably it's worth to be explicit, golang doesn't have limit on line length, so "OctaviaMultipleListeners" is fine.

luis5tb: Sounds good! I'll modify it

luis5tb:
holding it until kuryr-kubernetes adds the support for it

PR #445, opened 2020-01-15
Bug 1798191: Bump to ovn2.12

pecameron:
New paths for components

SDN-643

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -51,7 +51,7 @@ spec:
       # (container) -> (host)
       # /etc/openvswitch -> /var/lib/ovn/etc - ovsdb data
       # /var/lib/openvswitch -> /var/lib/ovn/data - ovsdb pki state
-      # /run/openvswitch -> tmpfs - sockets & pids
+      # /run/openvswitch -> tmpfs - sockets
       # /env -> configmap env-overrides - debug overrides
       containers:
       # ovn-northd: convert network objects in nbdb to flows in sbdb
@@ -70,7 +70,6 @@ spec:
 
           exec ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off \
-            --pidfile=/var/run/openvswitch/ovn-northd.pid \
             --ovnnb-db "{{.OVN_NB_DB_LIST}}" \
             --ovnsb-db "{{.OVN_SB_DB_LIST}}" \
             -p /ovn-cert/tls.key \
@@ -86,6 +85,8 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /run/openvswitch/
           name: run-openvswitch
+        - mountPath: /run/ovn/
+          name: run-ovn
         - mountPath: /env
           name: env-overrides
         - mountPath: /ovn-cert # not needed, but useful when exec'ing in to pod.
@@ -187,6 +188,8 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /run/openvswitch/
           name: run-openvswitch
+        - mountPath: /run/ovn/
+          name: run-ovn
         - mountPath: /env
           name: env-overrides
         - mountPath: /ovn-cert
@@ -294,6 +297,8 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /run/openvswitch/
           name: run-openvswitch
+        - mountPath: /run/ovn/
+          name: run-ovn
         - mountPath: /env
           name: env-overrides
         - mountPath: /ovn-cert
@@ -330,7 +335,7 @@ spec:
           fi
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/run/openvswitch/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/tmp/ovnk-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
             --db "{{.OVN_NB_DB_LIST}}")
@@ -349,7 +354,7 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["/bin/bash", "-c", "kill $(cat /run/openvswitch/ovnk-nbctl.pid)"]
+              command: ["/bin/bash", "-c", "kill $(cat /tmp/ovnk-nbctl.pid) && unset OVN_NB_DAEMON"]
         volumeMounts:
         - mountPath: /etc/openvswitch/
           name: etc-openvswitch
@@ -359,6 +364,8 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /run/openvswitch/
           name: run-openvswitch
+        - mountPath: /run/ovn/
+          name: run-ovn
         - mountPath: /run/ovnkube-config/
           name: ovnkube-config
         - mountPath: /env
@@ -395,6 +402,9 @@ spec:
           path: /var/lib/ovn/data
       - name: run-openvswitch
         emptyDir: {}
+      - name: run-ovn
+        hostPath:
+          path: /var/run/ovn
       - name: ovnkube-config
         configMap:
           name: ovnkube-config
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -35,7 +35,7 @@ spec:
       containers:
       # ovsdb and ovs-vswitchd
       - name: ovs-daemons
-        image: {{.OvnImage}}
+        image: "{{.OvnImage}}"
         command:
         - /bin/bash
         - -c
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -46,8 +46,16 @@ spec:
             source "/env/${K8S_NODE}"
             set +o allexport
           fi
+          # Determine the ovn rundir.
+          if [[ -f /usr/bin/ovn-appctl ]] ; then
+              # ovn-appctl is present. Use new ovn run dir path.
+              OVNCTL_DIR=ovn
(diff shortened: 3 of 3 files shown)


pecameron:
@alexanderConstantinescu @rcarrillocruz @dcbw PTAL
This only has effect when ovn-kube image has ovn2.12 
I need to get this in to get ovn-kubernetes PR72 to pass ci

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -70,7 +70,7 @@ spec:
 
           exec ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off \
-            --pidfile=/var/run/openvswitch/ovn-northd.pid \
+            --pidfile=ovn-northd.pid \
```
danwinship: According to the man pages (at least for 2.11) this will make it put the pidfile in `.`

However, do we actually need to be writing the pid files at all? It looks like the only pid file we ever read is the ovn-nbctl daemon's, and the others are "write-only" and we could just remove the `--pidfile` arguments...

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -330,7 +336,7 @@ spec:
           fi
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/run/openvswitch/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=ovnk-nbctl.pid \
```
danwinship: the container's `preStop` hook still does `kill $(cat /run/openvswitch/ovnk-nbctl.pid)`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -394,7 +402,11 @@ spec:
         hostPath:
           path: /var/lib/ovn/data
       - name: run-openvswitch
-        emptyDir: {}
+        hostPath:
+          path: /var/run/openvswitch
```
danwinship: This makes it so that the container's `/run/openvswitch` would be preserved between runs of different containers, which I don't think we want...

alexanderConstantinescu: So, if I am not mistaken (and I think I would like to "call a friend" on this one: @dcbw ). Those locations contain the OVN DB. Which we do not want to preserve between runs because:

1) I am pretty sure we can not upgrade and will crash when going from non-HA -> HA and vice versa (if we would ever, not sure we would though as non-HA is only in version 4.1/4.2 - tech-preview?)
2) Our source of truth is k8s and we sync directly with it on startup anyways, and thus re-write the DB. 

So, yes, I don't _think_ we want this. 

Why did you add this though @pecameron ? Is it required for 2.12?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: Changed. Dropped unneeded pid files.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: fixed

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: The unix socket needs to be everywhere it is referenced. This fixed the problem.

pecameron:
No nodes, no pods....
/test e2e-aws-ovn

pecameron:
"Error: ResourceLimitExceeded: Task limit exceeded"
/test e2e-aws-ovn

pecameron:
tcp 10.0.140.52:6443: connect: connection refused
/test e2e-aws-ovn

pecameron:
Cant't reach api server
/test e2e-aws-ovn

pecameron:
Ended up with 2 nodes (not 3).
/test e2e-aws-ovn

dcbw:
@pecameron @alexanderConstantinescu one option, if we don't want to persist the database but we do want to share the pid files and sockets, is to create our own /var/run/ovn-kubernetes and tell all our components to write their pid/sockets there with --pidfile and --unixctl. But leave the database path at the default and *don't* map that path onto the host.

dcbw:
@pecameron if you look at the ovnkube node logs, you'll see they all fail with "timed out waiting sbdb for node".  If you look at the code there, it's likely the path to ovn-controller's control socket is still wrong. either that or ovn-controller can't talk to the southbound, but since it does that via TCP I think that's less likely than a wrong control socket path.

pecameron:
@dcbw Thanks for looking at this, first run that got this far this week. Both ovnkube-node and ovn-controller complain about the SBDB

ovn-controller logs end with 
OVNSB commit failed, force recompute next time.
This should be over ssl, correct(?)
openshift-ovn-kubernetes_ovnkube-node-49sqn_ovn-controller.log

fatal msg="timed out waiting sbdb for node ip-10-0-143-132.us-east-2.compute.internal: exit status 1
Node should be using ssl to sdb (?)
e2e-aws-ovn/pods/openshift-ovn-kubernetes_ovnkube-node-fl86m_ovnkube-node.log

In pkg/util/ovs.go
runner.ovnRunDir = "/var/run/ovn/"

pkg/cluster/node.go
isOVNControllerReady
pid, err := ioutil.ReadFile(runDir + "ovn-controller.pid")  <<<< works
later
ctlFile := runDir + fmt.Sprintf("ovn-controller.%s.ctl", strings.TrimSuffix(string(pid), "\n")) <<< doesn't work
Or maybe ovn-controller can't talk to the SBDB?

Found a wrong version of openvswitch-selinux-extra-policy built a new image. Added some debug to ovnkube.

pecameron:
@dcbw Here is the version I think is ready to merge, PTAL
We'll see how the test goes. Expect to pass since ovnkube image is still 2.11 (PR72 has 2.12, so after this merges we can test that.)

pecameron:
Previous test that used the temp image failed to create nodes.

pecameron:
@dcbw Not sure what is wrong. Cluster came up with 6 nodes. Things look reasonable to me. PTAL

dcbw:
@pecameron I think your latest failures will probably be fixed by https://github.com/openshift/ovn-kubernetes/pull/83

dcbw:
@pecameron the duplicate IP PR merged for ovnkube, so you can re-test now.

pecameron:
@dcbw don't know what to make of these failures. The gcp tests are SDN. The ovn test does bring up ovn correctly. 
GCP
level=error msg="Error: Error applying IAM policy to project \"openshift-gce-devel-ci\": Too many conflicts.  Latest error: Error setting IAM policy for project \"openshift-gce-devel-ci\": googleapi: Error 409: There were concurrent policy changes. Please retry the whole read-modify-write with exponential backoff., aborted"

GCP upgrade
Jan 30 16:32:00.449: INFO: lookupDiskImageSources: gcloud error with [[]string{"instance-groups", "list-instances", "", "--format=get(instance)"}]; err:exit status 1
58
Jan 30 16:32:00.449: INFO:  > ERROR: (gcloud.compute.instance-groups.list-instances) could not parse resource []
59
What do you think?

dcbw:
@pecameron those look like install bootstrap flakes to me, or otherwise not networking related.

dcbw:
@pecameron but the AWS OVN failure looks perhaps legit, look for `Managed cluster should have no crashlooping pods in core namespaces over two minutes` in the build logs and one of the hits is at the end of the failure log for that specific test.

pecameron:
level=error msg="Cluster operator openshift-etcd Degraded is True with ConfigObservationDegradedError: ConfigObservationDegraded: error looking up self: lookup _etcd-server-ssl._tcp.ci-op-nd4331wx-583c6.origin-ci-int-aws.dev.rhcloud.com on 172.30.0.10:53: read udp 10.129.0.21:43654->172.30.0.10:53: i/o timeout\nConfigObservationDegraded: error looking up self: lookup _etcd-server-ssl._tcp.ci-op-nd4331wx-583c6.origin-ci-int-aws.dev.rhcloud.com on 172.30.0.10:53: read udp 10.129.0.21:55834->172.30.0.10:53: i/o timeout"

/retest
/test e2e-aws-ovn

pecameron:
Different set of failures this time. Trying again.
/retest
/test e2e-aws-ovn

pecameron:
e2e-aws-ovn
error: Couldn't find resource for "crd-publish-openapi-test-foo.example.com/v1,
"alertname\":\"KubeAPILatencyHigh\",
E kube-apiserver failed contacting the API: Get https://api.ci-op-h891bjwp-583c6.origin-ci-int-aws.dev.rhcloud.com:6443
The container could not be located when the pod was terminated
error deleting EBS volume "vol-0c95094660511ec2a" since volume is currently attached to "i-04c55ee4d769dbb2d"

pecameron:
@dcbw 445 passed tests. @alexanderConstantinescu @danwinship @rcarrillocruz Can we get this in? PTAL. I need it to test PR72 ovn2.12 image.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
pecameron: This actually passes tests. I will back out the change.

pecameron:
Back to a flaky CI....
release "release-latest" failed: the pod ci-op-kztttv4m/release-latest failed after 35s
/test e2e-aws-ovn

dcbw:
```
level=info msg="Credentials loaded from the \"default\" profile in file \"/etc/openshift-installer/.awscred\""
level=fatal msg="failed to fetch Terraform Variables: failed to fetch dependency of \"Terraform Variables\": failed to fetch dependency of \"Bootstrap Ignition Config\": failed to fetch dependency of \"Common Manifests\": failed to generate asset \"DNS Config\": getting public zone for \"origin-ci-int-aws.dev.rhcloud.com\": listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400, request id: 8de95699-cd1b-4fb2-b67b-d297e4222f61"
2020/02/04 20:51:09 Container setup in pod e2e-aws-ovn failed, exit code 1, reason Error
```
/test e2e-aws-ovn

PR #447, opened 2020-01-16
Dynamic HA 

alexanderConstantinescu:
This PR introduces the notion of "dynamic HA" (concretely meaning: the control plane nodes that ovnkube is deployed on, can both appear and disappear without affecting the deployment). 

**Today**: we have a deployment of ovnkube which - once the amount of control plane nodes change -  re-deploys ovnkube from scratch across all nodes. Issues have been seen by @pecameron related to this which caused the filing of https://github.com/openshift/cluster-network-operator/pull/421. That PR waits for the amount of control plane nodes to match the amount of nodes specified by the user during the cluster creation. If that condition does not apply we do not proceed with the deployment however, which is not optimal. 

This PR introduces a utilization of the `ovnkube-db` service DNS to generically target all ovnkube pods through CoreDNS. The benefit of this is will be: no new re-deployment of ovnkube when a new node pops up or disappears. The daemonset will deploy ovnkube-master on the new node and all pods in the cluster will continue to function as nothing happened since they're communicating with the svc DNS and not a specific list of IPs.

This is the chain of execution to make this work:

1) cluster boots and the CNO populates the `OVN_CONTROL_PLANE_IPS` in the ovnkube-master/node templates as to populate /etc/hosts with the initial cluster members and allow for initial communication to happen
2) pods boot and we change the `nsswitch.conf` to specify that we first want to perform DNS resolution on dns first and files (/etc/hosts) second.
3) we launch a daemon process which will watch when CoreDNS is up, and once it is, change /etc/resolv.conf to point to that instead of the host's nameserver. 

This PR needs a corresponding PR in ovn-kubernetes downstreams, that one adds `--no-leader-only` to "non-daemon-mode" triggered `ovn-nbctl` and `ovn-sbctl` commands. 

https://github.com/openshift/ovn-kubernetes/pull/78

/assign @danwinship @dcbw

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -31,8 +31,6 @@ spec:
   strategy:
     type: RollingUpdate
     rollingUpdate:
-      # by default, Deployments spin up the new pod before terminating the old one
-      # but we don't want that - because ovsdb holds the lock.
       maxSurge: 0
       maxUnavailable: 1
   template:
@@ -47,6 +45,12 @@ spec:
       serviceAccountName: ovn-kubernetes-controller
       hostNetwork: true
       priorityClassName: "system-cluster-critical"
+      hostAliases:
+      {{range .OVN_INITIAL_CONTROL_PLANE_IPS}}
+      - ip: "{{ . }}"
+        hostnames:
+        - "{{$.OVN_DB_SVC}}"
+      {{end}}
       # volumes in all containers:
       # (container) -> (host)
       # /etc/openvswitch -> /var/lib/ovn/etc - ovsdb data
@@ -62,11 +66,16 @@ spec:
         - -c
         - |
           set -xe
+
+          source /root/ovnkube-helper.sh
+
           if [[ -f /env/_master ]]; then
             set -o allexport
             source /env/_master
             set +o allexport
           fi
+
+          watch_core_dns_ep &
 
           exec ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off \
@@ -75,8 +84,12 @@ spec:
             --ovnsb-db "{{.OVN_SB_DB_LIST}}" \
             -p /ovn-cert/tls.key \
             -c /ovn-cert/tls.crt \
-            -C /ovn-ca/ca-bundle.crt 
+            -C /ovn-ca/ca-bundle.crt
         env:
+        - name: KUBERNETES_SERVICE_PORT
+          value: "{{.KUBERNETES_SERVICE_PORT}}"
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{.KUBERNETES_SERVICE_HOST}}"
         - name: OVN_LOG_LEVEL
           value: info 
         volumeMounts:
@@ -97,8 +110,6 @@ spec:
             cpu: 100m
             memory: 300Mi
         terminationMessagePolicy: FallbackToLogsOnError
-
-      # nbdb: the northbound, or logical network object DB. In raft mode 
       - name: nbdb
         image: "{{.OvnImage}}"
         command:
@@ -315,6 +326,9 @@ spec:
         - -c
         - |
           set -xe
+          
+          source /root/ovnkube-helper.sh
+          
           if [[ -f "/env/_master" ]]; then
             set -o allexport
             source "/env/_master"
@@ -329,10 +343,13 @@ spec:
             fi
           fi
 
+          watch_core_dns_ep &
+
           # start nbctl daemon for caching
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/run/openvswitch/ovnk-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
+            --no-leader-only \
             --db "{{.OVN_NB_DB_LIST}}")
 
           exec /usr/bin/ovnkube \
@@ -345,7 +362,7 @@ spec:
             --sb-address "{{.OVN_SB_ADDR_LIST}}" \
             --sb-client-privkey /ovn-cert/tls.key \
             --sb-client-cert /ovn-cert/tls.crt \
-            --sb-client-cacert /ovn-ca/ca-bundle.crt
+            --sb-client-cacert /ovn-ca/ca-bundle.crt 
         lifecycle:
           preStop:
             exec:
@@ -378,6 +395,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: KUBERNETES_SERVICE_PORT
+          value: "{{.KUBERNETES_SERVICE_PORT}}"
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{.KUBERNETES_SERVICE_HOST}}"
         ports:
         - name: metrics-port
           containerPort: 9102
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -20,6 +20,7 @@ import (
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
+const OVN_DB_SVC = "ovnkube-db.openshift-ovn-kubernetes.svc.cluster.local"
 const OVN_NB_PORT = "9641"
 const OVN_SB_PORT = "9642"
 const OVN_NB_RAFT_PORT = "9643"
@@ -50,13 +51,15 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["CNIBinDir"] = CNIBinDir
 	data.Data["OVN_NB_PORT"] = OVN_NB_PORT
 	data.Data["OVN_SB_PORT"] = OVN_SB_PORT
+	data.Data["OVN_DB_SVC"] = OVN_DB_SVC
 	data.Data["OVN_NB_RAFT_PORT"] = OVN_NB_RAFT_PORT
 	data.Data["OVN_SB_RAFT_PORT"] = OVN_SB_RAFT_PORT
-	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
-	data.Data["OVN_NB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
+	data.Data["OVN_NB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
+	data.Data["OVN_NB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
 	data.Data["OVN_MASTER_IP"] = bootstrapResult.OVN.MasterIPs[0]
+	data.Data["OVN_INITIAL_CONTROL_PLANE_IPS"] = bootstrapResult.OVN.MasterIPs
 	data.Data["OVN_MIN_AVAILABLE"] = len(bootstrapResult.OVN.MasterIPs)/2 + 1
 	data.Data["LISTEN_DUAL_STACK"] = listenDualStack(bootstrapResult.OVN.MasterIPs[0])
 
@@ -204,22 +207,6 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 		},
 	}
 	return &res, nil
-}
-
-func dbList(masterIPs []string, port string) string {
-	addrs := make([]string, len(masterIPs))
-	for i, ip := range masterIPs {
-		addrs[i] = "ssl:" + net.JoinHostPort(ip, port)
-	}
-	return strings.Join(addrs, ",")
-}
-
-func addrList(masterIPs []string, port string) string {
-	addrs := make([]string, len(masterIPs))
-	for i, ip := range masterIPs {
-		addrs[i] = "ssl://" + net.JoinHostPort(ip, port)
-	}
-	return strings.Join(addrs, ",")
 }
 
 func listenDualStack(masterIP string) string {
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -26,6 +26,12 @@ spec:
       hostNetwork: true
       hostPID: true
       priorityClassName: "system-node-critical"
+      hostAliases:
+      {{range .OVN_INITIAL_CONTROL_PLANE_IPS}}
+      - ip: "{{ . }}"
+        hostnames:
+        - "{{$.OVN_DB_SVC}}"
+      {{end}}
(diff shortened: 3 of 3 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -50,13 +51,15 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
-	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
-	data.Data["OVN_NB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
+	data.Data["OVN_NB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
+	data.Data["OVN_NB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
 	data.Data["OVN_MASTER_IP"] = bootstrapResult.OVN.MasterIPs[0]
+	data.Data["OVN_CONTROL_PLANE_IPS"] = bootstrapResult.OVN.MasterIPs
```
danwinship: maybe `OVN_INITIAL_CONTROL_PLANE_IPS`?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -67,6 +71,32 @@ spec:
             source /env/_master
             set +o allexport
           fi
+          
+          # Perform DNS resolution against the DNS resolver first and /etc/hosts second
+          sed -i 's/hosts:      files dns myhostname/hosts:      dns files myhostname/g' /etc/nsswitch.conf
```
danwinship: is this change needed before the `resolv.conf` switch? (ie, could this be done at the end of `watch_core_dns_ep`?)

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -67,6 +71,32 @@ spec:
...
+
+          function watch_core_dns_ep () {
```
danwinship: maybe this could be a helper script installed by ovn-kubernetes rather than being duplicated 3 times...

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -329,10 +364,34 @@ spec:
...
+            rm resolv.tmp
+          }
+
+          watch_core_dns_ep &
+
           # start nbctl daemon for caching
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/run/openvswitch/ovnk-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
+            --no-leader-only \
```
danwinship: The man page makes it sound like this is a bad idea... ("means that for read-only transactions it can report and act on stale data")

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -378,6 +437,14 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: OVN_HYBRID_OVERLAY_ENABLE
+          value: "{{ .OVNHybridOverlayEnable }}"
+        - name: OVN_HYBRID_OVERLAY_NET_CIDR
+          value: "{{ .OVNHybridOverlayNetCIDR }}"
```
danwinship: I think you accidentally brought these back in a bad git rebase. They aren't actually used

pkg/network/ovn_kubernetes.go
```diff
@@ -50,13 +51,15 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 	data.Data["OVN_NB_RAFT_PORT"] = OVN_NB_RAFT_PORT
 	data.Data["OVN_SB_RAFT_PORT"] = OVN_SB_RAFT_PORT
-	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
-	data.Data["OVN_NB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
-	data.Data["OVN_SB_ADDR_LIST"] = addrList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
+	data.Data["OVN_NB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_DB_LIST"] = "ssl:" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
+	data.Data["OVN_NB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_NB_PORT)
+	data.Data["OVN_SB_ADDR_LIST"] = "ssl://" + net.JoinHostPort(OVN_DB_SVC, OVN_SB_PORT)
```
danwinship: Not sure it really makes sense to template this stuff in if it is effectively hardcoded now

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -67,6 +71,32 @@ spec:
...
+          function watch_core_dns_ep () {
+            set +ex
+
+            while true; do
+              is_dns_up=$(kubectl get ep dns-default -n openshift-dns -o jsonpath='{.subsets[0].addresses[0].ip}' 2>/dev/null)
+              if [[ -n "${is_dns_up}" ]]; then
+                break
+              fi
+              echo "[$(date +%Y-%m-%dT%H:%M:%SZ)][watch_core_dns_ep] waiting for CoreDNS to start..."
+              sleep 5
```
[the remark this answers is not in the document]
alexanderConstantinescu: Increase the sleep? Or remove altogether?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -67,6 +71,32 @@ spec:
...
+            done
+
+            echo "[$(date +%Y-%m-%dT%H:%M:%SZ)][watch_core_dns_ep] CoreDNS is up, replacing inherited nameserver entry from host with CoreDNS service IP in /etc/resolv.conf"
+            core_dns_svc_ip=$(kubectl get svc dns-default -n openshift-dns -o jsonpath='{.spec.clusterIP}' 2>/dev/null)
+
+            cp /etc/resolv.conf resolv.tmp
+            sed -i "s/nameserver.*/nameserver $core_dns_svc_ip/g" resolv.tmp
+            cp -f resolv.tmp /etc/resolv.conf
```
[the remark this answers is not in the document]
alexanderConstantinescu: Actually that was my first try too: but I get this 
```
sh-4.2# sed -i "s/nameserver.*/nameserver 172.30.0.10/g" /etc/resolv.conf 
sed: cannot rename /etc/sedUHnoHS: Device or resource busy
```
Any idea?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: Indeed, that will speed things up during the first phase

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: Well, unfortunately that is a hard constraint for integrating all of this. Without `--no-leader-only` the `ovn-nbctl` commands risk talking to a follower and fail. Maybe we could have a discussion during the OVN weekly next week concerning this?

pkg/network/ovn_kubernetes.go
alexanderConstantinescu: I prefer hardcoding it in go rather than in bash, given the ipv4/ipv6 difference for host-port.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: Indeed my first thought as well, I am guilty of laziness and didn't go the extra mile to implement helper scripts which are packaged in our images. Will do however

bindata/network/ovn-kubernetes/ovnkube-master.yaml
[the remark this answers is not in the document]
danwinship: keeping the sleep 5 but only logging once a minute would be best, though complicated...

bindata/network/ovn-kubernetes/ovnkube-master.yaml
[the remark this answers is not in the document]
danwinship: huh. well, anyway, you can at least replace the first two commands with

    sed -e "s/nameserver.*/nameserver 172.30.0.10/g" /etc/resolv.conf > resolv.tmp

pkg/network/ovn_kubernetes.go
danwinship: there's no ipv4/ipv6 difference any more though because you're using a hostname instead of an IP

bindata/network/ovn-kubernetes/ovnkube-master.yaml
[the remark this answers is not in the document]
alexanderConstantinescu: Fair enough

bindata/network/ovn-kubernetes/ovnkube-master.yaml
danwinship: OK, so based on my reading the "Understanding Cluster Consistency" section of `ovsdb(7)`... it seems like as long as we're consistently talking to the same follower then we'll be fine; we might be slightly behind the leader at any point in time, but we'll eventually catch up. It would only be a problem if we talked to two different masters back-to-back when they were out of sync. (eg, if we create an object on master A and then query master B about its attributes). And it seems like `ovn-nbctl` daemon mode ought to mostly prevent that from happening, as long as the masters are stable. So maybe we're ok?

pkg/network/ovn_kubernetes.go
alexanderConstantinescu: OK, I got confused with the brackets around ipv6 addresses and thought it applied to ports too somehow. Honestly though I prefer keeping it templated here in case we ever change the service name. Would be less places in the code to change.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
danwinship: Hm... well, actually, etcd will not be synchronized with the ovn databases... So if there are situations where we assume that you can do:
- node 1 writes X=Y to ovn
- node 1 writes A=B to etcd
- node 2 sees that A=B has been written to etcd
- node 2 reads X from ovn and gets Y

then that might break now; node 2 might get the previous value of X instead

pecameron:
@alexanderConstantinescu are this PR and openshift/ovn-kubernetes#78 independent, they can go in any order and ci works on each independently of the other? We need to get ci to pass on each to get it to merge.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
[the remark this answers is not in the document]
alexanderConstantinescu: Done here: https://github.com/openshift/ovn-kubernetes/pull/78

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: Crap, you're right. I need to think about this but I wonder if we avoid the problem if ovnkube-master daemon only talks to the db deployed on his node. There will (hrmmm, should...) only be one ovnkube-master leader at any given point in time. ovnkube-node performs transactions using ovn-nbctl, but I'm not mistaken, those are only node specific and do not involve any transaction caused by changing k8s state.

alexanderConstantinescu:
> @alexanderConstantinescu are this PR and [openshift/ovn-kubernetes#78](https://github.com/openshift/ovn-kubernetes/pull/78) independent, they can go in any order and ci works on each independently of the other? We need to get ci to pass on each to get it to merge.

They are not independent. This depends on [openshift/ovn-kubernetes#78](https://github.com/openshift/ovn-kubernetes/pull/78), but that does not depend on this.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
alexanderConstantinescu: So, I think we should be alright if we remove `--db` when launching the daemon. `ovnkube-master` will thus only execute commands against the db member running on the same node (whether it's the leader or not). Master is the only one acting on k8s state. Well, we have multiple `AddServiceHandler` / `AddEndpointsHandler` in the code, but they all do different things and _should_ not overlap and act on the same OVN object - @dcbw, I guess your input is needed on this.

I will update the PR with this for now.

alexanderConstantinescu:
/hold

There seems to be something strange going on with OVN when ovnkube-master tries to perform the following type of transaction to a DB member which is not raft leader:

```
/usr/bin/ovn-nbctl --timeout=15 --wait=sb -- --may-exist lsp-add ip-10-0-131-121.ec2.internal openshift-marketplace_certified-operators-799765b8-mkjh7 -- lsp-set-addresses openshift-marketplace_certified-operators-799765b8-mkjh7 dynamic -- set logical_switch_port openshift-marketplace_certified-operators-799765b8-mkjh7 external-ids:namespace=openshift-marketplace external-ids:logical_switch=ip-10-0-131-121.ec2.internal external-ids:pod=true
```

The ovn daemon completely blocks and all transactions after that timeout. 

This does not happen when the transaction is sent to the raft leader, so obviously this PR exposes this flaw.

alexanderConstantinescu:
Well, this seems to have been a good idea but remains exactly just that: _an idea_

We cannot do this. The reason being, OVN Northbound and OVN Southbound run as two separate raft clusters.  `ovn-northd` needs to communicate with both raft master at all times, it cannot communicate with a follower and will not be considered up until that happens. If we do what I have done in this PR (meaning we don't provide `ovn-northd` with the complete set of cluster members and instead just a DNS) then `ovn-northd` will not be able to circle around and try all members in the cluster - as it usually does until it has successfully connected to both raft leaders. The DNS response from CoreDNS is always resolved to the same IP, it never circles the response to a secondary address as long as the first one is valid. `ovn-northd` will thus continuously just try to connect to the same server over and over again, and if that server is not the raft leader of both the southbound **and** northbound, then the cluster goes all south (that's a word play on OVN terminology, funny huh...). That is the reason the command above does not work, because `--wait=sb` forces the transaction to be confirmed as persisted in the southbound DB, which it is never as northd is not up. 

The only possible solution that could solve this is if CoreDNS would rotate its answer. I might reach out to the edge team for a question on that.

alexanderConstantinescu:
> If this works, I like it

This does not work. Having dug deep into the trenches of ovs and libunbound I am now certain we should not be doing this. For the interested reader, please see below as to why. 

Problem one:

ovs only seems to use the first IP in the DNS response from libunbound, see: https://github.com/openvswitch/ovs/blob/master/lib/dns-resolve.c#L290

Problem two:

`/etc/nsswitch.conf` has no effect on this. It's only used by libc and `gethostbyname`, which ovs does not use, as it uses libunbound instead. ovs defines two contexts when initializing the dns resolver object from libunbound, one context using `/etc/resolv.conf` and the other using `/etc/hosts`. There is no documentation in libunbound as to which takes precedence, but my tests show that it's `/etc/hosts`. This means that even if we configure CoreDNS to loadbalance the result (which I've done), northd will still only try to connect to the first IP returned from `/etc/hosts`. 

Problem three: 

Usually when creating a cluster the cluster "creator" will be elected leader in the first cycle. However if this turns out to not be the case, then northd will try to connect to the first entry in `/etc/hosts` which is not cluster leader and fail repeatedly, thus blocking cluster deployment and never reaching the phase where CoreDNS is up. 

On top of this, we also have the distributed problem listed above in the opened discussion. 

I am closing this PR.

PR #448, opened 2020-01-16
Support CA certificate provided by user

dulek:
In order to allow users to use self-signed certificates when
authenticating into OpenStack cloud, we need to make sure CA certificate
provided to the installer by the user is actually used by CNO and Kuryr.

In case of CNO this commit makes sure to configure gophercloud's HTTP
client to use the user CA when provided.

In case of Kuryr the CA certificate is injected into kuryr-controller's
/etc/ssl/certs.


MaysaMacedo:
Looks good. Just a question, should we mount the configmap even if the cert is not provided?

dulek:
> Looks good. Just a question, should we mount the configmap even if the cert is not provided?

Yup, I was actually forced to make it not mount it in that case, so this is handled now.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -850,8 +867,30 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
 		return nil, errors.Wrap(err, "failed to authenticate to OpenStack")
 	}
+	provider, err := openstack.NewClient(opts.IdentityEndpoint)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to authenticate to OpenStack")
+	}
+
+	// We need to fetch user-provided OpenStack cloud CA certificate and make gophercloud use it.
+	// Also it'll get injected into a ConfigMap mounted into kuryr-controller later on.
+	userCACert, err := getUserCACert(kubeClient)
```
luis5tb: do you need here : if err != nil?

dulek: Not necessarily, the idea is that when error happens this just proceeds as the cert was not set. And indeed that's the case as the ConfigMap will not exist when user had not set the certificates in install-config.yaml.

bindata/network/kuryr/003-config.yaml
```diff
@@ -73,6 +73,11 @@ data:
     user_domain_name = {{ default "\"\"" $AuthInfo.UserDomainName }}
     user_domain_id = {{ default "\"\"" $AuthInfo.UserDomainID }}
     region_name = {{ default "\"\"" .OpenStackCloud.RegionName }}
+{{ if .UserCACertificate }}
+    # There's no good way to just "append" user-provided certs to system ones,
+    # so just configure openstacksdk to use it.
+    cafile = /etc/ssl/certs/user-ca-bundle.crt
+{{ end }}
```
danwinship: pro tip: if you put a `-` after the `{{` here and on the `{{ end }}` line, it will delete the extra blank lines you'd otherwise get

pkg/platform/openstack/kuryr_bootstrap.go
luis5tb: awesome! thanks!

bindata/network/kuryr/003-config.yaml
dulek: Thanks, I fixed all the places.

dulek:
/hold

Seems like the name of the ConfgMap might change, I'll try updating this once I know the new one.

PR #449, opened 2020-01-16
Bug 1791278: Stop deploying kuryr-admission-controller if double listeners supported #

luis5tb:
If octavia amphora provider supports double listeners, i.e.,
exposing the same port but on different protocol (e.g., udp and tcp)
then there is no need for deploying the kuryr-admission-controller
to enforce dns resolution over TCP, as we can create both udp and
tcp listeners on port 53 for the DNS service

cherry-pick of #444

### DIFF
--- a/bindata/network/kuryr/007-service.yaml
+++ b/bindata/network/kuryr/007-service.yaml
@@ -1,4 +1,4 @@
----
+{{if .AdmissionController}}
 apiVersion: v1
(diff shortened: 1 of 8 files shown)


luis5tb:
This is needed to avoid the hack of using TCP for DNS resolution if the underlying OpenStack Octavia supports UDP loadbalancing

PR #451, opened 2020-01-20
sdn: use the correct socket path

rphillips:
Fix the socket path for the cni server socket


danwinship:
This fix is correct, but I think that code is useless anyway... as far as I remember, it was originally added (in openshift-ansible) to handle the case of upgrading from openshift-sdn running directly on the node to openshift-sdn running in a container, where it was difficult to synchronize the switchover for some reason. The check got copied over from openshift-ansible to cluster-network-operator in 4.0, but with sdn always running in a pod now, there shouldn't ever be any case where a new openshift-sdn process gets started while the old one is still running...

I was going to say "we should look through CI logs to see if the warning ever gets printed", except of course that since the check was broken, we know it doesn't. So, OK, let's merge this now and then check later...

/lgtm
/approve

PR #453, opened 2020-01-22
Kuryr bootstrap code split

dulek:
kuryr_bootstrap.go grown too much, this commit splits it into several
files with the OpenStack service criteria.


dulek:
/hold

This should get its chance once master is opened to 4.5.

PR #456, opened 2020-01-23
Bug 1807072: Adds route-override-cni and whereabouts for secondary network routing and IPAM

dougbtv:
Updates YAML templates for init containers, primarily. Condenses
the binary copy script into a single place to be used across other
CNI plugin binary copies.

Likely dependent on https://github.com/openshift/release/pull/6815

### DIFF
--- a/bindata/network/additional-networks/crd/001-crd.yaml
+++ b/bindata/network/additional-networks/crd/001-crd.yaml
@@ -26,3 +26,69 @@ spec:
           properties:
             config:
               type: string
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: ippools.whereabouts.cni.cncf.io
+spec:
+  group: whereabouts.cni.cncf.io
+  names:
+    kind: IPPool
+    plural: ippools
+  scope: ""
+  validation:
+    openAPIV3Schema:
+      description: IPPool is the Schema for Whereabouts for IP address allocation
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: IPPoolSpec defines the desired state of IPPool
+          properties:
+            allocations:
+              additionalProperties:
+                description: IPAllocation represents metadata about the pod/container
+                  owner of a specific IP
+                properties:
+                  id:
+                    type: string
+                required:
+                - id
+                type: object
+              description: Allocations is the set of allocated IPs for the given range.
+                Its indices are a direct mapping to the IP with the same index/offset
+                for the pool's range.
+              type: object
+            range:
+              description: Range is a RFC 4632/4291-style string that represents an
+                IP address and prefix length in CIDR notation
+              type: string
+          required:
+          - allocations
+          - range
+          type: object
+      type: object
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+  preserveUnknownFields: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -1,3 +1,88 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cni-binary-copy-script
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This is a script used to copy CNI binaries based on host OS
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+data:
+  cnibincopy.sh: |-
+    #!/bin/bash
+
+    DESTINATION_DIRECTORY=/host/opt/cni/bin/
+
+    # Perform validation of usage
+    if [ -z "$RHEL7_SOURCE_DIRECTORY" ] ||
+       [ -z "$RHEL8_SOURCE_DIRECTORY" ] ||
+       [ -z "$DEFAULT_SOURCE_DIRECTORY" ]; then
+      echo "FATAL ERROR: You must set env variables: RHEL7_SOURCE_DIRECTORY, RHEL8_SOURCE_DIRECTORY, DEFAULT_SOURCE_DIRECTORY"
+      exit 1
+    fi
+
+    if [ ! -d "$DESTINATION_DIRECTORY" ]; then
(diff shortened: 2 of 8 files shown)


bindata/network/additional-networks/crd/001-crd.yaml
```diff
@@ -13,3 +13,68 @@ spec:
...
+  name: ippools.whereabouts.cni.k8s.io
+spec:
+  group: whereabouts.cni.k8s.io
+  names:
+    kind: IPPool
+    plural: ippools
+  scope: ""
+  validation:
+    openAPIV3Schema:
+      description: IPPool is the Schema for the ippools API
```
danwinship: The documentation here is weird; it's there, but it's not enough to have any idea what any of this means unless you already know...

bindata/network/additional-networks/crd/001-crd.yaml
```diff
@@ -13,3 +13,68 @@ spec:
...
     shortNames:
     - net-attach-def
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: ippools.whereabouts.cni.k8s.io
+spec:
+  group: whereabouts.cni.k8s.io
```
danwinship: If whereabouts isn't an official Kubernetes project then it shouldn't be using a `k8s.io` API group...

bindata/network/additional-networks/crd/001-crd.yaml
```diff
@@ -13,3 +13,68 @@ spec:
...
+                description: IPAllocation represents metadata about the pod/container
+                  owner of a specific IP
+                properties:
+                  id:
+                    type: string
+                required:
+                - id
+                type: object
+              description: Allocations is the set of allocated IPs for the given range.
+                Its` indices are a direct mapping to the IP with the same index/offset
```
danwinship: (stray backtick)

bindata/network/multus/multus.yaml
```diff
@@ -1,3 +1,82 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fips-binary-copy-script
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This is a script used to copy binaries in a FIPS compliant fashion
```
danwinship: "copy binaries in a FIPS compliant fashion" makes it sound like you're using some cryptographically-secure version of `/usr/bin/cp`. This script really has nothing do to with FIPS per se, it's just that for FIPS reasons, we need separate RHEL7 and RHEL8 copies of CNI plugins, and this deals with that. I'd rename it all to be about installing CNI plugins rather than FIPS. And you can get rid of `DESTINATION_DIRECTORY`; it's always `/host/opt/cni/bin`.

Meanwhile, see also https://issues.redhat.com/browse/SDN-695. This script solves some of the problems there, though it still requires each daemonset to mount the right underlying files/dirs onto `/host/opt/cni/bin`, `/host/etc/os-release`, etc

bindata/network/multus/multus.yaml
```diff
@@ -1,3 +1,82 @@
...
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+data:
+  fipsbincopy.sh: |-
+    #!/bin/bash
+    set -x
```
danwinship: No, don't do that (`set -x`) please. It just fills logs with junk

bindata/network/multus/multus.yaml
```diff
@@ -29,101 +108,204 @@ spec:
...
+        - name: RHEL7_SOURCE_DIRECTORY
+          value: "/usr/src/plugins/rhel7/bin/"
+        - name: RHEL8_SOURCE_DIRECTORY
+          value: "/usr/src/plugins/rhel8/bin/"
+        - name: DEFAULT_SOURCE_DIRECTORY
+          value: "/usr/src/plugins/bin/"
+        - name: DESTINATION_DIRECTORY
+          value: "/host/opt/cni/bin/"
+      - name: routeoverride-cni
+        image: {{.RouteOverrideImage}}
```
danwinship: Do we really need to install this onto every node in every cluster? Can't we make this conditional on configuration somehow? (ie, it only gets deployed if you're using it). Similarly, does this really belong in the multus daemonset? It's really more "additional-networks" than "multus" isn't it?

Alternatively, perhaps we could copy all of the CNI binaries into a single image...

bindata/network/additional-networks/crd/001-crd.yaml
dougbtv: Agreed, it's like using the word in a definition, I'll touch this up.

bindata/network/additional-networks/crd/001-crd.yaml
dougbtv: Thanks for the eyes, this is an upstream oversight, it was intended to be cni.cncf.io.... I'll try to get it patched up...

bindata/network/multus/multus.yaml
dougbtv: +1, yeah I'll change the language here to talk about cni binary copy based on host OS, makes more sense.

bindata/network/multus/multus.yaml
dougbtv: Good considerations all. As for conditional copy, the only thing that really works this way is for DHCP CNI -- and in that case it's because it needs additional components (like the DHCP CNI running in daemon mode).

Previously, all the CNI binaries have been copied in these init containers for Multus... Well, the reference CNI plugins + the Multus binary, and now we add two more. I'll look at moving the CNI binary copies to be organized in the additional networks header...

I like the idea of condensing all of the CNI binaries into a single image, I think going forward that would likely be cleaner. However, there's a variety of upstream sources for these which adds some degree of complexity.

dougbtv:
New commit pushed addressing these issues:

* Updates language about CNI binary copy, also changes names to prefer "cni binary copy"
* Removes redundant `DESTINATION_DIRECTORY` from script
* Updates description of IPPool CRD (and removes stray backtick)

Still has route-override copied in the same location as the other CNI binaries in this daemonset.

Does not address the CRD namespace yet, I started an upstream conversation about it and looking into remedying that.

dougbtv:
Pushed another commit with updates to the CRD namespace

This depends on this PR: https://github.com/openshift/whereabouts-cni/pull/4

dougbtv:
/retest

images should be public now

dougbtv:
/retest

Still failing with an ErrImagePull but, I did that as soon as I found they were public, trying again in case there's some kind of delay.

dougbtv:
/retest

Still seeing `errimagepull` but for the first time I see the OSBS builds are working, so wondering if other parts of the build system have changed.

dougbtv:
New commit pushed with the *proper* names, updated to `quay.io/openshift/origin-multus-whereabouts-ipam-cni:4.3` and `quay.io/openshift/origin-multus-route-override-cni:4.3` (had older names previously)

bindata/network/additional-networks/crd/001-crd.yaml
```diff
@@ -26,3 +26,68 @@ spec:
           properties:
             config:
               type: string
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: ippools.whereabouts.cni.cncf.io
+spec:
```
abhat: add `preserveUnknownFields: true`

dougbtv: Got it, thanks for the eyes Aniket!

PR #457, opened 2020-01-29
Bug 1796114: Kuryr: Do not reconfigure if LB providers change

dulek:
When bootstrapping OpenStack resources CNO does autodetection of
available Octavia providers and if OVN provider is at hand Kuryr will
get configured to use it.

In order to make sure Kuryr won't get reconfigured when underlying
OpenStack cluster changes, an annotation marking the initial choice will
be saved to the kuryr-config ConfigMap created by CNO and if present CNO
will always choose that Octavia provider.

In order to force CNO to reconfigure Kuryr the annotation needs to be
manually removed by the user to indicate that this is intended and user
had followed upgrade documentation.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -60,6 +60,7 @@ const (
 	MinOctaviaVersionWithTagSupport        = "v2.5"
 	MinOctaviaVersionWithTimeouts          = "v2.1"
 	KuryrNamespace                         = "openshift-kuryr"
+	KuryrConfigMapName                     = "kuryr-config"
 	// NOTE(ltomasbo): Only OVN octavia driver supported on kuryr
 	OVNProvider = "ovn"
 )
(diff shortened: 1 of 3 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -820,16 +821,32 @@ func ensureCertificate(kubeClient client.Client, caPEM []byte, privateKey []byte
...
+	return cm, nil
+}
+
+func getUserCACert(kubeClient client.Client) (string, error) {
+	cm, err := getConfigMap(kubeClient, OpenShiftConfigNamespace, UserCABundleConfigMap)
+	if err != nil {
+		return "", err
 	}
-	err := kubeClient.Get(context.TODO(), client.ObjectKey{Namespace: OpenShiftConfigNamespace, Name: UserCABundleConfigMap}, cm)
+	return cm.Data["ca-bundle.pem"], nil
```
luis5tb: before we were doing string(cm.Data...), was that not needed?

dulek: Nah, it's `String` already. I copied it from the other operators code but my editor warned that conversion is redundant.

PR #460, opened 2020-01-30
Bug 1796562: sdn/ovn: don't allow ovs-vswitchd to clean up datapath flows on exit

dcbw:
When terminated gracefully, either via 'ovs-ctl stop' or TERM,
ovs-vswitchd cleans up datapath flows. This prevents any existing
flows from continuing to work, which kinda defeats the purpose of
a seamless upgrade.

@orgcandman @knobunc @danwinship @rcarrillocruz @smarterclayton

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -55,12 +55,15 @@ spec:
             fi
           done
 
-          # launch OVS
           function quit {
-              /usr/share/openvswitch/scripts/ovs-ctl stop
+              # Don't allow ovs-vswitchd to clear datapath flows on exit
+              kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true
+              kill $(cat /var/run/openvswitch/ovsdb-server.pid 2>/dev/null) 2>/dev/null || true
               exit 0
           }
           trap quit SIGTERM
+
+          # launch OVS
           /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random
 
           # Restrict the number of pthreads ovs-vswitchd creates to reduce the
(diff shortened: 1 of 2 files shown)


bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -132,7 +134,15 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["/usr/share/openvswitch/scripts/ovs-ctl", "stop"]
+              command:
+              - /bin/bash
+              - -c
+              - |
+                #!/bin/bash
+                set -e
```
juanluisvaladas: I have my doubts about this...
$ help set | grep -- -e
      -e  Exit immediately if a command exits with a non-zero status.

So  in other words if ```kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid)``` fails (for instance, if that PID wouldn't exist anymore for watherever reason) then we would never execute the ```kill $(cat /var/run/openvswitch/ovsdb-server.pid)```

I'm not sure if this is the behavior we want...

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -107,7 +109,15 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["/usr/share/openvswitch/scripts/ovs-ctl", "stop"]
+              command:
+              - /bin/bash
+              - -c
+              - |
+                #!/bin/bash
+                set -e
```
juanluisvaladas: Same thing.

dcbw: Good point, I was just cargo culting. Will remove.

bindata/network/openshift-sdn/sdn-ovs.yaml
dcbw: Fixed. Did || true where the existing script had -e and removed the -e for the preStop hook because failure there will just kill the container later anyway.

danwinship:
hm... we shouldn't need the `preStop` hook should we? Trapping `SIGTERM` should be sufficient. (This isn't new; it's redundant in the old code too, I just hadn't noticed it before.)

dcbw:
> hm... we shouldn't need the `preStop` hook should we? Trapping `SIGTERM` should be sufficient. (This isn't new; it's redundant in the old code too, I just hadn't noticed it before.)

@danwinship I think you're right, so I removed it.

dcbw:
Flake: 
[Feature:Builds] forcePull should affect pulling builder images  [Top Level] [Feature:Builds] forcePull should affect pulling builder images  ForcePull test case execution s2i [Suite:openshift/conformance/parallel]

/retest

PR #462, opened 2020-01-31
Kuryr: Handle unexpected errors fetching CA cert

dulek:
Other implementations of OpenStack cloudprovider CA certs handling print
errors other than NotFound when fetching the CA cert ConfigMap. This
commit makes sure CNO follows the same pattern.


dulek:
I'm not sure if this is still allowed to get into 4.4, it's technically not a bugfix.

luis5tb:
Seems like it may need a bugzilla associated to it

PR #464, opened 2020-01-31
Enables KIND deployments for CNO

trozet:
This patch enables using Kubernetes-in-Docker (KIND) for deploying CNO
with upstream K8S. This is useful for developers who need a quick
environment locally on their laptop. In addition, the script is capable
of building CNO and/or OVN-K8S from a local git workspace to use for
deployment.

### DIFF
--- a/hack/KIND_CNO.md
+++ b/hack/KIND_CNO.md
@@ -1,0 +1,67 @@
+# Kubernetes in Docker (KIND) CNO Deployment
+
+The [KIND](https://github.com/kubernetes-sigs/kind) deployment is used for reproducing an OpenShift networking
+environment with upstream K8S. The value proposition is really for developers who want to reproduce an issue or test a
+fix in an environment that can be brought up locally and within a few minutes.
+
+## How does it work?
+
+At a high level, the deployment will create a docker container per K8S node. This docker container will host its own
+containerd instance, and pods, kubelet, etc will all run inside of these docker containers. The ``kind.yaml`` file
+declares how many nodes will be deployed. Currently only a single node is supported. There are plans to support HA as
+well as more nodes. ``ovn-kind-cno.sh`` is the script that will handle deploying K8S, CNO, and consequently
+Multus/OVN-K8S.
+
+## Requirements
+
+The deployment should work locally on a laptop for a single node. To deploy, it is required to have golang
+(1.11 or later), kubectl, and docker installed. The KIND deployment requires at least Kubernetes v1.17 (installed by
+default) so building corresponding kubectl client is best.
+
+It is also required to install KIND before deploying. As of this writing the latest KIND release is v0.7.0, and
+installation instructions can be found at https://github.com/kubernetes-sigs/kind#installation-and-usage.
+
+## Usage
+
+Simply run ``ovn-kind-cno.sh`` to deploy. Additionally, a user may wish to build CNO or OVN-K8S locally. To build CNO
+simply execute:
+
+````
+$ BUILD_CNO=true ovn-kind-cno.sh
+```` 
+
+It is recommended to build CNO. In order to deploy changes are needed to the OVN yaml files, which are included within
+the CNO docker image. If CNO is not built, these files are patched live within the running CNO pod. If that pod is
+restarted post deployment, the changes will be lost, and OVN will stop functioning correctly.
+
+To build OVN K8S:
+
+````
+$ BUILD_OVN=true ovn-kind-cno.sh
+````
+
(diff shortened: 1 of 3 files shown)


trozet:
@abhat @dougbtv FYI

hack/KIND_CNO.md
```diff
@@ -0,0 +1,62 @@
...
+
+At a high level, the deployment will create a docker container per K8S node. This docker container will host its own
+containerd instance, and pods, kubelet, etc will all run inside of these docker containers. The ``kind.yaml`` file
+declares how many nodes will be deployed. Currently only a single node is supported. There are plans to support HA as
+well as more nodes. ``ovn-kind-cno.sh`` is the script that will handle deploying K8S, CNO, and consequently
+Multus/OVN-K8S.
+
+## Requirements
+
+To deploy, it is required to have golang, kubectl, and docker installed. The deployment should work locally on a laptop.
```
trozet: missing requirement for installing kind

hack/ovn-kind-cno.sh
```diff
@@ -0,0 +1,200 @@
...
+
+echo "Creating OVN CNO config"
+cat << EOF | kubectl create -f -
+apiVersion: config.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+  clusterNetwork:
+  - cidr: 10.244.0.0/16
```
juanluisvaladas: Probably we want to make it customizable via environment variables. I'm not quite sure if this doesn't colide with some of our internal ip addressing something like:
```
CLUSTER_CIDR=${CLUSTER_CIDR:-"10.244.0.0/16"}
SERVICE_NETWORK=${SERVICE_NETWORK:-"10.96.0.0/12"}
```
[...]
```
cat << EOF | kubectl create -f -
apiVersion: config.openshift.io/v1
kind: Network
metadata:
  name: cluster
spec:
  clusterNetwork:
  - cidr: ${CLUSTER_CIDR}
    hostPrefix: 24
  networkType: OVNKubernetes
  serviceNetwork:
  - ${SERVICE_NETWORK}
EOF
```

trozet: thanks. will fix

danwinship: > Probably we want to make it customizable via environment variables. I'm not quite sure if this doesn't colide with some of our internal ip addressing something like:

You should really use the default service range (`172.30.0.0/16`) just for recognizability.
And using a cluster CIDR in the 172.16.0.0 – 172.31.255.255 block as well ought to avoid both RH VPN and local wifi network IP ranges for most people.

hack/ovn-kind-cno.sh
```diff
@@ -0,0 +1,202 @@
...
+  kubectl create -f ${CNO_TEMPLATES}/$f
+done
+
+if [ "$BUILD_OVN" = true ] || [ "$BUILD_CNO" = true ]; then
+  mv deployment.yaml.bk $DEPLOYMENT_TEMPLATE
+  popd
+fi
+
+count=1
+until kubectl get pod -n openshift-network-operator -o jsonpath='{.items[0].metadata.name}' --field-selector status.phase=Running &> /dev/null ;do
```
abhat: reachability check fails here without a rule in iptables to allow the pod to reach out to the api server

hack/KIND_CNO.md
```diff
@@ -0,0 +1,67 @@
...
+the CNO docker image. If CNO is not built, these files are patched live within the running CNO pod. If that pod is
+restarted post deployment, the changes will be lost, and OVN will stop functioning correctly.
+
+To build OVN K8S:
+
+````
+$ BUILD_OVN=true ovn-kind-cno.sh
+````
+
+This will build a CNO docker image to use with the deployment from a local git workspace. By default the script will use
```
dcbw: @trozet "build a CNO" --> do you mean "build an ovn-kubernetes" here?

hack/ovn-kind-cno.sh
trozet: thanks I will add a rule to allow traffic in the Docker table.

hack/KIND_CNO.md
trozet: yeah thanks will fix.

hack/ovn-kind-cno.sh
trozet: I think these are default flannel values? I'll change it to what you suggested.

danwinship:
I didn't really look at this, but it doesn't touch any existing files so if it helps people then
/lgtm

PR #465, opened 2020-02-04
bug 1803193: Tighten security groups

MaysaMacedo:
This commit ensures the security group are limiting access
to the minimum required.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -62,6 +62,8 @@ const (
 	KuryrNamespace                         = "openshift-kuryr"
 	// NOTE(ltomasbo): Only OVN octavia driver supported on kuryr
 	OVNProvider = "ovn"
+	EtcdPort    = 2379
+	DNSPort     = 53
 )
 
 func GetClusterID(kubeClient client.Client) (string, error) {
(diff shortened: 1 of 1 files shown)


MaysaMacedo:
Still needs to be verified.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -1023,16 +1023,12 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		return nil, errors.Wrap(err, "failed to add rule opening traffic pod to pod")
 	}
 	for _, cidr := range podSubnetCidrs {
-		err = ensureOpenStackSgRule(client, masterSgId, cidr, -1, -1)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to add rule opening traffic to masters on %s", cidr)
-		}
 		err = ensureOpenStackSgRule(client, workerSgId, cidr, -1, -1)
```
dulek: Why would the traffic to the workers be still open?

MaysaMacedo: The console pods require traffic to be open on the worker, and the etcd pods required traffic to be open on the master.

dulek:
/lgtm

This seems okay, but I wonder if we aren't blocking some feature that is not required for bare install. Had you tried running conformance tests on that?

MaysaMacedo:
> /lgtm
> 
> This seems okay, but I wonder if we aren't blocking some feature that is not required for bare install. Had you tried running conformance tests on that?

@dulek I did run conformance on osp13 both with and without the changes, the number of failing tests was almost the same.  I also testes in osp16 with the changes, I'm testing without now to compare.

PR #472, opened 2020-02-11
WIP: Save the ovs flows before terminating the ovs pod

knobunc:
This saves the ovs flows when a pod is shut down and restores them
when a pod is restarted.  This should make the ovs behavior better
match that of a system daemon and make it respond to packets more
rapidly, rather than being empty when the pod restarts and waiting for
the sdn container to reprogam all of the rules.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -56,6 +56,10 @@ spec:
           done
 
           function quit {
+              # Save the flows
+              bridges=$(ovs-vsctl -- --real list-br)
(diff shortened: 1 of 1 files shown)


danwinship:
I thought we were killing OVS to prevent it from deleting the flows on shutdown now, so this should have no effect?

knobunc:
@danwinship not quite.  We kill it so that it doesn't disturb the cached kernel flows... but as soon as ovs starts, it will have no flows and wipe the kernel.  This changes it so that we save the flows at shutdown and restores them on startup so that we continue processing packets as we did before the shutdown.  Until the sdn pod starts and takes over...

Except for the things that it doesn't correctly know to resync :-(

smarterclayton:
Replaced by #476

PR #473, opened 2020-02-12
Bug 1802596: Remove erroneous ovnkube node command-line args

danwinship:
We were accidentally still passing some command line arguments that we also specified in the config file. Worse, we were passing empty strings for the values on the command line...

/cc @trozet


pecameron:
@trozet The nbaddress, certs, etc should all be in the configmap. The plumbing needs some cleanup to handle it. This is left over from when we went from tcp to ssl and allowed either. Now that we are always ssl (I think) it should be easy.

danwinship:
oops
/override ci/prow/e2e-gcp-ovn

PR #474, opened 2020-02-14
Update admission controller progression messaging

dougbtv:
The progression message should reflect that the admission controller is reliant on other systems being ready

Previously the CNO would output a message such as 'DaemonSet "openshift-multus/multus-admission-controller" is not available', which would cause users to believe that during a cluster bootstrapping error there was a problem with the multus-admission-controller. However, this is not a dependency of other systems, and therefore should have an error message reflecting that other systems may be involved.

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -86,7 +86,11 @@ func (status *StatusManager) SetFromPods() {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
 			dsProgressing = true
 		} else if ds.Status.NumberUnavailable > 0 {
-			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
+			if isNonCritical(ds) {
+				progressing = append(progressing, fmt.Sprintf("DaemonSet %q is waiting for other operators to become ready", dsName.String()))
+			} else {
+				progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
+			}
 			dsProgressing = true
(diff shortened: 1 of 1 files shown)


pkg/controller/statusmanager/pod_status.go
```diff
@@ -86,7 +86,11 @@ func (status *StatusManager) SetFromPods() {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
 			dsProgressing = true
 		} else if ds.Status.NumberUnavailable > 0 {
-			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
+			if strings.Contains(dsName.String(), "multus-admission-controller") {
```
dougbtv: Is there a better way to store this string as a constant, or pull it from another place? I couldn't quite find the right place to store it.

dougbtv:
/retest

Looks like a gcp flake with `failed to update lease`

pkg/controller/statusmanager/pod_status.go
danwinship: We already use the "non-critical" annotation to identify the multus admission controller, so just do `if isNonCritical(ds) {`.

pkg/controller/statusmanager/pod_status.go
```diff
@@ -86,7 +86,11 @@ func (status *StatusManager) SetFromPods() {
...
+				progressing = append(progressing, fmt.Sprintf("DaemonSet %q is waiting for other operators to become ready (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
```
danwinship: it doesn't make sense to include the `NumberUnavailable` here because we only want the special case message if it's unavailable on all nodes. So I guess the previous condition should be `isNonCritical(ds) && ds.Status.NumberAvailable == 0` and then don't print a number here

dougbtv:
New commit up which addresses using `isNonCritical(ds)` (thanks for pointing that out!) and updates the output to omit the count of nodes.

dougbtv:
/retest

Looks possibly like GCP infra causing a flake (maybe.)

danwinship:
/lgtm
thanks!

PR #477, opened 2020-02-14
[WIP] Start host openvswitch using systemctl

mccv1r0:
Start host openvswitch using systemctl

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -42,6 +42,18 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
+          set -x
+          if [[ -f /usr/bin/id ]]; then
+            /usr/bin/id openvswitch
+          else
+            echo "id command not found"
+          fi
+          if [[ -d /run/openvswitch ]]; then
+            ls -al /run/openvswitch
+          fi
+          if [[ -d /var/run/openvswitch ]]; then
+            ls -al /var/run/openvswitch
+          fi
           if [[ -f "/env/${K8S_NODE}" ]]; then
             set -o allexport
             source "/env/${K8S_NODE}"
@@ -50,6 +62,18 @@ spec:
           echo "$(date -Iseconds) - starting ovs-daemons"
           chown -R openvswitch:openvswitch /run/openvswitch
           chown -R openvswitch:openvswitch /etc/openvswitch
+          if [[ -d /run/openvswitch ]]; then
+            ls -al /run/openvswitch
+          fi
+          if [[ -d /var/run/openvswitch ]]; then
+            ls -al /var/run/openvswitch
+          fi
+          if [[ -f /var/run/openvswitch/ovs-vswitchd.pid ]] ; then
+            cat /var/run/openvswitch/ovs-vswitchd.pid
+          fi
+          if [[ -f /var/run/openvswitch/ovsdb-server.pid ]] ; then
+            cat /var/run/openvswitch/ovsdb-server.pid
+          fi
           function quit {
               # Don't allow ovs-vswitchd to clear datapath flows on exit
               kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true
@@ -57,9 +81,37 @@ spec:
               exit 0
           }
           trap quit SIGTERM
+          if [[ -S /var/run/openvswitch/db.sock ]]; then
+            echo "Found /var/run/openvswitch/db.sock"
+          else
+            echo "File /var/run/openvswitch/db.sock was not found"
+          fi
+          if [[ -S /run/openvswitch/db.sock ]]; then
+            echo "Found /run/openvswitch/db.sock"
+          else
+            echo "File /run/openvswitch/db.sock was not found"
+          fi
+          export SYSTEMD_IGNORE_CHROOT=yes
+          journalctl -xeu ovsdb-server --no-pager
+          #systemctl is-active --quiet ovsdb-server
+          journalctl -xeu openvswitch --no-pager
+          if (modprobe openvswitch) ; then
+            echo "modprobe openvswitch success?"
+          else
+            echo "modprobe openvswitch fail?"
+          fi
+          lsmod | grep openvswitch
+          if (systemctl is-active --quiet openvswitch) ; then
+            echo "OVS started by RHCOS"
+            systemctl is-active openvswitch
+            journalctl -xeu openvswitch --no-pager
+          else
+            echo "OVS started by ovs-node container, NOT RHCOS"
+          # Need container up v0.0 just to see if host starts ovs and friends
           /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
           /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+          fi
           echo "$(date -Iseconds) - ovs-daemons running"
 
           tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
@@ -82,9 +134,15 @@ spec:
           name: host-sys
           readOnly: true
         - mountPath: /run/openvswitch
-          name: run-openvswitch
+          name: host-run-openvswitch
+        - mountPath: /run/systemd
+          name: host-run-systemd
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
+        - mountPath: /var/log/openvswitch
+          name: host-var-log-openvswitch
+        - mountPath: /sys/fs/cgroup
+          name: sys-fs-cgroup
         - mountPath: /var/lib/openvswitch
           name: var-lib-openvswitch
         - mountPath: /env
@@ -120,22 +178,33 @@ spec:
               - key: network.operator.openshift.io/external-openvswitch
                 operator: DoesNotExist        
       volumes:
-      # used for iptables wrapper scripts
       - name: host-modules
         hostPath:
           path: /lib/modules
       - name: var-lib-openvswitch
         hostPath:
           path: /var/lib/openvswitch/data
+      - name: host-run-openvswitch
+        hostPath:
+          path: /run/openvswitch
+          type: Directory
+      - name: host-run-systemd
+        hostPath:
+          path: /run/systemd
+          type: Directory
       - name: etc-openvswitch
         hostPath:
           path: /var/lib/openvswitch/etc
-      - name: run-openvswitch
+      - name: host-var-log-openvswitch
         hostPath:
-          path: /run/openvswitch
+          path: /var/log/openvswitch
+          type: Directory
       - name: host-sys
         hostPath:
           path: /sys
+      - name: sys-fs-cgroup
+        hostPath:
+          path: /sys/fs/cgroup
       - name: env-overrides
         configMap:
           name: env-overrides
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -414,7 +414,9 @@ spec:
         hostPath:
           path: /var/lib/ovn/data
       - name: run-openvswitch
-        emptyDir: {}
+        hostPath:
+          path: /run/openvswitch
+          type: Directory
       - name: run-ovn
         hostPath:
           path: /var/run/ovn
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -41,12 +41,46 @@ spec:
         - -c
         - |
           set -e
+          set -x
+          if [[ -f /usr/bin/id ]]; then
+            /usr/bin/id openvswitch
+          else
+            echo "id command not found"
+          fi
+          if [[ -d /run/openvswitch ]]; then
+            ls -al /run/openvswitch
+          fi
+          if [[ -d /var/run/openvswitch ]]; then
+            ls -al /var/run/openvswitch
+          fi
+          if [[ -f /var/run/openvswitch/ovs-vswitchd.pid ]] ; then
+            cat /var/run/openvswitch/ovs-vswitchd.pid
+          fi
+          if [[ -f /var/run/openvswitch/ovsdb-server.pid ]] ; then
+            cat /var/run/openvswitch/ovsdb-server.pid
+          fi
           if [[ -f "/env/${K8S_NODE}" ]]; then
             set -o allexport
             source "/env/${K8S_NODE}"
             set +o allexport
           fi
           echo "$(date -Iseconds) - starting ovn-controller"
+          if (modprobe openvswitch) ; then
+            echo "modprobe openvswitch success?"
+          else
+            echo "modprobe openvswitch fail?"
+          fi
+          lsmod | grep openvswitch
+          if [[ -S /var/run/openvswitch/db.sock ]]; then
+            echo "Found /var/run/openvswitch/db.sock"
+          else
+            echo "File /var/run/openvswitch/db.sock was not found"
+          fi
+          if [[ -S /run/openvswitch/db.sock ]]; then
+            echo "Found /run/openvswitch/db.sock"
+          else
+            echo "File /run/openvswitch/db.sock was not found"
+          fi
           exec ovn-controller unix:/var/run/openvswitch/db.sock -vfile:off \
             --no-chdir --pidfile=/var/run/ovn/ovn-controller.pid \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
@@ -61,8 +95,11 @@ spec:
             fieldRef:
               fieldPath: spec.nodeName
         volumeMounts:
+        - mountPath: /host
+          name: host-slash
+          readOnly: false
         - mountPath: /run/openvswitch
-          name: run-openvswitch
+          name: host-run-openvswitch
         - mountPath: /run/ovn/
           name: run-ovn
         - mountPath: /etc/openvswitch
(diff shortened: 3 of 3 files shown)


bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -87,6 +90,25 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /env
           name: env-overrides
+        - mountPath: /run/systemd
+          name: run-systemd
+        - mountPath: /sysroot
+          name: host-sysroot
+        - mountPath: /var/log/openvswitch
+          name: var-log-openvswitch
+        - mountPath: /usr/shre/openvswitch/scripts
```
dcbw, line 99: @mccv1r0 /usr/share? Looks like we don't actually use it since we're using the container's scripts here.

dcbw:
@mccv1r0 can you:

1. mcc to clean this up and submit for real
2. add a check (systemctl status?) to make sure it's actually started before we return from the container, probably after the tail to make sure we get the first couple seconds or so of the log liens from OVS
3. make the container run only once (that might require making it an init container or something, not sure)

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,6 +42,7 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
+          set -x
```
dcbw, line 45: @mccv1r0 let's remove this for the final one

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,21 @@ spec:
...
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
           /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
 
           tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
           tail -F --pid=$(cat /var/run/openvswitch/ovsdb-server.pid) /var/log/openvswitch/ovsdb-server.log &
+          SYSTEMD_IGNORE_CHROOT=yes systemctl is-active --quiet openvswitch
+          status=$?
+          if [[ "$status" == 0 ]]; then
+              echo "OK"
+          else
```
dcbw, line 71: @mccv1r0 could you grab the output of `systemctl status` when it fails?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,21 @@ spec:
...
+          SYSTEMD_IGNORE_CHROOT=yes systemctl start openvswitch
           ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
           /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
 
           tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
           tail -F --pid=$(cat /var/run/openvswitch/ovsdb-server.pid) /var/log/openvswitch/ovsdb-server.log &
+          SYSTEMD_IGNORE_CHROOT=yes systemctl is-active --quiet openvswitch
+          status=$?
+          if [[ "$status" == 0 ]]; then
+              echo "OK"
```
dcbw, line 70: Can you make the message a bit more descriptive?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: Agreed

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: reissue `systemctl status` ?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: No problem.  This was just  a first pass to make sure what ran locally worked when translated to yaml.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,21 @@ spec:
               exit 0
           }
           trap quit SIGTERM
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
+          setenforce Permissive
```
cgwalters, line 60: Don't do this.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -87,6 +97,25 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /env
           name: env-overrides
+        - mountPath: /run/systemd
```
cgwalters, line 100: It may be simpler to mount the whole host, and use `chroot /host` for selected commands.  The MCD today just runs entirely chrooted into the host.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
cgwalters: Can you get/paste the selinux denials that drove you to do this?

cgwalters:
BTW please avoid making any *persistent* changes to the host as part of dynamic configuration.  See also https://github.com/coreos/fedora-coreos-tracker/issues/354

So concretely that means e.g. don't `systemctl enable` from a daemonset that is deriving its "source of truth" from a cluster config object (CRD/configmap/whatever); it's OK to `systemctl start` though as long as you "own" that service.

Also on this topic I'd like to move `openvswitch` out of the host by default into an extension system and have the SDN layer control its installation (and version too).

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: audit2allow provided: 

`
#============= openvswitch_t ==============
allow openvswitch_t container_var_run_t:file { append lock rename };
allow openvswitch_t container_var_run_t:sock_file { create unlink };
allow openvswitch_t spc_t:dir { getattr search };
allow openvswitch_t spc_t:file { getattr open read };
`

cgwalters: OK one general approach you can take that the MCD also does for the ostree side is to basically `systemd-run /bin/sh -c ...`, that way you "escape" the `spc_t` type and go back to the default.

Another approach is to run `chroot /host restorecon /path/to/files`.

cgwalters: On the `systemd-run` side, since you just have a shell script here, you can copy it out to the host and use `systemd-run` to run all or part from the host side.

squeed:
There's an existing enhancement for this: https://github.com/openshift/enhancements/pull/165

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,49 @@ spec:
...
-          ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
-          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+          chroot /host /bin/bash <<EOF
+          set -x
+          ln -s /run /var/run
+          #SYSTEMD_IGNORE_CHROOT=yes systemctl start ovsdb-server
+          export SYSTEMD_IGNORE_CHROOT=yes
+          echo "systemd ignore chroot set to " \$SYSTEMD_IGNORE_CHROOT
+          systemctl start ovsdb-server
+          if [[ $? == 0 ]]; then
```
cgwalters, line 67: I think writing shell script without `set -e` or more generally [strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) is just asking for unreliability at best.  If e.g. the `ln -s` above failed for some reason the script just blindly stumbles on, and something *might* fail later.

We need to ensure that failures are observable and also propagate up into the Kuberentes/OpenShift layers.  See also https://github.com/openshift/machine-config-operator/issues/1365

I think you should use `set -e` in this whole script, and then drop testing `$?`.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,49 @@ spec:
               exit 0
           }
           trap quit SIGTERM
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
-          ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
-          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+          chroot /host /bin/bash <<EOF
```
cgwalters, line 60: I think you likely want to use [quoted EOF](https://stackoverflow.com/questions/22697688/how-to-cat-eof-a-file-containing-code), otherwise e.g. I bet the `$?` is being expanded to `0` from the outer shell's last command success status...

mccv1r0:
These are the AVC denials:

``sh-4.4# cat mccovs.te

module mccovs 1.0;

require {
        type spc_t;
        type unconfined_service_t;
        type openvswitch_t;
        type container_var_run_t;
        type init_t;
        class process setsched;
        class dir { getattr search };
        class file { append getattr lock open read rename };
        class sock_file { create unlink };
}

#============= init_t ==============
allow init_t unconfined_service_t:process setsched;

#============= openvswitch_t ==============
allow openvswitch_t container_var_run_t:file { append lock rename };
allow openvswitch_t container_var_run_t:sock_file { create unlink };
allow openvswitch_t spc_t:dir { getattr search };
allow openvswitch_t spc_t:file { getattr open read };
sh-4.4#

bindata/network/ovn-kubernetes/006-ovs-node.yaml
cgwalters: (Or, write this in a compiled language like Go/Rust)

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -56,12 +57,49 @@ spec:
...
+          set -x
+          ln -s /run /var/run
```
cgwalters, line 62: This shouldn't be necessary?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
cgwalters, line 60: OK and looking at those denials...I think probably what we need to do is write this script out to the host at e.g. `/run/bin/openshift-cno-ovs` and then use:
`chroot /host systemd-run --unit=openshift-cno-ovs /run/bin/openshift-cno-ovs` or so to fully escape the container and execute with the host `unconfined_t` context.

It may also be possible to use `unconfined_t` instead of `spc_t`.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,26 +42,49 @@ spec:
...
-          ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
-          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
+          #cat > /run/bin/openshift-cno-ovs <<EOF
+          cat <<'EOF' > /run/bin/openshift-cno-ovs
+          #!/bin/bash
+          set -euo pipefail
+          set -x
+          export SYSTEMD_IGNORE_CHROOT=yes
+          echo "systemd ignore chroot set to " \$SYSTEMD_IGNORE_CHROOT
+          systemctl start ovsdb-server
```
smarterclayton, line 68: I don't think this is the correct approach - why wouldn't this operator add machine config injection to start this automatically?

I would expect CNO to contribute daemonsets AND an MCO config injector to all pools (except maybe those are invoking) that depends on OVS

cgwalters: There are definitely advantages to that, but a downside is that in order to avoid race conditions, the logic for rendering this config would need to be part of the bootstrap MCO phase at least I'd think.

Otherwise we'd have a situation where the MCO starts and renders a config *without* the CNO setup, then the CNO starts and injects a MC which forces another rollout.

It also gets weird to inject nontrivial *code* via systemd units...the MCO has some of the "kubernetes native shell script" stuff that's a template to run a static pod that runs shell, which got really unwieldly and ended up being moved into a separate operator.

smarterclayton: Yes, they need to be part of bootstrap MCO and CNO needs to contribute to bootstrap MCO

smarterclayton: Here's my top level expectation: "anyone can contribute an MC to the cluster that enables a systemd unit to start at boot of the masters trivially"

cgwalters: > Here's my top level expectation: "anyone can contribute an MC to the cluster that enables a systemd unit to start at boot of the masters trivially"

And then that operator owns upgrades to that MC right?

Yes, there are definitely advantages to this - actually if we took this approach we could probably move some of the "platform specific" stuff out of the MCO into separate operators, like baremetal-operator could own the `baremetal` platform MC stuff right now.

It'd probably be sufficient to extend `bootkube.sh` to have a static list of operators it walks over and asks them to dump machineconfig files into, then the bootstrap MCS just merges those in when it starts.

runcom: > It'd probably be sufficient to extend bootkube.sh to have a static list of operators it walks over and asks them to dump machineconfig files into, then the bootstrap MCS just merges those in when it starts. Then they also get injected into the cluster.

💯

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,23 +42,75 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
+          set -x
+          if [[ -f /usr/bin/id ]]; then
```
rbbratta, line 46: `-x` for executable?  

I don't understand all these checks, they are not atomic and thus worse than redundant because they hid the real error message.

`/usr/bin/id openvswitch || true`

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,23 +42,75 @@ spec:
...
+            /usr/bin/id openvswitch
+          else
+            echo "id command not found"
+          fi
+          if [[ -d /run/openvswitch ]]; then
+            ls -al /run/openvswitch
```
rbbratta, line 52: `ls -al /run/openvswitch/ || true`     trailing `/` forces treating it as a directory.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,23 +42,75 @@ spec:
...
           fi
           chown -R openvswitch:openvswitch /run/openvswitch
           chown -R openvswitch:openvswitch /etc/openvswitch
+          if [[ -d /run/openvswitch ]]; then
+            ls -al /run/openvswitch
+          fi
+          if [[ -d /var/run/openvswitch ]]; then
+            ls -al /var/run/openvswitch
+          fi
+          if [[ -f /var/run/openvswitch/ovs-vswitchd.pid ]] ; then
```
rbbratta, line 70: `-r` for readable.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
rbbratta: or no `|| true` with the `set -e`

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -42,23 +42,75 @@ spec:
...
+          export SYSTEMD_IGNORE_CHROOT=yes
+          journalctl -xeu ovsdb-server --no-pager
+          #systemctl is-active --quiet ovsdb-server
+          journalctl -xeu openvswitch --no-pager
+          if (modprobe openvswitch) ; then
+            echo "modprobe openvswitch success?"
+          else
+            echo "modprobe openvswitch fail?"
+          fi
+          lsmod | grep openvswitch
```
rbbratta, line 102: `grep live /sys/module/openvswitch/initstate`   maybe

bindata/network/ovn-kubernetes/006-ovs-node.yaml
rbbratta: or I guess it's not clear in which environment `id` is not present?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
rbbratta: I guess question here as well, when is `/var/run` not a symlink to `/run` ?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: > I don't understand all these checks, they are not atomic and thus worse than redundant because they hid the real error message.

Most (if not all) of these checks were added to debug and will not exist in the final PR.

cgwalters:
xref https://bugzilla.redhat.com/show_bug.cgi?id=1839065

sinnykumari:
can we get this PR rebased, wanted to test this with https://github.com/openshift/machine-config-operator/pull/1636/?

runcom:
@sinnykumari I think this is rebased now but not sure about all these failures in CI

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -50,16 +62,56 @@ spec:
...
+          if (systemctl is-active --quiet openvswitch) ; then
+            echo "OVS started by RHCOS"
+            systemctl is-active openvswitch
+            journalctl -xeu openvswitch --no-pager
+          else
+            echo "OVS started by ovs-node container, NOT RHCOS"
+          # Need container up v0.0 just to see if host starts ovs and friends
```
squeed, line 110: Can you fix the indentation here?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -50,16 +62,56 @@ spec:
...
+          if (modprobe openvswitch) ; then
+            echo "modprobe openvswitch success?"
+          else
+            echo "modprobe openvswitch fail?"
+          fi
+          lsmod | grep openvswitch
+          if (systemctl is-active --quiet openvswitch) ; then
+            echo "OVS started by RHCOS"
+            systemctl is-active openvswitch
+            journalctl -xeu openvswitch --no-pager
```
squeed, line 107: Probably just want to `exec journalctl -f -u openvswitch` and be done with the script.

sinnykumari:
> @sinnykumari I think this is rebased now but not sure about all these failures in CI

Can it because we would need https://github.com/openshift/machine-config-operator/pull/1636 merged first so that ovn related services are enabled on host to keep cluster install progressing further?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -50,16 +62,56 @@ spec:
...
+          if [[ -d /var/run/openvswitch ]]; then
+            ls -al /var/run/openvswitch
+          fi
+          if [[ -f /var/run/openvswitch/ovs-vswitchd.pid ]] ; then
+            cat /var/run/openvswitch/ovs-vswitchd.pid
+          fi
+          if [[ -f /var/run/openvswitch/ovsdb-server.pid ]] ; then
+            cat /var/run/openvswitch/ovsdb-server.pid
+          fi
           function quit {
```
squeed, line 77: Yikes. This needs to be moved to the "non-systemd case"

mccv1r0: This is all debug code

squeed: not the `kill -9`...

bindata/network/ovn-kubernetes/006-ovs-node.yaml
trozet, line 110: I don't really understand why we need to start OVS from this container. Shouldn't systemd OVS be configured as a service with restart=Always? Why does a pod need to start it?

PR #481, opened 2020-02-17
[WIP] Bug 1807670: Fixes to reliably save/restore flows

abhat:
This PR adds:
a) a mechanism to save/restore flows on ovs pods being stopped/started respectively. This only handles the case where the pod is restarted/recreated on a node after an equivalent of `oc delete pod ovs-xyz`. It doesn't handle the node reboot case, since during reboot it is expected that pod IPs will change.
b) a timestamp when adding logs for save/restore so they can be correlated with other logs.
c) removes the hostPID setting on the daemonset pods
d) removes the --monitor option when starting the ovs-vswitchd and ovsdb-server processes


abhat:
4.3 Backport of #524

PR #490, opened 2020-02-23
Bug 1797894: Add MachineNetworks to proxy.status.noPoxy

juanluisvaladas:
.

### DIFF
--- a/pkg/util/proxyconfig/no_proxy.go
+++ b/pkg/util/proxyconfig/no_proxy.go
@@ -2,6 +2,7 @@ package proxyconfig
 
 import (
 	"fmt"
+	"net"
 	"net/url"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


juanluisvaladas:
I added  github.com/openshift/installer as a dependency pointing to the master branch.
Should be safe because we're only using the ipnet.IPNet struct.

All the code changes are in pkg/util/proxyconfig/no_proxy.go
The rest are changes to Gopgk.* and adding github.com/openshift/installer as a dependency.

Gopkg.toml
```diff
@@ -60,6 +60,10 @@ required = [
   branch = "master"
   name = "github.com/openshift/api"
 
+[[constraint]]
+  branch = "master"
+  name = "github.com/openshift/installer"
```
danwinship: we should not be vendoring installer

pkg/util/proxyconfig/no_proxy.go
```diff
@@ -21,14 +22,20 @@ import (
 // are returned.
 func MergeUserSystemNoProxy(proxy *configv1.Proxy, infra *configv1.Infrastructure, network *configv1.Network, cluster *corev1.ConfigMap) (string, error) {
 	// TODO: This will be flexible when master machine management is more dynamic.
+	type machineNetworkEntry struct {
+		// CIDR is the IP block address pool for machines within the cluster.
+		CIDR ipnet.IPNet `json:"cidr"`
```
danwinship: Just make this a `string` and parse it yourself

juanluisvaladas:
/hold
haven't actually tested this yet

juanluisvaladas:
I'm doing the cherry pick before this is merged because this is blocking some people. Hopefully I won't need more changes so this will save some time.
/cherry-pick release-4.4
/cherry-pick release-4.3

danwinship:
> I'm doing the cherry pick before this is merged...

> @juanluisvaladas: once the present PR merges, I will cherry-pick it on top of release-4.4

DENIED!

PR #508, opened 2020-03-01
Bug 1808498: Ensure no API LB recreation happens upon Octavia upgrade

MaysaMacedo:
When upgrading Octavia from OSP13 to OSP16 the CNO checks if
tagging is supported and look for an API load balancer with tag.
However, the existent LB was created with OSP13, which does not
support tagging and so the tagging was added on description.
The CNO then tries to create a new LB API and fails due to
address already in use.

This commit fixes the issue by ensuring no LB exists with tags
and description, when tagging is supported, prior to creating
a new one.

### DIFF
--- a/pkg/platform/openstack/loadbalancer.go
+++ b/pkg/platform/openstack/loadbalancer.go
@@ -142,6 +142,37 @@ func ensureOpenStackLb(client *gophercloud.ServiceClient, name, vipAddress, vipS
 	if err != nil {
 		return "", errors.Wrap(err, "failed to extract LB list")
 	}
+
+	if len(lbs) == 0 && octaviaTagSupport {
+		// Attempt to retrieve API load balancer with description tagging
+		// to avoid another load balancer creation upon Octavia upgrade.
+		opts := loadbalancers.ListOpts{
+			Name:        name,
+			VipAddress:  vipAddress,
+			VipSubnetID: vipSubnetId,
+			Description: tag,
+		}
+		page, err = loadbalancers.List(client, opts).AllPages()
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get LB list")
+		}
+		lbs, err = loadbalancers.ExtractLoadBalancers(page)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to extract LB list")
+		}
+		if len(lbs) == 1 {
+			log.Printf("Tagging existing loadbalancer API %s", lbs[0].ID)
+			tags := []string{tag}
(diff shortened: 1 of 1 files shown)


dulek:
You need to gofmt this: `gofmt -w -s ./pkg/platform/openstack/loadbalancer.go`.

pkg/platform/openstack/loadbalancer.go
```diff
@@ -142,6 +142,26 @@ func ensureOpenStackLb(client *gophercloud.ServiceClient, name, vipAddress, vipS
...
+		}
+		page, err = loadbalancers.List(client, opts).AllPages()
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get LB list")
+		}
+		lbs, err = loadbalancers.ExtractLoadBalancers(page)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to extract LB list")
+		}
+	}
```
dulek: Shouldn't we tag it if we found it? That way on next loops we'll just make one call. Also it'll allow us to drop looking for LBs with tag in description after we'll drop support for Queens.

MaysaMacedo: Yup. Thanks for pointing this.

pkg/platform/openstack/loadbalancer.go
```diff
@@ -142,6 +142,37 @@ func ensureOpenStackLb(client *gophercloud.ServiceClient, name, vipAddress, vipS
...
+		}
+		if len(lbs) == 1 {
+			log.Printf("Tagging existing loadbalancer API %s", lbs[0].ID)
+			tags := []string{tag}
+			updateOpts := loadbalancers.UpdateOpts{
+				Tags: &tags,
+			}
+			_, err := loadbalancers.Update(client, lbs[0].ID, updateOpts).Extract()
+			if err != nil {
+				return "", errors.Wrap(err, "failed to update LB")
```
dulek: IMO this should not be fatal, just print a warning - LB should be fine and we'll try again on next run.

pkg/platform/openstack/loadbalancer.go
dulek: Hm, on the other hand it's better for the user to take a look - otherwise it can end up with LB never updated. Okay, let's leave it as is.

pkg/platform/openstack/loadbalancer.go
dulek: Hm, on the other hand it's better for the user to take a look - otherwise it can end up with LB never updated. Okay, let's leave it as is.

PR #509, opened 2020-03-02
Bug 1810137: Delete admission controller if double listeners is supported

MaysaMacedo:
In case Octavia is upgraded while the CNO is already running, the admission
controller resources are not deleted due to checking the previous Octavia
API version, as the version is only retrieved once when the CNO is starting.
This commit fixes the issue by ensuring the current API version is always
fetched.

### DIFF
--- a/pkg/platform/openstack/loadbalancer.go
+++ b/pkg/platform/openstack/loadbalancer.go
@@ -15,8 +15,6 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
 )
 
-var maxOctaviaVersion *semver.Version = nil
-
 func getMaxOctaviaAPIVersion(client *gophercloud.ServiceClient) (*semver.Version, error) {
 	allPages, err := apiversions.List(client).AllPages()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


dulek:
@MaysaMacedo: You're sure this is required? I think @luis5tb confirmed it will just get deleted automatically once it's out of template renders.

dulek:
The failure on verify job is valid BTW.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -490,6 +491,18 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		return nil, errors.Wrap(err, "failed to determine if Octavia supports double listeners")
 	}
 
+	log.Print("Checking presence of Admission Controller daemonset")
+	if octaviaMultipleListenersSupport {
+		daemonset := &appsv1.DaemonSet{ObjectMeta: metav1.ObjectMeta{Namespace: KuryrNamespace, Name: names.KURYR_DNS_ADMISSION_CONTROLLER}}
```
dulek: First and foremost shouldn't we remove the MutatingWebhookConfiguration? And also the Secrets?

PR #510, opened 2020-03-03
Update Progressing message for non-critical components

danwinship:
I was backporting #474 and realized that it solves `DaemonSet "openshift-multus/multus-admission-controller" is not available ...` (ie, "your install failed after the worker nodes were created but before multus-admission-controller could be deployed") but not `DaemonSet "openshift-multus/multus-admission-controller" is not yet scheduled on any nodes` (ie, "your install failed before the worker nodes were created").

(WIP until I see it doing the right thing in a test run)

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -82,15 +82,14 @@ func (status *StatusManager) SetFromPods() {
 
 		dsProgressing := false
 
-		if ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled {
+		if isNonCritical(ds) && ds.Status.DesiredNumberScheduled == 0 {
+			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is waiting for other operators to become ready", dsName.String()))
+			dsProgressing = true
+		} else if ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled))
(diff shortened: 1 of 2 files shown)


danwinship:
Statuses:

    Waiting for DaemonSet "openshift-multus/multus-admission-controller" to be created
    DaemonSet "openshift-multus/multus-admission-controller" is waiting for other operators to become ready
    DaemonSet "..." update is rolling out (0 out of 1 updated)
    DaemonSet "..." is waiting for other operators to become ready
    DaemonSet "..." is not available (awaiting 2 nodes)
    DaemonSet "..." is not available (awaiting 1 nodes)
    Progressing -> False

so this is not quite right. It shouldn't go back to "waiting for other operators" after it stops that.

danwinship:
ok, now we get:

    Waiting for DaemonSet "openshift-multus/multus-admission-controller" to be created
    DaemonSet "openshift-multus/multus-admission-controller" is waiting for other operators to become ready
    DaemonSet "openshift-multus/multus-admission-controller" update is rolling out (0 out of 1 updated)
    DaemonSet "openshift-multus/multus-admission-controller" is not available (awaiting 1 nodes)
    DaemonSet "openshift-multus/multus-admission-controller" is not available (awaiting 2 nodes)
    DaemonSet "openshift-multus/multus-admission-controller" update is rolling out (2 out of 3 updated)
    DaemonSet "openshift-multus/multus-admission-controller" is not available (awaiting 3 nodes)
    DaemonSet "openshift-multus/multus-admission-controller" is not available (awaiting 2 nodes)
    DaemonSet "openshift-multus/multus-admission-controller" is not available (awaiting 1 nodes)

    Progressing -> False

which is still a little weird, but it's consistent with our other reporting, which we ought to improve at some point. But (1) it stays at "waiting for other operators" until it starts being deployed, and (2) it doesn't go back to "waiting for other operators" after that.

PR #511, opened 2020-03-03
Increase termination grace period to 45s to allow for flows to get saved

abhat:
This pull request changes the termination grace period to 45s from 10s to allow for saving off flows prior to the pod getting a SIGKILL.


juanluisvaladas:
/retest
/lgtm
Until we have ovs as a system daemon, seems like a reasonable thing to do.

danwinship:
Do we have some reason/evidence for thinking that 10s is not enough time for it to save the flows and 45s is?

danwinship:
> Yes. For instance here

Thank you. PLEASE put information like this in PRs so that people trying to understand the code in the future can see why we changed things

/lgtm

smarterclayton:
I’m merging this to give us more soak in master on upgrades.  OVN failures are unrelated (and a bug in OVN if so).

PR #515, opened 2020-03-05
Fix quotes in hybrid ovn

pecameron:
PR 500 has a quote problem. This fixes it.


pecameron:
@dcbw PTAL

PR #516, opened 2020-03-06
Bug 1811022: Ensure correct Octavia API version is retrieved

MaysaMacedo:
We are using a Network client to retrieve the octavia API version,
this results in a wrong api version.
This commit fixes the issue by using the Load Balancer client.


dulek:
Should this be a bug so we can backport it to 4.4? Or part of 1808797 for example?

PR #520, opened 2020-03-09
Bug 1811748: Ensure removal of not rendered resources upon CNO recreation

MaysaMacedo:
If the CNO pod is recreated, the resources that are not rendered
anymore are not deleted. Due to The related objects field of the
cluster operator status being wiped out upon cno recreation.

This commit fixes the issue by ensuring the objects on the status
manager are only updated when the current status is filled.


danwinship:
> We use a "-s" on the commit

no, that's just upstream ovn-kubernetes. No openshift/ or kubernetes/ repos do that.

PR #524, opened 2020-03-10
Bug 1807638: Fixes to reliably save/restore flows.

abhat:
This PR adds:
a) a timestamp when adding logs for save/restore so they can be correlated with other logs.
b) removes the `hostPID` setting on the daemonset pods
c) removes the `--monitor` option when starting the `ovs-vswitchd` and `ovsdb-server` processes

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -73,7 +73,7 @@ spec:
 
           # launch OVS
           # Start the ovsdb so that we can prep it before we start the ovs-vswitchd
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random --no-monitor
 
           # Set the flow-restore-wait to true so ovs-vswitchd will wait till flows are restored
           ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-restore-wait=true
@@ -88,7 +88,7 @@ spec:
           fi
 
           # And finally start the ovs-vswitchd now the DB is prepped
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random --no-monitor
 
           /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
           /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null &&
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -24,7 +24,6 @@ spec:
     spec:
       serviceAccountName: sdn #needed to run privileged pods; not used for api access
       hostNetwork: true
-      hostPID: true
       priorityClassName: system-node-critical
       containers:
       - name: openvswitch
@@ -35,7 +34,6 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
-          TIMESTAMP=`date -u "+%Y-%m-%d %H:%M:%S"`
           chown -R openvswitch:openvswitch /var/run/openvswitch
           chown -R openvswitch:openvswitch /etc/openvswitch
 
@@ -57,16 +55,14 @@ spec:
 
           function quit {
               # Save the flows
-              echo "$TIMESTAMP info: Saving flows ..." 2>&1
+              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saving flows ..." 2>&1
               bridges=$(ovs-vsctl -- --real list-br)
               TMPDIR=/var/run/openvswitch /usr/share/openvswitch/scripts/ovs-save save-flows $bridges > /var/run/openvswitch/flows.sh
-              END_TS=`date -u "+%Y-%m-%d %H:%M:%S"`
-              echo "$END_TS info: Saved flows" 2>&1
+              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saved flows" 2>&1
 
               # Don't allow ovs-vswitchd to clear datapath flows on exit
               kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true
               kill $(cat /var/run/openvswitch/ovsdb-server.pid 2>/dev/null) 2>/dev/null || true
-              #kill $(jobs -p)
               exit 0
           }
           trap quit SIGTERM
@@ -77,7 +73,7 @@ spec:
 
           # Set the flow-restore-wait to true so ovs-vswitchd will wait till flows are restored
           ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-restore-wait=true
-           
+          
           # Restrict the number of pthreads ovs-vswitchd creates to reduce the
           # amount of RSS it uses on hosts with many cores
           # https://bugzilla.redhat.com/show_bug.cgi?id=1571379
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -57,10 +58,10 @@ spec:
 
           function quit {
               # Save the flows
-              echo "info: Saving flows ..." 2>&1
+              echo "$TIMESTAMP info: Saving flows ..." 2>&1
               bridges=$(ovs-vsctl -- --real list-br)
               TMPDIR=/var/run/openvswitch /usr/share/openvswitch/scripts/ovs-save save-flows $bridges > /var/run/openvswitch/flows.sh
-              echo "info: Saved flows" 2>&1
+              echo "$TIMESTAMP info: Saved flows" 2>&1
```
danwinship: This will print the same timestamp before and after...

abhat: good catch. will update.

abhat:
> Do all the logs use UTC timestamps? Also, ovs in OVNKubernetes should do this too.

Yeah, as far as I can tell. Agreed, I would like to get this working first before making the changes for OVNKubernetes.

dcbw:
@abhat is this just for testing purposes in CI, or for real production?

abhat:
> @abhat is this just for testing purposes in CI, or for real production?

I added this to debug why we were getting bailed out before the `terminationGracePeriodSeconds` expired. But I don't see a reason (unless it's too expensive) to not keep it in prod.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,11 +35,11 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
```
mrunalp: Can we also enable `set -x` ?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -71,11 +71,11 @@ spec:
 
           # launch OVS
           # Start the ovsdb so that we can prep it before we start the ovs-vswitchd
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random --no-monitor
 
           # Set the flow-restore-wait to true so ovs-vswitchd will wait till flows are restored
           ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-restore-wait=true
-          
+
```
dcbw: @abhat whitespace addition?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -86,18 +86,25 @@ spec:
           fi
 
           # And finally start the ovs-vswitchd now the DB is prepped
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random --no-monitor
```
dcbw: @abhat so the only downside to using --no-monitor is if the process segfaults or something, do we have allt he necessary health-checks to kill the container when those processes die? we've got one for the vswitchd I think (openshift-sdn's healthcheck.go) but not for the dbserver. Or it could be the other way around.

bindata/network/openshift-sdn/sdn-ovs.yaml
abhat: Will get rid of it.

bindata/network/openshift-sdn/sdn-ovs.yaml
abhat: Liveness probe here will make sure that kubelet will kill the pod if ovsdb or ovs-vswitchd are not responsive. https://github.com/openshift/cluster-network-operator/blob/9fb694ed7ec901e382585e23b16f1e83ac8a3af6/bindata/network/openshift-sdn/sdn-ovs.yaml#L144. Will that be sufficient?

danwinship:
Are the timestamps *really* useful? They were only added originally to figure out how long the save was taking

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -86,18 +86,25 @@ spec:
...
 
+          /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
+          /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null &&
+          /usr/bin/ovs-vsctl -t 5 --retry show > /dev/null
```
danwinship: If the goal of this is "make sure that the database has started" then there is probably a better way... At least, it seems like we should only need one of these commands?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -86,18 +86,25 @@ spec:
...
+          
           # Load any flows that we saved
-          echo "info: Loading previous flows ..." 2>&1
+          echo "$TIMESTAMP info: Loading previous flows ..." 2>&1
           if [[ -f /var/run/openvswitch/flows.sh ]]; then
-             echo "info: Execute flow script ..." 2>&1
+             echo "$TIMESTAMP info: Adding br0 if it doesn't exist ..." 2>&1
+             /usr/bin/ovs-vsctl --may-exist add-br br0 -- set Bridge br0 fail_mode=secure protocols=OpenFlow13
```
danwinship: I don't like having the startup script have knowledge about what options the openshift-sdn binary sets on `br0`. This can be fixed later, but we ought to leave a bug/issue/card around reminding us about it.

abhat:
> Are the timestamps _really_ useful? They were only added originally to figure out how long the save was taking

I left them in there; it will be useful debug information for the logs. I am OK taking them out too.

bindata/network/openshift-sdn/sdn-ovs.yaml
abhat: Created https://github.com/openshift/cluster-network-operator/issues/555. 
I understand the issue from maintainability perspective. Is there another reason why you don't like this?

bindata/network/openshift-sdn/sdn-ovs.yaml
abhat: The goal is to make sure database has started and the ovs-vswitchd is up so it can take ovs-ofctl commands.

abhat:
> Are the timestamps _really_ useful? They were only added originally to figure out how long the save was taking
They are potentially useful as debug information in the logs. Having said that, I am ok removing them.

bindata/network/openshift-sdn/sdn-ovs.yaml
danwinship: I would think probably `ovs-ctl status` alone should be sufficient for that? You definitely shouldn't need three different commands. I forget exactly what they're all doing in the readiness check, but that may not apply here.

bindata/network/openshift-sdn/sdn-ovs.yaml
danwinship: Just maintainability; if we change the details in openshift-sdn now, we'd have to also remember to change them in sync here

bindata/network/openshift-sdn/sdn-ovs.yaml
abhat: Yes, `ovs-ctl status` will indicate if the daemons are running. The other two commands are useful to check for socket readiness for ovs-vswitchd and ovsdb-server respectively. Given that restore flows only uses `ovs-ofctl`, I can get rid of the `ovs-vsctl` check. I would prefer to keep the `ovs-appctl` check since we have seen a lot of socket errors at start up in the logs before.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,11 +34,11 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
+          TIMESTAMP=`date -u "+%Y-%m-%d %H:%M:%S"`
```
danwinship: `$(foo)` is preferred to `` `foo` `` because backticks have weird quoting semantics and `$()` is saner. (Throughout the patch.)

But also, you're generating a timestamp once at the start of the function, and then reprinting that same identical timestamp in front of multiple different lines, which doesn't make much sense. You should either only print the timestamp once, at the start of the script, or else print a new timestamp each time

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -86,18 +87,24 @@ spec:
           fi
 
           # And finally start the ovs-vswitchd now the DB is prepped
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random
+          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random --no-monitor
 
+          /usr/share/openvswitch/scripts/ovs-ctl status > /dev/null &&
+          /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null
```
danwinship: I see no evidence that this is needed. In particular, `ovs-ctl restart` does not do this

abhat: OK will get rid of this check.

bindata/network/openshift-sdn/sdn-ovs.yaml
dcbw: Yeah I think that's fine.

bindata/network/openshift-sdn/sdn-ovs.yaml
dcbw: @abhat @danwinship should we add a comment here that about that?

smarterclayton:
If GCP OVN is broken we should just skip it.  Or does this PR break GCP OVN (I expect the former)

abhat:
This PR doesn't do anything to the OVN daemonsets. This is purely targeting openshift-sdn daemonsets. The CI config for the e2e-ovn-step-registry was added recently, but the job doesn't have `optional: true` tag on it, so although the job is not `Required` to merge, prow won't merge the PR.

smarterclayton:
I'm running an upgrade test in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-aws-upgrade/23805, so if that passes I'll green button this.

smarterclayton:
The upgrade run had a 1s disruption, but it was during master reboot and needs to be separately investigated.

This is LGTM, going to green button so we know that master has all fixes.

PR #530, opened 2020-03-11
Bug 1812583: Normalize CPU requests on nodes

smarterclayton:
The ovs container uses approximately 100m of node CPU in a reasonable
medium sized workload. Tune the default to be more realistic for medium
clusters. For large clusters, a dynamic scaling policy would be
necessary.

4.4 bug is 1812709


danwinship:
/lgtm

I'm pretty sure the current numbers were not scientifically arrived at so I have no problem with changing them.

PR #532, opened 2020-03-13
OVN: Ensure ovn-monitor-all=true before ovn-controller starts.

dceara:
ovn-controller starts.

"true".

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -56,6 +56,30 @@ spec:
               exit 0
           }
           trap quit SIGTERM
+
+          # Start the ovsdb so that we can prep it before ovn-controller
+          # connects to it. To achieve that we use a temporary private socket,
+          # db-private.sock.
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1808125
+          priv_db_sock=/var/run/openvswitch/db-private.sock
+          /usr/share/openvswitch/scripts/ovs-ctl start \
+            --ovs-user=openvswitch:openvswitch --db-sock=${priv_db_sock} \
+            --no-ovs-vswitchd
+
+          # The OVS DB should also be properly initialized before
+          # ovn-controller connects to it.
+          ovs-vsctl --db=unix:${priv_db_sock} -- init
+
+          # Ensure that ovn-controller will never read a value of
+          # ovn-monitor-all=false followed by an update with
+          # ovn-monitor-all=true.
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1808125
+          ovs-vsctl --no-wait --db=unix:${priv_db_sock} \
+            set Open_vSwitch . external_ids:ovn-monitor-all=true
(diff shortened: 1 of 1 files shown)


dcbw:
> How is the case where ovs or ovn-controller gets killed and restarted handled?

@pecameron the issue is when monitor-all is changed while ovn-controller is running (like by ovnkube --node). Here I believe the PR will ensure that the ovn-controller container waits until it is set before continuing, so that will work for restart/crash too.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -46,6 +46,28 @@ spec:
...
+          # Wait for ovn-monitor-all=true before ovn-controller starts to
+          # avoid a race condition in ovn-controller:
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1808125
+          trap 'kill $(jobs -p); exit 0' TERM
+          retries=0
+          while true; do
+            monitor_all=$(ovs-vsctl --columns _uuid find Open_vSwitch external_ids:ovn-monitor-all="true" 2> /dev/null | wc -l)
+            if [[ "${monitor_all}" -eq 0 ]]; then
+              echo "warning: OVS external_ids:ovn-monitor-all not set yet, waiting 5s ..." 2>&1
+              sleep 5 & wait
```
dcbw: @dceara let's reduce the wait from 5 seconds to 1 or something like that, and increase the retries to 2 minutes.

dceara: @dcbw Done. I'm also unsure about the e2e-gcp-ovn test failures. Scheduling a rerun.

/retest

dceara:
> > How is the case where ovs or ovn-controller gets killed and restarted handled?
> 
> @pecameron the issue is when monitor-all is changed while ovn-controller is running (like by ovnkube --node). Here I believe the PR will ensure that the ovn-controller container waits until it is set before continuing, so that will work for restart/crash too.
> 
> @dceara although, what happens if the ovsdb-server/vswitch container exits (which may wipe the database) and restarts, would there be a race where ovn-controller could connect to the db between ovsdb-server startup and where monitor-all=true is set in the script?

@pecameron @dcbw it’s only a problem if a SB reconnect happens at the same time. But it is indeed still a race. I’ll think more about it to see if this can be worked around without the core OVN fix.

dceara:
> > > How is the case where ovs or ovn-controller gets killed and restarted handled?
> > 
> > 
> > @pecameron the issue is when monitor-all is changed while ovn-controller is running (like by ovnkube --node). Here I believe the PR will ensure that the ovn-controller container waits until it is set before continuing, so that will work for restart/crash too.
> > @dceara although, what happens if the ovsdb-server/vswitch container exits (which may wipe the database) and restarts, would there be a race where ovn-controller could connect to the db between ovsdb-server startup and where monitor-all=true is set in the script?
> 
> @pecameron @dcbw it’s only a problem if a SB reconnect happens at the same time. But it is indeed still a race. I’ll think more about it to see if this can be worked around without the core OVN fix.

@dcbw @pecameron I think with the latest version it should be fine.

Thanks,
Dumitru

PR #533, opened 2020-03-13
Bug 1806854: Populate cluster-wide config maps with content from user-ca-bundle

alexanderConstantinescu:
Problem this solves: today the CNO reconciles configmaps with the label `config.openshift.io/inject-trusted-cabundle: true` by populating them with the content from `openshift-config-managed/trusted-ca-bundle`, whereas we are expected (apperently, I am not sure about this...) to populate them with the content from `openshift-config/user-ca-bundle`. Questions still exists in my mind as to what the expected behaviour should be w.r.t this, should all CMs be populated like this, or only some of them?

If someone can summon/tag a proxy/CA experienced engineer, then please feel free to do so. I have going through this article: https://github.com/openshift/enhancements/blob/master/enhancements/proxy/global-cluster-egress-proxy.md but it's still not clear which CA related configmap should be used to populate everything.

### DIFF
--- a/pkg/controller/configmap_ca_injector/controller.go
+++ b/pkg/controller/configmap_ca_injector/controller.go
@@ -95,8 +95,8 @@ func (r *ReconcileConfigMapInjector) Reconcile(request reconcile.Request) (recon
 
 	trustedCAbundleConfigMap := &corev1.ConfigMap{}
 	trustedCAbundleConfigMapName := types.NamespacedName{
-		Namespace: names.TRUSTED_CA_BUNDLE_CONFIGMAP_NS,
-		Name:      names.TRUSTED_CA_BUNDLE_CONFIGMAP,
+		Namespace: names.ADDL_TRUST_BUNDLE_CONFIGMAP_NS,
+		Name:      names.ADDL_TRUST_BUNDLE_CONFIGMAP,
 	}
 	err := r.client.Get(context.TODO(), trustedCAbundleConfigMapName, trustedCAbundleConfigMap)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


danwinship:
> If someone can summon/tag a proxy/CA experienced engineer, then please feel free to do so.

@danehans wrote the original code...

danehans:
> Problem this solves: today the CNO reconciles configmaps with the label config.openshift.io/inject-trusted-cabundle: true by populating them with the content from openshift-config-managed/trusted-ca-bundle, whereas we are expected (apperently, I am not sure about this...) to populate them with the content from openshift-config/user-ca-bundle.

@alexanderConstantinescu PTAL at the `TrustedCA` API [docs](https://github.com/openshift/api/blob/master/config/v1/types_proxy.go#L44-L69). The CNO CA injection controller populates any ConfigMaps labeled `config.openshift.io/inject-trusted-cabundle: true` with the certs from `openshift-config-managed/trusted-ca-bundle` ConfigMap. Before this can be accomplished, CNO's proxyconfig controller publishes the trusted CA bundle as a ConfigMap (i.e. `openshift-config-managed/trusted-ca-bundle`).

The proxyconfig controller is responsible for managing the `openshift-config-managed/trusted-ca-bundle` ConfigMap. The contents of this ConfigMap are based on `proxy.spec.trustedCA`. If `proxy.spec.trustedCA` is set to a ConfigMap containing valid CA certs (i.e. `openshift-config/user-ca-bundle`), the controller merges these certs with the RHCOS system trust bundle and publishes the combined bundle in `openshift-config-managed/trusted-ca-bundle`. If `proxy.spec.trustedCA` is unset, the controller populates `openshift-config-managed/trusted-ca-bundle` with _only_ the RHCOS system trust bundle certs.

CNO's CA injection controller injects the contents of `openshift-config-managed/trusted-ca-bundle` into the injection requestors ConfigMap (i.e. `config.openshift.io/inject-trusted-cabundle: true`). __Note:__ The ConfigMap referenced by `proxy.spec.trustedCA` must be in the `openshift-config` namespace. The product [docs](https://docs.openshift.com/container-platform/4.3/networking/enable-cluster-wide-proxy.html) may be helpful to better understand the cluster-wide egress proxy feature.

alexanderConstantinescu:
Thanks for clarifying @danehans. I just re-tested the bug and it seemed the CNO is doing what you said, so I am closing this PR as firstly: it's wrong, and secondly I am not sure what problem this bug is hitting exactly and I need to spend more time digging that out with the QE.

PR #535, opened 2020-03-16
Bug 1802557: Pass --node-name and --node-ip to openshift-sdn-node

danwinship:
https://github.com/openshift/sdn/pull/120 updates openshift-sdn-node to have `--node-name` and `--node-ip` flags; use them. (This won't pass tests until that merges.)

Fixes problems where openshift-sdn-node auto-detects the wrong IP

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -103,7 +103,10 @@ spec:
           cp -f /opt/cni/bin/openshift-sdn /host/opt/cni/bin/
 
           # Launch the network process
-          exec /usr/bin/openshift-sdn-node --proxy-config=/config/kube-proxy-config.yaml --v=${OPENSHIFT_SDN_LOG_LEVEL:-2}
+          exec /usr/bin/openshift-sdn-node \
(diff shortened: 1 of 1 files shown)


dcbw:
@danwinship 
```
level=error msg="Cluster operator network Degraded is True with ApplyOperatorConfig: Error while updating operator configuration: could not apply (apps/v1, Kind=DaemonSet) openshift-sdn/sdn: could not create (apps/v1, Kind=DaemonSet) openshift-sdn/sdn: DaemonSet.apps \"sdn\" is invalid: spec.template.spec.containers[0].env[4].valueFrom.fieldRef.fieldPath: Invalid value: \"spec.hostIP\": error converting fieldPath: field label not supported: spec.hostIP"
```

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -162,6 +165,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: K8S_NODE_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.hostIP
```
dcbw: @danwinship `status.hostIP` maybe?

PR #539, opened 2020-03-17
Bug 1814098: [release-4.3] OVN: Ensure ovn-monitor-all=true before ovn-controller starts.

dcbw:
Cherry-pick of https://github.com/openshift/cluster-network-operator/pull/532 and https://github.com/openshift/cluster-network-operator/pull/541 squashed together.

---
ovn-controller starts.

"true".

@dceara

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -59,6 +59,33 @@ spec:
               exit 0
           }
           trap quit SIGTERM
+
+          # Start the ovsdb so that we can prep it before ovn-controller
+          # connects to it. To achieve that we use a temporary private socket,
+          # db-private.sock.
+          # https://bugzilla.redhat.com/show_bug.cgi?id=1808125
+          # When called with '--db-sock', ovs-ctl doesn't point
+          # 'ovs-vsctl -- init' to the correct socket path so we have to
+          # ignore the error returned by ovs-ctl.
+          priv_db_sock=/var/run/openvswitch/db-private.sock
+          /usr/share/openvswitch/scripts/ovs-ctl start \
+            --ovs-user=openvswitch:openvswitch --db-sock=${priv_db_sock} \
+            --no-ovs-vswitchd || true
+
+          # The OVS DB should also be properly initialized before
+          # ovn-controller connects to it.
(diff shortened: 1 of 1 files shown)


danwinship:
> Until the bug is fixed in OVS/OVN code, this is a workaround delaying
> start up of ovn-controller until OVS is up and ovn-monitor-all is set to
> "true".

Doesn't the monitor-all fix really only matter for clusters larger than the largest supported OpenShift 4.3 ovn-kubernetes cluster? Would it be simpler to just revert setting monitor-all until the bug is fixed?

dcbw:
> Doesn't the monitor-all fix really only matter for clusters larger than the largest supported OpenShift 4.3 ovn-kubernetes cluster? Would it be simpler to just revert setting monitor-all until the bug is fixed?

@danwinship the fix reduces load on ovsdb-server southbound and allows faster and larger scale. I think it's something we want for > 100 nodes. @dceara could provide more information there.

dceara:
> > Doesn't the monitor-all fix really only matter for clusters larger than the largest supported OpenShift 4.3 ovn-kubernetes cluster? Would it be simpler to just revert setting monitor-all until the bug is fixed?
> 
> @danwinship the fix reduces load on ovsdb-server southbound and allows faster and larger scale. I think it's something we want for > 100 nodes. @dceara could provide more information there.

@dcbw, @danwinship, in our OVN-only scale testing, we brought up 300 (virtual) nodes and configured a simplified but similar topology to what ovn-kubernetes would configure. Nodes are spawned and provisioned in sequence and we noticed:
- with `ovn-monitor-all=false`: starting with iteration ~170, ovsdb-server for Southbound was becoming the bottleneck because of the way it processes conditional monitoring updates from ovn-controller. From that point on, the time to bring up a node started increasing linearly.
- with `ovn-monitor-all=true`: the time to bring up 300 nodes stayed constant as ovsdb-server doesn't need to filter the database contents for each jsonrpc connection from ovn-controllers, every time a new node is spawned. This enables ovsdb-server to serve pre-cached jsonrpc updates to ovn-controllers during one iteration therefore using less CPU to provision a new node.

Another side effect is that, at least in our testing, the amount of network traffic between ovn-controllers and ovsdb-server was reduced because there's no need anymore for each ovn-controller to send large jsonrpc conditional monitoring update messages every time a new node is spawned.

PR #541, opened 2020-03-18
OVN: Fix initial OVS DB preparation.

dceara:
When 'ovs-ctl start' is called with --db-sock=<value>, ovs-ctl doesn't
properly point its invocation of 'ovs-vsctl' to use the non-default unix
socket passed as argument. The ovsdb-server instance is however started
properly and the error returned can be safely ignored.

Fixes: cae57b4a547d ("OVN: Ensure ovn-monitor-all=true before ovn-controller starts.")


dceara:
@dcbw, please take a look, this fixes the issue introduced by https://github.com/openshift/cluster-network-operator/pull/532.

Thanks!

PR #542, opened 2020-03-18
Bug 1814099: [release-4.4] OVN: Ensure ovn-monitor-all=true before ovn-controller starts.

dcbw:
Cherry-pick of https://github.com/openshift/cluster-network-operator/pull/532 and https://github.com/openshift/cluster-network-operator/pull/541 squashed together.

---
ovn-controller starts.

"true".

@dceara

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -56,6 +56,33 @@ spec:
               exit 0
           }
           trap quit SIGTERM
+
+          # Start the ovsdb so that we can prep it before ovn-controller
+          # connects to it. To achieve that we use a temporary private socket,
+          # db-private.sock.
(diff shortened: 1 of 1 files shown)


danwinship:
So again, given that this is ugly, and the first attempt introduced another bug, and we don't have any customers who actually *need* monitor-all, and we expect to have a fix for the underlying problem soon, wouldn't it be better to just disable monitor-all for now?

dcbw:
@danwinship it's a scale fix which impacts our further scale testing. Are you perhaps suggesting we should skip the backports and just do this for 4.4/4.5 or even just 4.5+?

danwinship:
OK, if the bug blocks our own scale testing then that's a reason to work around it.

dcbw:
ovslib patch to fix this finally got acked; so this is no longer needed and we'll continue with the "real" solution by updating OVN with the ovslib change.

PR #544, opened 2020-03-18
Bug 1802481: Don't override containernetworking binaries in SDN and Kuryr

alexanderConstantinescu:
Containernetworking binaries are compiled for RHEL7 and RHEL8 [here](https://github.com/openshift/containernetworking-plugins/blob/master/Dockerfile). This was done during the effort last year when fixing this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1725832

That solution was applied to multus only and overlooked the fact that we have initcontainers in SDN and Kuryr that still copy the binaries blindly to the host without checking which OS version we're running on. This means that when we start the network plugin we can still end up in a situation where the binary cannot be executed without SIGSEGV-ing directly. 

Seeing as how Multus always runs, no matter the networking plugin used: just don't do anything and let Multus be the sole responsible, as such we should avoid stepping on each other toes. 

Assigning someone from SDN, Multus and Kuryr to review this PR. 

This will need back-ports to 4.2 (possibly even 4.1? @knobunc) 

/assign @vrutkovs @dougbtv @dcbw

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -76,11 +76,12 @@ data:
     fi
 
     cp -rf ${sourcedir}* $DESTINATION_DIRECTORY
+
     if [ $? -eq 0 ]; then
-        echo "Successfully copied files in ${sourcedir} to $DESTINATION_DIRECTORY"
+      echo "Successfully copied files in ${sourcedir} to $DESTINATION_DIRECTORY"
     else
-        echo "Failed to copy files in ${sourcedir} to $DESTINATION_DIRECTORY"
-        exit 1
+      echo "Failed to copy files in ${sourcedir} to $DESTINATION_DIRECTORY"
+      exit 1
     fi
 ---
 kind: DaemonSet
@@ -350,7 +351,7 @@ spec:
             path: {{ .MultusCNIConfDir }}
         - name: cnibin
           hostPath:
-            path: /var/lib/cni/bin
+            path: {{ .CNIBinDir }}
         - name: os-release
           hostPath:
             path: /etc/os-release
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -37,21 +37,6 @@ spec:
       hostNetwork: true
       hostPID: true
       priorityClassName: system-node-critical
-      initContainers:
-      - name: install-cni-plugins
-        image: {{ .CNIPluginsImage }}
-        command:
-        - /bin/bash
-        - -c
-        - |
-          #!/bin/bash
-          set -ex
-          cp -f /usr/src/plugins/bin/{loopback,host-local} /host/opt/cni/bin
-        volumeMounts:
-        - mountPath: /host/opt/cni/bin
-          name: host-cni-bin
-        securityContext:
-          privileged: true
       containers:
       - name: sdn
         image: {{.SDNImage}}
--- a/bindata/network/kuryr/005-daemon.yaml
+++ b/bindata/network/kuryr/005-daemon.yaml
@@ -24,21 +24,6 @@ spec:
       hostNetwork: true
       serviceAccountName: kuryr
       priorityClassName: system-node-critical
-      initContainers:
-      - name: install-cni-plugins
-        image: {{ .CNIPluginsImage }}
-        command:
-        - /bin/bash
-        - -c
-        - |
-          #!/bin/bash
-          set -ex
-          cp -f /usr/src/plugins/bin/loopback /host-cni-bin/
-        volumeMounts:
-        - name: bin
-          mountPath: /host-cni-bin
-        securityContext:
-          privileged: true
       containers:
       - name: kuryr-cni
         image: {{ .DaemonImage }}
--- a/pkg/network/multus.go
+++ b/pkg/network/multus.go
@@ -59,6 +59,7 @@ func renderMultusConfig(manifestDir, defaultNetworkType string, useDHCP bool) ([
 	data.Data["MultusCNIConfDir"] = MultusCNIConfDir
 	data.Data["SystemCNIConfDir"] = SystemCNIConfDir
 	data.Data["DefaultNetworkType"] = defaultNetworkType
+	data.Data["CNIBinDir"] = CNIBinDir
 
 	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network/multus"), &data)
 	if err != nil {


dcbw:
> Seeing as how Multus always runs, no matter the networking plugin used: just don't do anything and let Multus be the sole responsible, as such we should avoid stepping on each other toes.

@alexanderConstantinescu except it doesn't... you can turn Multus off via the CNO at cluster install time.

dcbw:
Another approach is to stop copying binaries with Multus entirely, and rely on the host itself having containernetwork-plugins RPMs in the right place (which would need copying from /usr/libexec/cni to wherever crio wants them)

alexanderConstantinescu:
> Another approach is to stop copying binaries with Multus entirely, and rely on the host itself having containernetwork-plugins RPMs in the right place (which would need copying from /usr/libexec/cni to wherever crio wants them)

Yeah, that had been on the table for a discussion (see: [here](https://bugzilla.redhat.com/show_bug.cgi?id=1725832#50) and [here](https://bugzilla.redhat.com/show_bug.cgi?id=1725832#c54)). @cgwalters : would that be something you guys could do as to ease this process? As you can see this RHEL7 vs RHEL8 issue came back to life.

danwinship:
Maybe we could still deploy the CNI-installing container even if multus was disabled...

(There was an idea a while back about forking multus and making it "the CNO CNI plugin", and having it always run as a wrapper around the default network plugin, even when `disableMultiNetwork` was set. (It just wouldn't do multiplexing in that case.). And then moving the anti-169.254.169.254 hack to there. And maybe some other stuff...)

dulek:
This seems to be fine with Kuryr if multus already copies the binaries.

alexanderConstantinescu:
> This seems to be fine with Kuryr if multus already copies the binaries.

I guess this is an "inverted race-condition" of sorts: if multus is the last one to write these binaries to the host then everything should work fine, if not: we risk having them segfaulting because the underlying OS is uncompatible. Maybe Kuryr is just a lot faster than multus :)

alexanderConstantinescu:
> d the default network plugin, even when `disableMultiNetwork` was set. (It just wouldn't do multiplexing in that cas

I could implement something like that, i.e: some k8s daemonset that always runs no matter the network plugin used and does that. I guess "anti-169.254.169.254" refers to the "block cloud provider metadata IP" which we have [here](https://github.com/openshift/sdn/blob/master/cmd/sdn-cni-plugin/openshift-sdn_linux.go#L130) and [here](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/cni/helper_linux.go#L242)? Anyways, that would mean more PRs across multiple repos, do we want that considering back-ports carried etc?

danwinship:
oh, that would not be a backport-able fix, that would only be for the future

alexanderConstantinescu:
> oh, that would not be a backport-able fix, that would only be for the future

Well, this is going to need a back-port all the way back to 4.1, I believe :man_shrugging: 

I am working on something which would do what we mentioned but not apply the "block cloud provider metadata", and would thus contain the fix to only this repo. I need to test and will push an update once I validate that it works.

manifests/0000_70_cluster-network-operator_05_daemonset.yaml
```diff
@@ -0,0 +1,117 @@
...
+          readOnly: true
+        env:
+        - name: RHEL7_SOURCE_DIRECTORY
+          value: "/usr/src/whereabouts/rhel7/bin/"
+        - name: RHEL8_SOURCE_DIRECTORY
+          value: "/usr/src/whereabouts/rhel8/bin/"
+        - name: DEFAULT_SOURCE_DIRECTORY
+          value: "/usr/src/whereabouts/bin/"
+      - name: cni-plugins-bin-copy
+        image: "quay.io/openshift/origin-container-networking-plugins:4.3"
```
alexanderConstantinescu: @danwinship : this is specifically what I am wondering. What will happen here when we deploy the CNO?

What I've done is create a daemonset whos intention is to run before the network plugin. I am however unsure of what binds these `image` fields to whatever will be deployed at "runtime". It's obvious that these fields are overwritten at some point in CI, but I have no clue of when and how? Am I missing some configuration to make this work?

danwinship: ah, if you want to add a new image there are a bunch of steps you have to go through.

but we shouldn't need a new image here; the existing CNI plugins image already has what we need; you just need a different DaemonSet to run it in the case where multus is disabled, right?

alexanderConstantinescu: > but we shouldn't need a new image here; the existing CNI plugins image already has what we need

That's what I want to do, just re-use those images (which are already defined in the CNO template [here](https://github.com/openshift/cluster-network-operator)) but here too. 

Basically what I want is to have a daemonset that runs before the CNO starts deploying any network plugin. This daemonset just copies these binaries on to every host in the cluster. But using it, we can be certain that it is the only thing that performs this operation and equally importantly we only have one version and location of the script that perform this (which now is defined in `manifests/0000_70_cluster-network-operator_03_configmap.yaml`).

This idea is thus:

1) Deploy this daemonset
2) Deploy the CNO
3) CNO deploys the network plugin(s)

This order also ensures that the binaries are there when the network plugin starts and requires them to start setting up the network.

danwinship: ah, sorry, yes, as long as you refer to the image by exactly the same name as it appears in `manifests/image-references`, then CI and ART will do the appropriate substitutions when building release images.

alexanderConstantinescu: Yeah, this seems to have built and deployed fine.

dcbw:
@alexanderConstantinescu needs another rebase, but otherwise seems OK to me?

alexanderConstantinescu:
@dcbw , rebase has been done. PTAL again

alexanderConstantinescu:
So this patch is ready to go now, I have fixed all the pending build issues. 

PTAL @dcbw @danwinship @dougbtv @vrutkovs

alexanderConstantinescu:
> Blah. Sorry, I don't think this is actually going to work, reliably. Multus and openshift-sdn both depend on the fact that their `initContainer` runs and installs the CNI binaries they depend on _before_ their main `container` runs. That's not guaranteed when the CNI copying is done by a different daemonset.
> 
> Although, openshift-sdn can't become fully ready until multus is ready. So if we solve the problem for multus, we solve it for openshift-sdn too.
> 
> So, I think the fix is that rather than moving all the initialization stuff out of the multus daemonset, we change it instead so that we just alway runs the multus daemonset, but with `{{ .if ... }}` conditionals added so we don't actually run the multus `entrypoint.sh` when multus is supposed to be disabled.
> 
> (Maybe it makes sense to keep the new `openshift-network-operator/network-plugin-install` container name though... although if we're using templates you'll have to move it to be installed at runtime via `bindata/` stuff rather than in `manifests/`.)

The problem I have with that is that things won't be explicit anymore. We'll have "multus" being named "network-plugin-install" and in some cases just doing "network plugin installation stuff"  and in other cases "doing multus stuff too".

That's why I placed all of this in `manifests/`, precisely so that it's created before the CNO itself. What I could do, to _guarantee_ that these binaries exists on the nodes before anything else gets done, is to run a check in the CNO at start-up time and confirm that this "network-plugin-install" daemonset is running 100 %, before rendering any resources. How about that?

danwinship:
> The problem I have with that is that things won't be explicit anymore. We'll have "multus" being named "network-plugin-install" and in some cases just doing "network plugin installation stuff" and in other cases "doing multus stuff too".

I see that as a feature. There's no reason to present Multus to the end user as being its own separate thing. OCP optionally supports multiple network attachments. CNO is one part of the infrastructure of this, CNI is another part, and Multus is a third, but from an OpenShift user's point of view, it's all just "the network".

(I would really like to fork multus and patch it locally to not identify itself as multus, since putting the word "multus" into log/error messages causes people to think that pod startup problems are multus's fault even when they're not.)

> What I could do, to _guarantee_ that these binaries exists on the nodes before anything else gets done, is to run a check in the CNO at start-up time and confirm that this "network-plugin-install" daemonset is running 100 %, before rendering any resources. How about that?

That works for the initial 3 masters at install time, but once we start adding worker nodes at run time, there's no guarantee that the network-plugin-install daemonset will deploy to the node before the multus and openshift-sdn ones.

> so maybe the right thing to do is to submit a patch to the CVO waiting for objects to actually be created and running according to the order provided?

No, the current behavior is what was intended. You're only ordering the *creation* of objects, not the "readiness" of those objects. But that's enough to let you make sure that a namespace is created before its contents, or that a CRD is created before any objects of that CRD type, or that a ConfigMap is created before a Pod that mounts it, etc.

alexanderConstantinescu:
> > The problem I have with that is that things won't be explicit anymore. We'll have "multus" being named "network-plugin-install" and in some cases just doing "network plugin installation stuff" and in other cases "doing multus stuff too".
> 
> I see that as a feature. There's no reason to present Multus to the end user as being its own separate thing. OCP optionally supports multiple network attachments. CNO is one part of the infrastructure of this, CNI is another part, and Multus is a third, but from an OpenShift user's point of view, it's all just "the network".
> 
> (I would really like to fork multus and patch it locally to not identify itself as multus, since putting the word "multus" into log/error messages causes people to think that pod startup problems are multus's fault even when they're not.)
> 
> > What I could do, to _guarantee_ that these binaries exists on the nodes before anything else gets done, is to run a check in the CNO at start-up time and confirm that this "network-plugin-install" daemonset is running 100 %, before rendering any resources. How about that?
> 
> That works for the initial 3 masters at install time, but once we start adding worker nodes at run time, there's no guarantee that the network-plugin-install daemonset will deploy to the node before the multus and openshift-sdn ones.
> 
> > so maybe the right thing to do is to submit a patch to the CVO waiting for objects to actually be created and running according to the order provided?
> 
> No, the current behavior is what was intended. You're only ordering the _creation_ of objects, not the "readiness" of those objects. But that's enough to let you make sure that a namespace is created before its contents, or that a CRD is created before any objects of that CRD type, or that a ConfigMap is created before a Pod that mounts it, etc.

All very reasonable and logical arguments :) I will update this PR

dougbtv:
I'm OK with Dan's suggestions and OK with it going into a `network-plugin-install` daemonset.

Additionally, regarding the output from Multus, @danwinship -- I agree wholeheartedly, just having the name "Multus" there doesn't mean it's from Multus, it's often a case of "kill the messenger" -- so I think we can solve this without a fork. We can make that log output configurable. I've created an issue to track it here @ https://github.com/intel/multus-cni/issues/472

alexanderConstantinescu:
Eureka! 

So, in my refactor of multus I started realizing the shape of the beast that I was creating with this patch. Keeping in mind that this will need to get back-ported to at the very least 4.2, I wanted to keep this patch sufficiently small to ease future back-ports. As

dulek:
@alexanderConstantinescu: Is this thing ready? Should we try it out with Kuryr or are new changes expected?

alexanderConstantinescu:
> @alexanderConstantinescu: Is this thing ready? Should we try it out with Kuryr or are new changes expected?

This is ready. I might have to address some further comments later on, but this is good to go. Let me know how the Kuryr CI runs

dulek:
/hold

Okay, please let me use hold until we have it tested. Unfortunately it might need to wait until next week.

alexanderConstantinescu:
> /hold
> 
> Okay, please let me use hold until we have it tested. Unfortunately it might need to wait until next week.

Sure, no problem. I don't expect any reviews today anyways, so next week is fine.

dulek:
@alexanderConstantinescu: Seems like Kuryr is fairly okay with this.

squeed:
I'm surprised we're seeing issues. The multus binaries were special because they need to bind to the system openssl. We built all the other CNI binaries statically.

alexanderConstantinescu:
> @alexanderConstantinescu: Seems like Kuryr is fairly okay with this.

Only "fairly" ok? Meaning something broke but not necessarily due to this patch?

> I'm surprised we're seeing issues. The multus binaries were special because they need to bind to the system openssl. We built all the other CNI binaries statically.

I can't completely say. I spent hours trying to figure out why the binary could not be executed, even strace just said "screw this":
```
strace -ff ./loopback_node0
execve("./loopback_node0", ["./loopback_node0"], 0x7ffd9766ae08 /* 64 vars */) = 0
arch_prctl(ARCH_SET_FS, 0x62fdf0)       = 0
--- SIGSEGV {si_signo=SIGSEGV, si_code=SI_KERNEL, si_addr=NULL} ---
+++ killed by SIGSEGV (core dumped) +++
Segmentation fault (core dumped)
``` 

And `dlv` returned this stack-trace:

```
dlv core ../loopback_node2 core.loopback_node0.1000.abaf9291b29f4a22b6e540f37b0da98c.253386.1584468315000000000000
Type 'help' for list of commands.
(dlv) stack -full
0  0x000000000044f9e8 in runtime.vdsoParseSymbols
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/vdso_linux.go:246
       info = (*runtime.vdsoInfo)(0x7ffedefc71e0)
       version = -1
1  0x000000000044fd8c in runtime.vdsoauxv
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/vdso_linux.go:279
       tag = 33
       val = 140732639649792
       info = runtime.vdsoInfo {valid: true, loadAddr: 140732639649792, loadOffset: 140732639649792,...+8 more}
2  0x000000000042996d in runtime.sysauxv
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/os_linux.go:259
       auxv = (unreadable read out of bounds)
       ~r1 = (unreadable empty OP stack)
       i = (unreadable could not find loclist entry at 0x32dd6 for address 0x42996c)
       val = (unreadable could not find loclist entry at 0x32e6a for address 0x42996c)
3  0x0000000000429737 in runtime.sysargs
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/os_linux.go:202
       argc = 1
       argv = (**uint8)(0x7ffedefc7738)
       n = (unreadable could not find loclist entry at 0x32bb7 for address 0x429736)
       auxv = *(unreadable read out of bounds)
4  0x0000000000439cf9 in runtime.args
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/runtime1.go:63
       c = 1
       v = (**uint8)(0x7ffedefc7738)
5  0x000000000045375b in runtime.rt0_go
   at /opt/rh/go-toolset-1.13/root/usr/lib/go-toolset-1.13-golang/src/runtime/asm_amd64.s:212
```

Which points to issues with the go-runtime (`vdso_linux.go`) which is quite precisely the problem you debugged last year.  

Multus (which is copying the right binary depending on the underlying OS: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus/multus.yaml#L136) was racing with the SDN (copying only the RHEL7 binary, https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/sdn.yaml#L42 ).

Then concerning the static linkage you mentioned, I don't know exactly.

squeed:
Argh, vdso. Bummer.

Dockerfile
```diff
@@ -9,6 +10,7 @@ RUN hack/build-go.sh; \
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /tmp/build/cluster-network-operator /usr/bin/
 COPY --from=builder /tmp/build/cluster-network-renderer /usr/bin/
+COPY --from=cli /usr/bin/oc /usr/bin/
```
squeed, line 13: Do we actually need `oc`?

squeed: disregard

manifests/0000_70_cluster-network-operator_04_daemonset.yaml
```diff
@@ -0,0 +1,157 @@
+kind: DaemonSet
```
squeed, line 1: This shouldn't be in manifests. It should be installed by the CNO, not the CVO.

manifests/0000_70_cluster-network-operator_03_configmap.yaml
```diff
@@ -0,0 +1,82 @@
+apiVersion: v1
```
squeed, line 1: This should be in `bindata`, not `manifests`.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -37,21 +37,6 @@ spec:
       hostNetwork: true
       hostPID: true
       priorityClassName: system-node-critical
-      initContainers:
```
squeed, line 40: so, wait, how are the sdn binaries installed now?

manifests/0000_70_cluster-network-operator_04_daemonset.yaml
```diff
@@ -0,0 +1,157 @@
...
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/os: linux
+      priorityClassName: "system-node-critical"
+      initContainers:
+      - name: multus-bin-copy
+        image: "quay.io/openshift/origin-multus-cni:4.3"
```
squeed, line 29: This needs proper release image substitution, like all other image references.

squeed:
We support disabling multus, so this needs to handle that.

squeed:
I don't like separating installation of plugins from deploying the node-level daemonsets. For example, for openshift-sdn, they're two binaries built at the same time from the same source code. They should be in the same container image, and ideally the same daemonset.

Also, as danwinship mentioned, now we can be in the scary case where openshift-sdn writes its network config file, thus signifying that it has finished installation and the network is up, but the CNI binary isn't on disk! That will result in some seriously confusing error conditions.

alexanderConstantinescu:
> We support disabling multus, so this needs to handle that.

It does. How do you figure that it does not? 

> I don't like separating installation of plugins from deploying the node-level daemonsets. For example, for openshift-sdn, they're two binaries built at the same time from the same source code. They should be in the same container image, and ideally the same daemonset.
> 
> Also, as danwinship mentioned, now we can be in the scary case where openshift-sdn writes its network config file, thus signifying that it has finished installation and the network is up, but the CNI binary isn't on disk! That will result in some seriously confusing error conditions.

It is guaranteed now, @danwinship comment's were not based on the latest changes. I am using node selectors now to enforce that the network-plugin does not run on nodes that are not labeled: ` node.openshift.io/network-plugin-install: ready`. I tested this during runtime too (removing the label) and everything works fine, i.e: the network plugin removes itself from the node.   

Like I mentioned in my previous comments, I don't want to copy that gigantic copy-script to every daemonset for every network plugin that gets deployed (currently multus, Kuryr and openshift-sdn). 

We already shot ourselves in the foot once, by not realizing that binaries were copied in more places than one. I only want one thing, and one thing only, to be copying these binaries, using only one script, and this thing should be doing it independently of the network plugin. Which is what it is now.

manifests/0000_70_cluster-network-operator_04_daemonset.yaml
alexanderConstantinescu: This is proper release image substitution for assets in `manifests`. We can't use template variables here. But CI understands this completely fine.

bindata/network/openshift-sdn/sdn.yaml
alexanderConstantinescu: By `network-plugin-install`

manifests/0000_70_cluster-network-operator_03_configmap.yaml
alexanderConstantinescu: I could place it there, sure. But I did not because I thought it would be clearer if this ran in `openshift-network-operator` namespace as I tried to do something generic and independent of the network plugin used. But you might be right as that would allow us to not copy all binaries but be selective about it

manifests/0000_70_cluster-network-operator_04_daemonset.yaml
squeed: for CVO-deployed objects, but this shouldn't be CVO-managed.

manifests/0000_70_cluster-network-operator_03_configmap.yaml
squeed: even more generically - the CVO should not be responsible for deploying dependent resources. Those are *always* the responsibility of the second-level operator.

squeed:
How do we handle possible version skew between openshift-sdn daemon and binary?

squeed:
> Like I mentioned in my previous comments, I don't want to copy that gigantic copy-script to every daemonset for every network plugin that gets deployed (currently multus, Kuryr and openshift-sdn).

That's what templates are for :-)

>We already shot ourselves in the foot once

We did that by being crazy and breaking ABI compatability between kernels. That's the root cause of any problems here. But we'll never fix that.

I'm still uncomfortable with decoupling the on-node daemon from the CNI binaries. What does that get us?

squeed:
One last concern - by adding another daemonset that pulls down *every* possible CNI container image, that is causing a lot of unnecessary image fetches that are startup-gating. Can you find a way to get rid of that?

In other words, we shouldn't be downloading kuryr and ovn-kubernetes for clusters with openshift-sdn.

manifests/0000_70_cluster-network-operator_03_configmap.yaml
danwinship: It can run in `openshift-network-operator` but still be in `bindata` rather than `manifests`

danwinship:
> I don't like separating installation of plugins from deploying the node-level daemonsets. For example, for openshift-sdn, they're two binaries built at the same time from the same source code. They should be in the same container image, and ideally the same daemonset.

So, first, this PR doesn't move anything to any different image. It only changes who copies things from images to the root filesystem.

Second, it doesn't change the installation of the `openshift-sdn` binary at all, only the `loopback` binary.

danwinship:
> I am using node selectors now to enforce that the network-plugin does not run on nodes that are not labeled: `node.openshift.io/network-plugin-install: ready`.

We can't do that. This will mess up the CNO's tracking of whether things are fully deployed or not, because when a node is missing that label, the DaemonSet will no longer match, so it will no longer even try to deploy to that node, so CNO will report that networking is fully Available even if something has gone wrong and prevented that label from ever getting set on some nodes.

danwinship:
All kuryr and openshift-sdn depend on is `loopback`. OVN-Kubernetes depends on that too but it doesn't bother to actually install it. Well, actually, none of the plugins themselves depend on it, it's *kubelet* that depends on `loopback`. There was [a thread about trying to clean this up on kubernetes-sig-network](https://groups.google.com/d/msg/kubernetes-sig-network/hlwUInyCDHI/A-j72fATAgAJ) last year but then nothing ever happened.

But anyway, maybe we should just install the loopback binary directly into machine-os-content (ie, the root filesystem image). (Or maybe all of the default CNI binaries?) Then this PR would only need to deal with the various multus-only extra binaries.

squeed:
So, I chatted with Alexander offline and realized where I had gone wrong.

However, we decided the best way forward is just for the mco to add a systemd dropin to `kubelet.service` (or maybe `crio.service`) that copies the loopback plugin from the podman `CNI_PATH` to crio's `CNI_PATH`.

cgwalters:
> However, we decided the best way forward is just for the mco to add a systemd dropin to kubelet.service (or maybe crio.service) that copies the loopback plugin from the podman CNI_PATH to crio's CNI_PATH.

OK I guess, but I think it'd be even cleaner to have crio support an array for CNI_PATH like `CNI_PATH=/run/openshift/cni/bin:/usr/lib/cni/bin`, then no copying is necessary.

danwinship:
yeah, it does actually. that was noticed shortly after

alexanderConstantinescu:
Filed: https://github.com/openshift/machine-config-operator/pull/1630 and tagged you all in it.

I will update this PR once that gets in.

alexanderConstantinescu:
I've updated the PR with the code needed for when openshift/machine-config-operator#1630 gets in. 

PTAL while keeping in mind that that patch goes in with this one.

dulek:
@alexanderConstantinescu: Same drill, can you ping me once this (and MCO piece) is ready to get tested with Kuryr? I understand this is a new approach that we better check and I should have an env ready to test this tomorrow.

alexanderConstantinescu:
https://github.com/openshift/machine-config-operator/pull/1630

is in, so 

/retest

@dulek could you please retry the Kuryr CI?

@danwinship & @squeed  PTAL again

bindata/network/multus/multus.yaml
```diff
@@ -75,12 +77,17 @@ data:
       sourcedir=$DEFAULT_SOURCE_DIRECTORY
     fi
 
-    cp -rf ${sourcedir}* $DESTINATION_DIRECTORY
+    if [ ! -z "$IGNORE_FILES" ]; then
+      cp -rf ${sourcedir}!($IGNORE_FILES) $DESTINATION_DIRECTORY
```
danwinship, line 81: I don't think we especially need to worry about *not* copying some files (as long as we're copying the right versions of them).

alexanderConstantinescu: Yeah, I did that to keep consistency between cluster "with multus" vs. "without", as mentioned a couple of comments above. But I'll remove it

dulek:
> [openshift/machine-config-operator#1630](https://github.com/openshift/machine-config-operator/pull/1630)
> 
> is in, so
> 
> /retest
> 
> @dulek could you please retry the Kuryr CI?
> 
> @danwinship & @squeed PTAL again

Alright, this seems to work on Kuryr deployment.

alexanderConstantinescu:
> Alright, this seems to work on Kuryr deployment.

Please remove the hold in that case.

/retest

dulek:
/hold cancel

Whoops, forgot, sorry about it!

PR #545, opened 2020-03-18
KIND deployments for CNO fails.

Billy99:
KIND deployments for CNO are failing due to a requirement for existence of
"cluster-config-v1" configMap introduced by merge c594414 on 03/09/2020. This
change adds a minimal "cluster-config-v1" configMap to KIND deployments.

Signed-off-by: Billy McFall <22157057+Billy99@users.noreply.github.com>


Billy99:
@trozet PTAL

trozet:
@danwinship PTAL

alexanderConstantinescu:
Instead of hard-coding `NUM_MASTER_NODES=1` maybe you could run a `grep -c "role\: control-plane" hack/kind.yaml` and remove the commented lines in that file? 

Besides that, thanks for fixing this!

Billy99:
> Instead of hard-coding `NUM_MASTER_NODES=1` maybe you could run a `grep -c "role\: control-plane" hack/kind.yaml` and remove the commented lines in that file?
> 
> Besides that, thanks for fixing this!

Thanks for the suggestion, Will do.

PR #546, opened 2020-03-19
Enable KIND debugging in ovn-kind-cno

nerdalert:
- Added a KIND debugging ENV var that takes an int
which corresponds to the log verbosity level in KIND.
- The default loglevel is set to 0 which aligns with
the default loglevel in upstream KIND.

Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>


trozet:
@danwinship 
/lgtm

danwinship:
um, I think
/skip
is the command that makes the bot notice that its test runs went awol

danwinship:
@nerdalert or you can try closing and then reopening the PR...

PR #551, opened 2020-03-20
Bug 1815613: [4.4] Update Progressing message for non-critical components

danwinship:
Backport of #474 and #510

/cc @dcbw


danwinship:
I think I confused the bots by filing this against the wrong branch and then quickly switching it...
/close

PR #552, opened 2020-03-20
Deploy the network metrics daemon as part of the CNO.

fedepaol:
This PR is to build the [network metrics daemon](https://github.com/openshift/network-metrics-daemon/) as part of CNO

### DIFF
--- a/bindata/network/network-metrics/000-rbac.yaml
+++ b/bindata/network/network-metrics/000-rbac.yaml
@@ -1,0 +1,38 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: metrics-daemon-sa
+  namespace: openshift-multus
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: metrics-daemon-role
+rules:
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get", "watch", "list"]
+  - apiGroups: ["k8s.cni.cncf.io"]
+    resources: ["network-attachment-definitions"]
+    verbs: ["get", "watch", "list"]
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
+    verbs: ["create"]
+  - apiGroups: ["authorization.k8s.io"]
+    resources: ["subjectaccessreviews"]
+    verbs: ["create"]
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: metrics-daemon-sa-rolebinding
+subjects:
+  - kind: ServiceAccount
+    name: metrics-daemon-sa
+    apiGroup: ""
+    namespace: openshift-multus
+roleRef:
+  kind: ClusterRole
+  name: metrics-daemon-role
+  apiGroup: rbac.authorization.k8s.io
--- a/bindata/network/network-metrics/002-prometheus.yaml
+++ b/bindata/network/network-metrics/002-prometheus.yaml
@@ -1,0 +1,77 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    name: monitor-network
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: monitor-network
+  namespace: openshift-multus
+spec:
+  endpoints:
+    - interval: 10s
+      port: metrics
+      honorLabels: true
+      bearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token"
+      scheme: "https"
+      tlsConfig:
+        caFile: "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
+        serverName: "network-metrics-service.openshift-multus.svc"
+  selector:
+    matchLabels:
+      service: network-metrics-service
+  namespaceSelector:
+    matchNames:
+      - openshift-multus
+---
+apiVersion: v1
+kind: Service
+metadata:
+  annotations:
+    prometheus.io/scrape: "true"
+    service.alpha.openshift.io/serving-cert-secret-name: metrics-daemon-secret
+  labels:
+    service: network-metrics-service
+  name: network-metrics-service
+  namespace: openshift-multus
+spec:
+  selector:
+    app: network-metrics-daemon
+  clusterIP: None
+  ports:
+    - name: metrics
+      port: 8443
+      targetPort: https
+  type: ClusterIP
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-multus
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-multus
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: prometheus-k8s
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-k8s
+    namespace: openshift-monitoring
--- a/manifests/0000_70_cluster-network-operator_03_deployment.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment.yaml
@@ -49,6 +49,8 @@ spec:
           value: "quay.io/openshift/origin-kuryr-cni:4.3"
         - name: KURYR_CONTROLLER_IMAGE
           value: "quay.io/openshift/origin-kuryr-controller:4.3"
+        - name: NETWORK_METRICS_DAEMON_IMAGE
+          value: "quay.io/openshift/origin-network-metrics-daemon:4.5"
         - name: POD_NAME
           valueFrom:
             fieldRef:
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -50,3 +50,8 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-kuryr-controller:4.3
+  - name: network-metrics-daemon
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-network-metrics-daemon:4.5
+
--- a/bindata/network/network-metrics/001-daemonset.yaml
+++ b/bindata/network/network-metrics/001-daemonset.yaml
@@ -1,0 +1,71 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: network-metrics-daemon
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the network metrics daemon on each node
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+    networkoperator.openshift.io/non-critical: ""
+spec:
+  selector:
+    matchLabels:
+      app: network-metrics-daemon
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: network-metrics-daemon
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+      tolerations:
+        - operator: Exists
+      containers:
+        - name: network-metrics-daemon
+          image: {{.NetworkMetricsImage}}
+          command:
+            - /usr/bin/network-metrics
+          args: ["--node-name", "$(NODE_NAME)"]
+          resources:
+            requests:
+              cpu: 10m
+              memory: 100Mi
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+        - name: kube-rbac-proxy
+          image: {{.KubeRBACProxyImage}}
+          args:
+            - --logtostderr
+            - --secure-listen-address=:8443
+            - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
+            - --upstream=http://127.0.0.1:9091/
+            - --tls-private-key-file=/etc/metrics/tls.key
+            - --tls-cert-file=/etc/metrics/tls.crt
+          ports:
+            - containerPort: 8443
+              name: https
+          resources:
+            requests:
+              cpu: 10m
+              memory: 20Mi
+          terminationMessagePolicy: FallbackToLogsOnError
+          volumeMounts:
+            - name: metrics-certs
+              mountPath: /etc/metrics
+              readOnly: True
+      volumes:
+        - name: metrics-certs
+          secret:
+            secretName: metrics-daemon-secret
+      serviceAccountName: metrics-daemon-sa
--- a/pkg/network/multus.go
+++ b/pkg/network/multus.go
@@ -22,12 +22,10 @@ func RenderMultus(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstruct
 		return nil, nil
 	}
 
-	var err error
 	out := []*uns.Unstructured{}
-	objs := []*uns.Unstructured{}
 
 	// enabling Multus always renders the CRD since Multus uses it
-	objs, err = renderAdditionalNetworksCRD(manifestDir)
+	objs, err := renderAdditionalNetworksCRD(manifestDir)
 	if err != nil {
 		return nil, err
 	}
@@ -39,6 +37,13 @@ func RenderMultus(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstruct
 		return nil, err
 	}
 	out = append(out, objs...)
+
+	objs, err = renderNetworkMetricsDaemon(manifestDir)
+	if err != nil {
+		return nil, err
+	}
+	out = append(out, objs...)
+
 	return out, nil
 }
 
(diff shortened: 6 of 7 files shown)


fedepaol:
Also, this requires https://github.com/openshift/release/pull/7877 to get in as the image is not on quay yet.

fedepaol:
@knobunc do you need if there is something else to do here to make the cluster aware of the new image (especially d/s)? 
It's referenced as the `NETWORK_METRICS_DAEMON_IMAGE` env variable but the only place it's set is set to point to quay.

I wonder if there's some black magic involved or I have to configure that somewhere else.

fedepaol:
The daemon is now available on quay, this is good to go.

bindata/network/network-metrics/000-ns.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    name: openshift-network-metrics
+    openshift.io/cluster-monitoring: "true"
+    openshift.io/run-level: "1"
```
juanluisvaladas, line 7: Since this isn't required for the SDN to run and the operator doesn't consider the daemonset critical, we should consider removing this and use an SCC instead (if necessary).

bindata/network/network-metrics/002-daemonset.yaml
```diff
@@ -0,0 +1,74 @@
...
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/worker: ""
+      tolerations:
+        - operator: Exists
+      containers:
+        - name: network-metrics-daemon
```
juanluisvaladas, line 31: Doesn't this container need the pod to have hostNetwork: true?

fedepaol: Nope, at the moment it listens for pods and publish metrics related to the secondary network names

bindata/network/network-metrics/000-ns.yaml
fedepaol: It turned out no special capabilities are needed, removed the annotation.

bindata/network/network-metrics/000-ns.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    name: openshift-network-metrics
+    openshift.io/cluster-monitoring: "true"
+  name: openshift-network-metrics
```
danwinship, line 7: Since this is only needed because of multus, it seems like it should go in the openshift-multus namespace?

bindata/network/network-metrics/001-rbac.yaml
```diff
@@ -0,0 +1,38 @@
...
+    verbs: ["get", "watch", "list"]
+  - apiGroups: ["k8s.cni.cncf.io"]
+    resources: ["network-attachment-definitions"]
+    verbs: ["get", "watch", "list"]
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
+    verbs: ["create"]
+  - apiGroups: ["authorization.k8s.io"]
+    resources: ["subjectaccessreviews"]
+    verbs: ["create"]
```
danwinship, line 24: why does it need tokenreviews and subjectaccessreviews create permission?

bindata/network/network-metrics/002-daemonset.yaml
```diff
@@ -0,0 +1,74 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: network-metrics-daemon
```
danwinship, line 5: The name "network-metrics-daemon" makes this sound much more generic than it really is. If this is really only about providing additional metrics for pods using multus, it should have a name that sounds more like that. (Likewise the https://github.com/openshift/network-metrics-daemon project)

bindata/network/network-metrics/002-daemonset.yaml
```diff
@@ -0,0 +1,74 @@
...
+          imagePullPolicy: Always
+          env:
+            - name: NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
danwinship, line 48: These get read by the network-metrics-daemon process? Stylistically, I think it's better to pass them as command-line flags, rather than having its configuration be a mix of command-line flags and environment variables.

bindata/network/network-metrics/002-daemonset.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - name: kube-rbac-proxy
+          image: {{.KubeRBACProxyImage}}
+          args:
+            - --logtostderr
+            - --secure-listen-address=:8443
+            - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
```
danwinship, line 54: we control the upstream; can't we make it default to that? (or figure it out from the cert or whatever?)

bindata/network/network-metrics/002-daemonset.yaml
```diff
@@ -0,0 +1,74 @@
+---
+kind: DaemonSet
```
danwinship, line 2: Why does this need to run on every node? It seems that it doesn't actually need access to the pods or the kernel for what it's doing. Can't there just be a single copy of it that accumulates the metrics for all nodes?

manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -47,6 +47,10 @@ spec:
           value: "quay.io/openshift/origin-kuryr-cni:4.3"
         - name: KURYR_CONTROLLER_IMAGE
           value: "quay.io/openshift/origin-kuryr-controller:4.3"
+        - name: NETWORK_METRICS_DAEMON_IMAGE
+          value: "quay.io/openshift/origin-network-metrics-daemon:4.5"
+        - name: KUBE_RBAC_PROXY_IMAGE
+          value: "quay.io/openshift/origin-kube-rbac-proxy:4.4"
```
danwinship, line 53: `KUBE_RBAC_PROXY_IMAGE` is already earlier in the file

manifests/image-references
```diff
@@ -46,3 +46,11 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-kuryr-controller:4.3
+  - name: network-metrics-daemon
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-network-metrics-daemon:4.5
+  - name: kube-rbac-proxy
```
danwinship, line 53: (likewise)

pkg/network/network_metrics.go
```diff
@@ -0,0 +1,49 @@
...
+	"path/filepath"
+
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/render"
+	"github.com/pkg/errors"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// RenderMultus generates the manifests of Multus
+func RenderNetworkMetricsDaemon(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstructured, error) {
```
danwinship, line 14: bad comment

pkg/network/network_metrics.go
```diff
@@ -0,0 +1,49 @@
...
+	if *conf.DisableMultiNetwork {
```
danwinship, line 15: rather than having `RenderNetworkMetricsDaemon` check this, you should just have this function be called as a subroutine of `RenderMultus`

bindata/network/network-metrics/003-prometheus.yaml
```diff
@@ -0,0 +1,83 @@
...
+metadata:
+  name: prometheus-k8s
+  namespace: openshift-network-metrics
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+      - endpoints
+      - pods
```
danwinship, line 65: What is this role used for? Why does it need those permissions?

bindata/network/network-metrics/001-rbac.yaml
fedepaol: It's because of the kube-rbac-proxy. We are using it for exposing the metrics endpoint securely, but it also requires those permissions as it performs TokenReviews and SubjectAccessReviews. I followed the same approach as https://github.com/openshift/cluster-network-operator/pull/563 and other operators, putting the rbac-proxy in the pod.

bindata/network/network-metrics/000-ns.yaml
fedepaol: will change.

bindata/network/network-metrics/002-daemonset.yaml
fedepaol: The initial idea was to even have this implemented in the multus admission controller, together with the metrics it is already exposing. 
The discussion that brought to implementing this as an independent daemon is because we will be able to extend it to collect metrics related to dpdk / vfio which need to be collected on the same node.

bindata/network/network-metrics/002-daemonset.yaml
fedepaol: As per my comment above, the general agreement was to use this to collect other metrics too, hence the name.

bindata/network/network-metrics/002-daemonset.yaml
fedepaol: The namespace env var is not read any more and I can probably remove it.
For the other one I agree, but I could not find a way to pass this kind of information as an argument. Using the downward api to expose pod fields as env variable seem to be the suggested approach.

Are you aware of an alternative approach to pass a pod field as an argument of the container?

bindata/network/network-metrics/002-daemonset.yaml
fedepaol: I can check, but is this holding this pr back?

bindata/network/network-metrics/003-prometheus.yaml
fedepaol: It's required by the prometheus pod for service discovery (https://github.com/coreos/prometheus-operator/blob/master/Documentation/rbac.md#prometheus-rbac) 
In this case, we use a `ServiceMonitor` which means that prometheus monitors all the pods backing that service. I think prometheus needs those permission to perform such resolution.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
fedepaol: Will remove it.

manifests/image-references
fedepaol: Will remove it

pkg/network/network_metrics.go
fedepaol: Agree, will change it.

fedepaol:
/hold
Implemented the changes for review but I am still spinning up a cluster for validation.

bindata/network/network-metrics/002-daemonset.yaml
danwinship: Yeah, you have to use the downward API to set the environment variable but then you just use the variable from the command-line rather than from the code. eg,

    args:
      ...
      - --node-name "${NODE_NAME}"

fedepaol: Ah, super simple! Thanks.

bindata/network/network-metrics/003-prometheus_rules.yaml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: openshift-monitoring
+  labels:
+    app: network-metrics
+    prometheus: k8s
+    role: alert-rules
+spec:
+  groups:
+    - name: network_with_name
+      rules:
+        - record: network:container_network_receive_bytes_total
```
lilic, line 16: As per the docs https://prometheus.io/docs/practices/rules/#naming-and-aggregation, the recording rules have the following pattern:
> level:metric:operations

bindata/network/network-metrics/003-prometheus_rules.yaml
```diff
@@ -0,0 +1,31 @@
...
+          expr: (container_network_receive_bytes_total) + on(namespace,pod,interface) group_left(network_name) ( pod_network_name_info )
```
lilic, line 17: Are you sure there are no selectors needed for these? 

Also curious what is pod_network_name_info?

brancz:
> but curious was there an enhancement for this, I am wondering about the background on this and more details and how this ties in with the network-metrics-daemon

Agreed. I would have expected to have seen an enhancement about this, but neither have we been pinged nor can I find an enhancement about this now.

bindata/network/network-metrics/003-prometheus_rules.yaml
fedepaol: > Left a few comments, but curious was there an enhancement for this, I am wondering about the background on this and more details and how this ties in with the network-metrics-daemon. Thanks!
> 
> cc @brancz

There is no enhancement, mostly because of my ignorance of needing one. I shared a doc with the network / multus team, and asked for validation on the monitoring team ml. @brancz replied to it.
The idea is to have a "dumb" (as in always 0 value) `pod_network_name_info` metric to "join" with the real network metrics to have a new metric with the network name.

fedepaol:
> > but curious was there an enhancement for this, I am wondering about the background on this and more details and how this ties in with the network-metrics-daemon
> 
> Agreed. I would have expected to have seen an enhancement about this, but neither have we been pinged nor can I find an enhancement about this now.

Actually, I sent an email to team-monitoring you replied to. Subject "os-monitoring New network metrics - looking for feedback"

bindata/network/network-metrics/003-prometheus_rules.yaml
brancz: Also, is the left hand side here supposed to be the cAdvisor metric? A typical join should also be done with multiplications.

brancz: I'm also missing why we need these recording rules, the join is pretty straight forward, for it to be done in ad-hoc querying. I'm saying that as this will add `# of containers * number of interfaces per container` number of time-series, which is pretty significant.

bindata/network/network-metrics/003-prometheus_rules.yaml
fedepaol: > As per the docs https://prometheus.io/docs/practices/rules/#naming-and-aggregation, the recording rules have the following pattern:
> 
> > level:metric:operations

I _tried_ to follow the pattern, but as I explained there is no real "operations" here.
How would you suggest to change this?

brancz:
I remember now. It would still be nice to have an enhancement so we have the entire history including implementation details written down. (plus it's just the standard procedure :) )

bindata/network/network-metrics/002-daemonset.yaml
brancz: Once other metrics come into play this is probably also a conversation we should have with upstream sig-instrumentation.

fedepaol:
> I remember now. It would still be nice to have an enhancement so we have the entire history including implementation details written down. (plus it's just the standard procedure :) )

Yes, it's mostly because of my ignorance, as it is my first contribution to "core" OpenShift. I tried to look for validation from all the relevant teams, but I missed that.
I am pretty sure it won't happen again.

bindata/network/network-metrics/003-prometheus_rules.yaml
fedepaol: Will try to reply to all in one:
- `pod_network_name_info` is a dumb metric that's needed for joining. The end result would be a metric with the same values as cAdvisor's one but with the network name defined by multus' 
- I can change the operation to use multiplication. Either is fine I think as the daemon's metric will always be 0
- The number is `# of pods * num of interfaces` (which is still significant). I wasn't sure about the overhead vs the ease of use, as it took a bit to put the query together, but probably is straightforward. If that is the case, I can remove the recording rules.

lilic:
> Yes, it's mostly because of my ignorance, as it is my first contribution to "core" OpenShift. I tried to look for validation from all the relevant teams, but I missed that.

No worries, was just curious, as I wanted to read about it to get the background. But will look at the emails, thanks!

bindata/network/network-metrics/000-rbac.yaml
```diff
@@ -0,0 +1,38 @@
+---
```
squeed, line 1: minor nit: If this is part of Multus, then shouldn't it just be in `bindata/network/multus`? It can't be installed separately, right?

(It's in the `openshift-multus` namespace, so the answer to that question is presently no)

fedepaol: It can't, as much as `multus-admission-controller` can't. I tried to stick to the same convention, see https://github.com/openshift/cluster-network-operator/tree/master/bindata/network/multus-admission-controller

It also look more clear what manifests belong to which component.
But if you prefer I can put the manifests under multus.

bindata/network/network-metrics/003-prometheus_rules.yaml
lilic: I think documenting the queries from those recording rules might be a good idea, but agree, if we don't use them anywhere we should not create them.

fedepaol: The idea was to make this more usable / discoverable, but I agree with the overhead. Probably documenting the queries alongside multus documentation is enough and less resource-intensive. Will drop the recording rules.

fedepaol:
@danwinship @squeed @brancz @lilic I _think_ I addressed all the questions / requests. 

Would you mind following up with this so we can move on once 4.6 gates open?
Thanks!

bindata/network/network-metrics/002-prometheus.yaml
```diff
@@ -0,0 +1,85 @@
...
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: monitor-network
+  namespace: openshift-multus
+spec:
+  endpoints:
+    - interval: 10s
+      port: metrics
+      metricRelabelings:
+        - sourceLabels: [pod]
```
brancz, line 16: this looks like you just want to `honorLabels: true`

fedepaol: Ah, exactly. Thanks!

fedepaol: Done, this needs the change on the daemon itself to get merged.

PR #554, opened 2020-03-24
Adding bond-cni binary installation

mmirecki:
I want to add the bond-cni [1] to openshift. We want to use it mainly with sriov. 
It might however be useful not just with sriov, but also on with some cni's available by default with the cluster-network-operator (host-device, macvlan, ...), so it makes sense to ship it in the cluster-network-operator.

[1] https://github.com/intel/bond-cni,  https://github.com/openshift/bond-cni

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -150,6 +150,24 @@ spec:
           value: "/usr/src/plugins/rhel8/bin/"
         - name: DEFAULT_SOURCE_DIRECTORY
           value: "/usr/src/plugins/bin/"
+      - name: bond-cni-plugin
+        image: {{.BondCNIPluginImage}}
+        command: ["/entrypoint/cnibincopy.sh"]
+        volumeMounts:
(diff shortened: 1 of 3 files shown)


danwinship:
> It might however be useful not just with sriov, but also on with some cni's available by default with the cluster-network-operator

We are trying to avoid supporting every possible networking configuration. If there is a specific feature that we need to support which involves using bond-cni with SR-IOV, then we can support it in that case, but we wouldn't want to support it for arbitrary other use cases.

I don't know the details of how SR-IOV is currently deployed so I don't know if that means it should still be in CNO or if it would go somewhere else.

If it does stay in CNO, we are trying to avoid continuously adding more and more container images that will need to be pulled onto every node, so if possible it would be good to just have this binary be installed in the sr-iov image or something

danwinship:
Is this still wanted? I heard Feng say something about maybe not needing this after all?

mmirecki:
Give some more time, it's still under discussion.

PR #559, opened 2020-03-27
Add an option to set retry attempts in ovn kind

nerdalert:
- Resolves an issue when trying to download the
  ovn-kubernetes image from quay on limited
  bandwidth (20mbps) by adding an option to
  adjust the retry account

- Also increased the default from 15 to 20 attempts

Signed-off-by: Brent Salisbury <bsalisbu@redhat.com>

### DIFF
--- a/hack/ovn-kind-cno.sh
+++ b/hack/ovn-kind-cno.sh
@@ -13,6 +13,8 @@ SERVICE_NETWORK=${SERVICE_NETWORK:-"172.30.0.0/16"}
 # Skip the comment lines and retrieve the number of Master nodes from kind.yaml file.
 NUM_MASTER_NODES=`grep "^[^#]" kind.yaml | grep -c "role\: control-plane"`
 OVN_KIND_VERBOSITY=${OVN_KIND_VERBOSITY:-0}
+# Increase the retries if on slower bandwidth to allow for image downloads to complete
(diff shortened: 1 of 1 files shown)


nerdalert:
A bit more detail on this and some logs. Without adjusting the retry count, the `ovn-kind-cno.sh` build script fails to complete on a 20mbps circuit. 

The issue is caused by the ovn-kubernetes image download from quay not completing in the allotted retries e.g.`Pulling image "quay.io/openshift/origin-ovn-kubernetes:4.3"` in the ovn-kube pods.

- Unsuccessful install log:
```
...
OVN pod not available yet on attempt 13
OVN pod not available yet on attempt 14
OVN pod not available yet on attempt 15
OVN-k8s pods are not running
```
- Successful install log with the timeouts added:
```
...
OVN pod not available yet on attempt 17
OVN pod not available yet on attempt 18
OVN pod not available yet on attempt 19
Sym-linking cni dirs for node ec5cc039c030
Deployment Complete!
```

Ty!

danwinship:
A nicer approach might be to `kubectl get -o yaml` the master node, which will show in the `status` which images it has. Although... they might all be nameless SHA hashes so maybe that wouldn't help?

danwinship:
I think you can also examine the pod YAML in greater detail to distinguish "still pulling the image" from "pulled the image but then something weird is happening"

PR #560, opened 2020-03-30
Remove logfile from ovnkube configmap as to avoid log rotation

alexanderConstantinescu:
For ovnkube logging we can chose to specify a logfile explicitly by passing the `logfile` parameter in the ovnkube configmap. However, /dev is constrained to 64MB in openshift pods, which means that for long running pods: when logging exceeds that limit we get broken pipes for logging.    

Part of the solution (the one fixed by this patch) is to change the `logfile` location to a unconstrained file system location. 

The other part is to switch from logrus to klog completely as logrus does not provide a feature of logging to a file and to stdout simultaneously, as klog does. This part means that we should back-port the "switch to klog" patch to 4.3.

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -17,9 +17,6 @@ data:
     ovn-config-namespace="openshift-ovn-kubernetes"
     apiserver="{{.K8S_APISERVER}}"
 
-    [logging]
-    logfile="/dev/stdout"
-
     [gateway]
     mode=local
     nodeport=true


juanluisvaladas:
It creates a file in /dev/ instead of printing it to stdout? Sounds more like a problem that container runtime people should look into.

Maybe as a workaround we can use /proc/self/fd/1 instead? That should print to stdout as well.

alexanderConstantinescu:
> It creates a file in /dev/ instead of printing it to stdout? Sounds more like a problem that container runtime people should look into.
> 
> Maybe as a workaround we can use /proc/self/fd/1 instead? That should print to stdout as well.

It just uses the file `/dev/stdout` (which points to `/proc/sefl/fd/1`). Anyways, we should not be doing this as it's not safe in any case, lumberjack - which we use in ovnkubernetes [here](https://github.com/openshift/ovn-kubernetes/blob/release-4.3/go-controller/pkg/config/config.go#L1081) - log-rotates the file according to [this](https://github.com/natefinch/lumberjack/blob/v2.0/lumberjack.go#L49). This will break things in /dev anyhow sooner or later as you can see [here](http://bugzilla.redhat.com/show_bug.cgi?id=1793108#c5) where:

```
sh-4.2# ls -talrh
lrwxrwxrwx. 1 root root   15 Mar 18 16:00 stdout-2020-03-19T18-03-52.057 -> /proc/self/fd/1
```

The log-rotated file now points to /proc/self/fd/1 instead of /dev/stdout

danwinship:
We shouldn't be logging to a file. eg, all of the ovn core components have:

    -vconsole:${OVN_LOG_LEVEL} -vfile:off

in their args. ovnkube is supposed to be doing the same thing.

By default, ovnkube doesn't log to a file, so I think the fix is just to remove the `[logging]` section from the generated config

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -18,7 +18,7 @@ data:
     apiserver="{{.K8S_APISERVER}}"
 
     [logging]
-    logfile="/dev/stdout"
+    logfile="/var/log/ovnkube.log"
```
pecameron: Casey explicitly changed this to stdout so that oc logs could pick it up.

alexanderConstantinescu:
> We shouldn't be logging to a file. eg, all of the ovn core components have:
> 
> ```
> -vconsole:${OVN_LOG_LEVEL} -vfile:off
> ```
> 
> in their args. ovnkube is supposed to be doing the same thing.
> 
> By default, ovnkube doesn't log to a file, so I think the fix is just to remove the `[logging]` section from the generated config

I'm fine with that.

alexanderConstantinescu:
Actually, come to think of it now that I saw @pecameron comment. 

Logrus seems to write to the /dev/stderr buffer if no specific output is specified, see [here](https://github.com/sirupsen/logrus/blob/master/logger.go#L81) so I am not sure if removing the logfile section is going to work. We might not get _all_ the output (only the error output) when doing `oc logs`. I need to test this and verify.

bindata/network/ovn-kubernetes/004-config.yaml
alexanderConstantinescu: That might be the case for logrus, but not for klog which is already in 4.4 and 4.5. I will check during the stand-up if we have any further plans on performing ovnkube related back-ports to 4.3, and then include the "switch to klog" to 4.3 which would fix this everywhere.

danwinship:
a simpler fix for 4.3 (than backporting the klog stuff) would be to just disable log rotation if the logfile is `"/dev/stdout"`

alexanderConstantinescu:
I've tested this across all branches for ovn (ie: 4.3, 4.4  and 4.5). Removing `logfile` from the configmap does not result in any problem for both `ovnkube-master` and `ovnkube-node`, we get all logs. 

I am not suprised this works for `ovnkube-master`, but `ovnkube-node` runs with `hostPID: true`, which would mean that the ovnkube process inside does not run as PID 1, which makes me wonder how the heck `oc logs` manages to capture the logs of the `ovnkube` process which pipes it's output to `proc/X/fd/1` / `proc/X/fd/2`? Have we changed this in version 4.X, i.e that logs are not captured only for PID 1? 

@squeed you're back tomorrow, care to fill me in on this in case you know something I'm overlooking?

danwinship:
PR title / commit message are wrong now

> which makes me wonder how the heck `oc logs` manages to capture the logs of the `ovnkube` process which pipes it's output to `proc/X/fd/1` / `proc/X/fd/2`?

Because we control what fd 1 and fd 2 are when the pod is started. (`oc logs` asks the apiserver, which asks the relevant kubelet, which asks cri-o, which is holding the other end of the pipe that the pod's stdout/stderr are attached to.)

alexanderConstantinescu:
> PR title / commit message are wrong now
> 
> > which makes me wonder how the heck `oc logs` manages to capture the logs of the `ovnkube` process which pipes it's output to `proc/X/fd/1` / `proc/X/fd/2`?
> 
> Because we control what fd 1 and fd 2 are when the pod is started. (`oc logs` asks the apiserver, which asks the relevant kubelet, which asks cri-o, which is holding the other end of the pipe that the pod's stdout/stderr are attached to.)

Aha! Excellent then. I have updated the commit message and

/retitle Remove logfile from ovnkube configmap as to avoid log rotation

PTAL again

alexanderConstantinescu:
stop spamming me...

/hold 

those ovn jobs are not going to start passing anytime soon I guess...

PR #562, opened 2020-03-31
Bug 1831112: Kuryr: Mount /run/netns to ensure netns access

dulek:
openshift/machine-config-operator#1568 moved pod namespaces from
/proc into /var/run/crio. As Kuryr needs access to them in order to
manipulate interfaces, we need to mount the new directory and this
commit does that.

Most likely the same change needs to be done for ovn-kubernetes, but
it's a bit out of my expertise.

### DIFF
--- a/bindata/network/kuryr/005-daemon.yaml
+++ b/bindata/network/kuryr/005-daemon.yaml
@@ -56,6 +56,9 @@ spec:
           mountPath: /host_proc
         - name: openvswitch
           mountPath: /var/run/openvswitch
+        - name: netns
+          mountPath: /run/netns
+          mountPropagation: HostToContainer
 {{ if (default true .DaemonEnableProbes) eq "true" }}
         readinessProbe:
           failureThreshold: 10
@@ -85,6 +88,9 @@ spec:
       - name: proc
         hostPath:
           path: /proc
+      - name: netns
+        hostPath:
+          path: /run/netns
       - name: openvswitch
         hostPath:
           path: /var/run/openvswitch


dulek:
/hold

I'd like to get @dcbw opinion here, maybe the breaking change should get reverted instead.

dulek:
/hold cancel

Or let's just merge this and revert it if needed - Kuryr QE needs to get unblocked ASAP as this blocks them from validating 4.4 and 4.3 fixes.

dulek:
/hold

Okay, this does not seem to help, somehow it's now "failed to open", probably permissions.

cgwalters:
(not super important nit; `/var/run` is always a symlink to `/run`, so it's cleaner to just reference that nowadays - it emphasizes strongly that the data is not persistent.  `/var` is about persistence)

haircommander:
/retitle Bug: 1831112 Kuryr: Mount /var/run/netns to ensure netns access

haircommander:
/retitle Bug 1831112: Kuryr: Mount /var/run/netns to ensure netns access

dulek:
This seems to be ready now.

MaysaMacedo:
LGTM. Should we wait for [openshift/machine-config-operator/pull/1689](https://github.com/openshift/machine-config-operator/pull/1689) ?

dulek:
> (not super important nit; `/var/run` is always a symlink to `/run`, so it's cleaner to just reference that nowadays - it emphasizes strongly that the data is not persistent. `/var` is about persistence)

Fixed!

> LGTM. Should we wait for openshift/machine-config-operator/pull/1689 ?

No, if openshift/machine-config-operator#1689 merges without this, we're broken, we should be first.

Interestingly we can't remove mounting `/proc` too, because even with the above merged, there is no upgrade path, so old containers will stay on `/proc`.

PR #563, opened 2020-03-31
Bug 1817562: Configure tls for multus metrics endpoint

aneeshkp:
Securing multus scrape endpoint with kube-rbac-proxy

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_deployment.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment.yaml
@@ -31,6 +31,8 @@ spec:
           value: "quay.io/openshift/origin-sdn:4.3"
         - name: KUBE_PROXY_IMAGE
           value: "quay.io/openshift/origin-kube-proxy:4.3"
+        - name: KUBE_RBAC_PROXY_IMAGE
+          value: "quay.io/openshift/origin-kube-rbac-proxy:4.3"
         - name: MULTUS_IMAGE
           value: "quay.io/openshift/origin-multus-cni:4.3"
         - name: MULTUS_ADMISSION_CONTROLLER_IMAGE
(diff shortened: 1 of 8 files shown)


manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -31,6 +31,8 @@ spec:
           value: "quay.io/openshift/origin-sdn:4.3"
         - name: KUBE_PROXY_IMAGE
           value: "quay.io/openshift/origin-kube-proxy:4.3"
+        - name: KUBE_RBAC_PROXY_IMAGE
+          value: "quay.io/openshift/origin-kube-rbac-proxy:latest"
```
lilic: Nit: Any specific reason this is latest and the other are pinned tags?

aneeshkp: No. I am not sure how it works ,  these get replaced by release manager I guess, typically like an art.yaml file that ties to image-references.yaml which will get replaced by an image name that the build system understands. Something like that.

aneeshkp: @juanluisvaladas I update the image tag  from latest to 4.3

aneeshkp:
@juanluisvaladas I updated the tag to 4.3

aneeshkp:
> What address is multus listening on? I see the kube-rbac-proxy points to 127.0.0.1:9091 but I don't see any change in multus config to listen on 127.0.0.1:9091 instead of 0.0.0.0:9091
@juanluisvaladas, I didn't change because 0.0.0.0 will also listen to 127.0.0.1, Do you think I should change bind-address to 127.0.0.1 only?

juanluisvaladas:
/lgtm
@aneeshkp thinking about it, you're right about listening on 0.0.0.0. It's probably innecessary to limit the listening address to 127.0.0.1, since we're not using the RBAC proxy to add any kind of authentication/authorization and someone may be reading the old endpoint.

PR #564, opened 2020-03-31
Bug 1818029: Stop waiting for failed loadbalancer.

gryf:
In case of LB which is in error state, don't wait for it, but return
immediately.


luis5tb:
should we actually try to delete the ERROR loadbalancer instead of simply return error? Otherwise the CNO could not get out of that situation, right?

MaysaMacedo:
@luis5tb I believe that with this current approach the CNO will try to recreate the LB on the next round of reconciliation.
Considering this is current behavior, it lgtm.

luis5tb:
Ohh, right, there is a checking for that! Sorry for the noise! Looks good to me!

dulek:
/lgtm

I'll leave /approve for someone else to do double-check.

MaysaMacedo:
/hold
Let's hold this till it's verified.

PR #565, opened 2020-03-31
Bug 1807648: Fixes to reliably save/restore flows

abhat:
This PR adds:
a) a timestamp when adding logs for save/restore so they can be correlated with other logs.
b) removes the hostPID setting on the daemonset pods
c) removes the --monitor option when starting the ovs-vswitchd and ovsdb-server processes


abhat:
This is a backport of PR#524

PR #567, opened 2020-04-02
Bug 1821166: [Back-port 4.3] Remove logfile from ovnkube configmap

alexanderConstantinescu:
Back-port of https://github.com/openshift/cluster-network-operator/pull/560 to 4.3

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -17,9 +17,6 @@ data:
     ovn-config-namespace="openshift-ovn-kubernetes"
     apiserver="{{.K8S_APISERVER}}"
 
-    [logging]
-    logfile="/dev/stdout"
-
     [gateway]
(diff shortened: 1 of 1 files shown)


alexanderConstantinescu:
Actually, do we care about ovnkube on 4.3 anymore? @knobunc @dcbw

dcbw:
Same question as the 4.4 backport; it's still using logrus, are we sure this is needed?

alexanderConstantinescu:
> Same question as the 4.4 backport; it's still using logrus, are we sure this is needed?

Yes! The original bug created by @jtaleric was found on 4.4. The question I have is: do we care about ovnkube 4.3 given that VZ now seems to want 4.4 and higher, and ovnkube is tech-preview for everyone else on 4.3?

alexanderConstantinescu:
/retitle Bug 1821166: [Back-port 4.3] Remove logfile from ovnkube configmap

PR #572, opened 2020-04-06
Adds missing OVN NB SSL args to ovnkube master

trozet:
For OVN NB DB interaction with the ovn-nbctl command, we use ovn-nbctl
daemon mode, where we do not need to use SSL arguments to directly
interact with the nbctl socket in the pod. However, SSL arguments are
needed for non-nbctl commands like ovsdb client. This patch adds those
arguments.

In addition, the nbctl-daemon-mode argument was missing as well. This
argument indicates to ovn-kubernetes that it needs to use nbctl daemon
mode. Even though we were previously not setting this, we were exporting
the proper bash env var to allow the process to automatically use daemon
mode. However, we should pass the correct arguments to be consistent.
Due to an OVN bug, we need to temporarily symlink the ovn-nbctl control
socket to /var/run/ovn as it is accidentally placed in
/var/run/openvswitch today.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -331,10 +331,13 @@ spec:
           fi
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/tmp/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
             --db "{{.OVN_NB_DB_LIST}}")
+
+           # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
+          ln -sf $OVN_NB_DAEMON /var/run/ovn/
 
           exec /usr/bin/ovnkube \
             --init-master "${K8S_NODE}" \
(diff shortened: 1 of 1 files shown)


trozet:
@dcbw @danwinship PTAL

dcbw:
Build issues?

/retest

trozet:
It almost passed ovn-aws, but needs:
https://github.com/openshift/origin/pull/24833

danwinship:
So is there some underlying bug that you're fixing? Or is this just code cleanliness?

abhat:
> So is there some underlying bug that you're fixing? Or is this just code cleanliness?

The PR is needed for go-ovn bindings when we get rid of the shell invocations.

trozet:
> > So is there some underlying bug that you're fixing? Or is this just code cleanliness?
> 
> The PR is needed for go-ovn bindings when we get rid of the shell invocations.

@danwinship it's needed right now because with the SCTP code we use ovsdbclient to check if OVN supports SCTP, and to do that we need NB SSL args.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -331,11 +331,14 @@ spec:
...
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/tmp/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
             --db "{{.OVN_NB_DB_LIST}}")
 
+           # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
+          ln -s $OVN_NB_DAEMON /var/run/ovn/
```
dcbw, line 340: @trozet why not `-sf` ?

PR #574, opened 2020-04-08
Bug 1809205: Configure TLS for OpenShift SDN metrics endpoint

juanluisvaladas:
We use port 29101 for plain HTTP metrics which is firewalled and expose 29102 over HTTPS.

---
Uses RBAC proxy just like https://github.com/openshift/cluster-network-operator/pull/563
Because  the pod uses `hostNetwork: true` and I don't want to add additional ports, the regular http metrics endpoint is exposed in `127.0.01:9101` and the https one on `0.0.0.0:9101`

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -229,3 +229,62 @@ spec:
           path: /var/lib/cni/networks/openshift-sdn
       tolerations:
       - operator: Exists
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: sdn-metrics
+  namespace: openshift-sdn
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the OpenShift networking components (kube-proxy and openshift-sdn).
+      It expects that OVS is running on the node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: sdn-metrics
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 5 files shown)


danwinship:
>  Because the pod uses `hostNetwork: true` and I don't want to add additional ports, the regular http metrics endpoint is exposed in `127.0.01:9101` and the https one on `0.0.0.0:9101`

I agree that not stealing another exposed port is good, but having 127.0.0.1 behave differently from every other interface seems confusing. You could just use a firewalled port for the http server. (eg, 29101)

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -227,5 +248,8 @@ spec:
       - name: host-var-lib-cni-networks-openshift-sdn
         hostPath:
           path: /var/lib/cni/networks/openshift-sdn
+      - name: metrics-certs
+        secret:
+          secretName: router-metrics-certs-default
```
danwinship, line 253: this should presumably be `sdn-metrics-certs-default`?

juanluisvaladas: Oh, it should be tls-metrics-sdn. I copied some bits from the ingress operator in an attempt to make things consistent.

juanluisvaladas:
Odd... Looks like the service serving certificate isn't working. Let's give it another shot.
/test e2e-gcp

dcbw:
> Odd... Looks like the service serving certificate isn't working. Let's give it another shot.
> /test e2e-gcp

@juanluisvaladas is the thing that creates the service serving cert going to work without network being up yet?

juanluisvaladas:
@dcbw We delayed this on purpose for 4.6. Can you approve it?

PR #575, opened 2020-04-08
Allow to only enable hybrid overlay and not handle Windows networks

rcarrillocruz:
This allows to enable the hybrid overlay without the need to specify
hybridClusterNetwork which is only needed for clusters that connect
to Windows nodes


squeed:
The kubernetes convention is that, if no configuration is needed, passing an empty object (instead of nil) enables that feature.

So, I don't think you actually need to change the type. Just the documentation around it.

squeed:
Seems reasonable.
/lgtm

PR #576, opened 2020-04-08
WIP [release-4.4] mount /var/run shared for ovnkube

haircommander:
this is a part of https://github.com/openshift/cluster-network-operator/pull/573. I hit with a heavy hammer yesterday to verify the approach was correct, now I need to figure out which yaml fix was needed (I suspect I don't need to make all the changes in 573)

creating a PR because it's the only way I know how to test it /shrug

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -171,6 +171,7 @@ spec:
         - mountPath: /run/netns
           name: host-run-netns
           readOnly: true
+          mountPropagation: Bidirectional
         # for installing the CNI plugin binary
         - mountPath: /cni-bin-dir
           name: host-cni-bin


fidencio:
I've tested this PR and it **does** solve the issue I've faced with kata, when combined with https://github.com/cri-o/cri-o/pull/3530 (plus two patches @haircommander must likely would to have added to that cri-o PR ;-)).

It's worth to mention that I've faced some OVN weirdness, such as: 
`evel=error msg="Error while checking pod to CNI network \"multus-cni-network\": neither IPv4 nor IPv6 found when retrieving network status: [Unexpected com...`

I'm not confident to say the error faced aboved is related to this patch, as my environment is far from stable (it's an azure cluster spawned using cluster bot with kata installed via a bleeding edge, with known issues, DaemonSet);

Thanks for digging this out, @haircommander!

PR #577, opened 2020-04-08
Bug 1821291: Normalize CPU requests on nodes

abhat:
The ovs container uses approximately 100m of node CPU in a reasonable
medium sized workload. Tune the default to be more realistic for medium
clusters. For large clusters, a dynamic scaling policy would be
necessary.


abhat:
Backport for #530

danwinship:
there's already a backport, #531

PR #578, opened 2020-04-08
Bug 1822308: Adds missing OVN NB SSL args to ovnkube master

trozet:
For OVN NB DB interaction with the ovn-nbctl command, we use ovn-nbctl
daemon mode, where we do not need to use SSL arguments to directly
interact iwth the nbctl socket in the pod. However, SSL arguments are
needed for non-nbctl commands like ovsdb client. This patch adds those
arguments.

In addition, the nbctl-daemon-mode argument was missing as well. This
argument indicates to ovn-kubernetes that it needs to use nbctl daemon
mode. Even though we were previously not setting this, we were exporting
the proper bash env var to allow the process to automatically use daemon
mode. However, we should pass the correct arguments to be consistent.
Due to an OVN bug, we need to temporarily symlink the ovn-nbctl control
socket to /var/run/ovn as it is accidentally placed in
/var/run/openvswitch today.

(cherry picked from commit 3a85ce8ed24d5a2717d52e1ba89e4321c586573b)

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -369,10 +369,13 @@ spec:
           fi
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/tmp/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
             --db "{{.OVN_NB_DB_LIST}}")
+
+           # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
+          ln -sf $OVN_NB_DAEMON /var/run/ovn/ || true
 
           exec /usr/bin/ovnkube \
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -369,11 +369,14 @@ spec:
...
 
           # start nbctl daemon for caching
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/tmp/ovnk-nbctl.pid \
+          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
             --db "{{.OVN_NB_DB_LIST}}")
 
+           # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
+          ln -sf $OVN_NB_DAEMON /var/run/ovn/
```
danwinship, line 378: Hm... this will cause deployments to start failing immediately once OVN is fixed, right? ("failed to create symbolic link: File exists") You should make it only make the link if `$OVN_NB_DAEMON` is in /var/run/openvswitch.

trozet: thanks will add || true. ln -sf to the same file will error so just need to avoid failing for this case

PR #579, opened 2020-04-08
mount /var/run/netns rslave in ovnkube

haircommander:
When trying to transfer cri-o to manage its network namespaces in openshift, we have run into problems with multus. Specifically we see the error:
2020-04-07T17:46:52Z [error] delegateAdd: error invoking DelegateAdd - "ovn-k8s-cni-overlay": error in getting result from AddNetwork: CNI request failed with status 400: '[openshift-dns/dns-default-98tll] failed to configure pod interface: failed to open netns "/var/run/netns/76716600-f7fd-462f-b7df-ae054dbd144e": unknown FS magic on "/var/run/netns/76716600-f7fd-462f-b7df-ae054dbd144e": 1021994
'

through testing, I've found the netns is definitely mounted as an nsfs and not tmpfs, so I suspect we are seeing https://github.com/containernetworking/plugins/issues/69

to fix this, attempt mounting /var/run/netns as HostToContainer in ovnkube container

I have verified this works for 4.4 (by creating a cluster with https://github.com/openshift/cluster-network-operator/pull/576, replacing crio and pinns binaries, running crio using managed namespaces, and verifying the node comes up (as well as the namespaces are in /var/run/netns). Thus, this PR is ready for full review

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -163,6 +163,7 @@ spec:
         - mountPath: /run/netns
           name: host-run-netns
           readOnly: true
+          mountPropagation: HostToContainer
         # for installing the CNI plugin binary
         - mountPath: /cni-bin-dir
           name: host-cni-bin


fidencio:
I've tested the 4.4 version of this PR and it does solve the issue I've faced with kata, when combined with cri-o/cri-o#3530 (plus two patches @haircommander must likely would like to have added to that cri-o PR ;-)).

It's worth to mention that I've faced some OVN weirdness, such as:
evel=error msg="Error while checking pod to CNI network \"multus-cni-network\": neither IPv4 nor IPv6 found when retrieving network status: [Unexpected com...

I'm not confident to say the error faced aboved is related to this patch, as my environment is far from stable (it's an azure cluster spawned using cluster bot with kata installed via a bleeding edge, with known issues, DaemonSet);

Thanks for digging this out, @haircommander!

haircommander:
/retest

another day, another attempt at happy tests

haircommander:
/retest

plz

haircommander:
@dcbw @danwinship @alexanderConstantinescu happy green tests, PTAL

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -163,6 +163,7 @@ spec:
         - mountPath: /run/netns
           name: host-run-netns
           readOnly: true
+          mountPropagation: Bidirectional
```
danwinship, line 166: reading the kube docs it seems like this should only need to be `HostToContainer`, not `Bidirectional`

haircommander: alright, I'm going to give that a try

fidencio:
Works like a charm for both OCI and VM runtime types.

haircommander:
commits squashed and comment updated :smile: PTAL @danwinship

haircommander:
let's see if robots can do the work for me

/bugzilla refresh

PR #580, opened 2020-04-08
Changes to get hostnames passed in via the common name in the CR

abhat:
The certificates generated by the rotation controller today don't work with the go-ovn client bindings because they are generated without any SANs. The hostname that we pass in to the certs is a dummy common name "ovn". When the certificates are being verified by the crypto/tls library, the VerifyHostname() API will look for the hostname of the server to be present in the SANs as either an IP address/DNS name.

This change is to enable hostnames to be passed in via the pki CR template parameter for common name. The hostnames passed to the CertRotationController get eventually included as IPAddresses in the certificate template and allow us to create a valid certificate that doesn't fail the IP SAN check.

### DIFF
--- a/bindata/network/ovn-kubernetes/006-pki.yaml
+++ b/bindata/network/ovn-kubernetes/006-pki.yaml
@@ -7,4 +7,4 @@ metadata:
   namespace: openshift-ovn-kubernetes
 spec:
   targetCert:
-    commonName: ovn
+    commonName: {{.OVN_MASTER_ADDR_LIST}}
--- a/pkg/controller/pki/pki_controller.go
+++ b/pkg/controller/pki/pki_controller.go
@@ -211,7 +211,7 @@ func newPKI(config *netopv1.OperatorPKI, clientset *kubernetes.Clientset, mgr ma
 			Validity:  OneYear / 2,
 			Refresh:   OneYear / 4,
 			CertCreator: &certrotation.ServingRotation{
-				Hostnames: func() []string { return []string{spec.TargetCert.CommonName} },
+				Hostnames: func() []string { return strings.Split(spec.TargetCert.CommonName, " ") },
 
 				// Force the certificate to also be client
 				CertificateExtensionFn: []crypto.CertificateExtensionFunc{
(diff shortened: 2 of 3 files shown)


danwinship:
> This change is to enable hostnames to be passed in via the pki CR
> template parameter for common name.

why do we need that?

abhat:
> > This change is to enable hostnames to be passed in via the pki CR
> > template parameter for common name.
> 
> why do we need that?

Updating the description.

bindata/network/ovn-kubernetes/006-pki.yaml
```diff
@@ -7,4 +7,4 @@ metadata:
   namespace: openshift-ovn-kubernetes
 spec:
   targetCert:
-    commonName: ovn
+    commonName: {{.OVN_MASTER_ADDR_LIST}}
```
danwinship, line 10: Maybe the field should be `commonNames` with an "s" now? And maybe array-valued rather than taking a single string with multiple values?

Also, if we need to do this now, maybe we should stop doing the silly thing where we only generate one cert and use it as both client and server cert for every host, and instead generate multiple certs?

abhat: Yeah, I guess we could change the field to `commonNames` since this is an internal-only CRD. This was more like proof-of-concept draft PR to make sure the certs get generated with the right `IPAddresses` in the SAN field of the certificate.

If we need to generate different certs for client and server, maybe we can tackle that in a separate PR.

squeed: I would rather have separate CommonName and AdditionalNames fields.

squeed:
So, most OVN-Kubernetes users, if they use TLS, will be using the built in PKI that comes with ovs: http://docs.openvswitch.org/en/latest/howto/ssl/

From my testing, I recall that the ovs-pki also creates certificates without a CommonName that matches the DNS name. (However, we should probably verify this). If that is the case, then we'll **definitely** need ovn-kubernetes to match OVS' behavior.

abhat:
> openssl x509 -text -in

I tried generating the certs using ovs-pki and you are right, that it doesn't generate certs with CN matching the DNS names. But I am wondering why that is a hard requirement for ovn-kubernetes? Do we allow for certs to be passed in to the CNO instead of the PKI CR reconciliation to generate them?

PR #582, opened 2020-04-09
ovn: use normal OVN DB address syntax

dcbw:
ovnkube has supported normal syntax for a while, let's use it and clean up some tech debt.


dcbw:
@danwinship fixed the unit test failure.

PR #581, opened 2020-04-09
Bug 1822720: Configure TLS for OVN metrics endpoints

juanluisvaladas:
This is a quick workaround for getting OVN metrics using TLS to unblock us...

I will modify OVN kubernetes to be capable to expose HTTPS directly in the near future.

https://issues.redhat.com/browse/SDN-912

PR #583, opened 2020-04-09
Bug 1822720: Configure TLS for OVN metrics endpoints

dcbw:
Continuation of https://github.com/openshift/cluster-network-operator/pull/581 with a small bugfix to pass unit tests.

@juanluisvaladas

### DIFF
--- a/bindata/network/ovn-kubernetes/monitor.yaml
+++ b/bindata/network/ovn-kubernetes/monitor.yaml
@@ -1,35 +1,42 @@
-
+---
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    app: ovnkube-master
+    app: ovnkube-master-metrics
   annotations:
     networkoperator.openshift.io/ignore-errors: ""
-  name: monitor-ovn-master
+  name: monitor-ovn-master-metrics
   namespace: openshift-ovn-kubernetes
 spec:
   endpoints:
   - interval: 30s
     port: metrics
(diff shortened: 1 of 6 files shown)


bindata/network/ovn-kubernetes/monitor.yaml
```diff
@@ -1,4 +1,4 @@
-
+---
```
abhat, line 1: nit: the separator seems un-necessary since there is nothing before.

dcbw: @abhat yeah that part was in Juan's original PR

dcbw:
OK, this is going to be tricky. We need the secret created by the service-ca-controller to bring up the kube-rbac-proxy to secure our metrics. But the service-ca-controller is a pod-network process, which means that the ovnkube pods have to be up. But they won't be up until the kube-rbac-proxy is running, which requires certificates, which requires service-ca-controller.

danwinship:
> OK, this is going to be tricky. We need the secret created by the service-ca-controller to bring up the kube-rbac-proxy to secure our metrics. But the service-ca-controller is a pod-network process

Right. This is why we roll our own certificates for the OVN databases.

IIRC even if we made service-ca-controller hostNetwork it still depends on other things that aren't ready as early as we need it.

juanluisvaladas:
/retest 
All look like flakes

juanluisvaladas:
/close 
https://github.com/openshift/cluster-network-operator/pull/581 merged

PR #584, opened 2020-04-09
Kuryr: Open metric endpoint ports from pod subnets

dulek:
Some of the prometheus metrics endpoints we have in the cluster are
running on the host networking. Those endpoints need to be accessible
from the pod subnet. This means that we should add some more security
group rules on the masters and workers SGs to allow prometheus pods to
reach those endpoints.

This commit does so, while doing some refactoring to make increasing
numer of SG rules more manageable in Kuryr bootstrap code.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -50,9 +50,14 @@ const (
 	KuryrNamespace                         = "openshift-kuryr"
 	KuryrConfigMapName                     = "kuryr-config"
 	// NOTE(ltomasbo): Only OVN octavia driver supported on kuryr
-	OVNProvider = "ovn"
-	EtcdPort    = 2379
-	DNSPort     = 53
+	OVNProvider              = "ovn"
+	etcdPort                 = 2379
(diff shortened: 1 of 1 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -362,41 +367,49 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+			sgRule{workerSgId, cidr, 9000, 9999, rules.ProtocolUDP},
+		)
+	}
+
+	log.Print("Allowing required traffic")
+	for _, rule := range sgRules {
+		err = ensureOpenStackSgRule(client, rule.sgId, rule.cidr, rule.minPort, rule.maxPort, rule.protocol)
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to add UDP rule opening traffic to masters from %s on %d", cidr, DNSPort)
+			return nil, errors.Wrapf(err, "failed to add %s rule opening traffic from %s in rule %s on ports %d:%d",
```
MaysaMacedo, line 410: I guess it should be "in security group"

PR #586, opened 2020-04-13
Adds failsafe for OVN nbctl daemon socket

trozet:
If/When OVN gets a backport fix for the wrong nbctl socket location, our
CI will break because the symlink will fail. This adds a failsafe.


trozet:
@danwinship @dcbw PTAL

PR #589, opened 2020-04-13
Backport fixes to support upstream->downstream merges

dcbw:
Backports of #466, #467, and #572 

@trozet @knobunc @fepan @openshift/networking

PR #467, opened 2020-02-07
Remove support for ovn2.11

pecameron:
Completes the switch from ovn2.11 to ovn2.12 by removing support for the
now obsolete ovn2.11

SDN-643
https://issues.redhat.com/browse/SDN-643

Signed-off-by: Phil Cameron <pcameron@redhat.com>

PR #466, opened 2020-02-04
Bug 1803697: Changing OVN DB readinessProbe to check membership status instead of DB status

alexanderConstantinescu:
This PR fixes the recently changed `readinessProbe` for `nbdb` and `sbdb` to check each pod's membership status instead of the general cluster status. 

Some background concerning this change (as to understand the implemented check better): 

We should check raft cluster status using `ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound`, which outputs 

```
sh-4.2#  ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound       
f782
Name: OVN_Northbound
Cluster ID: f72d (f72d04b2-f90e-4ce0-98a4-3636e668d1e7)
Server ID: f782 (f7826705-a5a9-4e4c-b2a0-935da3d72743)
Address: ssl:10.0.146.224:9643
Status: cluster member
Role: follower
Term: 3
Leader: 1b1b
Vote: 1b1b

Election timer: 1000
Log: [2, 1699]
Entries not yet committed: 0
Entries not yet applied: 0
Connections: ->0368 ->1b1b <-1b1b <-0368
Servers:
    0368 (0368 at ssl:10.0.137.205:9643)
    f782 (f782 at ssl:10.0.146.224:9643) (self)
    1b1b (1b1b at ssl:10.0.134.97:9643)
```

The following will be returned in the case that the pod is not a member of the cluster (has left or otherwise):

```
sh-4.2#  ovs-appctl -t /var/run/openvswitch/ovnnb_db.ctl  cluster/status OVN_Northbound
unknown cluster
ovs-appctl: /var/run/openvswitch/ovnnb_db.ctl: server returned an error
```

/assign @dcbw

PR #590, opened 2020-04-13
Bug 1823560: backport fixes to support upstream->downstream merges

dcbw:
4.3 version of https://github.com/openshift/cluster-network-operator/pull/589

Backports of #466, #467, and #572 . These are needed to make nbctl daemon mode work while at the same time being able to check local OVS database to figure out if SCTP is supported by OVN.

@trozet @knobunc @fepan @openshift/networking


dcbw:
No longer required for 4.3

PR #591, opened 2020-04-14
Enabling pprof for ovnkubernetes by default

aakarshg:
This will help us get pprof data for ovnkube-master by default for debugging.


squeed:
/hold
until #581 merges, or a blocking 4.5 bug is opened so that we absolutely do not push a release that exposes pprof.

PR #596, opened 2020-04-16
Check for existing cluster based on presence of install state file

bostrt:
Terraform files will not be present in installation directory for all types of clusters


bostrt:
I'm wrong, an OCP 4.3 installation on AWS will only have terraform, no ignition. Apologies, I made this patch weeks ago and just submitted it without double-checking. 

I'll fix the PR to check for some other common file in the installation directory.

PR #601, opened 2020-04-20
Bug 1826322: sdn: fix typo in openshift-sdn run script

squeed:
Thanks to the magic of bash, this was never caught!

The only effect was that overriding verbosity is not possible.


danwinship:
/override ci/prow/e2e-metal-ipi

@squeed: "Not mergeable. Needs bugzilla/valid-bug label."

PR #605, opened 2020-04-24
Migrate from dep to go modules dependency management

vishnoianil:
Modules is an official dependency management system
starting from v1.14. This new dependency management
system was introduced in go version 1.11. To build
this code in go 1.14, you need to explictly off the
GO111MODULE, so that it can pickup the legacy dep
management system.


danwinship:
Can you do this as two commits, first doing a "make update-deps" to get the latest versions of everything, and *then* switching to go modules, so we can then see that it doesn't screw up any of the dependencies?

abhat:
I prefer keeping the vendor dir for offline and more hermetic builds. Even with go mod you should be able to use the vendor dir by passing in `GOFLAGS=-mod=vendor`. Look at `go mod vendor` to create the vendor dir as per how go.mod and go.sum want it.

danwinship:
> I prefer keeping the vendor dir

oh, I didn't even look at the details of the patch. Keeping the vendor dir is a non-negotiable requirement, for all openshift modules.

squeed:
Agreed, we have to have `/vendor`
/hold

vishnoianil:
> I prefer keeping the vendor dir for offline and more hermetic builds. Even with go mod you should be able to use the vendor dir by passing in `GOFLAGS=-mod=vendor`. Look at `go mod vendor` to create the vendor dir as per how go.mod and go.sum want it.



> Can you do this as two commits, first doing a "make update-deps" to get the latest versions of everything, and _then_ switching to go modules, so we can then see that it doesn't screw up any of the dependencies?

GO Modules recommend to use go mod vendor to generate the correct vendor directory artifacts (also to make sure that modules.txt in vendor directory match with go.mod). So it might fail if i will just push the patch with go.mod related changes and without vendor directory update. So end up doing both the step in one to make sure that CI passes with these changes.

vishnoianil:
> I prefer keeping the vendor dir for offline and more hermetic builds. Even with go mod you should be able to use the vendor dir by passing in `GOFLAGS=-mod=vendor`. Look at `go mod vendor` to create the vendor dir as per how go.mod and go.sum want it.

@abhat 
Yes did that with my new PR (https://github.com/openshift/cluster-network-operator/pull/616). 
Just for my understanding, i believe we consume the generate images of CNO in the deployment, so all the deps will get stored in the image when we generate the images, so i am wondering for what other builds do we need to keep all the artifact in source code repos? The build that generates the images ? i assume those must be building the code ?

vishnoianil:
> > I prefer keeping the vendor dir
> 
> oh, I didn't even look at the details of the patch. Keeping the vendor dir is a non-negotiable requirement, for all openshift modules.

@danwinship 
Yup, i did that in my new PR.PTAL Thanks

vishnoianil:
I opened following new PR for this issue to avoid naming conflict resolutsions, closing this one.
https://github.com/openshift/cluster-network-operator/pull/616 PTAL.

PR #607, opened 2020-04-26
Bug 1829233: Recreate DNS service upon Octavia upgrade

MaysaMacedo:
When Octavia is upgraded from OSP13 to OSP16 double listeners becomes
supported and the load balancers that requires TCP and UDP on same
port should be recreated.

This commit ensure the DNS service is recreated upon Octavia upgrade.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
@@ -49,6 +50,8 @@ const (
 	MinOctaviaVersionWithTimeouts          = "v2.1"
 	KuryrNamespace                         = "openshift-kuryr"
 	KuryrConfigMapName                     = "kuryr-config"
+	DNSNamespace                           = "openshift-dns"
+	DNSServiceName                         = "dns-default"
 	// NOTE(ltomasbo): Only OVN octavia driver supported on kuryr
 	OVNProvider              = "ovn"
 	etcdPort                 = 2379
@@ -176,12 +179,12 @@ func getCloudProviderCACert(kubeClient client.Client) (string, error) {
 	return cm.Data["ca-bundle.pem"], nil
 }
 
-func getSavedOctaviaProvider(kubeClient client.Client) (string, error) {
+func getSavedAnnotation(kubeClient client.Client, annotation string) (string, error) {
 	cm, err := getConfigMap(kubeClient, KuryrNamespace, KuryrConfigMapName)
 	if err != nil {
 		return "", err
 	}
-	return cm.Annotations[names.KuryrOctaviaProviderAnnotation], nil
+	return cm.Annotations[annotation], nil
 }
 
 // Logs into OpenStack and creates all the resources that are required to run
@@ -512,7 +515,7 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	//    b. List providers and look for OVN one.
 	//    c. If it's present configure Kuryr to use it.
 	//    d. In case of any issues just use whatever the default is.
-	octaviaProvider, err := getSavedOctaviaProvider(kubeClient)
+	octaviaProvider, err := getSavedAnnotation(kubeClient, names.KuryrOctaviaProviderAnnotation)
 	if err != nil && !apierrors.IsNotFound(err) { // Ignore 404, just do the normal discovery then.
 		return nil, errors.Wrap(err, "failed to get kuryr-config ConfigMap")
 	}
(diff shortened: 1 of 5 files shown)


MaysaMacedo:
/hold cancel
Verified the DNS service is recreated upon upgrade

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -542,6 +545,32 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, errors.Wrap(err, "failed to get kuryr-config ConfigMap")
+	}
+
+	maxOctaviaVersion, err := getMaxOctaviaAPIVersion(lbClient)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get max octavia api version")
+	}
+
+	if octaviaVersion != "" {
```
dulek, line 558: Can you add a comment explaining why we need all this and the logic behind it? For sure it'll be useful in 6 months. ;)

dulek:
Don't we need bug tied to this if we're planning to backport it?

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: Yup.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -542,6 +545,39 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+
+	maxOctaviaVersion, err := getMaxOctaviaAPIVersion(lbClient)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get max octavia api version")
+	}
+
+	// In case the Kuryr config-map is annotated with an Octavia version different
+	// than the current Octavia version, and older than the version that multiple
+	// listeners becomes available and the Octavia provider is amphora, an Octavia
+	// upgrade happened and UDP listeners are now allowd to be created.
```
luis5tb, line 561: not allowed

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -542,6 +545,39 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+	// By recreating the OpenShift DNS service a new load balancer amphora is in
+	// place with all required listeners.
+	log.Print("Checking Octavia upgrade happened")
+	if octaviaVersion != "" && octaviaProvider == "default" {
+		savedOctaviaVersion := semver.MustParse(octaviaVersion)
+		multipleListenersVersion := semver.MustParse(MinOctaviaVersionWithMultipleListeners)
+		if !savedOctaviaVersion.Equal(maxOctaviaVersion) && savedOctaviaVersion.LessThan(multipleListenersVersion) {
```
luis5tb, line 568: don't we need to check that the current octavia version (octaviaVersion) supports multipleListeners?

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: I meant `now allowed` as this is the result of the upgrade

luis5tb: ups... sorry, I made a typo also in the comment: I mean it should be "now allowed" instead of "now allowd"

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -542,6 +545,39 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+	// listeners becomes available and the Octavia provider is amphora, an Octavia
+	// upgrade happened and UDP listeners are now allowed to be created.
+	// By recreating the OpenShift DNS service a new load balancer amphora is in
+	// place with all required listeners.
+	log.Print("Checking Octavia upgrade happened")
+	if octaviaVersion != "" && octaviaProvider == "default" {
+		savedOctaviaVersion := semver.MustParse(octaviaVersion)
+		multipleListenersVersion := semver.MustParse(MinOctaviaVersionWithMultipleListeners)
+		if !savedOctaviaVersion.Equal(maxOctaviaVersion) && savedOctaviaVersion.LessThan(multipleListenersVersion) && octaviaMultipleListenersSupport {
+			dnsService := &v1.Service{
```
squeed, line 569: Will this cause disruption to the cluster?

MaysaMacedo: It will cause a small downtime of few seconds on the DNS resolution, as a new load balancer with the double listeners is getting created.  After the load balancer is in place, the cluster should work normally.

luis5tb: with amphora octavia driver I would say not just a few seconds, but around a minute. Note though this only happens upon OpenStack octavia upgrade between octavia version not supporting UDP to the one supporting it -- on the same port as the TCP one.

squeed:
/override ci/prow/e2e-gcp-ovn
(Because this is an Octavia change)

pkg/platform/openstack/kuryr_bootstrap.go
squeed: That is... a bit concerning. What are the end-user expectations around this upgrade?

MaysaMacedo:
/hold
Let me just perform one last test prior to merging it.

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: The path for Octavia upgrade from version 13 to 16 is still in progress.
Octavia team suggestion for the problem being solved here is to make the operator open a support ticket requesting for the upgrade of a Load Balancer. However, we can't rely on that as our approach for enforcing DNS resolution over TCP would be gone as soon as the upgrade happens.

We expect to work on other approaches for this solution to make it more ideal in following releases.

PR #608, opened 2020-04-28
Bug 1825019: Setting resource request for kube-proxy deployment

alexanderConstantinescu:
This is done as to avoid `QoSClass: BestEffort` status which will fail certain e2e tests. I applied the same requests we have for openshift-sdn. But please advise if someone has an idea on a more scientifically correct way of setting these. 

/assign @danwinship


danwinship:
"same as openshift-sdn" makes sense; most of openshift-sdn's work is the embedded kube-proxy
/lgtm

PR #611, opened 2020-05-01
Bug 1831680: Prevent patching the HybridOverlayConfig after install time

JacobTanenbaum:
Since adding a hybrid network after install time is not a supported
operation throw an error when processesing the config.

### DIFF
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -262,6 +262,34 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
 	errs := isOVNKubernetesChangeSafe(prev, next)
 	g.Expect(errs).To(BeEmpty())
 
+	// try to add a new hybrid overlay config
+	hybridOverlayConfigNext :=
+		operv1.HybridOverlayConfig{
+			HybridClusterNetwork: []operv1.ClusterNetworkEntry{
+				{CIDR: "10.132.0.0/14", HostPrefix: 23},
+			},
+		}
+	next.DefaultNetwork.OVNKubernetesConfig.HybridOverlayConfig = &hybridOverlayConfigNext
+
+	errs = isOVNKubernetesChangeSafe(prev, next)
+	g.Expect(errs).To(HaveLen(1))
+	g.Expect(errs[0]).To(MatchError("cannot start a hybrid overlay network after install time"))
+
+	//try to change a previous hybrid overlay
+	hybridOverlayConfigPrev :=
+		operv1.HybridOverlayConfig{
+			HybridClusterNetwork: []operv1.ClusterNetworkEntry{
+				{CIDR: "10.135.0.0/14", HostPrefix: 23},
+			},
+		}
+	prev.DefaultNetwork.OVNKubernetesConfig.HybridOverlayConfig = &hybridOverlayConfigPrev
+	errs = isOVNKubernetesChangeSafe(prev, next)
+	g.Expect(errs).To(HaveLen(1))
+	g.Expect(errs[0]).To(MatchError("cannot edit a running hybrid overlay network"))
+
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes_test.go
```diff
@@ -262,6 +262,34 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
 	errs := isOVNKubernetesChangeSafe(prev, next)
 	g.Expect(errs).To(BeEmpty())
 
+	// try to add a new hybrid overlay config
+	hybridOverlayConfigNext :=
+		operv1.HybridOverlayConfig{
+			HybridClusterNetwork: []operv1.ClusterNetworkEntry{
+				operv1.ClusterNetworkEntry{CIDR: "00.0.0.0", HostPrefix: 80},
```
pecameron, line 269: "00.  should be "0. Is HostPrefix: 80 valid? Also 0.0.0.0 is special, I would start somewhere else.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -262,6 +262,34 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
...
+
+	errs = isOVNKubernetesChangeSafe(prev, next)
+	g.Expect(errs).To(HaveLen(1))
+	g.Expect(errs[0]).To(MatchError("cannot start a hybrid overlay network after install time"))
+
+	//try to change a previous hybrid overlay
+	hybridOverlayConfigPrev :=
+		operv1.HybridOverlayConfig{
+			HybridClusterNetwork: []operv1.ClusterNetworkEntry{
+				operv1.ClusterNetworkEntry{CIDR: "00.0.0.1", HostPrefix: 80},
```
pecameron, line 282: "00.  change to "0.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -262,6 +262,34 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
...
+			},
+		}
+	prev.DefaultNetwork.OVNKubernetesConfig.HybridOverlayConfig = &hybridOverlayConfigPrev
+	errs = isOVNKubernetesChangeSafe(prev, next)
+	g.Expect(errs).To(HaveLen(1))
+	g.Expect(errs[0]).To(MatchError("once set cannot change ovn-kubernetes Hybrid Overlay Config"))
```
pecameron, line 288: Once set, cannot....
Maybe change string in the tested code?

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: because it is unit tests and just a string it doesnt really matter but changing to CIDR "10.132.0.0/14" and hostsubnet 23

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: changed to "10.135.0.0/14"

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: changed to "cannot edit a running hybrid overlay network"

squeed:
@JacobTanenbaum this needs a bz since we're past feature freeze.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -273,4 +301,5 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
 	g.Expect(errs).To(HaveLen(2))
 	g.Expect(errs[0]).To(MatchError("cannot change ovn-kubernetes MTU"))
 	g.Expect(errs[1]).To(MatchError("cannot change ovn-kubernetes genevePort"))
+
```
dcbw, line 304: @JacobTanenbaum can you remove the extra newline?

JacobTanenbaum: removed

PR #612, opened 2020-05-04
Bug 1829824: Remove dead member from LB pool.

gryf:
During installation a bootstrap node is used as an API node at first,
but then gets removed. In Kuryr's bootstrap we never remove it from API
loadbalancer pool which leads to Octavia stating that LB is in degraded
state.

With those changes we prevent it by making sure on reconciliation we
remove loadbalancer members that we don't need there anymore. This
should help with scaledown of masters too.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -457,9 +457,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	log.Print("Creating OpenShift API loadbalancer pool members")
 	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port)$", clusterID))
 	portList, err := listOpenStackPortsMatchingPattern(client, tag, r)
+	addresses := make([]string, 0)
 	for _, port := range portList {
 		if len(port.FixedIPs) > 0 {
 			portIp := port.FixedIPs[0].IPAddress
+			addresses = append(addresses, portIp)
 			log.Printf("Found port %s with IP %s", port.ID, portIp)
 
 			// We want bootstrap to stop being used as soon as possible, as it will serve
@@ -479,6 +481,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		} else {
 			log.Printf("Matching port %s has no IP", port.ID)
 		}
+	}
+
+	err = purgeOpenStackLbPoolMember(client, poolId, addresses)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed on purging invalid LB members from LB pool")
 	}
 
 	log.Print("Ensuring certificates")
--- a/pkg/platform/openstack/loadbalancer.go
+++ b/pkg/platform/openstack/loadbalancer.go
@@ -411,3 +411,35 @@ func listOpenStackOctaviaProviders(client *gophercloud.ServiceClient) ([]provide
 		return providersList, nil
 	}
 }
+
+// Iterate on pool members and check their address against provided list
+// addresses of current master/bootstrap nodes. Remove all surplus members,
+// which address doesn't exists on that list.
+func purgeOpenStackLbPoolMember(client *gophercloud.ServiceClient, poolId string, addresses []string) error {
+	page, err := pools.ListMembers(client, poolId, nil).AllPages()
+	if err != nil {
+		return errors.Wrap(err, "failed to get LB member list")
+	}
+
+	members, err := pools.ExtractMembers(page)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract LB members list")
+	}
+
+	for _, member := range members {
+		found := false
+		for _, address := range addresses {
+			if address == member.Address {
+				found = true
+				break
+			}
+		}
+		if !found {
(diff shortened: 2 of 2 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -481,6 +483,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		}
 	}
 
+	err := purgeOpenStackLbPoolMember(client, poolId, addresses)
```
MaysaMacedo, line 486: the variable err already exists, no need to recreate it, just to update.

pkg/platform/openstack/loadbalancer.go
```diff
@@ -411,3 +411,34 @@ func listOpenStackOctaviaProviders(client *gophercloud.ServiceClient) ([]provide
 		return providersList, nil
 	}
 }
+
+// Iterate on pool members and check their address against provided list.
+// Remove all surplus members, which address doesn't existst on that list.
```
MaysaMacedo, line 416: typo: exist. Nit: better to specify what is the provided list.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -481,6 +483,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed on purging LB pool.")
```
MaysaMacedo, line 488: Nit: LB pool member.

gryf: Actually, I'm purging pool from invalid members ;)

pkg/platform/openstack/kuryr_bootstrap.go
gryf: ack.

pkg/platform/openstack/loadbalancer.go
gryf: ack.

pkg/platform/openstack/loadbalancer.go
```diff
@@ -411,3 +411,35 @@ func listOpenStackOctaviaProviders(client *gophercloud.ServiceClient) ([]provide
...
+	if err != nil {
+		return errors.Wrap(err, "failed to get LB member list")
+	}
+
+	members, err := pools.ExtractMembers(page)
+	if err != nil {
+		return errors.Wrap(err, "failed to extract LB members list")
+	}
+
+	for _, member := range members {
```
luis5tb, line 429: as we have already added the "new members", does it make sense to compare here members and addresses length first, and if the same assume members are already up to date?

gryf: Seems tempting, although it won't prevent us from making less Octavia calls, right? And this will ensure, that members addresses are the same as for nodes addresses (as it *should* be) :)

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -481,6 +483,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 		}
 	}
 
+	err = purgeOpenStackLbPoolMember(client, poolId, addresses)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed on purging LB pool from invalid members.")
```
MaysaMacedo, line 488: Perhaps this could be rephrased to "Failed on purging invalid LB members from LB pool"
The `DeleteMember` function removes a members from a pool. [1]

[1] https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/pools/requests.go#L428-L429

gryf: Ok!

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -457,9 +457,11 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	log.Print("Creating OpenShift API loadbalancer pool members")
 	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port)$", clusterID))
 	portList, err := listOpenStackPortsMatchingPattern(client, tag, r)
+	addresses := make([]string, 0)
```
dulek, line 460: No need for `make()` in that case, just `var addresses []string` to create a nil slice should be fine, `append` will be fine with it.

pkg/platform/openstack/loadbalancer.go
dulek: I support @gryf here, we could miss kind of a failover of master VM. Regardless of if it's possible or not, it's almost no cost to a do full comparison. It's just golang makes it look long and clumsy. ;)

pkg/platform/openstack/loadbalancer.go
```diff
@@ -411,3 +411,35 @@ func listOpenStackOctaviaProviders(client *gophercloud.ServiceClient) ([]provide
...
+		found := false
+		for _, address := range addresses {
+			if address == member.Address {
+				found = true
+				break
```
dulek, line 434: Instead of using `found` flag you could just use loops labels: mark loop in 429 as `members_loop` and make this a `continue members_loop`.

PR #615, opened 2020-05-05
Bug 1822296: Expose raft (nb-db/sb-db) election-timer and ovn-controller inactivit…

vishnoianil:
…y-probe.

These timers are currently set to fixed value based on my current observation
from scale tests. We might have to increase these values in future based on
the scale we will be supporting for upcoming release. Currently election-timer
values are limited by the raft jsonrpc inactivity-probe time of 5 seconds as well.
To further increase the election-timer value, we need to disable jsonrpc
inactivity-probe.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -185,39 +185,41 @@ spec:
                   fi
                 done
 
-                retries=0
-                while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
-                  | grep "Role: candidate" ); do
-                  if [[ ! -z "${is_candidate}" ]]; then
-                    (( retries += 1 ))
-                    if [[ "${retries}" -gt 10 ]]; then
-                      echo "Cluster node (nb-db raft) is in candidate role for prolonged time. Continuing..."
-                    fi
-                    sleep 2
-                  else
-                    break
-                  fi
-                done
-
-                is_leader=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
-                  | grep "Role: leader")
-                if [[ ! -z "${is_leader}" ]]; then
-                  while [[ ${current_election_timer} != ${election_timer} ]]; do
-                    max_electinon_timer=$((${current_election_timer} * 2))
-                    if [[ ${election_timer} -le ${max_electinon_timer} ]]; then
-                      if ! ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${election_timer}; then
-                        echo "Failed to set nb-db raft election timer ${election_timer}. Exiting..."
-                        exit 2
+                if [[ ${election_timer} -ne ${current_election_timer} ]]; then
+                  retries=0
+                  while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
+                    | grep "Role: candidate" ); do
+                    if [[ ! -z "${is_candidate}" ]]; then
+                      (( retries += 1 ))
+                      if [[ "${retries}" -gt 10 ]]; then
+                        echo "Cluster node (nb-db raft) is in candidate role for prolonged time. Continuing..."
                       fi
-                      current_election_timer=${electinon_timer}
+                      sleep 2
                     else
-                      if ! ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${max_electinon_timer}; then
-                        echo "Failed to set nb-db raft election timer ${max_election_timer}. Exiting..."
-                        exit 2
-                      fi
-                    current_election_timer=${max_electinon_timer}
+                      break
                     fi
                   done
+
+                  is_leader=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
+                    | grep "Role: leader")
+                  if [[ ! -z "${is_leader}" ]]; then
+                    while [[ ${current_election_timer} != ${election_timer} ]]; do
+                      max_election_timer=$((${current_election_timer} * 2))
+                      if [[ ${election_timer} -le ${max_election_timer} ]]; then
+                        if ! ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${election_timer}; then
+                          echo "Failed to set nb-db raft election timer ${election_timer}. Exiting..."
+                          exit 2
+                        fi
+                        current_election_timer=${election_timer}
+                      else
+                        if ! ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${max_election_timer}; then
+                          echo "Failed to set nb-db raft election timer ${max_election_timer}. Exiting..."
+                          exit 2
+                        fi
+                      current_election_timer=${max_election_timer}
+                      fi
+                    done
+                  fi
                 fi
 
         readinessProbe:
@@ -353,39 +355,41 @@ spec:
                   fi
                 done
 
-                retries=0
-                while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
-                  | grep "Role: candidate" ); do
-                  if [[ ! -z "${is_candidate}" ]]; then
-                    (( retries += 1 ))
-                    if [[ "${retries}" -gt 10 ]]; then
-                      echo "Cluster node (sb-db raft) is in candidate role for prolonged time. Continuing..."
-                    fi
-                    sleep 2
-                  else
-                    break
-                  fi
-                done
-
-                is_leader=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
-                  | grep "Role: leader")
-                if [[ ! -z "${is_leader}" ]]; then
-                  while [[ ${current_election_timer} != ${election_timer} ]]; do
-                    max_electinon_timer=$((${current_election_timer} * 2))
-                    if [[ ${election_timer} -le ${max_electinon_timer} ]]; then
-                      if ! ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${election_timer}; then
-                        echo "Failed to set sb-db raft election timer ${election_timer}. Exiting..."
-                        exit 2
+                if [[ ${election_timer} -ne ${current_election_timer} ]]; then
+                  retries=0
+                  while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
+                    | grep "Role: candidate" ); do
+                    if [[ ! -z "${is_candidate}" ]]; then
+                      (( retries += 1 ))
+                      if [[ "${retries}" -gt 10 ]]; then
+                        echo "Cluster node (sb-db raft) is in candidate role for prolonged time. Continuing..."
                       fi
-                      current_election_timer=${electinon_timer}
+                      sleep 2
                     else
-                      if ! ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${max_electinon_timer}; then
-                        echo "Failed to set sb-db raft election timer ${max_election_timer}. Exiting..."
-                        exit 2
-                      fi
-                    current_election_timer=${max_electinon_timer}
+                      break
                     fi
                   done
+
+                  is_leader=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
+                    | grep "Role: leader")
+                  if [[ ! -z "${is_leader}" ]]; then
+                    while [[ ${current_election_timer} != ${election_timer} ]]; do
+                      max_election_timer=$((${current_election_timer} * 2))
+                      if [[ ${election_timer} -le ${max_election_timer} ]]; then
+                        if ! ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${election_timer}; then
+                          echo "Failed to set sb-db raft election timer ${election_timer}. Exiting..."
+                          exit 2
+                        fi
+                        current_election_timer=${election_timer}
+                      else
+                        if ! ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${max_election_timer}; then
+                          echo "Failed to set sb-db raft election timer ${max_election_timer}. Exiting..."
+                          exit 2
+                        fi
+                      current_election_timer=${max_election_timer}
+                      fi
+                    done
+                  fi
                 fi
 
         readinessProbe:
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -151,9 +151,12 @@ spec:
               - /bin/bash
               - -c
               - |
+                set -x
                 MASTER_IP="{{.OVN_MASTER_IP}}"
                 if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
                   echo "$(date -Iseconds) - nbdb - postStart - waiting for master to be selected"
+
+                  # set the connection and disable inactivity probe
                   retries=0
                   while ! ovn-nbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_NB_PORT}}{{.LISTEN_DUAL_STACK}} -- set connection . inactivity_probe=60000; do
                     (( retries += 1 ))
@@ -164,6 +167,61 @@ spec:
                   sleep 2
                   done
                 fi
+
+                election_timer="${OVN_NB_RAFT_ELECTION_TIMER}"
+                echo "Setting nb-db raft election timer to ${election_timer} ms"
+                retries=0
+                while current_election_timer=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
+                  | grep "Election" | sed "s/.*:[[:space:]]//"); do
+                  if [[ -z "${current_election_timer}" ]]; then
+                    (( retries += 1 ))
+                    if [[ "${retries}" -gt 10 ]]; then
+                      echo "Failed to get current nb-db raft election timer value after multiple attempts. Exiting..."
+                      exit 1
+                    fi
+                    sleep 2
+                  else
+                    break
+                  fi
+                done
+
+                retries=0
+                while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
+                  | grep "Role: candidate" ); do
+                  if [[ ! -z "${is_candidate}" ]]; then
+                    (( retries += 1 ))
+                    if [[ "${retries}" -gt 10 ]]; then
+                      echo "Cluster node (nb-db raft) is in candidate role for prolonged time. Continuing..."
+                    fi
+                    sleep 2
+                  else
+                    break
+                  fi
+                done
+
+                is_leader=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
+                  | grep "Role: leader")
+                if [[ ! -z "${is_leader}" ]]; then
+                  while [[ ${current_election_timer} != ${election_timer} ]]; do
+                    max_electinon_timer=$((${current_election_timer} * 2))
+                    if [[ ${election_timer} -le ${max_electinon_timer} ]]; then
+                      ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${election_timer}
+                      if [[ $? != 0 ]]; then
+                        echo "Failed to set nb-db raft election timer ${election_timer}. Exiting..."
+                        exit 2
+                      fi
+                      current_election_timer=${electinon_timer}
+                    else
+                      ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound ${max_electinon_timer}
+                      if [[ $? != 0 ]]; then
+                        echo "Failed to set nb-db raft election timer ${max_election_timer}. Exiting..."
+                        exit 2
+                      fi
+                    current_election_timer=${max_electinon_timer}
+                    fi
+                  done
+                fi
+
         readinessProbe:
           initialDelaySeconds: 30
           timeoutSeconds: 5
@@ -177,6 +235,8 @@ spec:
         env:
         - name: OVN_LOG_LEVEL
           value: info 
+        - name: OVN_NB_RAFT_ELECTION_TIMER
+          value: "{{.OVN_NB_RAFT_ELECTION_TIMER}}"
         - name: K8S_NODE_IP
           valueFrom:
             fieldRef:
@@ -216,7 +276,7 @@ spec:
         - /bin/bash
         - -c
         - |
-          set -xe
+          set -x
           if [[ -f /env/_master ]]; then
             set -o allexport
             source /env/_master
@@ -261,9 +321,12 @@ spec:
               - /bin/bash
               - -c
               - |
+                set -x
                 MASTER_IP="{{.OVN_MASTER_IP}}"
                 if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
                   echo "$(date -Iseconds) - sdb - postStart - waiting for master to be selected"
+
+                  # set the connection and disable inactivity probe
                   retries=0
                   while ! ovn-sbctl --no-leader-only -t 5 set-connection pssl:{{.OVN_SB_PORT}}{{.LISTEN_DUAL_STACK}} -- set connection . inactivity_probe=60000; do
                     (( retries += 1 ))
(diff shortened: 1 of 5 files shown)


vishnoianil:
@dcbw @trozet PTAL, thanks.

dcbw:
First thought is that we shouldn't duplicate code if we can help it. @squeed do we move back to having a separate script that we can source and call functions from?

vishnoianil:
> First thought is that we shouldn't duplicate code if we can help it. @squeed do we move back to having a separate script that we can source and call functions from?

I was thinking the same. Frankly saying, it looked pretty ugly to me, but i didn't wanted to disrupt too much in the deployment yaml. I had another thought to align this part of the code  with the upstream ovn-kubernetes `ovndb-raft-functions` + ovnkube.sh approach, but problem with that is the way we define our daemonsets for ovnkube-master and ovnkube-node, so we probably will have to manage our own ovnkube.sh, which i believe is not desired because we don't push any custom code in our downstream ovn-kubernetes (everything is contained in CNO). 

Even if we move this logic to separate script (more cleaner approach), i would like to eventually get rid of this logic, because in my opinion it's unnecessary from consumer's point of view. I opened the following RFE to internal openvswitch project to provide support to set election-timer at the start-up of the raft cluster. That would make our life pretty simple.

https://bugzilla.redhat.com/show_bug.cgi?id=1831755

I am opening another bugzilla against ovn to wrap this feature in ovn-ctl (which we use to start the raft clusters) so user can pass election-timer through it.

vishnoianil:
OVN Bugzilla : https://bugzilla.redhat.com/show_bug.cgi?id=1831778

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -268,6 +329,61 @@ spec:
                   sleep 2
                   done
                 fi
+
+                election_timer="${OVN_SB_RAFT_ELECTION_TIMER}"
+                echo "Setting sb-db raft election timer to ${election_timer} ms"
+                retries=0
+                while current_election_timer=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
+                  | grep "Election" | sed "s/.*:[[:space:]]//"); do
```
rbbratta, line 337: sometimes to be clear on non-greediness in matching up to a delimiter we can use a negated character class containing the delimiter

long way of asking if  `"s/[^:]*:[[:space:]]//"`  might be slightly better.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -268,6 +329,61 @@ spec:
...
+                done
+
+                is_leader=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
+                  | grep "Role: leader")
+                if [[ ! -z "${is_leader}" ]]; then
+                  while [[ ${current_election_timer} != ${election_timer} ]]; do
+                    max_electinon_timer=$((${current_election_timer} * 2))
+                    if [[ ${election_timer} -le ${max_electinon_timer} ]]; then
+                      ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${election_timer}
+                      if [[ $? != 0 ]]; then
```
rbbratta, line 371: `if` can check exit status directly without having to use `$?` , and we can use `!` to check for failure.

```bash
if ! ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound ${election_timer}; then
    echo "Failed to set sb-db raft election timer ${election_timer}. Exiting..."
    exit 2
fi
```

pecameron:
@vishnoianil  @dcbw I thought we avoided control knobs like this. @smarterclayton made the point a while ago that we should figure out the correct setting without bothering the admin.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
vishnoianil: @rbbratta I believe grep on constant string just before this sed makes it pretty less greedy ?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
vishnoianil: I am not an scripting expert but will it log the error message from ovs-appctl in case of failure, if it's used in if statement ?

vishnoianil:
> @vishnoianil @dcbw I thought we avoided control knobs like this. @smarterclayton made the point a while ago that we should figure out the correct setting without bothering the admin.

@pecameron I believe that's the idea as of now. These knobs are exposed so that we can set it to some fixed value based on our scale testing, and the cluster can be deployed with those initail raft timer values. 

Currently there is no way to set these values at the deployment time, admin will have to find the nb/sb db leader and fire ovs-appctl command to change the timer, and also need to do the same across all the worker nodes in the deployment to change the inactivity-probe.

smarterclayton:
As long as this is not in a public API for customers that is fine.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -44,7 +44,13 @@ spec:
         - name: ROUTE_OVERRRIDE_CNI_IMAGE
           value: "quay.io/openshift/origin-multus-route-override-cni:4.4"
         - name: OVN_IMAGE
-          value: "quay.io/openshift/origin-ovn-kubernetes:4.3"
+          value: "docker.io/avishnoi/ovn-kube-f-dev:latest-v2"
```
smarterclayton, line 47: This can't merge like this (i assume this is for testing?

vishnoianil: @smarterclayton Thanks for catching this, i think i missed it during my self reviewing the patch. Pushing the patch with the fix.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: yes it will display output, stdout and stderr are the same.

```bash
$ if ! ls /tmp/nosuch ; then echo yes ; fi
ls: cannot access '/tmp/nosuch': No such file or directory
yes
```

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: if there were multiple `:` in the Election line the sed would match the second.   We usually try to avoid `.*` in general

but bikeshedding aside, you could use awk
```bash
| awk '/Election timer:/ {print $3}'
```
or grep   (using PCRE positive lookbehind assertion)
```bash
| grep -oP '(?<=Election timer:\s)[[:digit:]]+'
```

vishnoianil: Fixed.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
vishnoianil: Pushed the patch.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
vishnoianil: fixed.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -161,6 +163,59 @@ spec:
                   sleep 2
                   done
                 fi
+
+                election_timer="${OVN_NB_RAFT_ELECTION_TIMER}"
```
squeed, line 167: Does this affect ovnkube-master? Would it make sense for it to be a separate container? It could just apply changes as necessary.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -45,6 +45,12 @@ spec:
           value: "quay.io/openshift/origin-multus-route-override-cni:4.4"
         - name: OVN_IMAGE
           value: "quay.io/openshift/origin-ovn-kubernetes:4.3"
+        - name: OVN_NB_RAFT_ELECTION_TIMER
```
squeed, line 48: This should be added the operator configuration CRD, which is defined here: https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L37

squeed: However, that is a long process, and won't be in in time for 4.5. The problem is that these fields will be instantly overwritten by the CVO.

squeed:
I think this PR is basically fine as written, albeit a bit awkward (as everyone has mentioned). I'd just like to see the configuration knob somewhere else.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
vishnoianil: I think setting this timer while the master pods startup will give us bit more deterministic behavior. If user start cluster with high number of worker nodes, we can get into the death loop of raft partition. 
Also i believe the general guideline is that we don't want to allow user to change this bahavior dynamically.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
vishnoianil: Personally i don't like this PR as well. Mainly because the method that OVN provides us to set the election-timer is not very operation friendly. I already raised a RFE for OVN team to improve this and take election-timer as a parameter to ovn-ctl. That will make the entire logic pretty simple and we can write more stable logic to set the timer. So i am hoping that we will get rid of this logic, once OVN provides us a better way to set this timer.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
squeed: Right, what I'm saying is that this should be a separate container in the master pod, rather than gating the master container.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
squeed: That makes sense. I'm arguing that this value should either be hard-coded in the code, or available as a configuration knob.

dcbw:
@vishnoianil I'm confused as to why this PR has 35 commits... Can you rebase it to get down to the single commit(s) that make the change?

vishnoianil:
> @vishnoianil I'm confused as to why this PR has 35 commits... Can you rebase it to get down to the single commit(s) that make the change?

@dcbw yes i am working on it, multiple rebase and merge caused this mess.

vishnoianil:
> > @vishnoianil I'm confused as to why this PR has 35 commits... Can you rebase it to get down to the single commit(s) that make the change?
> 
> @dcbw yes i am working on it, multiple rebase and merge caused this mess.

@dcbw done

bindata/network/ovn-kubernetes/ovnkube-master.yaml
vishnoianil: If we think of allowing user to change this value dynamically, i think having separate container makes sense to me. But if it's one time startup config, there is nothing much for this container to do, it will come up, set the value and die, and not sure CNO will be happy with that.

manifests/0000_70_cluster-network-operator_03_deployment.yaml
vishnoianil: I am more in the favor of having it as a configuration knob, because for raft based implementation, choosing approximately right election-timer depends on the workload.

dcbw: I agree with @squeed that it should be hard-coded for now, until we have a better idea whether it acutally does need to be changed. It should *not* be user-visible configuration at any rate, especially not in the API.

vishnoianil: @dcbw In general we want to keep these numbers as minimal as possible. Currently these numbers are set for 200 nodes (for 50 pods per node workload), so the moment number of nodes increases, you will have to increase it, there is no question about it. We can hardcode it for 500 nodes as of now (60 seconds-- frankly it's very high), and live with it until and unless something else comes up that require it to change it.
For election-timer i don't have strong opinion against hardcoding, because you can change this value anyways directly on the pod if needed.
So for hardcoding what approach do we need to take ? using operator CRD? is it okay to make operator CRD to be aware of SB DB deployment model (raft or ha)?

vishnoianil:
> Since there is a limit on how much you can increase the timer in one shot, should we let the user know the maximum they can increase to versus accepting any value?

@jtaleric This patch takes care of that. It increases the value double everytime, till it reaches the set value.

dcbw:
@vishnoianil looks like a legit failure in the OVN tests:

```
        "ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured|PrometheusRemoteWriteDesiredShards\",alertstate=\"firing\",severity!=\"info\"} >= 1": {
            s: "promQL query: ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured|PrometheusRemoteWriteDesiredShards\",alertstate=\"firing\",severity!=\"info\"} >= 1 had reported incorrect results:\n[{\"metric\":{\"__name__\":\"ALERTS\",\"alertname\":\"TargetDown\",\"alertstate\":\"firing\",\"job\":\"ovnkube-node\",\"namespace\":\"openshift-ovn-kubernetes\",\"service\":\"ovn-kubernetes-node\",\"severity\":\"warning\"},\"value\":[1591657715.838,\"1\"]}]",
        },
```

dcbw:
@vishnoianil can you rebase?

dcbw:
>  I looked at all the master and node logs, seems like the timer values are set fine. what does this error real means, my guess is it's suspecting that one of the ovnkube-node is down?

@vishnoianil would be good to look further into the issue and see when exactly the ovnkube node that prometheus is complaining about is down. And then also see if the same issue with ovnkube-node happens on master without your PR. eg ensure the issue is not caused by this PR.

vishnoianil:
> > I looked at all the master and node logs, seems like the timer values are set fine. what does this error real means, my guess is it's suspecting that one of the ovnkube-node is down?
> 
> @vishnoianil would be good to look further into the issue and see when exactly the ovnkube node that prometheus is complaining about is down. And then also see if the same issue with ovnkube-node happens on master without your PR. eg ensure the issue is not caused by this PR

@dcbw yes, i did dig deeper. I watched the ci jobs through the job console and i don't see any ovnkube-node pod restarting. Only pod that restarted was the first ovnkube-master pod 

```
ovnkube-master-72jh8   4/4     Running   2          43m
ovnkube-master-cbzpx   4/4     Running   0          43m
ovnkube-master-v6dxh   4/4     Running   0          43m
ovnkube-node-255rm     2/2     Running   0          32m
ovnkube-node-n9v4m     2/2     Running   0          43m
ovnkube-node-rvhkq     2/2     Running   0          43m
ovnkube-node-swhth     2/2     Running   0          32m
ovnkube-node-vhxng     2/2     Running   0          32m
ovnkube-node-zjnln     2/2     Running   0          43m
ovs-node-5ftxz         1/1     Running   0          43m
ovs-node-8vrth         1/1     Running   0          32m
ovs-node-jp25c         1/1     Running   0          32m
ovs-node-knq48         1/1     Running   0          43m
ovs-node-slscn         1/1     Running   0          43m
ovs-node-zmgss         1/1     Running   0          32m
```
And looking at the description of this pod i see following warning

```
Events:
  Type     Reason       Age                From                                          Message
  ----     ------       ----               ----                                          -------
  Normal   Scheduled    <unknown>          default-scheduler                             Successfully assigned openshift-ovn-kubernetes/ovnkube-node-n9v4m to ci-op-mzq6mf93-b3a20-9nlss-master-1
  Warning  FailedMount  51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  MountVolume.SetUp failed for volume "ovn-ca" : failed to sync configmap cache: timed out waiting for the condition
  Normal   Pulled       51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Container image "registry.svc.ci.openshift.org/ci-op-mzq6mf93/stable@sha256:d819f3c65b58443f197ba19642b9505504575a37774c33111593dfbfffb1fc60" already present on machine
  Normal   Created      51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Created container ovn-controller
  Normal   Started      51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Started container ovn-controller
  Normal   Pulled       51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Container image "registry.svc.ci.openshift.org/ci-op-mzq6mf93/stable@sha256:d819f3c65b58443f197ba19642b9505504575a37774c33111593dfbfffb1fc60" already present on machine
  Normal   Created      51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Created container ovnkube-node
  Normal   Started      51m                kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Started container ovnkube-node
  Warning  Unhealthy    50m (x6 over 50m)  kubelet, ci-op-mzq6mf93-b3a20-9nlss-master-1  Readiness probe failed:
```

I will trigger the job of master branch to see if it's happening with the master branch of not.

dcbw:
Latest failures show failure to start ovn-controller due to br-int not being accessible, that is likely due to https://github.com/openshift/machine-config-operator/pull/1636 and will be fixed by https://github.com/openshift/machine-config-operator/pull/1830

vishnoianil:
> Latest failures show failure to start ovn-controller due to br-int not being accessible, that is likely due to [openshift/machine-config-operator#1636](https://github.com/openshift/machine-config-operator/pull/1636) and will be fixed by [openshift/machine-config-operator#1830](https://github.com/openshift/machine-config-operator/pull/1830)

@dcbw Yes, i already commented on relevant PR's. If any of the PR can be merge will resolve the current failure. Seems like we are planning to merge the revert PR in MCO.

dcbw:
e2e-ovn-step-registry failure needs investigation.

windows-hybrid is `Error: Unable to find matching route for Route Table (rtb-05f05fe1aa7c2967b) and destination CIDR block (0.0.0.0/0)` so an AWS error.

windows-hybrid-network failure is an AWS lease failure because jobs are too busy.

dcbw:
```
2020/06/22 14:49:06 Copied 2.60MB of artifacts from release-latest to /logs/artifacts/release-latest
{"component":"entrypoint","file":"prow/entrypoint/run.go:164","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2020-06-22T18:45:53Z"}
2020/06/22 18:45:53 error: Process interrupted with signal interrupt, cancelling execution...
```
I don't even know what this means...

/retest

vishnoianil:
For windows-hybrid-network tests, it shows that postStart script fails for one of ovnkube-master nodes because it was not able to set the election-timer to 5000 from 4000, although setting of 1000 to 2000 and 2000 to 4000 was successful.

```
ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound
++ is_leader='Role: leader'
++ [[ ! -z Role: leader ]]
++ [[ 1000 != 5000 ]]
++ max_election_timer=2000
++ [[ 5000 -le 2000 ]]
++ ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound 2000
++ current_election_timer=2000
++ [[ 2000 != 5000 ]]
++ max_election_timer=4000
++ [[ 5000 -le 4000 ]]
++ ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound 4000
++ current_election_timer=4000
++ [[ 4000 != 5000 ]]
++ max_election_timer=8000
++ [[ 5000 -le 8000 ]]
++ ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/change-election-timer OVN_Northbound 5000
+election timer change pending.
+ovs-appctl: /var/run/ovn/ovnnb_db.ctl: server returned an error
++ echo 'Failed to set nb-db raft election timer 5000. Exiting...'
++ exit 2
+\"",
```
Looking at the pod for which postStart is failing, seems like pod received signal 15 and terminated, and that's the reason election-timer setting from 4000 to 5000 failed.
```
2020-06-22T20:42:35Z|00004|ovsdb_server|INFO|ovsdb-server (Open vSwitch) 2.13.0
2020-06-22T20:42:35Z|00005|raft|INFO|Election timer changed from 1000 to 2000
2020-06-22T20:42:35Z|00006|raft|INFO|Election timer changed from 2000 to 4000
2020-06-22T20:42:35Z|00001|fatal_signal(log_fsync0)|WARN|terminating with signal 15 (Terminated)
2020-06-22T20:42:35Z|00007|fatal_signal|WARN|terminating with signal 15 (Terminated)
2020-06-22T20:42:35Z|00001|fatal_signal(urcu1)|WARN|terminating with signal 15 (Terminated)
```
and that's the reason we can see the failure event in the events.json. Although after the pod restart, timer was set properly to 5000.

```
2020-06-22T20:42:54Z|00004|ovsdb_server|INFO|ovsdb-server (Open vSwitch) 2.13.0
2020-06-22T20:42:54Z|00005|raft|INFO|Election timer changed from 4000 to 5000
2020-06-22T20:42:57Z|00006|raft|INFO|ssl:10.0.202.193:41778: learned server ID 2411
```
So i doubt that's causing the following test failure in this job, which runs once the cluster is up and running.

```
FAIL: TestWSU/VM_specific_tests/VM_0/Run_the_WSU_against_the_same_VM_again/East-west_networking (41.06s)
```

Similary ovn-hybrid-step-registry registry events.json shows and event from ovnkube-master reporting an error while setting up the election-timer (but this time for sb-db cluster).
```
+++ ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound
+++ grep 'Role: leader'
++ is_leader='Role: leader'
++ [[ ! -z Role: leader ]]
++ [[ 1000 != 5000 ]]
++ max_election_timer=2000
++ [[ 5000 -le 2000 ]]
++ ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound 2000
++ current_election_timer=2000
++ [[ 2000 != 5000 ]]
++ max_election_timer=4000
++ [[ 5000 -le 4000 ]]
++ ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/change-election-timer OVN_Southbound 4000
+election timer change pending.
+ovs-appctl: /var/run/ovn/ovnsb_db.ctl: server returned an error
++ echo 'Failed to set sb-db raft election timer 4000. Exiting...'
++ exit 2
```

And seems like the reason is same -- pod terminated with SIGTERM

```
2020-06-22T20:45:46Z|00012|raft|INFO|Election timer changed from 1000 to 2000
2020-06-22T20:45:46Z|00001|fatal_signal(log_fsync0)|WARN|terminating with signal 15 (Terminated)
2020-06-22T20:45:46Z|00013|fatal_signal|WARN|terminating with signal 15 (Terminated)
2020-06-22T20:45:46Z|00001|fatal_signal(urcu1)|WARN|terminating with signal 15 (Terminated)
```
Looking at the pods info, all the ovn-kubernetes pods seems fine here

```
openshift-ovn-kubernetes                           ovnkube-master-dwmw6                                                  4/4     Running             2          42m   10.0.160.238   ip-10-0-160-238.us-west-1.compute.internal   <none>           <none>
openshift-ovn-kubernetes                           ovnkube-master-j5hzl                                                  4/4     Running             1          42m   10.0.155.8     ip-10-0-155-8.us-west-1.compute.internal     <none>           <none>
openshift-ovn-kubernetes                           ovnkube-master-zrvdc                                                  4/4     Running             0          42m   10.0.214.53    ip-10-0-214-53.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-cc7q9                                                    2/2     Running             0          32m   10.0.211.34    ip-10-0-211-34.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-krlvr                                                    2/2     Running             0          42m   10.0.160.238   ip-10-0-160-238.us-west-1.compute.internal   <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-mrmmq                                                    2/2     Running             0          42m   10.0.214.53    ip-10-0-214-53.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-plmzj                                                    2/2     Running             0          32m   10.0.148.18    ip-10-0-148-18.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-rf4sc                                                    2/2     Running             0          32m   10.0.135.2     ip-10-0-135-2.us-west-1.compute.internal     <none>           <none>
openshift-ovn-kubernetes                           ovnkube-node-txvcq                                                    2/2     Running             0          42m   10.0.155.8     ip-10-0-155-8.us-west-1.compute.internal     <none>           <none>
openshift-ovn-kubernetes                           ovs-node-555w6                                                        1/1     Running             0          42m   10.0.155.8     ip-10-0-155-8.us-west-1.compute.internal     <none>           <none>
openshift-ovn-kubernetes                           ovs-node-gswz4                                                        1/1     Running             0          32m   10.0.211.34    ip-10-0-211-34.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovs-node-jk9pb                                                        1/1     Running             0          42m   10.0.214.53    ip-10-0-214-53.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovs-node-l422t                                                        1/1     Running             0          32m   10.0.148.18    ip-10-0-148-18.us-west-1.compute.internal    <none>           <none>
openshift-ovn-kubernetes                           ovs-node-nzcm7                                                        1/1     Running             0          42m   10.0.160.238   ip-10-0-160-238.us-west-1.compute.internal   <none>           <none>
openshift-ovn-kubernetes                           ovs-node-snxvn                                                        1/1     Running             0          32m   10.0.135.2     ip-10-0-135-2.us-west-1.compute.internal     <none>           <none>
```
and network operator also seems okay

```
marketplace                                0.0.1-2020-06-22-201338   True        False         False      36m
monitoring                                 0.0.1-2020-06-22-201338   True        False         False      29m
network                                    0.0.1-2020-06-22-201338   True        False         False      41m
node-tuning                                0.0.1-2020-06-22-201338   True        False         False      40m
```
So all the logs points that these failures are possibly caused by some other issue and not this PR. 

@dcbw do we collect ovsdb-server logs with our must gather or anywhere in the ci logs ? I tried to find, but didn't find one. If not, i think we should collect it, that can give some insight on issues like this.
Retriggering the failed jobs again to see if these events are consistent.

dcbw:
> do we collect ovsdb-server logs with our must gather or anywhere in the ci logs ? I tried to find, but didn't find one. If not, i think we should collect it, that can give some insight on issues like this.
> Retriggering the failed jobs again to see if these events are consistent.

@vishnoianil they should be part of the ovs pod logs, but they won't be very detailed because we dont' do debug logging...

dcbw:
@vishnoianil can we do a retry loop waiting for the pending election timer change to take effect, instead of just exiting with error 11?

dcbw:
retry loop can be a follow-up

PR #616, opened 2020-05-05
Bug 1833372: Migrate from dep to go modules dependency management

vishnoianil:
Modules is an official dependency management system
    starting from v1.14. This new dependency management
    system was introduced in go version 1.11. To build
    current code in go 1.14, you need to explictly set
    GO111MODULE to off, so that it can pickup the legacy dep
    management system.
    I migrated the code to Modules based on the best practices
    suggested in the following links from kubernetes community.
    Please refer these urls for more details.

*.* https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md#motivation
*.* https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away
*.* https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring


vishnoianil:
@dcbw @danwinship @squeed @abhat I opened this PR and my old PR (https://github.com/openshift/cluster-network-operator/pull/605) were getting into too many renaming conflict because of the vendor directory artifacts.

danwinship:
Again, please do "make update-deps" first, commit, and then migrate to go modules on top of that, to make it clear what is just "dependencies that have changed since the last time we updated" vs "things that changed because of using go mod"

vishnoianil:
> Again, please do "make update-deps" first, commit, and then migrate to go modules on top of that, to make it clear what is just "dependencies that have changed since the last time we updated" vs "things that changed because of using go mod"

@danwinship 
Probably a very naive question, but where do you want me to run "make update-deps"? i don't see any Makefile in the CNO code, so that will fail. Really appriciate any pointer here.

danwinship:
oh... huh... we need to migrate it to using the standard build stuff too...
try `dep ensure -update` then

vishnoianil:
> oh... huh... we need to migrate it to using the standard build stuff too...
> try `dep ensure -update` then

Oh yeah, i am trying that, i thought there was something else i need to try. Once we move to Modules, recommended way to udpate is "go mod vendor", so is there any other method we use to update the dependency in general?

pecameron:
@vishnoianil I vaguely remember issues with having the files in vendor vs download. Maybe others remember. A very long time ago I stated on this and we decided to not do it. I don't remember why.

vishnoianil:
@dcbw @danwinship The above two tests are failing because CNO's ci jobs are using go1.13. I pushed following PR to upgrade to go1.13. I have seen other operators are also using go1.13 once they moved to Modules.

https://github.com/openshift/release/pull/8920

vishnoianil:
> @vishnoianil I vaguely remember issues with having the files in vendor vs download. Maybe others remember. A very long time ago I stated on this and we decided to not do it. I don't remember why.

As far as i explored, the main rationale is to avoid build/jobs unstability because of the network issues. Algthough with go 1.14 they provided mirror proxy's for better availability of the artifacts, but still network can drop connection anytime and that can result in build failures, so having a local copy of artifact is helpful in avoiding these kind of failures.

hack/verify-style.sh
```diff
@@ -13,12 +13,7 @@ if [[ ! $(which go) ]]; then
 fi
 if [[ ! $(which golint) ]]; then
   echo "golint not found on PATH. To install:"
-  echo "go get -u github.com/golang/lint/golint"
```
vishnoianil, line 16: @squeed PTAL at this change, with GO Modules, we don't need dep so removed it.

But do we need a check for `which golint`? given that we are not doing any linting check in this script?

vishnoianil:
PT is failing prow/verify test failure because it was trying to install `dep` with go get -u, which is not needed. I removed it through the following patch to openshift/release
https://github.com/openshift/release/pull/8952

Once that PR goes in, all the tests should pass on this PR.

@danwinship @squeed @dave-tucker PTAL, thanks

vishnoianil:
@danwinship Looking at the failures for e2e-windows-hybrid-network job, doesn't seems to be related to this PR.

trozet:
@vishnoianil a sync was done for upstream ovn-k8s to downstream which contains a hybrid overlay fix. Windows passed there:

https://github.com/openshift/ovn-kubernetes/pull/162

That just merged 30 min ago, so I think we just need to rerun and see if it passes with latest code.

/retest

hack/verify-style.sh
squeed: oh, good point, installing golint is silly.

squeed:
You're doing a lot in this PR:
- bumping client-go and controller-runtime from 1.16 to 1.17 (they should be 1.18 if possible)
- upgrading other dependencies
- switching to go modules
- some other random cleanup

Can you split this in to multiple commits in the same PR? That makes it much easier to verify correctness.

danwinship:
> You're doing a lot in this PR:

yeah, I guess the discussion didn't happen in the PR but the problem is that if you run "dep update" it just crashes, so it's difficult to split this into any more commits

squeed:
> if you run "dep update" it just crashes, so it's difficult to split this into any more commits

Oh, great. That's fun. Well, if it's a huge bother, I suppose it doesn't matter that much.

vishnoianil:
> You're doing a lot in this PR:
> 
> * bumping client-go and controller-runtime from 1.16 to 1.17 (they should be 1.18 if possible)
> * upgrading other dependencies
> * switching to go modules
> * some other random cleanup
> 
> Can you split this in to multiple commits in the same PR? That makes it much easier to verify correctness.
@squeed yeah, initially I tried to break it in multiple PRs but these things are bit intertwined :-/. I can still break it in two commits by separting out the clean-up part, but not sure it will make this PR look any better. Sorry for the inconvinience. I am rebasing the branch and see if the gcp and windows hybrid network job passes.

PR #620, opened 2020-05-07
bug 1831006: For third party plugins enable testing of IPv6 single stack

JacobTanenbaum:
since the clusterCIDR is user configurable determine if the
given cluster is IPv4 or IPv6 single stack based on that value
for network plugins. this will allow third party plugins that
do support single stack to function

### DIFF
--- a/pkg/network/kube_proxy.go
+++ b/pkg/network/kube_proxy.go
@@ -121,7 +121,18 @@ func FillKubeProxyDefaults(conf, previous *operv1.NetworkSpec) {
 	}
 
 	if conf.KubeProxyConfig.BindAddress == "" {
-		conf.KubeProxyConfig.BindAddress = "0.0.0.0"
+		// TODO: this will probably need to change for dual stack
+		ip, _, err := net.ParseCIDR(conf.ClusterNetwork[0].CIDR)
+		if err != nil {
+			// this should not happen
+			return
+		}
+		if ip.To4() != nil {
+			conf.KubeProxyConfig.BindAddress = "0.0.0.0"
+		} else {
+			conf.KubeProxyConfig.BindAddress = "::"
+		}
+
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/network/kube_proxy.go
```diff
@@ -121,7 +121,12 @@ func FillKubeProxyDefaults(conf, previous *operv1.NetworkSpec) {
 	}
 
 	if conf.KubeProxyConfig.BindAddress == "" {
-		conf.KubeProxyConfig.BindAddress = "0.0.0.0"
+		// TODO: this will probably need to change for dual stack
+		if net.ParseIP(conf.ClusterNetwork[0].CIDR).To4() != nil {
```
danwinship, line 125: `ParseIP` takes an IP address not a CIDR string so this won't work right.

There are some nicer "is this IPv6" methods in upstream k8s now. Add `utilnet "k8s.io/utils/net"` to the imports, and use `utilnet.IsIPv6CIDRString(conf.ClusterNetwork[0].CIDR)` here. (Well, and put a "`!`" before it, or swap the then and else clauses around.)

JacobTanenbaum: I went to replace `ParseIP` with `ParseCIDR` but the k8s method is much better. Changed this around.

danwinship:
try `dep ensure -update k8s.io/utils` or if that doesn't work `dep ensure k8s.io/utils/net` ?

danwinship:
and if that doesn't work, go back to using `.To4()` :man_shrugging:. @vishnoianil is fixing CNO to use go modules in master, but we're not going to want to backport that to 4.4 so if this PR is getting backported, you don't want to depend on that

JacobTanenbaum:
@danwinship could not seem to get the update dep to work and went back to `.To4()`

PR #624, opened 2020-05-07
[release-4.4] bug 1833048: For third party plugins enable testing of IPv6 single stack

JacobTanenbaum:
master PR: https://github.com/openshift/cluster-network-operator/pull/620

since the clusterCIDR is user configurable determine if the
given cluster is IPv4 or IPv6 single stack based on that value
for network plugins. this will allow third party plugins that
do support single stack to function

### DIFF
--- a/pkg/network/kube_proxy.go
+++ b/pkg/network/kube_proxy.go
@@ -121,7 +121,18 @@ func FillKubeProxyDefaults(conf, previous *operv1.NetworkSpec) {
 	}
 
 	if conf.KubeProxyConfig.BindAddress == "" {
-		conf.KubeProxyConfig.BindAddress = "0.0.0.0"
+		// TODO: this will probably need to change for dual stack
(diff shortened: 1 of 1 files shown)


aojea:
@JacobTanenbaum should we change the status in this bug ?

> expected dependent Bugzilla bug 1831006 to be in one of the following states: VERIFIED, RELEASE_PENDING, CLOSED (ERRATA), but it is ON_QA instead

danwinship:
> should we change the status in this bug ?

The bot's error messages do not distinguish between "your bug is wrong and you need to fix it" vs "your bug is correct but it doesn't allow us to merge this PR right now". This is the latter; the backport until can't be merged until the original master bug has been verified by QE

aojea:
@JacobTanenbaum is this PR still needed or is the bugzilla tha should not be closed?

dcbw:
@knobunc patch manager didn't pick this up last week for some reason... are you going to grab it this week?

PR #627, opened 2020-05-07
OWNERS: add abhat

abhat:
This PR adds abhat to the OWNERS file as reviewer and approver.


knobunc:
Adding valid-bug because this is not appropriate to associate with a bug.

PR #628, opened 2020-05-07
bug 1831680: Want to be able to enable hybrid networking only if HybridOverlayConfig is nil

JacobTanenbaum:
after install time it should be possible to enable hybrid networking but only if
a hybridcluster network is empty


dcbw:
@JacobTanenbaum actually maybe our conversation yesterday was a bit unclear. We don't want to allow enabling anything after install. But I thought we didn't allow enabling Hybrid Overlay unless you explicitly set a ClusterSubnets for it, but we want to allow enabling it without cluster subnets specified (in which case the ovnkube defualt would get used).

JacobTanenbaum:
this is not needed

PR #629, opened 2020-05-08
Bug 1832872: Remove dead member from LB pool.

gryf:
During installation a bootstrap node is used as an API node at first,
but then gets removed. In Kuryr's bootstrap we never remove it from API
loadbalancer pool which leads to Octavia stating that LB is in degraded
state.

With those changes we prevent it by making sure on reconciliation we
remove loadbalancer members that we don't need there anymore. This
should help with scaledown of masters too.

(cherry picked from commit 0ccf62b3f058ebc1e9d8ddfa7a661098b37b1a21)


luis5tb:
holding it as we found one issue that could be triggered due to this... Only until we are sure it is not related

PR #630, opened 2020-05-08
Bug 1833372: Drop 'dep verify' from verify-style

danwinship:
ci/prow/verify is failing in #616 because of weird interactions between "go get" and "migrating to go modules", but we can't fix either side without breaking the other unless we do this first

@squeed @dcbw @rcarrillocruz 
(cc @vishnoianil )


vishnoianil:
@danwinship i removed the dep check in my PR as well, do you want me to remove it ?

vishnoianil:
LGTM

PR #632, opened 2020-05-09
Expose Kuryr metrics endpoints to Prometheus server

MaysaMacedo:
This commit expose the Prometheus endpoints in Kuryr
to the Prometheus server.

### DIFF
--- a/bindata/network/kuryr/005-daemon.yaml
+++ b/bindata/network/kuryr/005-daemon.yaml
@@ -75,6 +75,9 @@ spec:
             port: {{ default 8090 .DaemonProbesPort }}
           initialDelaySeconds: 15
 {{ end }}
+        ports:
+        - name: metrics-port
+          containerPort: 9655
       volumes:
       - name: bin
         hostPath:
--- a/bindata/network/kuryr/monitor.yaml
+++ b/bindata/network/kuryr/monitor.yaml
@@ -1,0 +1,115 @@
+
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app: kuryr-controller
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: monitor-kuryr-controller
+  namespace: openshift-kuryr
+spec:
+  endpoints:
+  - interval: 60s
+    port: metrics
+    scrapeTimeout: 30s
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+    - openshift-kuryr
+  selector:
+    matchLabels:
+      app: kuryr-controller
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kuryr-controller
+  name: kuryr-controller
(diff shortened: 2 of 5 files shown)


bindata/network/kuryr/005-daemon.yaml
```diff
@@ -75,6 +75,9 @@ spec:
             port: {{ default 8090 .DaemonProbesPort }}
           initialDelaySeconds: 15
 {{ end }}
+        ports:
+        - name: metrics-port
+          containerPort: 9103
```
dulek, line 80: Damn, I'm pretty sure one day it'll get taken by another service and we'll never know until we deploy, because we don't have any CI with Kuryr.

bindata/network/kuryr/monitor.yaml
```diff
@@ -0,0 +1,114 @@
...
+metadata:
+  labels:
+    app: kuryr-cni
+  name: kuryr-cni
+  namespace: openshift-kuryr
+spec:
+  selector:
+    app: kuryr-cni
+  clusterIP: None
+  publishNotReadyAddresses: true
```
dulek, line 76: Why?

bindata/network/kuryr/monitor.yaml
```diff
@@ -0,0 +1,114 @@
...
+    - openshift-kuryr
+  selector:
+    matchLabels:
+      app: kuryr-cni
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: kuryr-cni
```
dulek, line 69: How does it work here? It doesn't matter which kuryr-cni gets asked for metrics? I'd say it should be somehow separated per node?

bindata/network/kuryr/005-daemon.yaml
MaysaMacedo: I will add a line for Kuryr in [here](https://github.com/prometheus/prometheus/wiki/Default-port-allocations) with a new port, then we will be safe.

dulek: Ha, nice. :D So that one's already taken.

9409 and 9410 are taken by ovn-kubernetes, so I bet we can reserve a port too.

bindata/network/kuryr/monitor.yaml
MaysaMacedo: All kuryr-cni pods get their metrics collected.

bindata/network/kuryr/monitor.yaml
MaysaMacedo: This service only serves as a way to configure prometheus targets and not really to load-balance. With that field set it's ensured the notReadyAddresses of subsets are published.

bindata/network/kuryr/005-daemon.yaml
MaysaMacedo: I'll allocate other ports on that wiki and here when another round of review is taken here. As stated in the wiki, we must only allocate when ready to share with others.

PR #633, opened 2020-05-11
Bug 1820472: Finalize namespace when it is stuck in 'Terminating' after migration

pliurh:
As controller-runtime client has yet support subresources like `spec\finalizers`,
using client-go to invoke the Finalize API of Namespace.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -7,11 +7,15 @@ import (
 
 	"github.com/pkg/errors"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
 // ApplyObject applies the desired object against the apiserver,
@@ -47,6 +51,30 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
 	}
 	if err != nil {
 		return errors.Wrapf(err, "could not retrieve existing %s", objDesc)
+	}
+
+	if obj.GetKind() == "Namespace" {
+		ns := &v1.Namespace{}
+		if err := kscheme.Scheme.Convert(existing, ns, nil); err != nil {
+			return errors.Wrapf(err, "could not convert namespace %s", objDesc)
+		}
+		if ns.Status.Phase == v1.NamespaceTerminating {
+			log.Printf("finalize Namespace %s", objDesc)
+			ns.Spec.Finalizers = []v1.FinalizerName{}
+			// Get a config to for client-go
+			cfg, err := config.GetConfig()
+			if err != nil {
+				return errors.Wrapf(err, "could not create rest config")
+			}
+			clientset, err := kubernetes.NewForConfig(cfg)
+			if err != nil {
+				return errors.Wrapf(err, "could not create clientset")
+			}
+			if _, err := clientset.CoreV1().Namespaces().Finalize(ns); err != nil {
+				return errors.Wrapf(err, "could not finalize namespace %s", objDesc)
+			}
+			return nil
+		}
 	}
 
 	// Merge the desired object with what actually exists


pkg/apply/apply.go
```diff
@@ -49,6 +53,30 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
+			// Get a config to for client-go
+			cfg, err := config.GetConfig()
+			if err != nil {
+				return errors.Wrapf(err, "could not create rest config")
+			}
+			clientset, err := kubernetes.NewForConfig(cfg)
+			if err != nil {
+				return errors.Wrapf(err, "could not create clientset")
+			}
+			if _, err := clientset.CoreV1().Namespaces().Finalize(ns); err != nil {
```
juanluisvaladas, line 73: I may be wrong, but I think you can do instead:
if _, err := client.CoreV1().Namespaces().Finalize(ns); err 

And avoid instantiating a new client, removing the lines 64-72

pliurh: The `client` here is not a k8s client-go clientset, but a controller-runtime client which doesn't support the `finalize` call of the namespace object. https://github.com/kubernetes-sigs/controller-runtime/issues/573

alexanderConstantinescu: That's not going to work. He wants to access the `Finalize` method, which is not exposed by `k8sclient.Client`, but there might be a smoother way to do this than instantiating a new clientset like this?

alexanderConstantinescu:
I am not sure convinced this should be done in `ApplyObject`, we already have a place for cleaning up old resources which are not deployed by the CNO anymore, that's done in `deleteRelatedObjectsNotRendered` in `status_manager.go`. Maybe adapt that method to do this?

pkg/apply/apply.go
```diff
@@ -49,6 +53,30 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
 	}
 
+	if obj.GetKind() == "Namespace" {
+		ns := &v1.Namespace{}
+		if err := kscheme.Scheme.Convert(existing, ns, nil); err != nil {
+			return errors.Wrapf(err, "could not convert namespace %s", objDesc)
+		}
+		if ns.Status.Phase == v1.NamespaceTerminating {
+			log.Printf("finalize Namespace %s", objDesc)
+			ns.Spec.Finalizers = []v1.FinalizerName{}
```
squeed, line 63: Hang on a moment - you can't just go and delete all finalizers. You should only delete the finalizer for which you are responsible.

squeed:
/hold
I don't think this is quite right - removing *all* finalizers is pretty dangerous.

pkg/apply/apply.go
pliurh: This logic will only be triggered when user hit a problem with OVN after the migration and node reboot. At that time all the openshift API is not available (as OVN doesn't work), so the finalizer of `openshift-sdn` namespace cannot be removed. In such situation, the user cannot rollback, as the `openshift-sdn` is hung in the `Terminating`, new objects cannot be recreated. 
So either we ask user to manually invoke the `finalize` API before rolling back, or we add this logic in the code before recreating the `openshift-sdn` namespace.

pliurh:
> I am not sure convinced this should be done in `ApplyObject`, we already have a place for cleaning up old resources which are not deployed by the CNO anymore, that's done in `deleteRelatedObjectsNotRendered` in `status_manager.go`. Maybe adapt that method to do this?

Normally, it will work. However, if the OVN doesn't work after the migration. The finalizers in the namespace cannot be removed as expect. And the namespaces will be stuck in `Terminating` forever, which prevent user from rolling back to `openshift-sdn`.

alexanderConstantinescu:
> > I am not sure convinced this should be done in `ApplyObject`, we already have a place for cleaning up old resources which are not deployed by the CNO anymore, that's done in `deleteRelatedObjectsNotRendered` in `status_manager.go`. Maybe adapt that method to do this?
> 
> Normally, it will work. However, if the OVN doesn't work after the migration. The finalizers in the namespace cannot be removed as expected. And the namespaces will be stuck in `Terminating` forever, which prevent user from rolling back to `openshift-sdn`.

I understand what you mean, I read the BZ. I am not saying "you should not do this", rather: "you might think about not doing it there and instead in `deleteRelatedObjectsNotRendered`" - because that's where we cleanup old resources left around from previous deployments (which the openshift-sdn namespace qualify as, after an upgrade to ovn-kubernetes)

squeed:
Yeah, as written this can't be merged. We cannot just go deleting all namespace finalizers. That will cause all sorts of unpleasantness, like possibly orphaned pods.

pliurh:
Let me groom the logic a bit. In this patch `finalize` call will only be invoked, when the `openshift-sdn` namespace is stuck in the `Terminating` state, AND the user chose to rollback. So the `openshift-sdn` will be rendered and created by CNO again. So it will neither affect namespace in a normal state nor after successfully migration when the namespace can be deleted as expected.

The reason why we shall not put this logic in `deleteRelatedObjectsNotRendered` is that, we cannot tell whether the openshift-sdn is stuck in `Terminating` forever which requires a force deletion, OR it's just waiting for all the resources in the namespace being deleted.

pliurh:
> Yeah, as written this can't be merged. We cannot just go deleting all namespace finalizers. That will cause all sorts of unpleasantness, like possibly orphaned pods.

There will be no orphaned pods, all the openshift-sdn pods will be removed before the reboot. And this finalized shall only be triggered after the reboot by the rollback operation. However, it could be an issue for the openshift API controlled objects under openshift-sdn namespace. To solve that, either 
1. we don't delete the namespace during the migration, 
2. or we manipulate the order of object deletion in `deleteRelatedObjectsNotRendered`, to make sure the openshift-sdn pods won't be deleted before all the openshift-api objects get deleted.

But I don't know whether option 2 is practical or not.

pliurh:
replace by #641

PR #636, opened 2020-05-11
Bug 1834474: ovnkube: set NB/SB database inactivity probes to 60 seconds

dcbw:
Multiple northds run for HA in active/passive mode where the active
northd holds a lock. If that northd loses connectivity to the database
or is killed without releasing the lock, ovsdb-server will clear
the lock after twice the inactivity probe. But if that probe is set
to 0 (disabled) that will never happen, and a new northd will never
grab the lock and continue reconciling NB->SB.

Set the DB inactivity probe to something greater than 0 to ensure
that a northd will always eventually become active.


danwinship:
/hold
shouldn't merge without passing CI but e2e-gcp-ovn seems currently broken

PR #638, opened 2020-05-12
Bug 1836175: Bump ovn-kind-cno.sh script to use 4.5 image

rcarrillocruz:


hack/ovn-kind-cno.sh
```diff
@@ -88,7 +88,7 @@ if [ "$BUILD_OVN" = true ]; then
   pushd dist/images
   sudo cp -f ../../go-controller/_output/go/bin/* .
   cat << EOF | docker build -t origin-ovn-kubernetes:dev -f - .
-FROM quay.io/openshift/origin-ovn-kubernetes:4.3
+FROM quay.io/openshift/origin-ovn-kubernetes:4.4
```
squeed, line 91: why not 4.5? not that it really matters.

rcarrillocruz: no strong reason, just that 4.4 is what is released

danwinship:
It needs to vary with the CNO branch; CNO master should use ovn-kubernetes master/4.5, CNO `release-4.4` should use ovn-kubernetes 4.4, CNO `release-4.3` should use ovn-kubernetes 4.3.

You might be able to figure out the right version to use automatically with git commands, but it would probably be a lot simpler to just hardcode it on each branch, and bump it in master every time a new release branch splits off.

(so lgtm I think)

alexanderConstantinescu:
> @rcarrillocruz: you cannot LGTM your own PR.

LOL! :D

PR #639, opened 2020-05-13
clarify and tweek the threshold for NetworkPodsCrashlooping

JacobTanenbaum:
clarify the error message and increase the threshold so that the
alert will not fire during upgrade


JacobTanenbaum:
this is actually not our alert that is firing

PR #640, opened 2020-05-14
Support using common name as hostname for TLS server cert verification

abhat:
This change renders the daemonsets for ovnkube-master and ovnkube-node
using the same parameter that is used for the hostname parameter of
the PKI cert generated.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -31,6 +31,7 @@ const OVN_NB_RAFT_PORT = "9643"
 const OVN_SB_RAFT_PORT = "9644"
 const CLUSTER_CONFIG_NAME = "cluster-config-v1"
 const CLUSTER_CONFIG_NAMESPACE = "kube-system"
+const OVN_CERT_CN = "ovn"
 
 // renderOVNKubernetes returns the manifests for the ovn-kubernetes.
 // This creates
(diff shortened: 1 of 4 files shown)


abhat:
This PR replaces #580 and depends on the changes in upstream PR: https://github.com/ovn-org/ovn-kubernetes/pull/1335/commits/80b8f442025d647e42dff8ca9a60d2f3cedf456b

pkg/network/ovn_kubernetes.go
```diff
@@ -30,6 +30,7 @@ const OVN_NB_RAFT_PORT = "9643"
 const OVN_SB_RAFT_PORT = "9644"
 const CLUSTER_CONFIG_NAME = "cluster-config-v1"
 const CLUSTER_CONFIG_NAMESPACE = "kube-system"
+const OVN_CERT_CN = "ovnkube-master"
```
squeed, line 33: The certificate is used for more than just the master - every component uses it! That's ovn-northd, nbdb, sbdb, ovnkube-node, and ovn-controller.

I'd just call it "ovn" :p

squeed:
I suspect this is going to fail ovn CI until the upstream change is merged and merged down.

squeed:
Seems reasonable. Holding off on lgtm until the upstream change is merged down (correct me if it already is)

PR #641, opened 2020-05-15
Bug 1820472: Not delete namespace object when cleanup not rended objects

pliurh:
Namespace deletion cannot be done while openshift-apiserver is down. It will be stuck in 'Terminating' state forever, and that can block the sdn restore. So we choose not to delete the namespace by the cleanup function but leave it to the user to manually remove the namespace after the cluster is back to a normal state.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -85,6 +85,12 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				status.relatedObjects = append(status.relatedObjects, currentObj)
 				continue
 			}
+			if gvk.Kind == "Namespace" && gvk.Group == "" {
+				// BZ 1820472: During SDN migration, deleting a namespace object may get stuck in 'Terminating' forever if the cluster network doesn't working as expected.
+				// We choose to not delete the namespace here but to ask user do it manually after the cluster is back to normal state.
+				log.Printf("Object Kind is Namespace, skip")
+				continue
+			}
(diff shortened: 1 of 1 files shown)


pliurh:
@squeed @alexanderConstantinescu PTAL.

alexanderConstantinescu:
> @squeed @alexanderConstantinescu PTAL.

I am not sure if one can chose to go from "multus deployment -> non multus deployment", which would have the multus namespace linger around with this patch. 

@squeed, any input on that?

squeed:
> > @squeed @alexanderConstantinescu PTAL.
> 
> I am not sure if one can chose to go from "multus deployment -> non multus deployment", which would have the multus namespace linger around with this patch.
> 
> @squeed, any input on that?

That's correct - disabling multus after-the-fact is not supported.

I think this is a better approach; could you add a quick comment in the code as to why this was added, then I'll lgtm it?

pliurh:
@squeed PTAL

pkg/controller/statusmanager/status_manager.go
```diff
@@ -85,6 +85,12 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				status.relatedObjects = append(status.relatedObjects, currentObj)
 				continue
 			}
+			if gvk.Kind == "Namespace" {
```
squeed, line 88: Check that gvk.Group == "core" too.

pliurh: In the namespace `gvk` object, the group is empty string, so I change use `gvk.Group == ""` here.

PR #645, opened 2020-05-21
Bug 1838251: The DHCP CNI daemonset should mount the netns paths

dougbtv:
Otherwise, it can fail with out being able to locate the netns paths under OCP 4.5

### DIFF
--- a/bindata/network/multus/003-dhcp-daemon.yaml
+++ b/bindata/network/multus/003-dhcp-daemon.yaml
@@ -52,6 +52,9 @@ spec:
           mountPath: /host/run/cni
         - name: procpath
           mountPath: /host/proc
+        - name: netnspath
+          mountPath: /host/var/run/netns
+          mountPropagation: HostToContainer
       volumes:
         - name: socketpath
           hostPath:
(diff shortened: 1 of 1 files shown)


dougbtv:
/retitle Bug 1838251: The DHCP CNI daemonset should mount the netns paths

bindata/network/multus/003-dhcp-daemon.yaml
```diff
@@ -52,11 +52,19 @@ spec:
           mountPath: /host/run/cni
         - name: procpath
           mountPath: /host/proc
+        - name: netnspath
+          mountPath: /host/var/run/netns
+          mountPropagation: Bidirectional
```
dougbtv, line 57: Just an FYI that this was required otherwise I was getting complaints about wrong fs magic.

s1061123: Bidirectional or HostToContainer could be okey. see: https://github.com/openshift/cluster-network-operator/pull/579/files

danwinship: Yes, please make it `HostToContainer` if that works. More securer.

dougbtv:
Pushed new commit with `mountPropagation: HostToContainer`

PR #647, opened 2020-05-22
fix up dual-stack Service reconciling

danwinship:
We should not attempt to overwrite Service.Spec.IPFamily, if it's set (in a dual-stack cluster)

### DIFF
--- a/pkg/apply/merge.go
+++ b/pkg/apply/merge.go
@@ -76,7 +76,7 @@ func MergeDeploymentForUpdate(current, updated *uns.Unstructured) error {
 	return nil
 }
 
-// MergeServiceForUpdate ensures the clusterip is never written to
+// MergeServiceForUpdate ensures the ClusterIP/IPFamily is never modified
 func MergeServiceForUpdate(current, updated *uns.Unstructured) error {
 	gvk := updated.GroupVersionKind()
 	if gvk.Group == "" && gvk.Kind == "Service" {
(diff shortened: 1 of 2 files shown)


danwinship:
/hold
to stop retests since I don't plan for this to merge until after 4.6 opens anyway

pkg/apply/merge.go
```diff
@@ -76,18 +76,25 @@ func MergeDeploymentForUpdate(current, updated *uns.Unstructured) error {
...
 		if found {
 			return uns.SetNestedField(updated.Object, clusterIP, "spec", "clusterIP")
 		}
+
+		ipFamily, found, err := uns.NestedString(current.Object, "spec", "ipFamily")
+		if err != nil {
+			return err
+		}
+		if found {
+			return uns.SetNestedField(updated.Object, ipFamily, "spec", "ipFamily")
```
abhat, line 96: it would be desirable to add a log statement to indicate that the `Service.Spec.ipFamily` was not updated as a result of the merge.

abhat:
Probably worth considering updating `TestMergeService` to capture this behavior?

danwinship:
> Probably worth considering updating `TestMergeService` to capture this behavior?

Ha! Definitely, given that it showed that the previous version didn't actually work. Fixed now.

pkg/apply/merge.go
danwinship: The point of not updating it is that it's immutable and so trying to update it would be guaranteed to fail, so logging that we're not updating it doesn't provide any information.

(I guess I didn't totally explain the bug; the problem is that our template `Service` objects don't have `ipFamily` set in them, but once the service is created, it will get a default value of `IPv4`, but the way the code is currently written, we then end up trying to overwrite that with `ipFamily: ""` like in the YAML template, and that results in an error. The fix is to just ignore `ipFamily` on updates, since it can't be changed anyway.)

abhat: OK, that makes sense. If the field itself is immutable, there is little value. My point was to simply add a DBG log statement saying any updates to the `ipFamily` will be ignored because the field is immutable. But I guess I am OK with it either way.

PR #648, opened 2020-05-25
Bug 1838251: mount /var/run/netns rslave in DHCP daemonset

s1061123:
To fix that DHCP pod cannot get netns info due to #579, need to change the mount option for DHCP CNI server. This fix changes it.


s1061123:
Dupped to #645

PR #650, opened 2020-05-26
Bug 1837575: Allow connection between authentication net namespaces

rcarrillocruz:
Fixes #1837575

### DIFF
--- a/bindata/network/openshift-sdn/004-multitenant.yaml
+++ b/bindata/network/openshift-sdn/004-multitenant.yaml
@@ -71,6 +71,8 @@ netname: openshift-user-workload-monitoring
 # - openshift-service-catalog-controller-manager
 # - openshift-template-service-broker
 # - openshift-ansible-service-broker
+# - openshift-authentication
+# - openshift-authentication-operator
 apiVersion: network.openshift.io/v1
 kind: NetNamespace
 metadata:
@@ -126,4 +128,19 @@ metadata:
 netid: 1
 netname: openshift-service-catalog-controller-manager
 
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-authentication
+netid: 1
+netname: openshift-authentication
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-authentication-operator
+netid: 1
+netname: openshift-authentication-operator
 {{- end}}


squeed:
Hey @rcarrillocruz didn't we set up a CI job that tests multitenant? Is that being run here?

rcarrillocruz:
> Hey @rcarrillocruz didn't we set up a CI job that tests multitenant? Is that being run here?

I'm sure I did, it probably isn't enabled on every PR tho.
Will check on release repo and try manually if necessary.

squeed:
I seem to recall there being something weird about moving network namespace IDs... Since this has shipped, we might have some difficulties. Do you recall, @danwinship 

In any case, make sure that QE tests the upgrade case.

danwinship:
> > Hey @rcarrillocruz didn't we set up a CI job that tests multitenant? Is that being run here?
> 
> I'm sure I did, it probably isn't enabled on every PR tho.
> Will check on release repo and try manually if necessary.

There are periodic jobs, which are release-informing. eg, https://testgrid.k8s.io/redhat-openshift-ocp-release-4.5-informing#release-openshift-origin-installer-e2e-aws-sdn-multitenant-4.5

So that implies that we are not testing anything that gets broken by this bug I guess

> I seem to recall there being something weird about moving network namespace IDs... Since this has shipped, we might have some difficulties. Do you recall, @danwinship

sdn-controller does not watch for changes to NetNamespaces; it assumes that it is the only one modifying them.

In practice this probably doesn't matter, since CNO will update the NetNamespaces and the sdn-controller Deployment at about the same time, and and then sdn-controller will restart, and see the updated VNIDs on startup.

bindata/network/openshift-sdn/004-multitenant.yaml
```diff
@@ -126,4 +128,19 @@ metadata:
...
+netid: 1
+netname: openshift-authentication
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-authentication-operator
+netid: 1
+netname: openshift-authentication-operator
```
sttts, line 145: when do we need such a rule? For every cross-namespace communication?

sttts: Isn't it questionable to have these rules here in the network operator? The network operator does not know all software running on the cluster. Compare with prometheus monitor manifests, which were moved into the monitored components repos.

danwinship: I remember having talked about the idea of components putting annotations on their Namespaces but I don't remember why we did it this way instead. I think we wanted to avoid having other components have to worry about openshift-sdn multitenant mode because it's more-or-less deprecated anyway. (It's difficult to enable in openshift-sdn in 4.x, and it won't exist at all in ovn-kubernetes.)

danwinship: > when do we need such a rule? For every cross-namespace communication?

several of the more exciting namespaces (eg, openshift-monitoring, openshift-kube-apiserver) are made global so that anyone can talk to them and they can talk to anyone, and that covers almost all of the interactions.

sttts: ack

PR #652, opened 2020-05-28
Bug 1835497 : Prevent sb/nb db readiness probe from being zombie

vishnoianil:
Current implementation of sb/nb db containers readiness probe uses
exec to call the ovn-appctl to check the status of the sb/nb db raft
cluster. This exec call will be executed by another exec call that is
used to execute the probe. If sb/nb db raft cluster is busy and readiness
probe is fired, it will be stuck till the raft cluster respond with
success or failure. It might take time in seconds (even more than
`periodSeconds` time). Next readiness probe executes and kills the parent
that makes the child processes zombie and they never reap. With single
readiness probe failure it creates 3 zombie threads.

With higher scale, raft cluster can be busy for long duration
(10 to 100s depending on the load) and readiness probe will fail
more frequently and that can increase the zombie threads and slowly
 eat up the pthread_thread quota and newly scheduled or rebooted pods
 on the nodes might not be able to create threads and fails.

In this reported bug, sb raft cluster was not in healthy state and
CNO endup rebooting the node, but it fails to boot because it was not
allowed to create anymore threads, and this raft cluster never recovered.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -163,13 +163,14 @@ spec:
                 fi
         readinessProbe:
           initialDelaySeconds: 30
+          timeoutSeconds: 5
           exec:
             command:
             - /bin/bash
             - -c
             - |
               set -xe
-              exec /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound  2>/dev/null | grep ${K8S_NODE_IP} | grep -v Address -q
+              /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=3 cluster/status OVN_Northbound  2>/dev/null | grep ${K8S_NODE_IP} | grep -v Address -q
         env:
         - name: OVN_LOG_LEVEL
           value: info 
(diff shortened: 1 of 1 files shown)


vishnoianil:
@dcbw @abhat @squeed PTAL, thanks

squeed:
Good catch.

Why such a low timeout? Can you think of a better readiness probe?

vishnoianil:
> Good catch.
> 
> Why such a low timeout? Can you think of a better readiness probe?

@squeed 
In normal conditions, probe (ovn-appctl) should return in millisecond time, so if it doesn't respond in 3 second that means it's busy and might be busy for longer duration. Given that the probe runs periodically with 10 second interval and we wait for 3 failures to mark the container in unhleathy state, that overall gives us 30 second to check on the cluster status and mark it unhealthy, which i believe is resonable time duration?

With my experiment with horizontal scale test (increasing number of nodes), i have seen that the raft cluster can get busy for 2-10 seconds range for 200 nodes, so there is good chances that the ovn-appctl might be stuck for that duration and more frequently. So rather than ovn-appctl stuck for longer time, i think leveraging the `periodSeconds` gives us fast-fail and check-again method, which i believe will give us more deterministic readiness status.

If you have any thoughts to improve it, please share, we can evaluate that as well.

PR #655, opened 2020-05-29
Bug 1838343 - Improve the sb-db and nb-db readiness check to ensure it fails when cluster is not stable.

vishnoianil:


vishnoianil:
retest

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -744,15 +749,19 @@ spec:
...
         readinessProbe:
-          initialDelaySeconds: 60
+          initialDelaySeconds: 90
           timeoutSeconds: 5
           exec:
             command:
             - /bin/bash
             - -c
             - |
-              set -xe
```
squeed, line 754: We should probably have `set  -eo pipefail`, so if ovn-appctl fails, we still fail the readiness check.

vishnoianil: @squeed It will fail because in case of ovn-appctl failure, leader_status will be empty as well.

vishnoianil: @squat i take that back. You are right, check will pass and readiness probe will be success. I removed it because if grep doesn't match to any string, it return exit code 1, and that actually was failing the readiness probe as well, which it shouldn't.

squat: cc @squeed

vishnoianil:
@abhat @dcbw PTAL, Thanks

squeed:
Seems like a reasonable addition, and I think its a good idea.

@numansiddique does this look like a reasonable way to confirm the nbdb/sbdb is healthy and part of a cluster?

squeed:
Did some checks with this, I think it's good.
/lgtm
/approve

numansiddique:
> Seems like a reasonable addition, and I think its a good idea.
> 
> @numansiddique does this look like a reasonable way to confirm the nbdb/sbdb is healthy and part of a cluster?

Oops. I missed the notification. I agree. Its seems a reasonable way to me.

PR #656, opened 2020-06-01
Bug 1846396: Drop ovn-octavia provider limitation for multiprotocol listeners

luis5tb:
The ovn-octavia limitation of having several listener protocols
on the same port has been fixed, so the exception made for it
is not needed anymore.


dulek:
Any reason to keep it on hold?

luis5tb:
> Any reason to keep it on hold?

We need to merge first the kuryr-controller side

PR #660, opened 2020-06-04
Improve IPv6 testing and minor fixes

aojea:
Aggregated several commits in the PR because I was finding several issues.

It was reported that the cluster-network-operator wasn't configuring correctly kube-proxy bind-address if it was "::", so I added some unit tests to check that it works correctly.

The other fixes are explained in the commit messages.

### DIFF
--- a/pkg/network/kube_proxy_test.go
+++ b/pkg/network/kube_proxy_test.go
@@ -38,6 +38,35 @@ var config = operv1.NetworkSpec{
 	},
 }
 
+var configIPv6 = operv1.NetworkSpec{
+	ClusterNetwork: []operv1.ClusterNetworkEntry{
+		{
+			CIDR:       "fd00:1234::/48",
+			HostPrefix: 64,
+		},
+	},
+	KubeProxyConfig: &operv1.ProxyConfig{
+		BindAddress:        "::",
+		IptablesSyncPeriod: "1m",
+		ProxyArguments: map[string]operv1.ProxyArgumentList{
+			// string
+			"proxy-mode": {"blah"},
+
+			// duration
+			"iptables-min-sync-period": {"2m"},
+
+			// optional int
+			"iptables-masquerade-bit": {"14"},
+
+			// optional duration
+			"conntrack-tcp-timeout-close-wait": {"10m"},
+
+			// This will be overridden
+			"conntrack-max-per-core": {"5"},
+		},
+	},
+}
+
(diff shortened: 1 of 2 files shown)


rcarrillocruz:
you may want to rebase to single commit

aojea:
> you may want to rebase to single commit

I prefer not to, since every commit is totally independent and, in case of an error, it will be easier to revert just the offending one.

pkg/network/kube_proxy_test.go
```diff
@@ -98,6 +127,66 @@ winkernel:
...
+	errs := validateKubeProxy(&configIPv6)
+	g.Expect(errs).To(HaveLen(0))
+
+	cfg, err := kubeProxyConfiguration(map[string]operv1.ProxyArgumentList{
+		// special address+port combo
+		"metrics-bind-address":   {"fd00:1234::4"},
+		"metrics-port":           {"999"},
+		"conntrack-max-per-core": {"10"},
+	},
+		&configIPv6,
```
danwinship, line 142: I don't like the way this indents; can you add a newline before `map` so the rest of the map body up to the `},` indents another level?

pkg/network/kube_proxy_test.go
```diff
@@ -98,6 +127,66 @@ winkernel:
...
+func TestKubeProxyIPv6Config(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	errs := validateKubeProxy(&configIPv6)
+	g.Expect(errs).To(HaveLen(0))
+
+	cfg, err := kubeProxyConfiguration(map[string]operv1.ProxyArgumentList{
+		// special address+port combo
+		"metrics-bind-address":   {"fd00:1234::4"},
+		"metrics-port":           {"999"},
```
danwinship, line 139: use a port above 32767 here to test that fix?

danwinship:
> > you may want to rebase to single commit
> 
> I prefer not to, since every commit is totally independent and, in case of an error, it will be easier to revert just the offending one.

yeah. We don't want PRs with one real commit and a bunch of fixes to it, but if the commit is logically made up of multiple pieces, go with multiple commits

pkg/network/kube_proxy_test.go
aojea: done,

pkg/network/kube_proxy_test.go
aojea: done

PR #661, opened 2020-06-04
Bug 1824983: explicitly delete pidfiles when exiting

JacobTanenbaum:
saw this error message in the logs

Terminated:&ContainerStateTerminated{ExitCode:1,Signal:0,Reason:Error,Message:ovsdb-server: /var/run/openvswitch/ovsdb-server.pid: pidfile check failed (No such process)

when it looked like ovs was restarting causing the sdn pods to spin. explicitly delete

/var/run/openvswitch/ovs-vswitchd.pid and
/var/run/openvswitch/ovsdb-server.pid:

on ovs pod exit

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -67,6 +67,12 @@ spec:
           }
           trap quit SIGTERM
 
+          function deletePid {
+              rm /var/run/openvswitch/ovs-vswitchd.pid
+              rm /var/run/openvswitch/ovsdb-server.pid
+          }
+          trap deletePid EXIT
+
           # launch OVS
           # Start the ovsdb so that we can prep it before we start the ovs-vswitchd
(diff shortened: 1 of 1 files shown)


abhat:
> saw this error message in the logs
> 
> Terminated:&ContainerStateTerminated{ExitCode:1,Signal:0,Reason:Error,Message:ovsdb-server: /var/run/openvswitch/ovsdb-server.pid: pidfile check failed (No such process)
> 
> when it looked like ovs was restarting causing the sdn pods to spin. explicitly delete
> 
> /var/run/openvswitch/ovs-vswitchd.pid and
> /var/run/openvswitch/ovsdb-server.pid:
> 
> on ovs pod exit

So this fix seems to be counter-intuitive to the error message. The error message is saying that the pod termination failed because the pid file wasn't found. And we are trying to fix that by adding a `EXIT` handler that deletes the pid? Can you please explain the need for the change?

JacobTanenbaum:
@abhat it is more important that it runs on the previous iteration of the ovs pod. We want to make sure that we clean up the pidfile before another container tries to start since /var/run is mounted from the host when the container exits if the file remains it causes problems like this

PR #663, opened 2020-06-07
use KIND multinode cluster , add IPv6 support, fix multus webook problem

aojea:
for networking development and testing, using single node
environments mask a big part of the problems.

We should use multinode by default if possible.

Another improvements:

* the apiserver autodiscovers the API address, so we don't
  need to pass the kubeconfig

* fix multus webook problem

* bump KIND version to 0.8.1

* allow to configure IPv6 environments

* use kubectl wait instead of bash loops

### DIFF
--- a/hack/ovn-kind-cno.sh
+++ b/hack/ovn-kind-cno.sh
@@ -2,17 +2,27 @@
 set -euo pipefail
 
 # Version v1.17.0 or higher is required
-K8S_VERSION=${K8S_VERSION:-v1.17.0}
+K8S_VERSION=${K8S_VERSION:-v1.18.2}
 BUILD_OVN=${BUILD_OVN:-false}
 BUILD_CNO=${BUILD_CNO:-false}
 BUILD_MULTUS=${BUILD_MULTUS:-false}
 CNO_PATH=${CNO_PATH:-$GOPATH/src/github.com/openshift/cluster-network-operator}
 OVN_K8S_PATH=${OVN_K8S_PATH:-$GOPATH/src/github.com/ovn-org/ovn-kubernetes}
-CLUSTER_CIDR=${CLUSTER_CIDR:-"172.16.0.0/16"}
-SERVICE_NETWORK=${SERVICE_NETWORK:-"172.30.0.0/16"}
-# Skip the comment lines and retrieve the number of Master nodes from kind.yaml file.
-NUM_MASTER_NODES=`grep "^[^#]" kind.yaml | grep -c "role\: control-plane"`
-OVN_KIND_VERBOSITY=${OVN_KIND_VERBOSITY:-0}
+KIND_CONFIG=${KIND_CONFIG:-$HOME/kind-ovn-config.yaml}
+export KUBECONFIG=${HOME}/kube-ovn.conf
+NUM_MASTER_NODES=1
+OVN_KIND_VERBOSITY=${OVN_KIND_VERBOSITY:-5}
+
+# Default networks (same as in KIND)
+if [ "${IP_FAMILY:-ipv4}" = "ipv6" ]; then
+  CLUSTER_CIDR=${CLUSTER_CIDR:-"fd00:10:244::/48"}
+  SERVICE_NETWORK=${SERVICE_NETWORK:-"fd00:10:96::/112"}
+  HOST_PREFIX=64
+else
+  CLUSTER_CIDR=${CLUSTER_CIDR:-"10.244.0.0/16"}
+  SERVICE_NETWORK=${SERVICE_NETWORK:-"10.96.0.0/12"}
+  HOST_PREFIX=24
+fi
 
 # Check for docker
 if ! command -v docker; then
@@ -26,38 +36,30 @@ if ! command -v kubectl; then
   exit 1
 fi
 
-# Detect IP to use as API server
-API_IP=$(ip -4 addr | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | grep -v "127.0.0.1" | head -n 1)
-if [ -z "$API_IP" ]; then
-  echo "Error detecting machine IP to use as API server"
-  exit 1
-fi
-
-sed -i "s/apiServerAddress.*/apiServerAddress: ${API_IP}/" kind.yaml
-
 # Ensure reachability to host via Docker network
 if ! sudo iptables -C DOCKER-USER -j ACCEPT > /dev/null 2>&1; then 
   sudo iptables -I DOCKER-USER -j ACCEPT
 fi
 
+ # create the config file
+  cat <<EOF > ${KIND_CONFIG}
+# config for 1 control plane node and 2 workers (necessary for conformance)
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: ${IP_FAMILY:-ipv4}
+  disableDefaultCNI: true
+  podSubnet: ${CLUSTER_CIDR:-10.244.0.0/16}
+  serviceSubnet: ${SERVICE_NETWORK:-10.96.0.0/12}
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+EOF
+
+
 # Create KIND cluster
-kind create cluster --name ovn --kubeconfig ${HOME}/admin.conf --image kindest/node:${K8S_VERSION} --config=./kind.yaml -v ${OVN_KIND_VERBOSITY}
-export KUBECONFIG=${HOME}/admin.conf
-mkdir -p /tmp/kind
-sudo chmod 777 /tmp/kind
-count=0
-until kubectl get secrets -o jsonpath='{.items[].data.ca\.crt} &> /dev/null'
-do
-  if [ $count -gt 10 ]; then
-    echo "Failed to get k8s crt/token"
-    exit 1
-  fi
-  count=$((count+1))
-  echo "secrets not available on attempt $count"
-  sleep 5
-done
-kubectl get secrets -o jsonpath='{.items[].data.ca\.crt}' > /tmp/kind/ca.crt
-kubectl get secrets -o jsonpath='{.items[].data.token}' > /tmp/kind/token
+kind create cluster --name ovn --image kindest/node:${K8S_VERSION} --config=${KIND_CONFIG} -v ${OVN_KIND_VERBOSITY}
 
 echo -e "\n"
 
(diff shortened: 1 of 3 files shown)


hack/ovn-kind-cno.sh
```diff
@@ -26,38 +35,30 @@ if ! command -v kubectl; then
   exit 1
 fi
 
-# Detect IP to use as API server
```
aojea, line 29: The apiserver "autodectects" its external IP, and since we only have one interface in the nodes, there is no problem with "choosing" the wrong one

hack/ovn-kind-cno.sh
```diff
@@ -26,38 +35,30 @@ if ! command -v kubectl; then
...
+  serviceSubnet: ${SERVICE_CIDR:-10.96.0.0/12}
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+EOF
+
+
 # Create KIND cluster
-kind create cluster --name ovn --kubeconfig ${HOME}/admin.conf --image kindest/node:${K8S_VERSION} --config=./kind.yaml -v ${OVN_KIND_VERBOSITY}
```
aojea, line 44: because the controller is running in a pod, is able to get the right configuration without having to pass the kubeconfig

hack/ovn-kind-cno.sh
```diff
@@ -198,18 +183,14 @@ spec:
...
-    exit 1
-  fi
-  echo "OVN pod not available yet on attempt $count"
-  count=$((count+1))
-  sleep 10
-done
-sleep 10
+# Configuring secret for multus-admission-webhook
+# https://raw.githubusercontent.com/openshift/multus-admission-controller/master/hack/webhook-create-signed-cert.sh
+$CNO_PATH/hack/webhook-create-signed-cert.sh --service multus-admission-controller --namespace openshift-multus --secret multus-admission-controller-secret
```
aojea, line 188: multus admission webhook was failing because it was trying to mount the certificates from this secret, don't know where this has to be fixed though

hack/webhook-create-signed-cert.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
+# Original script found at: https://github.com/morvencao/kube-mutating-webhook-tutorial/blob/master/deployment/webhook-create-signed-cert.sh
```
aojea, line 2: I've found this here https://raw.githubusercontent.com/openshift/multus-admission-controller/master/hack/webhook-create-signed-cert.sh

manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -19,7 +19,6 @@ spec:
         image: quay.io/openshift/origin-cluster-network-operator:4.3
         command:
         - "/usr/bin/cluster-network-operator"
-        - "--url-only-kubeconfig=/etc/kubernetes/kubeconfig"
```
aojea, line 22: in KIND we don't need it, the controller is able to get the config from the pod environment.
Should I add a new deployment yaml and keep this?

aojea:
/assign @trozet @dcbw @danwinship 

Please bear in mind that my knowledge of the cluster-network-operator and openshift is minimum, so maybe some of the changes I added to the doc/script does not make sense in that context.
I especially confused about the apiserver IP, since the apiserver "autodiscovers" it, but the code mentions something a bootstrap problem, something is not happening with KIND, is this an openshift specific issue?

hack/webhook-create-signed-cert.sh
aojea: this makes me think if this is a bug and the operator should take care or check that the secret is created with the tls certificates

aojea: filed a bug because I could see that the kuryr code is creating the secret for its admission webhook
https://bugzilla.redhat.com/show_bug.cgi?id=1845033

squeed: The secret is created by the service-serving-cert controller, part of the openshift control plane.

squeed:
Ah, so you're deploying a vanilla kubernetes / KIND cluster and running the CNO on it? Neat. But tricky.

It mostly works, because the CNO needs to deploy the core networking functionality before the openshift-specific components of the cluster are running. This means we need to have a two-phase rollout in place - first the deployable set of components, then wait, then deploy the rest. In reality we're not that clever - instead, we just step over failures and set our status to Deploying until it all resolves.

So you're seeing the consequences of that decision: CNO + kind works enough to get the network up, but can't ever run to completion. It wants things like monitoring and the service-serving-cert.

So, is it worth implementing enough of the openshift functionality so that we can continue? I'm not sure; it's a lot of work. Probably better to just say "it will give you basic functionality, but don't expect everything".

danwinship:
> Ah, so you're deploying a vanilla kubernetes / KIND cluster and running the CNO on it? Neat. But tricky.

Note that this already exists; @trozet did the initial implementation a while back. This is just an update to what's already there.

danwinship:
>  I especially confused about the apiserver IP, since the apiserver "autodiscovers" it, but the code mentions something a bootstrap problem, something is not happening with KIND, is this an openshift specific issue?

The issue is that when you start CNO in a pod, kubelet will tell it `KUBERNETES_SERVICE_HOST=172.30.0.1` (or whatever the clusterIP of kubernetes.default is), but that doesn't actually work because there's no kube-proxy yet. We need to tell CNO what the actual direct apiserver IP is. In OCP there is a manually-maintained cloud loadbalancer IP that points back to the 3 masters, but here since there's only a single master you can just pass the IP of that node.

hack/webhook-create-signed-cert.sh
aojea: I closed the bug, thanks for the clarifications from you and Dan Winship

aojea:
/hold

need to fix this https://github.com/openshift/cluster-network-operator/pull/663#issuecomment-640604520

aojea:
/hold cancel

it works for IPv4, tested running e2e tests against the KIND cluster
PTAL

aojea:
and it works with IPv6 only clusters:

```
and it works with IPv6
kubectl get services -A
NAMESPACE                  NAME                          TYPE        CLUSTER-IP         EXTERNAL-IP   PORT(S)                  AGE
default                    kubernetes                    ClusterIP   fd00:10:96::1      <none>        443/TCP                  27m
kube-system                kube-dns                      ClusterIP   fd00:10:96::a      <none>        53/UDP,53/TCP,9153/TCP   26m
openshift-multus           multus-admission-controller   ClusterIP   fd00:10:96::c166   <none>        443/TCP,8443/TCP         18m
openshift-multus           network-metrics-service       ClusterIP   None               <none>        8443/TCP                 18m
openshift-ovn-kubernetes   ovn-kubernetes-master         ClusterIP   None               <none>        9102/TCP                 18m
openshift-ovn-kubernetes   ovn-kubernetes-node           ClusterIP   None               <none>        9103/TCP                 18m
openshift-ovn-kubernetes   ovnkube-db                    ClusterIP   None               <none>        9641/TCP,9642/TCP        18m
```

hack/ovn-kind-cno.sh
```diff
@@ -145,22 +149,9 @@ if [ "$BUILD_OVN" = true ] || [ "$BUILD_CNO" = true ]; then
...
-    echo "Network operator not running"
-    exit 1
-  fi
-  echo "Network operator pod not available yet on attempt $count"
-  count=$((count+1))
-  sleep 10
-
-done
-
-CNO_POD=$(kubectl get pod -n openshift-network-operator -o jsonpath='{.items[0].metadata.name}' --field-selector status.phase=Running)
```
nerdalert, line 160: `$CNO_POD` is still getting referenced at but undeclared after being deleted https://github.com/openshift/cluster-network-operator/blob/3346964e91c54d8577f5e54425b982701a86118d/hack/ovn-kind-cno.sh#L159 

I added it back in and the script seems to continue ok. Ty.

aojea:
@nerdalert if you are working on this, feel free to reuse the parts that are valid of this PR and send everything consolidated in one PR.

hack/ovn-kind-cno.sh
aojea: upps, I cleaned too much, it was clearly and oversight on my side 😄

aojea: recovered

nerdalert:
> I was able to get things up and running with `BUILD_OVN=true BUILD_CNO=true ./ovn-kind-cno.sh`.
> 
> I ran into issues when not building CNO with the options `BUILD_OVN=true ./ovn-kind-cno.sh` (also had to add the `$CNO_POD` back in to get this option to run):
> 
> ```
> Creating "cluster-config-v1" configMap with 1 master nodes
> configmap/cluster-config-v1 created
> Creating OVN CNO config
> network.config.openshift.io/cluster created
> Sym-linking cni dirs for node 9ef04a9e4197
> Sym-linking cni dirs for node 24402e252cbf
> Sym-linking cni dirs for node 93a200121f79
> 
> pod/ovs-node-dgsbq condition met
> pod/ovs-node-jgdrx condition met
> pod/ovs-node-nvnks condition met
> timed out waiting for the condition on pods/ovnkube-master-8khjb
> timed out waiting for the condition on pods/ovnkube-node-h89h5
> timed out waiting for the condition on pods/ovnkube-node-krhdv
> timed out waiting for the condition on pods/ovnkube-node-xtblf
> OVN-k8s pods are not running
> ```
> 
> Logs at https://gist.github.com/nerdalert/451896250569e0852e28a907a35198d7
> 
> Thanks for this. I'm going to poke around on getting hybrid overlay enabled. Ty!

Still stumped on why `BUILD_CNO=true BUILD_OVN=true ./ovn-kind-cno.sh` builds ok but `BUILD_OVN=true ./ovn-kind-cno.sh` does not. I went through and ran the same build with master and was able to reproduce the error there as well, so its not related to this PR. Everything looks like its working for me with regard to ovn-kubernetes for me passing the build ENVs.

LGTM with the caveat I don't know enough about Multus to say if its good to go there or not. Ty!

aojea:
/retest 
PTAL

aojea:
> Still stumped on why BUILD_CNO=true BUILD_OVN=true ./ovn-kind-cno.sh builds ok but BUILD_OVN=true ./ovn-kind-cno.sh does not. I went through and ran the same build with master and was able to reproduce the error there as well, so its not related to this PR. Everything looks like its working for me with regard to ovn-kubernetes for me passing the build ENVs.

same here, but if the images are different maybe the one that is used when is not built is outdated

```
    Image:         origin-ovn-kubernetes:dev
    Image ID:      sha256:6e08e12de795a32d90a0a4ddcd78ce45383232ff28e0e7762d89e8d87f02884f
```

> LGTM with the caveat I don't know enough about Multus to say if its good to go there or not. Ty!

great
/retest

aojea:
this job ci/prow/e2e-vsphere — Job failed. NEVER succeded :upside_down_face: 
does it make sense to run it?

juanluisvaladas:
/lgtm
I've using this branch for a while, it works great. e2e-vsphere is broken anyway and not required.
@danwinship maybe we can override it?

danwinship:
> e2e-vsphere is broken anyway and not required.
> @danwinship maybe we can override it?

if it's not required it doesn't need to be overridden

/lgtm

PR #668, opened 2020-06-15
Mark operator as degraded if there are any pods in CrashLoopBackOff State

tssurya:
This patch checks for the statuses of pods of those deployments and daemonsets
which are in a "hung" state. If any of the pods are in CrashLoopBackOff state
the operator will be marked as degraded.

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/names"
 
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -91,6 +92,10 @@ func (status *StatusManager) SetFromPods() {
 		} else if ds.Status.NumberUnavailable > 0 {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
 			dsProgressing = true
+			// Check for any pods in CrashLoopBackOff state and mark the operator as degraded if so.
+			if !isNonCritical(ds) {
+				hung = append(hung, status.CheckCrashLoopBackOffPods(dsName, ds.Spec.Selector.MatchLabels, "DaemonSet")...)
+			}
 		} else if ds.Status.NumberAvailable == 0 { // NOTE: update this if we ever expect empty (unscheduled) daemonsets ~cdc
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not yet scheduled on any nodes", dsName.String()))
 			dsProgressing = true
@@ -138,6 +143,10 @@ func (status *StatusManager) SetFromPods() {
 		if dep.Status.UnavailableReplicas > 0 {
 			progressing = append(progressing, fmt.Sprintf("Deployment %q is not available (awaiting %d nodes)", depName.String(), dep.Status.UnavailableReplicas))
 			depProgressing = true
+			// Check for any pods in CrashLoopBackOff state and mark the operator as degraded if so.
+			if !isNonCritical(dep) {
+				hung = append(hung, status.CheckCrashLoopBackOffPods(depName, dep.Spec.Selector.MatchLabels, "Deployment")...)
+			}
 		} else if dep.Status.AvailableReplicas == 0 {
 			progressing = append(progressing, fmt.Sprintf("Deployment %q is not yet scheduled on any nodes", depName.String()))
 			depProgressing = true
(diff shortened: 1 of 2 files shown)


pkg/controller/statusmanager/pod_status.go
```diff
@@ -91,6 +92,21 @@ func (status *StatusManager) SetFromPods() {
 		} else if ds.Status.NumberUnavailable > 0 {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
 			dsProgressing = true
+			// Check for any pods in CrashLoopBackOff state and mark the operator as degrade if so.
+			pods := &v1.PodList{}
+			err := status.client.List(context.TODO(), pods, client.InNamespace(dsName.Namespace), client.MatchingLabels(ds.Spec.Selector.MatchLabels))
```
squeed, line 97: suggestion: refactor this in to a function that takes namespace and label selector.

tssurya: yea, good point, was thinking the same to avoid code duplication, but just wanted a initial code screen first to see how the CI behaves.

tssurya: I also have to add tests

pecameron:
@tssurya This looks like network pods only? Also, I have encountered some non-network pods in CrashLoopBackoff during bringup. They are fine when cluster is fully up. Not sure why we try to start pods before cluster is ready.

tssurya:
> @tssurya This looks like network pods only? 

This patch actually includes all the pods in the cluster - as in any pod in any namespace which is stuck in the crash looping phase will be taken into consideration. Also this is a really simple patch to update the status of the operator accordingly.

> Also, I have encountered some non-network pods in CrashLoopBackoff during bringup. They are fine when cluster is fully up. Not sure why we try to start pods before cluster is ready.

 I am not sure about this, haven't seen this before (but I'm also relatively new to CNO and operators in general)

juanluisvaladas:
If pods are in crashloop backoff the daemonset should report as having less ready replicas than desired (if it's crashloopbackoff it's not ready). It looks me like duplicating the effort of the kube controller manager, is this resolving an observed problem or an improvement?

Also, some components aren't critical and therefore shouldn't be enough to mark the CO as degraded, those daemonsets created with the annotation `networkoperator.openshift.io/non-critical: "" ` should be ignored here.

squeed:
> If pods are in crashloop backoff the daemonset should report as having less ready replicas than desired (if it's crashloopbackoff it's not ready). It looks me like duplicating the effort of the kube controller manager, is this resolving an observed problem or an improvement?

@juanluisvaladas yeah, this is somewhat of a cosmetic improvement: if a pod is in CrashLoopBackoff, we will eventually report degraded because  `NumberUnavailable > 0`. However, we can do better: if we know that a pod is CrashLoopBackoff, we should go Degraded immediately.

tssurya:
> If pods are in crashloop backoff the daemonset should report as having less ready replicas than desired (if it's crashloopbackoff it's not ready). 

yea, this is true, the daemonset reports having less ready replicas :https://github.com/openshift/cluster-network-operator/blob/a10490e62709e9f96971db6adbc4ca6de735afa2/pkg/controller/statusmanager/pod_status.go#L92 and that's when I go in to check specifically for the crashloopbackoff case.

> It looks me like duplicating the effort of the kube controller manager, is this resolving an observed problem or an improvement?

It's trying to do both I guess, firstly fix an observed problem - say we have a pod which is alternating between crashloopbackoff and running or completed states, the existing logic which checks for the daemonset's status won't catch this among the hung ones and secondly its also an improvement where we are saying its better to let the user/op know that the network is degraded because something is crashing.

> 
> Also, some components aren't critical and therefore shouldn't be enough to mark the CO as degraded, those daemonsets created with the annotation `networkoperator.openshift.io/non-critical: "" ` should be ignored here.

I didn't know about this, I'll include the !isNonCritical check before doing this. Thanks for pointing this out.

juanluisvaladas:
> we are saying its better to let the user/op know that the network is degraded because something is crashing.

That's a pretty solid reason. Sounds like a good change.

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -1188,3 +1208,171 @@ func setLastPodState(t *testing.T, client client.Client, name string, ps podStat
 		t.Fatal(err)
 	}
 }
+
+func TestStatusManagerCheckCrashLoopBackOffPods(t *testing.T) {
```
juanluisvaladas, line 1212: Not really critical but having a test with the NonCriticalAnnotation would be a nice to have.

tssurya: yea, I'll add one with that, more coverage is always good.

pkg/controller/statusmanager/pod_status.go
```diff
@@ -138,6 +143,10 @@ func (status *StatusManager) SetFromPods() {
 		if dep.Status.UnavailableReplicas > 0 {
 			progressing = append(progressing, fmt.Sprintf("Deployment %q is not available (awaiting %d nodes)", depName.String(), dep.Status.UnavailableReplicas))
 			depProgressing = true
+			// Check for any pods in CrashLoopBackOff state and mark the operator as degrade if so.
+			if !isNonCritical(dep) {
+				hung = status.CheckCrashLoopBackOffPods(depName, dep.Spec.Selector.MatchLabels, hung, "Deployment")
```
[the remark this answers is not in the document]
tssurya: I'll tidy this up :)

pkg/controller/statusmanager/pod_status.go
```diff
@@ -290,6 +299,27 @@ func (status *StatusManager) setLastPodState(
 	})
 }
 
+func (status *StatusManager) CheckCrashLoopBackOffPods(dName types.NamespacedName, lb map[string]string, name string) []string {
```
squeed, line 302: A few minor cleanups:
* rename `lb` to `selector`, `name` to `kind.
* Write a quick docblock, something like

```
CheckCrashLoopBackOffPods checks for pods in the label selector with with 
any containers in state CrashLoopBackoff. It returns a human-readable string 
for any pod in such a state.

dName should be the name of a DaemonSet or Deployment.
```

squeed:
One minor cleanup, then this looks good.
/approve

I'll lgtm when it's good to go. We might have to skip some tests.

pkg/controller/statusmanager/pod_status.go
tssurya: ack, I'll fix these up

PR #671, opened 2020-06-16
Bug 1847313: Split etcd sg rule ports range into different sg rules

MaysaMacedo:
Depending on the workload running on the cluster,
the etcd leader elections can happen more frequently due to
timing out to connect to etcd port 2380. This commit can
workaround the issue by splitting the etcd port ranges into
different security group rules.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -55,7 +55,8 @@ const (
 	DNSServiceName                         = "dns-default"
 	// NOTE(ltomasbo): Only OVN octavia driver supported on kuryr
 	OVNProvider              = "ovn"
-	etcdPort                 = 2379
+	etcdClientPort           = 2379
+	etcdServerToServerPort   = 2380
 	dnsPort                  = 53
 	apiPort                  = 6443
 	routerMetricsPort        = 1936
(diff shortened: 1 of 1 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -380,18 +381,23 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+		// rules to avoid the issue of constant leader election changes
+		{masterSgId, workerSubnet.CIDR, etcdClientPort, etcdClientPort, rules.ProtocolTCP},
+		{masterSgId, workerSubnet.CIDR, etcdServerToServerPort, etcdServerToServerPort, rules.ProtocolTCP},
 	}
 
 	var decommissionedRules = []sgRule{
 		{podSgId, workerSubnet.CIDR, 0, 0, ""},
-		{masterSgId, openStackSvcCIDR, 2379, 2380, rules.ProtocolTCP},
+		{masterSgId, openStackSvcCIDR, etcdClientPort, etcdServerToServerPort, rules.ProtocolTCP},
+		{masterSgId, workerSubnet.CIDR, etcdClientPort, etcdServerToServerPort, rules.ProtocolTCP},
```
dulek, line 395: Uhm, had this rule ever existed?

MaysaMacedo: The installer creates it. This is the one that needs to be deleted and splitted.

dulek: Oh crap, so this is certainly a bigger problem. First of all it'll affect setups without Kuryr too, right? If so it probably should be changed in installer too.

Second issue is the fact that IIRC we're unable to modify those rules on upgrade, so older releases are stuck with old groups.

MaysaMacedo:
/hold
Till the CI issue is verified.

PR #672, opened 2020-06-16
openshift-sdn, ovn-kubernetes: adopt systemd-managed openvswitch if present

squeed:
Needed to remove some of the readiness probes because they don't work, but this doesn't actually lose much coverage.

And liveness probes are not useful in this context.

This contains a hack to pass a strange case that an artifact of CI upgrade tests. It won't be in the wild, and could be removed.

Note to reviewers: hide whitespace changes :-)

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -34,131 +34,128 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
-          chown -R openvswitch:openvswitch /var/run/openvswitch
-          chown -R openvswitch:openvswitch /etc/openvswitch
-
-          # if another process is listening on the cni-server socket, wait until it exits
-          retries=0
-          while true; do
-            if /usr/share/openvswitch/scripts/ovs-ctl status &>/dev/null; then
-              echo "warning: Another process is currently managing OVS, waiting 15s ..." 2>&1
-              sleep 15 & wait
-              (( retries += 1 ))
-            else
-              break
+          export SYSTEMD_IGNORE_CHROOT=yes
+          # Check to see if ovs is provided by the node:
+          if systemctl status ovsdb-server; then
+            echo "openvswitch is running in systemd"
+            # In some very strange corner cases, the owner for /run/openvswitch
+            # can be wrong, so we need to clean up and restart.
+            ovs_uid=$(chroot /host id -u openvswitch)
+            ovs_gid=$(chroot /host id -g openvswitch)
+            chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
+            if [[ ! -S /run/openvswitch/db.sock ]]; then
+              systemctl restart ovsdb-server
             fi
-            if [[ "${retries}" -gt 40 ]]; then
-              echo "error: Another process is currently managing OVS, exiting" 2>&1
-              exit 1
+
+            # Don't need to worry about restoring flows; this can only change if we've rebooted
+            exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
+          else
+
+            echo "openvswitch is running in container"
+            # if another process is listening on the cni-server socket, wait until it exits
+            retries=0
+            while true; do
+              if /usr/share/openvswitch/scripts/ovs-ctl status &>/dev/null; then
+                echo "warning: Another process is currently managing OVS, waiting 15s ..." 2>&1
+                sleep 15 & wait
+                (( retries += 1 ))
+              else
+                break
+              fi
+              if [[ "${retries}" -gt 40 ]]; then
+                echo "error: Another process is currently managing OVS, exiting" 2>&1
+                exit 1
+              fi
+            done
+
+            function quit {
+                # Save the flows
+                echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saving flows ..." 2>&1
+                bridges=$(ovs-vsctl -- --real list-br)
+                TMPDIR=/var/run/openvswitch /usr/share/openvswitch/scripts/ovs-save save-flows $bridges > /var/run/openvswitch/flows.sh
+                echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saved flows" 2>&1
+
+                # Don't allow ovs-vswitchd to clear datapath flows on exit
+                kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true
+                kill $(cat /var/run/openvswitch/ovsdb-server.pid 2>/dev/null) 2>/dev/null || true
+                exit 0
+            }
+            trap quit SIGTERM
+
+            function deletePid {
+                rm /var/run/openvswitch/ovs-vswitchd.pid
+                rm /var/run/openvswitch/ovsdb-server.pid
+            }
+            trap deletePid EXIT
+
+            chown -R openvswitch:openvswitch /var/run/openvswitch
+            chown -R openvswitch:openvswitch /etc/openvswitch
+
+            # launch OVS
+            # Start the ovsdb so that we can prep it before we start the ovs-vswitchd
+            /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random --no-monitor
+
+            # Set the flow-restore-wait to true so ovs-vswitchd will wait till flows are restored
+            ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-restore-wait=true
+
+            # Restrict the number of pthreads ovs-vswitchd creates to reduce the
+            # amount of RSS it uses on hosts with many cores
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1571379
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1572797
+            if [[ `nproc` -gt 12 ]]; then
+                ovs-vsctl --no-wait set Open_vSwitch . other_config:n-revalidator-threads=4
+                ovs-vsctl --no-wait set Open_vSwitch . other_config:n-handler-threads=10
             fi
-          done
 
-          function quit {
-              # Save the flows
-              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saving flows ..." 2>&1
-              bridges=$(ovs-vsctl -- --real list-br)
-              TMPDIR=/var/run/openvswitch /usr/share/openvswitch/scripts/ovs-save save-flows $bridges > /var/run/openvswitch/flows.sh
-              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Saved flows" 2>&1
+            # And finally start the ovs-vswitchd now the DB is prepped
+            /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovsdb-server --system-id=random --no-monitor
 
-              # Don't allow ovs-vswitchd to clear datapath flows on exit
-              kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true
-              kill $(cat /var/run/openvswitch/ovsdb-server.pid 2>/dev/null) 2>/dev/null || true
-              exit 0
-          }
-          trap quit SIGTERM
+            # Load any flows that we saved
+            echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Loading previous flows ..." 2>&1
+            if [[ -f /var/run/openvswitch/flows.sh ]]; then
+              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Adding br0 if it doesn't exist ..." 2>&1
+              /usr/bin/ovs-vsctl --may-exist add-br br0 -- set Bridge br0 fail_mode=secure protocols=OpenFlow13
+              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Created br0, now adding flows ..." 2>&1
+              sh -x /var/run/openvswitch/flows.sh
+              echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Done restoring the existing flows ..." 2>&1
+            fi
 
-          function deletePid {
-              rm /var/run/openvswitch/ovs-vswitchd.pid
-              rm /var/run/openvswitch/ovsdb-server.pid
-          }
-          trap deletePid EXIT
+            echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Remove other config ..." 2>&1
+            ovs-vsctl --no-wait --if-exists remove Open_vSwitch . other_config flow-restore-wait=true
+            echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Removed other config ..." 2>&1
 
-          # launch OVS
-          # Start the ovsdb so that we can prep it before we start the ovs-vswitchd
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --no-ovs-vswitchd --system-id=random --no-monitor
-
-          # Set the flow-restore-wait to true so ovs-vswitchd will wait till flows are restored
-          ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-restore-wait=true
-          
-          # Restrict the number of pthreads ovs-vswitchd creates to reduce the
-          # amount of RSS it uses on hosts with many cores
-          # https://bugzilla.redhat.com/show_bug.cgi?id=1571379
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -47,6 +47,14 @@ spec:
             source "/env/${K8S_NODE}"
             set +o allexport
           fi
+          export SYSTEMD_IGNORE_CHROOT=yes
+          # Check to see if ovs is provided by the node:
+          if systemctl status ovsdb-server; then
+            echo "openvswitch is running in systemd"
+            # Don't need to worry about restoring flows; this can only change if we've rebooted
+            exec tail -F /var/log/openvswitch-host/ovs-vswitchd.log /var/log/openvswitch-host/ovsdb-server.log
+          fi
```
mccv1r0, line 56: Do you mean /var/log/openvswitch  instead of openvswitch-host? 

Is the goal to block here, and never return?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -136,6 +144,14 @@ spec:
...
         hostPath:
           path: /sys
+      - name: host-run-systemd
+        hostPath:
+          path: /run/systemd
+          type: Directory
+      - name: host-var-log-openvswitch
+        hostPath:
+          path: /var/log/openvswitch
+          type: Directory
```
mccv1r0, line 154: When OVS is *not* started on the host, what creates this Directory?   The container will fail mount constraints until the host creates ```/var/log/openvwitch```

bindata/network/ovn-kubernetes/006-ovs-node.yaml
squeed: No, this is correct. /var/log/openvswitch is not a host directory.
The goal is to block here. We need to keep a process running so kubelet doesn't think the pod has died.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
squeed: Good point, though I don't think it's possible to get in that state. I'll update it.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
mccv1r0: > No, this is correct. /var/log/openvswitch is not a host directory.

Then shouldn't e.g. 

```
tail -F --pid=$(cat /var/run/openvswitch/ovs-vswitchd.pid) /var/log/openvswitch/ovs-vswitchd.log &
```

below also use ```/var/log/openvswitch-host``` ?

squeed: no... because that's for the containerized case.

squeed:
e2e-gcp has a single freaking test failure. the network came up just fine.

vishnoianil:
@squeed this is the only PR where gcp-ovn test passed, rest all PR are failing this test because CNO is degrading becayse br-int is not created on the nodes.

abhat:
I think we should revert: https://github.com/openshift/machine-config-operator/pull/1830 first.

squeed:
test failures are all flakes or infra issues
/retest

squeed:
Upgrades are failing, nuts:

```
● ovsdb-server.service - Open vSwitch Database Unit
   Loaded: loaded (/usr/lib/systemd/system/ovsdb-server.service; static; vendor preset: enabled)
   Active: active (running) since Wed 2020-06-17 08:33:56 UTC; 25min ago
  Process: 1177 ExecStart=/usr/share/openvswitch/scripts/ovs-ctl --no-ovs-vswitchd --no-monitor --system-id=random ${OVS_USER_OPT} start $OPTIONS (code=exited, status=0/SUCCESS)
  Process: 1170 ExecStartPre=/bin/sh -c if [ "$${OVS_USER_ID/:*/}" != "root" ]; then /usr/bin/echo "OVS_USER_OPT=--ovs-user=${OVS_USER_ID}" >> /run/openvswitch.useropts; fi (code=exited, status=0/SUCCESS)
  Process: 1146 ExecStartPre=/bin/sh -c rm -f /run/openvswitch.useropts; /usr/bin/echo "OVS_USER_ID=${OVS_USER_ID}" > /run/openvswitch.useropts (code=exited, status=0/SUCCESS)
  Process: 1135 ExecStartPre=/usr/bin/chown ${OVS_USER_ID} /var/run/openvswitch /var/log/openvswitch (code=exited, status=1/FAILURE)
 Main PID: 1234 (ovsdb-server)
    Tasks: 1 (limit: 95371)
   Memory: 20.0M
   CGroup: /system.slice/ovsdb-server.service
           └─1234 ovsdb-server /etc/openvswitch/conf.db -vconsole:emer -vsyslog:err -vfile:info --remote=punix:/var/run/openvswitch/db.sock --private-key=db:Open_vSwitch,SSL,private_key --certificate=db:Open_vSwitch,SSL,certificate --bootstrap-ca-cert=db:Open_vSwitch,SSL,ca_cert --user openvswitch:hugetlbfs --no-chdir --log-file=/var/log/openvswitch/ovsdb-server.log --pidfile=/var/run/openvswitch/ovsdb-server.pid --detach
openvswitch is running in systemd
==> /var/log/openvswitch-host/ovs-vswitchd.log <==
2020-06-17T08:59:30.577Z|09841|fatal_signal|WARN|could not unlink "/var/run/openvswitch/br0.snoop" (Permission denied)
2020-06-17T08:59:30.577Z|09842|stream_unix|ERR|/var/run/openvswitch/br0.snoop: binding failed: Permission denied
2020-06-17T08:59:30.577Z|09843|connmgr|ERR|failed to listen on punix:/var/run/openvswitch/br0.snoop: Permission denied
2020-06-17T08:59:30.583Z|09844|socket_util_unix|WARN|unlinking "/var/run/openvswitch/br0.mgmt": Permission denied
2020-06-17T08:59:30.583Z|09845|fatal_signal|WARN|could not unlink "/var/run/openvswitch/br0.mgmt" (Permission denied)
2020-06-17T08:59:30.583Z|09846|stream_unix|ERR|/var/run/openvswitch/br0.mgmt: binding failed: Permission denied
2020-06-17T08:59:30.583Z|09847|socket_util_unix|WARN|unlinking "/var/run/openvswitch/br0.snoop": Permission denied
2020-06-17T08:59:30.583Z|09848|fatal_signal|WARN|could not unlink "/var/run/openvswitch/br0.snoop" (Permission denied)
2020-06-17T08:59:30.583Z|09849|stream_unix|ERR|/var/run/openvswitch/br0.snoop: binding failed: Permission denied
2020-06-17T08:59:30.583Z|09850|connmgr|ERR|failed to listen on punix:/var/run/openvswitch/br0.snoop: Permission denied
```

squeed:
Huh: `chown[1135]: /usr/bin/chown: cannot access '/var/run/openvswitch': No such file or directory`

squeed:
Okay, that should fix it.
@mccv1r0 want to take a quick look? This isn't perfect, but it should get us moving forward.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -34,9 +35,24 @@ spec:
         - |
           #!/bin/bash
           set -euo pipefail
-          chown -R openvswitch:openvswitch /var/run/openvswitch
-          chown -R openvswitch:openvswitch /etc/openvswitch
+          
+          export SYSTEMD_IGNORE_CHROOT=yes
+          # Check to see if ovs is provided by the node:
+          if systemctl status ovsdb-server; then
```
juanluisvaladas, line 41: systemctl status won't return 0 if the service is stopped.
Shouldn't we do something that returns 1 if the unit is available to be executed even if it's stopped? For instance systemctl cat ovsdb-server

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -73,6 +89,9 @@ spec:
           }
           trap deletePid EXIT
 
+          chown -R openvswitch:openvswitch /var/run/openvswitch
```
juanluisvaladas, line 92: About the line:
              kill -9 $(cat /var/run/openvswitch/ovs-vswitchd.pid 2>/dev/null) 2>/dev/null || true

ovsdb-server has by default defined on the systemd unit Restart=on-failure, a kill -9 would restart it. I haven't actually tried it but I think systemctl kill would get the job done.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -73,6 +89,9 @@ spec:
...
+          chown -R openvswitch:openvswitch /etc/openvswitch
```
juanluisvaladas, line 93: If we are using systemd's ovsdb server shouldn't we use systemd's vswitchd as well?

squeed: we do.

bindata/network/openshift-sdn/sdn-ovs.yaml
squeed: this is for containerized mode only - and is unchanged by this pr

bindata/network/openshift-sdn/sdn-ovs.yaml
squeed: We should not be starting openvswitch. We should just handle the case where it is started.

squeed:
So, this PR was failing because system openvswitch was added first, and we needed to handle an impossible upgrade case.

Now that the MCO pr has been reverted, we can get this in, **then** enable system openvswitch.

squeed:
@juanluisvaladas all valid criticisms... but I didn't add that :-) Those are unchanged, left over from the containerized case.

squeed:
Tested this by manually running `systemd enable openvswitch && reboot` on a test cluster, worked perfectly.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -47,24 +47,40 @@ spec:
...
-              exit 0
-          }
-          trap quit SIGTERM
-          /usr/share/openvswitch/scripts/ovs-ctl start --ovs-user=openvswitch:openvswitch --system-id=random
-          ovs-appctl vlog/set "file:${OVS_LOG_LEVEL}"
-          /usr/share/openvswitch/scripts/ovs-ctl --protocol=udp --dport=6081 enable-protocol
-          echo "$(date -Iseconds) - ovs-daemons running"
+          export SYSTEMD_IGNORE_CHROOT=yes
+          # Check to see if ovs is provided by the node:
+          if systemctl status ovsdb-server; then
```
trozet, line 52: make this check if ovs is enabled in systemd, not check if it is actually up

trozet: add another check if it is not up in this if block, then we need to exit until MCO/systemd restarts OVS

squeed: done

squeed:
@dcbw @mccv1r0 please review; this should go in ASAP so we can flip the switch in MCO.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -34,131 +34,128 @@ spec:
...
-              exit 1
+          export SYSTEMD_IGNORE_CHROOT=yes
+          # Check to see if ovs is provided by the node:
+          if systemctl is-enabled ovsdb-server; then
+            echo "openvswitch is running in systemd"
+            # In some very strange corner cases, the owner for /run/openvswitch
+            # can be wrong, so we need to clean up and restart.
+            ovs_uid=$(chroot /host id -u openvswitch)
+            ovs_gid=$(chroot /host id -g openvswitch)
+            chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
```
mccv1r0, line 45: Doesn't this just move the problem?

When MCO and CNO have different uid/gid for the openvswitch user/group the CNO got permission denied for things created by MCO.  The result was OVN didn't come up.  Doing `chown -R` to set the uid/gid to what MCO uses doesn't change the fact that the CNO container is using a different uid/gid for the openvswitch user/group.  When a  CNO container tries to access these files using a different uid/gid are you sure that CNO will not hit the same permission denied problem(s)?

squeed: Everything except openvswitch runs as root. So, yes, we should be fine for permissions. We're also not writing anything to this directory from a container anymore.

squeed:
@trozet turns out `systemctl is-enabled` returns something incorrect on RHCOS, we need to use `systemctl status`.

trozet:
> @trozet turns out `systemctl is-enabled` returns something incorrect on RHCOS, we need to use `systemctl status`.

Then since you mount /host you can check for the openvswitch unit file existing or you need to check the return code from systemctl status. On my setup 3 indicates the service exists but is not up, while 4 means service does not exist.

bindata/network/openshift-sdn/sdn-ovs.yaml
trozet: my understanding here is that we are not trying to set the permissions for this container, we are trying to fix the permissions on the host for the ovs directory because it may have a conflict with kubelet. We had similar problems in openstack between OVS and nova. I guess you could argue that this really isn't this container's job and the problem should be fixed by MCO?

On the other hand while this container will not issue any ovs commands, the ovn-k8s container will. Does the uid/gid matter in that case to just call the binaries?

squeed:
> Then since you mount /host you can check for the openvswitch unit file existing or you need to check the return code from systemctl status. On my setup 3 indicates the service exists but is not up, while 4 means service does not exist.

The unit file already exists, it's just not enabled by default. So we can't use whether or not the service exists as a predicate.

We could use `is-enabled` as a predicate, but that's not reliable for a few reasons - first of all, and this is the case on RHCOS, unit without ah `[Install]` section are defined as enabled (even though they're not). Second of all, we might not enable the unit directly in the future; we could make it a dependency of a meta-service. That would break this predicate again.

I don't want to get in the business of checking for systemd configuration files on disk. They can live in a lot of places, so that's just setting ourselves up for trouble.

In any case, the chances of `openvswitch.service` being stopped on boot, then later enabled, are pretty slim. If it crashes, systemd will restart it. I think this is a reasonable test.

squeed:
All required CI runs are green. Failing jobs are oddities - their network came up.

bindata/network/openshift-sdn/sdn-ovs.yaml
squeed: I should have been more specific. If openvswitch is a systemd service, no containerized processes will write to `/run/openvswitch`. They'll only read.

trozet:
> > Then since you mount /host you can check for the openvswitch unit file existing or you need to check the return code from systemctl status. On my setup 3 indicates the service exists but is not up, while 4 means service does not exist.
> 
> The unit file already exists, it's just not enabled by default. So we can't use whether or not the service exists as a predicate.
> 
> We could use `is-enabled` as a predicate, but that's not reliable for a few reasons - first of all, and this is the case on RHCOS, units without an `[Install]` section are defined as enabled (even though they're not). Second of all, we might not enable the unit directly in the future; we could make it a dependency of a meta-service. That would break this predicate again.
> 
> I don't want to get in the business of checking for systemd configuration files on disk. They can live in a lot of places, so that's just setting ourselves up for trouble.
> 
> In any case, the chances of `openvswitch.service` being stopped on boot, then later enabled, are pretty slim. If it crashes, systemd will restart it. I think this is a reasonable test.

I see your point now. I did't realize OVS RPM was in 4.5 already, so checking unit file existence doesn't really help us. @dcbw fyi

bindata/network/openshift-sdn/sdn-ovs.yaml
mccv1r0: There were issues with br-int.mgmt being created but didn't have permission.  Is this OVS or OVN?

squeed: Where did you see these issues? In CI? Or in your PR?

mccv1r0:
I fed this and the MCO PR to cluster-bot and things seemed to behave.  The logging of e.g. ```echo "openvswitch is running in systemd"``` seemed to get lost.  Is this going to be an issue?

squeed:
> I fed this and the MCO PR to cluster-bot and things seemed to behave. The logging of e.g. `echo "openvswitch is running in systemd"` seemed to get lost. Is this going to be an issue?

Huh. Well, did the PR do the right thing? Do you have a link to the CI run?

mccv1r0:
> > I fed this and the MCO PR to cluster-bot and things seemed to behave. The logging of e.g. `echo "openvswitch is running in systemd"` seemed to get lost. Is this going to be an issue?
> 
> Huh. Well, did the PR do the right thing? Do you have a link to the CI run?

"things seemed to behave"  The cluster did come up.  I didn't get a link just kubeconfig so I can run ```oc logs```

squeed:
@mccv1r0 if this looks good to you, can you /lgtm?

PR #674, opened 2020-06-16
ovn-kubernetes: allow dual-stack configs

danwinship:
CNO is enforcing that ovn-kubernetes only accepts a single service CIDR, which is no longer true


rcarrillocruz:
Change looks good, but where is 'service-cidrs' consumed?

rcarrillocruz:
nm, config.go from ovn-k

PR #682, opened 2020-06-26
Bug 1851518: Expose raft (nb-db/sb-db) election-timer and ovn-controller inactivity probe.

vishnoianil:
These timers are currently set to fixed value based on my current observation
from scale tests. We might have to increase these values in future based on
the scale we will be supporting for upcoming release. Currently election-timer
values are limited by the raft jsonrpc inactivity-probe time of 5 seconds as well.
To further increase the election-timer value, we need to disable jsonrpc
inactivity-probe.

Signed-off-by: Anil Vishnoi avishnoi@redhat.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,5 +106,7 @@ Session.vim
 !.vscode/extensions.json
 .history
 
+### GoLand files ###
+.idea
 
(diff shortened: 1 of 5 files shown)


vishnoianil:
@dcbw PTAL, thanks

PR #688, opened 2020-07-01
Bug 1837953: Check if the db file has been correctly initialized with raft.

fedepaol:
If this does not happen, the master pod might fail with `ovsdb-server: ovsdb error: server does not belong to cluster`
and never recover. We make sure the current serverid is in the file by checking its presence in the prev_servers set.

As discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1837953, this is a workaround to recover from inconsistent local db state that may cause the master pod to CrashLoopBackOff.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -115,6 +115,25 @@ spec:
 
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
 
+          check_db_health() {
+            db="${1}"
+            if [[ ! -f "${db}" ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking ${db} health"
+            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            if [[ -z ${match} ]]; then
+                echo "Current server_id $serverid not found in ${db}, cleaning up"
+                rm -- "${db}"
+            else
+                  echo "${db} is healthy"
+            fi
+          }
+
+          check_db_health "/etc/ovn/ovnnb_db.db"
+
           MASTER_IP="{{.OVN_MASTER_IP}}"
           if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
             exec /usr/share/ovn/scripts/ovn-ctl \
@@ -220,6 +239,25 @@ spec:
           fi
 
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
+
+          check_db_health() {
+            db="${1}"
+            if [[ ! -f "${db}" ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking ${db} health"
+            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            if [[ -z ${match} ]]; then
+                echo "Current server_id $serverid not found in ${db}, cleaning up"
+                rm -- "${db}"
+            else
+                  echo "${db} is healthy"
+            fi
+          }
+
+          check_db_health "/etc/ovn/ovnsb_db.db"
 
           MASTER_IP="{{.OVN_MASTER_IP}}"
           if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
...
+            if [[ ! -f ${db} ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking $db health"
+            serverid=$(ovsdb-tool show-log ${db} | grep "server_id" | sed 's/server_id: \(.*\)/\1/')
+            match=$(ovsdb-tool show-log ${db} | grep "prev_servers:.*${serverid}(" || true)
+            if [[ -z ${match} ]]; then
+                echo "Current server_id $serverid not found in $db, cleaning up"
+                rm ${db}
```
rbbratta, line 129: could we add quoting and guard flag?
`rm -- "${db}"`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
...
 
+          check_db_health() {
+            db=$1
+            if [[ ! -f ${db} ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking $db health"
+            serverid=$(ovsdb-tool show-log ${db} | grep "server_id" | sed 's/server_id: \(.*\)/\1/')
+            match=$(ovsdb-tool show-log ${db} | grep "prev_servers:.*${serverid}(" || true)
```
rbbratta, line 126: please quote `${db}`
can we avoid `.*`  and greediness backtracking and be more precise.  It should be just spaces.

`grep "prev_servers: *${serverid}(`
or
`grep "prev_servers:[[:blank:]]*${serverid}(`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
...
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
 
+          check_db_health() {
+            db=$1
+            if [[ ! -f ${db} ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking $db health"
+            serverid=$(ovsdb-tool show-log ${db} | grep "server_id" | sed 's/server_id: \(.*\)/\1/')
```
rbbratta, line 125: can we quote `${db}` everywhere?

also might as well let sed do both the matching and replacing, and maybe we can use `[[:xdigit:]]\+` to match hex.

`ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p'`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
fedepaol: Done

bindata/network/ovn-kubernetes/ovnkube-master.yaml
fedepaol: Done, thanks for helping in making this better!

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
 
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
 
+          check_db_health() {
+            db=$1
```
rbbratta, line 119: ah, well technically we should quote here as well

`db="${1}"`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -221,6 +240,25 @@ spec:
...
```
rbbratta, line 244: and quote here

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: we should probably run these snippets through shellcheck regularly.

https://www.shellcheck.net/

bindata/network/ovn-kubernetes/ovnkube-master.yaml
fedepaol: Done

bindata/network/ovn-kubernetes/ovnkube-master.yaml
fedepaol: Done

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
...
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
 
+          check_db_health() {
+            db="${1}"
+            if [[ ! -f "${db}" ]]; then
+                  echo "${db} does not exist, skipping health check"
+                  return
+            fi
+            echo "Checking ${db} health"
+            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
```
rbbratta, line 125: right, a little to clever, forgot the leading and trailing whitespace in the sed.  Usually the form is `s/^.*\(foo\).*$/\1/`


`$(ovsdb-tool show-log /etc/ovn/ovnnb_db.db | sed -ne '/server_id:/s/ *server_id: *\([[:xdigit:]]\+\).*/\1/p')`



 
bad: extra leading space
```
sh-4.2# serverid=$(ovsdb-tool show-log /etc/ovn/ovnnb_db.db | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/a\1/p')
sh-4.2# echo "[${serverid}]"
[ aac7d]
```

better
````
sh-4.2# serverid=$(ovsdb-tool show-log /etc/ovn/ovnnb_db.db | sed -ne '/server_id:/s/ *server_id: *\([[:xdigit:]]\+\).*/\1/p')
sh-4.2# echo "[${serverid}]"
[ac7d]
```

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -115,6 +115,25 @@ spec:
 
           bracketify() { case "$1" in *:*) echo "[$1]" ;; *) echo "$1" ;; esac }
 
+          check_db_health() {
+            db="${1}"
```
rbbratta, line 119: this is okay, but technically I was wrong.  

It turns out, in bash, variable assignment is the one case that doesn't need quoting, because the shell only allows a single word after the `=`

http://rg1-teaching.mpi-inf.mpg.de/unixffb-ss98/quoting-guide.html#para:sh-assignments

But always quoting seems safest rule to memorize.  It's not easy being pedantic. :laughing:

Sorry for the thrash on all this.

fedepaol: Np, there's definetely room for improvement in my bash skills and I am happy to learn. Moreover, I can't say no to Cobra Commander :P

PR #690, opened 2020-07-02
Bump OVN Bootstrap timeout

alexanderConstantinescu:
Today we wait for 1 minute for all nodes to be up before booting OVN. With "OVS on the system" there is additional configuration done by NetworkManager / MCO networking startup scripts and the likes, which might require longer timeouts than what we currently support. So, let's bump the time.

Reconciliation happens every 5 min, so I am setting the timeout to 280 seconds to not block reconciliation during bring up. 

/assign @trozet

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -232,7 +232,7 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 
 	controlPlaneReplicaCount, _ := strconv.Atoi(rcD.ControlPlane.Replicas)
 
-	err := wait.PollImmediate(2*time.Second, 60*time.Second, func() (bool, error) {
+	err := wait.PollImmediate(5*time.Second, 280*time.Second, func() (bool, error) {
 		matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
 		if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
 			return false, err


trozet:
@abhat @dbw @danwinship can one of you please approve?

pkg/network/ovn_kubernetes.go
```diff
@@ -232,7 +232,7 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 
 	controlPlaneReplicaCount, _ := strconv.Atoi(rcD.ControlPlane.Replicas)
 
-	err := wait.PollImmediate(2*time.Second, 60*time.Second, func() (bool, error) {
+	err := wait.PollImmediate(5*time.Second, 280*time.Second, func() (bool, error) {
```
tssurya, line 235: out of curiosity,:
1) any particular reason why we also bumped the polling interval to 5secs? - just to reflect the relativity wrt to the timeout interval?
2) do we know approximately how long it takes with the new model?

alexanderConstantinescu: 1. Yes
2. No, neither I nor @trozet performed a real study on the time it normally takes, and what we should set these values to....but then again, that was never the case the first time we implemented this mechanism either. With OVS running on the node and attached to the primary NIC it has however been seen that nodes might boot a bit slower due to DHCP issues: https://github.com/openshift/machine-config-operator/pull/1860#issuecomment-653354832.

tssurya: understood, thanks for clarifying Alex.

PR #692, opened 2020-07-06
Bug 1853712: Fix kuryr templates format for ifs

luis5tb:


danielmellado:
overall LGTM, pretty interesting that the change in the library wasn't backwards compatible. Let's see CI

dulek:
It's better to use `getOr` instead of `default`, turns our `default` will treat `false` as no value, effectively only putting `true` there. Here's a diff of how it might look like (as it requires modification to `getOr` definition:

```
diff --git a/bindata/network/kuryr/005-daemon.yaml b/bindata/network/kuryr/005-daemon.yaml
index ac3403fe..f0cf9d9b 100644
--- a/bindata/network/kuryr/005-daemon.yaml
+++ b/bindata/network/kuryr/005-daemon.yaml
@@ -59,7 +59,7 @@ spec:
         - name: netns
           mountPath: /run/netns
           mountPropagation: HostToContainer
-{{ if (default true .DaemonEnableProbes) eq "true" }}
+{{ if eq (getOr . "DaemonEnableProbes" true) true }}
         readinessProbe:
           failureThreshold: 10
           httpGet:
diff --git a/bindata/network/kuryr/006-controller.yaml b/bindata/network/kuryr/006-controller.yaml
index 8d7dae38..66494476 100644
--- a/bindata/network/kuryr/006-controller.yaml
+++ b/bindata/network/kuryr/006-controller.yaml
@@ -30,7 +30,7 @@ spec:
       containers:
       - image: {{ .ControllerImage }}
         name: controller
-{{ if (default true .ControllerEnableProbes) eq "true" }}
+{{ if eq (getOr . "ControllerEnableProbes" true) true }}
         readinessProbe:
           failureThreshold: 10
           httpGet:
diff --git a/pkg/render/funcs.go b/pkg/render/funcs.go
index 17d61433..54d07c12 100644
--- a/pkg/render/funcs.go
+++ b/pkg/render/funcs.go
@@ -5,7 +5,7 @@ package render
 // getOr returns the value of m[key] if it exists, fallback otherwise.
 // As a special case, it also returns fallback if the value of m[key] is
 // the empty string
-func getOr(m map[string]interface{}, key, fallback string) interface{} {
+func getOr(m map[string]interface{}, key string, fallback interface{}) interface{} {
     val, ok := m[key]
     if !ok {
         return fallback

```

dulek:
Ah, I cannot approve this one.

PR #694, opened 2020-07-07
Bug 1837953: fixup server_id regex

rbbratta:
We need to search the whole line for the server_id.

Current regex is too aggressive.

The original example indicates that the prev_server line is a list of servers

```bash
$ ovsdb-tool show-log /etc/ovn/ovnnb_db.db | grep "server_id\|prev_servers"
 server_id: d5db
  prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643"
```
  The above DB is inconsistent because d5db is not part of prev_servers.

A consistent DB looks like this:
```bash
$ ovsdb-tool show-log /etc/ovn/ovnnb_db.db | grep "server_id\|prev_servers"
 server_id: 33b4
  prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643")
```

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -123,8 +123,10 @@ spec:
                   return
             fi
             echo "Checking ${db} health"
-            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
-            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            serverid=$(ovsdb-tool show-log "${db}" | grep -m 1 -oP '(?<=server_id: )[[:xdigit:]]+')
+            # prev_servers is a list.  We are searching for " $server_id(" in the list
+            # prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643")
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers:.* ${serverid}(" || true)
             if [[ -z ${match} ]]; then
                 echo "Current server_id $serverid not found in ${db}, cleaning up"
                 rm -- "${db}"
@@ -311,8 +313,10 @@ spec:
                   return
             fi
             echo "Checking ${db} health"
-            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
-            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            serverid=$(ovsdb-tool show-log "${db}" | grep -m 1 -oP '(?<=server_id: )[[:xdigit:]]+')
+            # prev_servers is a list.  We are searching for " $server_id(" in the list
+            # prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643")
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers:.* ${serverid}(" || true)
             if [[ -z ${match} ]]; then
                 echo "Current server_id $serverid not found in ${db}, cleaning up"
                 rm -- "${db}"


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -311,8 +311,11 @@ spec:
                   return
             fi
             echo "Checking ${db} health"
-            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
-            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            # remove leading and trailing space around server_id
+            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/ *server_id: *\([[:xdigit:]]\+\).*/\1/p')
```
danwinship, line 315: The sed can be simplified to just

        sed -ne 's/ *server_id: *\([[:xdigit:]]\+\).*/\1/p'

The `s///` rule already won't match if there's no "`server_id:`" in the string, so the extra `/server_id:/` at the front wasn't doing anything

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -311,8 +311,11 @@ spec:
...
+            # prev_servers is a list we are searching for " $server_id(" in the list
```
danwinship, line 316: Either that needs to be two sentences (period before "we") or the "the list" at the end is superfluous.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -311,8 +311,11 @@ spec:
...
+            # prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643"
```
danwinship, line 317: ok, now I'm totally just nitpicking but presumably there's a "`)`" missing at the end of that line...

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: I was originally matching the original grep plus sed formulation and the simpler regex `/server_id:/` would be faster to match that the substitution regex but it won't matter much.  

I finally check the ovsdb-tools source and there is a fixed amount of space after the `server_id: ` so we can use a positive lookbehind with `grep -o` and that should simplify things.  Also threw in a `-m 1` to exit after the first match for good measure, assuming that there should always be only one match.  The `-m 1` causes `ovsdb-tool` to complain about the broken pipe but that is okay since it's on stderr.

`grep -m 1 -oP '(?<=server_id: )[[:xdigit:]]+'`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: This was copied from the original bugzilla.  I have yet to see multiple `prev_servers` in testing, but I checked the source and there should always be a closing paren.

fedepaol:
Looks good, thanks for fixing! I think in my original (despictable) implementation I included this case, but it got eventually got lost while improving it.
I wonder if we should point the bz from this PR too. In that case, @rbbratta would you mind changing the title?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -123,8 +123,10 @@ spec:
                   return
             fi
             echo "Checking ${db} health"
-            serverid=$(ovsdb-tool show-log "${db}" | sed -ne '/server_id:/s/server_id: *\([[:xdigit:]]\+\)/\1/p')
-            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers: *${serverid}(" || true)
+            serverid=$(ovsdb-tool show-log "${db}" | grep -m 1 -oP '(?<=server_id: )[[:xdigit:]]+')
```
danwinship, line 126: :man_shrugging: I like the original version better... I feel like it's reasonable to expect future people reading this to understand `sed` syntax but not reasonable to expect them to know GNU-specific `grep` flags... Which may be a totally unreasonable expectation?

rbbratta: `grep -o` seems cleaner instead of trying to remove all the text surrounding the match.  It feels kinda sad that GNU maintainers add all these nice features (`-o` and PCRE, for example, ~20 years ago) and no one uses them.

fedepaol:
@rbbratta do you mind referring the bug in the title? 
The bug is half baked as of today and I am afraid we loose track.

dcbw:
/retest

@rbbratta yes, we should have the bug in the title as well.

rcarrillocruz:
i.e.
Bug X: <whatever description you want>

rcarrillocruz:
the description should follow the colon and a space

dcbw:
If this is a follow-on to the already fixed bug, then we need a new bug cloned from the original, target release 4.6, so that QE can verify this additional fix separate from the original one.

sdodson:
/retitle Bug 1837953: fixup server_id regex
This is a followup to fixes in https://github.com/openshift/cluster-network-operator/pull/688

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta, line 126: https://github.com/openvswitch/ovs/blob/master/ovsdb/ovsdb-tool.c#L881

```c
    printf(" server_id: "SID_FMT"\n", SID_ARGS(&h->sid));
```
https://github.com/openvswitch/ovs/blob/master/ovsdb/raft-private.h#L36
```c
#define SID_FMT "%04x"
```

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -123,8 +123,10 @@ spec:
...
+            # prev_servers is a list.  We are searching for " $server_id(" in the list
+            # prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643")
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers:.* ${serverid}(" || true)
```
rbbratta, line 129: https://github.com/openvswitch/ovs/blob/master/ovsdb/ovsdb-tool.c#L897
```c
        print_servers("prev_servers", h->snap.servers);
```

https://github.com/openvswitch/ovs/blob/master/ovsdb/ovsdb-tool.c#L828
```c
    for (size_t i = 0; i < n; i++) {
        if (i > 0) {
            printf(", ");
        }

        const struct shash_node *node = nodes[i];
        printf("%.4s(", node->name);

        const struct json *address = node->data;
        char *s = json_to_string(address, JSSF_SORT);
        fputs(s, stdout);
        free(s);

        putchar(')');
    }
```

dcbw:
that job is for openshift-sdn and this PR only touches ovn-kubernetes

PR #695, opened 2020-07-07
[WIP] Add support for IPv6 when Kuryr is used

luis5tb:


pkg/util/ip/addr.go
```diff
@@ -61,22 +61,23 @@ func lastIP(subnet net.IPNet) net.IP {
 
 // lastUsableIP returns second to last IP of a subnet
 func LastUsableIP(subnet net.IPNet) net.IP {
-	// FIXME(dulek): This should have proper IPv6 support, but for now whatever, Kuryr doesn't support it yet.
-	// This gives us last IP of the subnet…
 	ip := lastIP(subnet)
-	// …and this will be second to last (last usable)
-	ip[len(ip)-1] -= 1
-	return ip
+	return IterateIP(ip, -1)
```
danwinship, line 65: fwiw IPv6 doesn't have a subnet broadcast address like IPv4 does, so the "subtract one" step is only needed for IPv4

But anyway, consider if some or all of these functions could be replaced with `GetIndexedIP` or `BigForIP`+`AddIPOffset` from `k8s.io/utils/net`

squeed:
closing this for now; feel free to reopen.

PR #698, opened 2020-07-08
Bug 1856289: Block MCS and metadata for host-networking pods

dulek:
This commit adds an init container to kuryr-daemon pods that will
execute iptables commands to block access to machine-config-server and
metadata server for pods that run on host networking. Both should not be
required after ignition.

### DIFF
--- a/bindata/network/kuryr/005-daemon.yaml
+++ b/bindata/network/kuryr/005-daemon.yaml
@@ -24,6 +24,30 @@ spec:
       hostNetwork: true
       serviceAccountName: kuryr
       priorityClassName: system-node-critical
+      initContainers:
+        - name: block-mcs
+          image: {{ .CNIPluginsImage }}
+          securityContext:
+            privileged: true
+          command:
+          - /bin/sh
+          - -c
+          - |
+            #!/bin/sh
+
+            set -x -e
+
+            # Block MCS
+            iptables -A OUTPUT -p tcp -m tcp --dport 22623 -j REJECT
+            iptables -A OUTPUT -p tcp -m tcp --dport 22624 -j REJECT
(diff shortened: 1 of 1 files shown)


dulek:
I'm a bit not sure if this is required, as e.g. ovn-kubernetes does not block it. I'm checking the regular openshift-sdn now.

danwinship:
> I'm a bit not sure if this is required, as e.g. ovn-kubernetes does not block it

It does now. https://github.com/openshift/ovn-kubernetes/pull/170

There's some discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1849205 about the fact that we need to be doing this at a lower level rather than requiring every network plugin to implement this independently.

dulek:
> > I'm a bit not sure if this is required, as e.g. ovn-kubernetes does not block it
> 
> It does now. [openshift/ovn-kubernetes#170](https://github.com/openshift/ovn-kubernetes/pull/170)
> 
> There's some discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1849205 about the fact that we need to be doing this at a lower level rather than requiring every network plugin to implement this independently.

Oh, thank you for linking that patch, I wasn't aware of it and only checked ovn-kubernetes a while ago. I see it's not blocking the metadata server, should I remove that part from the patch?

danwinship:
yes, the eventual decision was that we only block the metadata server from pods, but we block the MCS from everywhere

MaysaMacedo:
Looks good to me. I assume it was tested?
Should a bz be linked?

dulek:
> Looks good to me. I assume it was tested?
> Should a bz be linked?

Yes, I tested it and it does work fine. Just created a BZ.

PR #701, opened 2020-07-08
Set timeoutSeconds to reflect expected timeout for OVS commands

alexanderConstantinescu:
Seen on a distant and heavily loaded CI cluster: if the cluster is heavily loaded we can expect `ovs-vsctl` to be slow. The commands executed by our `livenessProbe`/`readinessProbe` already take this into account, however the probes themselves do not. We thus restart OVS many more times than what we should, as those probes default to 1 second. This is not a real problem, but an optimization. The real problem is that we flush our DB when restarting, which is something which will need to be fixed later on.  

PS: I decided to set `timeoutSeconds` to whatever timeout the command has + a 1 second buffer

I will file a separate PR for release-4.5....but I will wait with filing that until tomorrow, once we, possibly, find and define a fix for the DB flush. As such there are less back-ports, CI runs etc.

/assign @dcbw @juanluisvaladas

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -161,6 +161,7 @@ spec:
               /usr/bin/ovs-vsctl -t 5 show > /dev/null
           initialDelaySeconds: 15
           periodSeconds: 5
+          timeoutSeconds: 6
         terminationGracePeriodSeconds: 45
       nodeSelector:
         kubernetes.io/os: linux
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -123,6 +123,7 @@ spec:
               /usr/bin/ovs-vsctl -t 5 show > /dev/null
           initialDelaySeconds: 15
           periodSeconds: 5
+          timeoutSeconds: 6
         terminationGracePeriodSeconds: 10
       nodeSelector:
         beta.kubernetes.io/os: "linux"


alexanderConstantinescu:
> * I try to find a better mechanism than timeouts. We will scale up and load the system to an unknown degree and picking a "good enough" value will be difficult.

I agree about the value being vaguely picked. But the values should reflect the commands timeout value. Right now, it's just wrong. And concerning "the better mechanism than timeouts" you mention, I am not sure what you are referring to. 

> * ovs should not be restarted except on rpm upgrade. ovs is being moved out of the cluster and to the system to avoid restarts during upgrade (which are very destructive).

I agree. With better timeout values we should be restarting much less / not at all. I tested this on CI prod and OVS had 0 restarts with proper values.

> * I would not probe ovs at all. When it looks dead, maybe.

What do you mean? How would you know "when it looks dead" without a probe? That's what probes are for.

> * It is always safe (and required) to delete the db on restart since it is a cache of the kernel flows. It rebuilds itself during normal processing. (I think, and may be wrong) that starting ovs clears the kernel flows.

Yes, I agree, it should be safe. But it's not right now, but that's another PR and a separate problem.

> * The "show" command can take a long time when there are a lot of flows.

Sure, but here I am just setting the timeout to whatever timeout the show command expects.

juanluisvaladas:
> >  I try to find a better mechanism than timeouts. We will scale up and load the system to an unknown degree and picking a "good enough" value will be difficult.

> I agree about the value being vaguely picked. But the values should reflect the command's timeout value. Right now, it's just wrong. And concerning "the better mechanism than timeouts" you mention, I am not sure what you are referring to.

I don't think that's half bad, the openvswitch revalidator considers over 2 seconds is unreasonably long: https://github.com/openvswitch/ovs/blob/master/ofproto/ofproto-dpif-upcall.c#L2598

So tolerating 6 sounds reasonable.

I'm not saying it's absolutely flawless, but it's definitely an improvement.
/lgtm

pecameron:
/lgtm
I am still concerned that this will not work on highly loaded clusters with lots of pod creation/deletion. I am assuming that normally the cluster will be loaded to the point that another node needs to be added. (Why pay for a node you don't need?)

I was not clear on my need to probe comment above. We don't need to probe if a secondary method works, like another pod failing to access ovs or times out requests. Ovs and other software that relies on probes should have a fast path to respond to the probe. The app maintains internal state and the probe looks at it and immediately responds as needed. With that, the timeout mostly relates to network latency.

As a footnote: ovs is a mature and very stable part of RHEL. ovs/ovn were separated because ovs rarely needs an upgrade.

danwinship:
/hold
These are not liveness probes, they're readiness probes, which mean the pod doesn't get restarted if they fail, so this isn't what is causing pod restarts.

alexanderConstantinescu:
> /hold
> These are not liveness probes, they're readiness probes, which mean the pod doesn't get restarted if they fail, so this isn't what is causing pod restarts.

Yes, you're right. The original problem happens on 4.5 where they are liveness probes. But I thought it would be better to fix this across all branches in all places, and I don't see why we should have misaligned readiness probes either?

alexanderConstantinescu:
@danwinship : could you please explain/remove your hold?

PR #702, opened 2020-07-09
Make openshift-kube-apiserver netid 0

juanluisvaladas:
Some operators have networkPolicies to allow it which won't work
otherwise.


danwinship:
NetworkPolicy mode does not allow multiple namespaces to have the same VNID. This will break things.

NetworkPolicy podSelectors/namespaceSelectors do not apply to hostNetwork pods. If people are trying to use namespaceSelectors to allow connections from hostNetwork pods, then those NetworkPolicies are incorrect and _should_ fail.

We may need to do _something_, _somewhere_ about this, but this isn't it.

PR #703, opened 2020-07-09
bump api version to take the most up to date  openshift-api

JacobTanenbaum:
in order to make the VXLAN port for hybrid networking configurable on the linux side the api version of the openshift-api needs to be updated. 

Splitting the PR to make to make the VXLAN configurable into multiple parts


danwinship:
hm... I'm not sure if we have best practices on how to manage `go.mod`... @juanluisvaladas you did some work with go mod. Do you have any thoughts on if the updates here are done correctly?

pkg/controller/pki/event_recorder.go
```diff
@@ -37,3 +37,7 @@ func (r *loggingRecorder) WithComponentSuffix(componentNameSuffix string) events
 func (r *loggingRecorder) ComponentName() string {
 	return "not implemented"
 }
+
+func (r *loggingRecorder) Shutdown() {
+	//no implemented
```
juanluisvaladas, line 42: Nit: I think it should be "noT implemented"

juanluisvaladas:
> Do you have any thoughts on if the updates here are done correctly?

It looks pretty good to me, what I always do is modify go.mod, then run go mod verify, go mod tidy, and finally go mod vendor. It looks like this is exactly what's been done here.

Usually I think it's better to do the go mod vendor in a separate commit (so that you don't have a commit with 2 files modified somewhat manually and 4000 files modified in /vendor but I don't think it's relevant enough to ask to modify the branch history.

/lgtm.

pkg/controller/pki/event_recorder.go
JacobTanenbaum: oops typo, fixed

PR #707, opened 2020-07-13
Disable conntrack for GENEVE

mccv1r0:
Disabling conntrack for GENEVE


dcbw:
this test doesn't run OVN at all, and this change is only for OVN, so the test is flaky

dcbw:
@mccv1r0 I guess technically we should be getting the Geneve port from the config and using that instead, eg `{{.GenevePort}}` templated into the file from the CNO

PR #709, opened 2020-07-13
Bug 1852112: Ignore hostPrefix validation for non-(sdn/ovn) plugins

tssurya:
For plugins like Calico which don't use the hostPrefix for anything,
it does not make sense to validate this field and force customers
to set specific values just to pass validation considering the
field is not even used in these plugins. This patch ensures that
we perform the validation only for OpenShiftSDN and OVNKubernetes.

### DIFF
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -6,9 +6,13 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/pkg/errors"
 )
+
+// list of known plugins that require hostPrefix to be set
+var pluginsUsingHostPrefix = sets.NewString(string(operv1.NetworkTypeOpenShiftSDN), string(operv1.NetworkTypeOVNKubernetes))
 
 // ValidateClusterConfig ensures the cluster config is valid.
 func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
@@ -39,15 +43,18 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 		if err != nil {
 			return errors.Errorf("could not parse spec.clusterNetwork %s", cnet.CIDR)
 		}
-		ones, bits := cidr.Mask.Size()
-		// The comparison is inverted; smaller number is larger block
-		if cnet.HostPrefix < uint32(ones) {
-			return errors.Errorf("hostPrefix %d is larger than its cidr %s",
-				cnet.HostPrefix, cnet.CIDR)
-		}
-		if int(cnet.HostPrefix) > bits-2 {
-			return errors.Errorf("hostPrefix %d is too small, must be a /%d or larger",
-				cnet.HostPrefix, bits-2)
+		// ignore hostPrefix if the plugin does not use it and has it unset
+		if pluginsUsingHostPrefix.Has(clusterConfig.NetworkType) || (cnet.HostPrefix != 0) {
+			ones, bits := cidr.Mask.Size()
+			// The comparison is inverted; smaller number is larger block
+			if cnet.HostPrefix < uint32(ones) {
+				return errors.Errorf("hostPrefix %d is larger than its cidr %s",
+					cnet.HostPrefix, cnet.CIDR)
+			}
+			if int(cnet.HostPrefix) > bits-2 {
+				return errors.Errorf("hostPrefix %d is too small, must be a /%d or larger",
+					cnet.HostPrefix, bits-2)
+			}
 		}
 		if err := pool.Add(*cidr); err != nil {
 			return err
--- a/pkg/network/cluster_config_test.go
+++ b/pkg/network/cluster_config_test.go
@@ -59,8 +59,19 @@ func TestValidateClusterConfig(t *testing.T) {
 	haveError(cc, "CIDRs 192.168.0.0/20 and 192.168.2.0/23 overlap")
 
 	cc = *ClusterConfig.DeepCopy()
+	cc.ClusterNetwork[1].HostPrefix = 0
+	res := ValidateClusterConfig(cc)
+	// Since the NetworkType is None, and the hostprefix is unset we don't validate it
+	g.Expect(res).Should(BeNil())
+
+	cc = *ClusterConfig.DeepCopy()
 	cc.ClusterNetwork[1].HostPrefix = 21
 	haveError(cc, "hostPrefix 21 is larger than its cidr 10.2.0.0/22")
+
+	cc = *ClusterConfig.DeepCopy()
+	cc.NetworkType = "OpenShiftSDN"
+	cc.ClusterNetwork[1].HostPrefix = 0
+	haveError(cc, "hostPrefix 0 is larger than its cidr 10.2.0.0/22")
 
 	// network type
 	cc = *ClusterConfig.DeepCopy()
--- a/README.md
+++ b/README.md
@@ -112,12 +112,12 @@ Users must supply at least two address pools - ClusterNetwork for pods, and Serv
 
 For future expansion, multiple `serviceNetwork` entries are allowed by the configuration but not actually supported by any network plugins. Supplying multiple addresses is invalid.
 
-Each `clusterNetwork` entry has an additional required parameter, `hostPrefix`, that specifies the address size to assign to each individual node.  For example,
+Each `clusterNetwork` entry has an additional parameter, `hostPrefix`, that specifies the address size to assign to each individual node.  For example,
 ```yaml
 cidr: 10.128.0.0/14
 hostPrefix: 23
 ```
-means 512 nodes would get blocks of size `/23`, or 512 addresses.
+means 512 nodes would get blocks of size `/23`, or 512 addresses. If the hostPrefix field is not used by the plugin, it can be left unset.
 
 IP address pools are always read from the Cluster configuration and propagated "downwards" into the Operator configuration. Any changes to the Operator configuration are ignored.
 
--- a/manifests/0000_70_cluster-network-operator_01_crd.yaml
+++ b/manifests/0000_70_cluster-network-operator_01_crd.yaml
@@ -165,10 +165,10 @@ spec:
               items:
                 description: ClusterNetworkEntry is a subnet from which to allocate
                   PodIPs. A network of size HostPrefix (in CIDR notation) will be
-                  allocated when nodes join the cluster. Not all network providers
-                  support multiple ClusterNetworks
+                  allocated when nodes join the cluster. If HostPrefix is not used by the plugin,
+                  it can be left unset. Not all network providers support multiple ClusterNetworks
                 type: object
-                required: ["cidr", "hostPrefix"]
+                required: ["cidr"]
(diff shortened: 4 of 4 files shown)


pkg/network/cluster_config.go
```diff
@@ -14,6 +14,8 @@ import (
 func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 	// Check all networks for overlaps
 	pool := iputil.IPPool{}
+	// Maintain a list of plugins that use hostPrefix
+	pluginsUsingHostPrefix := []string{"OpenShiftSDN", "OVNKubernetes"}
```
danwinship, line 18: These are defined as constants, `operv1.NetworkTypeOpenShiftSDN` and `operv1.NetworkTypeOVNKubnernetes`

pkg/network/cluster_config.go
```diff
@@ -39,15 +41,20 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
...
-		if cnet.HostPrefix < uint32(ones) {
-			return errors.Errorf("hostPrefix %d is larger than its cidr %s",
-				cnet.HostPrefix, cnet.CIDR)
-		}
-		if int(cnet.HostPrefix) > bits-2 {
-			return errors.Errorf("hostPrefix %d is too small, must be a /%d or larger",
-				cnet.HostPrefix, bits-2)
+		//ignore hostPrefix if the plugin does not use it
+		for _, plugin := range pluginsUsingHostPrefix {
+			if plugin == clusterConfig.NetworkType {
```
danwinship, line 46: Looping over `pluginsUsingHostPrefix` like this is a weird way to check this. You should make a helper function `networkTypeUsesHostPrefix` or something, and just check that from here.

pkg/network/cluster_config.go
```diff
@@ -39,15 +41,20 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
...
-		ones, bits := cidr.Mask.Size()
-		// The comparison is inverted; smaller number is larger block
-		if cnet.HostPrefix < uint32(ones) {
-			return errors.Errorf("hostPrefix %d is larger than its cidr %s",
-				cnet.HostPrefix, cnet.CIDR)
-		}
-		if int(cnet.HostPrefix) > bits-2 {
-			return errors.Errorf("hostPrefix %d is too small, must be a /%d or larger",
-				cnet.HostPrefix, bits-2)
+		//ignore hostPrefix if the plugin does not use it
```
danwinship, line 44: space after `//`

pkg/network/cluster_config.go
danwinship: Also, we should not allow a "bad" HostPrefix even if it's not being used. We should allow it to be `0` if the plugin doesn't use HostPrefix, but if it's not 0, then it should still be required to be valid.

tssurya: > Also, we should not allow a "bad" HostPrefix even if it's not being used. We should allow it to be `0` if the plugin doesn't use HostPrefix, but if it's not 0, then it should still be required to be valid.

hmm ok yea so that means the plugins that don't use this field can set it to zero, but if they haven't set it to 0 they have to comply with our definition of a valid hostPrefix. That sounds reasonable and you are right - allowing to set anything in that field and not check is bad UI. I'll also try to add a document note somewhere for reference.

tssurya:
Thanks Dan for reviewing so quickly, all your comments are valid specially about the changes in the installer (which I didn't know about but of course makes sense which is why the installer script is failing for them :/).

pkg/network/cluster_config.go
tssurya: ack.

pkg/network/cluster_config.go
tssurya: ack.

tssurya:
> So this will not actually entirely fix the bug. You will also need a patch to openshift/installer, to change the validation of `install-config.yaml` in the same way (or they won't even be able to get the cluster installed.).

https://github.com/openshift/installer/pull/3888

> Additionally, we should mark the `HostPrefix` fields in the `ClusterNetworkEntry` types in `openshift/api/config/v1/` and `openshift/api/operator/v1/` as optional, so that they aren't shown in `oc get` output when they are unset/0.

(thinking out loud) Do we need to do this ? As per definition we do allow 0 as minimum hostPrefix which == unset. So what's wrong if oc get shows hostPrefix as 0 ? The user/customer using the plugin will be aware that they have it unset and hence it shouldn't be confusing right ?

pkg/network/cluster_config.go
```diff
@@ -131,3 +135,19 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec, oldStatus *configv1.
...
+// validated or not depending on if the networkType uses the hostPrefix.
+// It returns true if the networkType requires hostPrefix or
+// if the networkType that does not require a hostPrefix has a non-zero hostPrefix.
+func networkTypeRequiresHostPrefix(networkType string, hostPrefix uint32) bool {
+	// Maintain a list of plugins that use hostPrefix
+	pluginsUsingHostPrefix := sets.NewString(string(operv1.NetworkTypeOpenShiftSDN), string(operv1.NetworkTypeOVNKubernetes))
+	if pluginsUsingHostPrefix.Has(networkType) {
+		return true
+	}
+	if hostPrefix != 0 {
```
abhat, line 149: Why do we care what the hostPrefix value is if the networkType is not one of `OpenshiftSDN` or `OVNKubernetes`?

tssurya: > Why do we care what the hostPrefix value is if the networkType is not one of `OpenshiftSDN` or `OVNKubernetes`?

@danwinship suggested that we should still perform the check if the networkType has set the hostPrefix (anything other than a 0) which makes sense since we expect the plugin to follow the OCP rules. This way we give two options to the plugins that don't use this field:

-  keep it unset in which case we won't care or validate,
-  if they set it they should ensure they set it as per our rules.

we can document this to keep it clear.

tssurya:
> > Additionally, we should mark the `HostPrefix` fields in the `ClusterNetworkEntry` types in `openshift/api/config/v1/` and `openshift/api/operator/v1/` as optional, so that they aren't shown in `oc get` output when they are unset/0.
> 
> (thinking out loud) Do we need to do this ? As per definition we do allow 0 as minimum hostPrefix which == unset. So what's wrong if oc get shows hostPrefix as 0 ? The user/customer using the plugin will be aware that they have it unset and hence it shouldn't be confusing right ?

ok, I think I found the answer. I don't know whether we should block oc get from displaying the hostPrefix when its 0, but I do realize I have to make changes to openshift/api repo nevertheless because all are interconnected. If I change it in the manifests of CNO to make `hostPrefix` as a not-required field I'll have to reflect this in the api repo.

related commit: https://github.com/openshift/api/pull/697

pkg/network/cluster_config.go
```diff
@@ -39,15 +40,18 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
...
-		// The comparison is inverted; smaller number is larger block
-		if cnet.HostPrefix < uint32(ones) {
-			return errors.Errorf("hostPrefix %d is larger than its cidr %s",
-				cnet.HostPrefix, cnet.CIDR)
-		}
-		if int(cnet.HostPrefix) > bits-2 {
-			return errors.Errorf("hostPrefix %d is too small, must be a /%d or larger",
-				cnet.HostPrefix, bits-2)
+		// ignore hostPrefix if the plugin does not use it and has it unset
+		if networkTypeRequiresHostPrefix(clusterConfig.NetworkType) || (!networkTypeRequiresHostPrefix(clusterConfig.NetworkType) && cnet.HostPrefix != 0) {
```
abhat, line 44: you don't need the `!networkTypeRequiresHostPrefix()` check since the || condition will get shorted at the first true.

pkg/network/cluster_config.go
abhat: I am not sure I follow. If the plugin is not the one managed by us, why should we care whether the hostPrefix gets used, is set, or not? I think @danwinship 's comment to make it optional makes sense so a field that is not set is not shown in the output of `oc get`

tssurya: > I am not sure I follow. If the plugin is not the one managed by us, why should we care whether the hostPrefix gets used, is set, or not? 

So when we make the field optional and the customer does not specify a value, the default value is zero in which case we skip validation. Initially I had done what you suggested which is to completely skip validation if its not either sdn or ovn. But what Dan suggested was that we should not allow anyone to set _any possible_ value to this field just because its not used which is why I changed it to include the validation if the plugin still sets it.

Basically CNO has a set of checks for hostPrefix field which every plugin should follow. If a plugin does not use it or does not want to follow it, they can unset it. However if they set this field, then they have to ensure they follow CNO's validation rules. See his comment https://github.com/openshift/cluster-network-operator/pull/709#discussion_r453935277.

> I think @danwinship 's comment to make it optional makes sense so a field that is not set is not shown in the output of `oc get`

yea I have made in optional in openshift/api#697 so that it is not shown in the output.

pkg/network/cluster_config.go
tssurya: oh hey yea you are right, stupid me...

pkg/network/cluster_config.go
```diff
@@ -131,3 +135,14 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec, oldStatus *configv1.
...
 	return &status
 }
+
+// networkTypeRequiresHostPrefix checks whether the networkType requires the hostPrefix field.
+// It returns true if the networkType requires hostPrefix.
+func networkTypeRequiresHostPrefix(networkType string) bool {
+	// list of known plugins that require hostPrefix to be set
+	pluginsUsingHostPrefix := sets.NewString(string(operv1.NetworkTypeOpenShiftSDN), string(operv1.NetworkTypeOVNKubernetes))
+	if pluginsUsingHostPrefix.Has(networkType) {
+		return true
```
danwinship, line 145: `pluginsUsingHostPrefix` should be defined as a top-level variable rather than being redefined every time `networkTypeRequiresHostPrefix` is called.

And in fact, once you do that, maybe we don't even need `networkTypeRequiresHostPrefix`. (Yeah, I know I had suggested it before but that was because there was more code then so it was cleaner to abstract it out. But if you move `pluginsUsingHostPrefix` out of the function then all you have to do is just change `if networkTypeRequiresHostPrefix(clusterConfig.NetworkType)` to `if pluginsUsingHostPrefix.Has(clusterConfig.NetworkType)` and then you don't need the helper func

pkg/network/cluster_config_test.go
```diff
@@ -58,10 +58,20 @@ func TestValidateClusterConfig(t *testing.T) {
 	cc.ClusterNetwork[0].CIDR = "192.168.2.0/23"
 	haveError(cc, "CIDRs 192.168.0.0/20 and 192.168.2.0/23 overlap")
 
+	cc = *ClusterConfig.DeepCopy()
+	cc.ClusterNetwork[1].HostPrefix = 0
+	res := ValidateClusterConfig(cc)
+	g.Expect(res).Should(BeNil())
```
danwinship, line 64: maybe add a comment here noting that `cc.NetworkType` is `None` so we don't validate `HostPrefix`

pkg/network/cluster_config.go
tssurya: yea makes sense,

PR #710, opened 2020-07-14
ovn-kubernetes: record ovn-northd pid

danwinship:
New upstream metrics require this, so ovn-appctl can find northd


dcbw:
this PR is for ovnkube, and the test is for openshift-sdn, no possible effect of this PR on that test

PR #711, opened 2020-07-15
Enable multicast support by default on the master

abhat:
The daemonset file currently doesn't enable multicast support by default. This change is to pass that as an argument during starting of the ovnkube-master pod.


danwinship:
hm...
/lgtm
but this should be fixed upstream to allow specifying this in the config file

abhat:
@dcbw mentioned that we may not want to do this upstream by default since it does incur some penalty.

danwinship:
Right, not saying we should change the default upstream, just that you should be able to specify the option in the config file rather than having to do it on the command line.

PR #712, opened 2020-07-16
Bug 1857738: [release-4.5] Set timeoutSeconds to reflect expected timeout for OVS commands

alexanderConstantinescu:
This is the release-4.5 back-port of https://github.com/openshift/cluster-network-operator/pull/701

Unlike https://github.com/openshift/cluster-network-operator/pull/701 we need to set the values in different places as the code varies a lot between these versions, so no cherry-pick involved here, however same logic applies to the value definition.


alexanderConstantinescu:
/retitle Bug 1857738: [release-4.5] Set timeoutSeconds to reflect expected timeout for OVS commands

PR #714, opened 2020-07-16
Bug 1856130: [release-4.5] bump golang.org/x/text/encoding to v0.3.3

aojea:
CVE-2020-14040 golang.org/x/text: possibility to trigger an
infinite loop in encoding/unicode could lead to crash.

Upgrade to golang.org/x/text/encoding to v0.3.3


alexanderConstantinescu:
Just FMI (for my information...): I didn't see a master PR with this patch, do you have one @aojea ?

aojea:
> Just FMI (for my information...): I didn't see a master PR with this patch, do you have one @aojea ?

Master is not affected @alexanderConstantinescu 

The fix is in golang.org/x/text v0.3.3,

- master is not affected since already vendor that version
https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224
- 4.7 ok
https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224
- 4.6 ok
https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224

4.5 and lower needs to upgrade the module to the v0.3.3 version

alexanderConstantinescu:
> > Just FMI (for my information...): I didn't see a master PR with this patch, do you have one @aojea ?
> 
> Master is not affected @alexanderConstantinescu
> 
> The fix is in golang.org/x/text v0.3.3,
> 
> * master is not affected since already vendor that version
>   https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224
> * 4.7 ok
>   https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224
> * 4.6 ok
>   https://github.com/openshift/cluster-network-operator/blob/e907dc569c81fd76712904545fa149f1736b018d/vendor/modules.txt#L224
> 
> 4.5 and lower needs to upgrade the module to the v0.3.3 version

ACK, thanks for clarifying. We might need a "dummy bug" in that case that targets 4.6, so that this PR can go in.

knobunc:
@aojea the bot will be fixed when https://github.com/openshift/release/pull/10440 merges.

PR #718, opened 2020-07-20
Bug 1858712: Revert "[release-4.5] Bug 1854072: Check if the db file has been correctly initialized with raft.

abhat:
Reverts openshift/cluster-network-operator#693


wking:
/retitle Bug 1854072: Revert "[release-4.5] Bug 1854072: Check if the db file has been correctly initialized with raft."

sdodson:
> /retitle Bug 1854072: Revert "[release-4.5] Bug 1854072: Check if the db file has been correctly initialized with raft."

I don't think that was the right thing to do since we're fixing the regresison but not the original bug.

/retitle Bug 1858712: Revert "[release-4.5] Bug 1854072: Check if the db file has been correctly initialized with raft.

PR #720, opened 2020-07-21
Bug 1846117: Bump ovn default MTU to 1442

tssurya:
This patch changes the default ovn MTU value to 1442.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -193,7 +193,14 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int) 
 	// (which may not be a node in the cluster).
 	// However, this can never change, so we always prefer previous.
 	if sc.MTU == nil {
-		var mtu uint32 = uint32(hostMTU) - 100 // 100 byte geneve header
+		var mtu uint32 = uint32(hostMTU) - 58 // 58 bytes geneve header
+		// if it's an ipv6 cluster or dual stack, ip header would be an extra 20 bytes.
+		for _, cn := range conf.ClusterNetwork {
+			if utilnet.IsIPv6CIDRString(cn.CIDR) {
+				mtu = mtu - 20
+				break
+			}
+		}
 		if previous != nil && previous.DefaultNetwork.OVNKubernetesConfig != nil &&
 			previous.DefaultNetwork.OVNKubernetesConfig.MTU != nil {
 			mtu = *previous.DefaultNetwork.OVNKubernetesConfig.MTU
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -189,7 +189,7 @@ func TestFillOVNKubernetesDefaults(t *testing.T) {
 	conf.DefaultNetwork.OVNKubernetesConfig = nil
 
 	// vars
-	m := uint32(8900)
+	m := uint32(8942)
 	p := uint32(6081)
 
 	expected := operv1.NetworkSpec{
@@ -214,6 +214,40 @@ func TestFillOVNKubernetesDefaults(t *testing.T) {
 	}
 
 	fillOVNKubernetesDefaults(conf, nil, 9000)
+
+	g.Expect(conf).To(Equal(&expected))
+
+	conf.ServiceNetwork = []string{
+		"172.30.0.0/16",
+		"fd02::/112",
+	}
+	conf.ClusterNetwork = append(config.ClusterNetwork, operv1.ClusterNetworkEntry{
+		CIDR: "fd01::/48", HostPrefix: 64,
+	})
+	conf.DefaultNetwork.OVNKubernetesConfig = nil
+	m = uint32(1422)
+	expected = operv1.NetworkSpec{
+		ServiceNetwork: []string{"172.30.0.0/16", "fd02::/112"},
+		ClusterNetwork: []operv1.ClusterNetworkEntry{
+			{
+				CIDR:       "10.128.0.0/14",
+				HostPrefix: 23,
+			},
+			{
+				CIDR:       "fd01::/48",
+				HostPrefix: 64,
+			},
+		},
+		DefaultNetwork: operv1.DefaultNetworkDefinition{
+			Type: operv1.NetworkTypeOVNKubernetes,
+			OVNKubernetesConfig: &operv1.OVNKubernetesConfig{
+				MTU:        &m,
+				GenevePort: &p,
+			},
+		},
+	}
+
+	fillOVNKubernetesDefaults(conf, nil, 1500)
 
 	g.Expect(conf).To(Equal(&expected))
 


pkg/network/ovn_kubernetes.go
```diff
@@ -193,7 +193,13 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 	// (which may not be a node in the cluster).
 	// However, this can never change, so we always prefer previous.
 	if sc.MTU == nil {
-		var mtu uint32 = uint32(hostMTU) - 100 // 100 byte geneve header
+		var mtu uint32 = uint32(hostMTU) - 50 // 50 byte geneve header
+		for _, cn := range conf.ClusterNetwork {
```
tssurya, line 197: ok, based on the comment in https://issues.redhat.com/browse/SDN-1063 from Dan, I am not really sure if this is the best of determining if the cluster is ipv6 or not.

pkg/network/ovn_kubernetes.go
```diff
@@ -193,7 +193,13 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 	// (which may not be a node in the cluster).
 	// However, this can never change, so we always prefer previous.
 	if sc.MTU == nil {
-		var mtu uint32 = uint32(hostMTU) - 100 // 100 byte geneve header
+		var mtu uint32 = uint32(hostMTU) - 50 // 50 byte geneve header
```
smalleni, line 196: I've seen the packets to be 1442 bytes with geneve. Looks like the geneve header is 58 bytes instead of 50.

tssurya: > I've seen the packets to be 1442 bytes with geneve. Looks like the geneve header is 58 bytes instead of 50.

ack, thanks Sai, I'll dig more into this and fix the number.

danwinship:
I can't find the place where we did the math before...

The MTU of an interface is the maximum size of an _ethernet payload_. So, eg, an interface with an MTU can carry packets of (at least) 1518 bytes: a 14-byte ethernet header plus a 1500-byte payload plus a 4-byte checksum. The 1500 byte payload could be, eg, a UDPv4 packet consisting of a 20-byte IPv4 header, an 8-byte UDP header, and a 1472-byte UDP payload.

If that UDP payload is a VXLAN packet, then (per [RFC 7348](https://tools.ietf.org/html/rfc7348#section-5)) it consists of 8 bytes for the fixed VXLAN header, followed by 14 bytes for the inner ethernet header, followed by the inner ethernet payload. (There is no inner trailing checksum.) That means the MTU (largest inner ethernet payload) of the VXLAN tunnel is 1500-20-8-8-14=1450 bytes, ie 50 bytes less than the MTU of the parent interface. OTOH, if the VXLAN packet is being sent over IPv6 then the outer UDPv6 packet has a 40-byte IPv6 header (assuming no IPv6 options) leaving less room for the inner ethernet payload, so the inner MTU is 1430. (It doesn't matter whether the _inner_ packets are IPv4 or IPv6 because the MTU gives the size of the inner ethernet payload, not the inner TCP/UDP payload.)

For Geneve (per [draft-ietf-nvo3-geneve-16](https://tools.ietf.org/html/draft-ietf-nvo3-geneve-16#section-3)), there is also a fixed 8-byte header, but also possibly options. According to [ovn-architecture(7)](http://www.openvswitch.org/support/dist-docs/ovn-architecture.7.html), OVN uses a single additional 32-bit option for datapath information, so that's an extra 4 bytes of Geneve metadata describing the option, and 4 bytes of actual OVN datapath data. So indeed, the MTU for OVN-Kubernetes should be 58 less than the parent MTU, not 50 less. I got this wrong in my comment in Jira.

So:
  - OpenShift SDN: base MTU minus 50
  - OVN Kubernetes: base MTU minus 58
  - IPv6/dual-stack (either): subtract another 20 bytes

And, also, if we implement "namespace tagging" / "packet marking" / whatever, that will probably require another 8 bytes in the OVN-Kubernetes case. This means that to deal with the possibility of that in the future, plus the possibility of upgrading clusters from single-stack IPv4 to dual-stack, we need to make sure it will be possible for us to change the MTU of the tunnel in a live cluster in the future.

tssurya:
> I can't find the place where we did the math before...
> 
> The MTU of an interface is the maximum size of an _ethernet payload_. So, eg, an interface with an MTU of 1500 can carry packets of (at least) 1518 bytes: a 14-byte ethernet header plus a 1500-byte payload plus a 4-byte checksum. The 1500 byte payload could be, eg, a UDPv4 packet consisting of a 20-byte IPv4 header, an 8-byte UDP header, and a 1472-byte UDP payload.
> 
> If that UDP payload is a VXLAN packet, then (per [RFC 7348](https://tools.ietf.org/html/rfc7348#section-5)) it consists of 8 bytes for the fixed VXLAN header, followed by 14 bytes for the inner ethernet header, followed by the inner ethernet payload. (There is no inner trailing checksum.) That means the MTU (largest inner ethernet payload) of the VXLAN tunnel is 1500-20-8-8-14=1450 bytes, ie 50 bytes less than the MTU of the parent interface. OTOH, if the VXLAN packet is being sent over IPv6 then the outer UDPv6 packet has a 40-byte IPv6 header (assuming no IPv6 options) leaving less room for the inner ethernet payload, so the inner MTU is 1430. (It doesn't matter whether the _inner_ packets are IPv4 or IPv6 because the MTU gives the size of the inner ethernet payload, not the inner TCP/UDP payload.)
> 
> For Geneve (per [draft-ietf-nvo3-geneve-16](https://tools.ietf.org/html/draft-ietf-nvo3-geneve-16#section-3)), there is also a fixed 8-byte header, but also possibly options. According to [ovn-architecture(7)](http://www.openvswitch.org/support/dist-docs/ovn-architecture.7.html), OVN uses a single additional 32-bit option for datapath information, so that's an extra 4 bytes of Geneve metadata describing the option, and 4 bytes of actual OVN datapath data. So indeed, the MTU for OVN-Kubernetes should be 58 less than the parent MTU, not 50 less. I got this wrong in my comment in Jira.
> 
> So:
> 
> * OpenShift SDN: base MTU minus 50
> * OVN Kubernetes: base MTU minus 58
> * IPv6/dual-stack (either): subtract another 20 bytes
> 

Oh this is beautifully explained! Thanks a ton Dan! So, for OVN its 1500-20(v4)-8(udp)-16(geneve)-14(inner-eth-hdr) = 1442. Plus another 20 if its ipv6 cluster.

> And, also, if we implement "namespace tagging" / "packet marking" / whatever, that will probably require another 8 bytes in the OVN-Kubernetes case. This means that to deal with the possibility of that in the future, plus the possibility of upgrading clusters from single-stack IPv4 to dual-stack, we need to make sure it will be possible for us to change the MTU of the tunnel in a live cluster in the future.

yea that would be ideal.

pkg/network/ovn_kubernetes.go
tssurya: You and Dan are right, did a packet capture and it sums upto 58 if there is nothing extra.

pkg/network/ovn_kubernetes.go
tssurya: @danwinship : PTAL whenever you have some time. The corresponding issue's priority was made high.

danwinship: This is not an urgent fix; if individual customers don't like the default MTU they can change it in their own cluster. (Additionally, based on discussion in other channels, the particular customer who wants this change actually wants it to be 1450, not 1442 anyway, so this won't even help them.)

I don't think we should merge this without also having code to allow changing the MTU in a deployed cluster.

danielmellado: +1, I don't think hardcoding MTU would do any good, that said, if there's a need to change the default value as per the following conversation, I'm good for it. But otherwise we'd be changing the hardcoded value as per every now and then customer's request.

tssurya: > +1, I don't think hardcoding MTU would do any good, that said, if there's a need to change the default value as per the following conversation, I'm good for it. But otherwise we'd be changing the hardcoded value as per every now and then customer's request.

agreed, hard-coding default MTU is not something this patch does. I was talking about the header length change.

tssurya:
@danwinship: based on your comment here: https://github.com/openshift/cluster-network-operator/pull/720#discussion_r467012539 , shall we close this PR and the corresponding bug (since we don't want to merge this now...) till we get the real feature of dymanic mtu change in a live cluster in ? I'd like to just run it by you before I close this.

danwinship:
The work that's here is correct; we just shouldn't merge it yet. I'm not sure we want to close the PR.

tssurya:
> The work that's here is correct; we just shouldn't merge it yet. I'm not sure we want to close the PR.

ack, ok let's leave this open then.

PR #732, opened 2020-07-29
Kuryr: Add new Custom Resource Definitions

dulek:
With 4.6 Kuryr uses KuryrPort custom resources to save Neutron port
information instead of annotating them on the pod. Also the
KuryrNetPolicy is replaced by updated format of KuryrNetworkPolicy CR.
This commit makes sure new CRDs are added, RBAC extended and Kuryr
configuration updated to support this.

### DIFF
--- a/bindata/network/kuryr/001-rbac.yaml
+++ b/bindata/network/kuryr/001-rbac.yaml
@@ -45,6 +45,9 @@ rules:
   - kuryrnets
   - kuryrnetworks
   - kuryrnetpolicies
+  - kuryrnetworkpolicies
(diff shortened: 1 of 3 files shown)


dulek:
/hold

This needs to wait for Kuryr patch.

dcbw:
`[sig-network] Conntrack should be able to preserve UDP traffic when server pod cycles for a NodePort service [Suite:openshift/conformance/parallel] [Suite:k8s]` is known flaky upstream and upstream Kube is doing a PR for it
/override ci/prow/e2e-gcp-ovn

PR #729, opened 2020-07-28
Bug 1819571: NetworkAttachmentDefinitions should have output for "oc explain" 

dougbtv:
This adds the apiversion, kind and metadata fields for explain to properly function.

@s1061123 can you check and see if this look OK? If it does, we can move forward with backports. Thanks!

PR #736, opened 2020-07-30
Bug 1823711: [backport 4.3] NetworkAttachmentDefinitions should have output for "oc explain"

dougbtv:
This adds the apiversion, kind and metadata fields for explain to properly function.

Backport of #729


dougbtv:
We decided to close the BZ as WONTFIX due to the minor impact on such an old version

PR #737, opened 2020-07-30
Add required setup for egressFirewall and egressIP

JacobTanenbaum:
grant ovn-kubernetes the ability to get,list, watch, and update egressfirewall and egressIP objects; add get, list, and watch to CustomResourceDefinitions in order to dynamically enable/disable egressfirewall based on the presence of its defining CRD

Also add the CRD for egressFirewall to the bindata for ovn-kubernetes

PR #739, opened 2020-07-31
Fixes permissions for CRDs with OVN

trozet:
With addition of Egress Firewall CRD, we now need to be able to watch
for this custom resource.


trozet:
close in favor of #737

PR #742, opened 2020-08-03
OVN: Fix control plane metrics alert job name

tssurya:
Following the change in PR #581 the job name should be renamed to
"ovnkube-master-metrics"

### DIFF
--- a/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
@@ -15,9 +15,9 @@ spec:
     - alert: NoRunningOvnMaster
       annotations:
         message: |
-          there is no running ovn-kubernetes master
+          There is no running ovn-kubernetes master
       expr: |
-        absent(up{job="ovnkube-master",namespace="openshift-ovn-kubernetes"} ==1)
+        absent(up{job="ovnkube-master-metrics", namespace="openshift-ovn-kubernetes"} == 1)
       for: 10m
       labels:
         severity: warning


dcbw:
@tssurya I would actually use "ovnkube-db" since that's the service that actually fronts the master pods. If we use the metrics service, then technically we're just watching to see if the metrics pods are running, I think. And the ovnkube master pods and the metrics proxy service are independant daemonsets

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -17,7 +17,7 @@ spec:
         message: |
           there is no running ovn-kubernetes master
       expr: |
-        absent(up{job="ovnkube-master",namespace="openshift-ovn-kubernetes"} ==1)
+        absent(up{job="ovnkube-master-metrics",namespace="openshift-ovn-kubernetes"} ==1)
```
abhat, line 20: If the job name is the name of the service, should it be `ovn-kubernetes-master-metrics` instead as per: [https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/monitor.yaml#L33](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/monitor.yaml#L33)

dcbw: The job name is a Kube Service.

dcbw: I think we should target the "ovnkube-db" service though per https://github.com/openshift/cluster-network-operator/pull/742#issuecomment-668134977

tssurya: I tried changing it to ovnkube-db, the alert still fires. I might have to add an additional ServiceMonitor for us to use this service from what I understand (https://coreos.com/blog/the-prometheus-operator.html)

Also I think its ok to leave it as ovnkube-master-metrics since in this instance the metrics pod's endpoint at port 9102 is indirectly connected to master container's 21902, it should be checking for the master pods' endpoint while scraping right ?

tssurya: Also I think the job name could be basically the label defined in the `jobLabel` field of the respective ServiceMonitor (https://coreos.com/operators/prometheus/docs/latest/api.html#servicemonitorspec)

tssurya: > Also I think its ok to leave it as ovnkube-master-metrics since in this instance the metrics pod's endpoint at port 9102 is indirectly connected to master container's 21902, it should be checking for the master pods' endpoint while scraping right ?

this assumption is wrong, it indeed only looks for metrics pods..

tssurya: all right so after some trial and errors, this is what I found: Whatever is the job name provided - in our case this is the `jobLabel` field value in the `ServiceMonitor` object which is `app: ovnkube-master` ; prometheus will scrape the endpoints which is the `instance:portno.` on all the pods which have the same label i.e `app:ovnkube-master` in our case. I think this is what @abhat  was also telling me on slack.

So for our particular alert query `up{job="ovnkube-master",namespace="openshift-ovn-kubernetes"} == 1` this is the output when all is well:



Name | endpoint | instance | job | namespace | pod | prometheus | service | Value
-- | -- | -- | -- | -- | -- | -- | -- | --
   up | metrics | 10.0.0.3:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-hnntq | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1
   up | metrics | 10.0.0.3:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-826kq | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1
   up | metrics | 10.0.0.4:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-frx9r | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1
   up | metrics | 10.0.0.4:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-xh5qb | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1
   up | metrics | 10.0.0.5:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-jppw6 | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1
   up | metrics | 10.0.0.5:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-zz42l | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 1

When I brought down the metrics pods this is what it looks like:
and the alert gets fired.

Name | endpoint | instance | job | namespace | pod | prometheus | service | Value
-- | -- | -- | -- | -- | -- | -- | -- | --
  up | metrics | 10.0.0.3:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-hnntq | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0
  up | metrics | 10.0.0.4:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-xh5qb | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0
  up | metrics | 10.0.0.5:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-jppw6 | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0


When I brought down the master pods this is what it looks like:
and the alert is fired.


Name | endpoint | instance | job | namespace | pod | prometheus | service | Value
-- | -- | -- | -- | -- | -- | -- | -- | --
   up | metrics | 10.0.0.3:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-ghvlw | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0
   up | metrics | 10.0.0.4:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-kmftc | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0
   up | metrics | 10.0.0.5:9102 | ovnkube-master | openshift-ovn-kubernetes | ovnkube-master-metrics-7jxjp | openshift-monitoring/k8s | ovn-kubernetes-master-metrics | 0


Another point to note is that this particular alert `NoRunningOVNKubeMaster` is duplicating what the global alert `TargetDown` does. TargetDown essentially scraps the up metrics for all openshift components. This alert is controlled by the cluster-monitoring-operator. Hence even if we don't have a seperate alert for OVNKube nodes, TargetDown takes care of this when the pod goes down (or used to take care, now it takes care of the ovnkube-node-metrics pods).

Nevertheless we can still continue to have our own alert since its been there from the beginning.

squeed:
Hang on, I'm not sure this is the best way to write this alert. We should instead be alerting on whether or not there is a leader. The query can then just be `max(ovnkube_master_leader) = 0`.

squeed:
An update: we should really alert on both conditions: a failure to scrape and no ovnkube master leader.

tssurya:
> An update: we should really alert on both conditions: a failure to scrape and no ovnkube master leader.

this metric `up` basically fails for both - if the endpoint is not configured properly and there is failure on scraping and if there is no ovnkube master pod which again results in not being able to get in touch with the pod.

tssurya:
> Hang on, I'm not sure this is the best way to write this alert. We should instead be alerting on whether or not there is a leader. The query can then just be `max(ovnkube_master_leader) = 0`.

ah ok, I guess its better to add this as well. From what I understand this alert will be fired if the leader election fails ? err, actually I don't know if there is a leader election in ovn like in sdn. Let me check.

trozet:
> > Hang on, I'm not sure this is the best way to write this alert. We should instead be alerting on whether or not there is a leader. The query can then just be `max(ovnkube_master_leader) = 0`.
> 
> ah ok, I guess its better to add this as well. From what I understand this alert will be fired if the leader election fails ? err, actually I don't know if there is a leader election in ovn like in sdn. Let me check.

Yeah there is. That is why gcp job alert was firing in the first place. In GCP there is some temporary outage to kapi server and the leader fails to renew his leadership lease. When that happens he shuts down and restarts and becomes leader again. The entire process takes about a second. This causes the alert to fire. However, the alert clause iiuc says "for 10m" which means as long as the pod comes back up within a 10 minute period the alert should not fire. Question is why does the alert fire if we were only down for a second? Does prometheus correctly detect that the pod is back up?

squeed:
Pending alerts will show up in the interface (but not actually fire), so maybe that's what's happening?

trozet:
> Pending alerts will show up in the interface (but not actually fire), so maybe that's what's happening?

I think it is firing

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/727/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn/1289934434367180800

fail [github.com/openshift/origin@/test/extended/util/prometheus/helpers.go:174]: Expected
    <map[string]error | len:1>: {
        "ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured|PrometheusRemoteWriteDesiredShards\",alertstate=\"firing\",severity!=\"info\"} >= 1": {
            s: "promQL query: ALERTS{alertname!~\"Watchdog|AlertmanagerReceiversNotConfigured|PrometheusRemoteWriteDesiredShards\",alertstate=\"firing\",severity!=\"info\"} >= 1 had reported incorrect results:\n[{\"metric\":{\"__name__\":\"ALERTS\",\"alertname\":\"NoRunningOvnMaster\",\"alertstate\":\"firing\",\"severity\":\"warning\"},\"value\":[1596381630.6,\"1\"]}]",
        },
    }
to be empty

tssurya:
> Yeah there is. That is why gcp job alert was firing in the first place. In GCP there is some temporary outage to kapi server and the leader fails to renew his leadership lease. When that happens he shuts down and restarts and becomes leader again. The entire process takes about a second. This causes the alert to fire. However, the alert clause iiuc says "for 10m" which means as long as the pod comes back up within a 10 minute period the alert should not fire. Question is why does the alert fire if we were only down for a second? Does prometheus correctly detect that the pod is back up?

Yes prometheus does come back to pending state from firing if the pod is back up - or at least it did in my nightly cluster.
Btw, this particular alert; the NoRunningOVNKubeMaster was always firing because it was incorrectly configured.

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -15,9 +15,9 @@ spec:
     - alert: NoRunningOvnMaster
       annotations:
         message: |
-          there is no running ovn-kubernetes master
+          There is no running ovn-kubernetes master
       expr: |
-        absent(up{job="ovnkube-master",namespace="openshift-ovn-kubernetes"} ==1)
+        absent(up{job="ovnkube-master-metrics", namespace="openshift-ovn-kubernetes"} == 1)
```
abhat, line 20: This can stay as is. The only thing we need is the `app: ovnkube-master` on the `ovnkube-db` service. Which it already does. So the only real change needed is in the `ServiceMonitor` where the `matchLabels` stanza should have `app: ovnkube-master`?

abhat:
So the current alert as it stands should just pick up `ovnkube-master` pods being down. Like @squeed mentioned we need separate alert for when there is no leader. And I am not so sure if we need to alert if the metrics scraper pods are missing. I mean it's good to know if the ovnkube-master-metrics pods are down, but the real deal is the ovnkube-master pods being down. Thoughts?

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
tssurya: > This can stay as is. The only thing we need is the `app: ovnkube-master` on the `ovnkube-db` service. Which it already does. So the only real change needed is in the `ServiceMonitor` where the `matchLabels` stanza should have `app: ovnkube-master`?

umm no, I mean if we do that - as in use the `ovnkube-db` service then we have to expose the metrics port on the `ovnkube-db` service so that prometheus can scrape that endpoint which basically defeats the purpose of having a `ovnkube-master-metrics` service.

abhat: I am thoroughly confused. The selector for the "app" it monitors is orthogonal to the port(s) it monitors on. 

```
endpoints:
  - interval: 30s
    port: metrics
    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
    scheme: https
    tlsConfig:
      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
      serverName: ovn-kubernetes-master-metrics.openshift-ovn-kubernetes.svc
``` 
is telling where to get the metrics from. The 

```
  selector:
    matchLabels:
      app: ovnkube-master
``` 
is telling what the "job" is. So technically you can still setup the endpoint to get the metrics from to be the port (`metrics`) specified on the [ovnkube-master-metrics service ](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/monitor.yaml#L28) and have that check for pods that have `app:ovnkube-master` label. No? Is my understanding incorrect?

dcbw:
> So the current alert as it stands should just pick up `ovnkube-master` pods being down. Like @squeed mentioned we need separate alert for when there is no leader. And I am not so sure if we need to alert if the metrics scraper pods are missing. I mean it's good to know if the ovnkube-master-metrics pods are down, but the real deal is the ovnkube-master pods being down. Thoughts?

IMO this can be 3 different things, in priority order:

1) alert when no ovnkube master pod is running in the cluster
2) alert when there is no ovnkube master leader
3) metrics pods are down

Each one can be a separate alert and thus PR. #1 is the most pressing need to get our CI flakes taken care of. So let's do that first, and then file Jira cards for the others?

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
tssurya: So Juan and I tested this, and we had to change the labels on the `ovn-kubernetes-master-metrics` service to match the master pods labels for the metric to be pulled from master pods. So we use the `ovn-kubernetes-master-metrics` instead of `ovnkube-db`

bindata/network/ovn-kubernetes/monitor.yaml
```diff
@@ -23,20 +23,20 @@ spec:
...
     matchLabels:
-      app: ovnkube-master-metrics
+      app: ovnkube-master
 ---
 apiVersion: v1
 kind: Service
 metadata:
   labels:
-    app: ovnkube-master-metrics
+    app: ovnkube-master
```
abhat, line 32: This should stay as `ovnkube-master-metrics`.

bindata/network/ovn-kubernetes/monitor.yaml
```diff
@@ -23,20 +23,20 @@ spec:
...
   name: ovn-kubernetes-master-metrics
   namespace: openshift-ovn-kubernetes
   annotations:
     service.beta.openshift.io/serving-cert-secret-name: ovn-master-metrics-cert
 spec:
   selector:
-    app: ovnkube-master-metrics
+    app: ovnkube-master
```
abhat, line 39: This makes sense.

abhat: This need not change. I think had it backwards.

bindata/network/ovn-kubernetes/monitor.yaml
tssurya: After our call yesterday, I tried it but it looks like the ports are not orthogonal. I haven't yet found a clear documentation on Prometheus that explains this. I have asked the monitoring-team to confirm this. So basically the scraping works only if I add a `metrics` port to `ovnkube-db` service as well.

So the way I see it, now we have two options:

1) Either we go back to my previous solution where we sort of umbrella all 6 pods under the `ovn-kubernetes-master-metrics` service and not use the `ovnkube-db` service at all - when it comes to metrics scraping.
2) We use both `ovnkube-db` and `ovn-kubernetes-master-metrics` but then define/expose the 9012 port on both. - I don't know if this will have an (undoing) effect on the TLS work done and all the http/https scraping stuff.

tssurya:
> @tssurya I would actually use "ovnkube-db" since that's the service that actually fronts the master pods. If we use the metrics service, then technically we're just watching to see if the metrics proxy pods are running, I think. And the ovnkube master pods and the metrics proxy pods are independent daemonsets

So we tested this. Even if we use the metrics service, when the master pods go down, it will be detected since the 9102 port used for scraping is proxying the 29102 localhost port.

tssurya:
The e2e-gcp-ovn job is failing on the Prometheus tests:

-  [sig-instrumentation] Prometheus when installed on the cluster should have important platform topology metrics [Suite:openshift/conformance/parallel]
- [sig-instrumentation][sig-builds][Feature:Builds] Prometheus when installed on the cluster should start and expose a secured proxy and verify build metrics [Suite:openshift/conformance/parallel] 
- [sig-instrumentation] Prometheus when installed on the cluster should have non-Pod host cAdvisor metrics [Suite:openshift/conformance/parallel] 
- [sig-instrumentation] Prometheus when installed on the cluster should have a AlertmanagerReceiversNotConfigured alert in firing state [Suite:openshift/conformance/parallel]
- [sig-instrumentation] Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics [Suite:openshift/conformance/parallel]
- [sig-instrumentation] Prometheus when installed on the cluster should report telemetry if a cloud.openshift.com token is present [Suite:openshift/conformance/parallel] 

Most of them are related to the the first failure `Prometheus when installed on the cluster should have important platform topology metrics ` I think and is being investigated here: https://bugzilla.redhat.com/show_bug.cgi?id=1852919

dcbw:
/override ci/prow/e2e-gcp-ovn

last 4 gcp test failures have either been `[sig-storage] PersistentVolumes [Feature:vsphere][Feature:ReclaimPolicy] [sig-storage] persistentvolumereclaim:vsphere [Feature:vsphere] should retain persistent volume when reclaimPolicy set to retain when associated claim is deleted [Suite:openshift/conformance/parallel] [Suite:k8s]` or a few storage related ones about local volumes.

PR #743, opened 2020-08-03
ovnkube: fix NoRunningOvnKubeMaster alert rule

dcbw:
The "job" must be a Kubernetes Service object that selects all the
backend pods we care about. We have two options:

1) ovnkube-db service: selects all ovnkube master pods

2) ovn-kubernetes-master-metrics service: selects all master metrics
pods which are a separate daemonset for proxying metrics over TLS

Go with #1 since that accurately represents the thing we really care
about: the ovnkube master pods.

@trozet @abhat @knobunc @squeed


dcbw:
Closing in favor of whatever happens in https://github.com/openshift/cluster-network-operator/pull/742

PR #744, opened 2020-08-04
DNM OSP e2e test

danielmellado:
This is a dummy patch to test OpenStack e2e tests. DO NOT MERGE


juanluisvaladas:
/hold
Just in case there is an accidental merge somehow

PR #746, opened 2020-08-05
WIP: Validate ovn db consistency and eventually delete it.

fedepaol:
This is a new attempt to continue what was reverted in https://github.com/openshift/cluster-network-operator/pull/723 , based on the discussion with @dceara on https://bugzilla.redhat.com/show_bug.cgi?id=1837953#c45 

There are some cases where the ovn db gets corrupted. The temporary remediation is to:
- identify the incosisntency (search the current id in prev_servers in the snapshot,
and in servers if not found there)
- delete the db
- kick the old server id from the raft cluster

To achieve that, the old id is passed from command to postStart via a file, as the
connection to the cluster needs to be established.

Leaving this WIP as I want to do some additional validation.
Also, it could be that the offending bug is fixed by the same fix for https://bugzilla.redhat.com/show_bug.cgi?id=1861087


fedepaol:
Update: I validated that deleting the master pods on a fresh cluster does not delete the dbs (as the previous implementation did).
I tried to force the "unhealthy status" by changing the regex, but the effect was that all the masters deleted their nbdb which I don't think it works (even the clusterid changed).

Trying to find a way to force only one master to perform the delete + kick steps.

fedepaol:
Update: by manually replacing the `if [[ -z ${match} ]]` with a check with the node ip, I validated the behaviour when removing the db on one node only.

On non leaders, it works as expected.
When deleting the leader db though, what I see is:
```
2020-08-07T15:53:59Z|00449|raft|INFO|ssl:10.0.0.5:44624: syntax "{"cluster":"a8fcb0ad-c219-4f33-b5b3-56cc62fff86e","comment":"heartbeat","from":"35fb413c-1792-47ac-8ac2-87d8b1755052","leader_commit":2901,"log":[],"prev_log_index":2901,"prev_log_term":12,"term":12,"to":"a0f7e381-a94c-4034-83ca-edf90f04e42a"}": syntax error: Parsing raft append_request RPC failed: wrong cluster a8fc (expected 9052)
2020-08-07T15:54:01Z|00450|raft|INFO|ssl:10.0.0.5:44624: syntax "{"cluster":"a8fcb0ad-c219-4f33-b5b3-56cc62fff86e","comment":"heartbeat","from":"35fb413c-1792-47ac-8ac2-87d8b1755052","leader_commit":2901,"log":[],"prev_log_index":2901,"prev_log_term":12,"term":12,"to":"a0f7e381-a94c-4034-83ca-edf90f04e42a"}": syntax error: Parsing raft append_request RPC failed: wrong cluster a8fc (expected 9052)
```

which I think makes sense, as the leader chose another clusterid when starting from scratch. At this point this feels like it's growing in complexity and I am afraid more unexpected dragons are behind the corner.

squeed:
> At this point this feels like it's growing in complexity and I am afraid more unexpected dragons are behind the corner.

Agreed. We should probably come up with a top-down solution for managing etcd set membership. The CNO already knows exactly *who* the members of the cluster should be. We can probably write a small controller that synchronizes that to ovn raft state.

Does it make sense to pull this out to a separate feature? Are there any obvious, failure-free cases we can handle with a simple shell script for now?

alexanderConstantinescu:
> > At this point this feels like it's growing in complexity and I am afraid more unexpected dragons are behind the corner.
> 
> Agreed. We should probably come up with a top-down solution for managing etcd set membership. The CNO already knows exactly _who_ the members of the cluster should be. We can probably write a small controller that synchronizes that to ovn raft state.
> 
> Does it make sense to pull this out to a separate feature? Are there any obvious, failure-free cases we can handle with a simple shell script for now?

I agree with this. And I've been thinking about it, and we are _almost_ there today. We already reconcile the bootstrap phase in the `operconfig_controller.go`, what we need to do is, change the `OVNBootstrapResult` to:

```go
type OVNBootstrapResult struct {
	MasterIPs []string
	Invite    []string // In case several masters join at once
	Kick      []string // In case several masters leave at once
	Bouncer   string   // Decided on one of MasterIPs to perform the kick(s)
}
```
The `Bouner` will kick out every master node in `Kick` which drops off the cluster. The nodes in `Invite` will only need to wipe the OVN DB location before starting (to ensure that no stale DB data might exist, in case that master already formed a part of the cluster). This should take care of all forms of replacing masters in a cluster, that being:

The same IP address leaves and comes back
```
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: All good | IP:1 ID:A | IP:2 ID:B | IP:3 ID:C | Bouncer:1 Kick:[] Invite:[] Masters:[1,2,3]  |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: 1 drops  |           | IP:2 ID:B | IP:3 ID:C | Bouncer:2 Kick:[1] Invite:[] Masters:[2,3]   |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: 1 joins  | IP:1 ID:D | IP:2 ID:B | IP:3 ID:C | Bouncer:1 Kick:[] Invite:[1] Masters:[1,2,3] |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
``` 

One IP address leaves and another joins
```
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: All good | IP:1 ID:A | IP:2 ID:B | IP:3 ID:C | Bouncer:1 Kick:[] Invite:[] Masters:[1,2,3]  |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: 1 drops  |           | IP:2 ID:B | IP:3 ID:C | Bouncer:2 Kick:[1] Invite:[] Masters:[2,3]   |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
| State: 4 joins  | IP:4 ID:D | IP:2 ID:B | IP:3 ID:C | Bouncer:4 Kick:[] Invite:[4] Masters:[4,2,3] |
+-----------------+-----------+-----------+-----------+----------------------------------------------+
``` 

Of course, the state of all of this would be stored in-memory in the CNO. Which makes all of this dependent upon _**not**_ having the node hosting the CNO being the node that drops of the cluster. To avoid that case we _could_ have the CNO annotate the ovnkube-master with the last set of `MasterIPs` that it saw before dying, and have the new CNO read that as to reconcile the state when it starts.

That should clean up this bash madness by a lot, the only doubt I still have is: what happens if the leader is the one leaving the cluster? But I assume the raft protocol will just assign the leadership to one of the remaining members, so it shouldn't be a problem if `1` is the leader, for example.

dceara:
> 
> That should clean up this bash madness by a lot, the only doubt I still have is: what happens if the leader is the one leaving the cluster? But I assume the raft protocol will just assign the leadership to one of the remaining members, so it shouldn't be a problem if `1` is the leader, for example.

That's correct, leadership will move to one of the remaining members, assuming that there are at least 2 members alive.

squeed:
Dumitru and I talked about being able to set the node's raft ID, so that we could easily determine **who** needs to be removed from the cluster. Otherwise, it's kind of a pain.

Another option would be for the ovn-k node to annotate itself with its raft ID, so we could compute and pass down the set of desired participants.

trozet:
> Dumitru and I talked about being able to set the node's raft ID, so that we could easily determine **who** needs to be removed from the cluster. Otherwise, it's kind of a pain.
> 
> Another option would be for the ovn-k node to annotate itself with its raft ID, so we could compute and pass down the set of desired participants.

@squeed I don't really understand this. We know who belongs in the cluster, because those very addresses are passed into ovn-kubernetes:
 servers: 23b1("tcp:[172.18.0.6]:6643"), 5c5d("tcp:[172.18.0.5]:6643"), f5e7("tcp:[172.18.0.3]:6643")

So whoever is in this list that was not passed to ovn-kubernetes can be kicked. Are you suggesting the problem scenario is when the same address is used again, but a different ID? I don't think that is possible, because I think the entry will be replaced in this list with the new ID for the same address.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -116,6 +116,35 @@ spec:
...
+            echo "Checking ${db} health"
+            serverid=$(ovsdb-tool show-log "${db}" | grep -m 1 -oP '(?<=server_id: )[[:xdigit:]]+')
+            # prev_servers is a list.  We are searching for " $server_id(" in the list
+            # prev_servers: 33b4("ssl:10.19.138.37:9643"), 720d("ssl:10.19.138.33:9643"), 7e4f("ssl:10.19.138.32:9643")
+            match=$(ovsdb-tool show-log "${db}" | grep "prev_servers:.* ${serverid}(" || true)
+            if [[ -z ${match} ]]; then
+                  echo "Current server_id $serverid not found in ${db} prev_servers list, checking servers"
+                  match=$(ovsdb-tool show-log "${db}" | grep "^ servers:.* ${serverid}(" || true)
+            fi
+            if [[ -z ${match} ]]; then
```
trozet, line 136: I'm not sure this is a comprehensive way to check if the db is healthy. There could be previous log records of the db joining the cluster, but then it could become corrupt later. We need to check if ovsdb-server is actually up, and if we can make queries to this particular db. For example you could have:

[root@ovn-control-plane3 ~]# ovsdb-tool show-log /etc/openvswitch/ovnnb_db.db | grep servers
 servers: 23b1("tcp:[172.18.0.6]:6643"), 5c5d("tcp:[172.18.0.5]:6643"), f5e7("tcp:[172.18.0.3]:6643")
 servers: 23b1("tcp:[172.18.0.6]:6643"), 5c5d("tcp:[172.18.0.5]:6643")

Where now f5e7 has left the cluster, but your grep would still see it there and think everything is OK.

fedepaol: That part was added as the result of the discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1837953#c40 .
Basically, if the snapshot does not happen, the current id is not dumped in the prev_servers list and we attempt to look for it.
Also, not sure if what you are describing could happen (or is a remote corner case) as we are looking for the current serverid, which I don't expect to disappear from the server list.

fedepaol:
Sorry I am late to the party. 

> > Dumitru and I talked about being able to set the node's raft ID, so that we could easily determine **who** needs to be removed from the cluster. Otherwise, it's kind of a pain.
> > Another option would be for the ovn-k node to annotate itself with its raft ID, so we could compute and pass down the set of desired participants.
> 
> @squeed I don't really understand this. We know who belongs in the cluster, because those very addresses are passed into ovn-kubernetes:
> servers: 23b1("tcp:[172.18.0.6]:6643"), 5c5d("tcp:[172.18.0.5]:6643"), f5e7("tcp:[172.18.0.3]:6643")
> 
> So whoever is in this list that was not passed to ovn-kubernetes can be kicked. Are you suggesting the problem scenario is when the same address is used again, but a different ID? I don't think that is possible, because I think the entry will be replaced in this list with the new ID for the same address.

Looking at the bz, the issue was that the id was different but (I guess, since the list is still of 3 elemets) with the same address: https://bugzilla.redhat.com/show_bug.cgi?id=1837953#c23 ,that's why (I think) the current ids are important as well.

Another point is, this snapshot / serverid discrepancy _could_ have been fixed by https://patchwork.ozlabs.org/project/openvswitch/patch/1596641273-5748-1-git-send-email-dceara@redhat.com/

squeed:
> because I think the entry will be replaced in this list with the new ID for the same address.

@trozet okay, good. That was previously not the case IIRC. I seem to recall an issue where the db was lost, a new ID was generated, and then we had 4 participants but same address:port. Maybe I'm mis-remembering.

alexanderConstantinescu:
> I'm thinking this can be handled in ovnkube-master nodes instead of in CNO:
> 
> 1. Kicking old raft member: ovnkube-master gets the addresses from arguments passed to it. ovnkube-master will restart via CNO if nodes are removed with the new addresses. On ovnkube-master start simply check if there are any unknown addresses and kick them.
> 2. Ensuring DB: ovnkube-master can run a goroutine to check for db corruption, and if corrupted wipe the db. We just need to mount the db file location to the container. We can also make this more intelligent, so that if somehow all dbs got corrupted and they came back in a clean state (we now are out of sync with k8s api), we can now force a sync by restarting the ovnkube-master leader.

I am not that convinced that this should all be done in upstream ovn-kubernetes, for: ovnkube-master would still be dependent on having an operator, such as the CNO, to provide it with this list of IPs...which is not the case for upstream. 

For 2. I don't see how ovnkube-master would delete the DB on any other node it's not running on. It could theoretically tag the API server object with a annotation, flagging this fact to ovnkube-node and then have ovnkube-node read it in a sync loop and delete the DB if that is the case....but, that seems contrived, for this very particular OCP specific use case (actually: VZ using OCP specific use case).

fedepaol:
Closing as the fix is happening in ovn-k8s

PR #748, opened 2020-08-05
Bug 1857387: Do not reset the election-timer value

vishnoianil:
If user manually configure the election-timer, restarting the pods
can reset the value to default value (5 second). This can be an issue
if pods are restarted at large scale -- resetting to 5 second might
disrupt the cluster at that scale and it might not converge.


vishnoianil:
@dcbw minor patch. PTAL, thanks.

PR #751, opened 2020-08-07
Bug 1870711: Move metrics to SDN DS

juanluisvaladas:
The SDN metrics pod needs a service serving certificate which depends on
the SDN. However during the upgrade we've seen that sometimes during the
upgrade tis causes problems because the metrics daemonset is deployed
before rolling out the old daemonset.

### DIFF
--- a/bindata/network/openshift-sdn/monitor.yaml
+++ b/bindata/network/openshift-sdn/monitor.yaml
@@ -3,7 +3,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    app: sdn-metrics
+    app: sdn
   annotations:
     networkoperator.openshift.io/ignore-errors: ""
   name: monitor-sdn
@@ -16,14 +16,14 @@ spec:
     scheme: https
     tlsConfig:
       caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
-      serverName: sdn-metrics.openshift-sdn.svc
+      serverName: sdn.openshift-sdn.svc
   jobLabel: app
   namespaceSelector:
     matchNames:
     - openshift-sdn
   selector:
     matchLabels:
-      app: sdn-metrics
+      app: sdn
 ---
 apiVersion: v1
 kind: Service
@@ -31,12 +31,12 @@ metadata:
   annotations:
     service.beta.openshift.io/serving-cert-secret-name: sdn-metrics-certs
   labels:
-    app: sdn-metrics
-  name: sdn-metrics
+    app: sdn
+  name: sdn
   namespace: openshift-sdn
 spec:
   selector:
-    app: sdn-metrics
+    app: sdn
   clusterIP: None
   publishNotReadyAddresses: true
   ports:
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -174,6 +174,65 @@ spec:
             command: ["test", "-f", "/etc/cni/net.d/80-openshift-network.conf"]
           initialDelaySeconds: 5
           periodSeconds: 5
+      - name: kube-rbac-proxy
+        image: {{.KubeRBACProxyImage}}
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
+          TLS_PK=/etc/pki/tls/metrics-certs/tls.key
+          TLS_CERT=/etc/pki/tls/metrics-certs/tls.crt
+
+          # As the secret mount is optional we must wait for the files to be present.
+          # The service is created in monitor.yaml and this is created in sdn.yaml.
+          # If it isn't created there is probably an issue so we want to crashloop.
+          TS=$(curl \
+            --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+            -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
+              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
+          )
+
+          TS=$(date -d "${TS}" +%s)
+          WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
+          HAS_LOGGED_INFO=0
+          
+          log_missing_certs(){
+              CUR_TS=$(date +%s)
+              if [[ "${CUR_TS}" -gt "WARN_TS"  ]]; then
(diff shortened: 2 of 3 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -174,6 +174,42 @@ spec:
...
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
+          TLS_PK=/etc/pki/tls/metrics-certs/tls.key
+          TLS_CERT=/etc/pki/tls/metrics-certs/tls.crt
+
+          # As the secret mount is optional we must wait for the files to be present.
+          while [[ ! -f ${TLS_PK} ||  ! -f ${TLS_CERT} ]] ; do
+            echo $(date -Iseconds) WARN: Waiting for sdn-metrcs-certs to be created and mounted. This is expected during the installation.
```
dcbw, line 190: @juanluisvaladas two things; how about INFO instead of WARN if this is expected? Second, should we ever exit the loop, or just do we just wait for something else to kill the pod?

juanluisvaladas: > Or, perhaps INFO for the first 5 minutes, then WARN? Especially if the secrets directory is optional and might not be mounted the first time the container is run.

I like this, but the container creation time isn't really correlated with the certificate creation time.

I'm not quite sure how long it actually takes to create, but the whole instalation takes 30 so how about this:
1- Check if the files are there.
2- If they aren't report it once,
3- Check periodically and don't report anything at all until service creation timestamp + 25 minutes of grace period (if the whole installation takes 30, 25 seems very generous to me). After 25 minutes report it as an error rather than a warning.
4- After a longer period, maybe an hour or something in that line maybe make it a fatal so that there is a crashloop back off and it triggers alarms and the CNO to report issues

juanluisvaladas: Didn't want to wait any further because this may be having an actual impact so I implemented this grace period anyway. Let me know if you don't like it @dcbw

bindata/network/openshift-sdn/monitor.yaml
```diff
@@ -16,27 +16,27 @@ spec:
...
-      serverName: sdn-metrics.openshift-sdn.svc
+      serverName: sdn.openshift-sdn.svc
   jobLabel: app
   namespaceSelector:
     matchNames:
     - openshift-sdn
   selector:
     matchLabels:
-      app: sdn-metrics
+      app: sdn-
```
squeed, line 26: extra hyphen?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -174,6 +174,70 @@ spec:
...
+          TLS_CERT=/etc/pki/tls/metrics-certs/tls.crt
+
+          # As the secret mount is optional we must wait for the files to be present.
+          # The service is created in monitor.yaml and this is created in sdn.yaml.
+          # If it isn't created there is probably an issue so we want to crashloop.
+          TS=$(curl \
+            --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+            -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn-metrics" |
+              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
```
squeed, line 195: Can you just use `jq` here?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -174,6 +174,70 @@ spec:
...
+                HAS_LOGGED_INFO=1
+              fi
+          }
+
+          while [[ ! -f "${TLS_PK}" ||  ! -f "${TLS_CERT}" ]] ; do
+            log_misisng_certs
+            sleep 5
+          done
+          
+          /usr/bin/kube-rbac-proxy \
```
squeed, line 222: should do an `exec` here so we know when the process dies.

bindata/network/openshift-sdn/monitor.yaml
abhat, line 26: `app: sdn` vs. `app: sdn-`?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -174,6 +174,70 @@ spec:
...
+            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn-metrics" |
+              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
+          )
+
+          TS=$(date -d "${TS}" +%s)
+          WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
+          FATAL_TS=$(( ${TS} + $(( 60 * 60)) ))
+          HAS_LOGGED_INFO=0
+          
+          log_misisng_certs(){
```
abhat, line 203: nit: log_missing_certs <- log_misisng_certs

abhat:
overall minor comments including 👍 to those from @squeed. 

We need similar changes for ovn-k.

squeed:
I think this is basically fine, though it would ultimately be better for us to expose the missing service-ca certificate in the operator directly. That way we can expose a more meaningful message to the end-user.

bindata/network/openshift-sdn/sdn.yaml
juanluisvaladas: No, the image doesn't have jq, kubectl or oc.

The image ships the urlgrabber python library so I probably could replace bash with a python script doing something like:
```
command:
  - python
  - -c
  - |
    <SNIP>
```

Not quite sure it's worth the effort now.

juanluisvaladas:
/retitle Bug 1870711: Move metrics to SDN DS

juanluisvaladas:
/retest
@squeed @abhat can you guys please have another look? I'm in a bit of a hurry because I want to do the same for OVN.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -174,6 +174,70 @@ spec:
...
+            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
+              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
+          )
+
+          TS=$(date -d "${TS}" +%s)
+          WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
+          FATAL_TS=$(( ${TS} + $(( 60 * 60)) ))
+          HAS_LOGGED_INFO=0
+          
+          log_missing_certs(){
```
squeed, line 203: A thought - what if we just don't do this? Since metrics will never start, we could alert on this if we liked.

juanluisvaladas: Removed the fatal because the lack of metrics should already trigger an alert.

PR #752, opened 2020-08-11
Bug 1866200: fix recording rules group name

s-urbaniak:
The `general.rules` group name conflicts with others, causing them to be merged with other rules in OpenShift console via Thanos Querier.

This fixes it.

/cc @openshift/openshift-team-monitoring

### DIFF
--- a/bindata/network/kuryr/alert-rules.yaml
+++ b/bindata/network/kuryr/alert-rules.yaml
@@ -10,7 +10,7 @@ metadata:
   namespace: openshift-kuryr
 spec:
   groups:
-  - name: general.rules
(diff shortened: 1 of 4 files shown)


squeed:
@s-urbaniak does this need a backport?

s-urbaniak:
@squeed thank you for approving so fast :raised_hands: no it does not need a backport as it really is an issue in thanos querier rules API, not being active in <4.6. Also we are planning to fix it in Thanos itself where consensus was reached as we speak in [1]. However this fix here is still nice as it a) solves the issue short-term and b) gives you a good semantical name for the rules you are deploying.

[1] https://github.com/thanos-io/thanos/issues/3017

PR #753, opened 2020-08-12
build: require golang 1.14

danwinship:
We noticed in https://github.com/openshift/release/pull/10862 that `Dockerfile` and `Dockerfile.rhel7` used different versions of golang. This updates everything to use 1.14.

Changing `go.mod` to specify 1.14 required re-running `go mod vendor` to update `vendor/modules.txt` to include the `## explicit` lines

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.14 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-network-operator
 COPY . .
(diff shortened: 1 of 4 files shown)


aojea:
/retest
2 failed tests in gcp OVN
```
[sig-storage] In-tree Volumes [Driver: azure-disk] [Testpattern: Pre-provisioned PV (ext4)] volumes should allow exec of files on the volume [Suite:openshift/conformance/parallel] [Suite:k8s]
[sig-storage] In-tree Volumes [Driver: gluster] [Testpattern: Dynamic PV (default fs)] volume-expand should not allow expansion of pvcs without AllowVolumeExpansion property [Skipped:ibmcloud] [Suite:openshift/conformance/parallel] [Suite:k8s]
```

PR #754, opened 2020-08-12
Bug 1868078: Fixes providing empty slice for HybridClusterNetwork

trozet:
In order to deploy without using HybridClusterNetwork with
OVN-Kubernetes, the user is required to specify:

    ovnKubernetesConfig:
      hybridOverlayConfig:
        hybridClusterNetwork: []

The code was assuming the slice always had values.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -91,7 +91,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVN_service_cidr"] = svcpools
 
 	if c.HybridOverlayConfig != nil {
-		if c.HybridOverlayConfig.HybridClusterNetwork != nil {
+		if len(c.HybridOverlayConfig.HybridClusterNetwork) > 0 {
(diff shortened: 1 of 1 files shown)


trozet:
@danwinship @rcarrillocruz

pkg/network/ovn_kubernetes.go
```diff
@@ -91,7 +91,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVN_service_cidr"] = svcpools
 
 	if c.HybridOverlayConfig != nil {
-		if c.HybridOverlayConfig.HybridClusterNetwork != nil {
+		if c.HybridOverlayConfig.HybridClusterNetwork != nil && len(c.HybridOverlayConfig.HybridClusterNetwork) > 0 {
```
danwinship, line 94: ok but if you feel like updating it you can remove the `nil` check, since `len(nil) == 0` so it's redundant

trozet: good catch. I just updated it.

trozet:
@danwinship @rcarrillocruz I had to rebase to pick up #753

Can you please lgtm?

danwinship:
/lgtm

> I had to rebase to pick up #753

I'm pretty sure that shouldn't be needed... CI locally merges the PR to master before testing; that's why it forces you to rebase if there are merge conflicts.

PR #755, opened 2020-08-13
Bug 1874385: OVN: Handle OVN db upgrades.

numansiddique:
The ovnkube master daemonset starts the ovsdb-servers using ovn-ctl commands - run_nb_ovsdb
and run_sb_ovsdb. These versions of the command don't daemonize the ovsdb-servers and hence
the code to handle the DB upragdes in these functions never gets run.

This results in the OVN dbs in the old schema. Upgraded ovn-controllers will not see
any new columns because of this resulting in datapath disruptions.

This patch adds the code to handle the db upgrade in the 'postStart' stage.

Bug 1868392: [FDP 20.F] OVN 2.13 breaks pod-pod networking across the projects on OCP

Signed-off-by: Numan Siddique <nusiddiq@redhat.com>

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -166,6 +166,22 @@ spec:
                   fi
                   sleep 2
                   done
+
+                  # Upgrade the db if required.
+                  DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
+                  DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                  schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                  db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                  target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                    :
+                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
+                  else
+                      echo "Upgrading database $schema_name from schema version $db_version to $target_version"
+                      ovsdb-client -t 30 convert "$DB_SERVER" "$DB_SCHEMA"
+                  fi
                 fi
                 #configure northd_probe_interval
                 OVN_NB_CTL="ovn-nbctl -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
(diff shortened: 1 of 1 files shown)


dcbw:
@numansiddique should the upgrade happen before the DB is started? You've put the block into the "postStart" lifecycle hook so the ovsdb-server process will already be running with the DB by the time the upgrade happens.

numansiddique:
Note: I tested this PR using ovn-kind-cno.sh locally and it works as expected.

squeed:
question - should we handle this on the master? what happens if it runs twice on a cluster?

numansiddique:
> twice

We could. run on leader too. Probably its better to run it on leader.
Are you fine moving the code to run on the leader.

When it is run twice, the second time it will be no-op because the schema version and the db version will be the same.

squeed:
If upgrading is idempotent, then it's probably fine to just run in all nodes.

This is handled by one of the ovn support scripts, right? Is there any way we can use that and still keep the neat process hierarchy? Or is that just too much bother?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -166,6 +166,22 @@ spec:
...
+                  db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                  target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                    :
+                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
+                  else
+                      echo "Upgrading database $schema_name from schema version $db_version to $target_version"
+                      ovsdb-client -t 30 convert "$DB_SERVER" "$DB_SCHEMA"
```
abhat, line 183: Will the timeout be sufficient for large db sizes? Or is the conversion orthogonal to the size?
Should we check for the return value of `convert` to make sure the upgrade was successful? What is our recourse if the conversion fails or causes data loss?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -371,6 +387,22 @@ spec:
...
```
abhat, line 404: Same as Line 183.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
numansiddique: I copied much of the code from here - https://github.com/openvswitch/ovs/blob/master/utilities/ovs-lib.in#L461

ovn-ctl scripts call upgrade_cluster() which does the samething.

But since the  daemonset here calls, run_xx_ovsdb, it  never calls upgrade_cluster().

I'm not sure about the what should be proper timeout value. I'm fine if we want to increase the timeout value to may be 60 seconds ? or higher.

If convert fails, we can probably attempt few times before we give up.
I'm not sure what to do in case uprade fails or there seems to be data loss. 

@dceara  Do you have any comments here ?

numansiddique:
> If upgrading is idempotent, then it's probably fine to just run in all nodes.
> 
> This is handled by one of the ovn support scripts, right? Is there any way we can use that and still keep the neat process hierarchy? Or is that just too much bother?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
numansiddique: I discussed with @dceara.

We both are of the opinion that - if the upgrade fails, the only thing we can do is delete all DBs and start over.

numansiddique:
> If upgrading is idempotent, then it's probably fine to just run in all nodes.
> 
> This is handled by one of the ovn support scripts, right? Is there any way we can use that and still keep the neat process hierarchy? Or is that just too much bother?

Yeah. But since the daemonset is calling - run_xx_ovsdb (instead of start_xx_ovsdb), the uprade_cluster() is never hit.

I've  a diry way to do it - https://github.com/numansiddique/ovn/commit/9f51a2df6665e0de17271c8df4fd67fe64675207

But I'm really not sure if that's the right way. This patch basically starts ovsdb-server in backgroud, it then calls the upgrade_cluster() code and then finally just waits for ovsdb-server to complete.

alexanderConstantinescu:
CI is....well, not very happy about anything, let's wait until at least tomorrow. 

/hold

alexanderConstantinescu:
I guess this is doomed to fail, but here goes anyway:

/test e2e-gcp-ovn

knobunc:
Forced this merge because we have a tangle of broken CI and we need to get this in to unblock 4.5.

PR #756, opened 2020-08-14
Bug 1868259: Fixes OVN upgrade with shared gateway mode

trozet:
With OCP upgrade today, CNO happens before MCO. When upgrading from 4.5
to 4.6, 4.6 OVN expects system OVS to be there and configured which will
not happen until MCO upgrade finishes. Therefore upon initial CNO
upgrade, we will run in local gateway mode until the node is rebooted
via MCO with a machine-os-content change to include system OVS and run
the ovs-configuration service.

We also need to temporarily use local gateway when converting from
openshift-sdn to ovn-kubernetes, so use the enabled system service path
for checking.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -495,6 +495,15 @@ spec:
             fi
           fi
 
+          gateway_mode_flags=
+          # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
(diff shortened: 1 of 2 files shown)


trozet:
@squeed @danwinship PTAL

Ideally I would like this to wait until https://github.com/openshift/release/pull/10582 is fixed, or perhaps get some manual verification that this works.

/hold

danwinship:
ah, makes sense
/lgtm

trozet:
@danwinship had to rebase to pick up #753. Do you mind adding lgtm again?

PR #760, opened 2020-08-19
[release-4.4] Bug 1858217: bump golang.org/x/text/encoding

aojea:
CVE-2020-14040 golang.org/x/text: possibility to trigger an
infinite loop in encoding/unicode could lead to crash.

Upgrade to golang.org/x/text/encoding to v0.3.3


aojea:
@juanluisvaladas PTAL

PR #761, opened 2020-08-20
Simplify CA Bundle injection for multus admission webhook

alexanderConstantinescu:
Preamble:

Coming soon will be a validating admission webhook for ovn-kubernetes. Leading to that: I investigated the different mechanisms used in OpenShift and in our CNO to generate certs/ca-bundles, realizing that nothing generates these in the same way. Multus has one way of doing it, Kuryr another, OVN another...deciding to focus on multus I however realized that multus was severely over-complicating things. 

The mechanism to inject the CA bundle into its webhook was:

1. Use `multus-admission-controller/004-configmap.yaml` to generate a configmap (`openshift-network-operator/openshift-service-ca`) with the annotation `service.beta.openshift.io/inject-cabundle: "true"` which is reconciled by the ServiceCA to inject the proper CA bundle. 
2. Reconcile that reconciled config map ourselves to read the injected value, and inject that value into the webhook. 

In fact, the ServiceCA reconciles any webhook with the annotation `service.beta.openshift.io/inject-cabundle: "true"` and injects the CA-bundle into the webhook directly. 

This patch thus removes one operator in the CNO, one config map and consolidates the webhook generation according to a more "standard" procedure for OpenShift. 

/assign @dougbtv


juanluisvaladas:
I've been checking this and I don't see where the openshift-service-ca is used at all, I can't find a single reference to it in the multus admission controller. Do we still need this configmap to be created? I think we may be able to also remove the annotation. 

@s1061123 can you PTAL?

juanluisvaladas:
Ah, you're right, I was refering to the annotation service.beta.openshift.io/inject-cabundle but this is in mutation webhook now, I didn't know this could be used outside a configmap.
/lgtm

danwinship:
I don't think we have any e2e of the multus admission webhook (grumble grumble) so this will need to be manually tested.

I don't really know any of this code so I'm not sure, but it looks like a nice simplification if it does indeed work.

alexanderConstantinescu:
> I don't think we have any e2e of the multus admission webhook (grumble grumble) so this will need to be manually tested.
> 
> I don't really know any of this code so I'm not sure, but it looks like a nice simplification if it does indeed work.

I tested that the admission control webhook got the proper CA bundle injected by the ServiceCA, earlier today. So I am sure that works. 

I will test that it properly works and validates any `network-attachment-definitions` next week when I am back. Unless @s1061123 / @dougbtv knows if such objects are created during a CI run? Which would sort that out already, given that some CI jobs have already passed.

alexanderConstantinescu:
FYI, using https://github.com/openshift/multus-admission-controller#example-of-the-admission-controller-in-action I just validated that it works:

```
aconstan@linux ~ $ cat tmp
apiVersion: "k8s.cni.cncf.io/v1"
metadata:
  name: bad-conf
spec:
  config: '{
      "cniVersion": "0.3.0"malFormattedJSON
      }
    }'
aconstan@linux ~ $ oc create -f tmp
Error from server: error when creating "tmp": admission webhook "multus-validating-config.k8s.io" denied the request: configuration string is not in JSON format
aconstan@linux ~ $ vi tmp
aconstan@linux ~ $ cat tmp
apiVersion: "k8s.cni.cncf.io/v1"
metadata:
  name: macvlan-conf
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "macvlan",
      "master": "eth0",
      "mode": "bridge",
      "ipam": {
        "type": "host-local",
        "subnet": "192.168.1.0/24",
        "rangeStart": "192.168.1.200",
        "rangeEnd": "192.168.1.216",
        "routes": [
          { "dst": "0.0.0.0/0" }
        ],
        "gateway": "192.168.1.1"
      }
    }'
aconstan@linux ~ $ oc create -f tmp
networkattachmentdefinition.k8s.cni.cncf.io/macvlan-conf created
aconstan@linux ~ $ oc get clusterversion
NAME      VERSION                                           AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.6.0-0.ci.test-2020-08-26-065839-ci-ln-1yvm47t   True        False         17m     Cluster version is 4.6.0-0.ci.test-2020-08-26-065839-ci-ln-1yvm47t

```

The message `configuration string is not in JSON format` comes from here: https://github.com/openshift/multus-admission-controller/blob/8c0ba31c0beebcd72d9d57513c0b47a2b6a545cf/pkg/webhook/webhook.go#L124

danwinship:
/lgtm cancel
/approve
@dougbtv can you /lgtm this if it makes sense to you?

alexanderConstantinescu:
@dougbtv, could you please take a look at this PR and let us know if what you think?

alexanderConstantinescu:
@dougbtv : PTAL

PR #763, opened 2020-08-25
Bug 1937594: Split SDN migration into 2 phase

pliurh:
Phase1, Users add the spec.migration feild in network.operator
CR. The only changes the migration field under the status of
Network.config. It will trigger MCO to provision ovs-configuration
service on every node and reboot. After rebooting, openshift-sdn
will use br-ex as L3 gateway.

Phase2, Users update the spec.networkType in network.config.
CNO starts to swap the network provider for the cluster to ovnkube.
Users need to manually reboot all nodes to make all pods attach to
the new cluster network.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -175,17 +175,14 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	// Compare against previous applied configuration to see if this change
 	// is safe.
 	if prev != nil {
-		// Check if the operator is put in the 'Network Migration' mode.
-		if _, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok {
-			// We may need to fill defaults here -- sort of as a poor-man's
-			// upconversion scheme -- if we add additional fields to the config.
-			err = network.IsChangeSafe(prev, &operConfig.Spec)
-			if err != nil {
-				log.Printf("Not applying unsafe change: %v", err)
-				r.status.SetDegraded(statusmanager.OperatorConfig, "InvalidOperatorConfig",
-					fmt.Sprintf("Not applying unsafe configuration change: %v. Use 'oc edit network.operator.openshift.io cluster' to undo the change.", err))
-				return reconcile.Result{}, err
-			}
+		// We may need to fill defaults here -- sort of as a poor-man's
+		// upconversion scheme -- if we add additional fields to the config.
+		err = network.IsChangeSafe(prev, &operConfig.Spec)
+		if err != nil {
+			log.Printf("Not applying unsafe change: %v", err)
+			r.status.SetDegraded(statusmanager.OperatorConfig, "InvalidOperatorConfig",
+				fmt.Sprintf("Not applying unsafe configuration change: %v. Use 'oc edit network.operator.openshift.io cluster' to undo the change.", err))
+			return reconcile.Result{}, err
 		}
 	}
 
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -164,5 +164,11 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec, oldStatus *configv1.
 		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.KuryrConfig.MTU)
 	}
 
+	// Set migration in the config status
+	if operConf.Migration != nil {
+		status.Migration = &configv1.NetworkMigration{NetworkType: string(operConf.Migration.NetworkType)}
+	} else {
+		status.Migration = nil
+	}
 	return &status
 }
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -232,6 +232,9 @@ func IsChangeSafe(prev, next *operv1.NetworkSpec) error {
 		errs = append(errs, err)
 	}
 
+	// Check the network migration
+	errs = append(errs, isMigrationChangeSafe(prev, next)...)
+
 	// Check the default network
 	errs = append(errs, isDefaultNetworkChangeSafe(prev, next)...)
 
@@ -456,20 +459,37 @@ func fillDefaultNetworkDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 }
 
 func isDefaultNetworkChangeSafe(prev, next *operv1.NetworkSpec) []error {
+
 	if prev.DefaultNetwork.Type != next.DefaultNetwork.Type {
-		return []error{errors.Errorf("cannot change default network type")}
+		if prev.Migration == nil {
+			return []error{errors.Errorf("cannot change default network type when not doing migration")}
+		} else {
+			if prev.Migration.NetworkType != next.DefaultNetwork.Type {
+				return []error{errors.Errorf("can only change default network type to the target migration network type")}
+			}
+		}
 	}
 
-	switch prev.DefaultNetwork.Type {
-	case operv1.NetworkTypeOpenShiftSDN:
-		return isOpenShiftSDNChangeSafe(prev, next)
-	case operv1.NetworkTypeOVNKubernetes:
-		return isOVNKubernetesChangeSafe(prev, next)
-	case operv1.NetworkTypeKuryr:
-		return isKuryrChangeSafe(prev, next)
-	default:
-		return nil
+	if prev.Migration == nil {
+		switch prev.DefaultNetwork.Type {
+		case operv1.NetworkTypeOpenShiftSDN:
+			return isOpenShiftSDNChangeSafe(prev, next)
+		case operv1.NetworkTypeOVNKubernetes:
+			return isOVNKubernetesChangeSafe(prev, next)
+		case operv1.NetworkTypeKuryr:
+			return isKuryrChangeSafe(prev, next)
+		default:
+			return nil
+		}
 	}
+	return nil
+}
+
+func isMigrationChangeSafe(prev, next *operv1.NetworkSpec) []error {
+	if prev.Migration != nil && next.Migration != nil && prev.Migration.NetworkType != next.Migration.NetworkType {
+		return []error{errors.Errorf("cannot change migration network type after migration is start")}
+	}
+	return nil
 }
 
 // ValidateAdditionalNetworks validates additional networks configs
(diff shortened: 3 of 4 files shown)


pliurh:
@trozet @abhat @squeed PTAL.

pliurh:
This PR depends on the https://github.com/openshift/machine-config-operator/pull/2015

pliurh:
/unhold
Since there will be no intermediary state (aka old ovn local gateway mode) available anymore. We have to take this approach.

pliurh:
@dcbw @trozet @abhat  PTAL

trozet:
/hold
I think you may not want to split into 2 phases after these go in:
https://github.com/openshift/cluster-network-operator/pull/801
https://github.com/openshift/ovn-kubernetes/pull/281

pliurh:
> /hold
> I think you may not want to split into 2 phases after these go in:
> #801
> [openshift/ovn-kubernetes#281](https://github.com/openshift/ovn-kubernetes/pull/281)

In 4.6, maybe yes. In long term, if the ovn-kube local gw mode will exist until openshift-sdn fades away in OCP, we can always relay on it as the intermediate stage during migration. But, if that is not the case, we will still need this 2-steps approach when the local gw mode is deprecated by ovn-kube.

trozet:
OK I'll leave it to your judgement. I just wanted to make you aware of the other changes before this merged so you could make the call.

/hold cancel

pliurh:
/hold

I'll test the local gw approach first with @trozet 's PRs. If it works, we will go that way in 4.6. And this PR can wait until 4.7.

pliurh:
/unhold
@trozet After testing with https://github.com/ovn-org/ovn-kubernetes/pull/2055, I found the gateway_mode_flags `--gateway-mode local --gateway-interface none` doesn't work anymore due to https://github.com/openshift/ovn-kubernetes/commit/1ba1ce885f089a023b4c803d85a2ffc7206eb98b. It breaks the local gateway mode.
```
I0312 08:00:41.404433   14571 gateway_init.go:162] Initializing Gateway Functionality
I0312 08:00:41.404649   14571 gateway_localnet.go:184] Node local addresses initialized to: map[10.131.0.2:{10.131.0.0 fffffe00} 127.0.0.1:{127.0.0.0 ff000000} 172.22.0.99:{172.22.0.0 ffffff00} 192.168.111.23:{192.168.111.0 ffffff00} 192.168.111.4:{192.168.111.4 ffffffff} ::1:{::1 ffffffffffffffffffffffffffffffff} fe80::899a:2541:edca:cbe8:{fe80:: ffffffffffffffff0000000000000000} fe80::d49c:f8ff:fe3a:98c8:{fe80:: ffffffffffffffff0000000000000000} fe80::dacf:b57b:2ab4:6b96:{fe80:: ffffffffffffffff0000000000000000} fe80::e07b:72ff:fe1b:c0a5:{fe80:: ffffffffffffffff0000000000000000}]
I0312 08:00:41.404820   14571 helper_linux.go:73] Found default gateway interface enp2s0 192.168.111.1
F0312 08:00:41.404860   14571 ovnkube.go:130] could not find IP addresses: failed to lookup link none: Link not found
```

Instead of spending time to fix the local gw mode in ovn-kube, I think it's better to refactor the migration with this PR, so that we don't need to rely on the ovn-kube intermediate stage anymore.

pliurh:
New CI test for this change https://github.com/openshift/release/pull/16861

alexanderConstantinescu:
/retitle Bug 1937594: Split SDN migration into 2 phase

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -176,7 +176,30 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	// is safe.
 	if prev != nil {
 		// Check if the operator is put in the 'Network Migration' mode.
-		if _, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok {
+		if val, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok || val != "start" {
```
trozet, line 179: How does this work? If the annotation doesn't exist how do you know it is in network migration mode? To me it would seem the opposite would be true. Should this be

```
if val, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; ok && val != "start" {
```

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -176,7 +176,30 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+			// Prepare migration, set status but not apply objects during preparation phase.
+			if val == "prepare" {
```
trozet, line 181: what if val is not prepare? should we log some error saying invalid value for migration?

pkg/controller/operconfig/operconfig_controller.go
pliurh: The condition here is to check if CNO shall bypass `IsChangeSafe`. So if the annotation doesn't exist or it exists but its value is not 'start', both mean CNO shall not bypass `IsChangeSafe` check and start the migration.

If the annotation exists and its value is 'prepare', the CNO comes to the migration preparing phase.

pkg/controller/operconfig/operconfig_controller.go
pliurh: fixed.

pkg/controller/operconfig/operconfig_controller.go
trozet: OK I don't think I understand CNO code well enough to review this. @abhat do you mind taking a look?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -176,7 +176,36 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
 	if prev != nil {
 		// Check if the operator is put in the 'Network Migration' mode.
-		if _, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok {
+		val, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]
+
+		if ok && val != "start" && val != "prepare" {
+			log.Printf("invalid annotation value for migration: %s", val)
+		}
+
+		if !ok || val != "start" {
```
abhat, line 185: if `!ok`, why do we want to continue checking for `"prepare"`?

abhat: is it better to just say `if ok && val == "prepare" {`

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -176,7 +176,36 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+			// Prepare migration, set status but not apply objects during preparation phase.
+			if val == "prepare" {
+				// Update Network.config.openshift.io.Status
+				status, err := r.ClusterNetworkStatus(context.TODO(), operConfig)
+				if err != nil {
+					log.Printf("Could not generate network status: %v", err)
```
abhat, line 191: nit: log should be consistent with the status message we are setting?

pkg/controller/operconfig/operconfig_controller.go
pliurh: fixed.

pkg/controller/operconfig/operconfig_controller.go
pliurh: Actually, these lines are copied from https://github.com/openshift/cluster-network-operator/blob/eb90c4ed641f302f5c1b5d514257c7f0698de66a/pkg/controller/operconfig/operconfig_controller.go#L327

And I found the log and the status messages aren't consistent in other existing codes. I'm not sure whether it's a protocol we need to follow.

trozet:
@danwinship

pkg/network/cluster_config.go
```diff
@@ -137,6 +137,11 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec, oldStatus *configv1.
 		status.NetworkType = string(operConf.DefaultNetwork.Type)
 	}
 
+	// Set networkType to empty string when preparing migration. So networkType value in spec can be consumed by MCO.
+	if migrationPrepare {
+		status.NetworkType = ""
```
danwinship, line 142: We can't do this. This is a public API; arbitrary components may be relying on the value of this field, for arbitrary purposes, and they might break if it is suddenly unset.

We may need to add a field to the config indicating that migration is in progress, so MCO can respond to that. (Maybe MCO could just read the annotation? That seems kind of bad though. If it's an actual API between CNO and MCO, then it should be actual API, not an annotation.)

danwinship (changes requested):

pkg/network/cluster_config.go
pliurh: When the migration starts, the cluster will be in a maintenance state. The cluster shall be considered as out-of-service until the whole process finishes. We don't guarantee any of the cluster components can behave as normal during this period. So as long as the MCO can work as expected, I suppose it should be fine. WDYT?

pliurh:
Requires MCO change in https://github.com/openshift/machine-config-operator/pull/2518

pliurh:
@danwinship PTAL

go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/mitchellh/copystructure v1.0.0 // indirect
 	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/onsi/gomega v1.10.2
-	github.com/openshift/api v0.0.0-20210402143208-92e9dab578e8
+	github.com/openshift/api v0.0.0-20210408195222-460636dd2fea
```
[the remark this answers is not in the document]
pliurh: https://github.com/openshift/cluster-network-operator/pull/1057

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -175,8 +175,8 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 	// Compare against previous applied configuration to see if this change
 	// is safe.
 	if prev != nil {
-		// Check if the operator is put in the 'Network Migration' mode.
-		if _, ok := operConfig.GetAnnotations()[names.NetworkMigrationAnnotation]; !ok {
+		// Check if the operator is put in the 'Network Migration' mode, skip IsChangeSafe if yes.
+		if operConfig.Spec.Migration == nil || operConfig.Spec.Migration.NetworkType != operConfig.Spec.DefaultNetwork.Type {
 			// We may need to fill defaults here -- sort of as a poor-man's
 			// upconversion scheme -- if we add additional fields to the config.
 			err = network.IsChangeSafe(prev, &operConfig.Spec)
```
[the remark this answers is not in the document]
pliurh: fixed.

PR #766, opened 2020-08-27
Bug 1880490: change the hybrid overlay config to use the ConfigMap instead of cli

JacobTanenbaum:
move the enabling of hybrid overlay into the ConfigMap for ovnkube like
all new arguments should be

Signed-off-by: Jacob Tanenbaum <jtanenba@redhat.com>

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -16,6 +16,9 @@ data:
     service-cidrs="{{.OVN_service_cidr}}"
     ovn-config-namespace="openshift-ovn-kubernetes"
     apiserver="{{.K8S_APISERVER}}"
+{{- if .OVNHybridOverlayEnable }}
+    no-hostsubnet-nodes="kubernetes.io/os=windows"
+{{- end  }}
  
     [ovnkubernetesfeature]
     enable-egress-ip=true
@@ -23,3 +26,13 @@ data:
     [gateway]
     mode=local
     nodeport=true
+{{ if .OVNHybridOverlayEnable }}
+    [hybridoverlay]
+    enabled=true
+    {{- if .OVNHybridOverlayNetCIDR }}
+    cluster-subnets="{{.OVNHybridOverlayNetCIDR}}"
+    {{- end }}
+    {{- if .OVNHybridOverlayVXLANPort}}
+    hybrid-overlay-vxlan-port="{{.OVNHybridOverlayVXLANPort}}"
+    {{- end }}
+{{- end  }}
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 
@@ -13,8 +14,6 @@ import (
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// vars
-var g = uint32(8061)
 var OVNKubernetesConfig = operv1.Network{
 	Spec: operv1.NetworkSpec{
 		ServiceNetwork: []string{"172.30.0.0/16"},
@@ -31,7 +30,7 @@ var OVNKubernetesConfig = operv1.Network{
 		DefaultNetwork: operv1.DefaultNetworkDefinition{
 			Type: operv1.NetworkTypeOVNKubernetes,
 			OVNKubernetesConfig: &operv1.OVNKubernetesConfig{
-				GenevePort: &g,
+				GenevePort: ptrToUint32(8061),
 			},
 		},
 	},
@@ -70,6 +69,7 @@ func TestRenderOVNKubernetes(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "openshift-ovn-kubernetes-node")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-ovn-kubernetes", "ovnkube-master")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-ovn-kubernetes", "ovnkube-node")))
+	g.Expect(objs).To(ContainElement(HaveKubernetesID("ConfigMap", "openshift-ovn-kubernetes", "ovnkube-config")))
 
 	// make sure all deployments are in the master
 	for _, obj := range objs {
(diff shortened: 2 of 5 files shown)


bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -20,3 +20,12 @@ data:
     [gateway]
     mode=local
     nodeport=true
+
+{{ if .OVNHybridOverlayEnable }}
```
danwinship, line 24: `OVNHybridOverlayEnable` is currently added to the template data as a `string`, so I don't think this will work; but you can just change it in `pkg/network/ovn_kubernetes.go` to set the value to a boolean `true` rather than a string `"true"`. (And likewise `false` instead of `""` in the unset case.)

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -20,3 +20,12 @@ data:
...
+    [hybridoverlay]
+    enabled=true
+    cluster-subnets="{{.OVNHybridOverlayNetCIDR}}"
```
danwinship, line 27: To be consistent with the bash, you need `{{if .OVNHybridOverlayNetCIDR}}`...`{{end}}` around this. I think that case is used with the non-Windows overlay mode.

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -20,3 +20,12 @@ data:
...
+    {{ if .OVNHybridOverlayVXLANPort}}
+    hybrid-overlay-vxlan-port="{{.OVNHybridOverlayVXLANPort}}"
+    {{ end }}
+{{ end  }}
```
danwinship, line 31: also, use `{{- if ... }}` and `{{- end }}` everywhere. The `-` after the `{{` makes it discard extra whitespace so that you don't end up with extra blank lines in the generated file (which doesn't _really_ matter, but it looks nicer and it's easy...)

pkg/network/ovn_kubernetes_test.go
```diff
@@ -70,6 +70,7 @@ func TestRenderOVNKubernetes(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "openshift-ovn-kubernetes-node")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-ovn-kubernetes", "ovnkube-master")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-ovn-kubernetes", "ovnkube-node")))
+	g.Expect(objs).To(ContainElement(HaveKubernetesID("ConfigMap", "openshift-ovn-kubernetes", "ovnkube-config")))
```
danwinship, line 73: I think we need more than this... You should be able to extract the generated `ovnkube.conf` with:

```
func extractOVNKubeConfig(objs []*uns.Unstructured) string {
	for _, obj := range objs {
		if obj.GetKind() == "ConfigMap" && obj.GetName() == "ovnkube-config" {
			val, ok, err := uns.NestedString(obj.Object, "data", "ovnkube.conf")
			g.Expect(ok).To(BeTrue())
			return string(val)
		}
	}
}
```

So, create a new test function, starting with the top of `TestRenderOVNKubernetes`. After calling `renderOVNKubernetes`, do `conffile := extractOVNKubeConfig(objs)` to get the generated `ovnkube.conf` file as a string and then you can scan for expected / not-expected substrings in that. You can do it a few times, modifying `config` and re-running `renderOVNKubernetes` on it to check different results. (`TestProxyArgs` in `openshift_sdn_test.go` is the same idea, although in that case the config file is YAML so it uses `uns.Unstructured` to parse it rather than just treating it as a string.)

bindata/network/ovn-kubernetes/004-config.yaml
danwinship: Actually... now that I think about it, it _will_ work this way, because `"true"` is a non-empty string, so it counts as `true`, and `""` is an empty string, so it counts as `false`, so you get the right result. But like, if the code was setting `OVNHybridOverlayEnable` to `"false"` in the false case, then this would break. So it would be clearer to use real boolean variables there.

bindata/network/ovn-kubernetes/004-config.yaml
JacobTanenbaum: I am not sure you are right 
```
          hybrid_overlay_flags=
          if [[ -n "{{.OVNHybridOverlayEnable }}" ]]; then
            hybrid_overlay_flags="--enable-hybrid-overlay --no-hostsubnet-nodes=kubernetes.io/os=windows"
            if [[ -n "{{.OVNHybridOverlayNetCIDR}}" ]]; then
              hybrid_overlay_flags="${hybrid_overlay_flags} --hybrid-overlay-cluster-subnets={{.OVNHybridOverlayNetCIDR}}"
            fi
            if [[ -n "{{.OVNHybridOverlayVXLANPort}}" ]]; then
              hybrid_overlay_flags="${hybrid_overlay_flags} --hybrid-overlay-vxlan-port={{.OVNHybridOverlayVXLANPort}}"
            fi
          fi```

is  the original bash so in order to define a HybridOverlayNetCIDR OVNHybridOverlay has to be enabled

File: bindata/network/ovn-kubernetes/004-config.yaml (hunk 1 above)
Comment (JacobTanenbaum) [reply bindata/network/ovn-kubernetes/004-config.yaml:23]: done

File: pkg/network/ovn_kubernetes_test.go (hunk 4 above)
Comment (JacobTanenbaum) [reply pkg/network/ovn_kubernetes_test.go:74]: added some more test cases

File: bindata/network/ovn-kubernetes/004-config.yaml (hunk 3 above)
Comment (JacobTanenbaum) [reply bindata/network/ovn-kubernetes/004-config.yaml]: I did this except for the first one so that a new line would appear between the sections like we expect

File: pkg/network/ovn_kubernetes_test.go [hunk 5]
```diff
@@ -139,6 +141,66 @@ func TestRenderOVNKubernetesIPv6(t *testing.T) {
...
+		expected            string
+		hybridOverlayConfig *operv1.HybridOverlayConfig
+	}
+	testcases := []testcase{
+		{
+			desc: "default",
+			expected: "[default]\nmtu=\"1500\"\ncluster-subnets=\"10.128.0.0/15/23,10.0.0.0/14/24\"\n" +
+				"encap-port=\"8061\"\n\n[kubernetes]\nservice-cidrs=\"172.30.0.0/16\"\n" +
+				"ovn-config-namespace=\"openshift-ovn-kubernetes\"\napiserver=\"" + fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")) + "\"\n\n" +
+				"[gateway]\nmode=local\nnodeport=true",
```
danwinship, line 156: ick. If you use backticks around the whole thing rather than double-quotes, then you won't have to backslash the internal quotes. Also, you could use actual newlines rather than `\n` in the string to make it more readable, although sometimes that's weird and not worth it because you lose the indentation.

hm, also, I think I'd set `KUBERNETES_SERVICE_HOST` / `KUBERNETES_SERVICE_PORT` at the start of the test case rather than rely on whatever happened to be in the environment. (And then you don't need to `Getenv` and `Sprintf` here, you can just hardcode the values into the expected output.)

pkg/network/ovn_kubernetes_test.go
```diff
@@ -139,6 +141,66 @@ func TestRenderOVNKubernetesIPv6(t *testing.T) {
...
+			expected: "[default]\nmtu=\"1500\"\ncluster-subnets=\"10.128.0.0/15/23,10.0.0.0/14/24\"\n" +
+				"encap-port=\"8061\"\n\n[kubernetes]\nservice-cidrs=\"172.30.0.0/16\"\n" +
+				"ovn-config-namespace=\"openshift-ovn-kubernetes\"\napiserver=\"" + fmt.Sprintf("https://%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT")) + "\"\n\n" +
+				"[gateway]\nmode=local\nnodeport=true\n\n[hybridoverlay]\nenabled=true\ncluster-subnets=\"10.132.0.0/14\"",
+			hybridOverlayConfig: &operv1.HybridOverlayConfig{
+				HybridClusterNetwork: []operv1.ClusterNetworkEntry{
+					{CIDR: "10.132.0.0/14", HostPrefix: 23},
+				},
+			},
+		},
```
danwinship, line 169: and two more cases to test the vxlan and enabled-but-no-clusternetworks cases?

pkg/network/ovn_kubernetes_test.go
```diff
@@ -139,6 +141,66 @@ func TestRenderOVNKubernetesIPv6(t *testing.T) {
...
+	g := NewGomegaWithT(t)
+
+	for i, tc := range testcases {
+		t.Run(fmt.Sprintf("%d:%s", i, tc.desc), func(t *testing.T) {
+			OVNKubeConfig := OVNKubernetesConfig.DeepCopy()
+			if tc.hybridOverlayConfig != nil {
+				OVNKubeConfig.Spec.DefaultNetwork.OVNKubernetesConfig.HybridOverlayConfig = tc.hybridOverlayConfig
+			}
+			//set a few files so that they are not machine dependant
+			mtu := uint32(1500)
```
danwinship, line 180: "files"?

danwinship:
/retest
looks like some CI burp?

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: done

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: done

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: done

pkg/network/ovn_kubernetes_test.go
```diff
@@ -139,6 +141,141 @@ func TestRenderOVNKubernetesIPv6(t *testing.T) {
...
+	type testcase struct {
+		desc                string
+		expected            string
+		hybridOverlayConfig *operv1.HybridOverlayConfig
+	}
+	testcases := []testcase{
+		{
+			desc: "default",
+			expected: `[default]
+mtu="1500"
```
danwinship, line 154: so as a hack, you could add a newline immediately after the backtick, to get `[default]` lined up with everything else, and likewise in the other cases, and then just update the comparison to use `strings.TrimSpace(tc.expected)` to fix it up...

pkg/network/ovn_kubernetes_test.go
```diff
@@ -344,3 +481,19 @@ func TestOVNKubernetesIsSafe(t *testing.T) {
...
+			val, ok, err := uns.NestedString(obj.Object, "data", "ovnkube.conf")
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(ok).To(BeTrue())
+			return string(val)
+		}
+	}
+	return ""
+}
+
+func ptrToUINT32(x uint32) *uint32 {
```
danwinship, line 497: should be `ptrToUint32` (eg, like `strconv.ParseUint`)

pkg/network/ovn_kubernetes_test.go
```diff
@@ -139,6 +141,141 @@ func TestRenderOVNKubernetesIPv6(t *testing.T) {
...
+
+	for i, tc := range testcases {
+		t.Run(fmt.Sprintf("%d:%s", i, tc.desc), func(t *testing.T) {
+			OVNKubeConfig := OVNKubernetesConfig.DeepCopy()
+			if tc.hybridOverlayConfig != nil {
+				OVNKubeConfig.Spec.DefaultNetwork.OVNKubernetesConfig.HybridOverlayConfig = tc.hybridOverlayConfig
+			}
+			//set a few inputs so that the tests are not machine dependant
+			mtu := uint32(1500)
+			OVNKubeConfig.Spec.DefaultNetwork.OVNKubernetesConfig.MTU = &mtu
```
danwinship, line 256:
```suggestion
			OVNKubeConfig.Spec.DefaultNetwork.OVNKubernetesConfig.MTU = ptrToUint32(1500)
```

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: done

pkg/network/ovn_kubernetes_test.go
JacobTanenbaum: done

JacobTanenbaum:
@danwinship thanks for the quick review! I addressed your nits and changed all the setting of `uint32` pointers to `ptrToUint32()`

danwinship:
/lgtm
alas we can't really skip e2e-gcp-ovn on this one...

JacobTanenbaum:
gcp-ovn test failed with transport is closing error 
/retest

PR #769, opened 2020-08-31
Bug 1874242: Revert "Bug 1857387: Do not reset the election-timer value"

trozet:
Reverts openshift/cluster-network-operator#748

This PR stops election leader timers from being set to the default value of 5 seconds, because 5 is never less than 1.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -218,7 +218,7 @@ spec:
                   fi
                 done
 
-                if [[ ${election_timer} -lt ${current_election_timer} ]]; then
+                if [[ ${election_timer} -ne ${current_election_timer} ]]; then
                   retries=0
                   while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound 2>/dev/null \
                     | grep "Role: candidate" ); do
@@ -390,7 +390,7 @@ spec:
                   fi
                 done
 
-                if [[ ${election_timer} -lt ${current_election_timer} ]]; then
+                if [[ ${election_timer} -ne ${current_election_timer} ]]; then
                   retries=0
                   while is_candidate=$(ovs-appctl -t /var/run/ovn/ovnsb_db.ctl cluster/status OVN_Southbound 2>/dev/null \
                     | grep "Role: candidate" ); do


trozet:
@vishnoianil FYI

We shouldn't be allowing users to change settings underneath us and try to guess what their intention was. If they need to change the setting it should be done via CNO.

dcbw:
@trozet @vishnoianil and rather than some knob in the CNO, if the value depends on cluster size, the CNO is fully capable of picking a number based on # nodes / # pods / whatever and deploying it to the cluster automatically, IMO

vishnoianil:
> @vishnoianil FYI
> 
> We shouldn't be allowing users to change settings underneath us and try to guess what their intention was. If they need to change the setting it should be done via CNO.

CNO is not guessing here, it's just trying to not override if user manually configure the value. OVN itself is a one big subsystem underneath and exposes it's own API that user can use to change config. If we really want user to not change the values i think we should not allow permission to user to use those utility tools. Unfortunately, CNO is not very flexible when it comes to changing the values, for that you need to stop CVO & it will restart all the pods, which you don't want to do in production environment. 

The option @dcbw mentioned below seems a bit better option, but until and unless cluster deny the new node joining if it passes that #node, you are always at the risk that user can add more nodes. If user add more nodes and raft start partitioning more frequently, what is our options ? Do we ask user to disable CVO and change the value through CNO?

vishnoianil:
> @trozet @vishnoianil and rather than some knob in the CNO, if the value depends on cluster size, the CNO is fully capable of picking a number based on # nodes / # pods / whatever and deploying it to the cluster automatically, IMO

@dcbw Please see my above comment.

vishnoianil:
> > @vishnoianil FYI
> > We shouldn't be allowing users to change settings underneath us and try to guess what their intention was. If they need to change the setting it should be done via CNO.
> 
> CNO is not guessing here, it's just trying to not override if user manually configure the value. OVN itself is a one big subsystem underneath and exposes it's own API that user can use to change config. If we really want user to not change the values i think we should not allow permission to user to use those utility tools. Unfortunately, CNO is not very flexible when it comes to changing the values, for that you need to stop CVO & it will restart all the pods, which you don't want to do in production environment.
> 
> The option @dcbw mentioned below seems a bit better option, but until and unless cluster deny the new node joining if it passes that #node, you are always at the risk that user can add more nodes. If user add more nodes and raft start partitioning more frequently, what is our options ? Do we ask user to disable CVO and change the value through CNO?

@trozet thanks for reverting the patch, when i tested the patch, i was testing it after setting the higher value manually and restarting it. Forgot the default 1 second case. Dumb mistake.

PR #770, opened 2020-08-31
Bug 1874457: Ensure router is present when Kuryr is used

MaysaMacedo:
Kuryr requires a router to be present on the cluster
to allow communication between Pod and Services. However,
during a FIP less no router is created by the installer.
This commit makes sure a router is created if not already
present.

### DIFF
--- a/pkg/platform/openstack/networking.go
+++ b/pkg/platform/openstack/networking.go
@@ -92,9 +92,31 @@ func ensureOpenStackSubnetpool(client *gophercloud.ServiceClient, name, tag stri
 	}
 }
 
-// Looks for a Neutron subnet by name and tag. Fails if subnet is not found
-// or multiple subnets match.
-func findOpenStackSubnet(client *gophercloud.ServiceClient, name, tag string) (subnets.Subnet, error) {
+// Looks for a Neutron Network by tag. Fails if network is not found
+// or multiple networks match.
+func findOpenStackNetwork(client *gophercloud.ServiceClient, tag string) (networks.Network, error) {
+	empty := networks.Network{}
+	opts := networks.ListOpts{Tags: tag}
+	page, err := networks.List(client, opts).AllPages()
+	if err != nil {
+		return empty, errors.Wrap(err, "failed to get network list")
+	}
+	networkList, err := networks.ExtractNetworks(page)
+	if err != nil {
+		return empty, errors.Wrap(err, "failed to extract networks list")
+	}
+	if len(networkList) == 1 {
+		return networkList[0], nil
+	} else if len(networkList) == 0 {
+		return empty, errors.New("network not found")
+	} else {
+		return empty, errors.New("multiple matching networks")
+	}
+}
+
+// Looks for a Neutron subnet by name and tag. In case of not found, looks for a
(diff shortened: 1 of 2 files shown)


MaysaMacedo:
/hold

Still needs to be tested.

pkg/platform/openstack/networking.go
```diff
@@ -183,7 +183,15 @@ func findOpenStackRouter(client *gophercloud.ServiceClient, name, tag string) (r
 	if len(routerList) == 1 {
 		return routerList[0], nil
 	} else if len(routerList) == 0 {
-		return empty, errors.New("router not found")
+		router, err := routers.Create(client, routers.CreateOpts{Name: name}).Extract()
```
luis5tb, line 186: we also need to connect the router to the VMs/Nodes network, right?

MaysaMacedo: True!

pkg/platform/openstack/networking.go
```diff
@@ -167,9 +208,10 @@ func ensureOpenStackSubnet(client *gophercloud.ServiceClient, name, tag, netId,
 	}
 }
 
-// Looks for a Neutron router by name and tag. Fails if router is not found
-// or multiple routers match.
-func findOpenStackRouter(client *gophercloud.ServiceClient, name, tag string) (routers.Router, error) {
+// Looks for a Neutron router by name and tag. If not found, provides
+// the router used by the costum Network. If no router exists, creates
```
luis5tb, line 212: custom

pkg/platform/openstack/networking.go
```diff
@@ -183,12 +225,53 @@ func findOpenStackRouter(client *gophercloud.ServiceClient, name, tag string) (r
 	if len(routerList) == 1 {
 		return routerList[0], nil
 	} else if len(routerList) == 0 {
-		return empty, errors.New("router not found")
+		ps, err := getOpenStackPortsByNetwork(client, networkID)
```
luis5tb, line 228: ps? perhaps worth to rename it to networkPorts?

pkg/platform/openstack/networking.go
```diff
@@ -183,12 +225,53 @@ func findOpenStackRouter(client *gophercloud.ServiceClient, name, tag string) (r
...
+				if err != nil {
+					return empty, errors.Wrap(err, "failed to get router list")
+				}
+				routerList, err := routers.ExtractRouters(page)
+				if err != nil {
+					return empty, errors.Wrap(err, "failed to extract routers list")
+				}
+				if len(routerList) == 1 {
+					return routerList[0], nil
+				}
```
dulek, line 244: If we have the ID of the router we can just do `routers.Get` and simplify this.

dulek (changes requested):
One small improvement that can be made.

pkg/platform/openstack/networking.go
MaysaMacedo: Indeed. Thank you!

MaysaMacedo:
Waiting for this [PR](https://github.com/openshift/installer/pull/4159) to get merged, before moving the hold.

PR #771, opened 2020-09-01
Bug 1874647: Initialize controller-runtime logging

danwinship:
While trying to figure out how we ended up with three simultaneous copies of CNO running in an IPv6 cluster, I found that we're currently dropping all logs from code in `sigs.k8s.io/controller-runtime`.

This is hacky but we should eventually stop using controller-runtime anyway, so...

### DIFF
--- a/cmd/cluster-network-operator/log.go
+++ b/cmd/cluster-network-operator/log.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"log"
+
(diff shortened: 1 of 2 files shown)


danwinship:
/hold
I want to see how much additional logging this generates...

danwinship:
/override ci/prow/e2e-gcp-ovn 
patch is plugin-independent; passing sdn jobs is sufficient

PR #772, opened 2020-09-01
Bug 1874638: Fixes missing nbctl daemon logging

trozet:

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -521,7 +521,10 @@ spec:
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            --db "{{.OVN_NB_DB_LIST}}")
+            --db "{{.OVN_NB_DB_LIST}}" --log-file=/run/ovn/ovn-nbctl.log)
+
+          # include nbctl daemon logging, allow for ovn-nbctl to create the log file
+          tail -f --retry /run/ovn/ovn-nbctl.log &
 
            # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
           ln -sf $OVN_NB_DAEMON /var/run/ovn/ || true


trozet:
@astoycos FYI

astoycos:
Thanks @trozet this should help in debugging future issues between `ovnkube-master` and `nbdb`

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -521,7 +521,11 @@ spec:
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            --db "{{.OVN_NB_DB_LIST}}")
+            --db "{{.OVN_NB_DB_LIST}} --log-file=/run/ovn/ovn-nbctl.log")
+
+          # include nbctl daemon logging, allow for ovn-nbctl to create the log file
+          sleep 1
+          tail -f /run/ovn/ovn-nbctl.log &
```
dcbw, line 528: @trozet `tail -f --retry` instead of a sleep 1?

danwinship, line 528: add `--retry` to the `tail` args and then it will work even if "sleep 1" isn't long enough.

(You could theoretically get rid of the "sleep 1" entirely, but then tail would log a pointless warning, so it's best to have both.)

trozet: oh never knew about that argument, nice I'll change it

danwinship:
Jinx! So the problem I was saying is, the way it is now, you'll get:

    tail: cannot open '/run/ovn/ovn-nbctl.log' for reading: No such file or directory
    tail: '/run/ovn/ovn-nbctl.log' has appeared;  following new file

ugh, and actually you get

    tail: warning: --retry only effective for the initial open

too. Way to be useless, `tail`...

trozet:
hmm not seeing the nbctl logs

trozet:
I see a bunch of ovn-nbctl logging now in the ovn-kube logs.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -521,7 +521,10 @@ spec:
           export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
             --detach \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            --db "{{.OVN_NB_DB_LIST}}")
+            --db "{{.OVN_NB_DB_LIST}}" --log-file=/run/ovn/ovn-nbctl.log)
+
+          # include nbctl daemon logging, allow for ovn-nbctl to create the log file
+          tail -f --retry /run/ovn/ovn-nbctl.log &
```
abhat, line 527: why do we need to tail it?

trozet: so we can see exactly in the ovnkube-master logs when the commands happen. Really helpful in debugging. I suspect as we migrate more and more to the golang bindings they will eventually go away, but it's a good way to track things as well that need to get migrated.

abhat: Sorry I should have provided more context. I meant, yes, its useful to tail it, but that data is already in the log file on disk. But I suppose the real value add is the interleaving of those commands with the master logs.

astoycos:
@abhat It's weird the merge bot is still not recognizing `ci/prow/e2e-openstack` as not required

bindata/network/ovn-kubernetes/ovnkube-master.yaml
rbbratta: `-f --retry` is `--follow=descriptor --retry` whereas the usual `-F` is `-follow=name --retry`

```
       -f, --follow[={name|descriptor}]
              output appended data as the file grows;

              an absent option argument means 'descriptor'
```

We usually want `--follow=name`, is there a reason we are only following descriptor.

Also, do we want `--pid=$(cat /var/run/ovn/ovn-nbctl.pid)`  as in the other tail commands to stop the tail once the daemon stops?

PR #773, opened 2020-09-02
Bug 1874696: Set hostPID for sdn ovs DaemonSet

pliurh:
Set hostPID to 'true', so that systemctl in container can talk to the host systemd which has host PID 1.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -24,6 +24,7 @@ spec:
     spec:
       serviceAccountName: sdn #needed to run privileged pods; not used for api access
       hostNetwork: true
+      hostPID: true
       priorityClassName: system-node-critical
       containers:
       - name: openvswitch


juanluisvaladas:
@pliurh I don't understand this PR:

> Set hostPID to 'true', so that systemctl in container can talk to the host systemd which has host PID 1.

What do you mean by talk exactly? signaling it?
As far as I know, systemd exposes all its DBus API on a unix socket.

pliurh:
> @pliurh I don't understand this PR:
> 
> > Set hostPID to 'true', so that systemctl in container can talk to the host systemd which has host PID 1.
> 
> What do you mean by talk exactly? signaling it?
> As far as I know, systemd exposes all its DBus API on a unix socket.

The `systemctl` need to talk to systemd which has the PID 1 in a host. However, in a container, the PID 1 is not the systemd. So we got `Failed to connect to bus: No data available` when executing `systemctl status ovsdb-server` in the container. Using host PID namespace where PID 1 is systemd, that solves the issue.

danwinship:
/hold
it used to be `hostPID` but we removed that in #524... the PR is not clear about why though. Do you remember, @abhat ?

I believe it ought to be possible for us to talk to systemd without being hostPID; like, if we mount the appropriate socket in `/run` or something...

pliurh:
Also I find we also have `hostPID` for `ovs-node` DaemonSet of OVN-kube. I suppose we shall keep it aligned in both DaemonSets, right?

abhat:
> /hold
> it used to be `hostPID` but we removed that in #524... the PR is not clear about why though. Do you remember, @abhat ?
> 
> I believe it ought to be possible for us to talk to systemd without being hostPID; like, if we mount the appropriate socket in `/run` or something...

Yeah the reason we removed that was to make sure the process hierarchy is maintained so that `terminationGracePeriodSeconds` is honored correctly by all the processes. This was needed to make sure that the save flows works reliably.

abhat:
> Also I find we also have `hostPID` for `ovs-node` DaemonSet of OVN-kube. I suppose we shall keep it aligned in both DaemonSets, right?

The save/restore flows was never done for OVN because we said we will get to system ovs by the time we have ovn-kube GA. So the hostPID setting is divergent.

abhat:
> > /hold
> > it used to be `hostPID` but we removed that in #524... the PR is not clear about why though. Do you remember, @abhat ?
> > I believe it ought to be possible for us to talk to systemd without being hostPID; like, if we mount the appropriate socket in `/run` or something...
> 
> Yeah the reason we removed that was to make sure the process hierarchy is maintained so that `terminationGracePeriodSeconds` is honored correctly by all the processes. This was needed to make sure that the save flows works reliably.

abhat:
Since we only run `tail` in the container now, we can probably change `hostPID: true` for 4.6, but like @danwinship mentioned, it should likely be possible to talk to systemd without hostPID by setting the right mounts.

pliurh:
> Since we only run `tail` in the container now, we can probably change `hostPID: true` for 4.6, but like @danwinship mentioned, it should likely be possible to talk to systemd without hostPID by setting the right mounts.

I tried with `chroot /host/ systemctl status ovsdb-server`, and it failed with the same error. So I don't think setting mounts can solve this issue.

juanluisvaladas:
According to the issue you mention this is a systemctl limitation rather than systemd limitation and dbus api still works.

Can you use busctl or gdbus instead of systemctl to make the API call?

pliurh:
> According to the issue you mention this is a systemctl limitation rather than systemd limitation and dbus api still works.
> 
> Can you use busctl or gdbus instead of systemctl to make the API call?

What is the drawback of using hostPID? Now the ovs pod is only for collecting the logs.

juanluisvaladas:
OK since the container is there only for logs, I don't really care how it's done. Either way it's fine for me.

danwinship:
> Now the ovs pod is only for collecting the logs.

well, we still have the code to run OVS in the container there too. Why is that still there? Did we just not remove it? Is it needed for upgrades?

abhat:
I guess for a >4.6 cluster, that is dead-code, hopefully we should never start ovs in the container (except when we do like here). I think @squeed left it in there as a fall-back mostly.

abhat:
FWIW, I am ok making hostPID to true for the save/restore functionality and termination ordering is moot now.

rbbratta:
On 4.6.0-0.nightly-2020-09-07-104243 the systemctl commands seem to run okay in the OVS pods without hostPid.  Does something change in the pods during migration to OVN?

rbbratta:
is the `export SYSTEMD_IGNORE_CHROOT=yes` set?  Or something else is going on.

```
sh-4.2# systemctl status
● ip-10-0-217-110
    State: running
     Jobs: 0 queued
   Failed: 0 units
    Since: Tue 2020-09-08 01:00:24 UTC; 56min ago
   CGroup: /
           └─kubepods.slice
             └─kubepods-burstable.slice
               └─kubepods-burstable-pod1100ef9f_d705_4e7b_9a07_9347b19cd12e.slice
                 └─crio-28510f94eccbb3690f055bf74b3cd51f969d2ed04b7ba898949f82a7dc9d21f4.scope
                   ├─   1 tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
                   ├─5036 /bin/sh
                   ├─5130 systemctl status
                   └─5131 more
```

pliurh:
> On 4.6.0-0.nightly-2020-09-07-104243 the systemctl commands seem to run okay in the OVS pods without hostPid. Does something change in the pods during migration to OVN?

It is not about migration at all (actually it blocks migration from working). I found this issue on a fresh-installed cluster (4.6.0-0.ci-2020-09-06-060329) with openshift-sdn. 

`export SYSTEMD_IGNORE_CHROOT=yes` doesn't help either.

```
[root@nfvsdn-18 dev-scripts]# oc rsh -n openshift-sdn ovs-bvzwt
sh-4.4# export SYSTEMD_IGNORE_CHROOT=yes
sh-4.4# systemctl status openvswitch
Failed to connect to bus: No data available
```

pliurh:
Tested with 4.6.0-0.nightly-2020-09-07-224533, no such issue found. Not sure what changed the behavior. However, I still think we shall set `hostPID` to true. So that the systemctl behavior will not be affected.

juanluisvaladas:
> I guess for a >4.6 cluster, that is dead-code, hopefully we should never start ovs in the container (except when we do like here). I think @squeed left it in there as a fall-back mostly.

How confident are we on not using the fall-back? If we aren't sure enough toremove it we should just talk to  systemd through dbus.

abhat:
> > I guess for a >4.6 cluster, that is dead-code, hopefully we should never start ovs in the container (except when we do like here). I think @squeed left it in there as a fall-back mostly.
> 
> How confident are we on not using the fall-back? If we aren't sure enough toremove it we should just talk to systemd through dbus.

As discussed earlier today, we need to have the fall-back code in place for upgrade/rollback cases, because the ordering is not guaranteed. So in that case, we need to make sure that we leave the code as is and figure out a way to query using dbus potentially. @juanluisvaladas PTAL with @pliurh.

PR #777, opened 2020-09-07
Bug 1874591: autodect mtu support for ipv6

aojea:
The code was taking into account IPv4 routes only.
It consider IPv6 routes now


aojea:
It seems that the e2e-openstack and e2e-vsphere job never worked :thinking:

PR #778, opened 2020-09-07
Bug 1884430: Move kube-rbac-proxy to general ovnkube DeamonSet

bond95:
Follow up of this PR: https://github.com/openshift/cluster-network-operator/pull/751

Basically it's fixing same issue as for SDN. To avoid deploying metric's DeamonSet before OVN, metrics was moved to OVN DeamonSet.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -309,7 +309,80 @@ spec:
         - name: nb-db-raft-port
           containerPort: {{.OVN_NB_RAFT_PORT}}
         terminationMessagePolicy: FallbackToLogsOnError
-      
+
+      - name: kube-rbac-proxy
+        image: {{.KubeRBACProxyImage}}
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -euo pipefail
+          TLS_PK=/etc/pki/tls/metrics-cert/tls.key
+          TLS_CERT=/etc/pki/tls/metrics-cert/tls.crt
+          # As the secret mount is optional we must wait for the files to be present.
+          # The service is created in monitor.yaml and this is created in sdn.yaml.
+          # If it isn't created there is probably an issue so we want to crashloop.
+          retries=0
+          while [[ "${retries}" -lt 100 ]]; do
+            TS=$(
+              curl \
+                -s \
+                --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+                -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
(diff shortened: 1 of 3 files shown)


bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -286,63 +343,6 @@ spec:
       - name: ovn-cert
         secret:
           secretName: ovn-cert
```
danwinship, line 345: I believe the `volume` entry in both `ovnkube-node.yaml` and `ovnkube-master.yaml` needs to be made `optional: true` (as it is in `bindata/network/openshift-sdn/sdn.yaml`), because the secret can't be created until after the network is already up, but if the volume is required then the network won't come up without the secret having already been created.

juanluisvaladas: Tests are failing precisely because of this

juanluisvaladas:
@bond95 can you add this to the PR as well?
https://github.com/openshift/cluster-network-operator/pull/786

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -309,7 +309,80 @@ spec:
...
+          retries=0
+          while [[ "${retries}" -lt 100 ]]; do
+            TS=$(
+              curl \
+                -s \
+                --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+                -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+                "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/ovn-kubernetes/services/ovnkube-master" |
+                  python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
+            )
```
juanluisvaladas, line 335: replace this line with
) || :

So that we don't get the error treated as a fatal. And the same with the node one.

juanluisvaladas (changes requested):

juanluisvaladas:
@abhat this includes PRs #751 and #786 . I've verified it works on a cluster manually. Can you PTAL?

juanluisvaladas:
/retest
@danwinship can you PTAL? this includes PRs #751 and #786 . I've verified it works on a cluster manually

danwinship:
/lgtm
but it needs a bugzilla bug too

juanluisvaladas:
/retitle Bug 1884430 Move kube-rbac-proxy to general ovnkube DeamonSet

juanluisvaladas:
/retitle Bug 1884430: Move kube-rbac-proxy to general ovnkube DeamonSet

tssurya:
@juanluisvaladas : can we please make sure we don't break that alerts-job thing again ? We have already bounced around this label change quiet a few times:
1) https://github.com/openshift/cluster-network-operator/pull/742/files
2) https://github.com/openshift/cluster-network-operator/pull/436/files

juanluisvaladas:
@tssurya what needs to be changed? I see tests are passing so it should be fine

PR #781, opened 2020-09-07
Bug 1875438: [release-4.4] OVN: Handle OVN db upgrades

alexanderConstantinescu:
4.4 back-port of #755

/assign @knobunc

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -169,6 +169,22 @@ spec:
                   fi
                   sleep 2
                   done
+
+                  # Upgrade the db if required.
+                  DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
+                  DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                  schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                  db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                  target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                    :
+                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
+                  else
+                      echo "Upgrading database $schema_name from schema version $db_version to $target_version"
+                      ovsdb-client -t 30 convert "$DB_SERVER" "$DB_SCHEMA"
+                  fi
                 fi
         readinessProbe:
           initialDelaySeconds: 30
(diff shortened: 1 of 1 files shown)


alexanderConstantinescu:
Why are we retesting? Everything required has already passed, right?

dcbw:
> Why are we retesting? Everything required has already passed, right?

No OVN tests have passed, which is what we care about here right? Unless the github test list is just completely lying to me.

alexanderConstantinescu:
@trozet @numansiddique @dcbw , I've added a commit which deletes the NB/SB DB on a detected downgrade. PTAL and let me know if that makes sense?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -169,6 +169,23 @@ spec:
...
+                  DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                  schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                  db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                  target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                  if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                    :
+                  elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), destroy the DB"
+                      rm -rf /etc/ovn/ovnnb_db.db
```
trozet, line 184: I wonder if this will have the same implications that we are discussing with wiping all the dbs in upstream where ovn-controller and ovn-northd need to also be reset. @abhat FYI

For this particular case, northd probably doesn't matter because all of the master nodes are going to be bounced. However, for ovn-controller and sb_db the nodes will reject database updates until they get restarted, which might be OK on worker nodes since those will be upgraded after master nodes. Still what about ovnkube-node pods on the masters. What if they have come up and connected to the previous db before masters are done? Do we need to reset those?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -322,7 +323,8 @@ spec:
                   if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
                     :
                   elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
-                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
+                      echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), destroy the DB"
+                      rm -rf /etc/ovn/ovnsb_db.db
```
trozet, line 327: Also I'm not sure putting this 2nd downgrade commit into 4.4 is right. The 4.5 and 4.6 commits for fixing upgrades do not have this commit right? Maybe it would be better to just stick to the original commit and get that in with this PR, file another bug about possible db downgrade case and get that commit into 4.6,4.5,4.4? Or is this a one time thing we only need in 4.4?

alexanderConstantinescu: Yeah, it was more for consistency between version + having one less PR to do. But yes, we will never downgrade on 4.4. I will remove it.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
numansiddique: This code is executed after the nbdb and sbdb containers/pods are started and ovsdb-servers are already running. After deleting the db, I think its better to restart the containers. Otherwise there would be wierd behavior.

knobunc:
/override ci/prow/e2e-gcp

This is an ovn only change, and is critical.  It should not be blocked by unrelated flakes.

PR #782, opened 2020-09-09
Bug 1854306: Remove ovs bridges when sdn/ovnkube pod is terminated

pliurh:
Remove bridge br0 when sdn pod is terminated.
Remove bridge br-int and br-local when ovnkube-node pod is terminated.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -167,7 +167,12 @@ spec:
         lifecycle:
           preStop:
             exec:
-              command: ["rm","-f","/etc/cni/net.d/80-openshift-network.conf", "/host/opt/cni/bin/openshift-sdn"]
+              command:
+              - /bin/bash
+              - -c
+              - |
+                rm -f /etc/cni/net.d/80-openshift-network.conf /host/opt/cni/bin/openshift-sdn
+                ovs-vsctl --timeout=30 --if-exists del-br br0
         readinessProbe:
           exec:
             # openshift-sdn writes this file when it is ready to handle pod requests.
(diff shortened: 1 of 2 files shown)


pliurh:
@abhat @trozet PTAL.

danwinship:
/hole

We can't do this; it will make upgrades extremely disruptive, because traffic from all pods will stop when the bridge is deleted, and then when sdn restarts, it will have to kill every pod on the node because it can't reattach the old pods to a new bridge without having the information from the old ovsdb.

Is this for migration? Don't the nodes get rebooted in between sdn and ovn-kubernetes? I was going to say we could just always erase the ovsdb at boot time, but IIRC there's some reason we don't want to do that...

danwinship:
ahem
/hold

pliurh:
> Is this for migration? Don't the nodes get rebooted in between sdn and ovn-kubernetes? I was going to say we could just always erase the ovsdb at boot time, but IIRC there's some reason we don't want to do that...

Yes, it's for migration. And yes, the nodes get rebooted during migration. However, the bridges will still be there after reboot, as there is no logic that removes them. 
If the bridges shall not be removed here. Can we add this logic in the ovs-configuration.service? I have another PR that is also related to the ovsdb cleanup during the migration. https://github.com/openshift/machine-config-operator/pull/2066

danwinship:
Commented on the MCO PR. Yes, I think it would make sense to merge this into that.

pliurh:
The keepalived fix has been merged. So I remove the workaround of deleting `keepalived.conf`.

pliurh:
moved the logic to openshift/machine-config-operator#2066

PR #785, opened 2020-09-09
Bug 1874696: Remove systemctl calls

juanluisvaladas:
/assign @pliurh 

When systemctl is run in a separate PID namespace where PID 1 isn't systemd, it doesn't behave the normal way: https://github.com/systemd/systemd/issues/11300

The solution for this is to entirely remove calls to systemctl calls.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -35,8 +35,23 @@ spec:
           #!/bin/bash
           set -euo pipefail
           export SYSTEMD_IGNORE_CHROOT=yes
+
+          # systemctl cannot be used in a separate PID namespace to reach
+          # the systemd running in PID 1. Therefore we need to use the dbus API
+          systemctl_restart(){
+            gdbus call \
+              --system \
+              --dest org.freedesktop.systemd1 \
+              --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
+              --method org.freedesktop.systemd1.Unit.Restart "replace"
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1//-/_2d}"
+          }
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if [[ -f '/host/usr/local/bin/configure-ovs.sh' ]]; then
             echo "openvswitch is running in systemd"
             # In some very strange corner cases, the owner for /run/openvswitch
             # can be wrong, so we need to clean up and restart.
@@ -44,9 +59,8 @@ spec:
             ovs_gid=$(chroot /host id -g openvswitch)
             chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
             if [[ ! -S /run/openvswitch/db.sock ]]; then
-              systemctl restart ovsdb-server
+              systemctl_restart ovsdb-server
             fi
-
             # Don't need to worry about restoring flows; this can only change if we've rebooted
             exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
           else
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -48,8 +48,24 @@ spec:
             set +o allexport
           fi
           export SYSTEMD_IGNORE_CHROOT=yes
+
+          # systemctl cannot be used in a separate PID namespace to reach
+          # the systemd running in PID 1. Therefore we need to use the dbus API
+          systemctl_restart(){
+            gdbus call \
+              --system \
+              --dest org.freedesktop.systemd1 \
+              --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
+              --method org.freedesktop.systemd1.Unit.Restart "replace"
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1//-/_2d}"
+          }
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if [[ -f '/host/usr/local/bin/configure-ovs.sh' ]]; then
             echo "openvswitch is running in systemd"
             # In some very strange corner cases, the owner for /run/openvswitch
             # can be wrong, so we need to clean up and restart.
@@ -57,7 +73,7 @@ spec:
             ovs_gid=$(chroot /host id -g openvswitch)
             chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
             if [[ ! -S /run/openvswitch/db.sock ]]; then
-              systemctl restart ovsdb-server
+              systemctl_restart ovsdb-server
             fi
             # Don't need to worry about restoring flows; this can only change if we've rebooted
             exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log


bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,34 @@ spec:
           #!/bin/bash
           set -euo pipefail
           export SYSTEMD_IGNORE_CHROOT=yes
-          # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          # systemctl may not be used in a separate PID namespace to reach
+          # the systemd running in PID 1. Therefore we need to use the dbus API
+          systemctl_is_active(){
+            return gdbus call \
```
pliurh, line 41: s/return//g

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,34 @@ spec:
...
+            gdbus call \
+                  --system \
+                  --dest org.freedesktop.systemd1 \
+                  --object-path /org/freedesktop/systemd1/unit/"${1}"_2eservice \
+                  --method o0rg.freedesktop.systemd1.Unit.Restart "replace"
+          }
+
+          # Check to
 see if ovs is provided by the node:
+          if systemctl_is_active ovsdb-server; then
```
pliurh, line 57: s/ovsdb-server/ovsdb_server/g

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,34 @@ spec:
...
             echo "openvswitch is running in systemd"
             # In some very strange corner cases, the owner for /run/openvswitch
             # can be wrong, so we need to clean up and restart.
             ovs_uid=$(chroot /host id -u openvswitch)
             ovs_gid=$(chroot /host id -g openvswitch)
             chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
             if [[ ! -S /run/openvswitch/db.sock ]]; then
-              systemctl restart ovsdb-server
+              systemctl_restart ovsdb-server
```
pliurh, line 65: s/ovsdb-server/ovsdb_server/g

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -54,7 +72,7 @@ spec:
             echo "openvswitch is running in container"
             # if another process is listening on the cni-server socket, wait until it exits
             retries=0
-            while true; do
+            0while true; do
```
pliurh, line 75: s/0while/while/g

pliurh (commented):
After fixing the nits, It works as expected in my environment.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,34 @@ spec:
...
+          systemctl_restart(){
+            gdbus call \
+                  --system \
+                  --dest org.freedesktop.systemd1 \
+                  --object-path /org/freedesktop/systemd1/unit/"${1}"_2eservice \
+                  --method o0rg.freedesktop.systemd1.Unit.Restart "replace"
+          }
+
+          # Check to
 see if ovs is provided by the node:
```
pliurh, line 56: One more. There is an invisible character that breaks the YAML file.

bindata/network/openshift-sdn/sdn-ovs.yaml
pliurh: s/ovsdb-server/ovsdb_2dserver/g

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,34 @@ spec:
...
+                   --object-path /org/freedesktop/systemd1/unit/"${1}"_2eservice \
+                   --method org.freedesktop.DBus.Properties.Get "org.freedesktop.systemd1.Unit" "ActiveState"  |
+                     fgrep "(<'active'>,)" -q
+          }
+          systemctl_restart(){
+            gdbus call \
+                  --system \
+                  --dest org.freedesktop.systemd1 \
+                  --object-path /org/freedesktop/systemd1/unit/"${1}"_2eservice \
+                  --method o0rg.freedesktop.systemd1.Unit.Restart "replace"
```
pliurh, line 53: s/o0rg/org/g

pliurh:
@juanluisvaladas Could you put the bz id 1874696 in the title?

juanluisvaladas:
/retitle Bug 1874696: Replace systemctl calls with DBus API calls
/retest

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,39 @@ spec:
...
+            gdbus call \
+              --system \
+              --dest org.freedesktop.systemd1 \
+              --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
+              --method org.freedesktop.systemd1.Unit.Restart "replace"
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1}" | sed 's/-/_2d/g'
```
rbbratta, line 58: bash can do simple replaces itself using parameter expansion
```bash
echo "${1//-/_2d}"
```

      ${parameter/pattern/string}
              Pattern substitution.  The pattern is expanded to produce a pattern just as in pathname expansion, Parameter is expanded and the longest match of pattern against its value is replaced with string.  The match is performed using the rules described under Pattern Matching below.  If pattern begins
              with  /,  all matches of pattern are replaced with string.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,39 @@ spec:
...
+          }
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if systemctl_is_active ovsdb-server; then
```
rbbratta, line 62: do we need this in `bindata/network/ovn-kubernetes/006-ovs-node.yaml` as well?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,39 @@ spec:
...
           export SYSTEMD_IGNORE_CHROOT=yes
+          # systemctl may not be used in a separate PID namespace to reach
+          # the systemd running in PID 1. Therefore we need to use the dbus API
+          systemctl_is_active(){
+            gdbus call \
+              --system \
+              --dest org.freedesktop.systemd1 \
+              --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
+              --method org.freedesktop.DBus.Properties.Get "org.freedesktop.systemd1.Unit" "ActiveState"  |
+                fgrep "(<'active'>,)" -q
```
rbbratta, line 46: this won't match the other possible states: `reloading, inactive, failed, activating, deactivating`

There was a thread somewhere on slack about this.  I think we really only care if the systemd unit is "enabled", but "enabled" is too hard to track.  Even if the unit was started correctly on boot but has since restarted or failed we still want to know that it is supposed to be running on the host and not in the container.

`systemctl status ovsdb-server` exit status is blunt indicator that `ovsdb-server` is running or restarted, etc.

I think a better check would be to check for `ActiveEnterTimestamp > (<uint64 0>,) `  because that indicates that at some point since boot `ovsdb-server` was running, so it is our intention to run `ovsdb-server` on the host and not in a container, and it does not depend on whatever the current state of `ovsdb-server` is, be it `activating` or `reloading`, etc.

cc @trozet

juanluisvaladas:
/hold
to address @rbbratta 's comment

juanluisvaladas:
@rbbratta I used `ExecMainPD` instead of `ActiveEnterTimestamp` because ActiveEnterTimestamp means the process has been active at some point and the process may be running but not active.

The ovsdb-server currently has type=forking so it's effectively the same, but if at some point in the future they moved to `Type=dbus` or `Type=notify` we should be covered.

juanluisvaladas:
PS: I'm keeping the hold until I see a CI run to see it works, if it works I'll copy this to `bindata/network/ovn-kubernetes/006-ovs-node.yaml`

juanluisvaladas:
@rbbratta @pliurh can you please validate the changes?
/hold cancel

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,42 @@ spec:
...
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1//-/_2d}"
+          }
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if systemctl_has_started ovsdb-server; then
```
danielmellado, line 65: I'm not going to block this, and I'm aware that this is bash within a yaml file and so, but let me give you a few comments.
* If we're going to be reusing this, it'd be worth to add the functions in bash to another folder and call them within there.
* I see a lot of comments, which I appreciate, within the code, but none on the PR/Commit. Please address that out as well as it makes things easier when reviewing.
* This might also be worth a release note.

danielmellado (approved):
Won't be blocking this, LGTM, but please address the comments for future reference.

bindata/network/openshift-sdn/sdn-ovs.yaml
juanluisvaladas: > If we're going to be reusing this, it'd be worth to add the functions in bash to another folder and call them within there.
That'd be good, but since it's embedded and these are static files injected into CNO, I don't know how could we do that.
If you have any suggestion about how to do that I'm happy to hear about it.

> I see a lot of comments, which I appreciate, within the code, but none on the PR/Commit. Please address that out as well as it makes things easier when reviewing.

Will address that.

> This might also be worth a release note.

The code I've just modified has never been released, this was introduced in 4.6 so none of our customers are actually running this code, there is a release note saying that ovs now runs in systemd, which I believe is sufficient.

juanluisvaladas:
/hold
This is flaky.
Gdbus has problems connecting to the system bus. I don't understand yet why.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,31 @@ spec:
...
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1//-/_2d}"
+          }
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if [[ -f '/etc/systemd/system/multi-user.target.wants/openvswitch.service' ]]; then
```
rbbratta, line 54: I think Casey had a comment about this

https://github.com/openshift/cluster-network-operator/pull/672#issuecomment-647980159

> We could use is-enabled as a predicate, but that's not reliable for a few reasons - first of all, and this is the case on RHCOS, units without an [Install] section are defined as enabled (even though they're not). Second of all, we might not enable the unit directly in the future; we could make it a dependency of a meta-service. That would break this predicate again.

> I don't want to get in the business of checking for systemd configuration files on disk. They can live in a lot of places, so that's just setting ourselves up for trouble

juanluisvaladas: We have a big issue here, the problem of the `Failed to connect to bus: No data available` error happens both with gdbus and systemctl as we can see here:
https://storage.googleapis.com/origin-ci-test/pr-logs/pull/25495/pull-ci-openshift-origin-master-e2e-gcp/1306219874485276672/artifacts/e2e-gcp/gather-extra/pods/openshift-sdn_ovs-5x4hr_openvswitch.log

We're two weeks aware from the code freeze and this is blocking several people. Debugging gdbus/systemctl will take time and the error is likely produced by underlying issues, and heavily dependant on timing. While I agree tihs solution isn't great as long as we can guarantee we're not enabling the ovs service in a different way during the 4.6 time frame we're pretty safe.

The code to run ovs outside 4.6 should be removed in 4.7 anyway so keeping this (which again is ugly) for 3 months doesn't look like such a bad decision...

danwinship: > The code to run ovs outside 4.6 should be removed in 4.7 anyway so keeping this (which again is ugly) for 3 months doesn't look like such a bad decision...

I'm not sure what you're suggesting? Are you saying we should drop this PR? That we should merge it as it is now?

juanluisvaladas: > I'm not sure what you're suggesting? Are you saying we should drop this PR?
No, what I'm trying to say is that because the fallback to start the ovs daemons in the container is only necessary for the 4.5 to 4.6 upgrade, eventually the vast majority of the start up script will be gone in 4.7.

> That we should merge it as it is now?
Yes, checking if the file is there has a bunch of problems but since it has such a limited scope I think it's viable.

juanluisvaladas:
/retest
Let's hope for a build with artifacts...

juanluisvaladas:
/hold cancel
to see if at least ONE run gets some logs related to the OVS DS in SDN.

OVN seems to be fine.

juanluisvaladas:
/hold
CI is completely mangled, there is o point having this thing running....

bindata/network/openshift-sdn/sdn-ovs.yaml
rbbratta: I'm fine with establishing a exact convention for detecting that host-OVS is enabled.   As long as MCO and CNO agree on the convention we should be okay, but this means we should put a warning in MCO to indicate that MCO should not change the `openvswitch.service` setting without updating CNO.

rbbratta: Here maybe
https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/units/openvswitch.service.yaml#L2

juanluisvaladas: @danwinship @pliurh thoughts?

abhat: @rbbratta you mean the location of the service file that we will check for here in the CNO yes?

juanluisvaladas:
Depends on https://github.com/openshift/machine-config-operator/pull/2102/

juanluisvaladas:
Blocked by: https://github.com/openshift/machine-config-operator/pull/2066

When it merges we can test this one.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,16 +35,31 @@ spec:
...
+          }
+          svc_encode_name(){
+            # systemd encodes some characters, so far we only need to encode
+            # the character "-" but there may be more in the future.
+            echo "${1//-/_2d}"
+          }
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if [[ -f '/usr/local/bin/configure-ovs.sh' ]]; then
```
danwinship, line 54: I think you need `/host` there
(and in ovnkube below)

danwinship (changes requested):
yeah, I think this should work

danwinship:
> Blocked by: [openshift/machine-config-operator#2066](https://github.com/openshift/machine-config-operator/pull/2066)

This is no longer relevant, right? This PR can merge?

juanluisvaladas:
> This is no longer relevant, right? This PR can merge?

It's still relevant and we still need it to merge. The one that is no longer relevant is openshift/machine-config-operator#2102 .
Otherwise `/usr/local/bin/configure-ovs` is not present with openshift-sdn.

pliurh:
> > This is no longer relevant, right? This PR can merge?
> 
> It's still relevant and we still need it to merge. The one that is no longer relevant is [openshift/machine-config-operator#2102](https://github.com/openshift/machine-config-operator/pull/2102) .
> Otherwise `/usr/local/bin/configure-ovs` is not present with openshift-sdn.

`/usr/local/bin/configure-ovs` is always present on 4.6 node, and not present on 4.5 node. And openshift/machine-config-operator#2066 won't change that.

juanluisvaladas:
/hold cancel
actually teh file is created anyway, it's only the service is not enabled, but we don't need it anyway

juanluisvaladas:
/retest
@danwinship can you lgtm? I already changed the path to include /host/.

wking:
Checking [Prow][1], two consecutive update jobs have timed out after 4h30m.  Possibly something related to this change?  Not much in [the gathered assets][2] to help debug this, but now that we're multi-step perhaps we can get a timeout on just the test step so we leave time to gather resources even if that times out... 

[1]: https://prow.ci.openshift.org/?pull=785&job=pull-ci-openshift-cluster-network-operator-master-e2e-upgrade
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/785/pull-ci-openshift-cluster-network-operator-master-e2e-upgrade/1308837080670932992/artifacts/e2e-upgrade/

juanluisvaladas:
I don't think these failures are related to the changes in this PR. For instance this test seems to behave as expected on our end:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/785/pull-ci-openshift-cluster-network-operator-master-e2e-upgrade/1309032575599120384

For instance, here OVS is running as expected.:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/785/pull-ci-openshift-cluster-network-operator-master-e2e-upgrade/1309032575599120384/artifacts/e2e-upgrade/gather-extra/pods/openshift-sdn_ovs-krx67_openvswitch.log

juanluisvaladas:
Yep azure is broken
` level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply Terraform: error(AzureEventualConsistencyFailure) from Infrastructure Provider: Failed to find a resource that was recently created usualy caused by Azure's eventual consistency delays." `

danwinship:
The failure seen in `e2e-upgrade` here is unique to this PR; it passes (or fails during the e2e-tests) for other PRs

danwinship:
So it seems like in e2e-upgrade (which is SDN-based as opposed to e2e-gcp-ovn-upgrade, which passed), "something" is happening which breaks the network entirely, such that that job eventually times out, and then the post-job must-gather fails as well. So it seems like the patch is wrong, somehow; it does the right thing on initial install, but not on reboot/restart.

danwinship:
Bringing up a cluster with this PR, everything seems to work, and neither killing OVS pods nor rebooting nodes seems to trigger any bad behavior

wking:
You need to bring up a recent nightly, and the update to a release image with this PR.  Possibly installing this PR, updating to a nightly, and tgen updating back to this PR would reproduce whatever is happening in CI.

dcbw:
@juanluisvaladas after some discussion, we thing that we can delete this entire block:

```
            echo "openvswitch is running in systemd"
            # In some very strange corner cases, the owner for /run/openvswitch
            # can be wrong, so we need to clean up and restart.
            ovs_uid=$(chroot /host id -u openvswitch)
            ovs_gid=$(chroot /host id -g openvswitch)
            chown -R "${ovs_uid}:${ovs_gid}" /run/openvswitch
            if [[ ! -S /run/openvswitch/db.sock ]]; then
              systemctl restart ovsdb-server
            fi
```

because

(a) Tim thinks we don't have the UID/GID problem anymore because OVS will now always start before kubelet (which is the source of the ownership problem in the first place when it creates mounts for the container)
(b) we should fix permissions from MCO/configure-ovs.sh rather than the container, which can get the ownership wrong because it's inside a contianer
(c) we shouldn't ever need to restart ovsdbserver from the container when it's been started from the host

Can you strip out that block from both sdn/ovn and see if that helps things?

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -35,17 +35,10 @@ spec:
           #!/bin/bash
           set -euo pipefail
           export SYSTEMD_IGNORE_CHROOT=yes
+
           # Check to see if ovs is provided by the node:
-          if systemctl status ovsdb-server; then
+          if [[ -f '/host/usr/local/bin/configure-ovs.sh' ]]; then
             echo "openvswitch is running in systemd"
-            # In some very strange corner cases, the owner for /run/openvswitch
```
trozet, line 41: just add this back, because we know now the root cause of the e2e-upgrade job failure is not this. We can remove it later once @squeed is back.

knobunc:
The proposed fix works. The e2e-upgrade job was simply failing because it is trying to upgrade from the broken 4.6->PR 4.6. 4.5->4.6 should work.

/override ci/prow/e2e-gcp

abhat:
[`test upgrade openshift/cluster-network-operator#808 openshift/cluster-network-operator#785 aws,ovn`](https://meet.google.com/linkredirect?authuser=0&dest=https%3A%2F%2Fprow.ci.openshift.org%2Fview%2Fgs%2Forigin-ci-test%2Flogs%2Frelease-openshift-origin-installer-launch-aws%2F1311032386108002304) succeeded so 4.6 to 4.6 upgrades are working AOK.

abhat:
Upgrade jobs succeeded from 4.6 to 4.6 this time on SDN: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1311054422679228416

abhat:
@knobunc, need another override on the e2e-upgrade job.

PR #786, opened 2020-09-10
Bug 1883903:Add retries to SDN's RBAC proxy

juanluisvaladas:
Because kube-proxy may not be initialized by the time the RBAC proxy
starts it may crashloop for a while. Doesn't have any actual impact but
the restarts show in oc get pod and people may worry about that.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -188,12 +188,27 @@ spec:
           # As the secret mount is optional we must wait for the files to be present.
           # The service is created in monitor.yaml and this is created in sdn.yaml.
           # If it isn't created there is probably an issue so we want to crashloop.
-          TS=$(curl \
-            --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-            -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
-              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
-          )
+          retries=0
+          while [[ "${retries}" -lt 100 ]]; do
+            TS=$(
+              curl \
+                -s \
+                --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+                -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+                "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
+                  python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
+            )
+            if [ ! -z "TS" ]; then
+              break
+            fi
+            (( retries += 1 ))
+            echo $(date -Iseconds) INFO: Failed to get sdn service from API. Retry "${retries}"/100 1>&2
+            sleep 20
+          done
+          if [ "${retries}" -ge 20 ]; then
+            echo $(date -Iseconds) FATAL: Unable to get sdn service from API.
+            exit 1
+          fi
 
           TS=$(date -d "${TS}" +%s)
           WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
@@ -204,7 +219,7 @@ spec:
               if [[ "${CUR_TS}" -gt "WARN_TS"  ]]; then
                 echo $(date -Iseconds) WARN: sdn-metrics-certs not mounted after 20 minutes.
               elif [[ "${HAS_LOGGED_INFO}" -eq 0 ]] ; then
-                echo $(date -Iseconds) INFO: sdn-metrics-certs not mounted. Waiting one hour.
+                echo $(date -Iseconds) INFO: sdn-metrics-certs not mounted. Waiting 20 minutes.
                 HAS_LOGGED_INFO=1
               fi
           }


danielmellado (changes requested):
Overall LGTM, although, could you please put up a warning message should the retry fails?

juanluisvaladas:
@danielmellado Done

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -188,12 +188,20 @@ spec:
...
           # If it isn't created there is probably an issue so we want to crashloop.
-          TS=$(curl \
-            --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
-            -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-            "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
-              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
-          )
+          retries=0
+          until [ "$retries" -ge 20 ]
+          do
```
danwinship, line 193:
```suggestion
          while [[ "${retries}" -lt 20 ]]; do
```

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -188,12 +188,20 @@ spec:
...
+            TS=$(curl -s \
+              --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+              -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+              "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
+                python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])' &&
+                  break ||
+                  echo "WARN: Failed to get sdn service from API" 1>&2
```
danwinship, line 200: If this is _expected_ it should be a "warning". And you should say you're retrying.

Also, can you use `if`/`else` rather than `&&`/`||`? (I realize the variable assignment might make that tricky, so maybe no...)

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -188,12 +188,20 @@ spec:
...
+            )
+            retries=$(( retries + 1 ))
+            sleep 15
+          done
 
           TS=$(date -d "${TS}" +%s)
           WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
```
danwinship, line 207: hm... github won't let me review outside the context of the diff, but below here I see:

```
              if [[ "${CUR_TS}" -gt "WARN_TS"  ]]; then
                echo $(date -Iseconds) WARN: sdn-metrics-certs not mounted after 20 minutes.
              elif [[ "${HAS_LOGGED_INFO}" -eq 0 ]] ; then
                echo $(date -Iseconds) INFO: sdn-metrics-certs not mounted. Waiting one hour.
```

"one hour" should be "20 minutes" shouldn't it?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -188,12 +188,20 @@ spec:
...
-              python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])'
-          )
+          retries=0
+          until [ "$retries" -ge 20 ]
+          do
+            TS=$(curl -s \
+              --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+              -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+              "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/openshift-sdn/services/sdn" |
+                python -c 'import json,sys; print(json.load(sys.stdin)["metadata"]["creationTimestamp"])' &&
```
danielmellado, line 198: Just a nitpick, but are we 100% sure that the json python module would be available within the system? Maybe there's something around that installs it and I'm not aware of it, but if that's not the case, we should also catch that exception or make sure it's installed.

juanluisvaladas: It's a RHEL container so we're guaranteed to have it.

bindata/network/openshift-sdn/sdn.yaml
juanluisvaladas: Yes, will slide that in the PR

juanluisvaladas:
@danwinship while I was addressing it, I realized it never never exited if the retries were never successful, so besides addressing your comments I also added these new lines:

```
          if [ "${retries}" -ge 20 ]; then
            echo $(date -Iseconds) FATAL: Unable to get sdn service from API.
            exit 1
          fi
```

bindata/network/openshift-sdn/sdn.yaml
danwinship: all the stylistic suggestions here still apply, here and elsewhere. (single-line rather than multi-line and `while` rather than `until` because those are both more standard. `[[` rather than `[` and `${retries}` rather than `$retries` because they have fewer gotchas and it matches the kube/OCP shell script style guide that was never finalized...

danwinship:
Do we need a similar patch for any of the other daemonsets?

juanluisvaladas:
Yes, ovnkube-node and ovnkube-master daemonsets will need this as well, but because https://github.com/openshift/cluster-network-operator/pull/778 is not merged yet I can't modify it here, so I asked @bond95 to add this code in his PR

aojea:
seems the aws and gcp jobs never passed, should this PR affecting it somehow?
https://prow.ci.openshift.org/pr-history/?org=openshift&repo=cluster-network-operator&pr=786

juanluisvaladas:
/hold
There is something wrong here.

juanluisvaladas:
Still flaky
/hold

juanluisvaladas:
There's something wrong, just added a -x to get more logging, don't review yet

juanluisvaladas:
@danwinship I had to replace the if curl | python; then break; fi for a curl | python && break.

With the variable asignation it wasn't working as I expected.

juanluisvaladas:
There was an issue doing the break inside the subshell. Now this is fixed and ready to merge.Can you please lgtm?

juanluisvaladas:
/retitle Bug 1883903:Add retries to SDN's RBAC proxy

PR #793, opened 2020-09-16
Bug 1854306: Run ovn in shared gw mode when configure-ovs.sh run with parameter OVNKubernetes

pliurh:
ovs-configuration.service is enabled for both ovn-kube and openshift-sdn. So using the parameter of configure-ovs.sh as the condition to tell whether ovnkube shall be in a shared gw or local gw mode. In 4.5, this parameter is not added yet.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -531,7 +531,7 @@ spec:
           gateway_mode_flags=
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
-          if [ -f /etc/systemd/system/network-online.target.wants/ovs-configuration.service ]; then
+          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
             gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
             gateway_mode_flags="--gateway-mode local --gateway-interface none"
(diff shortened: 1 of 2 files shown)


pliurh:
/hold
don't merge before https://github.com/openshift/machine-config-operator/pull/2066 merged

pliurh:
@trozet I am not sure whether we still need this logic anymore since we will not run ovn-kube in old local gw mode.

danwinship:
/assign @trozet 
but I think this is irrelevant now

(if it's not, then you need to rebase because there's currently a merge commit there)

pliurh:
> /assign @trozet
> but I think this is irrelevant now
> 
> (if it's not, then you need to rebase because there's currently a merge commit there)

It's still relevant. Now the ovs-configuration.service is always enabled on 4.6 nodes, so we cannot use the file as the condition anymore. This PR makes the DaemonSets can be rendered correctly in the following 3 scenarios:
1. 4.6 pods on 4.5 host, no ovs-configuration.service, run with `--gateway-interface none`.
2. 4.6 pods on 4.6 host, ovs-configuration.service rendered with OVNKubernetes, run with `--gateway-interface br-ex`.
3. The intermediate stage of migration, 4.6 pods on 4.6 host, ovs-configuration.service rendered with Openshift-SDN, run with `--gateway-interface none`.

trozet:
for ovn 4.5-> 4.6 this change doesn't matter, but it is required for openshift sdn-> ovn conversion

trozet (approved):
/lgtm

PR #801, opened 2020-09-22
Fixes gateway mode parameters for OVN

trozet:
We have now switched back to using local gateway mode in ovn-kubernetes.
This is currently forced in ovn-k8s with a hack. This patch sets it
correctly in CNO and we will revert the hack in ovn-k8s.

Additionally, this passes the gateway-interface of "none" when
ovs-configuration is not available. This allows us to do upgrade via

https://github.com/openshift/ovn-kubernetes/pull/281

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -532,9 +532,9 @@ spec:
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
           if [ -f /etc/systemd/system/network-online.target.wants/ovs-configuration.service ]; then
-            gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+            gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
-            gateway_mode_flags="--gateway-mode local"
+            gateway_mode_flags="--gateway-mode local --gateway-interface none"
           fi
 
           # start nbctl daemon for caching
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -124,9 +124,9 @@ spec:
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
           if [ -f /etc/systemd/system/network-online.target.wants/ovs-configuration.service ]; then
-            gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+            gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
-            gateway_mode_flags="--gateway-mode local"
+            gateway_mode_flags="--gateway-mode local --gateway-interface none"
           fi
 
           exec /usr/bin/ovnkube --init-node "${K8S_NODE}" \


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -532,9 +532,9 @@ spec:
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
           if [ -f /etc/systemd/system/network-online.target.wants/ovs-configuration.service ]; then
-            gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+            gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
-            gateway_mode_flags="--gateway-mode local"
+            gateway_mode_flags="--gateway-mode local --gateway-interface none"
```
danwinship, line 537: When the docs talk about "If none specified", it means "if you don't specify it", not "if you specify `none`". This will tell it to look for an interface named `none` and fail if it's not found.

danwinship (commented):
So the idea is that in the 4.5 node case, this would make ovn-kubernetes move `eth0` into `br-ex` itself rather than us doing it in `ovs-configuration.service`, but we expect that to work fine for this limited case for the duration of the upgrade?

Considering all the cases:

- 4.5 node, 4.5 CNO, 4.5 OVS - obviously that works
- ~4.5 node, 4.6 CNO, 4.5 OVS - CNO will specify `--gateway-mode local` (because the node is 4.5), OVS will use old-style local gateway mode (because it's 4.5), it works~
- ~4.5 node, 4.5 CNO, 4.6 OVS - CNO will specify `--gateway-mode local` (because it's 4.5), OVS will use new-style local gateway mode and manually move `eth0` into `br-ex`, it works~
- 4.5 node, 4.6 CNO, 4.6 OVS - CNO will specify `--gateway-mode local` (because the node is 4.5), OVS will use new-style local gateway mode and manually move `eth0` into `br-ex`, it works
- 4.6 node, either CNO or OVS is 4.5 - this wouldn't work but we don't believe that can happen
- 4.6 node, 4.6 CNO, 4.6 OVS - obviously that works

danwinship:
> * 4.5 node, 4.6 CNO, 4.5 OVS - CNO will specify `--gateway-mode local` (because the node is 4.5), OVS will use old-style local gateway mode (because it's 4.5), it works
> * 4.5 node, 4.5 CNO, 4.6 OVS - CNO will specify `--gateway-mode local` (because it's 4.5), OVS will use new-style local gateway mode and manually move `eth0` into `br-ex`, it works

oh... also, those can't happen. I mean, we could have a cluster with a 4.6 CNO and 4.5 OVN, but the OVN pods in that case would be using the configuration that had been generated by the 4.5 CNO still. And there's no way to have 4.5 CNO and 4.6 OVN at all

bindata/network/ovn-kubernetes/ovnkube-master.yaml
trozet: which docs?

trozet:
> So the idea is that in the 4.5 node case, this would make ovn-kubernetes move `eth0` into `br-ex` itself rather than us doing it in `ovs-configuration.service`, but we expect that to work fine for this limited case for the duration of the upgrade?

No, the idea here is we don't want ovn-kubernetes to move any interface into OVS. in a 4.5-> 4.6 upgrade case, CNO gets upgraded before MCO. Meaning that this will happen:

1. 4.5 Cluster starts upgrade
2. CNO gets upgraded, which rolls out 4.6 ovn-kubernetes pods
3. Normally ovnkube-node starts (with new local gw) will attempt to move the interface cause there is no br-ex. However because "none" was specified as the gateway-interface, it knows not to do that part. So no interface is moved, but local gw will still function.
4. MCO rolls out upgrades per node asynchronously. Eventually the node gets rebooted with machine content change including ovs-configuration service.
5. ovs-configuration service runs on the reboot, before kubelet ever comes up and migrates the interface into br-ex
6. kubelet starts with ovnkube-node, local gw continues to function with "none"
7. CNO refreshes and changes the args to ovnkube-node to be gateway-interface "br-ex" (because ovs-configuration is now present on the machine)
8. ovnkube-node restarts with gateway-interface br-ex, and then sets up the OVN GR and other "shared" topology components

danwinship:
oh, sorry, I missed that there was a linked ovn-kubernetes PR

PR #802, opened 2020-09-22
Add trozet to owners

trozet:

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -7,17 +7,19 @@ reviewers:
   - dcbw
   - JacobTanenbaum
   - juanluisvaladas
-  - pecameron
   - rcarrillocruz
   - squeed
+  - trozet
   - tssurya
 approvers:
   - abhat
   - danwinship
   - dcbw
+  - fepan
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -7,7 +7,7 @@ reviewers:
   - dcbw
   - JacobTanenbaum
   - juanluisvaladas
-  - pecameron
+  - trozet
```
danwinship, line 10: This breaks the sorting...

danwinship (commented):
You should update openshift/sdn too for consistency. Also, the rest of Feng's team was supposed to get added as `reviewers` to these, but then no one did that...

OWNERS
danwinship: also maybe add @fepan as an approver but not a reviewer (for emergency merges when no one else is around?)

trozet: I didn't know it was sorted. Will do.

danwinship:
/hold
(to stop it from retesting again)

PR #803, opened 2020-09-24
Bug 1854306: Set a 30s timeout for kubectl command in ovnkube-node

pliurh:
By default the timeout of 'kubectl get' is infinity. If apiserver don't response,
This command can take ~15min to fail(socket timeout). It's too long. This patch
set a 30s timeout to 'kubectl get'.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -105,7 +105,8 @@ spec:
           iptables -t raw -A OUTPUT -p udp --dport {{.GenevePort}} -j NOTRACK
           retries=0
           while true; do
-            db_ip=$(kubectl get ep -n ${ovn_config_namespace} ovnkube-db -o jsonpath='{.subsets[0].addresses[0].ip}')
+            # TODO: change to use '--request-timeout=30s', if https://github.com/kubernetes/kubernetes/issues/49343 is fixed. 
+            db_ip=$(timeout 30 kubectl get ep -n ${ovn_config_namespace} ovnkube-db -o jsonpath='{.subsets[0].addresses[0].ip}')
             if [[ -n "${db_ip}" ]]; then
               break
             fi


pliurh:
@trozet @danwinship PTAL

danwinship (commented):
So under what circumstances is this a problem, and what exactly is the effect? If the apiserver eventually starts responding then won't the kubectl complete then? And if it doesn't eventually start responding then does it really matter if we bail out or keep waiting?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -105,7 +105,9 @@ spec:
           iptables -t raw -A OUTPUT -p udp --dport {{.GenevePort}} -j NOTRACK
           retries=0
           while true; do
-            db_ip=$(kubectl get ep -n ${ovn_config_namespace} ovnkube-db -o jsonpath='{.subsets[0].addresses[0].ip}')
+            # wait 30s for kubectl get to return
+            # TODO: change to use '--request-timeout=30s', if https://github.com/kubernetes/kubernetes/issues/51952 is fixed.
```
danwinship, line 109: oops, meant to comment here: the first part ("wait 30s for kubectl get to return") doesn't seem necessary (it's pretty obvious from either "timeout 30" or "--request-timeout=30s"). For the second part, that issue was closed as a duplicate of 49343, so link there instead?

pliurh:
> So under what circumstances is this a problem, and what exactly is the effect? If the apiserver eventually starts responding then won't the kubectl complete then? And if it doesn't eventually start responding then does it really matter if we bail out or keep waiting?

It happens during the SDN migration. After MCO trigger reboot, node come back with br-ex created, the ovnkube-node container will hang there for a long time with following logs. After the socket timeout in ~15mins, the pod will be restart, and then can work as expected. I suspect it was caused by the keepalived VIP floating.

```
++ kubectl get ep -n openshift-ovn-kubernetes ovnkube-db -o 'jsonpath={.subsets[0].addresses[0].ip}'
Unable to connect to the server: read tcp 192.168.111.5:40626->192.168.111.5:6443: read: connection timed out
+ db_ip=
```

bindata/network/ovn-kubernetes/ovnkube-node.yaml
pliurh: fixed.

PR #809, opened 2020-09-29
Validate poolBatchPorts number against min and max in configuration.

gryf:


pkg/network/kuryr.go
```diff
@@ -157,6 +157,15 @@ func validateKuryr(conf *operv1.NetworkSpec) []error {
 		octaviaServiceNet = &octaviaServiceNetObj
 	}
 
+	if kc != nil && kc.EnablePortPoolsPrepopulation && kc.PoolBatchPorts != nil && *kc.PoolBatchPorts > 0 {
```
luis5tb, line 160: the prepopulation should be independent of the checkings

pkg/network/kuryr.go
```diff
@@ -157,6 +157,15 @@ func validateKuryr(conf *operv1.NetworkSpec) []error {
...
+		if kc.PoolMinPorts > 0 && *kc.PoolBatchPorts < kc.PoolMinPorts {
```
luis5tb, line 161: we should also warn about batchports to at least be 1, right?

pkg/network/kuryr.go
gryf: Ah, right. I didn't read the description for it carefully enough.

pkg/network/kuryr.go
gryf: Right!

luis5tb:
Looks good to me! Thanks!

MaysaMacedo:
Looks good!

squeed:

bz 1890130

PR #812, opened 2020-09-29
Bug 1883662: Tune sb-db raft cluster election-timer

vishnoianil:
At current scale test (100 nodes, 3000+ services, 15K+ pods), sb-db
cluster is partitioning very frequently with the current default
election time of 5 seconds. This makes the cluster
stable and once it goes to unstable state, it takes longer to recover.
Given the huge number of flows being installed, using any utility tool to
dump the logical flows from sb-db causes the cluster partition with the
current election timer.


vishnoianil:
@jtaleric FYI.

@dcbw @trozet @abhat Minor patch, PTAL, thanks

dcbw:
ovn-step-registry failure is `[sig-api-machinery][Feature:APIServer][Late] kubelet terminates kube-apiserver gracefully [Suite:openshift/conformance/parallel]`

PR #813, opened 2020-09-30
Set image versions of CNO deployment to latest

rcarrillocruz:
This doesn't affect production deploys nor CI runs, as those
versions are replaced to sane ones.
However when using hack/run_locally we need this otherwise
there is a mismatch between 4.3/4.4 code and master (which is what
we usually use for developing)

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_deployment.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment.yaml
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: network-operator
-        image: quay.io/openshift/origin-cluster-network-operator:4.3
+        image: quay.io/openshift/origin-cluster-network-operator:latest
         command:
         - "/usr/bin/cluster-network-operator"
         - "--url-only-kubeconfig=/etc/kubernetes/kubeconfig"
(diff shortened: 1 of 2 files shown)


nerdalert (commented):
FYI I am failing pretty early in the script when I try to build with `hack/ovn-kind-cno.sh`. Still looking at why. Not sure if others are seeing the same. Ty.

```
deployment.apps/network-operator created
clusteroperator.config.openshift.io/network created
deployment.apps/network-operator condition met
tar: ovnkube-node.yaml: Cannot open: File exists
tar: Exiting with failure status due to previous errors
command terminated with exit code 2
```

rcarrillocruz:
let's agree if we want to land this on 4.6 or we wait when master is 4.7

squeed:
this might not work; the last time I checked, mirroring to `latest` was totally broken and had images from `4.1` So do please check the image hashes.

rcarrillocruz:
https://quay.io/repository/openshift/origin-ovn-kubernetes?tab=tags 

Latest matches 4.6 .

Moreover, I had to do this locally in order to get it working.

squeed:
cool, glad that's working again.
/lgtm
/skip ci/prow/e2e-openstack
/skip ci/prow/e2e-metal-ipi

squeed:
You need to update both `image-references` and the CNO deployment.

trozet (approved):
/lgtm

PR #814, opened 2020-09-30
Fix ovn-kind-cno

trozet:
HA now works by specifying NUM_MASTER_NODES as 3. Updates to CNO
versions so we pull the latest versions of the containers.


trozet:
@nerdalert @aojea @danielmellado

aojea:
/lgtm
Thanks

hack/ovn-kind-cno.sh
```diff
@@ -90,7 +99,7 @@ if [ "$BUILD_OVN" = true ]; then
   pushd dist/images
   sudo cp -f ../../go-controller/_output/go/bin/* .
   cat << EOF | docker build -t origin-ovn-kubernetes:dev -f - .
-FROM quay.io/openshift/origin-ovn-kubernetes:4.5
+FROM quay.io/openshift/origin-ovn-kubernetes:4.6
```
abhat, line 102: `quay.io/openshift/origin-ovn-kubernetes:latest` vs. `4.6`?

hack/ovn-kind-cno.sh
```diff
@@ -149,6 +158,12 @@ if [ "$BUILD_OVN" = true ] || [ "$BUILD_CNO" = true ]; then
   popd
 fi
 
+# CNO needs to get scheduled on first node, because thats where KIND will put /etc/kubernetes/kubeconfig
+until kubectl get pod -n openshift-network-operator -o wide | grep "ovn-control-plane ";do
```
abhat, line 162: exit the script if this operation times out after a few retries?

hack/ovn-kind-cno.sh
```diff
@@ -202,17 +217,20 @@ done
 # wait until resources are created
 sleep 30
 
-if ! kubectl wait -n openshift-ovn-kubernetes --for=condition=ready pods --all --timeout=300s ; then
-  echo "OVN-k8s pods are not running"
+# Wait for pods to come up and be ready. Note metrics pods do not currently work.
```
abhat, line 220: is there a bz for tracking why metrics pods are not coming up and excluded from the readiness check?

hack/ovn-kind-cno.sh
aojea: I do not recommend to use latest and better pin the versions, so you always know what you are using and everybody has the same version

manifests/0000_70_cluster-network-operator_03_deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
       - name: network-operator
-        image: quay.io/openshift/origin-cluster-network-operator:4.3
+        image: quay.io/openshift/origin-cluster-network-operator:4.6
```
abhat, line 19: s/4.6/latest

hack/ovn-kind-cno.sh
Billy99: I don't know which is right, but @rcarrillocruz push #813 to move them all to latest.

rcarrillocruz: Using latest is fine. These scripts are used mainly for development, i.e. master, so it's more than likely that we want to test CNO latest with everything latest.
If we set to 4.6 then it's a moving target, every release we need to change it. Even within the same release, there may be times where a breaking change doesn't work against CNO master.

aojea: fair enough,

rcarrillocruz:
As mentioned, I already had up https://github.com/openshift/cluster-network-operator/pull/813/files which tackled the versions of the images in the deployment manifest. So either we land https://github.com/openshift/cluster-network-operator/pull/813/files first and then this just touches the ovn kind script or you change your PR to switch all images to latest and I close my PR.

aojea:
> As mentioned, I already had up https://github.com/openshift/cluster-network-operator/pull/813/files which tackled the versions of the images in the deployment manifest. So either we land https://github.com/openshift/cluster-network-operator/pull/813/files first and then this just touches the ovn kind script or you change your PR to switch all images to latest and I close my PR.

I think ricardos PR is cleaner, since the problem for OVN KIND seems related to the images version and this just adds a new functionality to use HA deployments

hack/ovn-kind-cno.sh
trozet: well this script isn't really part of the product or anything, I don't know that we need a bz for it. But someone can feel free to fix it in a future patch.

trozet:
> > As mentioned, I already had up https://github.com/openshift/cluster-network-operator/pull/813/files which tackled the versions of the images in the deployment manifest. So either we land https://github.com/openshift/cluster-network-operator/pull/813/files first and then this just touches the ovn kind script or you change your PR to switch all images to latest and I close my PR.
> 
> I think ricardos PR is cleaner, since the problem for OVN KIND seems related to the images version and this just adds a new functionality to use HA deployments

I'm not convinced we want to use latest. Mainly because we use branches in CNO for releases. Therefore after 4.6 branch is pulled and we are on 4.7, if you deploy in 4.6 you will get "latest" which isn't right...it should be 4.6. Yes there is more of an upkeep with updating the versions each release, but on the flip side, when you check out a particular release of the repo, you know what you're getting.

hack/ovn-kind-cno.sh
trozet: I'd rather keep versions pinned correctly in appropriate branches:
https://github.com/openshift/cluster-network-operator/pull/814#issuecomment-701551960

hack/ovn-kind-cno.sh
trozet: yeah I can add that to avoid some unforeseen infinite loop

hack/ovn-kind-cno.sh
abhat: Ok, yeah that makes sense. Let's pin the versions for each branch appropriately. it just becomes a maintenance issue over time. does it make sense to create a version file for each release branch and then pick the version as a tag in each of the clients?

trozet:
/hold

going to modify the versions to detect from git revision

aojea:
@trozet please check this PR https://github.com/openshift/cluster-network-operator/pull/936 too, maybe some of the issues have been identified by @juanluisvaladas  already

PR #819, opened 2020-10-02
kube-proxy: drop liveness probe

danwinship:
As near as I can tell, the only effect of the kube-proxy liveness probe is to make bad situations worse, by repeatedly killing kube-proxy when it's already struggling. (eg, https://bugzilla.redhat.com/show_bug.cgi?id=1880680)

We already removed the kube-proxy-based liveness probe from the SDN pods (#385), and the various upstream kube-proxy deployments do not have a liveness probe either.

~Note that if kube-proxy is failing that will already cause metrics alerts to fire, so there is at least visibility if there is a problem.~

/assign @dcbw

### DIFF
--- a/bindata/kube-proxy/alert-rules.yaml
+++ b/bindata/kube-proxy/alert-rules.yaml
@@ -1,0 +1,62 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    prometheus: k8s
+    role: alert-rules
+  annotations:
+    networkoperator.openshift.io/ignore-errors: ""
+  name: networking-rules
+  namespace: openshift-kube-proxy
+spec:
+  groups:
+  - name: cluster-network-operator-kube-proxy.rules
+    rules:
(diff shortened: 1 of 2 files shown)


abhat:
LGTM. Do we need a bz for this?

Also, while this is a low-risk change, do we want to defer it till 4.6.z?

abhat:
> Note that if kube-proxy is failing that will already cause metrics alerts to fire, so there is at least visibility if there is a problem.

What is the trigger for the metric alert? Does it fire on pod restart or when it takes too long to do a restore?

dcbw:
/lgtm

but I think we only have alerts for the SDN, not the standalone kube-proxy.

danwinship:
so, added metrics, copied from openshift-sdn, but we don't really have any good way to test these do we?

We could pretty easily hack openshift-sdn to allow running with an external kube-proxy. It would break multitenant mode and EgressNetworkPolicy, but "normal" stuff should work fine, and then we could have an "e2e-sdn-external-proxy" test to test external kube-proxy...

danwinship:
/hold
working on standalone-kube-proxy testability and feature-completeness first...

PR #821, opened 2020-10-04
Use dbus to detect if OVS is running on host

trozet:
This should be cleaner than trying to check a symlink, and will actually
tell us if OVS is running or not.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -45,6 +45,18 @@ spec:
               --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
               --method org.freedesktop.systemd1.Unit.Restart "replace"
           }
+          # returns true if OVS is running in systemd
+          is_systemd_ovs(){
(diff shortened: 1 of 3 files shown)


bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -45,14 +45,26 @@ spec:
               --object-path /org/freedesktop/systemd1/unit/"$(svc_encode_name ${1})"_2eservice \
               --method org.freedesktop.systemd1.Unit.Restart "replace"
           }
+          # returns true if OVS is running in systemd
+          is_systemd_ovs(){
+            local state=$(gdbus call --system --dest org.freedesktop.systemd1 \
+              --object-path /org/freedesktop/systemd1/unit/openvswitch_2eservice \
+              --method org.freedesktop.DBus.Properties.Get \
+              org.freedesktop.systemd1.Unit ActiveState)
+            if [[ "$state" == *\'active\'* || "$state" == *\'activating\'* ]]; then
```
rbbratta, line 54: shouldn't we match `reloading` as well, which is why we tried `ActiveEnterTimestamp > 0`

abhat:
This is no longer needed since we don't need all the container code handling in 4.7.

PR #824, opened 2020-10-05
Bug 1885165: Fix ovnkube metrics

juanluisvaladas:
With PR #778 (commit 518118b) we changed the service monitor definition,
and the daemonset serviceaccount.
    
With this PR we fix the serverName in the serviceMonitors, the
serviceAccount permissions and the prometheusRule alert expression.


juanluisvaladas:
/retitle [WIP] Bug 1885165: Fix master control plane job name

juanluisvaladas:
Finally managed to deploy a full cluster with this from scratch, verified myself that this works.

PR #827, opened 2020-10-06
Remove operator-sdk dependencies

rcarrillocruz:
We should no longer use operator-sdk, this was long overdue.
This change pulls the few things we use from operator-sdk (namely in cmd/) and remove them from go.mod.
This will help us bump to kubernetes 0.19, as operator-sdk doesn't seem maintained and is pinned on 0.18.2.


rcarrillocruz:
@squeed

abhat:
> _No description provided._

Can you please add a description for the PR?

cmd/cluster-network-operator/k8sutil.go
```diff
@@ -40,7 +40,7 @@ const (
 	ClusterRunMode RunModeType = "cluster"
 )
 
-var log = logf.Log.WithName("k8sutil")
+var logger = logf.Log.WithName("k8sutil")
```
squeed, line 43: Delete this logger; we don't need this fine-grained logging.

cmd/cluster-network-operator/leader.go
```diff
@@ -29,7 +27,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var log = logf.Log.WithName("leader")
+var logger2 = logf.Log.WithName("leader")
```
squeed, line 30: Just call this `logger` and use it for everything.

cmd/cluster-network-operator/main.go
```diff
@@ -80,7 +78,7 @@ func main() {
 	}
 
 	// become leader
-	err = leader.Become(context.TODO(), LOCK_NAME)
+	err = Become(context.TODO(), LOCK_NAME)
```
squeed, line 81: now the naming is confusing; the function should be called `BecomeLeader()` or something like that.

squeed:
few minor naming nits but otherwise fine.

cmd/cluster-network-operator/main.go
rcarrillocruz: ack

cmd/cluster-network-operator/leader.go
rcarrillocruz: ack

cmd/cluster-network-operator/k8sutil.go
rcarrillocruz: ack

juanluisvaladas:
Can you describe what has been done manually and what's automated?
Is there any manual change other than go.mod?

cmd/cluster-network-operator/k8sutil.go
```diff
@@ -72,7 +69,6 @@ func GetOperatorNamespace() (string, error) {
 		return "", err
 	}
 	ns := strings.TrimSpace(string(nsBytes))
-	log.V(1).Info("Found namespace", "Namespace", ns)
```
squeed, line 75: You don't need to get rid of these; they have access to `var logger` in the other file.

rcarrillocruz:
> Can you describe what has been done manually and what's automated?
> Is there any manual change other than go.mod?

Put context on the description, but tl;dr moved the few things used from operator-sdk from vendor to CNO cmd/ folder (which is where we use it).

cmd/cluster-network-operator/constants.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package k8sutil
+package main
```
danwinship, line 15: can you just move whichever constants we care about into the other file?

likewise, drop any bits from the other files that we aren't using

cmd/cluster-network-operator/leader.go
```diff
@@ -29,26 +27,26 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var log = logf.Log.WithName("leader")
+var logger = logf.Log.WithName("leader")
```
danwinship, line 30: rather than using the controller-runtime logging, it would be good to just make this files use `"log"` like the rest of CNO. (I mean, we should make the rest of CNO use klog really, but, this would at least get rid of one logger.) Note that if you do this then `cmd/cluster-network-operator/log.go` can go away

cmd/cluster-network-operator/main.go
```diff
@@ -25,7 +23,7 @@ import (
 func printVersion() {
 	log.Printf("Go Version: %s", runtime.Version())
 	log.Printf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH)
-	log.Printf("operator-sdk Version: %v", sdkVersion.Version)
+	log.Printf("operator-sdk Version: %v", Version)
```
danwinship, line 26: I don't think there's any point in logging this any more

danwinship (commented):
/lgtm cancel

cmd/cluster-network-operator/main.go
rcarrillocruz: yeah, just left it in case *something* was consuming it...

cmd/cluster-network-operator/leader.go
rcarrillocruz: ack

cmd/cluster-network-operator/constants.go
rcarrillocruz: Which other file, you mean any constant within cmd/cluster-network-operator/* onto constants.go ?

danwinship: whichever other file(s) were using the constants from here. (I was thinking when I wrote the comment that you were only copying over leader.go, so that's what I meant...)

maybe it makes sense to keep it as a separate file but in that case maybe the new files should all move to `pkg/util/operatorsdk` or something like that

cmd/cluster-network-operator/k8sutil.go
danwinship: as Casey said, these should be kept, just using the existing logging system. (which is now `"log"` rather than "the logger in the other file")

danwinship (commented):
looks good except the k8sutil logging

cmd/cluster-network-operator/k8sutil.go
rcarrillocruz: Done

danielmellado:
overall /lgtm, but let's check the CI issues

PR #830, opened 2020-10-07
Bug 1886127: 4.5: handle partial downgrade from 4.6

squeed:
In 4.6, openvswitch is executed as a systemd service. This is not the case for 4.5. However, changes are always rolled out to the SDN first, including on rollback.
    
This means we need to handle, even if sub-optimally, the case where we're running the 4.5 daemonset on a 4.6 MCO-configured node. In other words, openvswitch is running as a systemd service.
    
Fixes: 1885848

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -37,6 +37,14 @@ spec:
           chown -R openvswitch:openvswitch /var/run/openvswitch
           chown -R openvswitch:openvswitch /etc/openvswitch
 
+          if [[ -L '/host/etc/systemd/system/network-online.target.wants/ovs-configuration.service' ]]; then
+            echo "openvswitch is running in systemd"
+            # Don't need to worry about restoring flows; this can only change if we've rebooted
+            rm /var/run/openvswitch/flows.sh || true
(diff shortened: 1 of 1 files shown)


squeed:
Okay, this is going to annoy the bugzilla bot, because it is a 4.5 bug that blocks 4.6.
@knobunc how would you feel about manually setting valid-bug in this case?

knobunc:
Set valid-bug directly since this is a downgrade issue.  There is no problem in 4.6, we just need to handle the possibility that 4.5 has ovs on the host before the MCO reboots the node to revert the change.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -37,6 +37,14 @@ spec:
           chown -R openvswitch:openvswitch /var/run/openvswitch
           chown -R openvswitch:openvswitch /etc/openvswitch
 
+          if [[ -L '/host/etc/systemd/system/network-online.target.wants/ovs-configuration.service' ]]; then
```
rbbratta, line 40: using the new file?
` if [ -f /host/var/run/ovs-config-executed ]; then`

knobunc:
/hold

@abhat is making a new pr.

abhat:
Replaced by: #833

PR #834, opened 2020-10-09
Bug 1886834: Make ovn-cert secret mandatory for both master and node

juanluisvaladas:
With commit 518118bb9 we accidentlly made ovn-cert secret mount
optional, this secret is never optional.


juanluisvaladas:
/retitle Bug 1886834: Make ovn-cert secret mandatory for both master and node

juanluisvaladas:
we could override e2e-aws-sdn-single, the change is only in ovn...

PR #835, opened 2020-10-09
Bug 1886855: Make ovn-cert secret mandatory for both master and node

juanluisvaladas:
With commit 518118bb9 we accidentlly made ovn-cert secret mount
optional, this secret is never optional.


juanluisvaladas:
/retitle Bug 1886855: Make ovn-cert secret mandatory for both master and node

abhat:
@knobunc can we pull this in 4.6.0?

dcbw:
Merging as this issue appears to be is an install test blocker

PR #836, opened 2020-10-09
Bug 1880591: Adds missing check for ovs-config-executed

trozet:

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -820,7 +820,7 @@ spec:
           gateway_mode_flags=
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -820,7 +820,7 @@ spec:
           gateway_mode_flags=
           # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
           # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
+          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service && [ -f /host/var/run/ovs-config-executed ]; then
```
pliurh, line 823: I don’t think it’ll impact the SDN migration.  But I doubt if we really need to check the existence of ovs-config-executed here. Since the ovs-configuration.service will always run before the ovn-kube pods. So I think checking `grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service` is sufficient.

squeed:

This CI run is faulty, tracked in bz 1890130

sdodson:
once more

PR #837, opened 2020-10-13
Bug 1886127: [4.5] sdn-ovs: fix liveness probe for downgrade case

squeed:
ovs-appctl doesn't like running in a different pid namespace, so the liveness probe fails when downgrading from 4.6 (which runs it in systemd).

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -139,33 +139,38 @@ spec:
         livenessProbe:
           exec:
             command:
-            # Validate that ovs-vswitchd and ovsdb are running. These three commands check
-            # 1. that the pids in the pid file are up
-            # 2. that ovs-vswitchd is responding to queries
-            # 3. that ovsdb is responding to queries
-            # 4. if br0 is configured, that its management process is responding
+            # Validate that ovs-vswitchd and ovsdb are running. Check that
+            # 1. that ovs-vswitchd is responding to queries
+            # 2. that ovsdb is responding to queries
+            #
+            # Don't bother executing this if we're in systemd, as the liveness
+            # probe isn't helpful.
             - /bin/bash
             - -c
             - |
               #!/bin/bash
+              if [ -f /host/var/run/ovs-config-executed ]; then
+                exit 0
+              fi
               /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null &&
-              /usr/bin/ovs-vsctl -t 5 show > /dev/null &&
-              if /usr/bin/ovs-vsctl -t 5 br-exists br0; then /usr/bin/ovs-ofctl -t 5 -O OpenFlow13 probe br0; else true; fi
+              /usr/bin/ovs-vsctl -t 5 show > /dev/null
           initialDelaySeconds: 15
           periodSeconds: 5
           timeoutSeconds: 21
         readinessProbe:
           exec:
             command:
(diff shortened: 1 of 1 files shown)


squeed:
@knobunc because the bug hierarchy doesn't match what the bot wants, this will probably need manual valid-bug tagging.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -139,33 +139,32 @@ spec:
...
             - /bin/bash
             - -c
             - |
               #!/bin/bash
-              /usr/bin/ovs-appctl -T 5 ofproto/list > /dev/null &&
-              /usr/bin/ovs-vsctl -t 5 show > /dev/null &&
-              if /usr/bin/ovs-vsctl -t 5 br-exists br0; then /usr/bin/ovs-ofctl -t 5 -O OpenFlow13 probe br0; else true; fi
+              PID=$(cat /var/run/openvswitch/ovs-vswitchd.pid)
+              /usr/bin/ovs-appctl -t "/var/run/openvswitch/ovs-vswitchd.${PID}.ctl" -T 5 ofproto/list > /dev/null &&
+              /usr/bin/ovs-vsctl -t 5 show > /dev/null
```
abhat, line 154: we don't care for the br0 to exist anymore? I thought that was needed for liveness to make sure new pods can come up and flows be configured for them?

squeed: that doesn't affect openvswitch readiness though. It's not a useful check, really.

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -139,33 +139,32 @@ spec:
...
-            # 1. that the pids in the pid file are up
-            # 2. that ovs-vswitchd is responding to queries
-            # 3. that ovsdb is responding to queries
-            # 4. if br0 is configured, that its management process is responding
+            # Validate that ovs-vswitchd and ovsdb are running. Check that
+            # 1. that ovs-vswitchd is responding to queries
+            # 2. that ovsdb is responding to queries
+            #
+            # Need the manual target file because ovs-appctl doesn't like
+            # being in a different namespace from ovs-vswitchd.
```
danwinship, line 147: If OVS is running in systemd we shouldn't be liveness-probing it anyway... I'd just do

    if [ -f /host/var/run/ovs-config-executed ]; then
      exit 0
    fi

for both probes

squeed: Right, I thought about that, but I want to halt rollouts if OVS isn't running, regardless of who should be running it.

abhat: But ovs should always be running (systemd has the unit set to "Restart=Always") if its running under systemd. I think that is what @danwinship was referring to.

squeed: And, as we've found, there are lots of cases where we got this wrong. The readiness probe protects us from taking down the cluster if we miss yet another corner case.

I really think we should keep the readiness probe.

danwinship: We should certainly remove the liveness probe anyway

danwinship: (if not the readiness probe)

squeed:
OK, updated.

PR #839, opened 2020-10-16
Use kube-rbac-proxy for standalone kube-proxy metrics

danwinship:
As with openshift-sdn and ovn-kubernetes, we should run the standalone kube-proxy metrics behind an authenticated proxy.

To avoid possible problems with upgrading old clusters, I made it conditional on using the new metrics port.

### DIFF
--- a/bindata/kube-proxy/kube-proxy.yaml
+++ b/bindata/kube-proxy/kube-proxy.yaml
@@ -66,8 +66,6 @@ spec:
         ports:
         - name: healthz
           containerPort: {{.HealthzPort}}
-        - name: metrics
-          containerPort: {{.MetricsPort}}
         livenessProbe:
           httpGet:
             path: /healthz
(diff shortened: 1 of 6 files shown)


squeed:
Can you think of a cleverer way to roll this out such that all clusters get this? There would be a bit of scraping failures as the ServiceMonitor and Daemonset race, but my instinct is that's not much of a problem. (kube-proxy monitoring is nice but not stupendously critical).

danwinship:
I don't really know a lot about monitoring, and I don't have much sense of what is and is not breaking.

squeed:
Given that we don't have any alerts defined for kube-proxy (oops, we should, I'll file a ticket) then I think it's safe to just roll this out and accept a bit of disruption. I've asked the monitoring folks either way.

squeed:
Yeah, TargetDown fires if scrape fails for more than 10 minutes; I think we don't need the affordance for old clusters.

danwinship:
> Given that we don't have any alerts defined for kube-proxy (oops, we should, I'll file a ticket)

#819 adds them but I didn't want to merge that until we had CI for standalone kube-proxy (https://github.com/openshift/release/pull/12502) which is blocked by this because we fail "Prometheus when installed on the cluster should start and expose a secured proxy and unsecured metrics"

danwinship:
@squeed fixed to unconditionally deploy kube-rbac-proxy. I had to keep the possibility of continuing to use the old port though, since we allow the user to explicitly request that port in the operator config and it seems wrong to not use that port in that case? (But cleaned up some other stuff around specifying explicit ports in the kube-proxy config...)

PR #841, opened 2020-10-19
Bug 1891023: Fix ovn rbac proxy init script secret name

juanluisvaladas:
We were pointing to both wrong service name and namespace.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -457,7 +457,7 @@ spec:
                 -s \
                 --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
                 -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
-                "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}/api/v1/namespaces/ovn-kubernetes/services/ovnkube-master" |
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -467,7 +467,7 @@ spec:
             echo $(date -Iseconds) INFO: Failed to get ovnkube-master service from API. Retry "${retries}"/100 1>&2
             sleep 20
           done
-          if [ "${retries}" -ge 20 ]; then
+          if [ "${retries}" -ge 100 ]; then
```
tssurya, line 470: why are we changing the retries from 20 to 100? In sdn we use 20. Anyways let's see, now that the namespace part is fixed we should not be hitting this that often.

juanluisvaladas: Becase we retry 100 times;
          while [[ "${retries}" -lt 100 ]]; do
So we should fatal with the same amount. It should probably be a constant though

juanluisvaladas: Because we retry 100 times:
          while [[ "${retries}" -lt 100 ]]; do

This should have never been 20.

juanluisvaladas:
/retest
/retitle Fix ovn rbac proxy init script secret name
Verified that this is working as intended

juanluisvaladas:
/override e2e-aws-sdn-multi

We're not modifying sdn at all.

PR #843, opened 2020-10-21
Enable OVS hardware offload for OpenShift

zshi-redhat:


zshi-redhat:
API change: https://github.com/openshift/api/pull/766

squeed:
/hold
until the API change is in.

Otherwise, looks reasonable.

pkg/network/ovn_kubernetes.go
```diff
@@ -108,6 +108,22 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVNHybridOverlayVXLANPort"] = ""
 	}
 
+	if c.HardwareOffloadConfig != nil {
+		if c.HardwareOffloadConfig.EnableHardwareOffload != nil {
+			data.Data["OVNEnableHardwareOffload"] = c.HardwareOffloadConfig.EnableHardwareOffload
```
squeed, line 113: this is a `*bool`, so you'll have to handle the nil case.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -72,6 +72,13 @@ spec:
                 exit 0
             }
             trap quit SIGTERM
+
+            # Enable OVS hardware offload
```
pliurh, line 76: From OCP 4.6, ovs runs in systemd instead of the container. So this logic will not be invoked.

pliurh (commented):
We shall use MCO to inject the configuration change to the ovs systemd service.

pliurh:
My thought is that it's unnecessary to involve CNO to enable the hardware offload. We can use the sriov network operator to generate 
1. an 'offload' MachineConfigPool, which users can put selected nodes in; 
2. a MachineConfig which applies the offload configuration to the nodes in that pool. 

In the MachineConfig we can 
1. inject hardware offload config to openswitch systemd service;
2. create a new systemd service to configure the NIC to switchdev mode.

Additionally, the sriov network operator takes care of the firmware setup for the NICs. With such a design, we don't have to add any cluster level API. Users only need to interact with the sriov network operator API, and put nodes in the predefined pool.

zshi-redhat:
> My thought is that it's unnecessary to involve CNO to enable the hardware offload. We can use the sriov network operator to generate
> 
>     1. an 'offload' MachineConfigPool, which users can put selected nodes in;

CNO could also creates such a pool and apply/inject offload config (e.g. hw-offload, tc-policy) to system ovs configuration service.

Would we prefer to using CNO or SRIOV?

> 
>     2. a MachineConfig which applies the offload configuration to the nodes in that pool.
> 
> 
> In the MachineConfig we can
> 
>     1. inject hardware offload config to openswitch systemd service;
> 
>     2. create a new systemd service to configure the NIC to switchdev mode. It will set the numVFs, load/unload driver, set switchdev mode, set NIC name...

This part is about SR-IOV device configuration and in-relevant to system ovs, could be handled totally by sriov operator.

> 
> 
> Additionally, the sriov network operator takes care of the firmware setup for the NICs. With such a design, we don't have to add any cluster level API. Users only need to interact with the sriov network operator API, and put nodes in the predefined pool.

It doesn't sound like OVS tc-policy is a proper API definition for SR-IOV Operator. We may prefer to having it in CNO API.

juanluisvaladas:
I think it would make more sense to enable it on ovn kube node. I think we should also enable it for openshift sdn.

pliurh:
> It doesn't sound like OVS tc-policy is a proper API definition for SR-IOV Operator. We may prefer to having it in CNO API.

If we really want to expose the `tc-policy` flag to users, it's fine to have CNO to have those new APIs and create the MCP and MC. But part of the content of the MC needs to be patched by the SR-IOV operator. As only the SRIOV operator knows which NIC shall be configured.

pliurh:
> I think it would make more sense to enable it on ovn kube node. I think we should also enable it for openshift sdn.

I don't think we have a plan to enable offload for openshift-sdn.

zshi-redhat:
closing as we are adding all the ovs-hw-offload logic inside sriov operator.

PR #844, opened 2020-10-21
Bug 1890130: openshift-sdn: multitenant: join openshift-etcd-operator to etcd

squeed:
It needs to be able to reach etcd, so it needs to be in netns 1.

Fixes: bz 1890130

### DIFF
--- a/bindata/network/openshift-sdn/004-multitenant.yaml
+++ b/bindata/network/openshift-sdn/004-multitenant.yaml
@@ -66,6 +66,7 @@ netname: openshift-user-workload-monitoring
 # Namespaces which are part of the "control plane" and need to reach each other:
 # - kube-system
 # - openshift-etcd
+# - openshift-etcd-operator
 # - openshift-apiserver
 # - openshift-service-catalog-apiserver
 # - openshift-service-catalog-controller-manager
@@ -95,6 +96,15 @@ metadata:
   name: openshift-etcd
 netid: 1
 netname: openshift-etcd
+
+---
+apiVersion: network.openshift.io/v1
+kind: NetNamespace
+metadata:
+  name: openshift-etcd-operator
+netid: 1
+netname: openshift-etcd-operator
+
 
 ---
 apiVersion: network.openshift.io/v1


squeed:
/skip ci/prow/e2e-gcp-ovn

no ovn changes.

squeed:
well, the multitenant cluster has come up and seems mostly fine, but we might have a bit more whack-a-mole to find connection issues.

squeed:
looks like one node went unreachable - investigating (though this should have had no effect)

/test e2e-aws-sdn-multi

squeed:
update: no real useful information; of course the node's logs are inaccessible. It's a worker node, so definitely not affected by this change.

squeed:
argh, it looks like load took down a node:
```
RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = Kubelet may be retrying requests that are timing out in CRI-O due to system load
```

I think this will go in, eventually.

juanluisvaladas:
Thje pod predicate nodeports error also happens in networkpolicy, we can ignore this.
https://search.ci.openshift.org/?search=Pod+Predicate+NodePorts+failed&maxAge=48h&context=1&type=bug%2Bjunit&name=&maxMatches=5&maxBytes=20971520&groupBy=job

squeed:
ordinarily I would just `/skip` this, but given that this is literally the CI job I'm trying to fix...

squeed:
still crio creation timing out; this is really weird. Again, it doesn't immediately seem like a problem caused by my change.

squeed:
/hold
something is definitely wrong; bumping osdn verbosity.

danwinship:
> /skip ci/prow/e2e-gcp-ovn

(the word you're looking for is `/override`)

squeed:
An update: I think there's a deadlock in openshift-sdn in multitenant mode. This PR exposes it, but I can't fix it in this repository.

I think we should force-merge this, so the test results are meaningful

squeed:
It looks like multitenant mode is stuck on an sdn deadlock. Given that this change doesn't break anything, and gets us closer to fixing openshift-sdn multitenant, I'm going to override the job so we can make progress.

/override ci/prow/e2e-gcp-ovn

PR #852, opened 2020-10-27
Add EgressRouter type and CRD

tssurya:
Part of [SDN-1249](https://issues.redhat.com/browse/SDN-1249)

This CRD allows the user to specify the required parameters for
the egress-router feature. Using these fields, the controller
in CNO for egress-router will create a NAD [pod & service] which
will be absorbed by the egress-router-cni plugin.

Sample CR yaml:
```
apiVersion: network.operator.openshift.io/v1
kind: EgressRouter
metadata:
  name: egress-router-5
spec:
  addresses: [
    {
      ip: "192.168.3.10/24",
      gateway: "192.168.3.1",
    },
  ]
  mode: Redirect
  redirect: {
    redirectRules: [
      {
        destinationIP: "203.0.113.25/30",
        port: 65,
        protocol: TCP,
        targetPort: 75
      },
      {
        destinationIP: "203.0.113.26/30",
        port: 65535,
        protocol: SCTP,
      },
      {
        destinationIP: "203.0.113.27/30",
        port: 65534,
        protocol: UDP,
      },
    ],
    fallbackIP: "203.0.113.28/30"
  }
  networkInterface: {}
```
Output:
```
oc create -f manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
customresourcedefinition.apiextensions.k8s.io/egressrouters.network.operator.openshift.io created

 oc create -f egress_router.yaml 
egressrouter.network.operator.openshift.io/egress-router-1111 created

```

```
oc describe egressrouter egress-router-5
Name:         egress-router-5
Namespace:    default
Labels:       <none>
Annotations:  <none>
API Version:  network.operator.openshift.io/v1
Kind:         EgressRouter
Metadata:
  Creation Timestamp:  2021-02-22T19:34:13Z
  Generation:          1
  Managed Fields:
    API Version:  network.operator.openshift.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:spec:
        .:
        f:addresses:
        f:mode:
        f:networkInterface:
          .:
          f:macvlan:
            .:
            f:mode:
        f:redirect:
          .:
          f:fallbackIP:
          f:redirectRules:
    Manager:         kubectl-create
    Operation:       Update
    Time:            2021-02-22T19:34:13Z
  Resource Version:  37818
  Self Link:         /apis/network.operator.openshift.io/v1/namespaces/default/egressrouters/egress-router-5
  UID:               c5ef79f2-6000-42bc-ba03-9f7a53c7650a
Spec:
  Addresses:
    Gateway:  192.168.3.1
    Ip:       192.168.3.10/24
  Mode:       Redirect
  Network Interface:
    Macvlan:
      Mode:  Bridge
  Redirect:
    Fallback IP:  203.0.113.28/30
    Redirect Rules:
      Destination IP:  203.0.113.25/30
      Port:            65
      Protocol:        TCP
      Target Port:     75
      Destination IP:  203.0.113.26/30
      Port:            65535
      Protocol:        SCTP
      Destination IP:  203.0.113.27/30
      Port:            65534
      Protocol:        UDP
Events:                <none>
```


squeed:
Question

squeed:
/cc @danielmellado 

This is a CRD that the network operator itself will be reacting to, right? In other words, there is a control loop in the CNO that creates things (e.g. services, NADs) in respons to an EgressRouter CR? If so, then this CRD should be in `manifests`.

The model you're following is OperatorPKI

squeed:
Actually wait. This is public API, right? Then it might need to go in to `openshift/api`. Let me think

squeed:
Hmm, no, I don't think we need to put this in openshift/api. Sorry for the noise!

tssurya:
> /cc @danielmellado
> 
> This is a CRD that the network operator itself will be reacting to, right? In other words, there is a control loop in the CNO that creates things (e.g. services, NADs) in respons to an EgressRouter CR?

yes that is the intention from the last discussion I had with Daniel based on what you told him.

 > If so, then this CRD should be in `manifests`.

oops, I should probably move this into there then. Will update this after I test this in the setup and add all the missing pieces.

squeed:
I just realized this is an OVN type, not a CNO type. That is a bit more complicated, then. Perhaps this belongs in `bindata` after all.

tssurya:
> I just realized this is an OVN type, not a CNO type. That is a bit more complicated, then. Perhaps this belongs in `bindata` after all.

well I am also confused on what should go into manifests versus bindata, trying to look into the reconciliation code part to figure this out. I basically followed suit to egress IP and firewall and put it in the same file :D

bindata/network/ovn-kubernetes/001-crd.yaml
```diff
@@ -262,3 +262,195 @@ status:
...
+              description: Specification of the desired behavior of EgressRouter.
+              properties:
+                egressRouters:
+                  description: EgressRouters is the list of egress router pods requested.
+                  items:
+                    description: EgressRouter is a single egressrouter pod configuration object.
+                      Each egress router pod must have exactly one of the following three
+                      fields defined; `ip`, `podIP` or `ipConfig`.
+                    properties:
+                      name:
```
tssurya, line 314: `name` is the name of the NAD. The pod name will be the same

tssurya: add `type`: `egress-router`

bindata/network/ovn-kubernetes/001-crd.yaml
```diff
@@ -262,3 +262,195 @@ status:
...
+                        description: The name of the network
+                        type: string
+                      interfaceType:
+                        description: Type of interface to create/use. e.g; ipvlan, macvlan
```
tssurya, line 318: let's focus on macvlan

bindata/network/ovn-kubernetes/001-crd.yaml
```diff
@@ -262,3 +262,195 @@ status:
...
+                                If not provided, the pod can connect to any destination.
+                              items:
+                                type: string
+                              type: array
+                          required:
+                          - name
+                          - addresses
+                          type: object
+                        type: array
+                      ipConfig:
```
tssurya, line 379: remove `ipConfig` and `podIP` for now. Daniel is focussing on `ip` for the beginning. We can revisit this.

abhat:
is there a types.go that defines this CRD's type somewhere?

danielmellado:
> is there a types.go that defines this CRD's type somewhere?

if you refer to the controller, nope, it's not there yet, but I'm working on it ;)

tssurya:
> is there a types.go that defines this CRD's type somewhere?

hmm good point, you mean something like this: https://github.com/ovn-org/ovn-kubernetes/blob/81c3416d25b0a92ce2ec4e0e48d5bfeeb90fe906/go-controller/pkg/crd/egressfirewall/v1/types.go ? Need to look into this

bindata/network/ovn-kubernetes/001-crd.yaml
```diff
@@ -262,3 +262,195 @@ status:
...
   conditions: []
   storedVersions: []
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  creationTimestamp: null
+  name: egressrouters.k8s.ovn.org
+spec:
+    group: k8s.ovn.org
```
squeed, line 272: so @danielmellado and I were chatting, and I think the group should probably be `network.operator.openshift.io`

If that's the case, you wouldn't write the CRD directly. Rather, you'd create a struct in `pkg/apis/...`, and `hack/update-codegen.sh` would generate the CRD for you.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -15,41 +13,23 @@ spec:
...
-        CA certificate will have a CommonName of \"<namespace>_<name>-ca@<timestamp>\",
-        where <timestamp> is the last rotation time."
+      description: "OperatorPKI is a simple certificate authority. It is not intended for external use - rather, it is internal to the network operator. The CNO creates a CA and a certificate signed by that CA. The certificate has both ClientAuth and ServerAuth extended usages enabled. \n  More specifically, given an OperatorPKI with <name>, the CNO will manage: - A Secret called <name>-ca with two data keys:   - tls.key - the private key   - tls.crt - the CA certificate - A ConfigMap called <name>-ca with a single data key:   - cabundle.crt - the CA certificate(s) - A Secret called <name>-cert with two data keys:   - tls.key - the private key   - tls.crt - the certificate, signed by the CA \n The CA certificate will have a validity of 10 years, rotated after 9. The target certificate will have a validity of 6 months, rotated after 3 \n The CA certificate will have a CommonName of \"<namespace>_<name>-ca@<timestamp>\", where <timestamp> is the last rotation time."
```
tssurya, line 18: gah, wtf, I dunno why this came alone. TODO: remove this file change.

bindata/network/ovn-kubernetes/001-crd.yaml
tssurya: @squeed : thanks, I've updated the patch accordingly. Just need to clean some pki stuff into another patch :D.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// EgressRouter is a feature allowing the user to define an egress router
+// that acts as a bridge between pods and external systems. The egress router runs
+// a service that redirects egress traffic originating from a pod or a group of
+// pods to a remote external system or multiple destinations as per configuration.
+
+//  More specifically, given an EgressRouter with <name>, the CNO will create and manage:
+// - A service called <name>
+// - An egress pod called <name>-ca
```
tssurya, line 14: remove `-ca`

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
tssurya: done in PR #868

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default=egress-router
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
+
+	// Type of interface to create/use. e.g; ipvlan, macvlan
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Optional
+	// +optional
+	InterfaceType string `json:"interfaceType,omitempty"`
```
tssurya, line 52: should I make macvlan the default ?

danielmellado: Sure, let's go with macvlan as default (it's the default one on the cni plugin anyways!). Thanks!

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+// - A service called <name>
+// - An egress pod called <name>
+// - An NAD called <name>
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// EgressRouter is a single egressrouter pod configuration object.
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=egressrouters,scope=Namespaced
+type EgressRouter struct {
```
tssurya, line 22: add // +kubebuilder:printcolumn:

hack/update-codegen.sh
```diff
@@ -35,3 +35,5 @@ HEADER="# This file is automatically generated. DO NOT EDIT"
 # Add a new CRD? Duplicate these lines
 echo "${HEADER}" > manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
 oc annotate --local -o yaml "${RELEASE_PROFILE}" -f _output/crds/network.operator.openshift.io_operatorpkis.yaml >> manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
+echo "${HEADER}" > manifests/0000_70_cluster-network-operator_05_egr_crd.yaml
+oc annotate --local -o yaml "${RELEASE_PROFILE}" -f _output/crds/network.operator.openshift.io_egressrouters.yaml >> manifests/0000_70_cluster-network-operator_05_egr_crd.yaml
```
squeed, line 39: the final file should be called `0000_70_cluster-network-operator_01_...` - we want to create the CRD before the operator deployment.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +kubebuilder:validation:Required
+	Spec EgressRouterSpec `json:"spec"`
+
+	// +kubebuilder:validation:Optional
+	// +optional
+	Status []EgressRouterStatus `json:"status,omitempty"`
```
squeed, line 31: This shouldn't be an array.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// EgressRouterSpec contains the configuration for an egress router.
+// +k8s:openapi-gen=true
+// +kubebuilder:validation:Required
+type EgressRouterSpec struct {
+	// The name of the network attachment definition
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
```
squeed, line 41: Do we need this? Or can we just use the name of the CR itself?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:default=egress-router
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
```
squeed, line 46: What does Type mean? What are valid values?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Type of interface to create/use. e.g; ipvlan, macvlan
```
squeed, line 48: Is `type` the CNI plugin that is used, or would you like to leave a layer of indirection?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// IPConfigObject contains the configuration for the ip arguments
+type IPConfigObject struct {
+	// List of IP addresses to configure on the interface.
+	// +kubebuilder:validation:Required
+	Addresses []IPAddress `json:"addresses"`
+
+	// IP address of the next-hop gateway, if it cannot be automatically determined.
+	// +kubebuilder:validation:Pattern=`^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$`
```
squeed, line 82: Does this validation pass ipv6 :-)

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// +kubebuilder:validation:Required
+	Type string `json:"type"`
+
+	// Type of interface to create/use. e.g; ipvlan, macvlan
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Optional
+	// +optional
+	InterfaceType string `json:"interfaceType,omitempty"`
+
+	// Arguments specific to the interfaceType
```
squeed, line 54: Can you give an example of this?

hack/update-codegen.sh
tssurya: oh heh thanks for teaching me what the numbering is for XD. I thought its the order of adding new crds or something (feels stupid now that I think about it).

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// +kubebuilder:validation:Required
+	EgressRouter string `json:"egressRouter,omitempty"`
+
+	// Observed status of the egress router pod
+	// +kubebuilder:validation:Required
+	Status string `json:"status,omitempty"`
+}
+
+// IPConfigObject contains the configuration for the ip arguments
+type IPConfigObject struct {
```
squeed, line 76: The convention in kubernetes APIs is not to use the word `Object` as a suffix. So the type would just be called `IPConfig`. This holds for all of the types here.

pkg/apis/network/v1/egressrouter_types.go
tssurya: its supposed to be fixed to the default literal: https://github.com/openshift/egress-router-cni/blob/master/README.md#network-configuration-reference

pkg/apis/network/v1/egressrouter_types.go
tssurya: ack will change this.

pkg/apis/network/v1/egressrouter_types.go
tssurya: It's actually https://github.com/openshift/cluster-network-operator/pull/852/files/0f00146c02e1d87171d53892ff4f73d9c96ac4a7#diff-f01bf54a5d3bc5ccf27f44954fbfe2124577e4564082f801bf967209983b8f7dR101 - inspired of off https://github.com/openshift/egress-router-cni/blob/master/README.md#network-configuration-reference. Truth be told, I don't know if we need this @danielmellado : could you comment on if we need the interfaceargs field for macvlan? Do we really need the mode/master fields

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	Gateway string `json:"gateway"`
+
+	// List of CIDR blocks that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Destinations []DestinationCIDR `json:"destinations"`
+}
+
+// IPAddress is the address to configure on the router's interface.
+// +kubebuilder:validation:Pattern=`^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$`
```
squeed, line 94: This looks suspiciously like a v4 only pattern :-)

pkg/apis/network/v1/egressrouter_types.go
tssurya: it won't :D . good catch . I admit I shamelessly copied it from openshift/api repo. Lemme check the portions for ipv6 validation

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// Observed status of the egress router pod
+	// +kubebuilder:validation:Required
+	Status string `json:"status,omitempty"`
+}
+
+// IPConfigObject contains the configuration for the ip arguments
+type IPConfigObject struct {
+	// List of IP addresses to configure on the interface.
+	// +kubebuilder:validation:Required
+	Addresses []IPAddress `json:"addresses"`
```
squeed, line 79: Does this need a netmask? If so, convention is to use CIDR notation (e.g. `address/mask`).

pkg/apis/network/v1/egressrouter_types.go
tssurya: I think we can use the name of the CR itself, name of CR == name of NAD == name of router == name of pod == name of svc. @danielmellado : I think we talked about this before and you were not particular on this right ? So I guess we can safely remove this field and stick to using the name of the CRD?

pkg/apis/network/v1/egressrouter_types.go
tssurya: hmm let me read more on this and fix it

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// IP address of the next-hop gateway, if it cannot be automatically determined.
+	// +kubebuilder:validation:Pattern=`^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$`
+	// +kubebuilder:validation:Optional
+	// +optional
+	Gateway string `json:"gateway"`
```
squeed, line 85: If you support dual stack, you will need multiple gateways.

Usually, gateway and cidr are tied together (though this doesn't exactly match the kernel's understanding of this) So, I'd suggest:

```go
type Spec struct {
    ...
    Addresses []EgressRouterAddress
    AllowedDestinations []CIDR
}
type EgressRouterAddress {
    IP CIDR
    Gateway string
}

// (we might already have a JSON-serializable cidr type lying around somewhere)
type CIDR string
```

pkg/apis/network/v1/egressrouter_types.go
danielmellado: works for me in that part, I mean, hopefully, both, NAD, svc and pod would be hidden from the user anyway.

pkg/apis/network/v1/egressrouter_types.go
danielmellado: I swear I read this at some point, lol

pkg/apis/network/v1/egressrouter_types.go
danielmellado: yep, IPv4 again.

pkg/apis/network/v1/egressrouter_types.go
danielmellado: good catch there casey!

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,124 @@
...
+	// Type of interface to create/use. e.g; ipvlan, macvlan
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Optional
+	// +optional
+	InterfaceType string `json:"interfaceType,omitempty"`
+
+	// Arguments specific to the interfaceType
+	// +kubebuilder:validation:Optional
+	// +optional
+	InterfaceArgs InterfaceArgsObject `json:"interfaceArgs,omitempty"`
```
squeed, line 57: Kubernetes has a different convention for "expanded args". For example, check out how volumes are configured.

Basically, you want to do something like this:

```
struct Spec {
    // some global configuration
    Foo string

    NetworkInterface EgressRouterInterface
}

type EgressRouterInterface struct {
    MacVLAN XVLANConfig
    IPVLAN XVLANConfig
}

type XVLANConfig {
    Mode string
    Master string
}
```

which results in the following yaml:
```yaml
spec:
    foo: bar
    networkInterface:
        macvian: {}
```

pkg/apis/network/v1/egressrouter_types.go
tssurya: err yes sample yaml:

```
apiVersion: network.operator.openshift.io/v1
kind: EgressRouter
metadata:
  name: egress-router
spec:
  egressRouters: [
    {
      name: "egress-router-1",
      type: "egress-router",
      ip: {
        addresses: ["192.168.3.10/24"],
        gateway: "192.168.3.1",
        destinations: ["10.1.2.3/32"]

      },
      interfaceType: "macvlan",
      interfaceArgs: {
        mode: "blah"
      },
    },
  ]
```

pkg/apis/network/v1/egressrouter_types.go
tssurya: super useful, thanks so much,!

pkg/apis/network/v1/egressrouter_types.go
tssurya: cool, ok so I'm removing `spec.name`.
Consensus: to use the CR's name.

pkg/apis/network/v1/egressrouter_types.go
tssurya: spoke with @danielmellado on slack. was agreed since this is always going to be a constant we don't need it.

pkg/apis/network/v1/egressrouter_types.go
tssurya: yea it was supposed to stand for the CNI plugin that is used, but since that is always going to be egress-router-cni, we don't have to have that field. So removing that for now.

pkg/apis/network/v1/egressrouter_types.go
tssurya: done.

pkg/apis/network/v1/egressrouter_types.go
tssurya: just making a note here, ipvlan is not going to be supported from what I gathered from daniel. So I'll add only macvlan, and then we can add ipvlan later on if there is plans to support it

pkg/apis/network/v1/egressrouter_types.go
tssurya: yea I just reuse the CIDR type field for the destination and addresses fields.

pkg/apis/network/v1/egressrouter_types.go
tssurya: hmm ok, isn't it enough we keep both gateways and cidrs as lists? do we need to tie them?

tssurya: actually I don't mind keeping them together _since_ that makes more sense (each subnet (should)get its a gateway).

pkg/apis/network/v1/egressrouter_types.go
tssurya: ok, so I did some research and nowhere on openshift do we check for ipv6 validity using pattern because it is a HUGE regex. Instead people use the `net` library in the code and parse it. So I'll cut the validation here. @danielmellado : you might need to keep this point in mind when you use the value from these fields (example: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/ovn/egressip.go#L309).

danielmellado: yeah, let's cut this off, actually, the same has happened in some another validations now that I do recall. It's a huge thing to validate both IPv4 and 6. I'll parse it within the controller code so no worries.

pkg/apis/network/v1/egressrouter_types.go
tssurya: oh yea actually I just realised that this should not be a list here, since it marks the status of the single router in the spec, this would automatically become a list in the EgressRouterList kind which is what I wanted.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,121 @@
...
+	// IP is the address to configure on the router's interface. Can be IPv4 or IPv6.
+	// +kubebuilder:validation:Required
+	IP CIDR `json:"ip"`
+	// IP address of the next-hop gateway, if it cannot be automatically determined. Can be IPv4 or IPv6.
+	// +kubebuilder:validation:Optional
+	// +optional
+	Gateway string `json:"gateway,omitempty"`
+}
+
+// CIDR represents an IPv4 or IPV6 CIDR block.
```
squeed, line 99: Probably want to add a simple pattern matcher here - it won't be perfect (a proper one is like 4kb) but something simple would be good. It should be dual-stack :-).

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,121 @@
...
+type CIDR string
+
+// EgressRouterStatus contains the observed status of EgressRouter. Read-only.
+type EgressRouterStatus struct {
+	// Name of the egress router network
+	// +kubebuilder:validation:Required
+	EgressRouter string `json:"egressRouter,omitempty"`
```
squeed, line 106: Is this necessary?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Observed status of the egress router pod
+	// +kubebuilder:validation:Required
+	Status string `json:"status,omitempty"`
```
squeed, line 110: Might as well have the status follow the usual standard of a []Conditions array - with conditions of Deploying, Available, Degraded, etc.

squeed:
One minor suggestion, then this looks fine to me
ping @danwinship since he is good at APIs.

pkg/apis/network/v1/egressrouter_types.go
tssurya: hmm I guess not.. I keep getting confused with this list of routers versus single router.
okay so the EgressRouterStatus object will tell the status of the of the corresponding router in the EgressRouter object. So technically this field is not needed, however I added it for displaying it as a part of the CLI output since we decided not to have the `name` field in the main object and to pick it from the CR's metadata: https://github.com/openshift/cluster-network-operator/pull/852/files#diff-9f879d82ac7bec3c4bd2277d645b009b8c6ca77f034e90d1524a026f9dfe2272R22-R28

We have something similar for egressIPs and egressFirewall so thought of keeping the same format.

pkg/apis/network/v1/egressrouter_types.go
tssurya: yea I removed the pattern because the actual pattern for ipv6 was too huge, instead daniel has agreed to manually add the check in the admission controller.

pkg/apis/network/v1/egressrouter_types.go
tssurya: ack, I'll update the patch to change to this.

tssurya: done.

juanluisvaladas (commented):
I think we should consider this severe enough to require changes before merging, if we release it and later make the API more restrictive, is it technically breaking the API even if previously allowed values wouldn't work?

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,142 @@
...
+                description: Specification of interface to create/use. e.g; macvlan
+                properties:
+                  macVLAN:
+                    description: Arguments specific to the interfaceType macvlan
+                    properties:
+                      master:
+                        description: Name of the master inteface. Usually used for macvlan and ipvlan. Need not be specified if it can be inferred from the IP address.
+                        minLength: 1
+                        type: string
+                      mode:
```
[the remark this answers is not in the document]
tssurya: this is a very very good point thanks for pointing this out, tbh I did not read the code and the only info I had on these fields were from here: https://github.com/openshift/egress-router-cni/blob/master/README.md#interface-types-and-platform-support which is why I just left them as string type.

pkg/apis/network/v1/egressrouter_types.go
tssurya: ok you are right, I don't need this field because Name gets displayed auto on output, my knowledge of k8s api sucks.

 [dsal@bkr-hv02  ~]$ oc get egr  
NAME            EGRESSROUTERS   CONDITION   STATUS 
egress-router

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,151 @@
...
+                properties:
+                  macVLAN:
+                    description: Arguments specific to the interfaceType macvlan
+                    properties:
+                      master:
+                        description: Name of the master inteface. Need not be specified if it can be inferred from the IP address.
+                        minLength: 1
+                        type: string
+                      mode:
+                        description: Mode depicts the mode that is used for the macvlan interface. The default mode is "vepa".
```
juanluisvaladas, line 83: The default mode is bridge. 
https://github.com/openshift/egress-router-cni/blob/master/pkg/macvlan/macvlan.go#L198-L200

Also I would expect the description to include something like:
```
The following modes are allowed:
- Bridge
- Private
- VEPA (Or Vepa, whatever you decide to do)
- Passthrough
```
The reason why I expect this is because `oc explain` (or `kubectl`) don't show the enum values, you need to document them manually.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,151 @@
...
+                        enum:
+                        - MacvlanModeBridge
```
juanluisvaladas, line 85: This is what the customer writes in the yaml. While `MacvlanModeBridge` is a really good name for a constant in `types.go`, we typically don't add that kind of prefix in the actual value of the string. `Bridge` would be adequate, and the same with the rest.
VEPA is an acronym so I think it should be `VEPA` instead of `Vepa`, but I don't care too much about that detail, either way is gfine by me.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+// +kubebuilder:validation:Required
+type EgressRouterInterface struct {
+	// Arguments specific to the interfaceType macvlan
+	// +kubebuilder:validation:Required
+	MacVLAN MacVLANConfig `json:"macVLAN"`
+}
+
+// MacvlanMode defines the different types of modes that are supported for the macvlan interface.
+// source: https://man7.org/linux/man-pages/man8/ip-link.8.html
+type MacvlanMode string
```
juanluisvaladas, line 60: This has to be updated

juanluisvaladas (changes requested):

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
tssurya: hmm yea actually you are right, will update this.

pkg/apis/network/v1/egressrouter_types.go
juanluisvaladas: actually it's fine, ignore it

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:validation:Required
+	IP IPConfig `json:"ip"`
+}
+
+// EgressRouterInterface contains the configuration of interface to create/use.
+// +kubebuilder:validation:Required
+type EgressRouterInterface struct {
+	// Arguments specific to the interfaceType macvlan
+	// +kubebuilder:validation:Required
+	MacVLAN MacVLANConfig `json:"macVLAN"`
```
danwinship, line 55: `MacVLAN` is wrong, capitalization-wise. It's a VLAN based on MAC addresses, so if "VLAN" is capitalized, "MAC" should be too. But the interface type is usually called just "macvlan" in all lowercase, in which case this should be

    Macvlan MacvlanConfig `json:"macvlan"`

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+// MacVLANConfig consists of arguments specific to the macvlan EgressRouterInterfaceType
+// +kubebuilder:validation:Required
+type MacVLANConfig struct {
+	// Mode depicts the mode that is used for the macvlan interface; one of Bridge|Private|VEPA|Passthru. The default mode is "Bridge".
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=Bridge;Private;VEPA;Passthru
+	// +kubebuilder:default:Bridge
+	Mode MacvlanMode `json:"mode"`
+
+	// Name of the master inteface. Need not be specified if it can be inferred from the IP address.
```
danwinship, line 92:
```suggestion
	// Name of the master interface. Need not be specified if it can be inferred from the IP address.
```

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Optional
+	// +optional
+	Master string `json:"master,omitempty"`
+}
+
+// IPConfig contains the configuration for the ip arguments
```
danwinship, line 99:
```suggestion
// IPConfig contains the configuration for the IP arguments
```

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+	// +kubebuilder:validation:Required
+	Addresses []EgressRouterAddress `json:"addresses"`
+
+	// List of CIDR blocks that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination.
+	// +kubebuilder:validation:Optional
+	// +optional
+	AllowedDestinations []CIDR `json:"allowedDestinations,omitempty"`
+}
+
+// EgressRouterAddress contains a pair of ip cidr and gateway to be configured on the router's interface
```
danwinship, line 112:
```suggestion
// EgressRouterAddress contains a pair of IP CIDR and gateway to be configured on the router's interface
```

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+// +kubebuilder:validation:Required
+type IPConfig struct {
+	// List of IP addresses to configure on the interface.
+	// +kubebuilder:validation:Required
+	Addresses []EgressRouterAddress `json:"addresses"`
+
+	// List of CIDR blocks that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination.
+	// +kubebuilder:validation:Optional
+	// +optional
+	AllowedDestinations []CIDR `json:"allowedDestinations,omitempty"`
```
danwinship, line 109: I don't think `AllowedDestinations` and `IPConfig` necessarily go together. If we plan to extend this to support the egress router HTTP proxy mode, then we'd need a different kind of `AllowedDestinations` with hostnames for that.

danwinship (commented):
Was there an enhancement filed about this? It's hard to say for sure that this is right without knowing how it's going to be processed. Also, this doesn't support everything we currently support with egress routers (eg, HTTP proxy mode). Are we going to support that later? That has implications for how the currently-available features are organized.

/hold

danielmellado:
> Was there an enhancement filed about this? It's hard to say for sure that this is right without knowing how it's going to be processed. Also, this doesn't support everything we currently support with egress routers (eg, HTTP proxy mode). Are we going to support that later? That has implications for how the currently-available features are organized.
> 
> /hold

Hi Dan, this goes tied to the egress-router-cni, which in its original state, it's not even addressing HTTP proxy mode. That's be addressed later, but for now we'd need to merge this tied to my controller patch. In any case, I'll address an amend to the original EP, as this was later discussed on.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,210 @@
...
+// - A service called <name>
+// - An egress pod called <name>
+// - An NAD called <name>
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// EgressRouter is a single egressrouter pod configuration object.
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=egressrouters,scope=Namespaced
+// +kubebuilder:resource:shortName=egr
```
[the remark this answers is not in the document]
tssurya: oh ok, I just mimic-ed egressip. I'll remove this.

pkg/apis/network/v1/egressrouter_types.go
tssurya: hmm I wasn't aware of the HTTP proxy mode design plan, was not included in the enhancement: https://github.com/openshift/enhancements/blob/master/enhancements/network/egress-router.md. The API portion was roughly based on https://github.com/openshift/egress-router-cni/blob/master/README.md keeping in mind the portions daniel said he plans to implement. I don't know if this particular card/epic of writing a CRD needs a separate enhancement in itself, but in any case it should be included in the main enhancement.

pkg/apis/network/v1/egressrouter_types.go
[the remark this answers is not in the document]
tssurya: > This should not be `Required`; that would imply that macvlan was the only supported type. There's some kubebuilder validation thing you can do on `EgressRouterInterface` to say "at least one subfield must be present", which I think is what is intended here.
> 
yea that is what I intend to do, but AFAIK macvlan _is_ the only supported type ? (from what I gathered from @danielmellado). However I guess in the future we could have more. So I'll try and find out how to do the `make sure you set one of these!!!`

> Also, I don't remember exactly how "required" and "default" interact, but it should be possible for someone to not specify the `macvlan` section at all when creating the object, and just have the defaults get filled in.

yea I'll remove the Required and set default and see how that works.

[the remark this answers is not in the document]
danwinship: If macvlan was really the only supported type then you shouldn't have to specify it explicitly. But I'm pretty sure the plan is still to _eventually_ support ipvlan and cloud-specific types.

danwinship:
> Hi Dan, this goes tied to the egress-router-cni, which in its original state, it's not even addressing HTTP proxy mode.

HTTP proxy mode is a different way of _using_ egress-router-cni, not additional functionality that would be added to egress-router-cni itself. (egress-router-cni is basically a generic way to get a pod with an external macvlan; there are then multiple ways you could use such a pod usefully.) So you didn't have to think about it while implementing egress-router-cni, but we do need to think about it when implementing the EgressRouter CRD.

pkg/apis/network/v1/egressrouter_types.go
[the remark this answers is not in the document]
tssurya: ack, this has been updated now in the PR, the `Required` and `default` don't go together. So I made the `networkInterface` field optional and set the default of that value to `macvlan`. So even if the user doesn't specify an interface, macvlan will get set automatically.

Sample yaml:
```
apiVersion: network.operator.openshift.io/v1
kind: EgressRouter
metadata:
  name: egress-router-11
spec:
  ip: {
    addresses: [
      {
        ip: "192.168.3.10/24",
        gateway: "192.168.3.1",
      },
    ],
    allowedDestinations: ["10.1.2.3/32"],
  }
```

Sample output:

```oc get egressrouter egress-router-11 -oyaml
apiVersion: network.operator.openshift.io/v1
kind: EgressRouter
metadata:
  creationTimestamp: "2020-12-08T15:07:02Z"
  generation: 1
  managedFields:
  - apiVersion: network.operator.openshift.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        .: {}
        f:ip:
          .: {}
          f:addresses: {}
          f:allowedDestinations: {}
        f:networkInterface:
          .: {}
          f:macvlan:
            .: {}
            f:mode: {}
    manager: kubectl-create
    operation: Update
    time: "2020-12-08T15:07:02Z"
  name: egress-router-11
  namespace: default
  resourceVersion: "79686"
  selfLink: /apis/network.operator.openshift.io/v1/namespaces/default/egressrouters/egress-router-11
  uid: 904225be-8006-431b-be0f-7820909deae8
spec:
  ip:
    addresses:
    - gateway: 192.168.3.1
      ip: 192.168.3.10/24
    allowedDestinations:
    - 10.1.2.3/32
  networkInterface:
    macvlan:
      mode: Bridge
```

pkg/apis/network/v1/egressrouter_types.go
tssurya: ok, so we'd have probably either an array of cidrs for the normal router scenario and an array of strings for the proxy mode. hmm I'll try to make this field more generic.

tssurya:
> > Hi Dan, this goes tied to the egress-router-cni, which in its original state, it's not even addressing HTTP proxy mode.
> 
> HTTP proxy mode is a different way of _using_ egress-router-cni, not additional functionality that would be added to egress-router-cni itself. (egress-router-cni is basically a generic way to get a pod with an external macvlan; there are then multiple ways you could use such a pod usefully.) So you didn't have to think about it while implementing egress-router-cni, but we do need to think about it when implementing the EgressRouter CRD.

right, so the `allowedDestinations` field right now is just an array of strings so we can put whatever we want into that and the admission controller /egress-router-cni image will parse the strings and check for validity.

@danwinship Q: Do we really need to move it out of `IPConfig` into its own entity because we could have hostnames and/or ips based on the mode ? Or are we ok with documenting this and keeping it in `IPConfig`. I am ok either ways, just wanted to know what is API-ishly correct.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// EgressRouterDegraded indicates that the router's current state does not match its
+	// desired state over a period of time resulting in a lower quality of service.
+	EgressRouterDegraded EgressRouterStatusConditionType = "Degraded"
+
+	// EgressRouterUpgradeable indicates whether the router is in a state that is safe to upgrade. When status is `False`
+	// administrators should not upgrade their cluster and the message field should contain a human readable description
+	// of what the administrator should do to allow the operator to successfully update.  A missing condition, True,
+	// and Unknown are all treated by the CNO as allowing an upgrade.
+	EgressRouterUpgradeable EgressRouterStatusConditionType = "Upgradeable"
```
danwinship, line 165: "upgradeable" mostly only makes sense at the level of operators. I don't think we need an upgradeable condition here. (If we have use for it later we can add it later...)

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,228 @@
...
+// +kubebuilder:validation:Required
+type IPConfig struct {
+	// List of IP addresses to configure on the pod's secondary interface.
+	// +kubebuilder:validation:Required
+	Addresses []EgressRouterAddress `json:"addresses"`
+
+	// List of destinations that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination.
+	// Examples:
+	// {1} ["<dst-ip-addr/mask>"] (["203.0.113.25/30"])
+	// {2} ["<src-port> <protocol> <dst-ip-addr/mask>", "<src-port> <protocol> <dst-ip-addr/mask> <dst-port>"] (["80 tcp 203.0.113.25/30", "8080 tcp 203.0.113.26 80", "8443 tcp 203.0.113.26/30 443"])
```
danwinship, line 124: So that was fine for the plain text config mode, but if we have a proper JSON config it seems like we should split that out into multiple fields rather than having it all as plain text and then parsing it

danwinship:
> @danwinship Q: Do we really need to move it out of `IPConfig` into its own entity because we could have hostnames and/or ips based on the mode ? Or are we ok with documenting this and keeping it in `IPConfig`. I am ok either ways, just wanted to know what is API-ishly correct.

I don't think the configured IP addresses and the list of blocked destinations are really part of a single idea where they belong witj each other more than they belong with the other fields. So I'd split it out. And yeah, then `IPConfig` could go away.

pkg/apis/network/v1/egressrouter_types.go
tssurya: ack will change this.

pkg/apis/network/v1/egressrouter_types.go
tssurya: hmm ok, I'll do a new struct for the `allowedDestinations`, where by we can have the port/hostname/ip fields separately.

tssurya:
> I don't think the configured IP addresses and the list of blocked destinations are really part of a single idea where they belong witj each other more than they belong with the other fields. So I'd split it out. And yeah, then `IPConfig` could go away.

ack, I'll push an update here and I guess I should also change it in https://github.com/openshift/egress-router-cni/pull/34 to reflect the type.

tssurya:
> I don't think the configured IP addresses and the list of blocked destinations are really part of a single idea where they belong witj each other more than they belong with the other fields. So I'd split it out. And yeah, then `IPConfig` could go away.

ack, I'll push an update here and I guess I should also change it in https://github.com/openshift/egress-router-cni/pull/34 to reflect the type.

danwinship (commented):
(I commented on the CRD rather than the golang source but you can figure it out...)

So, the whole "allowedDestinations" thing is specific to `mode: Redirect`... and the HTTP proxy and DNS proxy modes have totally different configuration. So probably the `allowedDestinations` should really be like `redirectConfig` or something like that (with the expectation that there will later also be `httpProxyConfig` etc, and only one of those would be set at a time, like with `openshiftsdnConfig` and `ovnKubernetesConfig` in the CNO config)?

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,169 @@
...
+                      minimum: 0
+                      type: integer
+                    protocol:
+                      description: Protocol can either be TCP or UDP.
+                      enum:
+                      - TCP
+                      - UDP
+                      type: string
+                    srcPort:
+                      description: SrcPort allows specifying the port number at the source.
```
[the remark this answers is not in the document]
tssurya: > "SrcPort" makes it sound like you're talking about the local port that the client binds to, but that's not what this is supposed to be at all. (I know you just copied this naming from the old docs.)
> 
> The "SrcPort" is the port on the egress router IP, and the "DstPort" is the port on the destination IP. So
> 
> ```
> allowedDestinations:
>   - destinationCIDR: 1.2.3.4/32
>     srcPort: 8080
>     dstPort: 80
>   - destinationCIDR: 5.6.7.8/32
>     srcPort: 8081
>     dstPort: 80
> ```
> 
> means if the user connects to port 8080 on the egress router, it gets forwarded to 1.2.3.4:80, but if they connect to port 8081, it gets forwarded to 5.6.7.8:80.
> 
> Renaming "srcPort" to "port" and "dstPort" to "targetPort" would match the naming used in Services

heh, actually I meant for the `srcPort` to refer to the port on the router... I can see how this is confusing and might mean the port on the client side that in turn connects to the router. I can rename it to `port` and `targetPort` if that makes it clearer.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,169 @@
...
+                  description: AllowedDestination consists of parameters used to define the remote destination that the router pod is allowed to connect to. If not provided, the router pod can connect to any destination.
+                  properties:
+                    destinationCIDR:
+                      description: DestinationCIDR specifies the destination's IPaddress.
+                      type: string
+                    dstPort:
+                      description: DstPort allows specifying the port number at the remote destination.
+                      maximum: 65535
+                      minimum: 0
+                      type: integer
```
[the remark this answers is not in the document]
tssurya: actually all the fields are optional. If no ip address or no allowedDestinations is specified, the router pod should be able to connect to any destination as per 3.11. Or at least that's what I thought. But.... I think I was mistaken considering the whole concept doesn't actually make sense if no destination is specified.

So I have another question. What happens if the user provides:
```
8443   tcp 203.0.113.25/32 443
8443   tcp 203.0.113.26/32 443
```
as input ? Will the router redirect traffic to both remote servers then ?

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,169 @@
...
+                      description: IP is the address to configure on the router's interface. Can be IPv4 or IPv6.
+                      type: string
+                  required:
+                  - ip
+                  type: object
+                type: array
+              allowedDestinations:
+                description: "List of destinations that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination. Examples: {1} [{destinationCIDR: 203.0.113.25/30}] {2} [ { \t    srcPort: 80, \t    protocol: UDP, \t    destinationCIDR: 203.0.113.25/30 }, { \t    srcPort: 8443, \t    protocol: TCP, \t    destinationCIDR: 203.0.113.26/30, \t    dstPort: 443 } ]"
+                items:
+                  description: AllowedDestination consists of parameters used to define the remote destination that the router pod is allowed to connect to. If not provided, the router pod can connect to any destination.
```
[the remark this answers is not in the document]
tssurya: `allowedDestinations` is an array of `allowedDestination`. So no `s` there.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,169 @@
...
+                      type: string
+                    ip:
+                      description: IP is the address to configure on the router's interface. Can be IPv4 or IPv6.
+                      type: string
+                  required:
+                  - ip
+                  type: object
+                type: array
+              allowedDestinations:
+                description: "List of destinations that the pod is allowed to connect to via this interface. If not provided, the pod can connect to any destination. Examples: {1} [{destinationCIDR: 203.0.113.25/30}] {2} [ { \t    srcPort: 80, \t    protocol: UDP, \t    destinationCIDR: 203.0.113.25/30 }, { \t    srcPort: 8443, \t    protocol: TCP, \t    destinationCIDR: 203.0.113.26/30, \t    dstPort: 443 } ]"
```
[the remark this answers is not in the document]
tssurya: ack, will remote them.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,169 @@
...
+                    destinationCIDR:
+                      description: DestinationCIDR specifies the destination's IPaddress.
+                      type: string
+                    dstPort:
+                      description: DstPort allows specifying the port number at the remote destination.
+                      maximum: 65535
+                      minimum: 0
+                      type: integer
+                    protocol:
+                      description: Protocol can either be TCP or UDP.
```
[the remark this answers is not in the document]
tssurya: hmm these things are actually not controllable via the CRD right ? I could set TCP as the default one, but I can't really base the protocol versus port relation through kubebuilder, we would have to validate that in the controller.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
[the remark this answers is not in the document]
tssurya: > The other fields should all be marked as optional, and should document what happens when they aren't specified. (eg, in the case of `dstPort`, if it's not specified then it allows to all ports)

hmm this is tricky, according to 3.11 documentation if `targetPort` is not specified, then we redirect traffic to `port` on the remote host.

but looking at the iptables in the implementation side, it doesn't seem like that, although that's doable.

tssurya:
> (I commented on the CRD rather than the golang source but you can figure it out...)
> 
> So, the whole "allowedDestinations" thing is specific to `mode: Redirect`... and the HTTP proxy and DNS proxy modes have totally different configuration. So probably the `allowedDestinations` should really be like `redirectConfig` or something like that (with the expectation that there will later also be `httpProxyConfig` etc, and only one of those would be set at a time, like with `openshiftsdnConfig` and `ovnKubernetesConfig` in the CNO config)?

hmm, good point, although most of the `dns mode` allowedDestination fields are the same as for `redirect mode` except for the `hostname` which is extra. So I'm worried the `DNSConfig` might be a duplicate of `redirectConfig` (https://docs.openshift.com/container-platform/3.11/admin_guide/managing_networking.html#admin-guide-deploying-an-egress-dns-proxy-pod). However the http proxy mode has totally different configuration, so we will need a new Config for that.

I guess right now we can go ahead with the `redirectConfig` and deal with the DNS/http modes later. 

This is also what we do right now for the macvlan config. I can't find a way to ensure only one is specified at a time though....

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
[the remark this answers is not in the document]
tssurya: ok so I've gone for the `no port without protocol` and `no protocol without port` rules. Also makes it consistent with what was supported in 3.11.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,186 @@
...
+                      type: string
+                  required:
+                  - ip
+                  type: object
+                type: array
+              modeConfig:
+                description: ModeConfig represents router's mode configuration.
+                properties:
+                  allowedDestinationsRedirect:
+                    description: List of destinations that the pod is allowed to connect to in redirect mode. If unspecified, the router will not be configured to connect to specific remote destination. Therefore if the mode is "Redirect" then this field must be set.
```
tssurya, line 62: Q: I wonder what happens if the destination is not set. Need to test how this is handled today, from looking at the code, it looks like it will error out: https://github.com/openshift/egress-router-cni/blob/dadd5efd4ea9c10788feabb04392e868d126a8a6/pkg/macvlan/macvlan.go#L377

tssurya: We might need to add this logic to admission controller, because I am not sure how to handle the relational assignment in kubebuilder.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,267 @@
...
+	IP CIDR `json:"ip"`
+
+	// PortProtocolConfig allows the user to define the port, protocol and targetPort fields for the remote destination.
+	// +kubebuilder:validation:Optional
+	PortProtocolConfig PortProtocolRedirectConfig `json:"portProtocolConfig,omitempty"`
+}
+
+// PortProtocolRedirectConfig defines the port, protocol and targetPort fields for the remote destination in redirect mode.
+type PortProtocolRedirectConfig struct {
+	// Port allows specifying the port number on the router.
```
tssurya, line 156: Note to reviewers: this has been split out of `RedirectConfig` purposefully to ensure more control on the fields. For example these fields have to be optional from up top but once a `port` is specified then we need the `protocol` also to be specified and vice versa and so they both have to  be required fields as well.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
tssurya: update: might need to try one of the `kubebuilder:validation:MinItems`/ `kubebuilder:validation:ExclusiveMinimum` or something. I remember trying something like this for the macvlan config... but couldn't find the right validator...

tssurya: ok so it is https://github.com/openshift/cluster-network-operator/pull/125/files/c15bf3c625a7a7833774696f85f2d1531cc39faa#r269165577 #709 
`minProperties` and `maxProperties`. For some weird reason they don't get detected in the autogeneration. Had to fix it manually. But they work as expected.

`The EgressRouter "egress-router-1115" is invalid: spec.modeConfig: Invalid value: 2: spec.modeConfig in body should have at least 2 properties`

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,275 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EgressRouter is a feature allowing the user to define an egress router
+// that acts as a bridge between pods and external systems. The egress router runs
+// a service that redirects egress traffic originating from a pod or a group of
+// pods to a remote external system or multiple destinations as per configuration.
+
```
squeed, line 11: Minor nit: I think you need a `//` because whitespace between comments is meaningful for go.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,275 @@
...
+	ProtocolTypeUDP ProtocolType = "UDP"
+
+	// ProtocolTypeSCTP refers to the SCTP protocol
+	ProtocolTypeSCTP ProtocolType = "SCTP"
+)
+
+// EgressRouterModeConfig represents a single mode's configuration.
+// Mode must be specified along with exactly one "Config" that matches the mode.
+// Each config consists of parameters used to define the remote destination that the router pod is allowed to connect to for that mode.
+type EgressRouterModeConfig struct {
```
squeed, line 125: You don't need this struct if you don't want. Look at the way Volumes are configured: https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#volume-v1-core

I would just do something like

```go
type EgressRouterSpec struct {
    Mode EgressRouterMode
    Redirect *RedirectConfig
}
type RedirectConfig struct {
    AllowedDestinations []IPDestination
}
etc
```

pkg/apis/network/v1/egressrouter_types.go
tssurya: done.

pkg/apis/network/v1/egressrouter_types.go
tssurya: yea, the reason why I split it out was to make it more modular and then use the `minProperties` and `maxProperties` on it. Since `networkInterface` and `addresses` fields are required fields, I guess we could still validate if one of the config's is provided.

tssurya: https://etherpad.opendev.org/p/CubpCqAkGAadpbruAF5K

tssurya: @danwinship : so casey came up with this idea (etherpad), where we could just combine the single IP destination case on 3.11 with fallbackIP idea for the multiple destinations into a single field called `FallBackIP` and then put the rest into a separate struct and make the `ip, port, protocol` parts `required`. Let me know in case you have any suggestions/changes regarding this approach. We sort of need to get this CRD in asap.

The only difference between this approach and previous one I had was that now we can't accept a format that looks like:
```
destinations: [ipA, ipB, ipC]
```
but then that use case is not supported in 3.11 so should be fine.

squeed: If there's a use case that this config doesn't support, then we should certainly try and support it. I'm not sure what an array if IPs would mean though - that traffic would be spread across them on a per-connection basis?

tssurya: > If there's a use case that this config doesn't support, then we should certainly try and support it. I'm not sure what an array of IPs would mean though - that traffic would be spread across them on a per-connection basis?

Yea I guess? although it doesn't make much sense, because we don't do any control on connection basis on the router end. Based on IP table rule matches it might be sent to both destinations right? I had the same-ish (a bit different) question: https://github.com/openshift/cluster-network-operator/pull/852#discussion_r577546026 on what would happen and if we have such a use case or not.

As far as I can see in 3.11 we don't support this use case, there can only be one line with a single IP (https://github.com/openshift/images/blob/master/egress/router/egress-router.sh#L70-L109)

The reason I had this doubt was because of the lines here: https://github.com/openshift/egress-router-cni/blob/master/README.md#network-configuration-reference in the `destination` field part.

I think we can safely discard this use case unless there is a strong reason that the CU might want to do something like that.

squeed: agreed :+1:

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=egressrouters,scope=Namespaced
+// +kubebuilder:printcolumn:name="Condition",type=string,JSONPath=".status.conditions[*].type"
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.conditions[*].status"
+type EgressRouter struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinProperties:=4
```
squeed, line 29: do you need the `MaxProperties` and `MinProperties`?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+	// +kubebuilder:validation:Minimum:=1
+	Port uint16 `json:"port"`
+
+	// Protocol can be TCP, SCTP or UDP.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum="TCP";"UDP";"SCTP"
+	Protocol ProtocolType `json:"protocol"`
+
+	// TargetPort allows specifying the port number on the remote destination to which the traffic gets redirected to.
+	// If unspecified, the traffic is redirected to the destination IP on all ports.
```
squeed, line 111: "If unspecified, the value from Port is used."

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+	// If redirect rules are specified and no fallback IP is provided, connections on other ports will simply be rejected.
+	// +kubebuilder:validation:Optional
+	FallbackIP CIDR `json:"fallbackIP,omitempty"`
+}
+
+// L4RedirectRule consists of L4parameters used to define a single remote destination in the redirect mode.
+type L4RedirectRule struct {
+	// IP specifies the remote destination's IP address. Can be IPv4 or IPv6.
+	// +kubebuilder:validation:Required
+	DestinationIP CIDR `json:"destinationIP"`
```
squeed, line 97: This isn't a CIDR, it's a single IP address.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+
+	// FallbackIP specifies the remote destination's IP address. Can be IPv4 or IPv6.
+	// If no redirect rules are specified, all traffic from the router are redirected to this IP.
+	// If redirect rules are specified, then any connections on any other port (undefined in the rules) on the router will be redirected to this IP.
+	// If redirect rules are specified and no fallback IP is provided, connections on other ports will simply be rejected.
+	// +kubebuilder:validation:Optional
+	FallbackIP CIDR `json:"fallbackIP,omitempty"`
+}
+
+// L4RedirectRule consists of L4parameters used to define a single remote destination in the redirect mode.
```
squeed, line 93: Wording suggestion:

"L4RedirectRule defines a DNAT redirection from a given port to a destination IP and port."

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+type L4RedirectRule struct {
+	// IP specifies the remote destination's IP address. Can be IPv4 or IPv6.
+	// +kubebuilder:validation:Required
+	DestinationIP CIDR `json:"destinationIP"`
+
+	// Port allows specifying the port number on the router.
```
squeed, line 99: "Port is the port number to which clients should send traffic to be redirected."

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
+package v1
```
sttts, line 1: wrong repo. This must go to openshift/api.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+// - An egress pod called <name>
+// - A NAD called <name>
+//
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+//
+// EgressRouter is a single egressrouter pod configuration object.
+// +k8s:openapi-gen=true
+// +kubebuilder:resource:path=egressrouters,scope=Namespaced
+// +kubebuilder:printcolumn:name="Condition",type=string,JSONPath=".status.conditions[*].type"
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.conditions[*].status"
```
sttts, line 23: There are new tags about API compatibility cc @sanchezl

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+type EgressRouter struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinProperties:=4
+	// +kubebuilder:validation:MaxProperties:=4
```
sttts, line 30: what should this mean?

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EgressRouter is a feature allowing the user to define an egress router
```
sttts, line 7: lower-case. These are docs for JSON, not Golang types

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,273 @@
...
+// +kubebuilder:printcolumn:name="Condition",type=string,JSONPath=".status.conditions[*].type"
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.conditions[*].status"
+type EgressRouter struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinProperties:=4
+	// +kubebuilder:validation:MaxProperties:=4
+	Spec EgressRouterSpec `json:"spec"`
```
sttts, line 31: missing description

pkg/apis/network/v1/egressrouter_types.go
tssurya: yea else users might give 
- redirect and http configs and we won't be able to control that. OR
- they might not provide any mode config and we won't be able to catch that.

pkg/apis/network/v1/egressrouter_types.go
tssurya: not according to 3.11 implementation: https://github.com/openshift/images/blob/master/egress/router/egress-router.sh#L70-L109

but we can change this redirect it to the same port, makes more sense.

sttts:
This is the wrong repo. APIs should go to openshift/api repo in order to have the necessary visibility and to be covered by changes across all core APIs.

/hold

pkg/apis/network/v1/egressrouter_types.go
tssurya: ah forgot to update this from the previous iteration, in reality the CIDR is just a string, but I agree its misleading. will fix this.

pkg/apis/network/v1/egressrouter_types.go
squeed: This is a CRD that the CNO consumes - do we need to put it in openshift/api?

tssurya: there was a discussion on this and as long as we document this properly it should be fine right ? Is there a strict rule for mapping all types to `openshift/api`? We use this types.go to auto-generate the yaml file, hence asking.

pkg/apis/network/v1/egressrouter_types.go
tssurya: it means there should be exactly 4 fields set at a time for this struct in the yaml.

danielmellado:
> This is the wrong repo. APIs should go to openshift/api repo in order to have the necessary visibility and to be covered by changes across all core APIs.
> 
> /hold

@squeed mind commenting on this? I would've loved to be aware of this fact before we got to this stage...

pkg/apis/network/v1/egressrouter_types.go
tssurya: done.

pkg/apis/network/v1/egressrouter_types.go
tssurya: It gave me a "go-lint" suggestion to match it with the struct name when I make it small, that's why the R is in caps.

tssurya:
> This is the wrong repo. APIs should go to openshift/api repo in order to have the necessary visibility and to be covered by changes across all core APIs.
> 
> /hold

moved the api changes to https://github.com/openshift/api/pull/851

PR #853, opened 2020-10-27
Detect Nodes Network MTU on Installations with Kuryr

MaysaMacedo:
same MTU as the host.

### DIFF
--- a/pkg/platform/openstack/networking.go
+++ b/pkg/platform/openstack/networking.go
@@ -6,6 +6,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/routers"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/mtu"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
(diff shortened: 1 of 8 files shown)


MaysaMacedo:
Need  to test with the support added on Kuryr.

pkg/platform/openstack/networking.go
```diff
@@ -114,6 +115,21 @@ func findOpenStackNetwork(client *gophercloud.ServiceClient, tag string) (networ
 	}
 }
 
+// Gets the MTU of a Network
+func getOpenStackNetworkMTU(client *gophercloud.ServiceClient, networkID string) (int, error) {
+	networkMTU := 0
+	type NetworkMTU struct {
+		networks.Network
+		mtu.NetworkMTUExt
+	}
```
gryf, line 124: I'm not sure how it will work with golang - but shouldn't that type definition be rather defined outside of the function?

gryf (commented):
I'm not sure how it is work in golang, but

pkg/platform/openstack/networking.go
MaysaMacedo: The scope of this `struct` is only meant to be for inside the function as that is the only place it's used.I believe there is no need to make it a available at package level.

MaysaMacedo:
> Need to test with the support added on Kuryr.

worked fine.

pkg/platform/openstack/networking.go
gryf: I just hope, the compiled code will not create this struct again, and again while calling this function :)

PR #856, opened 2020-10-28
Add connectivity checker controller

rcarrillocruz:
Add network diagnostics controller to CNO
    
This leverages PodNetworkConnectivityChecks (https://github.com/openshift/enhancements/blob/241bfa6388a2bde7d78c4be7f0479fe99784daed/enhancements/kube-apiserver/stability-point-to-point-network-check.md) to perform network checks periodically.
The checks are:
    
From regular pod to kubeapiserver/openshiftapiserver service and endpoints and LBs.
From regular pod to pods on every node running a hello-openshift service.


rcarrillocruz:
@sanchezl can you please comment how EventRecorder gets initialized on other operators ?

sanchezl (commented):
> can you please comment how EventRecorder gets initialized on other operators ?

We use a framework for most of our operator commands that creates the event.Recorder for us. I don't see a proper event.Recorder in use in any of the existing controllers, but the pki controller does have a [loggingRecorder](https://github.com/openshift/cluster-network-operator/blob/ec56a6df59276d4ec91f1491989a4a4b293e349c/pkg/controller/pki/event_recorder.go#L11) that simply writes to Stdout. You can start with that, or use `events.NewRecorder()` to create a new recorder that actually creates events.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+		operatorClient,
+		operatorcontrolplaneClient,
+		apiextensionsClient,
+		kubeInformersForNamespaces,
+		configInformers,
+		apiextensionsInformers,
+		eventRecorder,
+	)
+
+	go connectivityCheckController.Run(ctx, 1)
```
sanchezl, line 273:
```suggestion
	go connectivityCheckController.Run(ctx, 1)
	apiextensionsInformers.Start(ctx.Done())
	kubeInformersForNamespaces.Start(ctx.Done())
```

sanchezl: Start your informers.

sanchezl:
@rcarrillocruz See https://github.com/rcarrillocruz/cluster-network-operator/pull/1 for some thoughts on the context.

rcarrillocruz:
Running this patch with cluster network operator running locally it gets stuck with:

I1030 10:16:01.494974  252285 base_controller.go:66] Waiting for caches to sync for ConnectivityCheckController

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,280 @@
...
+	eventRecorder := &loggingRecorder{}
+	kubeClient, err := kubernetes.NewForConfig(protoKubeConfig)
+	if err != nil {
+		return err
+	}
+	configClient, err := configv1client.NewForConfig(kubeConfig)
+	if err != nil {
+		return err
+	}
+	operatorClient, _, err := genericoperatorclient.NewClusterScopedOperatorClient(kubeConfig, operatorv1.GroupVersion.WithResource("openshiftapiservers"))
```
sanchezl, line 238:
```suggestion
	operatorClient, dynamicInformers, err := genericoperatorclient.NewClusterScopedOperatorClient(kubeConfig, operatorv1.GroupVersion.WithResource("openshiftapiservers"))
```

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,280 @@
...
+		configInformers,
+		apiextensionsInformers,
+		eventRecorder,
+	)
+
+	go connectivityCheckController.Run(ctx, 1)
+	apiextensionsInformers.Start(ctx.Done())
+	kubeInformersForNamespaces.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
+
```
sanchezl, line 277:
```suggestion
	dynamicInformers.Start(ctx.Done())
```

sanchezl (commented):
Looks like one more informer factory to start.

rcarrillocruz:
We will need to tweak the period each check is run , right now is every second (in library-go)

rcarrillocruz:
I need to templatize the manifests, so the RBAC and check-endpoints is only added in case the flag $check_endpoints is enabled

rcarrillocruz:
I also need to put this on SDN

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -178,6 +178,39 @@ spec:
             command: ["test", "-f", "/etc/cni/net.d/80-openshift-network.conf"]
           initialDelaySeconds: 5
           periodSeconds: 5
+{{- if .EnableConnectivityChecks }}
+      - name: sdn-check-endpoints
+        image: quay.io/openshift/origin-cluster-kube-apiserver-operator:latest
```
squeed, line 183: obviously this needs to be changed to the real image reference when this merges.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -178,6 +178,39 @@ spec:
             command: ["test", "-f", "/etc/cni/net.d/80-openshift-network.conf"]
           initialDelaySeconds: 5
           periodSeconds: 5
+{{- if .EnableConnectivityChecks }}
```
squeed, line 181: This is useful for all network providers, right? It should probably be its own daemonset. It can be in the cluster-network-operator namespace, though.

squeed:
Yeah, it would be good to think about how this could be made more generic - do we even need to care about the underlying network provider?

bindata/network/openshift-sdn/sdn.yaml
rcarrillocruz: For sure

bindata/network/openshift-sdn/sdn.yaml
rcarrillocruz: PodNetworkConnectivityCheck resources test connection from pods managed by the operator to a target endpoint.
Then check-endpoints tool is run in a sidecar container and that checks the resources defined to its pod and updates the status (did it resolve ok, did it connect ok).
The PR right now has as placeholder checks from every ovnkube-node to every kube-apiserver endpoint/svc and internal/external LBs which I tried to discuss on our last team meeting 3 days ago but we didn't have time.
I guess there is value to have these generic tests to key targets like kube/openshift apiserver endpoints/services and LBs but run it within the operator namespace in a pod running a dummy container.
But I also I guess there is value in keeping at least pod checks for every ovnkube-node to ovn DB to complement the dbchecker that checks the DB is healthy and stale raft members.

Thoughts?

@squeed @trozet @danwinship @dcbw @abhat

squeed: Well, we can create multiple `PodNetworkConnectivityCheck` objects, right? Then we should certainly create ones for generic "the cluster network is working" tests.

If you think it's useful for us to probe that ovn-nbdb is serving, then we can create a second resource. However, given that it's host-network, it's probably sufficient to just set a ReadinessProbe...?

rcarrillocruz: For more context:

https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/stability-point-to-point-network-check.md

rcarrillocruz: Yeah, as a matter of fact the PR does create multiple PodNetworkConnectivityChecks object.
The question now is:

From which pod (a generic one non-hostnetwork from CNO namespace) to where.
kubeapiserver endpoints/svc AND openshift-apiserver ?
internal/external LBs are good also?
Anything else?

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -178,6 +178,39 @@ spec:
...
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+        ports:
+          - name: check-endpoints
+            containerPort: 17698
```
danwinship, line 207: Since this is a `hostNetwork` pod, this is a host port. However, this isn't in one of the open ranges (9000-9999, 30000-32767) so it will be blocked between nodes by default....

It seems like possibly in cluster-kube-apiserver-operator, the listening port only gets used for the health checks (from the local node), and the declared `check-endpoints` port is not actually used? In which case since we are not doing health checks on these containers, the port (and the `--listen`) may be unnecessary.

(If we _are_ grabbing a host port, even if just for local health checks, it will need to be registered in https://github.com/openshift/enhancements/blob/master/enhancements/network/host-port-registry.md)

cmd/cluster-network-operator/main.go
```diff
@@ -102,16 +103,29 @@ func main() {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-stopCh
+		cancel()
+	}()
+	defer cancel()
+
+	log.Print("Starting the Cmd.")
+	func() {
+		if err := connectivitycheck.Start(ctx, mgr.GetConfig()); err != nil {
```
danwinship, line 116: Why does this have to be started explicitly from `main()` rather than being started like all the other controllers?

If the problem is that it needs access to a `Context` but we don't currently pass that to the Controller start funcs, then just fix the Controller start func code

manifests/0000_70_cluster-network-operator_01_crd.yaml
```diff
@@ -341,6 +341,11 @@ spec:
                 network support should be disabled. If unset, this property defaults
                 to 'false' and multiple network support is enabled.
               type: boolean
+            enableConnectivityChecks:
```
danwinship, line 344: (this will need to be updated in `openshift/api` as well)

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiextensionsinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+type OVNKubernetesConnectivityCheckController interface {
```
danwinship, line 31: This should not be ovn-kubernetes-specific

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	c := ovnKubernetesConnectivityCheckController{
+		ConnectivityCheckController: connectivitycheckcontroller.NewConnectivityCheckController(
+			"openshift-ovn-kubernetes",
+			operatorClient,
+			operatorcontrolplaneClient,
+			apiextensionsClient,
+			apiextensionsInformers,
+			[]factory.Informer{
+				operatorClient.Informer(),
+				kubeInformersForNamespaces.InformersFor("openshift-ovn-kubernetes").Core().V1().Pods().Informer(),
```
danwinship, line 54: Ideally we should not have hardcoded references to ovn-kubernetes anywhere in the code except `pkg/network/ovn_kubernetes.go`. The preferred way to do something like this is to add annotations to objects in `bindata/network/ovn-kubernetes/` and then act generically on the annotations here. (`pkg/controller/operconfig` can notice when it creates/reconciles an object with annotations and then inform this controller, if that would help)

(Except it sounds like probably this isn't going to involve the ovn-kubernetes namespace at all anyways)

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	// kas default service IP
+	templates = append(templates, c.getTemplatesForKubernetesDefaultServiceCheck(syncContext.Recorder())...)
+	// each kas endpoint IP
+	templates = append(templates, c.getTemplatesForKubernetesServiceEndpointsChecks(syncContext.Recorder())...)
+	// each api load balancer hostname
+	templates = append(templates, c.getTemplatesForApiLoadBalancerChecks(syncContext.Recorder())...)
+
+	pods, err := c.podLister.List(labels.Set{"app": "ovnkube-node"}.AsSelector())
+	if err != nil {
+		syncContext.Recorder().Warningf("EndpointDetectionFailure", "failed to list openshift-apiserver pods: %v", err)
```
danwinship, line 100: error message doesn't match what the code is doing

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+func (c *connectivityCheckTemplateProvider) generate(ctx context.Context, syncContext factory.SyncContext) ([]*v1alpha1.PodNetworkConnectivityCheck, error) {
+	var templates []*v1alpha1.PodNetworkConnectivityCheck
+	// kas service IP
+	templates = append(templates, c.getTemplatesForKubernetesServiceMonitorService(syncContext.Recorder())...)
+	// kas default service IP
+	templates = append(templates, c.getTemplatesForKubernetesDefaultServiceCheck(syncContext.Recorder())...)
+	// each kas endpoint IP
+	templates = append(templates, c.getTemplatesForKubernetesServiceEndpointsChecks(syncContext.Recorder())...)
+	// each api load balancer hostname
+	templates = append(templates, c.getTemplatesForApiLoadBalancerChecks(syncContext.Recorder())...)
```
danwinship, line 96: `...ForAPILoadBalancer...`

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,277 @@
...
+	pods, err := c.podLister.List(labels.Set{"app": "ovnkube-node"}.AsSelector())
+	if err != nil {
+		syncContext.Recorder().Warningf("EndpointDetectionFailure", "failed to list openshift-apiserver pods: %v", err)
+		return nil, nil
+	}
+
+	var checks []*v1alpha1.PodNetworkConnectivityCheck
+	for _, pod := range pods {
+		if len(pod.Spec.NodeName) == 0 {
+			// apiserver pod hasn't been assigned a node yet, skip
```
danwinship, line 107: again, not "apiserver" pods here

pkg/controller/connectivitycheck/connectivity_check_template.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	for _, option := range options {
+		option(check)
+	}
+	return check
+}
+
+// WithTlsClientCert option specifies the name of the secret in the check namespace that
+// contains a tls client certificate (and key) to use when performing the check.
+func WithTlsClientCert(secretName string) func(*v1alpha1.PodNetworkConnectivityCheck) {
```
danwinship, line 31: `WithTLSClientCert`

go.mod
```diff
@@ -16,12 +16,14 @@ require (
 	github.com/mitchellh/reflectwalk v1.0.1 // indirect
 	github.com/onsi/gomega v1.10.1
 	github.com/openshift/api v0.0.0-20200827090112-c05698d102cf
+	github.com/openshift/client-go v0.0.0-20200827190008-3062137373b5
```
danwinship, line 19: can you do the vendor updates in a separate commit?

danwinship (commented):
So is the idea that this will replace the one in cluster-kube-apiserver-operator? Or will we be running connectivity checkers from both components?

> _No description provided._

Please fix that. Every PR should explain exactly what it's doing in its initial comment.

bindata/network/network-diagnostics/000-ns.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
squeed, line 1: extremely minor nit - this should just be in `bindata/network-diagnostics`, IMHO.

bindata/network/network-diagnostics/network-check-tester.yaml
```diff
@@ -0,0 +1,54 @@
...
+    matchLabels:
+      app: network-check-tester
+  template:
+    metadata:
+      labels:
+        app: network-check-tester
+    spec:
+      serviceAccountName: network-check-tester
+      containers:
+      - name: network-check-tester-container
```
squeed, line 18: What does this do? I don't think we can reference `busybox` as an image.

bindata/network/openshift-sdn/sdn.yaml
squeed: Be careful, if you check external LB, we may fall afoul of offline / proxied clusters.

cmd/cluster-network-check-endpoints/main.go
```diff
@@ -0,0 +1,16 @@
+package main
```
squeed, line 1: It looks like you're referencing the image `cluster-kube-apiserver-operator` in the deployment; does this code need to be here?

pkg/cmd/checkendpoints/controller/backoff_recorder.go
```diff
@@ -0,0 +1,241 @@
+package controller
```
squeed, line 1: Is it possible to consume the upstream library-go version of this? What changes would we need? This is a lot of code.

squeed:
If we're running exactly one connectivity check controller; why not just run it in-process in the CNO?

squeed:
/hold
to stop the CI bleeding - feel free to unhold whenever. Oh, my poor inbox.

pkg/cmd/checkendpoints/controller/backoff_recorder.go
rcarrillocruz: oh, i was not aware it was on library-go, just copied over from CNO pki controller, thx!

cmd/cluster-network-check-endpoints/main.go
rcarrillocruz: yeah, that was prior to pulling down the cluster-kube-apiserver-operator code for check-endpoints into CNO, it's a left over.
I guess I will need to decouple this cmd/cluster-network-check-endpoints into its own PR, land it, get it built, so then I can reference it in the deployment, I don't think I can self-reference on CI ?

bindata/network/network-diagnostics/network-check-tester.yaml
rcarrillocruz: Yeah, this was just for testing. So basicly, I need a pod to act as a src to run the connectivity checks so in the PR I have this pod with a busybox container doing nothing and the check-endpoints sidecar containers that's doing the check business logic.

bindata/network/network-diagnostics/000-ns.yaml
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_template.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/cmd/checkendpoints/controller/backoff_recorder.go
rcarrillocruz: hmm, I guess you mean https://github.com/openshift/library-go/blob/master/pkg/operator/events/recorder_logging.go ?
That gives quite ugly logging?
e.g.

2020/11/10 18:27:40 reconciling (/v1, Kind=ServiceAccount) openshift-ovn-kubernetes/ovn-kubernetes-controller
W1110 18:27:40.982403  698742 recorder_logging.go:46] &Event{ObjectMeta:{dummy.1646359beaf46f13  dummy    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []  []},InvolvedObject:ObjectReference{Kind:Pod,Namespace:dummy,Name:dummy,UID:,APIVersion:v1,ResourceVersion:,FieldPath:,},Reason:EndpointDetectionFailure,Message:PodNetworkConnectivityCheck.controlplane.operator.openshift.io/network-check-tester-rcarrillocruz-diag-6-dstf9-worker-b-pgd98-to-openshift-apiserver-service-cluster -n openshift-apiserver: the namespace of the provided object does not match the namespace sent on the request,Source:EventSource{Component:,Host:,},FirstTimestamp:2020-11-10 18:27:40.982251283 +0100 CET m=+27.659866000,LastTimestamp:2020-11-10 18:27:40.982251283 +0100 CET m=+27.659866000,Count:1,Type:Warning,EventTime:0001-01-01 00:00:00 +0000 UTC,Series:nil,Action:,Related:nil,ReportingController:,ReportingInstance:,}
2020/11/10 18:27:41 update was successful

bindata/network/network-diagnostics/network-check-tester.yaml
rcarrillocruz: which image should I use instead for this scenario?

rcarrillocruz:
> If we're running exactly one connectivity check controller; why not just run it in-process in the CNO?

Because the intent is to also run regular pod checks.
The idea is to cover:

Generic pod -> Generic pod in every node
Generic pod -> Generic pod service
Generic pod -> kubepiserver service and endpoints
Generic pod -> openshift-apiserver service and endpoints
Generic pod -> LBs

And well, cos it's better to isolate this new feature on its own namespace with its own Kube objects to not pollute CNO I guess.

bindata/network/network-diagnostics/network-check-tester.yaml
rcarrillocruz: nm, as per our slack chat doesn't make sense to have a dummy container, just run check-endpoints on the pod.

cmd/cluster-network-operator/main.go
rcarrillocruz: Because it's a library-go operator and CNO is controller-runtime, thus the split. Sure, I could 'try' to make the operator play as a controller-runtime controller but that would get messy.
Having said that, this will probably change now since @squeed change for making CNO a library-go controller is in flight and this PR is probably better off rebased when that merges.

manifests/0000_70_cluster-network-operator_01_crd.yaml
rcarrillocruz: yep

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

bindata/network/network-diagnostics/network-check-tester.yaml
rcarrillocruz: done

cmd/cluster-network-check-endpoints/main.go
rcarrillocruz: Filed https://github.com/openshift/cluster-network-operator/pull/872/files to decouple

bindata/network/openshift-sdn/sdn.yaml
rcarrillocruz: https://github.com/openshift/cluster-network-operator/pull/872

rcarrillocruz:
https://github.com/openshift/cluster-network-operator/pull/872 needs to land before this

bindata/network/openshift-sdn/sdn.yaml
rcarrillocruz: per @squeed , doesn't matter to split, thus I closed #872 and just added cluster-network-operator image to deployments.

rcarrillocruz:
It seems I need to decouple check-endpoints after all, the deployment hungs because it pulls CNO image from registry and does not have the check-endpoints binary.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	// each oas endpoint IP
+	templates = append(templates, c.getTemplatesForOpenShiftAPIServerServiceEndpointsChecks(syncContext.Recorder())...)
+	// each api load balancer hostname
+	templates = append(templates, c.getTemplatesForAPILoadBalancerChecks(syncContext.Recorder())...)
+	// generic pod service IP
+	templates = append(templates, c.getTemplatesForGenericPodServiceCheck(syncContext.Recorder())...)
+	// each generic pod endpoint IP
+	templates = append(templates, c.getTemplatesForGenericPodServiceEndpointsChecks(syncContext.Recorder())...)
+
+	pods, err := c.diagnosticsPodLister.List(labels.Set{"app": "network-check-source"}.AsSelector())
```
alexanderConstantinescu, line 120: Are we sure this is consistently done before the render phase and that this call actually returns all pods?

alexanderConstantinescu (commented):
Given that it's no longer a `WIP` I started taking a look. I managed to review about 50 % of it and will have another go later on.

bindata/network-diagnostics/network-check-source.yaml
```diff
@@ -0,0 +1,88 @@
...
+      app: network-check-source
+  template:
+    metadata:
+      labels:
+        app: network-check-source
+    spec:
+      serviceAccountName: network-diagnostics
+      containers:
+      - name: check-endpoints
+        image: quay.io/rcarrillocruz/origin-cluster-network-operator:latest
```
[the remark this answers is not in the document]
rcarrillocruz: oops, done

bindata/network-diagnostics/network-check-source.yaml
```diff
@@ -0,0 +1,88 @@
...
+        imagePullPolicy: IfNotPresent
+        terminationMessagePolicy: FallbackToLogsOnError
+        command:
+          - cluster-network-check-endpoints
+        args:
+          - --listen
+          - 0.0.0.0:17698
+          - --namespace
+          - $(POD_NAMESPACE)
```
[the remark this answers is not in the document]
rcarrillocruz: done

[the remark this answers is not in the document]
rcarrillocruz: This is fine, it's Kubernetes envvars substitution:

https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,41 @@
...
+      matchLabels:
+        app: network-check-target
+  template:
+    metadata:
+      labels:
+        app: network-check-target
+    spec:
+      containers:
+      - image: openshift/hello-openshift
+        imagePullPolicy: Always
```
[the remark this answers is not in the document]
rcarrillocruz: done

bindata/network-diagnostics/001-rbac.yaml
```diff
@@ -0,0 +1,77 @@
...
+  name: network-diagnostics
+rules:
+- apiGroups: [""]
+  resources:
+  - endpoints
+  - namespaces
+  - pods
+  - services
+  - nodes
+  - secrets
```
squeed, line 21: should it really be able to access all secrets across all namespaces? I'd rather not if possible...

bindata/network-diagnostics/network-check-source.yaml
```diff
@@ -0,0 +1,88 @@
...
+      app: network-check-source
+  template:
+    metadata:
+      labels:
+        app: network-check-source
+    spec:
+      serviceAccountName: network-diagnostics
+      containers:
+      - name: check-endpoints
+        image: quay.io/openshift/origin-cluster-network-operator:latest
```
squeed, line 19: this needs to be parametereized.

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: network-check-target
+  namespace: openshift-network-diagnostics
+spec:
+  selector:
+      matchLabels:
+        app: network-check-target
+  template:
```
squeed, line 10: should add a readiness probe

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,41 @@
...
+    metadata:
+      labels:
+        app: network-check-target
+    spec:
+      containers:
+      - image: openshift/hello-openshift
```
squeed, line 16: this needs to be parameterized.

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,41 @@
...
+        imagePullPolicy: IfNotPresent
+        name: network-check-target-container
+        ports:
+        - containerPort: 8080
+          protocol: TCP
+        resources: {}
+        terminationMessagePath: /dev/termination-log
```
squeed, line 23: is this correct? do we write a real termination message? or should we pull it from stderr?

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,41 @@
...
+      serviceAccount: default
+      terminationGracePeriodSeconds: 10
+      priorityClassName: "system-cluster-critical"
```
squeed, line 26: is this really critical? otherwise, it would be good to remove this.

cmd/cluster-network-operator/main.go
```diff
@@ -102,16 +103,29 @@ func main() {
...
+			log.Fatal(err)
```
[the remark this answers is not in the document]
squeed: if the only thing that could cause this to fail is the apiserver being in a bad state, then this is acceptable. if other misconfigurations could block network rollout, then we need to think more carefully.

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: done

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: yeah, will change this. It's some daemonset boilerplate i got from somewhere

bindata/network-diagnostics/001-rbac.yaml
rcarrillocruz: I will get back to you on this, but I had to add that as I was getting issues from library-go connectivity check controller.
Will remove, run it and put a gist the thing I got, can't remember off the top of my head.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	if err != nil {
+		syncContext.Recorder().Warningf("EndpointDetectionFailure", "failed to list network-check-source pods: %v", err)
+		return nil, nil
+	}
+
+	var checks []*v1alpha1.PodNetworkConnectivityCheck
+	for _, pod := range pods {
+		if len(pod.Spec.NodeName) == 0 {
```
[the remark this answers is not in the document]
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	return templates
+}
+
+type endpointInfo struct {
+	hostName string
+	port     string
+	nodeName string
+}
+
+func withTarget(label, nodeName string) func(check *v1alpha1.PodNetworkConnectivityCheck) {
```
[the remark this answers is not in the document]
rcarrillocruz: Why?

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: yeah, it will when they are deployed.
FWIW, there is only one network-check-source.

pkg/controller/connectivitycheck/connectivity_check_controller.go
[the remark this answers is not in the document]
alexanderConstantinescu: Because this is not calling `withTarget` with a label and a node name:

https://github.com/openshift/cluster-network-operator/blob/0542083f9f705c934096215413acc25202ce0650/pkg/controller/connectivitycheck/connectivity_check_controller.go#L220

nor is this:

https://github.com/openshift/cluster-network-operator/blob/0542083f9f705c934096215413acc25202ce0650/pkg/controller/connectivitycheck/connectivity_check_controller.go#L333

and most (almost all) places where it is being called.

pkg/controller/connectivitycheck/connectivity_check_controller.go
alexanderConstantinescu: I understand that there is only one network-check-source. However that will be deployed after the render phase is completed by the CNO, and I don't understand how we guarantee that this line is called _after_ that part so that `pods` is not empty?

pkg/controller/connectivitycheck/connectivity_check_controller.go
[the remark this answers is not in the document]
rcarrillocruz: well, it is here:

https://github.com/openshift/cluster-network-operator/blob/0542083f9f705c934096215413acc25202ce0650/pkg/controller/connectivitycheck/connectivity_check_controller.go#L184

The reason why in those lines you referred are not nodes is because they don't apply to a node as they are services and LBs respectively.

So I take when you mean 'label and nodeName' you really mean nodeName?

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: Ah ok, you said pods, that's why I pointed out it's only *one* pod.
Why it matters if they are deployed after?
This will reconcile as the pod come up, and the controller doesn't update the network checks till it it can perform the check between source and dest.

bindata/network-diagnostics/001-rbac.yaml
```diff
@@ -0,0 +1,77 @@
...
+subjects:
+- kind: ServiceAccount
+  name: network-diagnostics
+  namespace: openshift-network-diagnostics
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: system:openshift:controller:network-diagnostics-check-endpoints
```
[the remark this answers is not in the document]
rcarrillocruz: It's not a typo. I pulled down this code from kube-apiserver operator repo since we will end up owning all this:

https://github.com/openshift/cluster-kube-apiserver-operator/blob/0a69c78cea511a8c8e70b3910c637ffaf2a1dc36/bindata/v4.1.0/kube-apiserver/check-endpoints-rolebinding.yaml

But I don't care about var naming or argument naming bikeshedding, I can change.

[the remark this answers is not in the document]
rcarrillocruz: Actually I take that back, it seems it's a Kubernetes convention:

https://kubernetes.io/docs/reference/access-authn-authz/rbac/

bindata/network-diagnostics/001-rbac.yaml
```diff
@@ -0,0 +1,77 @@
...
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: extension-apiserver-authentication-reader
+subjects:
+  - kind: User
+    name: system:serviceaccount:openshift-network-diagnostics:network-diagnostics
```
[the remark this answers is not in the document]
rcarrillocruz: Kubernetes seems to have the convention of using 'system:serviceaccount:<namespace>:<service account>

bindata/network-diagnostics/001-rbac.yaml
```diff
@@ -0,0 +1,77 @@
...
+- kind: ServiceAccount
+  name: network-diagnostics
+  namespace: openshift-network-diagnostics
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: system:openshift:controller:network-diagnostics-check-endpoints
+  namespace: kube-system
```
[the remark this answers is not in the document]
rcarrillocruz: That's fair, the apiserver folks put the RoleBinding in there.
I'll see if putting it on openshift-network-diagnostics is fine.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,403 @@
...
+	apiextensionsInformers := apiextensionsinformers.NewSharedInformerFactory(apiextensionsClient, 10*time.Minute)
+	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
+		"",
+		"openshift-config",
+		"openshift-config-managed",
+		"openshift-cluster-network-operator",
+		"openshift-network-diagnostics",
+		metav1.NamespaceSystem,
+		"openshift-kube-apiserver",
+		"openshift-apiserver",
```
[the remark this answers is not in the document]
rcarrillocruz: This won't matter, eventually it will and the controller reconciles.

danwinship:
You've noticed the failing test, right?

```
[It] ensure control plane pods do not run in best-effort QoS [Suite:openshift/conformance/parallel]
  github.com/openshift/origin/test/extended/operators/qos.go:20
Nov 19 12:53:44.406: FAIL: 
6 pods found in best-effort QoS:
openshift-network-diagnostics/network-check-target-b4n25 is running in best-effort QoS
openshift-network-diagnostics/network-check-target-hzzjd is running in best-effort QoS
openshift-network-diagnostics/network-check-target-j4sph is running in best-effort QoS
openshift-network-diagnostics/network-check-target-jmqm2 is running in best-effort QoS
openshift-network-diagnostics/network-check-target-n4b2k is running in best-effort QoS
openshift-network-diagnostics/network-check-target-x86g6 is running in best-effort QoS
```

rcarrillocruz:
> You've noticed the failing test, right?
> 
> ```
> [It] ensure control plane pods do not run in best-effort QoS [Suite:openshift/conformance/parallel]
>   github.com/openshift/origin/test/extended/operators/qos.go:20
> Nov 19 12:53:44.406: FAIL: 
> 6 pods found in best-effort QoS:
> openshift-network-diagnostics/network-check-target-b4n25 is running in best-effort QoS
> openshift-network-diagnostics/network-check-target-hzzjd is running in best-effort QoS
> openshift-network-diagnostics/network-check-target-j4sph is running in best-effort QoS
> openshift-network-diagnostics/network-check-target-jmqm2 is running in best-effort QoS
> openshift-network-diagnostics/network-check-target-n4b2k is running in best-effort QoS
> openshift-network-diagnostics/network-check-target-x86g6 is running in best-effort QoS
> ```

Thanks, I think it's due to missing resources body, I added same as Multus daemonset and pushed.

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: done

bindata/network-diagnostics/network-check-source.yaml
rcarrillocruz: done

rcarrillocruz:
https://github.com/openshift/api/pull/790

rcarrillocruz:
The vendorized API changes is just for testing, will revert that as soon as the openshift/api PR lands

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
squeed: right, that's a way of fully-qualifying references to subjects. that's not what you're doing here - just naming a RoleBinding. This should just be a normal name.

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
squeed: Alexander is right - you should qualify this using the normal way,.

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
squeed: Well, this is a RoleBinding, not a ClusterRoleBinding. So the namespace **is** relevant. However, I'm not necessarily sure that you need these rights - where in the code would it make these api calls?

bindata/network-diagnostics/001-rbac.yaml
squeed: Still need to remove this.

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,53 @@
...
+            memory: 150Mi
+        readinessProbe:
+          httpGet:
+            port: 8080
+            path: /
+          initialDelaySeconds: 30
+          timeoutSeconds: 10
+      serviceAccount: default
+      terminationGracePeriodSeconds: 10
+      tolerations:
```
squeed, line 39: I don't think we want to tolerate anything. If a node is tainted, we shouldn't deploy on to it.

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1
+kind: DaemonSet
```
squeed, line 2: Is there's a reason this is on every node? Wouldn't a deployment be better?

manifests/image-references
```diff
@@ -58,4 +58,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-network-metrics-daemon:latest
-
+  - name: hello-openshift
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-hello-openshift:latest
```
squeed, line 64: This will have to change to a proper ImageReference provided by the CVO

squeed: sorry, nevermind!

pkg/controller/connectivitycheck/connectivity_check_controller.go
squeed: If you like, you could create the pods in this controller, rather than in the network controller. That would be a reasonable logical division.

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: Because we want to do a check from a pod to every node to cover the whole cluster.

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: fair, this I cargo culted, will amend.

rcarrillocruz: done

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: The rationale we agreed was:

From a pod to a pod on every node
From a pod to every kubeapiserver service/endpoint
From a pod to every openshiftapiserver service/endpoint
From a pod to LBs

bindata/network-diagnostics/001-rbac.yaml
rcarrillocruz: So you suggest having a Role with just secrets, so it's bound to openshift-network-diagnostics namespace and then bind it to the network-diagnostics ServiceAccount?

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
rcarrillocruz: Right, I cargo culted what apiserver folks put as RoleBinding name as schem per https://github.com/openshift/cluster-kube-apiserver-operator/blob/0a69c78cea511a8c8e70b3910c637ffaf2a1dc36/bindata/v4.1.0/kube-apiserver/check-endpoints-rolebinding.yaml#L4 , so you confirm that scheme is not usual?

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
rcarrillocruz: I misread, I thought were talking about the naming bikeshedding again.
Ok, will switch to ServiceAccount, thanks.

cmd/cluster-network-operator/main.go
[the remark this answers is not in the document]
rcarrillocruz: That's fair, better be safe, I will change.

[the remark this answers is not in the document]
rcarrillocruz: done

cmd/cluster-network-operator/main.go
```diff
@@ -102,16 +103,29 @@ func main() {
...
+	stopCh := signals.SetupSignalHandler()
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-stopCh
+		cancel()
+	}()
+	defer cancel()
+
+	log.Print("Starting the Cmd.")
+	func() {
```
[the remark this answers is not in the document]
rcarrillocruz: done

bindata/network-diagnostics/network-check-target.yaml
rcarrillocruz: Actually ,I take that back.
I need the toleration, as I want the hello-openshift to be deployed on masters too, so the source pod can do a network check against the whole cluster.
I removed it from network-check-source tho, as that I don't care which node is.

bindata/network-diagnostics/001-rbac.yaml
squeed: Yes, a Role rather than a ClusterRole is the right thing to do here.

We **definitely** don't want to have full access to all secrets across all namespaces.

bindata/network-diagnostics/network-check-target.yaml
squeed: Hmm. I suppose we don't have a better answer than tolerating all taints. Bah.

bindata/network-diagnostics/001-rbac.yaml
rcarrillocruz: ++

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
squeed: There's no need for us to name things in that manner. This is a private role for our operator's internal use. A simple, descriptive name is sufficient.

bindata/network-diagnostics/network-check-source.yaml
```diff
@@ -0,0 +1,92 @@
...
+          - name: check-endpoints
+            containerPort: 17698
+            protocol: TCP
+        resources:
+          requests:
+            memory: 50Mi
+            cpu: 10m
+
+---
+apiVersion: v1
```
squeed, line 55: If this is the traffic source, why does it need a Service?
If the service is just for metrics, then it should be a headless service.

bindata/network-diagnostics/network-check-target.yaml
squeed: gotcha.

bindata/network-diagnostics/network-check-source.yaml
rcarrillocruz: Ah good point, it's indeed for Prometheus, I'll change.

bindata/network-diagnostics/001-rbac.yaml
rcarrillocruz: done

bindata/network-diagnostics/network-check-source.yaml
rcarrillocruz: done

rcarrillocruz:
> > > _No description provided._
> > 
> > 
> > Please fix that. Every PR should explain exactly what it's doing in its initial comment.
> 
> Still applies. Please fix the initial description of the PR. You don't want future OCP hackers to send terminators back in time to kill you because they can't figure out what your PR does

Yeah, I added to commit description a while ago, it seems GH only adds it to PR description on PR creation.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+		"openshift-oauth-apiserver",
```
danwinship, line 383: This set of namespaces seems unnecessarily large

danwinship (commented):
There are various unaddressed comments from other reviews still (eg about the rolebindings)

pkg/network/render.go
```diff
@@ -409,3 +419,22 @@ func renderMultusAdmissionController(conf *operv1.NetworkSpec, manifestDir strin
 	out = append(out, objs...)
 	return out, nil
 }
+
+// renderDiagnostics renders the connectivity checks
+func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstructured, error) {
```
JacobTanenbaum, line 424: nit: the comment and the function name are different

bindata/network-diagnostics/network-check-target.yaml
```diff
@@ -0,0 +1,54 @@
...
+metadata:
+  name: network-check-target
+  namespace: openshift-network-diagnostics
+  annotations:
+    kubernetes.io/description: |
+      This daemonset deploys the network-check-target pods that run
+      a dummy hello-openshift app to be checked by network-check-source pod
```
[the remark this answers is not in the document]
rcarrillocruz: oh nice, didn't know that

manifests/0000_70_cluster-network-operator_01_crd.yaml
```diff
@@ -342,6 +342,13 @@ spec:
                 network support should be disabled. If unset, this property defaults
                 to 'false' and multiple network support is enabled.
               type: boolean
+            disableNetworkDiagnostics:
+              description: disableNetworkDiagnostics specifies whether or not PodNetworkConnectivityCheck
+                CRs from a test pod to every node, apiserver and LB should be disabled
+                or not. If unset, this property defaults to 'false' and network
```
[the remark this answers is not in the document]
rcarrillocruz: yep, that was the discussion i had with api folks. The rationale was as it is going to be enabled by default, they rather wanted to have 'disableFoo' flag, and well consistency with multiNetwork

pkg/network/render.go
rcarrillocruz: done

manifests/0000_70_cluster-network-operator_01_crd.yaml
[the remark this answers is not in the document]
rcarrillocruz: fixed the or not thingy

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
rcarrillocruz: done

manifests/0000_70_cluster-network-operator_01_crd.yaml
rcarrillocruz: done

pkg/cmd/checkendpoints/controller/backoff_recorder.go
rcarrillocruz: ended up using the LoggingRecorder from pki, making its own package

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
rcarrillocruz: If I don't have this rolebinding, the check-endpoints binary gives:

`W1202 21:48:26.242220       1 requestheader_controller.go:193] Unable to get configmap/extension-apiserver-authentication in kube-system.  Usually fixed by 'kubectl create rolebinding -n kube-system ROLEBINDING_NAME --role=extension-apiserver-authentication-reader --serviceaccount=YOUR_NS:YOUR_SA'`

bindata/network-diagnostics/001-rbac.yaml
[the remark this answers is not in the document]
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+	eventRecorder := &eventrecorder.LoggingRecorder{}
+	kubeClient, err := kubernetes.NewForConfig(protoKubeConfig)
+	if err != nil {
+		return err
+	}
+	configClient, err := configv1client.NewForConfig(kubeConfig)
+	if err != nil {
+		return err
+	}
+	operatorClient, dynamicInformers, err := genericoperatorclient.NewClusterScopedOperatorClient(kubeConfig, operatorv1.GroupVersion.WithResource("openshiftapiservers"))
```
squeed, line 361: You should use the operatorClient from `pkg/client` now

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+
+func withTarget(label, target string) func(check *v1alpha1.PodNetworkConnectivityCheck) {
+	return WithTarget(label + "-" + target)
+}
+
+func Start(ctx context.Context, kubeConfig *rest.Config) error {
+	protoKubeConfig := rest.CopyConfig(kubeConfig)
+	protoKubeConfig.ContentType = "application/vnd.kubernetes.protobuf,application/json"
+	eventRecorder := &eventrecorder.LoggingRecorder{}
+	kubeClient, err := kubernetes.NewForConfig(protoKubeConfig)
```
squeed, line 353: you can get this from `pkg/client`

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		return err
+	}
+	apiextensionsClient, err := apiextensionsclient.NewForConfig(kubeConfig)
+	if err != nil {
+		return err
+	}
+	apiextensionsInformers := apiextensionsinformers.NewSharedInformerFactory(apiextensionsClient, 10*time.Minute)
+	kubeInformersForNamespaces := v1helpers.NewKubeInformersForNamespaces(kubeClient,
+		"",
```
squeed, line 375: can you get rid of this line? this means it's a non-namespace-scoped informer, which can be overly aggressive.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+		configInformers,
+		apiextensionsInformers,
+		eventRecorder,
+	)
+
+	go connectivityCheckController.Run(ctx, 1)
+	apiextensionsInformers.Start(ctx.Done())
+	kubeInformersForNamespaces.Start(ctx.Done())
+	dynamicInformers.Start(ctx.Done())
+	configInformers.Start(ctx.Done())
```
squeed, line 401: Shouldn't you start the informers before running the controller?

squeed: and wait for cache sync? But if you switch to `pkg/client` then that's taken care for you.

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -0,0 +1,404 @@
...
+	"k8s.io/client-go/kubernetes"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+type NetworkConnectivityCheckController interface {
+	connectivitycheckcontroller.ConnectivityCheckController
+}
+
+func NewNetworkConnectivityCheckController(
```
squeed, line 38: It would be nice to get an overview of what this controller does. What configuration does it consume? What does it create in response?

pkg/operator/operator.go
```diff
@@ -94,6 +95,7 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 	go o.manager.Start(ctx.Done())
 	go logLevelController.Run(ctx, 1)
 	go managementStateController.Run(ctx, 1)
+	go connectivitycheck.Start(ctx, o.ccfg.KubeConfig)
```
squeed, line 98: it looks like this already returns, so no need to run it in a goroutine.

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: it seems there are methods left to implement:

pkg/controller/connectivitycheck/connectivity_check_controller.go:387:70: cannot use kubeClient (type *"github.com/openshift/cluster-network-operator/pkg/client".Client) as type kubernetes.Interface in argument to NewNetworkConnectivityCheckController:
        *"github.com/openshift/cluster-network-operator/pkg/client".Client does not implement kubernetes.Interface (missing AdmissionregistrationV1 method)

Given the short time to deliver, let's refactor later.

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: same as before, let's refactor later as it seems it lacks method to be implemented.

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
rcarrillocruz: done

pkg/operator/operator.go
rcarrillocruz: Done

abhat:
@rcarrillocruz noticed that the network operator reconcile loop is showing a bunch of errors with unregistered GVK

```
E1204 19:27:19.939188       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:20.031564       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:20.135990       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:23.402481       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:28.910603       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:28.918887       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:29.213138       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:29.221347       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:29.726810       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:30.635573       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
E1204 19:27:30.643651       1 base_controller.go:250] "ConnectivityCheckController" controller failed to sync "key", err: cannot decode "pkg/operator/connectivitycheckcontroller/manifests/controlplane.operator.openshift.io_podnetworkconnectivitychecks.yaml": no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "pkg/runtime/scheme.go:101"
I1204 19:31:11.621560       1 log.go:181] Reconciling update to openshift-multus/multus
```

Holding off for now till you can ack these are benign or we have a fix.

abhat:
Looks like the openstack-ovn job has been failing all over the place. Nothing to do with the PR itself. Cancelling the hold.

PR #858, opened 2020-10-30
WIP: Adds Namespace finalizers

dougbtv:
*NOTE* This is a WIP (in a number of ways). I'm posting it to test it in-cluster with cluster-bot.

Things it needs:

* Change in architecture to fit into the right spot (I kinda jammed it into a convenient place)
* Dan Winship suggested that we do this by labels on namespaces, this would be preferable to the static way this performs, should be straight forward to implement the label scheme.
* It has some trace output that needs to be removed.
* Rebase and squash
* It has some yaml spec changes that shouldn't be there.


dougbtv:
This is obsolete.

PR #859, opened 2020-11-02
Bug 1893362: Ensure tail processes exit with parent

cgwalters:
There's a great Linux-specific feature that allows sending a signal
to a child process when its parent dies.  This allows strongly
"lifecycle binding" two processes together and avoiding the default
Unix behavior where the child process will just be reparented.

I think this will help avoid problems like
https://bugzilla.redhat.com/show_bug.cgi?id=1893362
where when the bash process exits, the tail process sticks around.
Perhaps something isn't removing the pid file?  In any case when
this pod is requested to terminate, the tail processes should die too.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -62,8 +62,14 @@ spec:
             if [[ ! -S /run/openvswitch/db.sock ]]; then
               systemctl_restart ovsdb-server
             fi
+            # We need to explicitly exit on SIGTERM, see https://github.com/openshift/cluster-dns-operator/issues/65
+            function quit {
+                exit 0
+            }
+            trap quit SIGTERM
             # Don't need to worry about restoring flows; this can only change if we've rebooted
-            exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
+            tail --pid=$BASHPID -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log &
+            wait
           else
 
             echo "openvswitch is running in container"
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -75,12 +75,19 @@ spec:
             if [[ ! -S /run/openvswitch/db.sock ]]; then
               systemctl_restart ovsdb-server
             fi
+            # We need to explicitly exit on SIGTERM, see https://github.com/openshift/cluster-dns-operator/issues/65
+            function quit {
+                exit 0
+            }
+            trap quit SIGTERM
             # Don't need to worry about restoring flows; this can only change if we've rebooted
-            exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
+            tail --pid=$BASHPID -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log &
+            wait
           else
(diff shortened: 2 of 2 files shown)


cgwalters:
(Not tested)

cgwalters:
e.g. https://github.com/coreos/coreos-assembler/commit/2118af4b78fb69dae6d44c0eed13a1ce6e9ab02f

cgwalters:
Hm actually the new problem may be the ` exec tail -F` path.  Looking at that.

cgwalters:
OK updated.
/hold cancel

cgwalters:
What I've been doing to test is:

```
$ oc run tailtest --image=registry.fedoraproject.org/fedora:33 --restart=Never -- bash -c 'function quit {
                                                  echo got SIGTERM
                                                  exit 0
                                                  }
                                                  trap quit SIGTERM
                                                  setpriv --pdeathsig TERM -- tail -f /etc/hosts & wait; echo done'
$ oc delete pod/tailtest
```

Terminates right away.  But if I remove the `setpriv` or the `trap` it hangs.

cgwalters:
Why is this bash code duplicated in openshift-sdn and ovn-kubernetes btw?

dcbw:
> Why is this bash code duplicated in openshift-sdn and ovn-kubernetes btw?

@cgwalters because we don't have a bash library script that we add to all the containers that we can call?

cgwalters:
> @cgwalters because we don't have a bash library script that we add to all the containers that we can call?

OK, that seems not too hard to fix but we can do that later.

You might say there's a long tail of problems here.

cgwalters:
Argh, util-linux in RHEL8 is too old.  That's rather annoying.

cgwalters:
/test e2e-agnostic-upgrade 
/test e2e-azure-ovn
Flakes

cgwalters:
This one is passing CI, can I get the approves and lgtms and :tada: emoji etc.?

aojea:
/lgtm
impressive lesson on linux signals and processes :smile:

cgwalters:
Also needs a `/approve`, thanks!

bindata/network/openshift-sdn/sdn-ovs.yaml
```diff
@@ -62,8 +62,14 @@ spec:
...
               systemctl_restart ovsdb-server
             fi
+            # We need to explicitly exit on SIGTERM, see https://github.com/openshift/cluster-dns-operator/issues/65
+            function quit {
+                exit 0
+            }
+            trap quit SIGTERM
             # Don't need to worry about restoring flows; this can only change if we've rebooted
-            exec tail -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log
+            tail --pid=$BASHPID -F /host/var/log/openvswitch/ovs-vswitchd.log /host/var/log/openvswitch/ovsdb-server.log &
```
squeed, line 71: Question: why does `exec tail ...` not exit when it gets a SIGTERM? Shouldn't that always exit? Bash is gone...

cgwalters: By default in Linux (possibly Unix in general) pid 1 has `SIGTERM` set to ignore by default...  For basically bad reasons this is still the default with pid namespaces; it clearly would have been better to change it but we can't now.

https://vagga.readthedocs.io/en/latest/pid1mode.html
https://hynek.me/articles/docker-signals/
etc.

squeed: OHHHHHHHH. Wow.

squeed: (frantically checking my other pods...)

cgwalters:
vsphere looks like provisioning failures.
upgrade...hmmm, looks fairly red across the board.

PR #865, opened 2020-11-04
fix CRD / Codegen

squeed:
- make sure codegen is vendored in 
- bump controller-tools
- add in the manually generated annotations


danielmellado:
Hi @squeed it does seem like the CI is complaining about the magnet, mind taking a look before I cherry-pick this? Thanks!

hack/update-codegen.sh
```diff
@@ -20,7 +20,7 @@ if  ! ( command -v controller-gen > /dev/null ); then
   fi
 fi
 
-"${SCRIPT_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh" deepcopy \
+bash "${SCRIPT_ROOT}/vendor/k8s.io/code-generator/generate-groups.sh" deepcopy \
```
danwinship, line 21: why?

hack/verify-style.sh
```diff
@@ -23,8 +23,8 @@ trap 'rc=$?' ERR
 # Go to the root of the repo
 cd "$(git rev-parse --show-cdup)"
 
-GOFILES=$(find . -path ./vendor -prune -o -name '*.go' | grep -v vendor | grep -v pkg/operator/assets)
-GOPKGS=$(go list ./... | grep -v '/vendor/' | grep -v '/generated/' | grep -v pkg/operator/assets)
+GOFILES=$(find . -path ./vendor -prune -o -name '*.go' | grep -v vendor)
```
danwinship, line 26: fwiw

    GOFILES=$(find . -path ./vendor -prune -o -name '*.go' -print)

works too. (ie, explicitly telling it to print filenames that match the `-name` predicate overrides the default behavior of printing filenames that match any predicate)

hack/verify-style.sh
```diff
@@ -23,8 +23,8 @@ trap 'rc=$?' ERR
...
+GOPKGS=$(go list ./cmd... ./pkg/...)
```
danwinship, line 27: the old version was written before `go` knew about `vendor`. Just

    GOPKGS=$(go list ./...)

will work too. (And picks up the `version` subdir, which I'm not sure if you had meant to omit)

hack/update-codegen.sh
```diff
@@ -6,14 +6,12 @@ set -o pipefail
 
 SCRIPT_ROOT="$(dirname "${BASH_SOURCE[0]}")/.."
 
-if  ! ( command -v controller-gen > /dev/null ); then
-  # Need to take an unreleased version of controller-tools
-  # because of controller-tools bug 302
-  echo "controller-gen not found, installing sigs.k8s.io/controller-tools@83f6193..."
+if ( ! ( command -v controller-gen > /dev/null )  || test "$(controller-gen --version)" != "Version: v0.4.0" ); then
+  echo "controller-gen not found or out-of-date, installing sigs.k8s.io/controller-tools@v0.4.0..."
```
danwinship, line 10: if [[ $(controller-gen --version 2>/dev/null) != "Version: v0.4.0" ]]; then

?

also, "incorrect version", not "out of date", though do we really want to be blocking people from using v0.4.1+? Are we assuming API incompatibilities?

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -1,12 +1,12 @@
-
----
-apiVersion: apiextensions.k8s.io/v1beta1
+# This file is automatically generated. DO NOT EDIT
+apiVersion: apiextensions.k8s.io/v1
```
damemi, line 2: Not sure who to ask, but do you also want to update the CRD in https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/001-crd.yaml#L2 to `v1` as well? (from https://bugzilla.redhat.com/show_bug.cgi?id=1881636#c4, this BZ was originally for upgrading all the CRDs in openshift/api to v1)

If so, you may want to do it here and link this PR to that bugzilla

hack/update-codegen.sh
squeed: because the shell script isn't executable.

hack/update-codegen.sh
squeed: I hate it when I do codegen (in any of its forms) and the output changes in ways I don't understand. So, I prefer to pin versions.

hack/verify-style.sh
squeed: all of this is silly - filed SDN-1342 to switch to golangci-lint.

hack/verify-style.sh
squeed: We need to skip `dependencymagnet`, hence the changes here. (And version will go away soonish)

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
squeed: This is just to fix codegen; let's make it a separate PR.

squeed:
This isn't going to pass until https://github.com/openshift/release/pull/13489 merges.

squeed:
hey @tssurya, I know this is blocking you - want to review it?

hack/update-codegen.sh
danielmellado: What would be changed without this pinning exactly?

squeed:
/hold
something's broken.

hack/update-codegen.sh
squeed: @danielmellado if we don't pin a specific version, then people who run `hack/update-codegen.sh` might get different output depending on the version of the binary installed on their local machine.

squeed:
Test failures are spurious. I think we can merge this. Someone want to lgtm?

tssurya:
> hey @tssurya, I know this is blocking you - want to review it?

oops, too late to the party, I need to get better with doing reviews diligently but thanks for doing this, I'll rebase my patch on this.

PR #866, opened 2020-11-04
Revert "Add ovn-db-pod label on the master pods"

abhat:
Reverts openshift/cluster-network-operator#828

Changing the match-labels that included the `ovn-db-pod=true` label breaks upgrades from 4.6.1 to 4.7. Reverting this change and instead relying on matchExpressions style syntax for the ovndb-manager and ovn_db metrics collection.


abhat:
Closing in favor of #867

PR #867, opened 2020-11-04
Bug 1891716: Remove spurious label in the pod selector of the master daemonset.

abhat:
Remove the ovn-db-pod=true label from matchLabels since the daemonset
pod selector doesn't need to select on that label.


trozet (approved):
/lgtm

PR #868, opened 2020-11-05
Update PKI crd

tssurya:
Ran ./hack/update-codegen.sh 
There were some discrepencies for the PKI CRD when I was creating the CRD for the egress-router in PR https://github.com/openshift/cluster-network-operator/pull/852. So pulled out this change separately to keep it clean.

/assign @squeed


tssurya:
/close 

I did not know when I opened this PR that this was already being done in https://github.com/openshift/cluster-network-operator/pull/865

PR #869, opened 2020-11-05
[WIP]  Add IPsec CertificateSigningRequest controller

markdgray:
In order to keep private keys private to nodes, this patch
introduces a basic CertificateSigningRequest controller to CNO. This
allows a client to generate a cert/private key pair locally
and request for the certificate to be signed by a CNO CA.

In this implementation, a new CA is generated at each
run of the CNO. Future iterations, will attempt to use the
OperatorPKI CA that is rotated and shared via the ovn-ca
secret. Future iterations will also integrate this controller
as part of the OperatorPKI
    
In order to test, a script is included in the root directory
which creates certs and requests for them to be signed, it
can be run as follows:
    
    gen_sign_cert.sh <common name>
    
where, <common name> is the CN of the generated certificate.
    
### DIFF
--- a/gen_sign_cert.sh
+++ b/gen_sign_cert.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+if [ -z "$1" ]; then
+	echo "Usage: $0 <common name>"
+	exit
+fi
+
+cn=$1
+echo "Generating cert for common name = " ${cn}
+openssl genrsa -out test.key 2048
+
+openssl req -new -text \
+          -extensions v3_req \
+          -addext "subjectAltName = DNS:${cn}" \
+          -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
+          -key test.key \
+          -out test.csr
+
+csr_64=$(cat test.csr | base64 | tr -d "\n")
+
+cat <<EOF | oc apply -f -
+apiVersion: certificates.k8s.io/v1
+kind: CertificateSigningRequest
+metadata:
+  name: test
+spec:
+  groups:
+  - system:authenticated
+  request: ${csr_64}
+  signerName: network.openshift.io/ipsec
+  usages:
+  - client auth
+EOF
+
+cert_64=$(oc get csr/test -o jsonpath='{.status.certificate}')
+
+oc get csr/test -o jsonpath='{.status.certificate}' | base64 -d | openssl x509 -outform pem -text
+
+oc delete csr/test
--- a/pkg/controller/ipsec/ipsec_controller.go
+++ b/pkg/controller/ipsec/ipsec_controller.go
@@ -1,0 +1,185 @@
+package ipsec
+
+import (
+	"context"
+	c "crypto"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"errors"
+	"log"
+	"math/big"
+	"time"
+
+	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"github.com/openshift/library-go/pkg/crypto"
+	csrv1 "k8s.io/api/certificates/v1"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const ipsecSignerName = "network.openshift.io/ipsec"
+
+var ca *crypto.TLSCertificateConfig
+
+// Add and Start it when the Manager is Started.
+func Add(mgr manager.Manager, status *statusmanager.StatusManager) error {
+	return add(mgr, newReconciler(mgr, status))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager) reconcile.Reconciler {
+	return &ReconcileCSR{client: mgr.GetClient(), scheme: mgr.GetScheme(), status: status}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("ipsec-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to CetificateSigningRequest resource
+	err = c.Watch(&source.Kind{Type: &csrv1.CertificateSigningRequest{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Make an ephemeral (in-memory) CA.
+	// TODO: This is temporary as it should use the Operator PKI ovn-ca secret
+	ca, _ = crypto.MakeSelfSignedCAConfig("ipsec-ca", 365)
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileCSR{}
+
+// ReconcileCSR reconciles a cluster CertificateSigningRequest object
+type ReconcileCSR struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+	status *statusmanager.StatusManager
+}
+
+// Reconcile
+func (r *ReconcileCSR) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	csr := &csrv1.CertificateSigningRequest{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, csr)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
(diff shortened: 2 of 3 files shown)


pkg/controller/ipsec/ipsec_controller.go
```diff
@@ -0,0 +1,185 @@
...
+
+	cert, err := extractCertificateRequest(csr.Spec.Request)
+	if err != nil {
+		log.Println("Could extract certificate from request")
+		return reconcile.Result{}, nil
+	}
+	cacert, err := extractCert(ca)
+	if err != nil {
+		log.Println("Could not extract CA certificate")
+		return reconcile.Result{}, nil
```
aojea, line 106: should we return error?

pkg/controller/ipsec/ipsec_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	signedCert, err := signCertificate(newCertificateTemplate(), cert.PublicKey, cacert, ca.Key)
+	if err != nil {
+		log.Println("Could not sign request")
+		return reconcile.Result{}, nil
```
aojea, line 112: err?, same for the other places, seems on err we always return nil

pkg/controller/ipsec/ipsec_controller.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		// Error reading the object - requeue the request.
+		log.Println(err)
+
+		return reconcile.Result{}, err
+	}
+
+	// Only handle CSRs for this signer
+	if csr.Spec.SignerName != ipsecSignerName {
+		return reconcile.Result{}, nil
```
aojea, line 95: log may be useful for debugging here

gen_sign_cert.sh
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  groups:
+  - system:authenticated
+  request: ${csr_64}
+  signerName: network.openshift.io/ipsec
+  usages:
+  - client auth
+EOF
+
+cert_64=$(oc get csr/test -o jsonpath='{.status.certificate}')
```
aojea, line 35: not familiar with openshift auth, does this requires to configure some roles and permissions in the cluster
https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#authorization

pkg/controller/ipsec/ipsec_controller.go
markdgray: In this case, it may be right to do so as we would probably want the reconciliation to be requeued. I will update this.

pkg/controller/ipsec/ipsec_controller.go
markdgray: In some cases, I don't think I should return error as the reconciliation will be requeued - especially in cases in which the CSR is malformed. In those cases, I should set the status appropriately and not try to reconcile again. TBH, I havent though too much about this yet as I just wanted to get the basics working.

pkg/controller/ipsec/ipsec_controller.go
markdgray: Yeah, I will add it here.

gen_sign_cert.sh
markdgray: Me either yet. I had a quick read. What I would like to do here (at a high-level), is only authorize the OVN role for CSRs to the SignerName "network.openshift.io/ipsec" or even "network.openshift.io". However, I would automatically sign anyone with that role. 

By the way, thanks for the review @aojea.

pkg/controller/ipsec/ipsec_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// Watch for changes to CetificateSigningRequest resource
+	err = c.Watch(&source.Kind{Type: &csrv1.CertificateSigningRequest{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	// Make an ephemeral (in-memory) CA.
+	// TODO: This is temporary as it should use the Operator PKI ovn-ca secret
```
squeed, line 59: I think it would be better for ipsec to use its own CA, and just store that in a Secret.

pkg/controller/ipsec/ipsec_controller.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// Only handle CSRs for this signer
+	if csr.Spec.SignerName != ipsecSignerName {
+		return reconcile.Result{}, nil
+	}
+
+	cert, err := extractCertificateRequest(csr.Spec.Request)
+	if err != nil {
+		log.Println("Could extract certificate from request")
```
squeed, line 100: You should also integrate with the StatusManager - if the CA is in a bad state, you should set yourself Degraded with a useful message.

squeed: (This applies to most cases where you would error out).

pkg/controller/ipsec/ipsec_controller.go
squeed: Yup, your instinct is correct - return `nil`, set status.

squeed:
So, a thought occurs to me: this will allow anyone who can create a CertificateSigningRequest object to request an ipsec certificate. Is that a problem? What are the security implications?

I need to see who can create CertificateSigningRequests - I'm not sure what the "default" role contains.

gen_sign_cert.sh
```diff
@@ -0,0 +1,39 @@
...
+openssl genrsa -out test.key 2048
+
+openssl req -new -text \
+          -extensions v3_req \
+          -addext "subjectAltName = DNS:${cn}" \
+          -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
+          -key test.key \
+          -out test.csr
+
+csr_64=$(cat test.csr | base64 | tr -d "\n")
```
rbbratta, line 19: `cat` is not necessary

`csr_64=$(<test.csr base64 | tr -d "\n")`

gen_sign_cert.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cat <<EOF | oc apply -f -
```
rbbratta, line 21: `cat` is usually never needed

`<<EOF oc apply -f -`

pkg/controller/ipsec/ipsec_controller.go
markdgray: No problem with that but what is your thought process. I thought it would be better to have a centralized ovn-ca that could be reused? If I dont do this, I guess I can create another cert-rotation-controller like you have done.

pkg/controller/ipsec/ipsec_controller.go
markdgray: ok

markdgray:
> So, a thought occurs to me: this will allow anyone who can create a CertificateSigningRequest object to request an ipsec certificate. Is that a problem? What are the security implications?
> 
> I need to see who can create CertificateSigningRequests - I'm not sure what the "default" role contains.

I haven't look into this yet but I am assuming I could check the role of the user that created the CSR and use that to decide if I will sign.

gen_sign_cert.sh
markdgray: Ok

gen_sign_cert.sh
markdgray: ok

markdgray: it doesnt work like that?

gen_sign_cert.sh
markdgray: it doesnt work like that?

rbbratta: err, `base64` takes a file arg, and `-w 0` disables wrapping, so we don't need the shell.

`csr_64=$(base64 -w 0 test.csr)`

rbbratta: of course `-w 0` is only GNU coreutils, and doesn't work on busybox or Mac OS X BSD base64, so whatever. :man_shrugging:

pkg/controller/ipsec/ipsec_controller.go
markdgray: @squeed I could use the operator PKI to create a new CA. E.g. not instantiate another cert rotatation controller.

gen_sign_cert.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
squeed, line 1: This is just for testing, right? Or do we expect nodes to execute this?

pkg/controller/ipsec/ipsec_controller.go
squeed: Basically, the ovn-ca is really intended to only be used for ovn database communications; it would be better that ipsec uses a different ca hierarchy, since they're separate.

If all you need is a CA, then the cert-rotation-controller isn't necessary; it's only a few lines of code to create the Secret with the CA. Create, like, a 100-year CA and just be done with it.

gen_sign_cert.sh
markdgray: It's just temporary for testing. However, each node will run something like this.

pkg/controller/ipsec/ipsec_controller.go
markdgray: I think I should rotate the CA. I will use the operatorpki resource to create another certrotation controller and call it signer-ca or something like that. Hopefully I will have a more complete PR by the end of the week and you can review.

squeed: > I think I should rotate the CA.

Can ipsec handle that case? Remember, rotating the CA is not atomic. linux ipsec will need to support multiple CAs at once.

markdgray: Good question but I am more worried about OVS which may not. I am not sure but probably something we will need to eventually do. For the moment, I have integrated with operator pki but can change in a future rework if required.

markdgray:
Deprecating as I have a new PR that has full implementation of IPsec

PR #863, opened 2020-11-03
Start to move to a library-go style operator

squeed:
- Build with a Makefile
- Enable the standard controllers:
    - ClusterOperatorLoggingController
    - OperatorManagementStateController
- Use the standard controller config

Also, remove the renderer, it's not useful anymore.

/cc @danwinship

PR #875, opened 2020-11-12
Bug 1889701: Use openshift/api as the source of truth for CNO CRD

rcarrillocruz:


squeed:
Much easier, thanks!
/lgtm

squeed:
actually, hang on. Do we even need this file at all, or is it created somewhere else?
/lgtm cancel

rcarrillocruz:
I see quite a bit of discrepancies between the CRD on openshift/api vs CNO. I think I'll push a copy straight to openshift/api first, looks like the CI issues may be due to some issues with the fields

rcarrillocruz:
We need to wait for https://github.com/openshift/api/pull/770/files to land, otherwise we'll keep getting the oc explain test failure.

rcarrillocruz:
Closing as https://github.com/openshift/cluster-network-operator/pull/863 implements this

PR #879, opened 2020-11-13
Simplify kube-rbac-proxy startup script

juanluisvaladas:
CNO tries to create the objects defined in bindata's files in
lexicographic order, so we can rely on the secret being created when
this starts.

I don't know what happens if during an upgrade the CNO
fails to create the services, so it's probably a good idea not to
backport this to 4.6.

/assign @tssurya

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -192,29 +192,7 @@ spec:
           # As the secret mount is optional we must wait for the files to be present.
           # The service is created in monitor.yaml and this is created in sdn.yaml.
           # If it isn't created there is probably an issue so we want to crashloop.
-          retries=0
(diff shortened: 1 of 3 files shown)


squeed:
AIUI, the Secret isn't created by the CNO, it's created by the service-cert controller. So we can't actually rely on it existing.

If you wanted to create an empty secret, that might work. I believe the mount would be automatically updated when the serving certs are created.

juanluisvaladas:
@squeed this code tests that the service is created, not that the secret is created. We wait for the cert to be mounted later:
https://github.com/openshift/cluster-network-operator/blob/cea0ebfcd15da13aaa7efbedc0b3e6c885866e40/bindata/network/openshift-sdn/sdn.yaml#L209-L212

squeed:
Oh duh.

/approve
/lgtm

PR #880, opened 2020-11-13
Assign openshift sdn traffic to system priority level

tkashem:
Assign sdn traffic to `system` priority level, this will bring sdn traffic to the same priority level as kubelet traffic. 

Currently the kube-apiserver does not allocate the sdn traffic to any dedicated concurrency pool, so treating it no different than the default cluster workload. This change will allow the kube-apiserver to allocate the sdn traffic to a known priority level `system` shared by kubelet.

For more information on priority and fairness - https://kubernetes.io/blog/2020/04/06/kubernetes-1-18-feature-api-priority-and-fairness-alpha/


manifests/0000_70_cluster-network-operator_04_kubeapiserver_flowschema.yaml
```diff
@@ -0,0 +1,33 @@
...
+      clusterScope: true
+      namespaces:
+      - '*'
+      resources:
+      - '*'
+      verbs:
+      - '*'
+    subjects:
+    - kind: ServiceAccount
+      serviceAccount:
```
tkashem, line 31: did i get the service account right?

squeed: We have two service accounts, `sdn-controller` and `sdn`. They're both system priority.

squeed:
One small change (both SAs) but otherwise seems reasonable.

manifests/0000_70_cluster-network-operator_04_kubeapiserver_flowschema.yaml
tkashem: @squeed I added `sdn-controller` as you suggested.

squeed:
Looks reasonable
/approve
I'll leave lgtm to @deads2k, since I can't speak for apiserver QOS.

deads2k:
@tkashem this is probably worth a backport.

dcbw:
@tkashem can you rebase? Thanks!

tkashem:
@dcbw rebase completed, it needs re-lgtm.

danwinship:
Um, this seems wrong; the `sdn` and `sdn-controller` ServiceAccounts are only created if openshift-sdn is selected as the network plugin, so this will be referring to non-existent ServiceAccounts when ovn-kubernetes, kuryr, or a third-party network plugin is in use. Even if that doesn't make things fail, we definitely want this traffic priority thing to be applying at least to both openshift-sdn and ovn-kubernetes...

tkashem:
> Um, this seems wrong; the sdn and sdn-controller ServiceAccounts are only created if openshift-sdn is selected as the network plugin, so this will be referring to non-existent ServiceAccounts when ovn-kubernetes, kuryr, or a third-party network plugin is in use. Even if that doesn't make things fail, we definitely want this traffic priority thing to be applying at least to both openshift-sdn and ovn-kubernetes...

@danwinship so this implies that we have to create a flowschema based on the network stack installed by the network operator.

squeed:
Yes, @danwinship is right. I shouldn't have approved this.

PR #884, opened 2020-11-19
Bug 1948718: IBM Cloud manifest profile patch

csrwng:
Part 2 of cluster profiles support for IBM Cloud:
- Bumps github.com/openshift/build-machinery-go to latest version
- Adds targets to Makefile for updating/verifying profile patches
- Adds profile patch for ibm-cloud-managed

On IBM Cloud the network operator needs to run on worker nodes, thus the
master node selector needs to be removed from the operator deployment
manifest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,16 @@ all: build
 include $(addprefix ./vendor/github.com/openshift/build-machinery-go/make/, \
 	golang.mk \
 	targets/openshift/deps-gomod.mk \
+	targets/openshift/operator/profile-manifests.mk \
 )
+
+# This will include additional actions on the update and verify targets to ensure that profile patches are applied
+# to manifest files
+# $0 - macro name
+# $1 - target name
+# $2 - profile patches directory
+# $3 - manifests directory
+$(call add-profile-manifests,manifests,./profile-patches,./manifests)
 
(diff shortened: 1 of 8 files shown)


csrwng:
/hold
Depends on cluster profiles support merging in CVO

danwinship:
Why does this need both a patch to the deployment _and_ a separate copy of the deployment?

Oh, wait, I see; the separate copy is generated from the patch. But in that case, then why do we need to commit it? Can't it just be built at build time? Failing that can we at least get a comment in the generated file indicating that it's autogenerated?

csrwng:
@danwinship The comment in the generated manifest sounds like a good idea. I will submit a change to `build-machinery-go` to add that.

csrwng:
Submitted https://github.com/openshift/build-machinery-go/pull/32
Will update this PR with an updated vendor of that repo when that merges.

csrwng:
@danwinship re-bumped build-machinery-go with the additional comment on the generated manifest

csrwng:
@danwinship had to rebase

csrwng:
Rebased

csrwng:
@danwinship can you retag w lgtm again? Thank you.

PR #886, opened 2020-11-20
Enable IPsec

markdgray:
This PR enables IPsec for the Cluster Network Operator. There are a number of components to this:

* Re-vendor in order to pull in the latest changes to openshift/api which include updates to the CNO crd.
* Addition of a CertificateSigningRequest controller that watches for CertificateSigningRequest resources and signs them using a CA key pair that has been created by the OperatorPKI.
* Modification of the ovnkube-master and ovs-nodes manifests and creation of a new ovn-ipsec Daemonset that generate necessary keys and configure OVN and OVS to enable IPsec across the network. This change can only be made at cluster installation time.

This PR is dependent on a number changes in other projects that have not been merged:
* **openshift/ovn-kubernetes:** https://github.com/openshift/ovn-kubernetes/pull/344 (this pulls in the libreswan and openvswitch-ipsec packages)
* **ovs:** https://patchwork.ozlabs.org/project/openvswitch/list/?series=215446 (this fixes a bug in ovs-monitor-ipsec)
* **kernel:** https://bugzilla.redhat.com/show_bug.cgi?id=1891818 (kernel geneve fix needs to be backported to RHEL 8.3)

This PR has been minimally tested with IPsec enabled but will require the kernel patch for full testing by the CI (which may throw up some issues). However, this should pass the current CI as IPsec is disabled by default and cannot be enabled after cluster initialization.


markdgray:
@squeed @abhat @rcarrillocruz @trozet @danwinship @alexanderConstantinescu

markdgray:
@squeed @rcarrillocruz  @trozet @alexanderConstantinescu @danwinship @abhat : @dcbw suggested adding you for review

abhat:
As per https://github.com/openshift/api/pull/783/files#r528769454, consider the `Enable` flag based on the `IPsecConfig` being present. If the IPsecConfig is empty struct, it can be interpreted as the feature is disabled.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,142 @@
...
+          /usr/libexec/ipsec/_stackmanager start
+          # Check for nss database status and migration
+          /usr/sbin/ipsec --checknss
+          # Check for nflog setup
+          /usr/sbin/ipsec --checknflog
+          # Start the actual IKE daemon
+          /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
+
+          # Required until latest upstream package is added to RHEL
+          curl https://github.com/lzhecheng/ovs/commit/869b06356e389079861962160e864df609d033e5.patch > ipsec.patch
```
squeed, line 72: just FYI, we can't merge something like this.

If this won't be in RHEL in time, then we need to fix this in the dockerfile. And the patch needs to be in the repo; not downloaded.

bindata/network/ovn-kubernetes/002-rbac.yaml
```diff
@@ -65,6 +65,14 @@ rules:
 - apiGroups: ['authorization.k8s.io']
   resources: ['subjectaccessreviews']
   verbs: ['create']
+- apiGroups: [certificates.k8s.io]
+  resources: ['certificatesigningrequests']
+  verbs:
+  - create
+  - get
+  - delete
```
squeed, line 73: Probably just needs get & create?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -80,10 +80,60 @@ spec:
                 exit 0
             }
             trap quit SIGTERM
+{{ if .EnableIPsec }}
```
squeed, line 83: The OVS daemonsets will be going away; is there another place this could live?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -80,10 +80,60 @@ spec:
...
+            kubectl delete --ignore-not-found=true csr/$(hostname)
+
+            cat <<EOF | kubectl apply -f -
+            apiVersion: certificates.k8s.io/v1
+            kind: CertificateSigningRequest
+            metadata:
+              name: $(hostname)
+            spec:
+              groups:
+              - system:authenticated
```
squeed, line 113: I don't think we want this group; the certificate shouldn't actually have any k8s rights.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -80,10 +80,60 @@ spec:
...
+              request: ${csr_64}
+              signerName: network.openshift.io/signer
+              usages:
+              - client auth
+          EOF
+
+            kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' | base64 -d | openssl x509 -outform pem -text -out ipsec-cert.pem
```
squeed, line 120: Presumably we need to wait for this to be signed, right?

enableIPsec.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+oc  patch  network.operator.openshift.io/cluster --type=merge  -p  '{"spec": {"defaultNetwork" :{ "ovnKubernetesConfig" :{ "ipsecConfig" :{ "enable":true}}}}}'
```
squeed, line 2: This should be in `hack` or deleted.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileCSR{}
+
+// ReconcileCSR reconciles a cluster CertificateSigningRequest object
```
squeed, line 68: Can you add a bit more documentation here?

pkg/controller/signer/signer.go
```diff
@@ -0,0 +1,84 @@
...
+
+func newCertificateTemplate(certReq *x509.CertificateRequest) *x509.Certificate {
+
+	template := &x509.Certificate{
+		Subject: certReq.Subject,
+
+		SignatureAlgorithm: x509.SHA256WithRSA,
+
+		NotBefore:    time.Now().Add(-1 * time.Second),
+		NotAfter:     time.Now().Add(24 * time.Hour),
```
squeed, line 24: Can we recover from an expired certificate? It's not uncommon for whole clusters to be shut down for days, especially in internal training environments. Or when :hankey: really happens.

pkg/controller/signer/signer.go
```diff
@@ -0,0 +1,84 @@
...
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		BasicConstraintsValid: true,
+	}
+
+	return template
+}
+
+func signCertificate(template *x509.Certificate, requestKey c.PublicKey, issuer *x509.Certificate, issuerKey c.PrivateKey) (*x509.Certificate, error) {
```
squeed, line 37: nit: `signCSR`

squeed:
/hold
because this is still a work in progress.

squeed:
How do we handle certificate rotation? What happens if a node misses its rotation deadline?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Yes, of course. This is currently building against a private ovn_k8s image. But this (and the environment variables a few lines below) will be fixed by this commit: https://github.com/openshift/ovn-kubernetes/pull/344

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: I guess that depends on where they are going and in what time frame. Do you have a suggestion? I could add something like this to ovn-kubernetes but because it needs to request signing from a controller that doesn't exist in the ovn kubernetes project, it didn't seem suitable to put it there. Will this go away in OCP 4.7 time frame?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: Yes, good point.

enableIPsec.sh
markdgray: All the content in the last commit will be deleted.

pkg/controller/signer/signer-controller.go
markdgray: Ok, no problem.

pkg/controller/signer/signer.go
markdgray: Ok, I can make that change.

markdgray:
> How do we handle certificate rotation? What happens if a node misses its rotation deadline?

We don't. I think we will need this but we can't use the cert rotation controller for this. I also don't think this was specified as part of the enhancement proposal. I could make the certificates have a long lifespan initially? Restarting the container will, of course, recreate and resign but as you mentioned, that Daemonset will be going away.

bindata/network/ovn-kubernetes/002-rbac.yaml
markdgray: Currently, I delete and recreate the certificate every time the Daemonset is started,

markdgray:
@squeed Thanks for the quick review!

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: Just thinking about this. I will move this code into the ipsec container which can then also generate the keys (on the host filesystem) and configure ovs with them.

astoycos: Agree with @squeed The OVS daemonsets are just alive to help with logging since ovs is back to running on the hosts. It wouldn't make sense to put it in upstream... Why not just have a custom downstream container startup script for the IPsec Daemon that handles this? (Sorry if this is already what you meant @markdgray )

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I will remove this now as the openvswitch 2.13-ipsec package is available. Just waiting for this PR to merge: https://github.com/openshift/ovn-kubernetes/pull/344

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: Yes, as long as I can run `ovs-vsctl` in the ipsec pod, I can move all this there. This will simplify the design.

pkg/controller/signer/signer.go
markdgray: Apologies, I meant to set this to longer. Thanks for catching it.

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: Correct. Cut and paste mistake

enableIPsec.sh
markdgray: In fact, I will just remove it now to save confusion. I thought it might someone with testing.

markdgray:
> As per https://github.com/openshift/api/pull/783/files#r528769454, consider the `Enable` flag based on the `IPsecConfig` being present. If the IPsecConfig is empty struct, it can be interpreted as the feature is disabled.

@abhat https://github.com/openshift/api/pull/794 I will update this PR when/if this gets accepted

markdgray:
@squeed @abhat Updated as per your comments. Any further issues?

bindata/network/ovn-kubernetes/006-ovs-node.yaml
markdgray: @astoycos I didn't add a script to downstream yet. I don't think it makes sense to do that now as the feature is not enabled downstream. Maybe future versions.

markdgray:
@squeed I made a few additional changes as I managed to get a cluster running with a customized kernel and run some basic tests which picked up a few integration issues. You can see the diff of the last push to see them.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+            ovn-nbctl set nb_global . ipsec=true
+          fi
+
+          while true; do
+            sleep 10
+          done
+        env:
+        - name: OVS_LOG_LEVEL
+          value: info
+        - name: K8S_NODE
```
alexanderConstantinescu, line 145: You don't seem to be using this environment variable anywhere, but instead use `hostname`, so this could be removed right?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+        - mountPath: /signer-ca
+          name: signer-ca
+        - mountPath: /etc/openvswitch
+          name: etc-openvswitch
+        resources:
+        # TODO what is reasonable here?
+          requests:
+            cpu: 100m
+            memory: 300Mi
+          limits:
```
alexanderConstantinescu, line 167: I don't know what is reasonable, but usually it's better to not set `limits` as that might cause the pod to be CPU/Mem throttled if it exceeds these limits, leading to bugs which are difficult to pin-point.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+            memory: 400Mi
+        terminationMessagePolicy: FallbackToLogsOnError
+        readinessProbe:
+          exec:
+            command:
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              ovs-appctl -t ovs-monitor-ipsec ipsec/status
```
alexanderConstantinescu, line 178: Do we need to parse the output and validate a specific status? Or is it fine if this just returns the exit code 0?

pkg/network/ovn_kubernetes.go
```diff
@@ -108,7 +108,22 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVNHybridOverlayVXLANPort"] = ""
 	}
 
+	if c.IPsecConfig != nil {
+		data.Data["EnableIPsec"] = c.IPsecConfig.Enable
```
alexanderConstantinescu, line 112: Nit: we don't really need to set `data.Data["EnableIPsec"]` right? The ovn-kubernetes IP sec daemonset will be rendered based on what happens in this function, the template variable has no impact, am I right?

alexanderConstantinescu (commented):
In general `/lgtm` from me, mainly minor nits here and there. I am not really "IP sec savvy", so in all honesty: I wasn't really able to validate the behavior of `ovn-ipsec` container in detail, but I trust you do :)

pkg/controller/signer/signer.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, errors.New("Expected a single certificate")
+	}
+	return certs[0], nil
+}
+
+func decodeCertificateRequest(pemBytes []byte) (*x509.CertificateRequest, error) {
+	block, _ := pem.Decode(pemBytes)
+	if block == nil || block.Type != "CERTIFICATE REQUEST" {
+		err := errors.New("PEM block type must be CERTIFICATE_REQUEST")
+		log.Println(err)
```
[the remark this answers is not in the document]
markdgray: Ok, I will update that.

pkg/controller/signer/signer.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	return x509.ParseCertificateRequest(block.Bytes)
+}
+
+func decodeCertificate(pemBytes []byte) (*x509.Certificate, error) {
+	block, _ := pem.Decode(pemBytes)
+	if block == nil || block.Type != "CERTIFICATE" {
+		err := errors.New("PEM block type must be CERTIFICATE")
+		log.Println(err)
```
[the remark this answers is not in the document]
markdgray: Ok, I will update that.

pkg/controller/signer/signer.go
```diff
@@ -0,0 +1,89 @@
...
+
+	return x509.ParseCertificate(block.Bytes)
+}
+
+func decodePrivateKey(pemBytes []byte) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode(pemBytes)
+	if block == nil || block.Type != "RSA PRIVATE KEY" {
+		fmt.Println(block.Type)
+		err := errors.New("PEM block type must be RSA PRIVATE KEY")
+		log.Println(err)
```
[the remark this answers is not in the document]
markdgray: Ok, I will update that.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Yes, you are right - I copied this spec file from the OVS one. I will do another parse of it to make sure I havent missed anything else.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Yeah, I will do that then. I wasnt sure if there was a convention here?

pkg/network/ovn_kubernetes.go
markdgray: I can remove that now. The last change I made removed templating of the ovnkube-master yaml (which was needed) but I didnt think of removing this.

markdgray:
Thanks for the review @alexanderConstantinescu

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
alexanderConstantinescu: I don't know of any official convention, but I would say that limits serve a bigger purpose for user deployed applications for which the CPU and Memory consumption has been studied in advance. In those cases, a user could define limits on pods as to avoid crashing a node if there's a memory leak in the component. This is however a system critical component, and we shouldn't be restricting it, because if the CPU / Mem spikes at a given moment (maybe because of too much traffic I guess?) then the pod gets throttled (i.e: it gets less access to the CPU) by the scheduler.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+      # ovs-monitor-ipsec and libreswan daemons
+      - name: ovn-ipsec
+        image: "{{.OvnImage}}"
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+
+          function cleanup()
```
squeed, line 45: This daemonset will be stopped on live nodes during upgrades - will this cause traffic to be dropped? (I assume so).

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+          }
+          trap cleanup SIGTERM
+
+          # Every time we restart the container, we will create a new key pair
+          echo "Configuring IPsec keys"
+
+          # We use the system-id as the CN. This is a requirement for OVN
+          cn=$(ovs-vsctl --retry -t 60 get Open_vSwitch . external-ids:system-id | tr -d "\"")
+
+          mkdir -p /etc/openvswitch/keys
```
squeed, line 64: This should probably be an InitContainer; that would be a bit more logical.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+          pushd /etc/openvswitch/keys
+          umask 077 && openssl genrsa -out ipsec-privkey.pem 2048
+          openssl req -new -text \
```
squeed, line 67: So, this generates a key and submits a CSR. That's fine... but this needs to be a bit more of an idempotent reconciliation loop that

1. Looks to see if a signed certificate exists on disk that is not about to expire
2. Generates a key + CSR if needed
3. Configures OVN-k
4. GOTO 1.

squeed: You might have to write this in go.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+          # Workaround for https://mail.openvswitch.org/pipermail/ovs-dev/2020-October/375734.html
+          OVS_LOGDIR=/var/log/openvswitch OVS_RUNDIR=/var/run/openvswitch OVS_PKGDATADIR=/usr/share/openvswitch /usr/share/openvswitch/scripts/ovs-ctl --ike-daemon=libreswan start-ovs-ipsec
+
+          # Wait for the nbdb container in the ovnkube-master pod to become ready before we try to write to
+          # it.
+          if kubectl wait --for=condition=Ready pod -l app=ovnkube-master -n ${ovn_kubernetes_namespace} --timeout 600s; then
+            echo "ovnkube-master is not Ready, unable to write to OVNNB DB"
+            exit 2
+          else
+            ovn-nbctl set nb_global . ipsec=true
```
squeed, line 136: Don't touch the NBDB in the ovnkube-node database - this needs to be somewhere else.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+          metadata:
+            name: $(hostname)
+          spec:
+            request: ${csr_64}
+            signerName: network.openshift.io/signer
+            usages:
+            - ipsec tunnel
+          EOF
+
+          until [ ! -z $(kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' 2>/dev/null) ]
```
squeed, line 92: Can't you just wait for the signed certificate to show up in the mount?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+                    echo "Unable to sign certificate after $counter seconds"
+                    exit 1
+            fi
+          done
+
+          kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' | base64 -d | openssl x509 -outform pem -text -out ipsec-cert.pem
+
+          kubectl delete csr/$(hostname)
+
+          cat /signer-ca/tls.crt | openssl x509 -outform pem -text > /etc/openvswitch/keys/ipsec-cacert.pem
```
squeed, line 106: All of these writes need to be atomic (moving a symlink), because we don't know when ovs-vswitchd will read the certificate.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: Okay, if certificate lifecycle isn't going to be solveable yet, then at least let's shortcut and not regen the certificate if it's not too old.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: In other words, write this to `/etc/openvswitch/keys/ipsec-cert-NNNN.pem` and then set a symlink

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+          ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=/etc/openvswitch/keys/ipsec-cert.pem \
+                                                     other_config:private_key=/etc/openvswitch/keys/ipsec-privkey.pem \
+                                                     other_config:ca_cert=/etc/openvswitch/keys/ipsec-cacert.pem
+
+          popd
```
squeed, line 112: ouch - let's just use absolute paths here.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+          # Workaround for https://github.com/libreswan/libreswan/issues/373
+          ulimit -n 1024
+
+          /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
+          # Check for kernel modules
+          /usr/libexec/ipsec/_stackmanager start
```
squeed, line 119: do we need to bind-mount `/lib/modules` ?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+      # /env -> configmap env-overrides - debug overrides
+      containers:
+      # ovs-monitor-ipsec and libreswan daemons
+      - name: ovn-ipsec
+        image: "{{.OvnImage}}"
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
```
squeed, line 43: This should have `set -euo pipefail` or something similar.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I did do that initially but it was causing me problems due to another issue. I will change it back.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Why should we worry about not regenerating the certificate? It simplifies things a lot if we dont check validity,etc. When we include better support of certificate rotation, we can refactor and maybe write it in go.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I wanted to do this in the nbdb in ovnkube-master, however, due to the way that pod is started (using ovn-ctl run_nb_ovsdb - which blocks) i couldn't easily template it in. Is this the correct place or could you suggest elsewhere?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: ok

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: ok

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Yeah, I will do this.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I can't do this because the signed-cert does not appear via a secret. It is shared as a resource. This is ok as the signed cert is a public key that can be shared safely.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
abhat: In what contexts do we stop/kill the ovn-ipsec and libreswan containers? Is this during cert rotation? If so, cleaning out the xfrm policy and xfrm state on each node by node basis will cause replay errors on other nodes sending traffic to the node where the container is being re-provisioned?

README.md
```diff
@@ -256,6 +256,16 @@ The hybridClusterNetwork `cidr` and hostPrefix are used when adding windows node
...
 
+#### Configuring IPsec with OVNKubernetes
+OVNKubernetes supports IPsec encryption of all pod traffic using the OVN IPsec functionality. Add the following to the `spec:` section of the operator config:
+
+```yaml
+spec:
+  defaultNetwork:
+    type: OVNKubernetes
+    ovnKubernetesConfig:
+      ipsecConfig:
```
squeed, line 267: super minor nit: this should be `ipsecConfig: {}` I think?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: ack.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: This isn't the right place; it needs to be in the master pod somewhere.

If it makes sense, it can be a PostStartHook on the correct container. Or it can be its own container. that configures ipsec and sleeps forever.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: We use it for overriding loglevels on a per-node basis - perhaps that mechanism is useful here?

Check out the env-overrides bit in the node daemonset.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: The way this is written, it will definitely blackhole the node during in-place upgrades - no good :-)

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+      labels:
+        app: ovn-ipsec
+        component: network
+        type: infra
+        openshift.io/component: network
+        kubernetes.io/os: "linux"
+    spec:
+      serviceAccountName: ovn-kubernetes-node
+      hostNetwork: true
+      hostPID: true
```
squeed, line 27: If at all possible, avoiding hostPID is good. I know it's not always easy to do with openvswitch.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: It's not necessary, but it does mean fewer things could go wrong on startup. If you're confident that cert rotations are foolproof, then you can ignore this suggestion. If there's a chance they cause issues - the fewer we do, the better :-)

squeed:
I think I'd like to see the key-setup part as an InitContainer - it fits a bit more logically with what I would expect. Then, start and monitor the ipsec processes in a normal container. Make sense?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Ok, based on Casey's comment, I will keep this.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: On second thoughts, I think for the moment this is sufficient as this will simultaneously test if ovs-monitor-ipsec and libreswan are running.

pkg/network/ovn_kubernetes.go
markdgray: I'll keep it in because I made a change to that needs this now!

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Thats not a bet I want to take :) I will check for cert end-date validity. Apart from anything else it will speed things up significantly.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Yeah, I was thinking about "PostStartHook" yesterday evening and I have implemented that successfully.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Just checking this, the way "ovs-monitor-ipsec" works is that it monitors OVSDB for changes and acts upon them. In the case of certificates, when it sees the certificate has been added, it will import that cert into NSS. It will then only update the cert if the location changes or if "ovs-monitor-ipsec" is restarted. As we only update the keys when restarting the pod, neither of these conditions will trigger so we are safe to do this here.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I think xfrm is built-in in this case as it seems to work and I dont see any xfrm modules when i do an `lsmod`

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+            spec:
+              request: ${csr_64}
+              signerName: network.openshift.io/signer
+              usages:
+              - ipsec tunnel
+          EOF
+
+            counter=0
+            until [ ! -z $(kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' 2>/dev/null) ]
+            do
```
squeed, line 78: Probably want a sleep here.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+            kubectl delete csr/$(hostname)
+
+            cat /signer-ca/tls.crt | openssl x509 -outform pem -text > /etc/openvswitch/keys/ipsec-cacert.pem
+          fi
+
+          # Updating the cerificates does not need to be an atomic operation as
+          # the will get read and loaded into NSS by the ovs-monitor-ipsec process
+          # which has not started yet.
+          ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=/etc/openvswitch/keys/ipsec-cert.pem \
```
squeed, line 97: Does this do anything if we've rotated the certificate, but the filenames haven't changed?

markdgray: No

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+        beta.kubernetes.io/os: "linux"
+      volumes:
+      - name: host-var-log-ovs
+        hostPath:
+          path: /var/log/openvswitch
+      - name: host-var-run-ovs
+        hostPath:
+          path: /var/run/openvswitch
+      - name: signer-ca
+        configMap:
```
markdgray, line 200: Note: This is a configMap now. This configMap is populated by the certificate rotation controller with the CA cert. Therefore we are not sharing any private keys between hosts.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: Right, that makes sense - do we need to do anything special to signal the processes to reload? Or just hope the node reboots in the next 6 months?

Maybe ovs just watches the file, in which case that's obviously fine too.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+          # Every time we restart this container, we will create a new key pair
+          echo "Configuring IPsec keys"
+
+          # If the certificate does not exist or it will expire in the next 6 months
+          # (15770000 seconds), we will generate a new one.
+          if [ ! -e /etc/openvswitch/keys/ipsec-cert.pem ] || [ ! openssl x509 -noout -dates -checkend 1999999999 ];
+          then
+            # We use the system-id as the CN. This is a requirement for OVN
+            cn=$(ovs-vsctl --retry -t 60 get Open_vSwitch . external-ids:system-id | tr -d "\"")
```
abhat, line 46: so this will wait for potentially 60s when ovs pod for some reason is down? Should we check if ovs is running before checking for system-id for CN?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,207 @@
...
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -exuo pipefail
+
+          # Every time we restart this container, we will create a new key pair
+          echo "Configuring IPsec keys"
+
+          # If the certificate does not exist or it will expire in the next 6 months
```
abhat, line 41: Can you add a comment describing the overall steps here? Who is responsible for approving the CSR that you are creating here etc. is missing?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
abhat: Again, some comments that describes the order of operations will be useful.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I think this is OK. This will (also) check for OVS by retrying the db conection for 60 seconds. I could extend it?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Ok

markdgray:
> > As per https://github.com/openshift/api/pull/783/files#r528769454, consider the `Enable` flag based on the `IPsecConfig` being present. If the IPsecConfig is empty struct, it can be interpreted as the feature is disabled.
> 
> @abhat [openshift/api#794](https://github.com/openshift/api/pull/794) I will update this PR when/if this gets accepted

I have merged in the API changes.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: As this is an init container now, if we restart this container the other container should restart and force a reload. The issue now is how do we force a reload of this container .. I was originally thinking of using the readiness probe to check for certificate expiry but I can't do that anymore.

markdgray: This is currently by the way. I could implement something more sophisticated in ovs-monitor-ipsec.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I had to look into this more because now I have a working openshift cluster and I could do some testing and investigation. Currently, both libreswan and ovs-monitor-ipsec do not allow a clean restart without flushing flows from xfrm (directly by libreswan but indirectly by ovs-monitor-ipsec - which is unpleasant. This functionality is in progress in libreswan and easy to implement in ovs. We can uncleanly exit (i.e. kill) both and this should work but its not ideal.

I guess the other option is to move libreswan to the host and do what was done with OVS. In this way, a container would never restart libreswan. ovs-monitor-ipsec could be modified to exit without flushing anything. Not sure how much effort it is to move libreswan to the host?

markdgray: Looking at https://github.com/libreswan/libreswan/issues/385 and seeing what was done with OVS in CNO before it moved to the host, perhaps restarting ovs-monitor-ipsec and pluto ungracefully will be ok. When the libreswan and ovs-monitor-ipsec features get upstreamed, I can modify to a cleaner solution. What do you think?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: @abhat @squeed Updated this

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: OVS should never be down; it's a systemd service. So I think this is fine.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,246 @@
...
+                        -addext "subjectAltName = DNS:${cn}" \
+                        -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
+                        -key /etc/openvswitch/keys/ipsec-privkey.pem \
+                        -out /etc/openvswitch/keys/ipsec-req.pem
+
+            csr_64=$(cat /etc/openvswitch/keys/ipsec-req.pem | base64 | tr -d "\n")
+
+            # The signer controller does not allow re-signing a key. We will
+            # delete the old key to be sure it is not there
+            kubectl delete --ignore-not-found=true csr/$(hostname)
```
squeed, line 71: oh duh, this should be `${K8S_NODE}`, not `$(hostname)`.

squeed: In all places in this script.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,246 @@
...
+          # Configure OVS with the relevant keys for this node. This is required by ovs-monitor-ipsec.
+          #
+          # Updating the certificates does not need to be an atomic operation as
+          # the will get read and loaded into NSS by the ovs-monitor-ipsec process
+          # which has not started yet.
+          ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=/etc/openvswitch/keys/ipsec-cert.pem \
+                                                     other_config:private_key=/etc/openvswitch/keys/ipsec-privkey.pem \
+                                                     other_config:ca_cert=/etc/openvswitch/keys/ipsec-cacert.pem
+        env:
+        - name: OVS_LOG_LEVEL
```
squeed, line 122: This seems unused (which is fine, just get rid of it).

markdgray: Yeah, in this container I think so.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,246 @@
...
+              #!/bin/bash
+              ovs-appctl -t ovs-monitor-ipsec ipsec/status
+          initialDelaySeconds: 120
+          periodSeconds: 30
+          timeoutSeconds: 40
+        terminationGracePeriodSeconds: 10
+      nodeSelector:
+        beta.kubernetes.io/os: "linux"
+      volumes:
+      - name: host-var-log-ovs
```
squeed, line 232: All the `hostPath` volumes should have `type: Directory`.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -0,0 +1,246 @@
...
+          path: /var/log/openvswitch
+      - name: host-var-run-ovs
+        hostPath:
+          path: /var/run/openvswitch
+      - name: signer-ca
+        configMap:
+          name: signer-ca
+      - name: etc-openvswitch
+        hostPath:
+          path: /var/lib/openvswitch/etc
```
squeed, line 243: Note that this isn't `/etc/openvswitch` on the host. Is that OK for this use case?

I did this because `ovn-nbdb` was trying to write to `/etc/openvswitch` and getting selinux issues. But these are ovs components, not ovn.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,221 @@
...
+	// Only handle CSRs for this signer
+	if csr.Spec.SignerName != signerName {
+		// Don't handle a CSR for another signerName. We don't need to log this as
+		// we will pollute the logs. We also don't need to requeue it.
+		return reconcile.Result{}, nil
+	}
+
+	if isCertificateRequestApproved(csr) {
+		// We will not recertify an approved certificate
+		return reconcile.Result{}, nil
```
squeed, line 120: What happens if we approve a CSR then crash?

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,221 @@
...
+	if len(csr.Status.Certificate) != 0 {
+		// Request already has a certificate. Nothing to do as we will, currently,
+		// not re-certify so don't requeue.
+		log.Printf("Cannot re-sign %s for %s. Please delete and re-create resource", request.Name, signerName)
+		return reconcile.Result{}, nil
+	}
+
+	// Get our CA that was created by the operatorpki.
+	caSecret := &corev1.Secret{}
+	r.client.Get(context.TODO(), types.NamespacedName{Namespace: "openshift-ovn-kubernetes", Name: "signer-ca"}, caSecret)
```
squeed, line 148: This could fail (if the OperatorPKI is behind for some reason), so we need to handle that (return error and it should automatically retry).

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// Decode the certificate request from PEM format.
+	certReq, err := decodeCertificateRequest(csr.Spec.Request)
+	if err != nil {
```
squeed, line 152: As a follow-up, all of these failures should be set in the CSR's `Status.Conditions` array. Given that this is internal API, it's not blocking.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
squeed: After talking with @markdgray, disregard this - the hostname is what matters.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I just need a persistent host directory somewhere. I think I can leave this.

squeed: (I don't know the right answer, just calling your attention to this).

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,221 @@
...
+		csr.Status.Conditions = append(csr.Status.Conditions, csrv1.CertificateSigningRequestCondition{
+			Type:    csrv1.CertificateApproved,
+			Status:  "True",
+			Reason:  "AutoApproved",
+			Message: "Automatically approved by " + signerName})
+		// Update status to "Approved"
+		csr, err = r.clientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.TODO(), request.Name, csr, metav1.UpdateOptions{})
+		if err != nil {
+			log.Printf("Could not update CSR status: %v", err)
+			return reconcile.Result{}, nil
```
squeed, line 135: Probably want to return `err` here, so we retry.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+	}
+
+	if len(csr.Status.Certificate) != 0 {
+		// Request already has a certificate. Nothing to do as we will, currently,
+		// not re-certify so don't requeue.
+		log.Printf("Cannot re-sign %s for %s. Please delete and re-create resource", request.Name, signerName)
```
squeed, line 142: Note that this will be called every time CNO restarts, so you might want a different log line. Something like
"Certificate %s already signed; skipping".

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: None of the other manifests that I looked at specified this directory type? I will make this change.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: It should be ok. The host OVS instance doesnt need the files that we put in this directory. ovn-vsctl sets the string in ovsdb, and then ovs-monitor-ipsec reads that string. As long as this is persistent, it could be anywhere. I think I will leave it.

pkg/controller/signer/signer-controller.go
markdgray: I updated this code. Because the watch doesn't have predicates, all updates, creates and deletes get handled by the reconciler. As I update the CSR twice in the flow (once for adding the approval and again for adding the certificate), there needs to be a few guards there to not process the updates.

pkg/controller/signer/signer-controller.go
markdgray: Added an error condition

pkg/controller/signer/signer-controller.go
markdgray: This was a bit of a job but I cleaned up the error handling. All errors print to stdout, most update the csr status conditions, and many also update the status manager (the ones that are signer-controller errors)

pkg/controller/signer/signer-controller.go
markdgray: Ok

pkg/controller/signer/signer-controller.go
markdgray: Done

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: Also it will need to be DirectoryOrCreate because in some cases (Kind), the directory does not exist on the host.

markdgray:
@abhat @squeed @alexanderConstantinescu All remaining comments resolved. Also rebased and revendored on top of @squeed 's update.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,269 @@
...
+		// Don't handle a CSR for another signerName. We don't need to log this as
+		// we will pollute the logs. We also don't need to requeue it.
+		return reconcile.Result{}, nil
+	}
+
+	if len(csr.Status.Certificate) != 0 {
+		// Request already has a certificate. There is nothing
+		// to do as we will, currently, not re-certify or handle any updates to
+		// CSRs. We can ignore this CSR update.
+		updateCSRStatusConditions(r, csr, "RecertifyFailure",
```
squeed, line 134: This is probably not an error - it means the certificate is signed and we're done. Just return and move on.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,269 @@
...
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("signer-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to CetificateSigningRequest resource
+	err = c.Watch(&source.Kind{Type: &csrv1.CertificateSigningRequest{}}, &handler.EnqueueRequestForObject{},
+		ignoreUpdatePredicate())
```
squeed, line 70: I wouldn't do this - controllers **must** be level-triggered.

pkg/controller/signer/signer-controller.go
squeed: Multiple reconciliations are fine, as long as you're level-triggered.

You need to handle objects in all possible states, so if a certificate is approved, then we move on and sign it.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		signerFailure(r, csr, "EncodeFailure",
+			fmt.Sprintf("Could not encode certificate: %v", err))
+		return reconcile.Result{}, nil
+	}
+
+	err = r.client.Status().Update(context.TODO(), csr)
+	if err != nil {
+		log.Printf("Could not update CSR status: %v", err)
+		r.status.SetDegraded(statusmanager.CertificateSigner, "UpdateFailure",
```
squeed, line 219: For this error, we want to retry the reconcile, so **don't** set degraded - instead, just return `err`.

Yeah, I know this is tricky.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,269 @@
...
+			fmt.Sprintf("Unable to update signed certificate for %v and signer %v: %v", request.Name, signerName, err))
+		return reconcile.Result{}, nil
+	}
+
+	log.Printf("Certificate signed, issued and approved for %s by %s", request.Name, signerName)
```
squeed, line 224: Make sure to un-set degraded in this case.

pkg/controller/signer/signer-controller.go
```diff
@@ -0,0 +1,269 @@
...
+	// certificate signing request to this signer is automatically approved. This
+	// is somewhat protected by permissions on the CSR resource.
+	// TODO: We may need a more robust way to do this later
+	csr.Status.Conditions = append(csr.Status.Conditions, csrv1.CertificateSigningRequestCondition{
+		Type:    csrv1.CertificateApproved,
+		Status:  "True",
+		Reason:  "AutoApproved",
+		Message: "Automatically approved by " + signerName})
+	// Update status to "Approved"
+	csr, err = r.clientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(context.TODO(), request.Name, csr, metav1.UpdateOptions{})
```
squeed, line 157: This would also be a transient failure, so return `err`

pkg/controller/signer/signer-controller.go
markdgray: Then why does the controller-runtime allow for it?

pkg/controller/signer/signer-controller.go
markdgray: Good point.

pkg/controller/signer/signer-controller.go
markdgray: I think you mean "only" return `err`. Ill update.

pkg/controller/signer/signer-controller.go
squeed: No clue, but it's definitely not recommended.

Besides, you'll get an `ADD` event every time the process is restarted, regardless of the state of the object.

pkg/controller/signer/signer-controller.go
squeed: As a rule, a temporary apiserver blip doesn't need to be degraded (but it could be...). We mark degraded when there's something we'd like to report to the administrator.

squeed: Clear as mud, I realize.

squeed:
Handling all controller states is tricky - and it doesn't need to be perfect, since this isn't a user-facing API. In other words, we don't really need to worry about every possible invalid state. But we should handle the case where we find an approved, unsigned certificate.

pkg/controller/signer/signer-controller.go
markdgray: Updated, if we approve and crash, it will attempt to resign.

pkg/controller/signer/signer-controller.go
markdgray: Done

markdgray:
> Handling all controller states is tricky - and it doesn't need to be perfect, since this isn't a user-facing API. In other words, we don't really need to worry about every possible invalid state. But we should handle the case where we find an approved, unsigned certificate.

Also done

markdgray:
> As per https://github.com/openshift/api/pull/783/files#r528769454, consider the `Enable` flag based on the `IPsecConfig` being present. If the IPsecConfig is empty struct, it can be interpreted as the feature is disabled.

Done

dcbw:
/test e2e-aws-ovn-windows
 msg=Error: Error creating VPC: VpcLimitExceeded: The maximum number of VPCs has been reached.

dcbw:
not supposed to be required yet https://github.com/openshift/release/pull/14068

PR #887, opened 2020-11-23
Bug 1948592: Add Egress Router Controller

danielmellado:
This PR adds the egress router controller for managing such objects.
It'll be cherry-picking #852, which adds CRD/Types for egress routers
objects. It's taking PKI Controller as a model to be built upon.

Signed-off-by: Daniel Mellado <dmellado@redhat.com>

### DIFF
--- a/pkg/controller/egress_router/egress_router_controller.go
+++ b/pkg/controller/egress_router/egress_router_controller.go
@@ -1,0 +1,216 @@
+package egress_router
+
+// egress router implements a controller for the egress router CNI plugin
+
+import (
+	"context"
+	"fmt"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"net"
+	"os"
+
+	"github.com/openshift/cluster-network-operator/pkg/apply"
+	"github.com/openshift/cluster-network-operator/pkg/render"
+	"github.com/pkg/errors"
+
+	"path/filepath"
+	"reflect"
+	"strings"
+	"time"
+
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+
+	netopv1 "github.com/openshift/api/networkoperator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// Attach control loop to the manager and watch for Egress Router objects
+func Add(mgr manager.Manager, status *statusmanager.StatusManager) error {
+	r, err := newEgressRouterReconciler(mgr, status)
+	if err != nil {
+		return err
+	}
+
+	// Create a new controller
+	c, err := controller.New("egress-router-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to primary resource EgressRouter.network.operator.openshift.io/v1
+	err = c.Watch(&source.Kind{Type: &netopv1.EgressRouter{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &EgressRouterReconciler{}
+var manifestDir = "bindata/"
+
+type egressrouter struct {
+	spec netopv1.EgressRouterSpec
+}
+
+type EgressRouterReconciler struct {
+	mgr    manager.Manager
+	client client.Client
+	status *statusmanager.StatusManager
+
+	egressrouters    map[types.NamespacedName]*egressrouter
+	egressrouterErrs map[types.NamespacedName]error
+}
+
+var ResyncPeriod = 5 * time.Minute
+
+func newEgressRouterReconciler(mgr manager.Manager, status *statusmanager.StatusManager) (reconcile.Reconciler, error) {
+
+	return &EgressRouterReconciler{
+		mgr:    mgr,
+		status: status,
+		client: mgr.GetClient(),
+
+		egressrouters:    map[types.NamespacedName]*egressrouter{},
+		egressrouterErrs: map[types.NamespacedName]error{},
+	}, nil
+}
+
+func (r EgressRouterReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	klog.Infof("Reconciling egressrouter.network.operator.openshift.io %s\n", request.NamespacedName)
+
+	obj := &netopv1.EgressRouter{}
+	err := r.mgr.GetClient().Get(context.TODO(), request.NamespacedName, obj)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Info("Egress Router %s seems to have been deleted\n", request.NamespacedName)
+			return reconcile.Result{}, nil
+		}
+		klog.Error(err)
+		return reconcile.Result{}, err
+	}
+
+	// Check to see if we already know this object
+	existing := r.egressrouters[request.NamespacedName]
+	if existing != nil {
+		if !reflect.DeepEqual(obj.Spec, existing.spec) {
+			klog.Infof("Egress Router %s has changed, refreshing\n", request.NamespacedName)
+			delete(r.egressrouters, request.NamespacedName)
+			existing = nil
+		}
+	}
+
+	if existing == nil {
+		klog.Infof("Creating a new Egress Router")
+		// Set owner reference to the controller
+		boolTrue := bool(true)
+		EgressRouterOwnerReferences := []v1.OwnerReference{
+			{
+				APIVersion: "network.operator.openshift.io/v1",
+				Kind:       "EgressRouter",
+				Name:       obj.Name,
+				UID:        obj.UID,
+				Controller: &boolTrue,
+			},
+		}
+		err := r.ensureEgressRouter(manifestDir, request.Namespace, obj, EgressRouterOwnerReferences)
+
+		if err != nil {
+			klog.Error(err)
+			r.egressrouterErrs[request.NamespacedName] =
+				errors.Wrapf(err, "could not reconcile Egress Router %s", request.NamespacedName)
+			r.setStatus()
+			return reconcile.Result{}, err
+		}
+
+		r.egressrouters[request.NamespacedName] = existing
+	}
+
+	if err != nil {
+		klog.Error(err)
+		r.egressrouterErrs[request.NamespacedName] =
+			errors.Wrapf(err, "could not reconcile Egress Router %s", request.NamespacedName)
+		r.setStatus()
+		return reconcile.Result{}, err
+	}
+
+	klog.Infof("successful reconciliation")
+	delete(r.egressrouterErrs, request.NamespacedName)
+	r.setStatus()
+	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
+}
+
+// setStatus summarizes the status of all Egress Router objects and updates the statusmanager
+// as appropriate.
+func (r *EgressRouterReconciler) setStatus() {
+	if len(r.egressrouterErrs) == 0 {
+		r.status.SetNotDegraded(statusmanager.EgressRouterConfig)
+	} else {
+		msgs := []string{}
+		for _, e := range r.egressrouterErrs {
+			msgs = append(msgs, e.Error())
+		}
+		r.status.SetDegraded(statusmanager.EgressRouterConfig, "EgressRouterError", strings.Join(msgs, ", "))
+	}
+}
+
+func (r *EgressRouterReconciler) ensureEgressRouter(manifestDir string, namespace string, router *netopv1.EgressRouter, EgressRouterOwnerReferences []v1.OwnerReference) error {
+	if len(router.Spec.Addresses) == 0 {
+		return fmt.Errorf("Error: router without addresses")
+	}
+	out := []*uns.Unstructured{}
+	data := render.MakeRenderData()
+	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+	data.Data["EgressRouterNamespace"] = namespace
+	if isItValidCidr(string(router.Spec.Addresses[0].IP)) {
+		data.Data["Addresses"] = router.Spec.Addresses[0].IP
+	}
+	if isItValidIPAddress(router.Spec.Addresses[0].Gateway) {
+		data.Data["Gateway"] = router.Spec.Addresses[0].Gateway
+	}
+	data.Data["AllowedDestinations"] = router.Spec.Redirect.RedirectRules
+	data.Data["mode"] = router.Spec.Mode
+	data.Data["network_interfaces"] = router.Spec.NetworkInterface
+	data.Data["EgressRouterPodImage"] = os.Getenv("EGRESS_ROUTER_CNI_IMAGE")
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "egress-router"), &data)
+	if err != nil {
+		return err
+	}
+	out = append(out, manifests...)
+
+	for _, obj := range out {
+		klog.Infof("Assigning owner references")
+		obj.SetOwnerReferences(EgressRouterOwnerReferences)
+		klog.Infof("Applying manifest")
+		if err := apply.ApplyObject(context.TODO(), r.client, obj); err != nil {
+			klog.Infof("could not apply egress router object: %v", err)
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isItValidCidr(cidr string) bool {
+	_, _, err := net.ParseCIDR(cidr)
+	if err != nil {
+		klog.Error(err)
+		return false
+	}
+	return true
+}
+
+func isItValidIPAddress(ip string) bool {
+	return net.ParseIP(ip) != nil
+}
--- a/bindata/egress-router/000-nad.yaml
+++ b/bindata/egress-router/000-nad.yaml
@@ -1,0 +1,25 @@
+---
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: egress-router-cni-nad
+  namespace: "{{.EgressRouterNamespace}}"
+  annotations:
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  config: '{
+    "cniVersion": "0.4.0",
+    "type": "egress-router",
+    "name": "egress-router-cni-nad",
+    "ip": {
+      "addresses": [
+        "{{.Addresses}}"
+        ],
+      "destinations": [
+        "{{.AllowedDestinations}}"
+        ],
+      "gateway": "{{.Gateway}}"
+        },
+      "log_file": "/tmp/egress-router-log",
+      "log_level": "debug"
+    }'
--- a/bindata/egress-router/001-deployment.yaml
+++ b/bindata/egress-router/001-deployment.yaml
@@ -1,0 +1,25 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: egress-router-cni-deployment
+  namespace: "{{.EgressRouterNamespace}}"
+  annotations:
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+  labels:
+    app: egress-router-cni
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: egress-router-cni
+  template:
+    metadata:
+      labels:
+        app: egress-router-cni
+      annotations:
+        k8s.v1.cni.cncf.io/networks: egress-router-cni-nad
+    spec:
+      containers:
+        - name: egress-router-cni-pod
+          image: "{{.EgressRouterPodImage}}"
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -21,6 +21,7 @@ import (
 	operatorv1helpers "github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	configv1 "github.com/openshift/api/config/v1"
+	op_netopv1 "github.com/openshift/api/networkoperator/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	netopv1 "github.com/openshift/cluster-network-operator/pkg/apis/network/v1"
 	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
@@ -103,6 +104,9 @@ func New(cfg, protocfg *rest.Config) (*Client, error) {
 		log.Fatal(err)
 	}
 	if err := machineapi.AddToScheme(c.Scheme()); err != nil {
+		log.Fatal(err)
+	}
+	if err := op_netopv1.Install(c.Scheme()); err != nil {
 		log.Fatal(err)
 	}
 
--- a/pkg/controller/add_networkconfig.go
+++ b/pkg/controller/add_networkconfig.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"github.com/openshift/cluster-network-operator/pkg/controller/clusterconfig"
 	configmapcainjector "github.com/openshift/cluster-network-operator/pkg/controller/configmap_ca_injector"
+	"github.com/openshift/cluster-network-operator/pkg/controller/egress_router"
 	"github.com/openshift/cluster-network-operator/pkg/controller/ingressconfig"
 	"github.com/openshift/cluster-network-operator/pkg/controller/operconfig"
 	"github.com/openshift/cluster-network-operator/pkg/controller/pki"
@@ -14,6 +15,7 @@ func init() {
 	// AddToManagerFuncs is a list of functions to create controllers and add them to a manager.
 	AddToManagerFuncs = append(AddToManagerFuncs,
 		pki.Add,
+		egress_router.Add,
 		proxyconfig.Add,
 		operconfig.Add,
 		clusterconfig.Add,
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -38,6 +38,7 @@ const (
 	InjectorConfig
 	PodDeployment
 	PKIConfig
+	EgressRouterConfig
 	RolloutHung
 	CertificateSigner
 	maxStatusLevel


juanluisvaladas:
@danielmellado shouldn't this be part of openshift/api as well? If so I think the right thing to do would be adding it to openshift/api and consume it from there, like @rcarrillocruz does with this https://github.com/openshift/cluster-network-operator/pull/875 .

@rcarrillocruz opinions?

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,128 @@
+# This file is automatically generated. DO NOT EDIT
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.4.0
+    include.release.openshift.io/ibm-cloud-managed: "true"
```
juanluisvaladas, line 7: What is this annotation for?

danielmellado:
> @danielmellado shouldn't this be part of openshift/api as well? If so I think the right thing to do would be adding it to openshift/api and consume it from there, like @rcarrillocruz does with this #875 .
> 
> @rcarrillocruz opinions?

oh, I see your point! Well, I'm explicitly waiting also on ricky's patch in order to avoid duplicating this to openshift/api, so no worries.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,128 @@
...
+                        and gateway to be configured on the router's interface
+                      properties:
+                        gateway:
+                          description: IP address of the next-hop gateway, if it cannot
+                            be automatically determined. Can be IPv4 or IPv6.
+                          type: string
+                        ip:
+                          description: IP is the address to configure on the router's
+                            interface. Can be IPv4 or IPv6.
+                          type: string
```
juanluisvaladas, line 66: If addresses allows a list, and this list may include both ipv4 and ipv6 addresses at the same time, then we need to be able to specify an object that allows both gateways rather than a string.
Also a pattern to validate would be nice:
https://github.com/openshift/cluster-network-operator/blob/f47200e8dbc052501e79192a5b7933e1195b1b91/bindata/network/openshift-sdn/001-crd.yaml#L48

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,128 @@
...
+                      - ip
+                      type: object
+                    type: array
+                  allowedDestinations:
+                    description: List of CIDR blocks that the pod is allowed to connect
+                      to via this interface. If not provided, the pod can connect
+                      to any destination.
+                    items:
+                      description: CIDR represents an IPv4 or IPV6 CIDR block.
+                      type: string
```
juanluisvaladas, line 77: Again a pattern would be nice, you can use this one for IPv4.
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/001-crd.yaml#L48

Pretty sure there is some regex somewhere for the IPv6 one.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
```diff
@@ -0,0 +1,128 @@
...
+                    description: Arguments specific to the interfaceType macvlan
+                    properties:
+                      master:
+                        description: Name of the master inteface. Usually used for
+                          macvlan and ipvlan. Need not be specified if it can be inferred
+                          from the IP address.
+                        minLength: 1
+                        type: string
+                      mode:
+                        description: Usually used for macvlan interface.
```
juanluisvaladas, line 95: Since mode doesn't take any random value, we should document explicitly which are the values allowed. It's extremely annoying when oc explain doesn't do this. Is there a default value?

Also it would be great if we could add a pattern here to make sure we only allow valid values.

manifests/0000_70_cluster-network-operator_01_pki_crd.yaml
```diff
@@ -20,21 +20,38 @@ spec:
...
+          signed by the CA \n The CA certificate will have a validity of 10 years,
+          rotated after 9. The target certificate will have a validity of 6 months,
+          rotated after 3 \n The CA certificate will have a CommonName of \"<namespace>_<name>-ca@<timestamp>\",
+          where <timestamp> is the last rotation time."
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
```
juanluisvaladas, line 40: Not that I really care, but since we're splitting into shorter lines, probably the More info link makes the line too long and should be yet another line.

pkg/apis/network/v1/egressrouter_types.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// VLANConfig consists of arguments specific to the EgressRouterInterfaceType
+// +kubebuilder:validation:Required
+type VLANConfig struct {
+	// Usually used for macvlan interface.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Optional
+	// +optional
+	Mode string `json:"mode,omitempty"`
```
juanluisvaladas, line 65: Again the mode isn't a random string, so it should be a type, the deploymentConfig is a good example:

https://github.com/openshift/api/blob/992bc3a41fe6c4e5bcf27dc8080f9e5c114c33fc/apps/v1/types.go#L110-L119

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+var _ reconcile.Reconciler = &EgressRouterReconciler{}
+
+type EgressRouterReconciler struct {
```
juanluisvaladas, line 52: If it's public it should have a comment with the description for godoc.
I know we already have a jillion places where we don't do it, but still.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+	mgr       manager.Manager
+	clientset *kubernetes.Clientset
+	status    *statusmanager.StatusManager
+
+	egressrouters map[types.NamespacedName]*egressrouter
+	egressrouterErrs map[types.NamespacedName]error
+}
+
+var ResyncPeriod = 5 * time.Minute
```
juanluisvaladas, line 61: I don't think this needs to be public? And if it needs to be public it should have a description

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+		clientset: clientset,
+
+		egressrouters:    map[types.NamespacedName]*egressrouter{},
+		egressrouterErrs: map[types.NamespacedName]error{},
+
+	}, nil
+}
+
+func (r EgressRouterReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Printf("Reconciling egressrouter.network.operator.openshift.io %s\n", request.NamespacedName)
```
juanluisvaladas, line 81: I think this should be an klog.Infof instead.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+
+	obj := &netopv1.EgressRouter{}
+	err := r.mgr.GetClient().Get(context.TODO(), request.NamespacedName, obj)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Printf("Egress Router %s seems to have been deleted\n", request.NamespacedName)
+			return reconcile.Result{}, nil
+		}
+		log.Println(err)
```
juanluisvaladas, line 91: klog.Errorf

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+// egress router imnplements a controller for the egress router CNI plugin
+
+import (
+	"context"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/certrotation"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"reflect"
+	"log"
```
juanluisvaladas, line 12: Use klog for consistency.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+			r.setStatus()
+			return reconcile.Result{}, err
+		}
+		r.egressrouters[request.NamespacedName] = existing
+	}
+	err = existing.sync()
+	if err != nil {
+		log.Println(err)
+		r.egressrouterErrs[request.NamespacedName] =
+			errors.Wrapf(err, "could not reconcile PKI %s", request.NamespacedName)
```
juanluisvaladas, line 120: You forgot to change this message. Shouldn't be PKI

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,159 @@
...
+		status:    status,
+		clientset: clientset,
+
+		egressrouters:    map[types.NamespacedName]*egressrouter{},
+		egressrouterErrs: map[types.NamespacedName]error{},
+
+	}, nil
+}
+
+func (r EgressRouterReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
juanluisvaladas, line 80: Again public function without  description

juanluisvaladas (changes requested):

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
squeed: magic. well, not really. part of https://github.com/openshift/cluster-version-operator/pull/404

squeed:
Hey @juanluisvaladas just FYI, we're reviewing API / CRD in #852

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
danielmellado: As commented on the link casey's reflecting, we'll handle such check on the controller code

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
danielmellado: ditto

pkg/controller/egress_router/egress_router_controller.go
danielmellado: will change all logs to klog overall, thanks!

squeed:
Seems basically reasonable. Time to start making some resources :-)

Do you intend to template the resources in `bindata` or describe them in go directly?

danielmellado:
Pull-request updated, HEAD is now e10d2e1bc515f9d1f8fa3139b3b8915b67ec638b

danielmellado:
Pull-request updated, HEAD is now e808e7c8c2090f00b66c1e0959da13900ed3c15b

danielmellado:
Pull-request updated, HEAD is now a9a9e2a9a04a35afd05cd2535d1097567970a31b

bindata/network/egress-router/000-nad.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: egress-router
+spec:
+  config: '{
+    "cniVersion": "0.4.0",
+    "type": "egress-router",
+    "name": "egress-router-config",
+    "ip": {
+      "addresses": [
+        "192.168.10.99/24"
```
squeed, line 13: This is just for testing, right?

danielmellado:
Yep, but to check that it would create that. I'll replace for a template
and real data afterwards.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+}
+
+type egressrouter struct {
+	spec       netopv1.EgressRouterSpec
+	controller factory.Controller
+	lastErr    error
+}
+
+func ensureEgressRouter(manifestDir string) ([]*uns.Unstructured, error) {
```
squeed, line 159: Presumably you need to pass through the namespace as well.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		if err != nil {
+			log.Println(err)
+			r.egressrouterErrs[request.NamespacedName] =
+				errors.Wrapf(err, "could not reconcile Egress Router %s", request.NamespacedName)
+			r.setStatus()
+			return reconcile.Result{}, err
+		}
+		r.egressrouters[request.NamespacedName] = existing
```
squeed, line 122: You don't ever set `existing` to anything, so it will always be `nil` here.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+
+func (r EgressRouterReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	log.Printf("Reconciling egressrouter.network.operator.openshift.io %s\n", request.NamespacedName)
+
+	obj := &netopv1.EgressRouter{}
+	err := r.mgr.GetClient().Get(context.TODO(), request.NamespacedName, obj)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Printf("Egress Router %s seems to have been deleted\n", request.NamespacedName)
```
squeed, line 90: If the egress router is deleted, you should clean things up - delete any local references, and (perhaps) delete any kubernetes resources. However, the easiest way to delete kubernetes resources is to set the OwnerReference.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+			delete(r.egressrouters, request.NamespacedName)
+			existing = nil
+		}
+	}
+
+	if existing == nil {
+		log.Printf("Creating a new Egress Router")
+		objs, err := ensureEgressRouter(manifestDir)
+		for _, obj := range objs {
+			log.Printf("Applying manifest")
```
squeed, line 111: Need to set the OwnerReference of each object here. It should point to the EgressRouter object that caused this to be created.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+			apply.ApplyObject(context.TODO(), r.client, obj)
```
squeed, line 112: This returns an error, you need to handle that.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	} else {
+		msgs := []string{}
+		for _, e := range r.egressrouterErrs {
+			msgs = append(msgs, e.Error())
+		}
+		r.status.SetDegraded(statusmanager.EgressRouterConfig, "EgressRouterError", strings.Join(msgs, ", "))
+	}
+}
+
+type egressrouter struct {
```
squeed, line 153: You never create any instances of this type. What does it do?

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	data := render.MakeRenderData()
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network/egress-router"), &data)
+	if err != nil {
+		return nil, err
+	}
+	out = append(out, manifests...)
+	return out, nil
+}
+
+func (p *egressrouter) sync() error {
```
squeed, line 170: You don't need this function

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		for _, e := range r.egressrouterErrs {
+			msgs = append(msgs, e.Error())
+		}
+		r.status.SetDegraded(statusmanager.EgressRouterConfig, "EgressRouterError", strings.Join(msgs, ", "))
+	}
+}
+
+type egressrouter struct {
+	spec       netopv1.EgressRouterSpec
+	controller factory.Controller
```
squeed, line 155: and you definitely don't need a controller.

pkg/controller/egress_router/egress_router_controller.go
danielmellado: Why wouldn't it be needed? I assumed that it would run the controller itself, after checking the PKICotnroller

bindata/network/egress-router/000-nad.yaml
danielmellado: Yep, I'll change that for a template in the next iteration

pkg/controller/egress_router/egress_router_controller.go
danielmellado: so do you mean to set the ownerReference as metadata here so we can garbage collect it later?

pkg/controller/egress_router/egress_router_controller.go
danielmellado: it gets called in EgressRouterReconciler struct, pretty much as PKI does, what would you suggest in this case?

danielmellado:
Pull-request updated, HEAD is now a7486ebf6bba3a959fe28e9b0bf7e3484c847489

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+		objs, err := ensureEgressRouter(manifestDir, request.Namespace, obj)
+		for _, obj := range objs {
+			log.Printf("Assigning owner references")
+			obj.SetOwnerReferences(EgressRouterOwnerReferences)
+			log.Printf("Applying manifest")
+			if err := apply.ApplyObject(context.TODO(), r.client, obj); err != nil {
+				log.Printf("could not apply operator config: %v", err)
+				r.status.SetDegraded(statusmanager.EgressRouterConfig, "UpdateEgressRouterConfig",
+					fmt.Sprintf("Error while trying to update operator configuration: %v", err))
```
squeed, line 134: "Error while reconciling EgressRouter %s/%s"

bindata/network/egress-router/001-deployment.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
squeed, line 1: nit: this shouldn't be in `bindata/network/egress-router`, rather `bindata/egress-router`.

squeed:
What happens on upgrades? Are all connections broken?

We may wish to just, well, never upgrade unless absolutely necessary. This will require some careful thought around how the Deployment object is managed.

bindata/network/egress-router/001-deployment.yaml
danielmellado: why? every other network related manifest is under /network. tbh it feels like consistent to me. i.e. multus, kuryr...

danielmellado:
> What happens on upgrades? Are all connections broken?
> 
> We may wish to just, well, never upgrade unless absolutely necessary. This will require some careful thought around how the Deployment object is managed.

As we discussed, while that might be true, given the time constraings I just want to produce something that 'works'. Happy to iterate over this in another patch after this one gets in.

pkg/controller/egress_router/egress_router_controller.go
danielmellado: Changed

danielmellado:
Pull-request updated, HEAD is now 1ebed8a09e623fe3b7868ce76a7531fb09c7fdf2

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,212 @@
...
+	if isItValidCidr(string(router.Spec.Addresses[0].IP)) {
+		data.Data["Addresses"] = router.Spec.Addresses[0].IP
+	}
+	if isItValidIPAddress(router.Spec.Addresses[0].Gateway) {
+		data.Data["Gateway"] = router.Spec.Addresses[0].Gateway
+	}
+	data.Data["AllowedDestinations"] = router.Spec.AllowedDestinations
+	data.Data["mode"] = router.Spec.Mode
+	data.Data["network_interfaces"] = router.Spec.NetworkInterface
+	data.Data["EgressRouterPodImage"] = "centos/tools"
```
rcarrillocruz, line 204: I'd double-check if this is valid for release.
When I did the pod network connectivity check I was flagged for not using an openshift org image for doing curl tests, had to switch to openshift/hello-openshift.

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
rcarrillocruz: I have recently been blocked by apiserver team for not using regex on IP fields:

https://github.com/openshift/api/pull/846

If this is a cidr, I believe you can just change type: string for type: cidr, and that makes the validation (cidr does the regex validation).

bindata/network/egress-router/001-deployment.yaml
rcarrillocruz: Yep, I don't see why this should not be within network?

pkg/controller/egress_router/egress_router_controller.go
danielmellado: Thanks for checking that, just in case I'll make it match with the one in [1] (registry.redhat.com/openshift3/ose-pod)

[1] https://docs.openshift.com/container-platform/4.7/networking/ovn_kubernetes_network_provider/deploying-egress-router-ovn-redirection.html

danielmellado:
Pull-request updated, HEAD is now 3e35166ca9533c01f004ba3e3b49b49a6be25c70

manifests/0000_70_cluster-network-operator_01_egr_crd.yaml
danielmellado: Note that this has been quite changed, now its on https://github.com/openshift/api/pull/851 and stts has also mentioned that, so this whole PR is deending on that one, so don't consider this issue. Thanks!

pkg/controller/egress_router/egress_router_controller.go
danielmellado: this is also used like that in the PKIcontroller, take a look ;)

bindata/network/egress-router/001-deployment.yaml
squeed: because things in `/bindata/network/` are rendered by the network (operconfig) controller. This is a separate controller, and thus should have its own folder.

(If there's anything in `/bindata/network` that isn't rendered by operconfig, it should ultimately be moved)

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,211 @@
...
+				Controller: &boolTrue,
+			},
+		}
+		objs, err := ensureEgressRouter(manifestDir, request.Namespace, obj)
+		for _, obj := range objs {
+			log.Printf("Assigning owner references")
+			obj.SetOwnerReferences(EgressRouterOwnerReferences)
+			log.Printf("Applying manifest")
+			if err := apply.ApplyObject(context.TODO(), r.client, obj); err != nil {
+				log.Printf("could not apply operator config: %v", err)
```
squeed, line 133: This line is wrong - "could not apply egress router object:"

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,211 @@
...
+				r.status.SetDegraded(statusmanager.EgressRouterConfig, "UpdateEgressRouterConfig",
```
squeed, line 134: This will have no effect, because `egressroutererrs` will be empty, and thus will be erased in `r.setStatus()`

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,211 @@
...
+		EgressRouterOwnerReferences := []v1.OwnerReference{
+			{
+				APIVersion: "network.operator.openshift.io/v1",
+				Kind:       "EgressRouter",
+				Name:       obj.Name,
+				UID:        obj.UID,
+				Controller: &boolTrue,
+			},
+		}
+		objs, err := ensureEgressRouter(manifestDir, request.Namespace, obj)
```
squeed, line 127: you need to check the value of `err` before you look at `objs`

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,211 @@
...
+		return false
+	}
+	return true
+}
+
+func isItValidIPAddress(ip string) bool {
+	return net.ParseIP(ip) != nil
+}
+
+func ensureEgressRouter(manifestDir string, namespace string, router *netopv1.EgressRouter) ([]*uns.Unstructured, error) {
```
squeed, line 190: suggestion: ensureEgressRouter should also apply the object to the apisever. That makes error handling in the controller simpler.

squeed: thus, it would probably be easier to write this if it were a method on the controller.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,211 @@
...
+	if isItValidCidr(string(router.Spec.Addresses[0].IP)) {
+		data.Data["Addresses"] = router.Spec.Addresses[0].IP
+	}
+	if isItValidIPAddress(router.Spec.Addresses[0].Gateway) {
+		data.Data["Gateway"] = router.Spec.Addresses[0].Gateway
+	}
+	data.Data["AllowedDestinations"] = router.Spec.Redirect.RedirectRules
+	data.Data["mode"] = router.Spec.Mode
+	data.Data["network_interfaces"] = router.Spec.NetworkInterface
+	data.Data["EgressRouterPodImage"] = "registry.redhat.com/openshift3/ose-pod"
```
squeed, line 204: before this ships, this will have to change to a CVO-provided image for offline clusters

danielmellado: which image you'd be comfortable with regarding CVO usage?

pkg/controller/egress_router/egress_router_controller.go
danielmellado: I'll remove this

pkg/controller/egress_router/egress_router_controller.go
danielmellado: done

pkg/controller/egress_router/egress_router_controller.go
danielmellado: done

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -0,0 +1,212 @@
...
+		r.status.SetDegraded(statusmanager.EgressRouterConfig, "EgressRouterError", strings.Join(msgs, ", "))
+	}
+}
+
+func (r *EgressRouterReconciler) ensureEgressRouter(manifestDir string, namespace string, router *netopv1.EgressRouter, EgressRouterOwnerReferences []v1.OwnerReference) error {
+	out := []*uns.Unstructured{}
+	data := render.MakeRenderData()
+	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+	data.Data["EgressRouterNamespace"] = namespace
+	if isItValidCidr(string(router.Spec.Addresses[0].IP)) {
```
squeed, line 172: one last question: Does the CRD ensure that this array can't be empty? Otherwise this will panic.

pkg/controller/egress_router/egress_router_controller.go
danielmellado: it does, we also thought about that. Thanks for pointing it out.

pkg/controller/egress_router/egress_router_controller.go
danielmellado: it does, we also thought about that. Thanks for pointing it out.

squeed: Well, it says that the array is required, but it can be empty. So a quick length check would be a good idea.

Then we're good to go.

squeed:
At this point, this seems reasonable, and we can fix anything else in a follow-up PR.

/lgtm

pkg/controller/egress_router/egress_router_controller.go
tssurya: ya I did have a `is minLength=1 check` defined, but was told to remove it or something in one of the iterations since it was not standard.

squeed:
Neat, thanks for retesting those jobs that were already all green, prow

squeed:
This is insane.
/override ci/prow/e2e-gcp-ovn
/override ci/prow/e2e-gcp

PR #894, opened 2020-12-02
increasing logging when waiting for Masters during bootstrap for OVN

JacobTanenbaum:
Waiting 280 seconds for the bootstrap to fail without any timeout is too
long and does not reflect to the user what is going on appropriatly.

adding a log that prints out every 15 seconds while waiting for the
master nodes to become avalible telling the user how many have been seen
and how many are expected.

Also changed a typo boostrapOVN() -> bootstrapOVN() and changed the
timeouts to a const so that they can be changted and be reflected in the
new logging without additional changes

Signed-off-by: Jacob Tanenbaum <jtanenba@redhat.com>

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -21,6 +21,7 @@ import (
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	types "k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -32,6 +33,9 @@ const OVN_SB_RAFT_PORT = "9644"
 const CLUSTER_CONFIG_NAME = "cluster-config-v1"
 const CLUSTER_CONFIG_NAMESPACE = "kube-system"
 const OVN_CERT_CN = "ovn"
+
+const OVN_MASTER_DISCOVERY_TIMEOUT = 280
+const OVN_MASTER_DISCOVERY_POLL = 5
 
 // renderOVNKubernetes returns the manifests for the ovn-kubernetes.
 // This creates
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -242,12 +246,23 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
...
-	err := wait.PollImmediate(5*time.Second, 280*time.Second, func() (bool, error) {
+	var heartBeat int
+
+	err := wait.PollImmediate(OVN_MASTER_DISCOVERY_POLL, OVN_MASTER_DISCOVERY_TIMEOUT, func() (bool, error) {
 		matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
 		if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
 			return false, err
 		}
-		return len(masterNodeList.Items) != 0 && controlPlaneReplicaCount == len(masterNodeList.Items), nil
+		if len(masterNodeList.Items) != 0 && controlPlaneReplicaCount == len(masterNodeList.Items) {
```
dcbw, line 256: @JacobTanenbaum I would just do

```
if len(masterNodeList ..... {
    return true, nil
}

heartBeat++
<all the rest>
```

JacobTanenbaum: done

pkg/network/ovn_kubernetes.go
```diff
@@ -242,12 +246,23 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
...
+			return true, nil
+		}
+
+		heartBeat++
+		if heartBeat%3 == 0 {
+			klog.V(2).Infof("Wating to complete bootstrapping OVN: found (%d) master nodes out of (%d) expected: timing out in %d seconds",
```
dcbw, line 262: @JacobTanenbaum "Waiting to complete OVN bootstrap: found ..."

pkg/network/ovn_kubernetes.go
```diff
@@ -242,12 +246,23 @@ func boostrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
...
+				len(masterNodeList.Items), controlPlaneReplicaCount, int(OVN_MASTER_DISCOVERY_TIMEOUT)-int(OVN_MASTER_DISCOVERY_POLL)*5)
```
dcbw, line 263: @JacobTanenbaum shouldn't the OVN_MASTER_DISCOVERY_POLL be `heartBeat`?

dcbw (changes requested):

pkg/network/ovn_kubernetes.go
dcbw: Actually, maybe OVN_MASTER_DISCOVERY_* should just be ints (eg `const XXXX int = 5`) and then multiple by time.Second for the PollImmediate and then you don't have to do casting in the Infof alongside math.

pkg/network/ovn_kubernetes.go
JacobTanenbaum: done

pkg/network/ovn_kubernetes.go
JacobTanenbaum: done

PR #896, opened 2020-12-03
Bug 1903660: Revert "Read install-config and wait for the requested amount of control plane nodes"

alexanderConstantinescu:
This patch reverts commit a14003deb02d669e0f035aa3969175506a71444b from #421 

When #421 was created one year ago OVN looked very different than it does today. We imposed a restriction on bootstrapping OVN which, today, is based on legacy reasons. Essentially, we wanted to wait for all master replicas that a user had defined before bootstrapping OVN. This was done for the following reasons:

1.  We did not want to rollout the networking stack multiple times during cluster install, as to not introduce networking downtime. This would have been the case if we started deploying OVN deployment 1 when we only knew about 2 nodes, and then had to re-deploy OVN deployment 2 when we had 3 nodes. Leading to networking downtime during the rollout.
2. We were concerned that the OVN raft "join point" (i.e: node that acts as pivot point for the cluster) would change between the rollout. This could have happened if the initial cluster consisted of two nodes such as: 10.0.0.3, 10.0.0.4...only to have a third node: 10.0.0.2 join later. As we sort the master nodes in increasing order according to their IP address, the pivot would have changed from 10.0.0.3 -> 10.0.0.2, which risked causing raft issues

One year later here we are, with a bunch of improvements which I think justifies removing this check:

1. We now have OVS running as a systemd service. This means that during rollout of the networking stack, we should have 0 impact on OVS and 0 networking downtime. 
2. We now have the possibility to replace masters, thus taking into account the fact "join point" can change, in which case we blow away our DBs and re-create them.      
    
Removing this check allows for better "elasticity" of the deployment of OVN and better supporting the myriad of ways that OpenShift can install in. 

PS: looking at the restriction imposed I wonder how any OVN cluster could have supported auto-scaling/adding master nodes to the cluster? But maybe that's not a "thing"?   

/assign @squeed @trozet @danwinship

### DIFF
--- a/pkg/controller/operconfig/cluster.go
+++ b/pkg/controller/operconfig/cluster.go
@@ -2,6 +2,7 @@ package operconfig
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"reflect"
 
@@ -50,14 +51,20 @@ func (r *ReconcileOperConfig) MergeClusterConfig(ctx context.Context, operConfig
 	// If there are changes to the "downstream" networkconfig, commit it back
 	// to the apiserver
 	log.Println("WARNING: Network.operator.openshift.io has fields being overwritten by Network.config.openshift.io configuration")
-	// Have to restore the typemeta due to some weird shared cache bug --cdc
+	if updateErr := r.UpdateOperConfig(operConfig); updateErr != nil {
+		errors.Wrap(err, updateErr.Error())
+	}
+	return nil
+}
+
+func (r *ReconcileOperConfig) UpdateOperConfig(operConfig *operv1.Network) error {
 	operConfig.TypeMeta = metav1.TypeMeta{APIVersion: operv1.GroupVersion.String(), Kind: "Network"}
 	us, err := k8sutil.ToUnstructured(operConfig)
 	if err != nil {
-		return errors.Wrapf(err, "failed to transmute operator config")
+		return fmt.Errorf("failed to transmute operator config, err: %v", err)
 	}
 	if err = apply.ApplyObject(context.TODO(), r.client, us); err != nil {
-		return errors.Wrapf(err, "could not apply (%s) %s/%s", operConfig.GroupVersionKind(), operConfig.GetNamespace(), operConfig.GetName())
+		return fmt.Errorf("could not apply (%s) %s/%s, err: %v", operConfig.GroupVersionKind(), operConfig.GetNamespace(), operConfig.GetName(), err)
 	}
 	return nil
 }
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"reflect"
 	"time"
 
 	"github.com/pkg/errors"
@@ -190,13 +191,24 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	newOperConfig := operConfig.DeepCopy()
+
 	// Bootstrap any resources
-	bootstrapResult, err := network.Bootstrap(&operConfig.Spec, r.client)
+	bootstrapResult, err := network.Bootstrap(newOperConfig, r.client)
 	if err != nil {
 		log.Printf("Failed to reconcile platform networking resources: %v", err)
 		r.status.SetDegraded(statusmanager.OperatorConfig, "BootstrapError",
 			fmt.Sprintf("Internal error while reconciling platform networking resources: %v", err))
 		return reconcile.Result{}, err
+	}
+
+	if !reflect.DeepEqual(operConfig, newOperConfig) {
+		if err := r.UpdateOperConfig(newOperConfig); err != nil {
+			log.Printf("Failed to update the operator configuration: %v", err)
+			r.status.SetDegraded(statusmanager.OperatorConfig, "UpdateOperatorConfig",
+				fmt.Sprintf("Internal error while updating operator configuration: %v", err))
+			return reconcile.Result{}, err
+		}
 	}
 
 	// Generate the objects
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -10,15 +10,15 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/pkg/errors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
+	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/render"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	utilnet "k8s.io/utils/net"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const OVN_NB_PORT = "9641"
@@ -63,7 +63,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVN_CONTROLLER_INACTIVITY_PROBE"] = os.Getenv("OVN_CONTROLLER_INACTIVITY_PROBE")
 	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
 	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
-	data.Data["OVN_MASTER_IP"] = bootstrapResult.OVN.MasterIPs[0]
+	data.Data["OVN_DB_CLUSTER_INITIATOR"] = bootstrapResult.OVN.ClusterInitiator
 	data.Data["OVN_MIN_AVAILABLE"] = len(bootstrapResult.OVN.MasterIPs)/2 + 1
 	data.Data["LISTEN_DUAL_STACK"] = listenDualStack(bootstrapResult.OVN.MasterIPs[0])
 	data.Data["OVN_CERT_CN"] = OVN_CERT_CN
@@ -248,7 +248,7 @@ func networkPluginName() string {
 	return "ovn-kubernetes"
 }
 
-func bootstrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
+func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
 	masterNodeList := &corev1.NodeList{}
 	matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
 	if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
@@ -276,12 +276,42 @@ func bootstrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) 
 
 	sort.Strings(ovnMasterIPs)
 
+	// clusterInitiator is used to avoid a split-brain scenario for the OVN NB/SB DBs. We want to consistently initialize
+	// any OVN cluster which is bootstrapped here, to the same initiator (should it still exists), hence we annotate the
+	// network.operator.openshift.io CRD with this information and always try to re-use the same member for the OVN RAFT
+	// cluster initialization
+	var clusterInitiator string
+	currentAnnotation := conf.GetAnnotations()
+	if cInitiator, ok := currentAnnotation[names.OVNRaftClusterInitiator]; ok && currentInitiatorExists(ovnMasterIPs, cInitiator) {
+		clusterInitiator = cInitiator
+	} else {
+		clusterInitiator = ovnMasterIPs[0]
+		if currentAnnotation == nil {
+			currentAnnotation = map[string]string{
+				names.OVNRaftClusterInitiator: clusterInitiator,
+			}
+		} else {
+			currentAnnotation[names.OVNRaftClusterInitiator] = clusterInitiator
+		}
+		conf.SetAnnotations(currentAnnotation)
+	}
+
 	res := bootstrap.BootstrapResult{
 		OVN: bootstrap.OVNBootstrapResult{
-			MasterIPs: ovnMasterIPs,
+			MasterIPs:        ovnMasterIPs,
+			ClusterInitiator: clusterInitiator,
 		},
 	}
 	return &res, nil
+}
+
+func currentInitiatorExists(ovnMasterIPs []string, configInitiator string) bool {
+	for _, masterIP := range ovnMasterIPs {
+		if masterIP == configInitiator {
+			return true
+		}
+	}
+	return false
 }
 
 func dbList(masterIPs []string, port string) string {
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -8,30 +8,23 @@ import (
 	"path/filepath"
 	"reflect"
 	"sort"
-	"strconv"
 	"strings"
-	"time"
 
-	yaml "github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"github.com/openshift/cluster-network-operator/pkg/render"
-	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	types "k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const OVN_NB_PORT = "9641"
 const OVN_SB_PORT = "9642"
 const OVN_NB_RAFT_PORT = "9643"
 const OVN_SB_RAFT_PORT = "9644"
-const CLUSTER_CONFIG_NAME = "cluster-config-v1"
-const CLUSTER_CONFIG_NAMESPACE = "kube-system"
 const OVN_CERT_CN = "ovn"
 
 const OVN_MASTER_DISCOVERY_TIMEOUT = 280
@@ -226,7 +219,6 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 }
 
 func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int) {
-
 	if conf.DefaultNetwork.OVNKubernetesConfig == nil {
 		conf.DefaultNetwork.OVNKubernetesConfig = &operv1.OVNKubernetesConfig{}
 	}
@@ -256,48 +248,15 @@ func networkPluginName() string {
 	return "ovn-kubernetes"
 }
 
-type replicaCountDecoder struct {
-	ControlPlane struct {
-		Replicas string `json:"replicas"`
-	} `json:"controlPlane"`
-}
-
 func bootstrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
-	clusterConfig := &corev1.ConfigMap{}
-	clusterConfigLookup := types.NamespacedName{Name: CLUSTER_CONFIG_NAME, Namespace: CLUSTER_CONFIG_NAMESPACE}
 	masterNodeList := &corev1.NodeList{}
-
-	if err := kubeClient.Get(context.TODO(), clusterConfigLookup, clusterConfig); err != nil {
-		return nil, fmt.Errorf("Unable to bootstrap OVN, unable to retrieve cluster config: %s", err)
+	matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
+	if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
+		return nil, err
 	}
 
-	rcD := replicaCountDecoder{}
-	if err := yaml.Unmarshal([]byte(clusterConfig.Data["install-config"]), &rcD); err != nil {
-		return nil, fmt.Errorf("Unable to bootstrap OVN, unable to unmarshal install-config: %s", err)
-	}
-
-	controlPlaneReplicaCount, _ := strconv.Atoi(rcD.ControlPlane.Replicas)
-
-	var heartBeat int
-
-	err := wait.PollImmediate(OVN_MASTER_DISCOVERY_POLL*time.Second, OVN_MASTER_DISCOVERY_TIMEOUT*time.Second, func() (bool, error) {
-		matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
-		if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
-			return false, err
-		}
-		if len(masterNodeList.Items) != 0 && controlPlaneReplicaCount == len(masterNodeList.Items) {
-			return true, nil
-		}
-
-		heartBeat++
-		if heartBeat%3 == 0 {
-			klog.V(2).Infof("Waiting to complete OVN bootstrap: found (%d) master nodes out of (%d) expected: timing out in %d seconds",
-				len(masterNodeList.Items), controlPlaneReplicaCount, OVN_MASTER_DISCOVERY_TIMEOUT-OVN_MASTER_DISCOVERY_POLL*heartBeat)
-		}
-		return false, nil
-	})
-	if err != nil {
-		return nil, fmt.Errorf("Unable to bootstrap OVN, expected amount of control plane nodes (%v) do not match found (%v): %s", controlPlaneReplicaCount, len(masterNodeList.Items), err)
+	if len(masterNodeList.Items) == 0 {
+		return nil, fmt.Errorf("unable to bootstrap OVN, no master nodes found")
 	}
 
 	ovnMasterIPs := make([]string, len(masterNodeList.Items))
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -153,7 +153,7 @@ spec:
           # end of db_part_of_cluster
           
           # Checks if cluster has already been initialized.
-          # If not it returns false and sets init_ip to MASTER_IP
+          # If not it returns false and sets init_ip to CLUSTER_INITIATOR_IP
           cluster_exists() {
             local db=${1}
             local port=${2}
@@ -167,13 +167,13 @@ spec:
               fi
             done
             # if we get here  there is no cluster, set init_ip and get out
-            init_ip=$(bracketify $MASTER_IP)
+            init_ip=$(bracketify $CLUSTER_INITIATOR_IP)
             return 1
           }
           # end of cluster_exists()
           
-          MASTER_IP="{{.OVN_MASTER_IP}}"
-          echo "$(date -Iseconds) - starting nbdb  MASTER_IP=${MASTER_IP}, K8S_NODE_IP=${K8S_NODE_IP}"
+          CLUSTER_INITIATOR_IP="{{.OVN_DB_CLUSTER_INITIATOR}}"
+          echo "$(date -Iseconds) - starting nbdb  CLUSTER_INITIATOR_IP=${CLUSTER_INITIATOR_IP}, K8S_NODE_IP=${K8S_NODE_IP}"
           initial_raft_create=true
           initialize="false"
           
@@ -213,7 +213,7 @@ spec:
               run_nb_ovsdb
             else
               # either we need to initialize a new cluster or wait for master to create it
-              if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
+              if [[ "${K8S_NODE_IP}" == "${CLUSTER_INITIATOR_IP}" ]]; then
                 exec /usr/share/ovn/scripts/ovn-ctl \
                 --db-nb-cluster-local-port={{.OVN_NB_RAFT_PORT}} \
                 --db-nb-cluster-local-addr=$(bracketify ${K8S_NODE_IP}) \
@@ -262,8 +262,8 @@ spec:
               - -c
               - |
                 set -x
-                MASTER_IP="{{.OVN_MASTER_IP}}"
-                if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
+                CLUSTER_INITIATOR_IP="{{.OVN_DB_CLUSTER_INITIATOR}}"
+                if [[ "${K8S_NODE_IP}" == "${CLUSTER_INITIATOR_IP}" ]]; then
                   echo "$(date -Iseconds) - nbdb - postStart - waiting for master to be selected"
 
                   # set the connection and disable inactivity probe
@@ -544,7 +544,7 @@ spec:
           # end of db_part_of_cluster
           
           # Checks if cluster has already been initialized.
-          # If not it returns false and sets init_ip to MASTER_IP
+          # If not it returns false and sets init_ip to CLUSTER_INITIATOR_IP
           cluster_exists() {
             local db=${1}
             local port=${2}
@@ -558,13 +558,13 @@ spec:
               fi
             done
             # if we get here  there is no cluster, set init_ip and get out
-            init_ip=$(bracketify $MASTER_IP)
+            init_ip=$(bracketify $CLUSTER_INITIATOR_IP)
             return 1
           }
           # end of cluster_exists()
           
-          MASTER_IP="{{.OVN_MASTER_IP}}"
-          echo "$(date -Iseconds) - starting sbdb  MASTER_IP=${MASTER_IP}"
+          CLUSTER_INITIATOR_IP="{{.OVN_DB_CLUSTER_INITIATOR}}"
+          echo "$(date -Iseconds) - starting sbdb  CLUSTER_INITIATOR_IP=${CLUSTER_INITIATOR_IP}"
           initial_raft_create=true
           initialize="false"
           
@@ -604,7 +604,7 @@ spec:
               run_sb_ovsdb
             else
               # either we need to initialize a new cluster or wait for master to create it
-              if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
+              if [[ "${K8S_NODE_IP}" == "${CLUSTER_INITIATOR_IP}" ]]; then
                 exec /usr/share/ovn/scripts/ovn-ctl \
                 --db-sb-cluster-local-port={{.OVN_SB_RAFT_PORT}} \
                 --db-sb-cluster-local-addr=$(bracketify ${K8S_NODE_IP}) \
@@ -651,8 +651,8 @@ spec:
               - -c
               - |
                 set -x
-                MASTER_IP="{{.OVN_MASTER_IP}}"
-                if [[ "${K8S_NODE_IP}" == "${MASTER_IP}" ]]; then
+                CLUSTER_INITIATOR_IP="{{.OVN_DB_CLUSTER_INITIATOR}}"
+                if [[ "${K8S_NODE_IP}" == "${CLUSTER_INITIATOR_IP}" ]]; then
                   echo "$(date -Iseconds) - sdb - postStart - waiting for master to be selected"
 
                   # set the connection and disable inactivity probe
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -25,7 +25,8 @@ type KuryrBootstrapResult struct {
 }
 
 type OVNBootstrapResult struct {
-	MasterIPs []string
+	MasterIPs        []string
+	ClusterInitiator string
 }
 
 type BootstrapResult struct {
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -37,6 +37,10 @@ const NonCriticalAnnotation = "networkoperator.openshift.io/non-critical"
 // that executing network migration (switching the default network type of the cluster) is allowed.
 const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migration"
 
+// OVNRaftClusterInitiator is an annotation on the networks.operator.openshift.io CR to indicate
+// which node IP was the raft cluster initiator. The NB and SB DB will be initialized by the same member.
+const OVNRaftClusterInitiator = "networkoperator.openshift.io/ovn-cluster-initiator"
+
 // KuryrOctaviaProviderAnnotation is used to save latest Octavia provider that was configured in order to
 // prevent from reconfiguring it automatically when underlying Octavia changes.
 const KuryrOctaviaProviderAnnotation = "networkoperator.openshift.io/kuryr-octavia-provider"
--- a/pkg/network/bootstrap.go
+++ b/pkg/network/bootstrap.go
@@ -9,14 +9,14 @@ import (
 )
 
 // Bootstrap creates resources required by SDN on the cloud.
-func Bootstrap(conf *operv1.NetworkSpec, client client.Client) (*bootstrap.BootstrapResult, error) {
-	switch conf.DefaultNetwork.Type {
+func Bootstrap(conf *operv1.Network, client client.Client) (*bootstrap.BootstrapResult, error) {
+	switch conf.Spec.DefaultNetwork.Type {
 	case operv1.NetworkTypeKuryr:
-		return openstack.BootstrapKuryr(conf, client)
+		return openstack.BootstrapKuryr(&conf.Spec, client)
 	case operv1.NetworkTypeOpenShiftSDN:
 		return nil, nil
 	case operv1.NetworkTypeOVNKubernetes:
-		return bootstrapOVN(client)
+		return bootstrapOVN(conf, client)
 	}
 
 	return nil, nil


alexanderConstantinescu:
Since you're very familiar with the whole replacing master business:  

/assign @abhat

squeed:
I would be wary of accidentally setting up split brain as the "join point" changes. I think we should persist it to applied-config, and only change it if it goes away.

squeed:
I think we handle this case, but just for my own understanding: what happens when we are creating a cluster, and we have masters B and C but not A. We start to create the nbdb objects, then master A shows up. Will this be a problem? How do we handle this?

alexanderConstantinescu:
> I think we handle this case, but just for my own understanding: what happens when we are creating a cluster, and we have masters B and C but not A....How do we handle this?

So here's what I _think_ happens

**Today**

If the cluster was originally created with 2 masters and we add a third: we won't bootstrap and subsequently not reconcile. However, the new node is added and since we are running daemonsets the current OVN stack will be deployed on it, and the node A will be able to join given that he knows B and C. This is bad because any required update to OVN-K, from that point on won't have any effect. 

**With this patch**

If A has a the lowest IP among all: it will become "join point", we bootstrap, reconcile and deploy OVN-K + 1. In this case however, the "join point" is useless because in `ovnkube-master.yaml`, the following will happen:

For node A:

1. We start by checking if there's a DB file and if we should initialize based on that. There won't be since this is a new node.
2. So on node A we will check if a cluster exists, we do this by just listing all pods in `openshift-ovn-kubernetes`, this will return the old pods running on node B and C (I guess in no guaranteed order), see: https://github.com/openshift/cluster-network-operator/blob/007f1314412c2b72304b57cb1e2f220db3d60852/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L161-L168
3. Since 2. will begin with node B and `db_part_of_cluster $db_pod $db $port` will return true, the `init_ip` will be set to node B: https://github.com/openshift/cluster-network-operator/blob/007f1314412c2b72304b57cb1e2f220db3d60852/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L137
4. node A will thus join the cluster targeting node B (or node C, which ever is returned as one from 2.), I guess both are fine at that point...but I am not sure.

For node B and node C:

They will already have a DB file, so they won't join anything, they'll just start. OVN DB checker will then take care of verifying the content of the DB and check that it's valid, i.e: that the `cluster/status` is correct, that all members listed in `OVN_NB_DB_LIST` for a part of the known NB/SB DB - which should be the case with node A know, since the CNO has reconciled it as such (@abhat , correct me if I am wrong). 

> We start to create the nbdb objects, then master A shows up. Will this be a problem? 

I don't think it's a problem. Traffic for existing pods won't be interrupted since OVS is running on the system. In-flight CNI requests might however be lost in case the ovnkube-master is processing such an event and is killed while the rollout is happening because of this new master addition. But that will be the case for any situation where ovnkube-master is killed while processing a CNI request: we need to make sure we finish that transaction before dying (if we are not already doing that today?).

abhat:
When a new master is added, why do we need to change the "join point" at all? Current code doesn't worry about the master IPs to be in a sorted order at all in the daemonset. If a new node joins the raft db cluster, it just becomes a follower. The logic to figure out what join-ip to use is based on OVN_MASTER_IP being passed which for the CNO is `OVN.MasterIPs[0]`. So long as the new master being added doesn't change the spatial ordering of the IPs in the slice, we will be OK.

abhat:
LGTM pending looking at fail jobs. 

Upon deeper inspection (fetching from the ramdisk, than the cache), it looks like from the daemonset perspective, we don't care about the ordering of the master IPs so long as the cluster already has a raft leader. 

https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L535 

takes care of figuring out the current raft leader. So this should technically work even if the order of the master IPs is changed across reconciliation.

alexanderConstantinescu:
Summarizing failed test results:

**e2e-aws-ovn-windows** 
```
level=info msg=Bootstrap gather logs captured here "/tmp/installer/log-bundle-20201207104827.tar.gz"
level=fatal msg=Bootstrap failed to complete: failed to wait for bootstrapping to complete: timed out waiting for the condition
2020/12/07 10:48:50 Failed to upload $KUBECONFIG: timed out waiting for the condition: stat /tmp/secret/kubeconfig: no such file or directory
error: failed to execute wrapped command: exit status 1
```

**e2e-azure-ovn** 
```
: [sig-network] pods should successfully create sandboxes by reading container expand_more0s | : [sig-network] pods should successfully create sandboxes by reading container expand_more | 0s
-- | -- | --
: [sig-network] pods should successfully create sandboxes by reading container expand_more | 0s
: [sig-network] pods should successfully create sandboxes by getting pod expand_more0s | : [sig-network] pods should successfully create sandboxes by getting pod expand_more | 0s
: [sig-network] pods should successfully create sandboxes by getting pod expand_more | 0s
: [sig-network] pods should successfully create sandboxes by other expand_more0s | : [sig-network] pods should successfully create sandboxes by other expand_more | 0s
: [sig-network] pods should successfully create sandboxes by other expand_more | 0s
: Run multi-stage test e2e-azure-ovn - e2e-azure-ovn-openshift-e2e-test container test | : Run multi-stage test e2e-azure-ovn - e2e-azure-ovn-openshift-e2e-test container test
: Run multi-stage test e2e-azure-ovn - e2e-azure-ovn-openshift-e2e-test container test
```

**e2e-gcp-ovn-upgrade**
```
level=error
level=error msg=Error: Request "Create IAM Members roles/compute.instanceAdmin serviceAccount:ci-op-rbzyvplj-8aaeb-fv48s-m@openshift-gce-devel-ci.iam.gserviceaccount.com for \"project \\\"openshift-gce-devel-ci\\\"\"" returned error: Batch request and retried single request "Create IAM Members roles/compute.instanceAdmin serviceAccount:ci-op-rbzyvplj-8aaeb-fv48s-m@openshift-gce-devel-ci.iam.gserviceaccount.com for \"project \\\"openshift-gce-devel-ci\\\"\"" both failed. Final error: Error applying IAM policy for project "openshift-gce-devel-ci": Error setting IAM policy for project "openshift-gce-devel-ci": googleapi: Error 400: Service account ci-op-r7kn26-openshift-g-zq4rk@openshift-gce-devel-ci.iam.gserviceaccount.com does not exist., badRequest
level=error
level=error msg=  on ../tmp/openshift-install-119494987/master/main.tf line 6, in resource "google_project_iam_member" "master-compute-admin":
level=error msg=   6: resource "google_project_iam_member" "master-compute-admin" {
level=error
level=error
```

**e2e-openstack-ovn**
```
level=info msg=Cluster operator machine-config Progressing is True with : Working towards 4.7.0-0.ci.test-2020-12-07-100455-ci-op-liibcpwk
level=error msg=Cluster operator machine-config Degraded is True with RequiredPoolsFailed: Unable to apply 4.7.0-0.ci.test-2020-12-07-100455-ci-op-liibcpwk: timed out waiting for the condition during syncRequiredMachineConfigPools: error pool worker is not ready, retrying. Status: (pool degraded: true total: 3, ready 0, updated: 0, unavailable: 3)
level=info msg=Cluster operator machine-config Available is False with : Cluster not available for 4.7.0-0.ci.test-2020-12-07-100455-ci-op-liibcpwk
level=info msg=Cluster operator network ManagementStateDegraded is False with : 
level=error msg=Cluster initialization failed because one or more operators are not functioning properly.
level=error msg=The cluster should be accessible for troubleshooting as detailed in the documentation linked below,
level=error msg=https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html
level=error msg=The 'wait-for install-complete' subcommand can then be used to continue the installation
level=fatal msg=failed to initialize the cluster: Cluster operator machine-config is still updating
```

**e2e-ovn-hybrid-step-registry - maybe something linked to this PR**
```
level=info msg=Cluster operator network ManagementStateDegraded is False with : 
level=error msg=Cluster operator network Degraded is True with RolloutHung: DaemonSet "openshift-ovn-kubernetes/ovnkube-node" rollout is not making progress - last change 2020-12-07T10:20:59Z
level=error msg=DaemonSet "openshift-network-diagnostics/network-check-target" rollout is not making progress - last change 2020-12-07T10:18:43Z
level=error msg=Deployment "openshift-network-diagnostics/network-check-source" rollout is not making progress - last change 2020-12-07T10:27:29Z
```

**e2e-ovn-step-registry - maybe something linked to this PR**
```
es/ovnkube-master" rollout is not making progress - pod ovnkube-master-2824b is in CrashLoopBackOff State
level=error msg=DaemonSet "openshift-ovn-kubernetes/ovnkube-master" rollout is not making progress - last change 2020-12-07T10:23:26Z
level=error msg=DaemonSet "openshift-ovn-kubernetes/ovnkube-node" rollout is not making progress - last change 2020-12-07T10:22:48Z
level=error msg=DaemonSet "openshift-network-diagnostics/network-check-target" rollout is not making progress - last change 2020-12-07T10:22:35Z
level=error msg=Deployment "openshift-network-diagnostics/network-check-source" rollout is not making progress - last change 2020-12-07T10:31:11Z

File: pkg/network/ovn_kubernetes.go [hunk 1]
```diff
@@ -256,48 +248,15 @@ func networkPluginName() string {
 	return "ovn-kubernetes"
 }
 
-type replicaCountDecoder struct {
-	ControlPlane struct {
-		Replicas string `json:"replicas"`
-	} `json:"controlPlane"`
-}
-
 func bootstrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
```
danwinship, line 251: If we're not doing anything special at install time any more, we shouldn't need any "bootstrap" code at all; we should just be able to do it all from `renderOVNKubernetes` without it needing a `bootstrapResult`

alexanderConstantinescu: Not quite. We still need to bootstrap and provide OVN (nbdb/sbdb/northd/ovnkube) with the entire list of master IPs when deploying the stack. Once upon a time I tried to remove that (#447) and failed miserably

danwinship: But how can you provide the entire list of master IPs if you don't wait for all the masters to be created?

danwinship: If the dbs, etc, actually need the _complete_ list of masters when they are first configured, then this PR is wrong. If they just need a list of _some_ masters and we can add to that list later and everything will keep working, then you don't need bootstrapping code, you just need `renderOVNKubernetes` to re-fetch the list of master IPs when the set of masters changes.

alexanderConstantinescu: >  If they just need a list of some masters and we can add to that list later and everything will keep working

That's it. 

> then you don't need bootstrapping code, you just need renderOVNKubernetes to re-fetch the list of master IPs when the set of masters changes.

Does it matter though? `renderOVNKubernetes` would need to do exactly what `bootstrapOVN` does today, i.e: get all master nodes's internal IP and render the template with it. Or am I overlooking something that would simply things further?

alexanderConstantinescu:
FYI: @rbbratta has tested that we still function when replacing the "NB DB join point", I think we can thus conclude that this patch should not harm us. We can replace the join point, as much as we can replace any other raft member in a OVN cluster.

pkg/network/ovn_kubernetes.go
danwinship: > Does it matter though?

Architecturally, there is supposed to be a distinction between bootstrap and non-bootstrap, though I don't know exactly what it is since I guess we re-run the bootstrap code even after install time? @squeed should know...

alexanderConstantinescu:
Update 2: it seems the PR https://github.com/openshift/cluster-network-operator/pull/909 helped solved the assisted installer teams issues and successfully deployed a 4.6 cluster with this patch, so that's another data point

pkg/network/ovn_kubernetes.go
squeed: Bootstrap was initially designed for Kuryr, which needs to create certain cloud-provider resources on install. Given that we also wanted to support the offline renderer, it seemed like a useful design.

It turns out we didn't need to render offline, so the distinction between "bootstrap" and "render" phases is no longer necessary. However, it does make unit testing a **lot easier**, so I'd like to keep it if possible.

I have a suggested design, which I'll add in a comment below, since it isn't really a line-level suggestion.

squeed:
Again, the correct solution to prevent unnecessary re-renders is is to have `bootstrapOVNKubernetes()` decide on the join IP but persist it in an annotation on the `network.operator.openshift.io` object. The basic steps are

1. Get the list of master IPs
2. Check the annotation for an existing join IP
3. If the existing join IP is in the list of master IPs, you're done
4. If the annotation is missing, or points to a node that no longer exists, select a join IP
5. Store that IP in the annotation.

We really should not risk, ever, OVN db split brain.

alexanderConstantinescu:
> 5\. Store that IP in the annotation.

Our code today is not super well structured around performing manipulations such as that, which in turn makes that patch a bit contrived. 

If we need to check an annotation on `network.operator.openshift.io` in the bootstrap part and then persist it back in case it matches what we expect or not, then that requires some not-so-clean code. I updated my patch with what that would look like so that you can see. Let me know if you feel comfortable keeping it or you prefer to trust the master-replace code to not have us end up in a split brain situation.  

Mind you, I would like to avoid re-writing too much of the CNO in this patch, so I applied it as to keep things in the order they are today.

abhat:
> > 5. Store that IP in the annotation.
> 
> Our code today is not super well structured around performing manipulations such as that, which in turn makes that patch a bit contrived.
> 
> If we need to check an annotation on `network.operator.openshift.io` in the bootstrap part and then persist it back in case it matches what we expect or not, then that requires some not-so-clean code. I updated my patch with what that would look like so that you can see. Let me know if you feel comfortable keeping it or you prefer to trust the master-replace code to not have us end up in a split brain situation.
> 
> Mind you, I would like to avoid re-writing too much of the CNO in this patch, so I applied it as to keep things in the order they are 
today.

Without getting into the weeds too much, what cases do we expect "re-rendering" to happen? Once the master IP is chosen by the daemon-set bring-up code, it doesn't change as far as OVN is concerned so long as that master IP is still a valid IP assigned to one of the masters.

I prefer that we keep the original patch as is and worry about "tracking" the master IP as an annotation or whatever other mechanism we choose as a separate PR. We need to unblock other teams who are waiting for this patch to land.

alexanderConstantinescu:
> Without getting into the weeds too much, what cases do we expect "re-rendering" to happen? Once the master IP is chosen by the daemon-set bring-up code, it doesn't change as far as OVN is concerned so long as that master IP is still a valid IP assigned to one of the masters.

It will happen whenever any template variable changes, so in this case: when `OVN_NB_DB_LIST` or `OVN_SB_DB_LIST` change (whenever a new master node is added) or `OVN_MASTER_IP` whenever the list of master nodes remains unchanged, but we've replaced the master with the lowest IP to a new one (well, this is no longer a problem given the second commit I've added).

> I prefer that we keep the original patch as is and worry about "tracking" the master IP as an annotation or whatever other mechanism we choose as a separate PR. We need to unblock other teams who are waiting for this patch to land.

Well, now that I've added the new commit I'll let you and @squeed work out on what should go in.

pkg/controller/operconfig/cluster.go
```diff
@@ -50,14 +51,20 @@ func (r *ReconcileOperConfig) MergeClusterConfig(ctx context.Context, operConfig
...
+		errors.Wrap(err, updateErr.Error())
+	}
+	return nil
+}
+
+func (r *ReconcileOperConfig) UpdateOperConfig(operConfig *operv1.Network) error {
 	operConfig.TypeMeta = metav1.TypeMeta{APIVersion: operv1.GroupVersion.String(), Kind: "Network"}
 	us, err := k8sutil.ToUnstructured(operConfig)
 	if err != nil {
-		return errors.Wrapf(err, "failed to transmute operator config")
```
squeed, line 57: minor nit: I doubt we want to swallow `err` here - wrapping it is useful for debugging. You can also do `fmt.Errorf("%w"` to wrap the error.

pkg/network/ovn_kubernetes.go
```diff
@@ -276,14 +276,40 @@ func bootstrapOVN(kubeClient client.Client) (*bootstrap.BootstrapResult, error)
 
 	sort.Strings(ovnMasterIPs)
 
+	var clusterInitiator string
```
squeed, line 279: A few sentences about what this does would be good. Something about picking a stable raft "join point"

pkg/controller/operconfig/cluster.go
alexanderConstantinescu: I am not swallowing the error. `UpdateOperConfig` returns its own error and is then `errors.Wrapf`-ed with any eventual previous error, on L55

pkg/network/ovn_kubernetes.go
alexanderConstantinescu: The const `OVNRaftClusterInitiator` has a doc bloc which explains this, but I could add a minor here as well

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -190,15 +191,26 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	oldOperConfig := operConfig.DeepCopy()
```
squeed, line 194: This is backwards; you need to mutate the `DeepCopy()` and leave the original one untouched - or else you touch the shared object in the cache.

pkg/network/ovn_kubernetes.go
```diff
@@ -70,7 +63,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVN_CONTROLLER_INACTIVITY_PROBE"] = os.Getenv("OVN_CONTROLLER_INACTIVITY_PROBE")
 	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
 	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
-	data.Data["OVN_MASTER_IP"] = bootstrapResult.OVN.MasterIPs[0]
+	data.Data["OVN_MASTER_IP"] = bootstrapResult.OVN.ClusterInitiator
```
squeed, line 66: While we're at it, let's change the name of this key.

pkg/network/ovn_kubernetes.go
squeed: thanks!

pkg/controller/operconfig/cluster.go
squeed: Right, but here in line 64 you're returning a new error without the value of `err` from line 62. (And the same on line 66/67)

squeed:
(sorry, a few comments got eaten when my cable modem took a nap).

This is definitely the safest approach. It's really, really important that we never allow for the possibility of any sort of raft split-brain. Regardless of who is waiting on this patch. And a critical part of raft stability is a stable join point during cluster initiation.

pkg/controller/operconfig/cluster.go
alexanderConstantinescu: Ahh, understood now! Thanks, I will fix it

alexanderConstantinescu:
> (sorry, a few comments got eaten when my cable modem took a nap).
> 
> This is definitely the safest approach. It's really, really important that we never allow for the possibility of any sort of raft split-brain. Regardless of who is waiting on this patch. And a critical part of raft stability is a stable join point during cluster initiation.

I've update the PR taking your comments into account. PTAL

squeed:
/approve

would like to watch the tests succeed (at least, spin up an ovn cluster) before lgtm.

alexanderConstantinescu:
> /approve
> 
> would like to watch the tests succeed (at least, spin up an ovn cluster) before lgtm.

So, most OVN jobs passed. These are the required jobs that didn't pass:

* `e2e-gcp` - flake, 5 / ~ 3000 tests passed. And it should not be impacted by this patch since it runs openshift-sdn.
* `e2e-aws-sdn-multi` same as above
* `e2e-agnostic-upgrade` also flaky, and in general bad: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-agnostic-upgrade

squeed:
ci/prow/e2e-azure-ovn has ovn crashlooping.

alexanderConstantinescu:
> ci/prow/e2e-azure-ovn has ovn crashlooping.

I've had a look at this but I am not able to fully determine what's going on. One of the masters doesn't seem to be able to start the NB DB, but nothing is logged at all indicating why: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/896/pull-ci-openshift-cluster-network-operator-master-e2e-azure-ovn/1346072233218412544/artifacts/e2e-azure-ovn/gather-extra/pods/openshift-ovn-kubernetes_ovnkube-master-zbgmp_nbdb.log

I re-launched the job and the cluster started up fine that time around, which worries me that some form of nasty race conditions is lurking in the background. Obviously, since networking is completely hosed, the must-gather pods never come up and I am unable to retrieve the OVN DBs to inspect further...I will re-launch the job a couple of times more and hope it fails again.

alexanderConstantinescu:
OK, I _think_ I've gotten to the bottom of the problem. I am emphasizing _think_ because I am not able to fully prove it. 

TL;DR: It seems we're missing https://github.com/ovn-org/ovn-kubernetes/pull/1930 for this to work. 

Essentially what is happening is we begin by rolling out OVN v1 which only contains a subset of all master nodes that will exists. OVN v1 starts deploying the SB and NB DBs. CNO reconciles, bootstraps, sees all master nodes and re-deploys OVN v2. Now, there can be cases where one/some of the pods in OVN v1 can be disjoint from the raft cluster for whatever reason (this is seen on the other master pods since they do not have it/them listed as cluster members) - maybe we deployed OVN v2 and stopped what they were doing at a critical moment.  

The pods that are disjoint from the raft cluster will start Crashlooping when OVN v2 runs, this is because the pods running in v2 will never call `usr/bin/ovn-appctl -t /var/run/ovn/ovnsb_db.ctl exit` (which is done from the post stop hook, which is not called for Crashlooping pods) and close the unix sockets that were opened in v1. v2 will thus think that the NB / SB DB is running already and never starts, this can be seen in the following logs where I've modified `ovn-ctl` with `set -x`:

```
/usr/share/ovn/scripts/ovn-ctl --db-sb-cluster-local-port=9644 --db-sb-cluster-remote-port=9644 --db-sb-cluster-local-addr=10.0.178.155 --db-sb-cluster-remote-addr=10.0.169.225 --no-monitor --db-sb-cluster-local-proto=ssl --db-sb-cluster-remote-proto=ssl --ovn-sb-db-ssl-key=/ovn-cert/tls.key --ovn-sb-db-ssl-cert=/ovn-cert/tls.crt --ovn-sb-db-ssl-ca-cert=/ovn-ca/ca-bundle.crt '--ovn-sb-log=-vconsole:info -vfile:off' run_sb_ovsdb
+ case $0 in
++ sed 's,/[^/]*$,,'
++ echo /usr/share/ovn/scripts/ovn-ctl
....
+ test '!' -d /etc/ovn
+ pidfile_is_running /var/run/ovn/ovnsb_db.pid
+ return
```

This is the reason the SB and NB containers never do anything as mentioned in https://github.com/openshift/cluster-network-operator/pull/896#issuecomment-754141340 but just crash instead, to them something is already running. 

What I am not able to figure out is why v1 never properly calls the post stop hook and performs the proper exit command, when the pods are supposed to be stopped.

Anyways, the patch https://github.com/ovn-org/ovn-kubernetes/pull/1930 will fix this. I am certain of that because the `ovn-dbchecker` container in the disjoint pods are logging:

```
oc logs -c ovn-dbchecker ovnkube-master-ckxs4 -n openshift-ovn-kubernetes
+ [[ -f /env/_master ]]
++ date '+%m%d %H:%M:%S.%N'
+ echo 'I0105 14:25:27.828910325 - ovn-dbchecker - start ovn-dbchecker'
I0105 14:25:27.828910325 - ovn-dbchecker - start ovn-dbchecker
+ exec /usr/bin/ovndbchecker --config-file=/run/ovnkube-config/ovnkube.conf --loglevel 4 --sb-address ssl:10.0.162.35:9642,ssl:10.0.171.60:9642,ssl:10.0.224.250:9642 --sb-client-privkey /ovn-cert/tls.key --sb-client-cert /ovn-cert/tls.crt --sb-client-cacert /ovn-ca/ca-bundle.crt --sb-cert-common-name ovn --nb-address ssl:10.0.162.35:9641,ssl:10.0.171.60:9641,ssl:10.0.224.250:9641 --nb-client-privkey /ovn-cert/tls.key --nb-client-cert /ovn-cert/tls.crt --nb-client-cacert /ovn-ca/ca-bundle.crt --nb-cert-common-name ovn
I0105 14:25:27.839426       1 config.go:1306] Parsed config file /run/ovnkube-config/ovnkube.conf
I0105 14:25:27.839468       1 config.go:1307] Parsed config: {Default:{MTU:8901 ConntrackZone:64000 EncapType:geneve EncapIP: EncapPort:6081 InactivityProbe:100000 OpenFlowProbe:180 RawClusterSubnets:10.128.0.0/14/23 ClusterSubnets:[]} Logging:{File: CNIFile: Level:4 LogFileMaxSize:100 LogFileMaxBackups:5 LogFileMaxAge:5} CNI:{ConfDir:/etc/cni/net.d Plugin:ovn-k8s-cni-overlay} OVNKubernetesFeature:{EnableEgressIP:true} Kubernetes:{Kubeconfig: CACert: APIServer:https://api-int.ci-op-qmv7s00p-5e450.origin-ci-int-aws.dev.rhcloud.com:6443 Token: CompatServiceCIDR: RawServiceCIDRs:172.30.0.0/16 ServiceCIDRs:[] OVNConfigNamespace:openshift-ovn-kubernetes MetricsBindAddress: OVNMetricsBindAddress: MetricsEnablePprof:false OVNEmptyLbEvents:false PodIP: RawNoHostSubnetNodes:kubernetes.io/os=windows NoHostSubnetNodes:nil} OvnNorth:{Address: PrivKey: Cert: CACert: CertCommonName: Scheme: northbound:false externalID: exec:<nil>} OvnSouth:{Address: PrivKey: Cert: CACert: CertCommonName: Scheme: northbound:false externalID: exec:<nil>} Gateway:{Mode:local Interface: NextHop: VLANID:0 NodeportEnable:true DisableSNATMultipleGWs:false V4JoinSubnet:100.64.0.0/16 V6JoinSubnet:fd98::/64} MasterHA:{ElectionLeaseDuration:60 ElectionRenewDeadline:30 ElectionRetryPeriod:20} HybridOverlay:{Enabled:true RawClusterSubnets:10.132.0.0/14 ClusterSubnets:[] VXLANPort:4789}}
I0105 14:25:27.843016       1 ovndbmanager.go:23] Starting DB Checker to ensure cluster membership and DB consistency
I0105 14:25:27.843060       1 ovndbmanager.go:44] Starting ensure routine for Raft db: /etc/ovn/ovnsb_db.db
I0105 14:25:27.843136       1 ovndbmanager.go:44] Starting ensure routine for Raft db: /etc/ovn/ovnnb_db.db
W0105 14:28:27.850413       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:28:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:28:27.852990       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:28:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:30:27.850314       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:30:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:30:27.852772       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:30:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:32:27.850286       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:32:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:32:27.852570       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:32:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:33:27.850534       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:33:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:33:27.852832       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:33:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:35:27.850761       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:35:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:35:27.853058       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:35:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:36:27.850541       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:36:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:36:27.852915       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:36:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:37:27.850595       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:37:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:37:27.852914       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:37:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:37:27.856333       1 ovndbmanager.go:168] Unable to get db pod list from kubeclient: Get "https://api-int.ci-op-qmv7s00p-5e450.origin-ci-int-aws.dev.rhcloud.com:6443/api/v1/namespaces/openshift-ovn-kubernetes/pods?labelSelector=ovn-db-pod%3Dtrue": dial tcp 10.0.146.117:6443: connect: connection refused
W0105 14:39:27.850785       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:39:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:39:27.853260       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:39:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:40:27.850252       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:40:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:40:27.853276       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:40:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:41:27.850897       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:41:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:41:27.853134       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:41:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:42:27.850345       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:42:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:42:27.852582       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:42:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:43:27.850602       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:43:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:43:27.853163       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:43:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:46:27.850520       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:46:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:46:27.852766       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:46:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:47:27.850426       1 ovndbmanager.go:100] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:47:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
W0105 14:47:27.852798       1 ovndbmanager.go:156] Unable to get cluster status for: /etc/ovn/ovnnb_db.db, stderr: 2021-01-05T14:47:27Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovn-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (No such file or directory)
, err: OVN command '/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound' failed: exit status 1
```

They can't do that because the socket returns a connection refused. This is something I verified from one of the disjoint pods.
 
```
ovs-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound
2021-01-05T13:02:08Z|00001|unixctl|WARN|failed to connect to /var/run/ovn/ovnnb_db.ctl
ovs-appctl: cannot connect to "/var/run/ovn/ovnnb_db.ctl" (Connection refused)
[root@ip-10-0-178-155 ~]# ls -l /var/run/ovn/ovnnb_db.ctl
srwxr-x---. 1 root root 0 Jan  5 12:53 /var/run/ovn/ovnnb_db.ctl
```

And given the implementation of that patch, it will call exit for us in this case and have them exit their bad state.

alexanderConstantinescu:
Another approach we could take is keeping the existing code and combining it with d9f8f8c5b13216e194bc94a0d2e59a579b3eb271, i.e: wait a bit for all master nodes to be up in the normal case, but not error out in case the replica count never meets the expected one and annotate `networks.operator.openshift.io` with the join point. It kind of brings the worst out of both worlds: still keeping the dependency on the `cluster-config-v1` config map and introducing non-bootstrapping code in the CNO boostrapping code...but that will certainly reduce the entropy which is killing us in this case, and unblock the assisted installer team.

alexanderConstantinescu:
> Another approach we could take is keeping the existing code and combining it with [d9f8f8c](https://github.com/openshift/cluster-network-operator/commit/d9f8f8c5b13216e194bc94a0d2e59a579b3eb271), i.e: wait a bit for all master nodes to be up in the normal case, but not error out in case the replica count never meets the expected one and annotate `networks.operator.openshift.io` with the join point. It kind of brings the worst out of both worlds: still keeping the dependency on the `cluster-config-v1` config map and introducing non-bootstrapping code in the CNO boostrapping code...but that will certainly reduce the entropy which is killing us in this case, and unblock the assisted installer team.

I opened PR: https://github.com/openshift/cluster-network-operator/pull/933 for that.

abhat:
> (sorry, a few comments got eaten when my cable modem took a nap).
> 
> This is definitely the safest approach. It's really, really important that we never allow for the possibility of any sort of raft split-brain. Regardless of who is waiting on this patch. And a critical part of raft stability is a stable join point during cluster initiation.

If we want to add this code as a fail-safe that is OK by me. But given how the daemonset code handles master IP changing (for master replacement or other use-cases), it's really not possible to have a split brain. Of course, there may be bugs lurking there like we found with ovn-org/ovn-kubernetes/pull/1930. But I expect those occurrences to be rare to none.

alexanderConstantinescu:
> I opened PR: #934 for that.

Another data point is that this PR worked for the assisted installer team. I will thus wait for ovn-org/ovn-kubernetes#1930 and its subsequent downstream PR to integrate, then test this PR again. I will also open a PR of #934 against master as a back-up, in case everything goes too slow for the prior solution (this needs to merge for the assisted installer team before code freeze).

alexanderConstantinescu:
The dependent PR in downstream ovn-kube has merged

/retest

alexanderConstantinescu:
The following jobs failed because of the same reasons mentioned in this PR:
```
e2e-ovn-step-registry 
e2e-ovn-ipsec-step-registry
e2e-ovn-hybrid-step-registry
```
thus, PR https://github.com/openshift/ovn-kubernetes/pull/406 doesn't fix this problem. The ovndb-checker does not manage to flush the DB at all. I only have logs from `e2e-ovn-step-registry ` to be able to verify what happened. This is what occurred:

The cluster started off with master nodes `10.0.183.89` and `10.0.214.15`. `10.0.148.219` joined in v2. We can see that the db-checker on `10.0.148.219` does not flush anything because it never experiences any problems connecting to the SB or NB DBs, see: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/896/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-step-registry/1350000456255934464/artifacts/e2e-ovn-step-registry/gather-extra/pods/openshift-ovn-kubernetes_ovnkube-master-m7j8k_ovn-dbchecker.log

The nodes `10.0.183.89` and `10.0.214.15` do experience problems connecting to the SB DB (which is the problem with this job): https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/896/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-step-registry/1350000456255934464/artifacts/e2e-ovn-step-registry/gather-extra/pods/openshift-ovn-kubernetes_ovnkube-master-lpsqk_ovn-dbchecker.log and https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/896/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-step-registry/1350000456255934464/artifacts/e2e-ovn-step-registry/gather-extra/pods/openshift-ovn-kubernetes_ovnkube-master-2z4mc_ovn-dbchecker.log but it's sporadic and the counter is being reset constantly or not incremented sufficiently consistently and thus the DB is never flushed for a proper restart during the run of the CI job (approx 1 hour).

I am thus closing this PR, since it means that it will probably happen in the future and the fact that we're rolling out v2 so quickly after v1 is breaking us. 

We have waited long enough, I will file a master PR of #934  and close this as it clearly does not work.

/close

abhat:
@alexanderConstantinescu the flush logic is not working because the ovndbmanager is not able to get a list of current pods from API server. Not because it doesn't see a problem connecting to ovn dbs. We need to understand why api server is not reachable at that point. If the kclient to be used should be an internal one, maybe that is the bug.

alexanderConstantinescu:
> Not because it doesn't see a problem connecting to ovn dbs. We need to understand why api server is not reachable at that point. If the kclient to be used should be an internal one, maybe that is the bug.

@abhat : I never said that. I said: "but it's sporadic and the counter is being reset constantly or not incremented sufficiently consistently and thus the DB is never flushed for a proper restart during the run of the CI job (approx 1 hour)." 

We can see that in the logs, here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/896/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-step-registry/1350000456255934464/artifacts/e2e-ovn-step-registry/gather-extra/pods/openshift-ovn-kubernetes_ovnkube-master-lpsqk_ovn-dbchecker.log

It's logging

```
Failed to get cluster status for: /etc/ovn/ovnsb_db.db, number of retries: 5
```

But never reaches resetting the DB part. 

And on node `10.0.148.219` it does log not being able to communicate with the API server, but that is only for a short while, and the problem goes away.

PR #900, opened 2020-12-04
Fix IPsec CI

markdgray:
IPsec CI is unable to start pods due to lack of openshift-ovn-kubernetes namespace.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -114,14 +114,6 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 
 	if c.IPsecConfig != nil {
 		data.Data["EnableIPsec"] = true
-		// Only render ipsec manifest if ipsec has been enabled at cluster
-		// installation time. We will never have to delete the ipsec pod
-		// because it cannot be disabled at runtime
-		ipsecManifests, err := render.RenderDir(filepath.Join(manifestDir, "network/ovn-kubernetes-ipsec"), &data)
-		if err != nil {
-			return nil, errors.Wrap(err, "failed to render ipsec manifest")
-		}
-		objs = append(objs, ipsecManifests...)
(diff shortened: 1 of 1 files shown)


dcbw:
/lgtm
/approve

@markdgray a different way to fix this would be to put the ipsec yaml in the same dir as the rest of the ovnkube stuff, and then the filenames will handle the ordering for you.

pkg/network/ovn_kubernetes.go
```diff
@@ -114,14 +114,6 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 
 	if c.IPsecConfig != nil {
```
abhat, line 115: data.Data["EnableIPsec"] = c.IPSecConfig != nil

abhat: this is a nit. you don't have to fix it if all the other jobs are passing right now. it's more important than fixing the nit and having jobs fail again because - CI gods

markdgray:
> /lgtm
> /approve
> 
> @markdgray a different way to fix this would be to put the ipsec yaml in the same dir as the rest of the ovnkube stuff, and then the filenames will handle the ordering for you.

The problem with that was that the renderdir function will render everything in the folder. Is there a way to template out an entire file depending on the configuration option? IIRC I try to add an if clause around the whole yaml but that did not work so in the end I thought this was the cleanest way to do it.

pkg/network/ovn_kubernetes.go
markdgray: I think it should pass so I'll continue playing the retest lottery

dcbw:
> The problem with that was that the renderdir function will render everything in the folder. Is there a way to template out an entire file depending on the configuration option? IIRC I tried to add an if clause around the whole yaml but that did not work so in the end I thought this was the cleanest way to do it.

@markdgray good point, you're probably right. This is the best way to do it for now.

PR #904, opened 2020-12-07
DNM: allow overriding openshift-sdn binary

squeed:
This is a hack for testing.


danwinship:
(repushed to be based directly off the 4.3.15 CNO, so that deploying it to a 4.3.15 cluster will not cause OVS pods to be restarted due to startup script changes)

danwinship:
/close
([SDN-1411](https://issues.redhat.com/browse/SDN-1411) has thoughts on "productizing" this)

PR #905, opened 2020-12-07
Bug 1906844: Disable EndpointSlice and EndpointSliceProxying feature gates and update to k8s 1.19.4.

abhat:
This PR adds two commits:
- Bump k8s version in CNO to 1.19.4
- Disable the kubeproxy configuration for EndpointSlice and EndpointSliceProxying feature gates. 

This obviates the need for: https://github.com/openshift/sdn/pull/227/commits/12b21f8339b04ebe429cdab71674cad13b70fdef


go.mod
```diff
@@ -26,19 +26,48 @@ require (
...
-	k8s.io/client-go v0.19.2
-	k8s.io/code-generator v0.19.2
-	k8s.io/component-base v0.19.2
+	k8s.io/client-go v0.19.4
+	k8s.io/code-generator v0.19.4
+	k8s.io/component-base v0.19.4
 	k8s.io/klog/v2 v2.4.0
-	k8s.io/kube-proxy v0.19.2
+	k8s.io/kube-proxy v0.19.4
+	k8s.io/kubernetes v1.19.4
```
danwinship, line 41: Since this is for master only, we really ought to go to `1.20.0-rc.0` if you can...

go.mod
```diff
@@ -26,19 +26,48 @@ require (
...
 	k8s.io/utils v0.0.0-20200729134348-d5654de09c73
 	sigs.k8s.io/controller-runtime v0.6.3
 	sigs.k8s.io/structured-merge-diff/v4 v4.0.2 // indirect
 )
+
+replace (
+	github.com/onsi/ginkgo => github.com/onsi/ginkgo v1.11.0
```
danwinship, line 48: this shouldn't need to be done as a `replace`; just specify the correct version above

pkg/util/k8s/kubeproxy.go
```diff
@@ -81,6 +82,11 @@ func GenerateKubeProxyConfiguration(args map[string]operv1.ProxyArgumentList) (s
 
 	kpc.NodePortAddresses = ka.getCIDRList("node-port-addresses")
 
+	if kpc.FeatureGates == nil {
+		kpc.FeatureGates = make(map[string]bool)
+	}
+	kpc.FeatureGates[string(features.EndpointSlice)] = false
+	kpc.FeatureGates[string(features.EndpointSliceProxying)] = false
```
danwinship, line 89: We don't want to do this here; in particular, we eventually want to support running standalone kube-proxy with `IPv6DualStack=true` (and thus `EndpointSlice=true`). (SDN-1325)

You want to add these to `kpcOverrides` in `pkg/network/openshift_sdn.go`, though that will also require changing `MergeKubeProxyArguments`, because we want to merge the individual `FeatureGates` keys from each config source (defaults, user, overrides). And then we need to set `kpc.FeatureGates` correctly from `ka` here. This will need unit tests...

go.mod
abhat: `1.20.0-rc.0` hasn't landed in `openshift/kubernetes` yet afaik. I can create a card to track updating CNO deps to 1.20 when it lands?

go.mod
abhat: the `apimachinery` [go.mod](https://github.com/openshift/kubernetes/blob/master/staging/src/k8s.io/apimachinery/go.mod#L24) needs to have that crazy incompatible ginkgo version. So this is just to get to making it work.

pkg/util/k8s/kubeproxy.go
abhat: Ack. I was a bit torn between doing it as per your proposal or the way I ended up implementing it generically for all kube-proxy configs. I was not aware of SDN-1235. I will change this per your proposal.

go.mod
abhat: It looks like if I try to bump to 1.20.0-rc.0, we hit errors with `openshift/api` repo not being bumped to 1.20 yet.

```
go: finding module for package k8s.io/api/settings/v1alpha1
github.com/openshift/cluster-network-operator/pkg/controller/pki imports
        github.com/openshift/library-go/pkg/operator/certrotation imports
        github.com/openshift/library-go/pkg/operator/resource/resourceapply imports
        github.com/openshift/api imports
        k8s.io/api/settings/v1alpha1: module k8s.io/api@latest found (v0.19.4, replaced by k8s.io/api@v0.20.0-rc.0), but does not contain package k8s.io/api/settings/v1alpha1
```
I am going to create a card for myself to bump CNO go mods to 1.20.0 once the api changes land.

abhat: Will track this as SDN-1405

pkg/util/k8s/kubeproxy.go
```diff
@@ -255,3 +256,30 @@ func (ka *kpcArgs) getPortRange(key string) string {
...
+
+// getFeatureGates parses feature-gates and returns a map[string]bool
+func (ka *kpcArgs) getFeatureGates(key string) map[string]bool {
+	val := ka.args[key]
+	if len(val) == 0 {
+		return nil
+	}
+	delete(ka.args, key)
+	featureGates := make(map[string]bool)
+	for _, v := range val {
```
danwinship, line 268: you could perhaps do

    featureGates := cliflag.NewMapStringBool(nil)
    for _, v := range val {
            err := featureGates.Set(v)
            if err != nil { ... }
    }
    return featureGates.Map

where `cliflag` is `"k8s.io/component-base/cli/flag"`

abhat: Thanks. So much more cleaner.

danwinship:
@abhat looks like they turned on "you need a bz" checking

/override ci/prow/e2e-openstack-ovn

abhat:
This PR introduces a change to openshift-sdn kube-proxy which ovn-kubernetes doesn't use. the failing job e2e-aws-ovn-windows seems to be hitting a prow issue: `Failed resolve MultiStageTestConfiguration: no workflow named openshift-upgrade-gcp`

Overriding:

PR #910, opened 2020-12-10
Bug 1901034: proxyconfig: get basedomain from dns config

staebler:
The noproxy field includes the names of the etcd servers. The domain name used is obtained from the etcdDiscoveryDomain field of the infrastructure resource. However, that field is deprecated and no longer populated. Instead, get the domain name from the baseDomain field of the dns resource.


sdodson:
@trozet @dcbw PTAL

staebler:
It looks like #930 will make this PR unnecessary.

staebler:
This is obsolete.
/close

PR #912, opened 2020-12-11
Bug 1894268: Join subnets

abhat:
Add the ability to configure v4 and v6 join subnets via operator manifests

Depends on: openshift/api/pull/824


abhat:
/hold
Need to drop https://github.com/openshift/cluster-network-operator/pull/912/commits/f0726ac29c19b315c92b82cc1571075614e7c6a9 and instead bump openshift/api once openshift/api#824 merges.

danwinship:
copying my comment from the api PR:

> This is really an ugly implementation detail that no user should ever have to think about. It would be better if we could find some way to not need a separate CIDR for this. Like, maybe ovn-kubernetes could just use the highest-numbered subnet from the cluster network CIDR or something like that?

PR #917, opened 2020-12-14
Bug 1908889: Bump CNO to k8s 1.20

abhat:
This commit bumps the CNO mods to k8s 1.20.
For CNO we don't have a direct dependency on
openshift/kubernetes, so we don't need to carry
any patches from there. Instead this change, let's
CNO pull directly from upstream k8s 1.20.


go.mod
```diff
@@ -16,58 +16,63 @@ require (
...
-	k8s.io/kube-scheduler => github.com/openshift/kubernetes/staging/src/k8s.io/kube-scheduler v0.0.0-20201105175657-1bb012236ace
-	k8s.io/kubectl => github.com/openshift/kubernetes/staging/src/k8s.io/kubectl v0.0.0-20201105175657-1bb012236ace
-	k8s.io/kubelet => github.com/openshift/kubernetes/staging/src/k8s.io/kubelet v0.0.0-20201105175657-1bb012236ace
-	k8s.io/kubernetes => github.com/openshift/kubernetes v0.0.0-20201105175657-1bb012236ace
-	k8s.io/legacy-cloud-providers => github.com/openshift/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20201105175657-1bb012236ace
-	k8s.io/metrics => github.com/openshift/kubernetes/staging/src/k8s.io/metrics v0.0.0-20201105175657-1bb012236ace
-	k8s.io/sample-apiserver => github.com/openshift/kubernetes/staging/src/k8s.io/sample-apiserver v0.0.0-20201105175657-1bb012236ace
-	k8s.io/sample-cli-plugin => github.com/openshift/kubernetes/staging/src/k8s.io/sample-cli-plugin v0.0.0-20201105175657-1bb012236ace
-	k8s.io/sample-controller => github.com/openshift/kubernetes/staging/src/k8s.io/sample-controller v0.0.0-20201105175657-1bb012236ace
+	k8s.io/api => k8s.io/api v0.20.0
```
squeed, line 52: Presumably you can just get rid of all of these overrides?

abhat: I tried to start with nothing in the `replace` section of `go.mod` but it starts complaining about:

```
[anbhat@atman-centos-dev cluster-network-operator]$ go mod vendor
go: k8s.io/kubernetes@v1.20.0 requires
        k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
```
I think this seems to be a way to get around how things are referenced from the staging directory within k/k.

squeed: Weird, you're right. Huh.

danwinship:
/hold
@abhat `Not mergeable. Needs bugzilla/valid-bug label.`

abhat:
@danwinship had added the hold to keep the prow bots from re-running the jobs forever. Now that there is a valid bz associated with the PR, canceling the hold.
/hold cancel

PR #920, opened 2020-12-16
Bug 1912409: Fix flow schema deployment

alexanderConstantinescu:
PR: https://github.com/openshift/cluster-network-operator/pull/880 added a deployment stanza for a specific kubernetes feature, but only did that for openshift-sdn and not for ovn-kubernetes, moreover it added it in the wrong place independently of the network plugin being deployed.

I have no idea how this works and if the sequence of deployment is important or not for this stuff, so @tkashem could you please advise? 

/assign @danwinship


bindata/network/ovn-kubernetes/007-flowschema.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
+kind: FlowSchema
+metadata:
+  name: openshift-ovn-kubernetes
+  annotations:
+    include.release.openshift.io/self-managed-high-availability: "true"
```
danwinship, line 6: This annotation only makes sense for things in `manifests/` (so should be removed both here and from the sdn one)

danwinship (commented):
feature documentation is https://kubernetes.io/docs/concepts/cluster-administration/flow-control/

It looks like this is supposed to say "requests made by the given ServiceAccounts have `system` priority", which I believe is the highest priority level. OK. I agree that both the controller and the node service accounts should be high priority for both openshift-sdn and ovn-kubernetes.

tkashem:
- the sdn flowschema is in `manifests` folder and the ovn flowschema is in `bindata`. Is it possible to have both in `bindata`? 

- if this is part of the operator reconciliation logic, i wonder what happens if I disable the priority & fairness feature on my cluster. access to the flowschema resource will result in an error from the apiserver. should the operator logic be resilient to handle such error?

alexanderConstantinescu:
> * the sdn flowschema is in `manifests` folder and the ovn flowschema is in `bindata`. Is it possible to have both in `bindata`?

No, I just moved the sdn flowschema to `bindata`, so that is not a concern. My question was: does this resource need to be created with the deployment of the CNO, or can it wait until the deployment of the network plugin?

bindata/network/ovn-kubernetes/007-flowschema.yaml
alexanderConstantinescu: ACK, will change.

tkashem:
> My question was: does this resource need to be created with the deployment of the CNO, or can it wait until the deployment of the network plugin?

I can't speak to the operator logic, but as far as the kube-apiserver and p&f are concerned, there is no dependency. A flowschema object can be created even though the specified user/service account do not exist yet.

alexanderConstantinescu:
/hold 

Not sure why it's been re-running all jobs when all required ones have already passed. Also not sure why I need a valid bug for getting this stuff in on master.

alexanderConstantinescu:
/retitle Bug 1912409: Fix flow schema deployment

bindata/network/ovn-kubernetes/007-flowschema.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: flowcontrol.apiserver.k8s.io/v1alpha1
```
sttts, line 1: this must not use v1alpha1. v1alpha1 is subject to going away. Please switch to v1beta1 asap.

alexanderConstantinescu: OK, I will file a PR momentarily for that. Thanks for letting us now

tkashem: @alexanderConstantinescu based on the discussion with @sttts we should leave the version to `v1alpha` for 4.7. But we should set the version to `v1beta` in 4.8. Please see https://bugzilla.redhat.com/show_bug.cgi?id=1913399#c1

PR #922, opened 2020-12-17
ovn-ipsec: post feature-freeze CI bug fixes

markdgray:


bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -164,6 +164,25 @@ spec:
           }
           trap cleanup SIGTERM
 
+          # Don't start IPsec until ovnkube-node has finished setting up the node
+          counter=0
+          until [ ! -z $(kubectl get pods -l app=ovnkube-node  --field-selector spec.nodeName=${K8S_NODE} -o jsonpath='{.items[0].metadata.name}' 2>/dev/null) ]
```
danwinship, line 169: It seems weird to ask kubernetes about this... can't you just do local checks for the expected interface/ovn/ovs state?

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -183,19 +183,16 @@ spec:
           fi
           echo "ovnkube-node has configured node."
 
+          # After a restart of this container (or on initial startup), we flush xfrm state and policy
+          # before we start pluto and ovs-monitor-ipsec in order to start in a known good state. This
+          # will result in a small interruption in traffic until pluto and ovs-monitor-ipsec start again.
+          ip x s flush
+          ip x p flush
```
danwinship, line 190: Is there any chance this might flush someone else's state? (ie, is there any plausible way some other operator might be doing something with "ip x"? Or is that command basically limited to IPsec stuff only, and so there's no way someone else could be using it without interfering with us anyway?)

pkg/network/ovn_kubernetes.go
```diff
@@ -231,6 +231,11 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 		conf.DefaultNetwork.OVNKubernetesConfig = &operv1.OVNKubernetesConfig{}
 	}
 
+	var encapOverhead uint32 = 100 // 100 byte geneve header
```
danwinship, line 234: We actually only need 58 bytes, and the rest is overkill. So you could get away with a total overhead of 58+46=104.

Well, unless the cluster is using IPv6 in which case we need another 20 bytes. See #720. We had avoided merging that because we didn't want to tighten up the MTU until we had the ability to change it after install time. Need to decide what we're doing here.

I guess one possibility would be to use an overhead of 124 (58 geneve, 20 IPv6, 46 IPsec) when using IPsec for now, and then once we have the ability to change it post-install, we can fix it to only use the 20 bytes for IPv6 if they're actually needed.

mccv1r0, line 234: nit: use a constant?  If possible, let the compiler calculate this value based on all the components that make it up.

danwinship:
Can you also grab https://github.com/danwinship/cluster-network-operator/commit/6f1e7b5e on to this branch?

pkg/network/ovn_kubernetes.go
```diff
@@ -231,6 +231,11 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
...
+	if conf.DefaultNetwork.OVNKubernetesConfig.IPsecConfig != nil {
+		encapOverhead += 46 // 46 byte IPsec header
+	}
+
```
mccv1r0, line 238: nit: same here

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: @danwinship - It is namespaced so it will only affect the default namespace (host namespace). Therefore, we should be the only ones modifying this.

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: @danwinship I guess I could but I don't know what to check yet. I want to ensure that the node's networking is set-up before I do anything in case anything changes underneath me. What would be a good thing to check for to ensure that is the case? the cni conf file?

pkg/network/ovn_kubernetes.go
markdgray: I feel that is a bit out of the scope of this PR. I want to only decrease the MTU by the IPsec overhead. I don't know the background to those changes. Maybe I can do something like this somewhere just to make it clearer to the reader:

```
encapOverhead += geneveOverhead
```

markdgray: Good idea.

pkg/network/ovn_kubernetes.go
markdgray: Good idea.

pkg/network/ovn_kubernetes.go
danwinship: well, note at least that you are currently reserving 146 bytes for overhead, but per the math above, even if you want to unconditionally leave room for IPv6, you only need 124 bytes. So you could just say

```
	var encapOverhead uint32 = 100
	if conf.DefaultNetwork.OVNKubernetesConfig.IPsecConfig != nil {
		encapOverhead = 124
	}
```

markdgray: The IPsec overhead is 46 bytes on top of Geneve so I am just adding another 46 onto the (already specified) overhead. Basically, I want to call out all the overheads separately and add them on as needed. I will rework this (it will be a separate PR - https://github.com/openshift/cluster-network-operator/pull/928 ) and you can have a look and comment there if you would like me to change

> Can you also grab [danwinship@6f1e7b5](https://github.com/danwinship/cluster-network-operator/commit/6f1e7b5e) on to this branch?

I'm splitting this PR up so that the BZs line up correctly but I will add it to another PR if you would like

markdgray: I did this to some extent. I need to split up the PRs see - https://github.com/openshift/cluster-network-operator/pull/928

bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
markdgray: I created a new PR for this so I could map a Bz: https://github.com/openshift/cluster-network-operator/pull/929

markdgray:
See:
* https://github.com/openshift/cluster-network-operator/pull/929
* https://github.com/openshift/cluster-network-operator/pull/928
* https://github.com/openshift/cluster-network-operator/pull/927

PR #925, opened 2020-12-18
Bug 1905331: Multus containers should specify resource requests & limits

dougbtv:
This adds a fairly low request and limit, as the Multus daemonset must run in order for a cluster to become ready.

Additionally, these are primarily bash scripts that copy binaries, with the notable exception being the admission controller (which gets a more liberal request/limit).

Happy to take another opinion on how these are set, mostly my thinking was "bash by itself can take a meg of program space, I'll multiply that by 10." and then these utilize very little CPU.


smarterclayton:
I pulled all the remaining bits of this into https://github.com/openshift/cluster-network-operator/pull/1038, please review

PR #927, opened 2020-12-22
Bug 1910097: ovn-ipsec: Add resource limits

markdgray:
Fix test case in e2e-ovn-ipsec-step-registry:
"Managed cluster should ensure control plane pods do not run in
best-effort QoS" by requesting cpu and memory resources.

### DIFF
--- a/bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
+++ b/bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
@@ -132,6 +132,10 @@ spec:
           name: signer-ca
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
+        resources:
+          requests:
+            cpu: 10m
+            memory: 100Mi
         terminationMessagePolicy: FallbackToLogsOnError
         terminationGracePeriodSeconds: 10
       containers:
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes-ipsec/ovn-ipsec.yaml
```diff
@@ -132,6 +132,10 @@ spec:
           name: signer-ca
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
+        resources:
+          requests:
+            cpu: 100m
```
alexanderConstantinescu, line 137: Nit: this value is 10 times higher than what we usually specify in our ovn-kube daemonsets (see: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/ovnkube-node.yaml), any particular reason for that? Is it normal because of the encryption?

markdgray: I copied it from the old ovs yaml file at the very start of this work and haven't looked at it critically since then. I think I asked for feedback on this number in a previous PR but I didn't get any suggestions. However, now I can try to measure it. This is what I got on a 6 node cluster. I am not sure how this will scale but I suspect it wont scale linearly with the number of nodes. Also this is not a 'limit'. Therefore, I think I can set to cpu: 10m and memory: 100Mi
```
NAME              CPU(cores)   MEMORY(bytes)   
ovn-ipsec-27zc9   1m           44Mi
```

PR #935, opened 2021-01-05
Bug 1908914: Don't upgrade ovn-kube master until node update is rolled out

squeed:
This includes the following changes:
    
- A few utility methods for updating lists of objects
- Some logic to determine if it's safe to upgrade ovn-k master
- Plumbing to trigger the operconfig controller if any of its generated resources was updated on the server

/cc @danwinship @alexanderConstantinescu


squeed:
/test e2e-gcp-ovn
gcp fart.

pkg/controller/operconfig/pod_controller.go
```diff
@@ -4,21 +4,30 @@ import (
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // newPodReconciler returns a new reconcile.Reconciler
-func newPodReconciler(status *statusmanager.StatusManager) *ReconcilePods {
-	return &ReconcilePods{status: status}
+// When an update is receieved, it will trigger a status update
```
danwinship, line 16: "received"

pkg/controller/operconfig/pod_controller.go
```diff
@@ -42,6 +51,14 @@ func (r *ReconcilePods) Reconcile(request reconcile.Request) (reconcile.Result,
 	}
 
 	log.Printf("Reconciling update to %s/%s\n", request.Namespace, request.Name)
+	// Trigger an update to the operconfig reconciliation loop
+	r.eventChan <- event.GenericEvent{
+		// Hard coded, yes, but since reconcile.Request doesn't include
+		// any kind of object meta, there's not much we can do.
+		Meta: &metav1.ObjectMeta{
```
danwinship, line 58: might be nice to abstract this somehow.
`r.reconcileNotifier.Notify()`, etc

pkg/network/ovn_kubernetes.go
```diff
@@ -342,3 +378,38 @@ func listenDualStack(masterIP string) string {
...
+		klog.V(3).Infof("Not upgrading ovn-kube master yet, ovn-kube node DaemonSet still at version %s, need %s",
+			nodeVersion, releaseVersion)
+		return false
+	}
+
+	status := existingNode.Status
+	progressing := (status.UpdatedNumberScheduled < status.DesiredNumberScheduled ||
+		status.NumberUnavailable > 0 ||
+		status.NumberAvailable == 0 ||
+		existingNode.Generation > status.ObservedGeneration)
```
danwinship, line 407: This seems very magical and really needs at least a comment

danwinship (commented):
I wonder if we should have a more generic solution? We had wanted something similar in the past, to control sdn vs sdn-ovs rollout, before we switched to system OVS. We might need something else in the future?

(eg, there could be an annotation on the ovnkube-master DS indicating that it depended on the ovnkube-node DS)

But maybe we don't need that right now

squeed:
/hold

small fix needed.

pkg/controller/operconfig/pod_controller.go
squeed: good point, done.

squeed:
> I wonder if we should have a more generic solution?

@danwinship I thought about that, but I couldn't come up with an ordering notation I liked. This sort of stuff is annoyingly complex to express in plain text (i.e. SystemD), and easy enough to write in code.

The reality is we're going to need to add more and more conditional logic to the operator.

pkg/network/ovn_kubernetes.go
squeed: Agreed.

squeed:
/hold cancel
Small issue fixed

squeed:
@danwinship i had to de-dup the event stream from ReconcilePods because, somehow, controller-tools is broadcasting "duplicate" events. It took a small amount of ResourceVersion duct-tape but now it doesn't spam updates anymore.

squeed:
Checked out the test failures
- some cloud failures - GCP had some weird volume quota issue, so a lot of storage tests failed
- azure ran out of load balancers, so ingress never completed
- ovnkube-master just... never was scheduled on a node. We wrote the correct status, though, so that's nice!


/retest

dcbw:
@squeed is there any way we can do some testcases for the upgrade functionality that's been changed here?

squeed:
@dcbw yeah, I can write a few test cases.

BTW, this does the exact wrong thing for downgrades. We need to decide if that matters.

squeed:
@dcbw - added some test cases, repushed. PTAL.

squeed:
CI results look good.
@dcbw this is ready for final review.

squeed:
(There are no required jobs failing)

squeed:
all green!

alexanderConstantinescu (commented):
For what concerns the code in this patch: lgtm, besides my minor comment.  

However, are we certain that ovnkube-node is completely independent of ovnkube-master? Can it function properly and rollout if ovnkube-master is having troubles? This will most likely break us if there's a OVN NB and SB DB schema upgrade between the version N and version N+1 that ovnkube-node might be dependent upon. Say ovnkube-node is dependent upon a specific column in the database that only exists in v2, but v2 is never rolled out because ovnkube-master (which deploys the DB) is still waiting for ovnkube-node to finish. We will have a hung deployment, right? 

I am also wondering: we'll soon be removing the check on the master node count (see: #896) which means ovnkube-master and ovnkube-node might boot with 2 master nodes, and then very soon thereafter rollout to v2 on 3 master nodes....if ovnkube-node experiences a problem in v2, because ovnkube-master is still running in v1 or because it tries to connect to master nodes which don't have ovnkube-master running and properly configured yet, we'll never be able to unblock that, because we'll hold the ovnkube-master deployment.  

I guess what I am saying is: maybe we should try to rollout ovnkube-master after having retried X amount of time to rollout ovnkube-node and seen it fail. It might be that rolling out ovnkube-master to v2, unblocks the ovnkube-node v2 deployment in some corner cases.

pkg/controller/operconfig/pod_controller.go
```diff
@@ -1,48 +1,80 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("ReconcilePods: failed to retrieve object: %w", err)
+	}
+
+	if existing.GetResourceVersion() == newObj.GetResourceVersion() {
+		return reconcile.Result{}, nil
+	}
+
+	log.Printf("ReconcilePods %s was updated - triggering updates", nsn)
```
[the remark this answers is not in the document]
squeed: yeah, we need to move cno over to klog exclusively; it's on the list of things to do for 4.8.

squeed:
> Say ovnkube-node is dependent upon a specific column in the database that only exists in v2, but v2 is never rolled out because ovnkube-master (which deploys the DB) is still waiting for ovnkube-node to finish. We will have a hung deployment, right?

Great question, @alexanderConstantinescu. The genesis of this PR is that the OVN upstream supports upgrades **as long as the nodes (ovn-controller) are updated first**. In other words, ovn-controller can handle being "ahead" of ovn-northd.

If rollout hangs, then we'll fire an alert and the administrator will investigate. And we won't upgrade further.

pkg/controller/operconfig/pod_controller.go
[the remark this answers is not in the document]
alexanderConstantinescu: ack

alexanderConstantinescu:
> > Say ovnkube-node is dependent upon a specific column in the database that only exists in v2, but v2 is never rolled out because ovnkube-master (which deploys the DB) is still waiting for ovnkube-node to finish. We will have a hung deployment, right?
> 
> Great question, @alexanderConstantinescu. The genesis of this PR is that the OVN upstream supports upgrades **as long as the nodes (ovn-controller) are updated first**. In other words, ovn-controller can handle being "ahead" of ovn-northd.
> 
> If rollout hangs, then we'll fire an alert and the administrator will investigate. And we won't upgrade further.

OK and understood, but do you have any objections to what I mentioned:

> maybe we should try to rollout ovnkube-master after having retried X amount of time to rollout ovnkube-node and seen it fail. It might be that rolling out ovnkube-master to v2, unblocks the ovnkube-node v2 deployment in some corner cases.

As such we might be self-healing in some corner cases and spare the ~~cluster admin~~ us (since he/she probably won't fix it but just open a support case to Red Hat instead) the work?

pkg/apply/merge.go
```diff
@@ -91,6 +93,17 @@ func mergeServiceForUpdate(current, updated *uns.Unstructured) error {
...
+			return err
+		}
+		if found {
+			err = uns.SetNestedStringSlice(updated.Object, clusterIPs, "spec", "clusterIPs")
+			if err != nil {
+				return err
+			}
+		}
+
 		ipFamily, found, err := uns.NestedString(current.Object, "spec", "ipFamily")
```
danwinship, line 107: In master, `ipFamily` no longer exists, but we should be copying `ipFamilies` and `ipFamilyPolicy`. (You'd only notice that being a problem if you were running with the dual-stack feature gate, which is why you didn't notice that.)

pkg/controller/operconfig/pod_controller.go
```diff
@@ -1,48 +1,80 @@
...
+	newObj := (existing.NewEmptyInstance()).(*uns.Unstructured)
+	if err := r.client.Get(context.TODO(), nsn, newObj); err != nil {
+		if apierrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("ReconcilePods: failed to retrieve object: %w", err)
+	}
+
+	if existing.GetResourceVersion() == newObj.GetResourceVersion() {
+		return reconcile.Result{}, nil
```
danwinship, line 70: This all seems weird and fragile (like we could end up in an infinite loop if there's a small bug in it)

squeed: It is weird! There's definitely something wrong with the plumbing in `controller-runtime` where the PodReconciler is triggered even when it doesn't change on the server.

I started to look in to it, but hadn't gotten to the bottom of it.

I wish I had a better answer. We could also just set the reconciliation time to 2 minutes and just sweep this under the rug.

squeed:
> As such we might be self-healing in some corner cases and spare the cluster admin us (since he/she probably won't fix it but just open a support case to Red Hat instead) the work?

@alexanderConstantinescu yeah, that's not a terrible idea. We already detect hung rollouts in other cases. We just need to decide which is the better course: potentially causing issues, or making upgrade progress.

I think we can detect hung daemonsets and continue.

squeed:
Okay, took @alexanderConstantinescu's feedback in to account and now we'll force an upgrade if the node rollout is hung at 90% (rounded to one node).

@danwinship I don't have an answer for the potential fragility of the code. It is what it is. And there's no good way to test that we don't set up an infinite loop later. (Maybe this would be a good e2e test...). But I've tested it, and it works and goes quiescent.

squeed:
test failures seem like cloud provider issues (and lots of successful tests)
/retest

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +372,53 @@ func listenDualStack(masterIP string) string {
...
+	// Copy-pasted from status_manager: Figure out if a given DaemonSet is
+	// not fully rolled out.
+	nodeProgressing := (status.UpdatedNumberScheduled < status.DesiredNumberScheduled ||
+		status.NumberUnavailable > 0 ||
+		status.NumberAvailable == 0 ||
+		existingNode.Generation > status.ObservedGeneration)
+
+	// If progressing, don't upgrade masters yet.
+	// Exception: a rollout hung at 90% is overrided.
+	nodeHung := existingNode.GetAnnotations()[names.RolloutHungAnnotation] == "true"
```
alexanderConstantinescu, line 408: I am still not convinced about this, but I am willing to let it go. The concern I have is: if a roll-out is hung at X % (meaning it's made no progress at all for `ProgressTimeout` = 10 min), it's probably never going to make it to 90 %, and it feels like we should be forcing it in those cases too, not only when it's close to the finish line. Do you have any concern with just checking for the annotation being true, and forcing the roll-out in that case?

pkg/apply/merge.go
squeed: Good catch, updated.

squeed:
No, if the roll-out is hung at 10%, we should **not** push further. We stand a real risk of causing outages on nodes that haven't upgraded.

squeed:
OK @danwinship, removed the "updating daemonsets triggers daemonset upgrades" logic. Now we just re-reconcile every 3 minutes. That should be fine.

squeed:
test failures are spurious - e2e-gcp-ovn and e2e-gcp-ovn-upgrade both succeeded.

dcbw:
aws capacity issues, for one
/retest

dcbw:
@squeed looks sane to me; did you still need to answer the downgrade question? Or are we good to go here.

markdgray:
FWIW, I had a look at this. Its been pretty thoroughly reviewed so I didn't see any issues.

squeed:
Since we have a bit more breathing room, I'm going to add direction detection and just fix this.

squeed:
/hold
going to add directionality

squeed:
Okay, reworked this to take direction in to account.
Ready for re-review.

squeed:
Looked in to the test failures; looks like openstack-ovn had etcd issues, but definitely some weird error messages. Other failures don't seem OVN related.

/retest

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
 		return ""
 	}
 }
+
+// shouldUpdateDaemonSets determines if we should roll out changes to
```
squeed, line 385: oops, need to rename this.

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
...
+	// wait for master rollout
+	if masterDirection == directionUnchanged && nodeDirection == directionDowngrade {
+		if daemonSetProgressing(existingMaster) {
+			log.Printf("Waiting for OVN-Kubernetes master update to roll out")
+			return false, true
+		}
+		return true, true
+	}
+
+	if masterDirection == directionDowngrade && nodeDirection == directionDowngrade {
```
squeed, line 478: note to self: add comment here.

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
...
+	progressing := (status.UpdatedNumberScheduled < status.DesiredNumberScheduled ||
+		status.NumberUnavailable > 0 ||
+		status.NumberAvailable == 0 ||
+		ds.Generation > status.ObservedGeneration)
+
+	if !progressing {
+		return false
+	}
+
+	// If we're hung, but max(90% of nodes, 1) have been updated, then act as if not progressing
```
danwinship, line 503: we probably want this behavior for the node DS but not the master DS?

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
...
+	if existingNode == nil || existingMaster == nil {
+		return true, true
+	}
+
+	nodeVersion := existingNode.GetAnnotations()["release.openshift.io/version"]
+	masterVersion := existingMaster.GetAnnotations()["release.openshift.io/version"]
+
+	// shortcut - we're all rolled out.
+	if nodeVersion == releaseVersion && masterVersion == releaseVersion {
+		return true, true
```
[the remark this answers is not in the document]
squeed: Hah. yeah. I'll add a comment, since I don't want to flip the "parity" of the function.

pkg/network/semver.go
```diff
@@ -0,0 +1,30 @@
+package network
+
+import "github.com/Masterminds/semver"
```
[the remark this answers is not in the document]
squeed: That hard-codes the meaning of "alpha" and "beta"; I just want to sort the "extra" bits lexicographically so that CI works vaguely.

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
...
+	}
+	// weird: master alread upgraded, node needs upgrade
+	if masterDirection == directionUnchanged && nodeDirection == directionUpgrade {
+		return true, true
+	}
+	// weird: master needs downgrade, node needs upgrade
+	// in a bad state, just roll out changes
+	if masterDirection == directionDowngrade && nodeDirection == directionUpgrade {
+		log.Printf("WARNING: OVN-Kubernetes daemonset versions in an inconsistent state")
+		return true, true
```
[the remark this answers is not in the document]
squeed: yeah, good point.

pkg/network/ovn_kubernetes.go
squeed: agreed.

pkg/network/ovn_kubernetes.go
```diff
@@ -334,3 +381,133 @@ func listenDualStack(masterIP string) string {
...
+	}
+	// 9 cases
+	// +------------+---------------+-----------------+---------------+
+	// |  vs. CNO   | Master older  |   Master same   | Master newer  |
+	// +------------+---------------+-----------------+---------------+
+	// | Node older | Update node   | update node     | error         |
+	// | Node same  | wait for node | done            | update master |
+	// | Node newer | error         | wait for master | update master |
+	// +------------+---------------+-----------------+---------------+
```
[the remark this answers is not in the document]
squeed: I reworded it a bit. it's naturally confusing, since none of these chars define directionality super well. Let's hope it's enough context to debug this years from now.

squeed:
@danwinship thanks, as always, for the thorough review. enjoy the ascii charts :-). Updated.

squeed:
test failures seem like flakes; at least, the network operator performed as expected.

unfortunately the ovn upgrade test - the most important test - failed due to GCP api failures
/retest

squeed:
/hold
e2e-gcp-ovn-upgrade is running openshift-sdn

squeed:
filed https://github.com/openshift/release/pull/15254 to fix upgrade bug.

squeed:
hmm, need to close in favor of a new PR.

PR #939, opened 2021-01-08
Add alerts for OVN DB high election timer intervals

palanik1:
This PR adds alerts for high OVN North and SouthBound DB election timer intervals.

Co-authored by : Sayandeep sayandes@in.ibm.com

Signed-off-by: Palani Kodeswaran palani.kodeswaran@in.ibm.com

### DIFF
--- a/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NBDBLongElectionTimer
+      annotations:
+        message: |
+          OVN NB DB raft cluster has a high election timer interval.
+      expr: |
+        ovn_db_cluster_election_timer{db_name="OVN_Northbound"} > 3000
+      for: 10m
+      labels:
+        severity: warning
+    - alert: SBDBLongElectionTimer
+      annotations:
+        message: |
+          OVN SBDB raft cluster has a high election timer interval.
+      expr: |
+        ovn_db_cluster_election_timer{db_name="OVN_Southbound"} > 3000
+      for: 10m
+      labels:
+        severity: warning


bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NBDBLongElectionTimer
+      annotations:
+        message: |
+          OVN NB DB has a high election timer interval.
+      expr: |
+        ovn_db_cluster_election_timer{db_name="OVN_Northbound"} > 16000
```
vishnoianil, line 74: This is to check the existing limit and warn if this increases in future? If that's the case, that we might have to put this value a bit higher, because these values are based on 100 nodes with full workload. It might change when we move to 500 nodes (not baremetal) with full workload.

But if the idea is to warn that the value must not be set beyond x value for production environment, than i think we need to change this value to 3 seconds. Although it will generate the warning, but that's in my opinion is correct warning anyways.

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NBDBLongElectionTimer
+      annotations:
+        message: |
+          OVN NB DB has a high election timer interval.
```
vishnoianil, line 72: nitpick: "NB DB" -> "NB DB raft cluster"

palanik1: will fix that :-)

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
palanik1: @vishnoianil The raft election is still being run only between the three masters, or is the number of masters also a function of the number of nodes in the cluster?

vishnoianil: @palanik1 Raft election is still being run between the three instances of NB or SB database. Election timer is affected by  how much load (db transaction and data replication) existing implementation can handle. WIth increase of the load, you may have to bump up the election timer.

palanik1: Right...it grows as a function of the number of the flows..which I guess can be approximated as a function of the number of nodes. Will make the threshold 3 sec for now.

squeed:
Sorry for not getting to this sooner.

Unfortunately, after talking with @dceara, it turns out the value we scrape isn't going to be useful. Specifically, this is just the configured election deadline, rather than the time it took to elect.

In order for this to work, we need to scrape some more values from `appctl cluster/status` and determine the election time from that.

/hold

palanik1:
@squeed Any pointers on whether raft updates the election completion timestamp in any of the OVS DBs?

squeed:
@palanik1 according to @dceara, the thing to do is scrape the raft election start & stop times: https://github.com/openvswitch/ovs/blob/master/ovsdb/raft.c#L4506

However, it just isn't going to be that useful. We're going to have to do a re-think here. The problem is: we don't care how long a particular election took, per se. We just would like to alert on elections taking a long time in general. And, they happen rarely, so we aren't going to get good data.

While I would really really like to see this sort of alert, I'm not sure that we can actually reasonably alert on this. We really want a histogram of election times, and for that, we need a stateful component watching edges.

palanik1:
@squeed Agree..

squeed:
I think I know the easiest way to figure this out: If the average value over time of whether or not there is a leader is below a certain threshold, then we can alert.

vishnoianil:
> @palanik1 according to @dceara, the thing to do is scrape the raft election start & stop times: https://github.com/openvswitch/ovs/blob/master/ovsdb/raft.c#L4506
> 
> However, it just isn't going to be that useful. We're going to have to do a re-think here. The problem is: we don't care how long a particular election took, per se. We just would like to alert on elections taking a long time in general. And, they happen rarely, so we aren't going to get good data.
> 
> While I would really really like to see this sort of alert, I'm not sure that we can actually reasonably alert on this. We really want a histogram of election times, and for that, we need a stateful component watching edges.

@palanik1 I thought the idea was to throw an alert if user increase the election timer beyond a certain value that we support. If user end up setting 32 seconds, system should register the alert.

Also i think having an alert when cluster is taking longer for concluding the election would be useful as well. This alert would be like an early sign when cluster started being unstable, even though it recovers eventually. I raised an RFE (https://bugzilla.redhat.com/show_bug.cgi?id=1810730) to expose more info in cluster/status info, so that we can better debug/alert/monitor cluster status. Seems like there is a patch that does provide this information in cluster/status which you can use to alert.
http://patchwork.ozlabs.org/project/openvswitch/patch/fda4ec79f829881aca8e7be0f266a8907b4e99a5.1605629227.git.lorenzo.bianconi@redhat.com/

If you want to write this alert for the older OVN version, probably tracking the leadership state : UNKNOWN can give you how long cluster was without leader if you are running periodic checks.

squeed:
Oh, I didn't realize that was the purpose. It doesn't look like the user can change this value...

palanik1:
@squeed @vishnoianil Makes sense. Just to clarify that I understand your comments, there are two types of alerts if I understand it right. 
1) User sets a high election timer interval as @vishnoianil mentions. I think the current PR will probably handle this one.
2) Time it actually takes to complete a raft election, which is a metric available in the latest ovsdb.

@squeed I believe the CNO ovnkube-master.yaml updates the cluster_election_timer interval after a couple of retries

https://github.com/openshift/cluster-network-operator/blob/9358ad75fead31dcee3e17e37f17acddb39b0aed/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L368

Would this make it a case to raise an alert for (1)

palanik1:
@vishnoianil @dceara @squeed what do you think?

vishnoianil:
> @vishnoianil @dceara @squeed what do you think?

@palanik1 @squeed User can change value through CNO or it can go directly and update the value using ovs-appctl directly (we don't enforce any permissions on ovs-appctl as such). If we assume these scenarios are unlikely, i think at this point (1) won't give much value. Once election-timer can be set as an initial configuration, probably having this election time makes bit more sense as it can notify if the user set the value higher than what ovn-k supports.

squeed:
Right, but it will be reset any time anything restart ovn-kubernetes.

I think this alert is not needed.

PR #941, opened 2021-01-08
Add Alerts for high number of ovn db logs not committed

sdsen:
This PR adds alerts for high number of logs not committed in OVN North and South bound DBs.

Co-authored by : Palani Kodeswaran palani.kodeswaran@in.ibm.com

Signed-off-by: Sayandeep <sayandes@in.ibm.com>

### DIFF
--- a/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NorthBoundLogsNotCommitted
+      annotations:
+        message: |
+          NorthBound DB has too many logs not committed.
+      expr: |
+        ovn_db_cluster_log_not_applied{db_name="OVN_Northbound"} > 100
+      for: 10m
+      labels:
+        severity: warning
+    - alert: SouthBoundLogsNotCommitted
+      annotations:
+        message: |
+          SouthBound DB has too many logs not committed.
+      expr: |
+        ovn_db_cluster_log_not_applied{db_name="OVN_Southbound"} > 100
+      for: 10m
+      labels:
+        severity: warning 


bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NorthBoundLogsNotCommitted
+      annotations:
+        message: |
+          NorthBound DB has too many logs not committed.
+      expr: |
+        ovn_db_cluster_log_not_applied{db_name="OVN_Northbound"} > 100
+      for: 10m
```
aojea, line 75: just curiosity, what is the criteria to use these numbers, 100 events for 10 minutes?

sdsen: The idea we had is to see if more than 100 (not sure if this applies to all cluster sizes)  logs have not been committed in a 10 minute span.  

This specific numbers 100/10min is something we have put based on empirical observations, any suggestions on how to get better thresholds ?

aojea: no idea sorry, was just curious
it LGTM then

markdgray:
Bit of a nit but could you change commit message to imperative: "**Add** alerts for high number of ovn db logs not committed"

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
markdgray: Have you tested on large clusters? Does this value increase with respect to cluster size?

sdsen: Intuitively, with growing cluster sizes the rate of log writes should increase and hence, this static bound will become less stringent (ie will miss more alert worthy cases, but false positives will go down).

But(to answer your question), no haven't validated via experimentations, any pointers on how we can do those ? (get access to scale test infra) ?

markdgray: No (and I wouldnt block this PR because of that comment). If you were curious, I guess the scale team would be deploying large clusters but not sure if it would be hitting any alert worthy cases.

abhat: this alert template as I understand it says if we for all the monitoring samples, we see 100 logs not committed for a 10 min. period raise an alert. Is that what we want?

palanik1: @abhat Right..we have now made it a rate so that it doesnt trigger continuously.

palanik1:
@squeed what do you think?

PR #943, opened 2021-01-11
Added alerts high rate of connection errors in OVN DBs

palanik1:
Co-authored by : Sayandeep Sen sayandes@in.ibm.com

Signed-off-by: Palani Kodeswaran <palankod@in.ibm.com>

This PR adds alerts for too many connection errors in OVN DBs.

### DIFF
--- a/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NBDBConnectionErrors
+      annotations:
+        message: |
+          OVN NBDB has a high rate of connection errors
+      expr: |
+          ( rate(ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Northbound"}[5m]) + rate(ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Northbound"}[5m]) ) > 100
+      for: 10m
+      labels:
+        severity: warning
+    - alert: SBDBConnectionErrors
+      annotations:
+        message: |
+          OVN SBDB has a high rate of connection errors
+      expr: |
+        ( rate(ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Southbound"}[5m]) + rate(ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Southbound"}[5m]) ) > 100
+      for: 10m
+      labels:
+        severity: warning 


bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NorthBoundConnectionsError
+      annotations:
+        message: |
+          NorthBound DB has too many errors in connections.
+      expr: |
+          (ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Northbound"} + ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Northbound"}) > 100
```
squeed, line 74: These values are counters, which means they are monotonic. So, you need to compute a rate of errors instead. Otherwise this alert will fire "forever", even if the rate of errors is one per hour, after 100 hours.

So the query might look something like

```
(rate(ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Northbound"})[5m] + rate(ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Northbound"})[5m] ) > 100
```

(This applies to the southbound alert as well)

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
       for: 10m
       labels:
         severity: warning
+    - alert: NorthBoundConnectionsError
```
squeed, line 69: suggestion: `NBDBConnectionErrors`

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
```diff
@@ -66,3 +66,21 @@ spec:
...
+      annotations:
+        message: |
+          NorthBound DB has too many errors in connections.
```
squeed, line 72: "OVN NBDB has a high rate of connection errors"

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
palanik1: Done.

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
palanik1: Done.

bindata/network/ovn-kubernetes/alert-rules-control-plane.yaml
palanik1: @squeed Good point..should have made it rate. Thanks.

squeed:
Seems reasonable. Would you mind starting a cluster (with cluster-bot), then triggering an upgrade (again, with cluser-bot) and ensuring that noise from upgrades doesn't trigger this?

Assuming that works, we can get this in.

PR #877, opened 2020-11-13
Bug 1914932: Put correct resource name in relatedObjects

tremes:

PR #873, opened 2020-11-11
Add operator.openshift.io/v1/network to relatedObjects for must-gather

rcarrillocruz:
Fixes [SDN-1344](https://issues.redhat.com/browse/SDN-1344)

PR #945, opened 2021-01-12
Bug 1914932: Put correct resource name in relatedObjects

alexanderConstantinescu:
PR: https://github.com/openshift/cluster-network-operator/pull/873 was missing a correct resource name for `networks.operator.openshift.io` to be exported properly to a must-gather. PR: https://github.com/openshift/cluster-network-operator/pull/877 tried fixing this, but was missing a piece to be able to pass the CI upgrade jobs, which this PR now adds. 

Essentially upgrade jobs upgrade from HEAD to HEAD + patch. So with #873: HEAD now has a resource name which is incorrect, if we now want to fix this we will incorrectly delete `networks.operator.openshift.io` on such upgrades. This is not yet a problem for any OpenShift release, since #873 is only on master. The only fix I could think of was to explicitly skip the deletion of the CRD for now. Once this PR has merged we can then delete that explicit skip, since master will have the correct resource name and we won't expect that to change thereafter. I will file a PR for that once this integrates. 

/assign @squeed

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -70,12 +70,10 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 	if status.relatedObjects == nil {
 		return
 	}
-
 	for _, currentObj := range co.Status.RelatedObjects {
 		var found bool = false
 		for _, renderedObj := range status.relatedObjects {
 			found = reflect.DeepEqual(currentObj, renderedObj)
-
 			if found {
 				break
 			}
@@ -95,6 +93,11 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				// BZ 1820472: During SDN migration, deleting a namespace object may get stuck in 'Terminating' forever if the cluster network doesn't working as expected.
 				// We choose to not delete the namespace here but to ask user do it manually after the cluster is back to normal state.
 				log.Printf("Object Kind is Namespace, skip")
+				continue
+			}
+			// @aconstan: remove this after having the PR implementing this change, integrated.
+			if gvk.Kind == "Network" && gvk.Group == "operator.openshift.io" {
+				log.Printf("Object Kind is network.operator.openshift.io, skip")
 				continue
 			}
 			log.Printf("Detected related object with GVK %+v, namespace %v and name %v not rendered by manifests, deleting...", gvk, currentObj.Namespace, currentObj.Name)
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu:
Hooray, upgrades pass! Let's review this and get it in! 

@squeed : could you review this tomorrow? 

/retitle Bug 1914932: Put correct resource name in relatedObjects

squeed:
Suggestion: just always add the object in `SetRelatedObjects`. Then you don't have to worry about skipping it on delete.

alexanderConstantinescu:
> Suggestion: just always add the object in `SetRelatedObjects`. Then you don't have to worry about skipping it on delete.

That's what's happening here: https://github.com/openshift/cluster-network-operator/blob/b418646912268460723ba65c35d81d43cbcf90dd/pkg/controller/operconfig/operconfig_controller.go#L250 , right? It's still getting deleted on the update jobs since the resource field changes between the versions. 

Anyhow, the goal is to remove the skip altogether in a succeeding PR as I mentioned in the description.

squeed:
That's a good point.
/lgtm
/retest

alexanderConstantinescu:
@danwinship : could you override e2e-aws-sdn-multi ? It's history looks grim https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-aws-sdn-multi and I don't want to hold this PR anymore because of it

alexanderConstantinescu:
/cherry-pick release-4.6 cancel

We shouldn't cherry-pick this as-is. I need another PR towards master after this one merges, as mentioned in the description.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -97,6 +95,11 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				log.Printf("Object Kind is Namespace, skip")
 				continue
 			}
+			// @aconstan: remove this after having the PR implementing this change, integrated.
```
danwinship, line 98: Ugh. If we add new objects to `relatedObjects` to get them must-gathered, then that means we'll end up deleting those objects when you downgrade to a version of CNO where they weren't part of `relatedObjects`. This may be the cause of the "openshift-sdn CRDs get deleted when downgrading" bug? (Not sure where that is or who was working on it...)

We need to avoid doing `deleteRelatedObjectsNotRendered` on downgrade. @squeed 

For now, the fix is probably to backport this hack to 4.6 and 4.5

alexanderConstantinescu: It is the cause of it, @juanluisvaladas verified it with PR: https://github.com/openshift/cluster-network-operator/pull/948. We need this back-ported.

PR #949, opened 2021-01-13
Bug 1915885: Kuryr: Support multiple nodes subnets

dulek:
Bare OpenShift on OpenStack recently added support for running worker
nodes in separate subnets. By default Kuryr assumes that workers run on
a single subnets. In order to support aforementioned feature Kuryr needs
to be reconfigured to get subnets by watching Machines.

While CNO no longer needs to configure Kuryr with the worker node subnet
ID, it still needs to know it in order to get MTU of that network and
create correct security groups allowing traffic to the nodes. Due to
other changes it's now required to determine that by looking at the
Machine objects too as primary subnet is no longer tagged by the
installer.

This also required adding machine-api-operator and cluster-api-provider-openstack to deps.


bindata/network/kuryr/001-rbac.yaml
```diff
@@ -58,7 +58,13 @@ rules:
...
   verbs:
   - use
-
+- apiGroups: ["machine.openshift.io"]
+  resources:
+  - machines
+  verbs:
+  - get
+  - list
+  - watch
```
MaysaMacedo, line 67: Are the verbs get and watch needed even when only List is used?

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -189,6 +194,55 @@ func getSavedAnnotation(kubeClient client.Client, annotation string) (string, er
 	return cm.Annotations[annotation], nil
 }
 
+func getMasterMachines(kubeClient client.Client) (*machineapi.MachineList, error) {
```
MaysaMacedo, line 197: Probably MachineSets could be used instead of Machine,` oc get MachineSets -n openshift-machine-api`, this is created since the first steps of installation and also has a label applied `machine.openshift.io/cluster-api-machine-role: worker`.
As the subnet is only needed for mtu and router discovery, it should be fine to rely on the master subnet as the mtu and router would need to be the same, so no strong opinions about this.

MaysaMacedo (commented):
Looks great. Just a few questions and small suggestion:

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: actually I see there are sg rules that uses the worker cidr, this could end up not being the same for master subnets

bindata/network/kuryr/001-rbac.yaml
MaysaMacedo: ah, I just saw this will be used on Kuryr itself. Disregard this comment.

dulek: We're actually doing watch (it's for Kuryr). I'll check if it's possible to get rid of get and list one.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: The master machines have no MachineSet, moreover it is totally valid to create Machines without MachineSets (just like Pods without any higher-level controllers like DaemonSet). So I'd rather stick to using lowest level resources - Machines.

bindata/network/kuryr/001-rbac.yaml
dulek: On the other head it's just `get` and `list`, nothing too crazy here. I'll leave those in case in the future Kuryr will actually need to list them.

pkg/platform/openstack/networking.go
```diff
@@ -132,7 +132,7 @@ func getOpenStackNetworkMTU(client *gophercloud.ServiceClient, networkID string)
 
 // Looks for a Neutron subnet by name and tag. In case of not found, looks for a
 // provided custom subnet by tag. Fails if any not found.
```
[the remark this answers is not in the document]
dulek: Thanks!

pkg/platform/openstack/kuryr_bootstrap.go
dulek: Ah, and regarding the SGs - the assumption is that if somebody spawns nodes on a separate subnet, then he or she is responsible for creating all the SG rules. Martin prepared a playbook for the bare SoS and I have one for Kuryr too.

MaysaMacedo: okay

dulek:
/hold

This needs to wait for the Kuryr piece.

dulek:
One more improvement to support older BYON deployments being upgraded from 4.6 to 4.7.

dulek:
/hold cancel

Kuryr piece is merged now.

pkg/bootstrap/types.go
```diff
@@ -8,7 +8,6 @@ type KuryrBootstrapResult struct {
 	ServiceSubnet            string
 	PodSubnetpool            string
 	WorkerNodesRouter        string
-	WorkerNodesSubnet        string
```
danielmellado, line 11: I'm assuming you're removing this as you now import "github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"

dulek: I'm removing this because we no longer need CNO to pass detected *main* worker nodes subnet to the Kuryr configuration. It's because we're now detecting that directly in kuryr-controller by watching the Machine objects in order to make sure we're able to react to new Machines being spawned on separate subnets.

dulek: BTW - CNO still needs to **detect** the _main_ worker nodes subnet (in a different way though, now it needs to list Machines due to changes in the installer) in order to have its CIDR and create correct SG rules for Kuryr. The ShiftOnStack assumption is that for nodes created on separate subnets it's the users responsibility to create the SG and SG rules matching that CIDR. That's why CNO only needs one _main_ worker nodes subnet.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -189,6 +194,57 @@ func getSavedAnnotation(kubeClient client.Client, annotation string) (string, er
...
+	return machineList, nil
+}
+
+func getWorkersSubnetFromMasters(client *gophercloud.ServiceClient, kubeClient client.Client, clusterID string) (subnets.Subnet, error) {
+	empty := subnets.Subnet{}
+	machines, err := getMasterMachines(kubeClient)
+	if err != nil {
+		return empty, err
+	}
+	for _, machine := range machines.Items {
```
aojea, line 216: what happens if `len(machines.Items) == 0`?

danielmellado: would that ever be reached?

dulek: This would be considered a weird situation as Machines for master nodes are created by the installer. With 4.7 we do not have another way of determining the main worker subnet, so it'd fail on `findOpenStackSubnetByNetworkTag()` execution by default. Interestingly in envs with BYON upgraded from 4.6 to 4.7 the `findOpenStackSubnetByNetworkTag()` would successfully find the network - thanks to guarantee that it's tagged in Neutron. This is gone in 4.7 due to users requests (some doesn't have permissions to tag networks).

Anyway if there are no Machines created for masters then there's not much we can do and I'd be happy that CNO fails bootstrapping process here. It will be retried if somehow the Machines come back.

dulek:
/test e2e-aws-sdn-multi
--

PR #950, opened 2021-01-13
Bug 1912820: openshift-apiserver Available is False with 3 pods not ready for a while during upgrade 

sanchezl:
Create a "dummy" PodNetworkConnectivityCheck instance that will [prevent the PodNetworkConnectivityCheck CRD from being deleted](https://github.com/openshift/library-go/blob/e59ac21aada35d4562cf96a1eaf6cec3929b9778/pkg/operator/connectivitycheckcontroller/connectivity_check_controller.go#L139-L146) by v4.6 ConnectivityCheckController instances during a 4.6->4.7 upgrade

### DIFF
--- a/manifests/0000_11_cluster-network-operator_00_upgrade_check.yaml
+++ b/manifests/0000_11_cluster-network-operator_00_upgrade_check.yaml
@@ -1,0 +1,13 @@
+apiVersion: controlplane.operator.openshift.io/v1alpha1
+kind: PodNetworkConnectivityCheck
+metadata:
+  name: 46-to-47-upgrade
+  namespace: openshift-network-operator
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  sourcePod: '46-to-47-upgrade'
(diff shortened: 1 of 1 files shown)


rcarrillocruz:
Does it need to be openshift-config namespace?
I don't think has perms for that namespace:

https://github.com/openshift/cluster-network-operator/blob/master/bindata/network-diagnostics/001-rbac.yaml

sanchezl:
> Does it need to be openshift-config namespace?

@rcarrillocruz, I just need it to not be deleted by a disabled connectivity check controller. Any namespace not used by a connectivity check controller will do. Given that `openshift-network-diagnostics` is [used by the network operator's connectivity check controller](https://github.com/openshift/cluster-network-operator/blob/c6a7d458d1bc6cc715f3ded595778f06988918a1/pkg/controller/connectivitycheck/connectivity_check_controller.go#L56), I've changed it to `openshift-network-operator` to keep it closer to home. 

> I don't think has perms for that namespace:

All known connectivity check controller instances run in operators using service accounts bound to `cluster-admin` roles. The actual `check-endpoints` instances are not involved in this fix.

danwinship:
This is ugly but it seems like the simplest fix given that it's right before code freeze
/approve
but I'll let @deads2k argue more forcefully against it or lgtm it

sanchezl:
superseded by openshift/cluster-network-operator#970

PR #952, opened 2021-01-13
Bug 1914402: ipsec: Add liveness probe

markdgray:
In the case of failure of ipsec daemons, add liveness
probe to detect so that the kubelet can restart the
container.

### DIFF
--- a/bindata/network/ovn-kubernetes/ipsec.yaml
+++ b/bindata/network/ovn-kubernetes/ipsec.yaml
@@ -242,6 +242,16 @@ spec:
           initialDelaySeconds: 120
           periodSeconds: 30
           timeoutSeconds: 40
+        livenessProbe:
+          exec:
+            command:
+            - /bin/bash
+            - -c
+            - |
+              #!/bin/bash
+              ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
+          initialDelaySeconds: 60
+          periodSeconds: 60
         terminationGracePeriodSeconds: 10
       nodeSelector:
         beta.kubernetes.io/os: "linux"
--- a/bindata/network/ovn-kubernetes/ipsec.yaml
+++ b/bindata/network/ovn-kubernetes/ipsec.yaml
@@ -238,7 +238,7 @@ spec:
             - -c
             - |
               #!/bin/bash
-              ovs-appctl -t ovs-monitor-ipsec ipsec/status
+              ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
           initialDelaySeconds: 120
           periodSeconds: 30
           timeoutSeconds: 40


markdgray:
ipsec is failing with some kind of aws quota

/retest

markdgray:
unrelated panic in ipsec case

bindata/network/ovn-kubernetes/ipsec.yaml
```diff
@@ -238,10 +238,20 @@ spec:
             - -c
             - |
               #!/bin/bash
-              ovs-appctl -t ovs-monitor-ipsec ipsec/status
+              ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
```
astoycos, line 241: Was this broken before?

bindata/network/ovn-kubernetes/ipsec.yaml
```diff
@@ -238,10 +238,20 @@ spec:
...
           initialDelaySeconds: 120
           periodSeconds: 30
           timeoutSeconds: 40
+        livenessProbe:
```
astoycos, line 245: So readiness and liveliness probe are the same for the ipsec containers?

markdgray: Yes.

bindata/network/ovn-kubernetes/ipsec.yaml
markdgray: Not broken but this is an improvement. I will split this into a separate commit to make it clear. Thanks for the review.

bindata/network/ovn-kubernetes/ipsec.yaml
astoycos: Is that alright? I'm not super familiar but I assume we would rather have two checks?

markdgray: A readiness probe should indicate when a pod is ready to receive traffic. This makes sense when you are building a service on top of the cluster as it will indicate when the service is ready to accept traffic. I am not sure how much it is a requirement for these networking pods. I don't think it hurts as it will at least give an indication when the pod is ready to receive IKE traffic (i.e. when the IPsec daemons are up). The liveness probe is for detecting failures of the pod. At the moment, I think the biggest failure that I want to detect is the daemons just falling over (e.g. with a segfault) so I think this liveness probe is appropriate. They are the same now but maybe in the future they might change. I don't think that I see an issue with that.

markdgray:
unrelated panic in pods/openshift-operator-lifecycle-manager_packageserver for ipsec job

/retest

markdgray:
ipsec: looks unrelated. machine config operator and one worker node was not ready.

/test e2e-ovn-ipsec-step-registry

markdgray:
ipsec has passed. As this PR is only for the case when IPsec is enabled. Any other failures are unrelated.

/retest

markdgray:
ipsec has passed. As this PR is only for the case when IPsec is enabled. Any other failures are unrelated.

/retest

PR #953, opened 2021-01-15
Bug 1916601: guard against the deletion of the network CRD: `network.operator.openshift.io`

alexanderConstantinescu:
4.6 backport of #945 

Given that 4.6 supports egress IP, egress firewall, etc, and that those objects have been added to "related objects" in 4.7, we need to add them in 4.6 too, since: on a downgrade those CRDs would be deleted. This is not bad since we'll have them in the must-gathers on 4.6. 

I am also removing the skip in `status_manager.go` (see: https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/statusmanager/status_manager.go#L99-L102) on the CRD `network.operator.openshift.io`, since on a 4.6 -> 4.5 - once we have the 4.5 backport in - it should not get deleted anymore since that CRD will be in the "related objects" in both versions.

/assign @squeed @juanluisvaladas @danwinship

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -236,6 +236,11 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		})
 	}
 
+	// Be very careful of the objects added to `relatedObjects`,
+	// if the `ObjectReference` points to a specific object and
+	// that object is not explcitly skipped in `deleteRelatedObjectsNotRendered`
+	// then we'll delete it on a downgrade. Background:
+	// https://github.com/openshift/cluster-network-operator/pull/945
 	relatedObjects = append(relatedObjects, configv1.ObjectReference{
 		Resource: "namespaces",
 		Name:     names.APPLIED_NAMESPACE,
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -91,6 +91,10 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				log.Printf("Object Kind is Namespace, skip")
 				continue
 			}
+			if gvk.Kind == "Network" && gvk.Group == "operator.openshift.io" {
+				log.Printf("Object Kind is network.operator.openshift.io, skip")
+				continue
+			}
 			log.Printf("Detected related object with GVK %+v, namespace %v and name %v not rendered by manifests, deleting...", gvk, currentObj.Namespace, currentObj.Name)
 			objToDelete := &uns.Unstructured{}
 			objToDelete.SetName(currentObj.Name)


pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -241,6 +241,29 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
...
+		Group:    "operator.openshift.io",
+		Resource: "networks",
+		Name:     "cluster",
+	})
+
+	// Add NetworkPolicy, EgressFirewall, EgressIP, for must-gather
+	relatedObjects = append(relatedObjects, configv1.ObjectReference{
+		Group:    "networking.k8s.io",
+		Resource: "NetworkPolicy",
+	})
```
danwinship, line 255: This is not actually needed to prevent objects from being deleted; we only delete objects when the `ObjectReference` points to a _specific_ object, not when it points to an entire class of objects.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -241,6 +241,29 @@ func (r *ReconcileOperConfig) Reconcile(request reconcile.Request) (reconcile.Re
 		Name:     names.APPLIED_NAMESPACE,
 	})
 
+	// Add operator.openshift.io/v1/network to relatedObjects for must-gather
+	relatedObjects = append(relatedObjects, configv1.ObjectReference{
+		Group:    "operator.openshift.io",
+		Resource: "networks",
+		Name:     "cluster",
+	})
```
danwinship, line 249: This _does_ prevent the network config from being deleted, but I think that's the wrong fix; what if someone upgrades to this version of 4.6 but then has to downgrade back to an earlier 4.6 for some reason?

In 4.5 and 4.6, we should add the hack to never delete the operator config, but we shouldn't add the operator config to `relatedObjects`. (And we should maybe add a comment warning that it's not safe to add new objects in 4.5 and 4.6.)

danwinship (changes requested):

pkg/controller/operconfig/operconfig_controller.go
alexanderConstantinescu: You're right, I completely overlooked that. Thanks for spotting it.

pkg/controller/operconfig/operconfig_controller.go
alexanderConstantinescu: > what if someone upgrades to this version of 4.6 but then has to downgrade back to an earlier 4.6 for some reason?

True as well. I will change and only keep the "never delete hack" + comment.

alexanderConstantinescu:
@danwinship : updated the PR, PTAL.

bparees:
@alexanderConstantinescu this PR does not seem to do what the pr title says.  It is not updating the set of relatedObjects at all.  I need more information on how this relates to the linked bug, and the 4.7 PR, before i can cherry pick approve it.

/hold

alexanderConstantinescu:
> I need more information on how this relates to the linked bug, and the 4.7 PR, before i can cherry pick approve it.

In 4.7 (PR: #873) we added the `network.operator.openshift.io` CRD to the list of related objects, as to have the CRD parsed by the CCX team, who's mechanism for gathering OCP CRD based metrics passes through the list of related objects. The resource name had however been mistakenly written, which resulted in both the CCX team not being able to parse the CRD and also issues when downgrading from 4.7 (what was then master) to 4.6. PR: #945 fixed that, by setting the correct name and adding a check guarding against the deletion of the CRD by the CNO's garbage collector. 

This PR is the back-port of the guarding check.  

/retitle Bug 1916601: guard against the deletion of the network CRD: `network.operator.openshift.io`

bparees:
@alexanderConstantinescu thanks for the explanation!

/hold cancel

rhpmali:
As I am not the QA contact for BZ 1916601 so did not follow the pre-merge workflow. But while testing its failed. Details are in https://bugzilla.redhat.com/show_bug.cgi?id=1916601#c3

bparees:
> As I am not the QA contact for BZ 1916601 so did not follow the pre-merge workflow. But while testing its failed. Details are in https://bugzilla.redhat.com/show_bug.cgi?id=1916601#c3

thanks, removing CPA for now.

alexanderConstantinescu:
> > As I am not the QA contact for BZ 1916601 so did not follow the pre-merge workflow. But while testing its failed. Details are in https://bugzilla.redhat.com/show_bug.cgi?id=1916601#c3
> 
> thanks, removing CPA for now.

@bparees :  I am new to this process. What is expected from me in this case? I have had a look at the must-gather, trying to understand why 4.6 -> 4.7 -> 4.6 does not work (again...). It's clearly not a problem in the CNO - as the CRD is still there when downgrading to 4.6 and the CNO is reporting that it skipped deleting the CRD (which is what this patch implements). 

The MCO is reporting degraded, and it's because:

```
2021-02-12T15:15:11.378046030Z E0212 15:15:11.378002       1 operator.go:314] timed out waiting for the condition during syncRequiredMachineConfigPools: pool master has not progressed to latest configuration: controller version mismatch for rendered-master-6bc549ba9b1e2c4e7fba579860116e63 expected 14a2b82d9f4c4d8b423f8f05f6926778ef36870d has 0023e696058bbdf6e14504117bfc31f208125c47: pool is degraded because nodes fail with "3 nodes are reporting degraded status on sync": "Node ip-10-0-221-73.us-east-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-master-6bc549ba9b1e2c4e7fba579860116e63: Failed to parse Ignition for validation: parsing Ignition config failed: unknown version. Supported spec versions: 2.2, 3.0, 3.1\", Node ip-10-0-138-205.us-east-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-master-6bc549ba9b1e2c4e7fba579860116e63: Failed to parse Ignition for validation: parsing Ignition config failed: unknown version. Supported spec versions: 2.2, 3.0, 3.1\", Node ip-10-0-170-240.us-east-2.compute.internal is reporting: \"unexpected on-disk state validating against rendered-master-6bc549ba9b1e2c4e7fba579860116e63: Failed to parse Ignition for validation: parsing Ignition config failed: unknown version. Supported spec versions: 2.2, 3.0, 3.1\"", retrying
```

Which is most likely caused by this: https://bugzilla.redhat.com/show_bug.cgi?id=1911841

bparees:
> @bparees : I am new to this process. What is expected from me in this case? 

normally we'd merge the PR and QE would test it.  But since in this case it seems like we have evidence this PR is not going to fix the bug, what's expected is to iterate on the PR until it fixes the bug.

If you think that this PR does in fact resolve the original issue, and that the verification effort undertaken found some other problem, then you should clarify what the expected verification steps are in the BZ this PR is addressing, so that QE can properly confirm the fix.

Since you seem to have dug into the recreate attempt and feel that this PR does address the original issue, i'll go ahead and re-tag it, but please update the BZ so that QE knows how to verify this fix once it merges.

PR #954, opened 2021-01-15
Bug 1903660: Don't error when expected master node amount is not met 

alexanderConstantinescu:
PR #896 tried to unblock the assisted installer team by reverting commit a14003d from #421. However the level entropy introduced by removing that commit can in some situations bring the OVN cluster down. Essentially, if the amount of master nodes are not met during the initial cluster deployment: we risk rolling out successive versions of the OVN DB cluster in a very short time span. Since the NB / SB DB raft clusters does not have the time to stabilize before the new version rolls out, we risk doing bad things in some cases. 

This PR does things as before, except we don't block the cluster deployment if the master node replica count is not met. Thus we will still wait for all master nodes that should be up, are up. However if they are not, we proceed with the deployment of the initial cluster and then proceed to roll out the second version once the new master node(s) join. This has been tested by the assisted installer team using PR #934, and it works for them. 

We will still annotate `network.operator.openshift.io` with the raft join point as #896 did, for cases such as the assisted installer use case.

/assign @squeed @abhat

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -153,7 +153,7 @@ spec:
           # end of db_part_of_cluster
           
           # Checks if cluster has already been initialized.
-          # If not it returns false and sets init_ip to MASTER_IP
+          # If not it returns false and sets init_ip to CLUSTER_INITIATOR_IP
           cluster_exists() {
             local db=${1}
             local port=${2}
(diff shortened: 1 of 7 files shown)


squeed:
This seems okay, but I'd like to look in to, also, why the ovndbchecker didn't handle this case correctly.

Basically, we should be able to handle this kind of churn. I know it's not pretty, but, ultimately, we need to deal with it.

alexanderConstantinescu:
> This seems okay, but I'd like to look in to, also, why the ovndbchecker didn't handle this case correctly.
> 
> Basically, we should be able to handle this kind of churn. I know it's not pretty, but, ultimately, we need to deal with it.

I already mentioned why, according to my findings that didn't happen, see: https://github.com/openshift/cluster-network-operator/pull/896#issuecomment-760895280

abhat:
/lgtm

We do need to figure out what the gaps are with the daemonset scripts not being able to handle rapid roll-outs.

PR #955, opened 2021-01-20
Add script to launch openshift-sdn in kind

juanluisvaladas:
Currently does not support multus.

### DIFF
--- a/hack/sdn-kind-cno.sh
+++ b/hack/sdn-kind-cno.sh
@@ -1,0 +1,236 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Version v1.19 or higher is required
+K8S_VERSION=${K8S_VERSION:-v1.20.2}
+BUILD_SDN=${BUILD_SDN:-false}
+BUILD_CNO=${BUILD_CNO:-false}
+BUILD_MULTUS=${BUILD_MULTUS:-false}
+CNO_PATH=${CNO_PATH:-$GOPATH/src/github.com/openshift/cluster-network-operator}
+SDN_PATH=${SDN_PATH:-$GOPATH/src/github.com/openshift/sdn}
+KIND_CONFIG=${KIND_CONFIG:-$HOME/kind-sdn-config.yaml}
+export KUBECONFIG=${HOME}/kube-sdn.conf
+NUM_MASTER_NODES=1
+SDN_KIND_VERBOSITY=${SDN_KIND_VERBOSITY:-5}
+
+# Default networks (same as in KIND)
+IP_FAMILY=ipv4
+CLUSTER_CIDR=${CLUSTER_CIDR:-"10.244.0.0/16"}
+SERVICE_NETWORK=${SERVICE_NETWORK:-"10.96.0.0/12"}
+HOST_PREFIX=24
+
+# Check for docker
+if ! command -v docker; then
+  echo "docker binary missing in PATH. Docker is required for this deployment"
+  exit 1
+fi
+
+# Check for kubectl
+if ! command -v kubectl; then
+  echo "kubectl binary missing in PATH. Please build/install kubectl."
+  exit 1
+fi
+
+# Ensure reachability to host via Docker network
+if ! sudo iptables -C DOCKER-USER -j ACCEPT > /dev/null 2>&1; then 
+  sudo iptables -I DOCKER-USER -j ACCEPT
+fi
+
+# create the config file
+  cat <<EOF > ${KIND_CONFIG}
+# config for 1 control plane node and 2 workers (necessary for conformance)
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+networking:
+  ipFamily: ipv4
+  disableDefaultCNI: true
+  podSubnet: ${CLUSTER_CIDR}
+  serviceSubnet: ${SERVICE_NETWORK}
+featureGates:
+  APIPriorityAndFairness: true
+runtimeConfig:
+  flowcontrol.apiserver.k8s.io/v1alpha1: true
+nodes:
+- role: control-plane
+- role: worker
+- role: worker
+EOF
+
+
+# Create KIND cluster
+kind create cluster --name sdn --image kindest/node:${K8S_VERSION} --config=${KIND_CONFIG} -v ${SDN_KIND_VERBOSITY}
+
+echo -e "\n"
+
+CNO_TEMPLATES=$CNO_PATH/manifests
+
+if [ "$BUILD_CNO" = true ]; then
+  echo "Building CNO"
+  pushd $CNO_PATH
+  CNO_IMAGE=$(BUILDCMD="docker build" ./hack/build-image.sh | grep 'Successfully tagged' | grep -Eo cluster-network-operator:.*)
+  if [ -z "$CNO_IMAGE" ]; then
+    echo "Error locating built CNO Image"
+    exit 1
+  fi
+  echo "Loading CNO image into KIND"
+  kind load docker-image $CNO_IMAGE --name sdn -v ${SDN_KIND_VERBOSITY}
+  popd
+fi
+
+if [ "$BUILD_SDN" = true ]; then
+  echo "Building SDN"
+  pushd $SDN_PATH
+  make clean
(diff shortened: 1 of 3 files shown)


juanluisvaladas:
Usage is pretty simple, just run the script `hack/sdn-kind-cno.sh` and everything should work out of the box

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,239 @@
...
+  pushd $SDN_PATH
+  make clean
+  #docker build --no-cache -f images/sdn/Dockerfile.fedora -t sdn-test .
+  popd
+  echo "Loading SDN docker image into KIND"
+  kind load docker-image sdn-test --name sdn -v ${SDN_KIND_VERBOSITY}
+fi
+
+OVS_PASSWD=$(docker run -t --rm --entrypoint 'grep' quay.io/openshift/origin-ovn-kubernetes:latest openvswitch /etc/passwd)
+NODES=$(docker ps | grep "kindest/node" | awk '{ print $1 }')
```
alexanderConstantinescu, line 95: Could this grep on "sdn"? I've used this script and I had two clusters running ("ovn" and "sdn") and this led to stuff accidentally being done on my "ovn" cluster

aojea: we should use kind api
```
NODE=$(kind get nodes --name ${KIND_CLUSTER_NAME})
```

juanluisvaladas: You're right, I've changed it for:
`NODES=$(docker ps --filter label='io.x-k8s.kind.cluster=sdn' --format '{{.Names}}')`

Which should be far more reliable.

juanluisvaladas: Oops just read you Antonio, will do that instead, sounds a bit more reliable

aojea: it´s a wrapper of your command :smile:

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,239 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Version v1.19 or higher is required
+K8S_VERSION=${K8S_VERSION:-v1.20.2}
+KIND_IMAGE=aojea/kindnode:crio-1.19.1
```
aojea, line 6: do you want to use crio?

aojea: ahh , it is commented later, ok

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,239 @@
...
+  - ${SERVICE_NETWORK}
+EOF
+
+echo "Creating CNO operator"
+for f in $(ls $CNO_TEMPLATES| grep 0000| grep -v credentials); do
+  kubectl create -f ${CNO_TEMPLATES}/$f
+done
+
+# OVS is expected to run in systemd but that not an option in kindest/noede
+# we need to deploy it in a pod.
```
aojea, line 172: question, should this be the first thing to install?

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,239 @@
...
+  echo "Sym-linking cni dirs for node $n"
+  docker exec $n rm -rf /opt/cni/bin
+  docker exec $n ln -s /var/lib/cni/bin /opt/cni/
+  docker exec $n rm -rf /etc/cni/net.d
+  docker exec $n mkdir -p /etc/cni/
+  docker exec $n ln -s /etc/kubernetes/cni/net.d /etc/cni/
+done
+
+# wait until resources are created.
+sleep 150
```
aojea, line 210: this sleep should not be needed since we have an active loop later with `kubectl wait`

juanluisvaladas: kubectl wait --for=condition=ready fails if there aren't any pods at all with that selector.
I added the sleep afterwards, anyway 150 isn't long enough so I'd like to keep it as is.

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,239 @@
...
+NODES=$(kind get nodes --name sdn)
+for n in $NODES; do
+  echo "Modifying node $n"
+  echo "Modifying os-release for Multus"
+  # required for Multus platform check
+  docker exec $n sed -i 's/ID=.*/ID=rhcos/' /etc/os-release
+  echo "Adding ovs user"
+  docker exec $n bash -c "echo $OVS_PASSWD >> /etc/passwd"
+  docker exec $n mkdir -p /var/run/crio/
+  docker exec $n ln -s /run/containerd/containerd.sock /var/run/crio/crio.sock
```
aojea, line 104: super lol

aojea:
/lgtm
I couldn´t test it but I trust you ;)

hack/sdn-kind-cno.sh
```diff
@@ -0,0 +1,238 @@
...
+
+
+
+echo "Deleting default kube-proxy"
+kubectl delete -n kube-system ds kube-proxy
+
+if [ "$BUILD_CNO" != true ]; then
+  echo "WARNING: patching CNO operator pod for OVN-K8S, deployment will no longer function if this pod is restarted"
+  CNO_POD=$(kubectl get pod -n openshift-network-operator -o jsonpath='{.items[0].metadata.name}' --field-selector status.phase=Running)
+  # kubectl -n openshift-network-operator  exec $CNO_POD sed -i '/host-run-netns/{n;s/readOnly.*/mountPropagation: Bidirectional/}' /bindata/network/ovn-kubernetes/ovnkube-node.yaml > /tmp/ovnkube-node.yaml
```
alexanderConstantinescu, line 194: Could you remove all commented lines of code?

alexanderConstantinescu (commented):
I've tested (and used) this patch to create openshift-sdn clusters, so it does work. I think I had a problem the first time I ran it w.r.t the `kind` version, which I think didn't support the feature `flowcontrol.apiserver.k8s.io/v1alpha1: true`...but it was pretty evident that I needed to upgrade the client, I believe...so no need to log something extra. 

The only change I'd like to see is removing all commented lines of code, and possibly: remove the `network-metrics-daemon` and `multus-admission-controller` from the bindata folder read by this script (if it's not too much work) since those components can't be deployed and remain stuck in `ContainerCreating` with a message saying: `selfLink was empty, can't make reference`

juanluisvaladas:
@alexanderConstantinescu can you approve it?

alexanderConstantinescu:
> @alexanderConstantinescu can you approve it?

Sorry, I can't. It's broken currently, there have been changes on master and you need to update `./hack/sdn-kind-cno.sh` with 

```diff
  docker exec sdn-control-plane chmod 666 /etc/kubernetes/kubeconfig
 
 # Create Proxy resource
-kubectl create -f https://raw.githubusercontent.com/openshift/api/e7fa4b871a25985ef0cc36c2fbd9f2cb4445dc9c/config/v1/0000_03_config-operator_01_proxy.crd.yaml
+kubectl create -f https://raw.githubusercontent.com/openshift/api/master/config/v1/0000_03_config-operator_01_proxy.crd.yaml
 
 # Create Network resource
-kubectl create -f https://raw.githubusercontent.com/openshift/api/e7fa4b871a25985ef0cc36c2fbd9f2cb4445dc9c/config/v1/0000_10_config-operator_01_network.crd.yaml
+kubectl create -f https://raw.githubusercontent.com/openshift/api/master/config/v1/0000_10_config-operator_01_network.crd.yaml
+
+# Create ingress resource
+kubectl create -f https://raw.githubusercontent.com/openshift/api/master/operator/v1/0000_50_ingress-operator_00-ingresscontroller.crd.yaml
 
 # Create cluster operator
 kubectl create -f https://raw.githubusercontent.com/openshift/machine-api-operator/050a65a2bdabcc2c2f17036de967c6bcee6d6a48/config/0000_00_cluster-version-operator_01_clusteroperator.crd.yaml

```

Could you do that? Otherwise I'll take over the work from here.

Once that is done, I'll lgtm

alexanderConstantinescu:
/lgtm

This works now. I am using it on a daily basis, thanks @juanluisvaladas !

PR #956, opened 2021-01-20
Bug 1954755: Multus configuration should allow for net-attach-defs in certain namespaces to be used across namespaces

dougbtv:
In OpenShift, we have a feature for Multus known as "namespace isolation" this allows pods to only
have annotations for additional networks that refer to the same namespace as a given pod. The expection
is the default namespace, where net-attach-defs in the default namespace can be referred to by pods
in any namespace.

Sometimes, operators need to manage their own net-attach-defs, and dumping them all into the default
namespace is not ideal. This extends the CNO to use the Multus "--global-namespaces" option, and sets
for the openshift-multus, as well as the SRIOV network operator namespace to be used.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -342,6 +342,7 @@ spec:
         - "--cni-version=0.3.1"
         - "--additional-bin-dir=/opt/multus/bin"
         - "--skip-multus-binary-copy=true"
+        - "--global-namespaces=default,openshift-multus,openshift-sriov-network-operator"
         resources:
           requests:
             cpu: 10m


dougbtv:
/assign @zshi-redhat 

Zenghui, if you don't mind, just give a check that I've got it correct for the SRIOV network operator :+1: Thanks!

zshi-redhat:
> /assign @zshi-redhat
> 
> Zenghui, if you don't mind, just give a check that I've got it correct for the SRIOV network operator  Thanks!

Doug, thanks for the quick fix! 
The SR-IOV namepsace has the exact name as we need.

zshi-redhat:
LGTM

pliurh:
@dougbtv What will happen if a net-att-def name is reused in those namespaces? How will multus choose which net-att-def CR shall be used?

zshi-redhat:
> @dougbtv What will happen if a net-att-def name is reused in those namespaces? How will multus choose which net-att-def CR shall be used?

I assume we cannot create two net-attach-defs with the same name. For net-attach-def with same name but in different namespace, user needs to specify namespace/name so it would be unqiue.

pliurh:
> For net-attach-def with same name but in different namespace, user needs to specify namespace/name so it would be unqiue.

What will happen if a user doesn't specify the namespace in this case?

zshi-redhat:
> > For net-attach-def with same name but in different namespace, user needs to specify namespace/name so it would be unqiue.
> 
> What will happen if a user doesn't specify the namespace in this case?

It will use pod namespace in case not explicitly defined..

dougbtv:
> > > For net-attach-def with same name but in different namespace, user needs to specify namespace/name so it would be unqiue.
> > 
> > 
> > What will happen if a user doesn't specify the namespace in this case?
> 
> It will use pod namespace in case not explicitly defined..

You still need to refer to the namespace in the annotation, e.g. `openshift-multus/foo` -- you can't refer to it without the namespace, otherwise it will just be local to the current namespace that the pod is within.

dougbtv:
/title Bug 1954755: Multus configuration should allow for net-attach-defs in certain namespaces to be used across namespaces

dougbtv:
/retitle Bug 1954755: Multus configuration should allow for net-attach-defs in certain namespaces to be used across namespaces

PR #958, opened 2021-01-25
Minor fixups to docs/operands.md

danwinship:
(mostly because I noticed that the indentation was wrong in the sections that were supposed to be `<pre>`)


danwinship:
docs-only fix

/override ci/prow/e2e-gcp
/override ci/prow/e2e-gcp-ovn

PR #959, opened 2021-01-26
Allow dual stack migration

danwinship:
Not sure what's happening with this wrt 4.7, and it needs a bunch of additional testing, but...

This fixes CNO to let you flip a single-stack configuration to dual-stack, and vice versa.

### DIFF
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -7,6 +7,7 @@ import (
 	operv1 "github.com/openshift/api/operator/v1"
 	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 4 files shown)


pkg/network/cluster_config.go
```diff
@@ -19,20 +20,33 @@ func ValidateClusterConfig(clusterConfig configv1.NetworkSpec) error {
 	// Check all networks for overlaps
 	pool := iputil.IPPool{}
 
-	if len(clusterConfig.ServiceNetwork) == 0 {
```
abhat, line 22: nit: we could keep this check as is to bail out early. and leave the dual-stackness+count checks for after we've parsed the `clusterConfig.ServiceNetwork`. Although for keeping the count+dual-stackness logic localized, this is fine too.

PR #963, opened 2021-01-28
Bug 1920159: Reduce CPU requests of ovs daemonset

smarterclayton:
Now that OVS has moved to the node in 4.7, the daemonset pods have
less to do. Looking at some production systems data, the OVS pod
uses about 5 millicores on 4.6 and less than 1 millicore on 4.7.
Reduce the request from 100m to 15m to retain some headroom for users
upgrading from 4.6 and running the new pod on the old nodes. In a
future release the daemonset will be fully removed.

/hold

Testing with other PRs

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -85,7 +85,7 @@ spec:
           readOnly: true
         resources:
           requests:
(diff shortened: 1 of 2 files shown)


smarterclayton:
/hold cancel

Combined tests looked ok.  Team should review and see the comments in the bug.

PR #964, opened 2021-02-01
Bug 1908916: disable ovsdb column diffs if supported

squeed:
Column diffs only work when the entire cluster is upgraded to ovsdb 2.15, so for now, ensure they're always disabled.

/cc @dcbw
/cc @numansiddique


dcbw:
/hold

We need to wait on this until we have a version of OVS 2.15 tagged into OCP; otherwise it will fail because 2.14 doesn't have the CLI option.

dcbw:
/retest
All failed due to the current ClusterBareMetalOperator prometheus problems that are being worked on.

PR #965, opened 2021-02-01
Bug 1920769: override the default node-selector for network-connectivity check

JacobTanenbaum:
If the default node selector is not overridden then during upgrade the
network never comes up because the network diagnostics assumes the
network-checker pods are on all nodes and if the default node selector
is present and not overridden they could be restricted.


abhat:
waiting for CI to pass before doing a LGTM. can you also link to a cluster-bot upgrade job here?

PR #967, opened 2021-02-02
Bug 1924137: Increase the initialDelaySeconds for liveness and readiness probes

abhat:
In cases where we run ovs in a container, sometimes ovs-vswitchd
will take a bit longer to come up as a daemon. In this case liveness
probes will start failing if the initial delay is too small and this
will prevent ovs-daemons container from coming up. This fixes the
initial delay parameter to make sure liveness and readiness probes
start after giving ovs a chance to start.

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -97,14 +97,14 @@ spec:
             command:
             - /usr/share/openvswitch/scripts/ovs-ctl
             - status
-          initialDelaySeconds: 15
+          initialDelaySeconds: 45
           periodSeconds: 5
         readinessProbe:
           exec:
             command:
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/006-ovs-node.yaml
```diff
@@ -97,14 +97,14 @@ spec:
             command:
             - /usr/share/openvswitch/scripts/ovs-ctl
             - status
-          initialDelaySeconds: 15
+          initialDelaySeconds: 45
```
markdgray, line 100: Is this a regression from a change in OVS?

abhat: It's unclear to me. but I think we saw this during a 4.4 to 4.5 upgrade, and it may be a function of how loaded the node is vs. when the ovs pods come up the first time. I think we saw this most in a scale setup.

PR #968, opened 2021-02-03
Update remaining CRDs and webhooks to v1

danwinship:
We were still using the v1beta1 versions of the APIs for the SDN and Multus CRDs and multus-admission-webhook.

CRD diffs are best viewed with `-w`

Relevant changes:
  - the CRD for `OverlappingRangeIPReservation` had claimed `preserveUnknownFields: true` before, but based on https://github.com/openshift/whereabouts-cni/blob/master/pkg/api/v1alpha1/overlappingrangeipreservation_types.go that does actually seem to be needed.
  - multus-admission-webhook now has to explicitly specify some things that were previously defaulted/guessed in v1beta1

### DIFF
--- a/bindata/network/additional-networks/crd/001-crd.yaml
+++ b/bindata/network/additional-networks/crd/001-crd.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: network-attachment-definitions.k8s.cni.cncf.io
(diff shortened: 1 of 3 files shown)


dougbtv:
Thanks Dan

/lgtm
/retest

danwinship:
/hold
bad bot

alexanderConstantinescu:
> /test e2e-gcp

Sorry about that, mistakenly opened up the wrong PR...

danwinship:
@dougbtv can you re-lgtm?

I had to change multus-admission-controller to `sideEffects: NoneOnDryRun`, which I believe is correct; ie, multus-admission-controller correctly checks if the dry-run flag was passed, and doesn't do anything in that case. Right?

PR #970, opened 2021-02-04
Bug 1912820: competing connectivitycheckcontrollers cause pod restarts during upgrades

sanchezl:
- pickup library-go#987
- remove podnetworkconnectivitycheck from CVO control


danwinship:
> `[proof: pickup openshift/library-go#987]`

I think people generally try to avoid using strings that will trigger github's automatic linking-via-commit-message, since it tends to totally spam the linked PR/issue (eg, another message every time you push an updated version of the commit)

danwinship:
lgtm
/hold
for the library-go change to actually merge

deads2k:
> * remove podnetworkconnectivitycheck from CVO control

For the future, using the CVO prevents any conditional creation, list/watch reaction speed, conditional removal, selective mutation, coordination between operators.  It is basically always better to avoid the CVO.

/lgtm

PR #975, opened 2021-02-09
SDN-1571 ipsec: Allow enablement/disablement at runtime

markdgray:
OVN-Kubernetes IPsec was originally designed to not allow
runtime enable of IPsec as per the enhancment proposal. This
commit removes this restriction.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -112,10 +112,29 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVNHybridOverlayVXLANPort"] = ""
 	}
 
+	// If IPsec is enabled for the first time, we start the daemonset. If it is
+	// disabled after that, we do not stop the daemonset but only stop IPsec.
+	//
+	// TODO: We need to do this as, by default, we maintain IPsec state on the
+	// node in order to maintain encrypted connectivity in the case of upgrades.
+	// If we only unrender the IPsec daemonset, we will be unable to cleanup
+	// the IPsec state on the node and the traffic will continue to be
+	// encrypted.
 	if c.IPsecConfig != nil {
-		data.Data["EnableIPsec"] = true
+		// IPsec is enabled
+		data.Data["OVNIPsecDaemonsetEnable"] = true
+		data.Data["OVNIPsecEnable"] = true
 	} else {
-		data.Data["EnableIPsec"] = false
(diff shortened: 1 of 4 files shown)


markdgray:
looks like these are unrelated to change
/retest

markdgray:
The failing openstack test shouldnt be caused by this change

/retest

pkg/network/ovn_kubernetes.go
```diff
@@ -222,9 +222,6 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 			errs = append(errs, errors.Errorf("cannot edit a running hybrid overlay network"))
 		}
 	}
-	if pn.IPsecConfig == nil && nn.IPsecConfig != nil {
-		errs = append(errs, errors.Errorf("cannot enable IPsec after install time"))
-	}
 	if pn.IPsecConfig != nil {
 		if !reflect.DeepEqual(pn.IPsecConfig, nn.IPsecConfig) {
```
dcbw, line 226: @markdgray wouldn't this block trigger and fail when disabling IPsec on the cluster? at least I'd that a DeepEqual() on `pn != nil` and `nn == nil` would return false, but maybe DeepEqual() holds untold surprises within itself...

squeed:
This is a good first start, and I think we can merge this as-is.

It seems like we keep the ipsec daemonset around so we can deconfigure, which is a good thing. Is there some way for the daemonset to report back that it has successfully deconfigured ipsec and can thus be deleted?

The easiest thing I can think of would be an annotation on the node. Then, in bootstrap, you check to see if every node has ipsec de-configured and can thus delete the daemonset. This can be a follow-up PR.

/approve

pkg/network/ovn_kubernetes.go
squeed: no, `next` (`nn`) would be nil and thus the conditional would skip.

markdgray: I meant to mark Dan's comment as resolved. I fixed this in the first iteration.

markdgray:
> This is a good first start, and I think we can merge this as-is.
> 
> It seems like we keep the ipsec daemonset around so we can deconfigure, which is a good thing. Is there some way for the daemonset to report back that it has successfully deconfigured ipsec and can thus be deleted?
> 
> The easiest thing I can think of would be an annotation on the node. Then, in bootstrap, you check to see if every node has ipsec de-configured and can thus delete the daemonset. This can be a follow-up PR.
> 
> /approve

Yes, I was thinking about an annotation as well but I thought it may have been messy. Please LGTM this and I will try your suggestion.

PR #976, opened 2021-02-09
cmd: stop using URL_ONLY_KUBECONFIG

squeed:
Previously, we used to parse the kubelet's kubeconfig to get the URL to the apiserver. But this is a strange coupling.

In openshift/machine-config-operator#2232 we added a special environment file that emulates the in-cluster config environment variables. So now we can just source that if it exists.

Fallback to the old mode for now, mostly out of paranoia for upgrade / downgrade.


squeed:
Test failures are uninteresting; the cluster came up just fine.

Rebased (though not needed), hopefully we go green soon.

PR #977, opened 2021-02-11
Update alerting rules so that node and pod labels get used correctly

abhat:
The node, pod, and namespace labels were not being used correctly in
our alerting rules. This commit fixes the rules to show the alert
messages correctly.


abhat:
This is not the right fix since we use templates for rendering in the CNO.

PR #979, opened 2021-02-12
Bug 1928157: Don't set ClusterOperator Version until rollout is complete

danwinship:
In the rewrite to use library-go we broke our status reporting so that now CNO reports that it has upgraded to a new version immediately after the new CNO binary starts running, rather than waiting for all of its operands to be rolled out.


danwinship:
/bugzilla refresh
/hold cancel
/assign @squeed 

works now; the remaining problem was that in some cases the library-go code would set a "ManagementStateDegraded" condition on the operator config, which would then thwart the check that assumed that the operator config would have no statuses set on a 4.6->4.7 upgrade

PR #980, opened 2021-02-12
Bug 1927244: Kuryr: Always set worker_nodes_subnets

dulek:
Turns out in UPI there are no Machine objects so we still need to
manually tell Kuryr the ID of the Neutron subnet the masters and workers
are running in. This commit makes sure that's done.

### DIFF
--- a/bindata/network/kuryr/003-config.yaml
+++ b/bindata/network/kuryr/003-config.yaml
@@ -43,6 +43,9 @@ data:
 
     endpoints_driver_octavia_provider = {{ default "default" .OctaviaProvider }}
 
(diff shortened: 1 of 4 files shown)


dulek:
/hold

This should wait for the Kuryr bit.

dulek:
/hold cancel

The Kuryr part is in.

dulek:
/assign @dcbw 

This is ready and @MaysaMacedo confirmed it solves the issue. It surely doesn't affect any other SDN than Kuryr and failures on e2e-openstack-ovn are unrelated. We'd like to get this in to make sure we can do an early 4.7.z backport.

PR #981, opened 2021-02-15
Allow from router support in CNO

abhat:
This PR adds support for 

- creating the host network namespace via bindata manifests
- watching the ingress controller objects and updating the labels on host network namespace.

### DIFF
--- a/pkg/controller/ingressconfig/ingressconfig_controller.go
+++ b/pkg/controller/ingressconfig/ingressconfig_controller.go
@@ -1,0 +1,139 @@
+package ingressconfig
+
+import (
+	"context"
+	"log"
+	"time"
+
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"github.com/openshift/cluster-network-operator/pkg/names"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+// The periodic resync interval.
+// We will re-run the reconciliation logic, even if the network configuration
+// hasn't changed.
+var ResyncPeriod = 3 * time.Minute
+
+// ManifestPaths is the path to the manifest templates
+// bad, but there's no way to pass configuration to the reconciler right now
+var ManifestPath = "./bindata"
+
+// Add creates a new ingressConfig controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager, status *statusmanager.StatusManager) error {
+	return add(mgr, newIngressConfigReconciler(mgr.GetClient()))
+}
+
+// newIngressConfigReconciler returns a new reconcile.Reconciler
+func newIngressConfigReconciler(client client.Client) *ReconcileIngressConfigs {
+	return &ReconcileIngressConfigs{client: client}
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r *ReconcileIngressConfigs) error {
+	// create a controller and register watcher for ingresscontroller resource
+	c, err := controller.New("ingress-config-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+	err = c.Watch(&source.Kind{Type: &operv1.IngressController{}}, &handler.EnqueueRequestForObject{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileIngressConfigs{}
+
+// ReconcileIngressConfigs watches for updates to ingress controller configuration
+// and sets the network policy related labels on the openshift-host-network namespace
+type ReconcileIngressConfigs struct {
+	client client.Client
+}
+
+// Reconcile sets the openshift-host-network namespaces' labels as per the
+// endpointPublishingStrategy of the `default` ingress controller object.
+// In particular, when the endpointPublishingStrategy is HostNetwork, it will
+// add the "policy-group.network.openshift.io/ingress="" label and also add
+// the "network.openshift.io/policy-group=ingress" label for legacy reasons
+// to the host network namespace.
+// When the endpointPublishingStrategy is changed to anything other than
+// HostNetwork, it reconciles and removes these labels from the host network
+// namespace.
+func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if request.Namespace != names.IngressControllerNamespace || request.Name != names.DefaultIngressControllerName {
+		return reconcile.Result{}, nil
+	}
+	log.Printf("Reconciling update to IngressController %s/%s\n", request.Namespace, request.Name)
+	ingressControllerConfig := &operv1.IngressController{TypeMeta: metav1.TypeMeta{APIVersion: operv1.GroupVersion.String(), Kind: "IngressController"}}
+	err := r.client.Get(context.TODO(), request.NamespacedName, ingressControllerConfig)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Printf("Ingress Controller configuration %s was deleted", request.NamespacedName.String())
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected, since we set
+			// the ownerReference (see https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		log.Printf("Unable to retrieve IngressController.operator.openshift.io object: %v", err)
+		return reconcile.Result{}, err
+	}
+	addLabel := ingressControllerConfig.Spec.EndpointPublishingStrategy != nil &&
+		ingressControllerConfig.Spec.EndpointPublishingStrategy.Type == operv1.HostNetworkStrategyType
+
+	err = r.updatePolicyGroupLabelOnNamespace(names.HostNetworkNamespace, addLabel)
+	if err != nil {
+		log.Printf("Error setting the host network label on namespace %s: %v", names.HostNetworkNamespace, err)
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
+}
+
+// setLabelsOnNamespace sets the labels specified on the target namespace using the client API
+func (r *ReconcileIngressConfigs) updatePolicyGroupLabelOnNamespace(targetNamespace string, add bool) error {
+	var err error
+	namespace := &corev1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: corev1.SchemeGroupVersion.String(), Kind: "Namespace"}}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: targetNamespace}, namespace)
+	if err != nil {
+		// FIXME: abhat - this needs to be handled better. Currently we have no good way to tell
+		// the difference as to whether the error is a result of
+		// a) host-network namespace should have existed, but GET failed
+		// b) host-network namespace should not have existed in the first place,
+		//    and therefore this code should ideally not even have been called.
+		// The right way to address this would be to not even spawn the ingress
+		// controller if we are running in the context of a third party plugin
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+	newNamespace := namespace.DeepCopy()
+	existingLabels := newNamespace.GetLabels()
+	if existingLabels == nil {
+		existingLabels = map[string]string{}
+	}
+	if !add {
+		delete(existingLabels, names.PolicyGroupLabelIngress)
+		delete(existingLabels, names.PolicyGroupLabelLegacy)
+	} else {
+		existingLabels[names.PolicyGroupLabelIngress] = names.PolicyGroupLabelIngressValue
+		existingLabels[names.PolicyGroupLabelLegacy] = names.PolicyGroupLabelLegacyValue
+	}
+
+	newNamespace.SetLabels(existingLabels)
+
+	return r.client.Patch(context.TODO(), newNamespace, client.MergeFrom(namespace))
+}
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -103,3 +103,26 @@ const KURYR_ADMISSION_CONTROLLER_SECRET = "kuryr-dns-admission-controller-secret
 
 // KURYR_WEB_HOOK_SECRET is the name of the secret used in the kuryr-dns-admission-controller DaemonSet
 const KURYR_WEBHOOK_SECRET = "kuryr-webhook-secret"
+
+// constants for namespace and custom resource names
+// namespace in which ingress controller objects are created
+const IngressControllerNamespace = "openshift-ingress-operator"
+
+// namespace representing host network traffic
+// this is also the namespace where to set the ingress label
+const HostNetworkNamespace = "openshift-host-network"
+
+// label for ingress policy group
+const PolicyGroupLabelIngress = "policy-group.network.openshift.io/ingress"
+
+// legacy label for ingress policy group
+const PolicyGroupLabelLegacy = "network.openshift.io/policy-group"
+
+// we use empty label values for policy groups
+const PolicyGroupLabelIngressValue = ""
+
+// value for legacy policy group label
+const PolicyGroupLabelLegacyValue = "ingress"
+
+// default ingress controller name
+const DefaultIngressControllerName = "default"
--- a/bindata/network/openshift-sdn/host-network/ns.yaml
+++ b/bindata/network/openshift-sdn/host-network/ns.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    policy-group.network.openshift.io/host-network: ""
+  annotations:
+    openshift.io/description: "Namespace for enabling network policy specification for host network traffic. Can be used to allow access to or from host network components"
+
--- a/bindata/network/openshift-sdn/host-network/resourcequota.yaml
+++ b/bindata/network/openshift-sdn/host-network/resourcequota.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: host-network-namespace-quotas
+  namespace: openshift-host-network
+spec:
+  hard:
+    limits.cpu: "0"
+    limits.memory: 0Ki
+    pods: "0"
+    count/deployments.apps: "0"
+    count/daemonsets.apps: "0"
--- a/bindata/network/ovn-kubernetes/host-network/ns.yaml
+++ b/bindata/network/ovn-kubernetes/host-network/ns.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    policy-group.network.openshift.io/host-network: ""
+  annotations:
+    openshift.io/description: "Namespace for enabling network policy specification for host network traffic. Can be used to allow access to or from host network components"
+
--- a/bindata/network/ovn-kubernetes/host-network/resourcequota.yaml
+++ b/bindata/network/ovn-kubernetes/host-network/resourcequota.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: host-network-namespace-quotas
+  namespace: openshift-host-network
+spec:
+  hard:
+    limits.cpu: "0"
+    limits.memory: 0Ki
+    pods: "0"
+    count/deployments.apps: "0"
+    count/daemonsets.apps: "0"
--- a/pkg/controller/add_networkconfig.go
+++ b/pkg/controller/add_networkconfig.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"github.com/openshift/cluster-network-operator/pkg/controller/clusterconfig"
 	configmapcainjector "github.com/openshift/cluster-network-operator/pkg/controller/configmap_ca_injector"
+	"github.com/openshift/cluster-network-operator/pkg/controller/ingressconfig"
 	"github.com/openshift/cluster-network-operator/pkg/controller/operconfig"
 	"github.com/openshift/cluster-network-operator/pkg/controller/pki"
 	"github.com/openshift/cluster-network-operator/pkg/controller/proxyconfig"
@@ -18,5 +19,6 @@ func init() {
 		clusterconfig.Add,
 		configmapcainjector.Add,
 		signer.Add,
+		ingressconfig.Add,
 	)
 }


bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,15 @@ metadata:
...
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    openshift.io/run-level: "0"
+    openshift.io/cluster-monitoring: "true"
+  annotations:
+    openshift.io/node-selector: ""
+    openshift.io/description: "OpenShift SDN Host Network Traffic"
```
squeed, line 23: nit: I would add a more detailed description here.

abhat: what would you like for this to say? 
`Namespace created for enabling network policies for host network traffic` sounds good?

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,16 @@ metadata:
   annotations:
     openshift.io/node-selector: "" #override default node selector
     openshift.io/description: "OpenShift SDN components"
+---
```
danwinship, line 12: I feel like this should probably be a separate YAML... the main `openshift-sdn` namespace is `000-ns.yaml` because it has to be created first, before anything else in that namespace. But the host network namespace is a totally separate thing. It can be in its own file, and doesn't need to be created in any particular order. (So just, like, `host-network-ns.yaml`)

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,16 @@ metadata:
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    openshift.io/run-level: "0"
```
danwinship, line 19: should not have that label; we don't want pods in this namespace to have any special privileges

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,16 @@ metadata:
...
+    openshift.io/cluster-monitoring: "true"
```
danwinship, line 20: I don't know what that label does but we probably don't need it for this

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,16 @@ metadata:
...
+    network.openshift.io/policy-group: "host-network"
+  annotations:
+    openshift.io/node-selector: ""
```
danwinship, line 23: don't need that either; there will be no pods in this namespace, so we don't need to override node restrictions on it

In fact there was discussion somewhere about maybe we should apply a quota to this namespace to ensure that pods _can't_ be created in it, and I think that's a good idea. I don't know how to do that though... (If it requires creating a second object though, then maybe we should split the host-network objects into their own bindata subdir?)

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+	client client.Client
+}
+
+// constants for namespace and custom resource names
+// namespace in which ingress controller objects are created
+const IngressControllerNamespace = "openshift-ingress-operator"
+
+// namespace representing host network traffic
+// this is also the namespace where to set the ingress label
+const HostNetworkNamespace = "openshift-host-network"
```
danwinship, line 35: probably should have this defined in `pkg/names`; we'll need to write it out in the ovn-kubernetes configmap later too

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+	newNamespace := namespace.DeepCopy()
+	existingLabels := newNamespace.GetLabels()
+	if existingLabels == nil {
+		existingLabels = map[string]string{}
+	}
+	for k, v := range labels {
+		if v == "" {
+			delete(existingLabels, k)
```
danwinship, line 107: this is kind of weird semantics given that `""` is a very commonly-used value for labels/annotations.

You could just pass a bool value into `setLabelsOnNamespace` rather than a map, and have it know what to do with that. That would actually simplify the code quite a bit anyway.

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+	var err error
+	namespace := &corev1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: corev1.SchemeGroupVersion.String(), Kind: "Namespace"}}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: targetNamespace}, namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Printf("Namespace %s was deleted", targetNamespace)
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected, since we set
+			// the ownerReference (see https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+			// Return and don't requeue
```
danwinship, line 93: That comment doesn't apply to this code; `targetNamespace` isn't the object we were asked to reconcile, it's an unrelated object that we just expected would exist.

And also, we can't just expect it to always exist anyway, since it will only exist for openshift-sdn and ovn-kubernetes.

I guess since we don't let you change network plugins at runtime, the right fix for that might be that we should only create the ingress config reconciler if the network plugin is sdn or ovn? (It might be nicer if, rather than hardcode in the code the fact that it's those two plugins, we could say "create the ingress config reconciler if the operconfig reconciler reconciles a Namespace whose `.Name == HostNetworkNamespace`, but that implies we're creating the ingress config reconciler after the operconfig reconciler starts running and I'm not sure how easy that would be...)

danwinship (commented):
(All comments on the sdn bindata apply to the ovn-kube bindata too)

squeed:
I agree with @danwinship, this should be in a separate folder.

To prevent this namespace from being used, we should create a Quota object that prevents the creation of pods, daemonsets, replicasets, deployments, and services. (see https://kubernetes.io/docs/concepts/policy/resource-quotas/)

bindata/network/openshift-sdn/000-ns.yaml
squeed: We would need this label if there are ServiceMonitor objects created in this namespace, which there won't be.

bindata/network/openshift-sdn/000-ns.yaml
```diff
@@ -9,3 +9,16 @@ metadata:
...
+    openshift.io/description: "Namespace for enabling network policy specification for host network traffic on OpenShift SDN"
```
squeed, line 24: Suggestion: "Placeholder namespace for host-network traffic in NetworkPolicy. Can be used to allow access to or from host network components."

@jboxman any thoughts?

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
+package operconfig
```
squeed, line 1: Do we even need a separate controller? Is this ever expected to change after cluster creation? Or can we just retrieve the Ingress object during the operconfig reconciliation loop and render accordingly?

Remember, we re-render every 3 minutes, so that's the longest we could go without reacting to a change.

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+// constants for namespace and custom resource names
+// namespace in which ingress controller objects are created
+const IngressControllerNamespace = "openshift-ingress-operator"
+
+// namespace representing host network traffic
+// this is also the namespace where to set the ingress label
+const HostNetworkNamespace = "openshift-host-network"
+
+// label to put on the host network namespace
+const HostNetworkLabel = "network.openshift.io/policy-group"
```
squeed, line 38: nit: `PolicyGroupLabel`

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+
+// label value to indicate use of host network ingress policy
+const HostNetworkLabelValue = "ingress"
```
squeed, line 41: PolicyGroupLabelValue

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,115 @@
...
+
+// default ingress controller name
+const DefaultIngressControllerName = "default"
+
+// Reconcile sets the openshift-host-network namespaces' labels as per the
+// endpointPublishingStrategy of the `default` ingress controller object.
+func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if request.Namespace == IngressControllerNamespace && request.Name == DefaultIngressControllerName {
```
squeed, line 49: hey @Miciah can there be more than one IngressController? Could they have different publishing strategies?

pkg/controller/operconfig/ingressconfig_controller.go
squeed: or a `*string` to indicate a deleted label.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: Users can delete and re-create the default ingress controller. I thought it was cleaner to have a separate controller than stuff all that logic into the operconfig_controller.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: I think from what @Miciah mentioned the answer is yes. But we decided that for now, we will only do the labeling of the namespace with `policy-group=ingress` for the default controller since that covers 80% of our customer use-caes and for other controllers the user may create (the remaining 20%), we can let them use the `policy-group=host-network` default label available on the namespace as a selector for network policies.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: For now, I think it's fine to create the reconciler for all plugin types and just apply the labels if the openshift-host-network namespace exists. We will of course only create the namespace (render the host-network dir) if the plugin type is ovn-k or openshift-sdn.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: done.

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,96 @@
...
+			return reconcile.Result{}, err
+		}
+		labels := map[string]*string{names.PolicyGroupLabel: nil}
+		log.Printf("Ingress Controller Config: %+v", ingressControllerConfig)
+		if ingressControllerConfig.Spec.EndpointPublishingStrategy != nil &&
+			ingressControllerConfig.Spec.EndpointPublishingStrategy.Type == operv1.HostNetworkStrategyType {
+			labelVal := names.PolicyGroupLabelIngressValue
+			labels[names.PolicyGroupLabel] = &labelVal
+		}
+		err = r.setLabelsOnNamespace(names.HostNetworkNamespace, labels)
```
squeed, line 57: Rather than hard-coding the namespace name, it would be better to _retrieve_ the host-network-equivalent namespace with a label selector. In other words, look for `network.openshift.io/policy-group = host-network`

That leaves the door open for third-party network providers to implement this functionality.

bindata/network/host-network/resourcequota.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: host-network-namespace-quotas
+  namespace: openshift-host-network
+spec:
+  hard:
```
squeed, line 7: You can also restrict pods here with
```yaml
pods: "0"
count/deployments.apps: "0"
count/daemonsets.apps: "0"
```

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,96 @@
...
+// and sets the network policy related labels on the openshift-host-network namespace
+type ReconcileIngressConfigs struct {
+	client client.Client
+}
+
+// Reconcile sets the openshift-host-network namespaces' labels as per the
+// endpointPublishingStrategy of the `default` ingress controller object.
+func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if request.Namespace == names.IngressControllerNamespace && request.Name == names.DefaultIngressControllerName {
+		log.Printf("Reconciling update to %s/%s\n", request.Namespace, request.Name)
```
squeed, line 34: nit: `Reconciling update to IngressControler ...`

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,96 @@
...
+				// the ownerReference (see https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
+				// Return and don't requeue
+				return reconcile.Result{}, nil
+			}
+			// Error reading the object - requeue the request.
+			log.Printf("Unable to retrieve IngressController.operator.openshift.io object: %v", err)
+			return reconcile.Result{}, err
+		}
+		labels := map[string]*string{names.PolicyGroupLabel: nil}
+		log.Printf("Ingress Controller Config: %+v", ingressControllerConfig)
```
squeed, line 51: This log is probably unnecessary.

squeed:
Crap; I can't believe I didn't think of this before: A namespace can't have more than one value for the same label key.

That means we can't have a namespace with both `network.openshift.io/policy-group: host-network` and `network.openshift.io/policy-group: ingress`. And the network operator will continuously fight itself, labeling the namespace back and forth.

That means we need to use separate labels and meaningless values, like **everybody else already figured out**. Bah.

Now, since `network.openshift.io/policy-group: ingress` is existing API, we'll have to keep that. And instead, label the host-network namespace with `policy-group.network.openshift.io/host-network: ""`

While we're at it, let's get Ingress labeled with `policy-group.network.openshift.io/ingress: ""`

@danwinship @Miciah.

bindata/network/host-network/resourcequota.yaml
abhat: isn't that redundant? Can someone create pods when memory and cpu allowed are `0`?

pkg/controller/operconfig/ingressconfig_controller.go
abhat: Yeah I had added it for debugging, but it also helps keep it since we don't expect too many objects to be reconciled (esp. if something were to go cuckoo here in prod environments).

abhat:
> Crap; I can't believe I didn't think of this before: A namespace can't have more than one value for the same label key.
> 
> That means we can't have a namespace with both `network.openshift.io/policy-group: host-network` and `network.openshift.io/policy-group: ingress`. And the network operator will continuously fight itself, labeling the namespace back and forth.
> 
> That means we need to use separate labels and meaningless values, like **everybody else already figured out**. Bah.
> 
> Now, since `network.openshift.io/policy-group: ingress` is existing API, we'll have to keep that. And instead, label the host-network namespace with `policy-group.network.openshift.io/host-network: ""`
> 
> While we're at it, let's get Ingress labeled with `policy-group.network.openshift.io/ingress: ""`
> 
> @danwinship @Miciah.

Good catch.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: The `setLabelsOnNamespace` API is only used to set the `policy-group:ingress` label on the namespace that OpenShift's native network plugins will use for host-network traffic. For other providers, they may choose a completely different mechanism to implement network policy for host network traffic. That also potentially means we will expose a vulnerability that anyone who can create a namespace can slap the host-network label on that namespace and have host-network traffic allowed to pods?

danwinship:
doh

> label the host-network namespace with `policy-group.network.openshift.io/host-network: ""`

can you match an empty string in a LabelSelector? (It doesn't get `omitempty`ed away?)

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -63,15 +47,16 @@ func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcil
...
 			return reconcile.Result{}, err
 		}
-		labels := map[string]string{HostNetworkLabel: ""}
+		labels := map[string]*string{names.PolicyGroupLabel: nil}
 		log.Printf("Ingress Controller Config: %+v", ingressControllerConfig)
 		if ingressControllerConfig.Spec.EndpointPublishingStrategy != nil &&
 			ingressControllerConfig.Spec.EndpointPublishingStrategy.Type == operv1.HostNetworkStrategyType {
-			labels[HostNetworkLabel] = HostNetworkLabelValue
+			labelVal := names.PolicyGroupLabelIngressValue
+			labels[names.PolicyGroupLabel] = &labelVal
```
danwinship, line 55: I mean, I guess this works, but it's still just making extra work. There are only two possibilities here: either you want to set `names.PolicyGroupLabel`, or you want to clear `names.PolicyGroupLabel`. You don't need a fully generic `setLabelsOnNamespace` function that iterates over a map of changes. You just need to set or clear `names.PolicyGroupLabel`.

squeed:
> I think there's somewhere else we're doing this too? Monitoring maybe?

Exactly right; we'll want to label [their namespace](https://github.com/openshift/cluster-monitoring-operator/blob/6ba1db4ca73eca18c28acef3b865a892fa26e719/manifests/0000_50_cluster-monitoring-operator_01-namespace.yaml) in the same manner. And update documentation.

The chances of Prometheus running in host-network or `openshift-ingress` are basically 0, so that's not an actual problem in this case.

squeed:
> can you match an empty string in a LabelSelector? (It doesn't get `omitempty`ed away?)

I believe so; the struct is such that they're not omitemptie'd away: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1081

bindata/network/host-network/resourcequota.yaml
squeed: I believe so; I believe you could create the pod, but scheduling will fail. Better to just block creation of the api object.

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -80,17 +65,13 @@ func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcil
...
 	namespace := &corev1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: corev1.SchemeGroupVersion.String(), Kind: "Namespace"}}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: targetNamespace}, namespace)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			log.Printf("Namespace %s was deleted", targetNamespace)
-			// Request object not found, could have been deleted after reconcile request.
-			// Owned objects are automatically garbage collected, since we set
-			// the ownerReference (see https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/).
-			// Return and don't requeue
+			log.Printf("Namespace %s was deleted or doesn't exist", targetNamespace)
```
danwinship, line 74: There are two possibilities here:
1. if we expected `openshift-host-namespace` to exist, and it doesn't, then that's an error
2. if we didn't expect `openshift-host-namespace` to exist, then we shouldn't even be here.

CNO _should not_ log a message about `openshift-host-namespace` not existing once every five minutes in a cluster where we didn't expect it to exist in the first place.

pkg/controller/operconfig/ingressconfig_controller.go
squeed: Labelling namespaces is a privileged operation, so that's not a risk.

The whole point is to allow third-party vendors to hook in to this system; We have a standardized label API; if they conform to that they would be automatically supported without any extra customization.

abhat: How do we handle multiple namespaces with the `host-network` label on them? Do we label all such namespaces with ingress label? I get a feeling that in the attempt to prematurely optimize, we may open up new cans of worms. If there is a legit case where a third-party vendor wants to implement network policy for host-network pods using the mechanism we provide, then we can make the change being suggested at that point.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: Makes sense. Ideally we don't even want to create a ingresscontrollerconfig reconciler if the plugin is not openshift-sdn or ovn-kubernetes. But I think at operconfig_controller creation time, we don't know. The alternative is to detect the plugin during each reconcile loop. I think that probably makes the most sense. Thoughts?

bindata/network/host-network/resourcequota.yaml
abhat: Done.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: Done.

pkg/controller/operconfig/ingressconfig_controller.go
squeed: We would certainly label multiple namespaces - that would be the exact behavior we would expect.

Basically, we're saying that "when the Ingress controller is in host-network mode, we must ensure that the host-network stub namespaces have `policy-group/ingress`. How do we know which namespaces are host-network? We look for the label-selector.

By not hard-coding the namespace name, it also means that we no longer need any code changes to openshift-sdn. Just label `default` with `policy-group/host-network`, and everything else takes care of itself.

abhat: PTAL at https://github.com/openshift/sdn/pull/262/files. The changes for openshift-sdn aren't too bad. I am still not convinced we need to look up the namespace especially when we don't have a third-party plugin that would use the mechanism we are providing here to enable network policies on host-network traffic. Not to mention that the third-party plugins still need to have code in place in order to use the labels that we are providing on whatever namespace they choose to classify host-network traffic into. So this isn't quite seamless. @danwinship wdyt?

bindata/network/host-network/ns.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    policy-group.network.openshift.io/host-network: ""
```
danwinship, line 7: Meh. This makes sense:

```
  - from:
      - namespaceSelector:
          matchLabels:
            network.openshift.io/policy-group: host-network
```

This is weird-looking:

```
  - from:
      - namespaceSelector:
          matchLabels:
            policy-group.network.openshift.io/host-network: ""
```

For other "policy-group" values, there will never be any overlap (ie, you aren't going to have monitoring and routers in the same namespace). Maybe we should just leave the other labels the way they are, and use something non-conflicting for the host-network label? Since it's kind of a hack anyway, maybe use a label that emphasizes its hackiness? Like, `network.openshift.io/policy-group-special: host-network` ?

Other than that, the PR looks good.

pkg/controller/operconfig/ingressconfig_controller.go
danwinship: > By not hard-coding the namespace name, it also means that we no longer need any code changes to openshift-sdn. Just label `default` with `policy-group/host-network`, and everything else takes care of itself.

The fact that we don't use `default` in openshift-sdn is intentional. (`default` is not supposed to be magical, and while we can't fix it at this point, I don't want to encourage people to think of it that way, especially since it leads to people thinking ovn-kubernetes is broken for not having the same behavior.)


I don't like having CNO modify Namespaces that it didn't create. If a third-party network plugin wants to create their own "host-network" namespace, they can have their own ingress config watcher too.

bindata/network/host-network/ns.yaml
abhat: Does `network.openshift.io/policy-group/host-network: ""` read better?

squeed: I would definitely prefer a consistent format, so `network.openshift.io/policy-group/host-network` reads better to me.

danwinship: no, it's the "matching an empty string" thing that I don't like. It breaks the ordinary semantics of labels.

abhat: would you feel better with `network.openshift.io/policy-group/host-network: "true"`?

danwinship: > would you feel better with `network.openshift.io/policy-group/host-network: "true"`?

we don't do that, and that's also annoying for users anyway because the quotes are unexpectedly mandatory (since `true` without quotes would be parsed as a boolean rather than a string)

I guess given the way labels work in k8s, `policy-group.network.openshift.io/host-network: ""` is correct. :slightly_frowning_face:

bindata/network/host-network/ns.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  # NOTE: this namespace is the portal for host network traffic
+  name: openshift-host-network
+  labels:
+    network.openshift.io/policy-group/host-network: "true"
```
danwinship, line 7: Is that even accepted as a label name? It's supposed to be `[<prefix>/]<name>` where neither the `<prefix>` nor the `<name>` can have a slash in it. So, `policy-group.network.openshift.io/host-network` (and `""`), as Casey originally suggested.

We could suggest that people could write the policy as:

    - namespaceSelector:
        matchExpressions:
          - key: policy-group.network.openshift.io/host-network
            operator: Exists

if they find the `""` confusing...

OK, last chance to bikeshed! Is "policy-group" still the best phrase? (Does using the `matchExpression` syntax change that?) We need "`network.openshift.io`" in the prefix, so saying "`networkpolicy`" rather than just "`policy`" feels unnecessary. Do we want to drop "`group`"? Should the ingress label use the word "router[s]" rather than "ingress" to avoid confusion with the NP meaning of "ingress"? (Probably!)

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,90 @@
...
+// ReconcileIngressConfigs watches for updates to ingress controller configuration
+// and sets the network policy related labels on the openshift-host-network namespace
+type ReconcileIngressConfigs struct {
+	client client.Client
+}
+
+// Reconcile sets the openshift-host-network namespaces' labels as per the
+// endpointPublishingStrategy of the `default` ingress controller object.
+func (r *ReconcileIngressConfigs) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	if request.Namespace == names.IngressControllerNamespace && request.Name == names.DefaultIngressControllerName {
```
danwinship, line 33: return early, don't indent the whole function

    if ! ... {
    }
    log.Printf("Reconciling...

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// setLabelsOnNamespace sets the labels specified on the target namespace using the client API
+func (r *ReconcileIngressConfigs) updatePolicyGroupLabelOnNamespace(targetNamespace string, add bool) error {
+	var err error
+	namespace := &corev1.Namespace{TypeMeta: metav1.TypeMeta{APIVersion: corev1.SchemeGroupVersion.String(), Kind: "Namespace"}}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: targetNamespace}, namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
```
danwinship, line 70: I continue to be unhappy about this; if we created the host-network namespace, and now it doesn't exist, then that's an error and should not just be ignored. (But if we _didn't_ create it, then we should not log any errors about it. Though, ideally we wouldn't even be here in that case.)

We at least need a FIXME

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+		// Error reading the object - requeue the request.
+		log.Printf("Unable to retrieve namespace object: %v", err)
+		return err
```
danwinship, line 74: don't need to log anything here. the caller will do that

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -52,11 +52,12 @@ func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager) *Re
...
-		scheme:        mgr.GetScheme(),
-		status:        status,
-		mapper:        mgr.GetRESTMapper(),
-		podReconciler: newPodReconciler(status),
+		client:                  mgr.GetClient(),
+		scheme:                  mgr.GetScheme(),
+		status:                  status,
+		mapper:                  mgr.GetRESTMapper(),
+		podReconciler:           newPodReconciler(status),
+		ingressConfigReconciler: newIngressConfigReconciler(mgr.GetClient()),
```
danwinship, line 60: The `podReconciler` is created here because the operconfig reconciler needs to call methods on it. The `ingressConfigReconciler` is independent and so should be created as its own top-level reconciler (added via `pkg/controller/add_networkconfig.go`

pkg/controller/operconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	newNamespace := namespace.DeepCopy()
+	existingLabels := newNamespace.GetLabels()
+	if existingLabels == nil {
+		existingLabels = map[string]string{}
+	}
+	if !add {
+		delete(existingLabels, names.PolicyGroupLabelIngress)
+	} else {
+		existingLabels[names.PolicyGroupLabelIngress] = names.PolicyGroupLabelIngressValue
```
danwinship, line 84: we should probably set/delete both the legacy and new label names/values

bindata/network/host-network/ns.yaml
abhat: I am going to lean toward: `network.openshift.io/policy-group/host-network: ""`, it reads better and like @squeed said provides a more consistent format.

bindata/network/host-network/ns.yaml
abhat: You are right, that syntax is not valid: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set. Rolling back to our original label format. `"policy-group.network.openshift.io/ingress": ""` seems most prudent. There are more confusing things in the pipeline when router gets deprecated by ingressv2 :). Might as well keep ingress.

bindata/network/host-network/ns.yaml
abhat: Yes, changed back to the original format @squeed had suggested.

pkg/controller/operconfig/ingressconfig_controller.go
abhat: we need additional context then if this is for `openshift-sdn` or `ovn-k`? I don't see a reason to touch existing labels here. In upgrade scenarios, we should preserve the existing labels if any on the default namespace and this doesn't touch that namespace at all.

pkg/controller/operconfig/operconfig_controller.go
abhat: Moved ingress config controller at the top-level.

pkg/controller/operconfig/ingressconfig_controller.go
danwinship: no, I mean, we should set both `network.openshift.io/policy-group: ingress` and `policy-group.network.openshift.io/ingress: ""` on the namespace, so that it works both with new NetworkPolicies targeting the new label name and with existing NetworkPolicies targeting the old label name.

abhat:
Hitting quota issues.
/test e2e-gcp-ovn

abhat:
@squeed @danwinship are you ok with this PR now? Need to over-ride the metal-ipi-ovn-ipv6 job, but I feel like this is ready to go in otherwise.

pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -0,0 +1,132 @@
...
+var _ reconcile.Reconciler = &ReconcileIngressConfigs{}
+
+// ReconcileIngressConfigs watches for updates to ingress controller configuration
+// and sets the network policy related labels on the openshift-host-network namespace
+type ReconcileIngressConfigs struct {
+	client client.Client
+}
+
+// Reconcile sets the openshift-host-network namespaces' labels as per the
+// endpointPublishingStrategy of the `default` ingress controller object.
```
squeed, line 66: Can you add a bit more description here? What labels are added in response to what settings.

pkg/network/openshift_sdn.go
```diff
@@ -81,6 +81,7 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	}
 
 	objs = append(objs, manifests...)
+
```
squeed, line 84: nit: unnecessary

pkg/names/names.go
```diff
@@ -103,3 +103,26 @@ const KURYR_ADMISSION_CONTROLLER_SECRET = "kuryr-dns-admission-controller-secret
...
+
+// namespace representing host network traffic
+// this is also the namespace where to set the ingress label
+const HostNetworkNamespace = "openshift-host-network"
+
+// label for ingress policy group
+const PolicyGroupLabelIngress = "policy-group.network.openshift.io/ingress"
+
+// legacy label for ingress policy group
+const PolicyGroupLabelIngressLegacy = "network.openshift.io/policy-group"
```
squeed, line 119: nit: PolicyGroupLabelLegacy

pkg/names/names.go
```diff
@@ -103,3 +103,26 @@ const KURYR_ADMISSION_CONTROLLER_SECRET = "kuryr-dns-admission-controller-secret
...
+
+// we use empty label values for policy groups
+const PolicyGroupLabelIngressValue = ""
```
squeed, line 122: nit: don't need to const this.

squeed:
A few nits, otherwise seems basically fine.

pkg/names/names.go
abhat: just curious why not?

squeed: it's an empty string :). And the label value will never be anything else.

PR #984, opened 2021-02-17
Enable golangci-lint; fix up lint issues

squeed:
This enables golangci-lint, which finds a larger number of code smells while having very few false positives.

Fix up or ignore all found issues.

Most changes are mechanical, but some include handling errors we previously ignored. So this is, technically, a code change.


pkg/controller/proxyconfig/controller.go
```diff
@@ -232,6 +228,7 @@ func (r *ReconcileProxyConfig) Reconcile(request reconcile.Request) (reconcile.R
 		}
 
 		// Create a configmap containing the merged proxy.trustedCA/system bundles.
+		//nolint:staticcheck // TODO:danehans maybe remove this line?
```
squeed, line 231: @danehans I think there is a bug here; this block of code seems to be useless.

pkg/network/ovn_kubernetes.go
```diff
@@ -92,14 +92,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 
-	var svcpools string
-	for _, net := range conf.ServiceNetwork {
-		if len(svcpools) != 0 {
-			svcpools += ","
-		}
-		svcpools += fmt.Sprintf("%s", net)
-	}
-	data.Data["OVN_service_cidr"] = svcpools
+	data.Data["OVN_service_cidr"] = strings.Join(conf.ServiceNetwork, ",")
```
dulek, line 95: Okay, that substitution is impressive.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -441,6 +442,9 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 
 	log.Print("Ensuring pods security group")
 	podSgId, err := ensureOpenStackSg(client, generateName("kuryr-pods-security-group", clusterID), tag)
+	if err != nil {
```
dulek, line 445: Makes sense.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -577,6 +581,9 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	log.Print("Creating OpenShift API loadbalancer pool members")
 	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port)$", clusterID))
 	portList, err := listOpenStackPortsMatchingPattern(client, tag, r)
+	if err != nil {
```
dulek, line 584: And here as well.

dulek (commented):
Kuryr part looks good, thank you!

squeed:
Update: it turns out that GitHub actions are disabled, so we'll have to remove that bit.

pkg/operator/operator.go
```diff
@@ -92,10 +92,17 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 
 	// Start controllers
 	klog.Info("Starting controller-manager")
-	go o.manager.Start(ctx.Done())
+	go func() {
+		err := o.manager.Start(ctx.Done())
+		if err != nil {
+			klog.Fatalf("Failed to start controller-runtime manager: %v", err)
```
fedepaol, line 98: Would it make sense to close the ctx.Done() channel so the operator reboots (assuming this is the entry point) ?

pkg/operator/operator.go
```diff
@@ -92,10 +92,17 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+		}
+	}()
 	go logLevelController.Run(ctx, 1)
 	go managementStateController.Run(ctx, 1)
-	connectivitycheck.Start(ctx, o.ccfg.KubeConfig)
+	if err := connectivitycheck.Start(ctx, o.ccfg.KubeConfig); err != nil {
+		klog.Errorf("Failed to start connectivitycheck controller: %v", err)
```
fedepaol, line 104: Would it make sense to return here? Otherwise we just log and leave the operator up half functioning

pkg/operator/operator.go
squeed: I tried to do this, but ctx.Done() is not closable (it's a <-chan).

pkg/operator/operator.go
squeed: the connectivity check controller is not critical functionality; if it fails to start up (which is very unlikely here) we still want to try and continue to progress the cluster.

pkg/operator/operator.go
fedepaol: Ah, then I guess it's fine as it is.
An alternative would be to create a context.WithCancel starting from ctx, not sure it's worth the effort

pkg/operator/operator.go
fedepaol: Understood, thanks

abhat:
Thanks for fixing these linter errors.
/lgtm

PR #985, opened 2021-02-18
Bug 1924425: Update protobuf version to address CVE-2021-3121

abhat:
This commit udpates the protobuf version used in CNO to 1.3.2
to address the CVE-2021-3121


abhat:
Upstream k/k fix is: kubernetes/kubernetes#98477

PR #986, opened 2021-02-19
Bug 1960751: DisableMultiNetwork should only disable reading annotations

dougbtv:
Instead of previously, where it would disable deploying Multus at all.

Still a WIP, I need to run through some testing with this along with the companion Multus change @ https://github.com/openshift/multus-cni/pull/87


dougbtv:
/retitle Bug 1960751: Adds option to not read annotations.
/hold

dougbtv:
/retitle Bug 1960751: DisableMultiNetwork should only disable reading annotations

PR #988, opened 2021-02-22
config: Set enable-profiling true by default and allow `enable-profiling` as a proxy argument

smarterclayton:
Turn it on default (protected by RBAC).  Allows a user to configure this post deployment.

Verified we RBAC protect this.  Because it's rbac protected (like kubelet and apiserver) on by default is acceptable.

```
curl https://localhost:9101/debug/pprof -k -H "Authorization: bearer $( oc sa get-token -n openshift-sdn default )"
Handling connection for 9101
Forbidden (user=system:serviceaccount:openshift-sdn:default, verb=get, resource=, subresource=)
```

### DIFF
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -52,6 +52,7 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 		"healthz-port":            {"10256"},
 		"proxy-mode":              {"iptables"},
 		"iptables-masquerade-bit": {"0"},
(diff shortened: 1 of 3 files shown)


squeed:
One minor gotcha: I don't think we bind to `localhost` for our metrics port. This means that bare-metal installations will have this accessible.

smarterclayton:
Which metrics?  The core one is auth protected.  Kube proxy / the health check are not deliberately but contain no critical info

smarterclayton:
@squeed did that answer your comment or did i misinterpret the question?

PR #989, opened 2021-02-22
Removing old kuryr-kubernetes CRDs

scavnicka:
When moving from annotation to CRD, old CRDs was left to allow user to upgrade from 4.5 to 4.6 version. Now we are going to upgrade to ocp 4.8, so everyone should already use CRD instead of annotations and we can remove the old CRDs, which was used when we have worked with annotations.


dulek:
/lgtm

Seems good to me, we should probably test this a bit on a live setup.

dulek:
Seems like you need to update the tests too: https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/kuryr_test.go#L70. I'd just update it to check for all the CRDs that still exist.

MaysaMacedo:
/hold

Holding it from merging, just so that we can test it.

MaysaMacedo:
Could you squash all the commits into the first commit  "Removing old kuryr-kubernetes CRDs"?

MaysaMacedo:
It worked fine.

danwinship:
"Needs bugzilla/valid-bug label." (ie, needs a bz now if you want it to merge in 4.8)

dulek:
/hold

I'm not sure if this is worth a BZ.

dulek:
/hold cancel

Well, we can merge it without the BZ now. ;)

PR #990, opened 2021-02-22
Bug 1932114: kuryr: Detect Nodes Network MTU

MaysaMacedo:
same MTU as the host.


MaysaMacedo:
hm, not sure how to proceed here. This is already on 4.7 https://github.com/openshift/kuryr-kubernetes/commit/17f99a71782caec0cbf73cb268fc55ef0d2dba06

PR #991, opened 2021-02-23
ovnkube: disable ovsdb-server column-diffs if ovn-ctl supports the option

dcbw:
Detect whether ovn-ctl supports the disable-column-diffs option, and set
it to "yes" if supported to disable column diffs for OVS 2.15 and later.


dcbw:
gcp failures are capacity issues (too many in-flight API calls)

dcbw:
@squeed

dcbw:
Closed in favor of https://github.com/openshift/cluster-network-operator/pull/964

PR #993, opened 2021-02-25
[SDN-1364] Update CNO for OVN acl audit logging

astoycos:
Changes for [SDN-1364](https://issues.redhat.com/browse/SDN-1364)

See [Enhancement](https://github.com/openshift/enhancements/pull/617) for
details on the changes

1. Allows `ovn-kubernetes.go` to digest the new API changes 

2. Adds a new sidecar container to the `ocnkube-node` daemonset named acl-audit-logging
It does the following 
        - Pipes the audit logs(which are persisted to the node at `var/log/ovn/` ) to stdout so they can be consumed by the logging 
          stack if desired 
        - Handles log rotation since COREOS doesn't include logrotate modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,24 @@ spec:
       ipsecConfig: {}
 ```
 
+#### Configuring Network Policy audit logging with OVNKubernetes 
+
+OVNKubernetes supports audit logging of network policy traffic events.  Add the following to the `spec:` section of the operator config: 
+
+```yaml
+spec:
+  defaultNetwork:
+    type: OVNKubernetes
+    ovnKubernetesConfig: 
+      policyAuditingConfig:
+        maxFileSize: 1
+        rateLimit: 5
+        destination: libc
+        syslogFacility: local0
+```
+
+To understand more about each field, and to see the default values check out the [Openshift api definition](https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L397)
+
 ### Configuring Kuryr-Kubernetes
 Kuryr-Kubernetes is a CNI plugin that uses OpenStack Neutron to network OpenShift Pods, and OpenStack Octavia to create load balancers for Services. In general it is useful when OpenShift is running on an OpenStack cluster, as you can use the same SDN (OpenStack Neutron) to provide networking for both the VMs OpenShift is running on, and the Pods created by OpenShift. In such case avoidance of double encapsulation gives you two advantages: improved performace (in terms of both latency and throughput) and lower complexity of the networking architecture.
 
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -48,12 +48,18 @@ spec:
             set -o allexport
             source "/env/${K8S_NODE}"
             set +o allexport
-          fi
+          fi  
+          
           echo "$(date -Iseconds) - starting ovn-controller"
           exec ovn-controller unix:/var/run/openvswitch/db.sock -vfile:off \
             --no-chdir --pidfile=/var/run/ovn/ovn-controller.pid \
+            --syslog-method="{{.OVNPolicyAuditDestination}}" \
+            --log-file=/var/log/ovn/acl-audit-log.log \
+            -vFACILITY:"{{.OVNPolicyAuditSyslogFacility}}" \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            -vconsole:"${OVN_LOG_LEVEL}"
+            -vconsole:"${OVN_LOG_LEVEL}" -vconsole:"acl_log:off" \
+            -vsyslog:"acl_log:info" \
+            -vfile:"acl_log:info"
         securityContext:
           privileged: true
         env:
@@ -80,12 +86,62 @@ spec:
           name: ovn-cert
         - mountPath: /ovn-ca
           name: ovn-ca
+        - mountPath: /var/log/ovn
+          name: node-log
+        - mountPath: /dev/log
+          name: log-socket
         terminationMessagePolicy: FallbackToLogsOnError
         resources:
           requests:
             cpu: 10m
             memory: 300Mi
+      - name: ovn-acl-logging
+        image: "{{.OvnImage}}"
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -euo pipefail
 
+          # Rotate audit log files when then get to max size (in bytes)
+          MAXFILESIZE=$(( "{{.OVNPolicyAuditMaxFileSize}}"*1000000 )) 
+          LOGFILE=/var/log/ovn/acl-audit-log.log
+          CONTROLLERPID=$(cat /run/ovn/ovn-controller.pid)
+
+          # Redirect err to null so no messages are shown upon rotation
+          tail -F ${LOGFILE} 2> /dev/null &
+
+          while true
+          do
+            # Make sure ovn-controller's logfile exists, and get current size in bytes 
+            if [ -f "$LOGFILE" ]; then 
+              file_size=`du -b ${LOGFILE} | tr -s '\t' ' ' | cut -d' ' -f1`
+            else 
+              ovs-appctl -t /var/run/ovn/ovn-controller.${CONTROLLERPID}.ctl vlog/reopen
+              file_size=`du -b ${LOGFILE} | tr -s '\t' ' ' | cut -d' ' -f1`
+            fi 
+            
+            if [ $file_size -gt $MAXFILESIZE ];then
+              echo "Rotating OVN ACL Log File"
+              timestamp=`date '+%Y-%m-%dT%H-%M-%S'`
+              mv ${LOGFILE} /var/log/ovn/acl-audit-log.$timestamp.log
+              ovs-appctl -t /run/ovn/ovn-controller.${CONTROLLERPID}.ctl vlog/reopen
+              CONTROLLERPID=$(cat /run/ovn/ovn-controller.pid)
+            fi
+
+            # sleep for 30 seconds to avoid wasting CPU 
+            sleep 30 
+          done
+        resources:
+          requests:
+            cpu: 10m
+            memory: 20Mi
+        terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /var/log/ovn
+          name: node-log
+        - mountPath: /run/ovn/
+          name: run-ovn
       - name: kube-rbac-proxy
         image: {{.KubeRBACProxyImage}}
         command:
@@ -138,7 +194,6 @@ spec:
         - name: ovn-node-metrics-cert
           mountPath: /etc/pki/tls/metrics-cert
           readOnly: True
-
       # ovnkube-node: does node-level bookkeeping and configuration
       - name: ovnkube-node
         image: "{{.OvnImage}}"
@@ -329,6 +384,13 @@ spec:
       - name: run-ovn
         hostPath:
           path: /var/run/ovn
+      # Used for placement of ACL audit logs 
+      - name: node-log
+        hostPath: 
+          path: /var/log/ovn
+      - name: log-socket
+        hostPath: 
+          path: /dev/log
       # For CNI server
       - name: host-run-ovn-kubernetes
         hostPath:
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -84,6 +84,10 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["NetFlowCollectors"] = ""
 	data.Data["SFlowCollectors"] = ""
 	data.Data["IPFIXCollectors"] = ""
+	data.Data["OVNPolicyAuditRateLimit"] = c.PolicyAuditConfig.RateLimit
+	data.Data["OVNPolicyAuditMaxFileSize"] = c.PolicyAuditConfig.MaxFileSize
+	data.Data["OVNPolicyAuditDestination"] = c.PolicyAuditConfig.Destination
+	data.Data["OVNPolicyAuditSyslogFacility"] = c.PolicyAuditConfig.SyslogFacility
 
 	var ippools string
 	for _, net := range conf.ClusterNetwork {
@@ -240,7 +244,8 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 	return out
 }
 
-// isOVNKubernetesChangeSafe currently returns an error if any changes are made.
+// isOVNKubernetesChangeSafe currently returns an error if any changes to immutable
+// fields are made.
 // In the future, we may support rolling out MTU or other alterations.
 func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 	pn := prev.DefaultNetwork.OVNKubernetesConfig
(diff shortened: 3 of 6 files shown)


astoycos:
/assign @russellb 

He're you can see the work I did to allow for viewing only audit logs via the sidecar container `acl-audit-logging` and also the log rotation mechanism I built in

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -78,12 +83,53 @@ spec:
...
+          #Get size in bytes**
+          file_size=`du -b /var/log/ovn/acl-audit-log.log | tr -s '\t' ' ' | cut -d' ' -f1`
+          if [ $file_size -gt $MaxFileSize ];then
+            echo "Rotating OVN ACL Log File"
+            timestamp=`date '+%Y-%m-%dT%H-%M-%S'`
+            mv /var/log/ovn/acl-audit-log.log /var/log/ovn/acl-audit-log.$timestamp.log
+            ovs-appctl -t /var/run/ovn/ovn-controller.$(cat /var/run/ovn/ovn-controller.pid).ctl vlog/reopen
+          fi
+
+          done
```
russellb, line 124: I'm not sure this is needed.  When I was looking at the OVN package last week, it looks like it already automatically configures logrotate to run `vlog/reopen` when rotation occurs.

Here's the upstream link to the logrotate config, but I saw it installed on my OpenShift cluster as well.

https://github.com/ovn-org/ovn/blob/master/rhel/etc_logrotate.d_ovn

or did you determine a reason why this may not work?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -46,12 +46,17 @@ spec:
...
           echo "$(date -Iseconds) - starting ovn-controller"
           exec ovn-controller unix:/var/run/openvswitch/db.sock -vfile:off \
             --no-chdir --pidfile=/var/run/ovn/ovn-controller.pid \
+            --syslog-method="{{.OVNPolicyAuditDestination}}" \
+            --log-file=/var/log/ovn/acl-audit-log.log \
+            -vFACILITY:"{{.OVNPolicyAuditSyslogFacility}}" \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            -vconsole:"${OVN_LOG_LEVEL}"
+            -vconsole:"${OVN_LOG_LEVEL}" -vconsole:"acl_log:off" \
+            -vfile:"acl_log:${OVN_LOG_LEVEL}" -vsyslog:"acl_log:${OVN_LOG_LEVEL}"
```
russellb, line 59: I was surprised to see both the file and syslog configs here.  why both?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -303,6 +349,16 @@ spec:
       - name: run-ovn
         hostPath:
           path: /var/run/ovn
+      # Used for placement of ACL audit logs 
+      - name: node-log
+        hostPath: 
+          path: /var/log/ovn
+      - name: log-socket
```
squeed, line 356: These should be the same - /dev/log and the journald /dev/log...

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -78,12 +83,53 @@ spec:
...
+          if [ $file_size -gt $MaxFileSize ];then
+            echo "Rotating OVN ACL Log File"
+            timestamp=`date '+%Y-%m-%dT%H-%M-%S'`
+            mv /var/log/ovn/acl-audit-log.log /var/log/ovn/acl-audit-log.$timestamp.log
+            ovs-appctl -t /var/run/ovn/ovn-controller.$(cat /var/run/ovn/ovn-controller.pid).ctl vlog/reopen
+          fi
+
+          done
+        securityContext:
+          privileged: true
```
squeed, line 126: Does this need to be privileged? I don't think so.

pkg/network/ovn_kubernetes.go
```diff
@@ -260,6 +264,24 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
...
+
+	if sc.PolicyAuditConfig.RateLimit == nil {
+		var ratelimit uint32 = uint32(20)
+		sc.PolicyAuditConfig.RateLimit = &ratelimit
+	}
+	if sc.PolicyAuditConfig.MaxFileSize == nil {
+		var maxfilesize uint32 = uint32(50000000)
+		sc.PolicyAuditConfig.MaxFileSize = &maxfilesize
+	}
+	if sc.PolicyAuditConfig.Destination == "" {
```
squeed, line 276: You're going to want to validate that this is a destination OVN understands; you shouldn't just pass this through directly to the shell script.  (It's also a shell-script unescape exploit).

squeed: e.g. setting a facility of `$(reboot)`...

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: Per our weekly meeting I was informed that the built in log rotate modules would not work out of the box with coreos images unfortunately.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: I now explicitly configure all of the ovs logging modules here 
1. Before the `ovnkube-node` container dumped all controller logs now it dumps all logs except the ACL audit logs to stdout 
2. The file output only contains ACL audit logging which is stored on node and also tailed by the `acl-audit-logging` sidecar container 
3. The feature can be configured to also send ACL audit logs to the underlying Host's journald process or over a udp socket which is when the syslog module is configured here.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: for some reason I had to had both for the host's `journald` process to pickup the logs, but I will experiment some more and try to just have one socket here.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: ACK I agree

pkg/network/ovn_kubernetes.go
astoycos: Yep I added validation on the API side of things, let me know if that's alright https://github.com/openshift/api/pull/853/files#diff-2e98c189fdf696589c6d28124557b9cf96a329e82f790b4e61209351cc68648dR382

astoycos:
API changes have landed 

/retest

README.md
```diff
@@ -267,6 +267,24 @@ spec:
...
+```yaml
+spec:
+ defaultNetwork:
+   type: OVNKubernetes
+   ovnKubernetesConfig: 
+     policyAuditingConfig:
+       maxFileSize: 1
+       rateLimit: 5
+       destination: libc
+	syslogFacility: local0
```
squeed, line 283: nit: indentation seems weird here

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -48,12 +48,17 @@ spec:
             set -o allexport
             source "/env/${K8S_NODE}"
             set +o allexport
-          fi
+          fi  
+          
           echo "$(date -Iseconds) - starting ovn-controller"
           exec ovn-controller unix:/var/run/openvswitch/db.sock -vfile:off \
             --no-chdir --pidfile=/var/run/ovn/ovn-controller.pid \
+            --syslog-method="{{.OVNPolicyAuditDestination}}" \
```
squeed, line 56: shouldn't these flags be passed only if audit logging is enabled?

README.md
astoycos: ACK I'll fix

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: So based on the upstream work this feature is always enabled internally in ovn-k.... However it will only be activated if the user passes the required annotations on a namespace's yaml.

If the user does not manually configure the audit-logging-config all of the values will default to the values I have specified (in the case of the `--syslog-method` flag it will just default to the OVN-K default of `null`)

Therefore I think maintaining a single path with defined defaults is easier here.

pkg/network/ovn_kubernetes.go
```diff
@@ -239,6 +243,11 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 			errs = append(errs, errors.Errorf("cannot edit IPsec configuration at runtime"))
 		}
 	}
+	if pn.PolicyAuditConfig != nil && nn.PolicyAuditConfig != nil {
+		if !reflect.DeepEqual(pn.IPsecConfig, nn.IPsecConfig) {
```
squeed, line 247: why not? This seems like a perfectly reasonable thing to be able to change?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -48,12 +48,17 @@ spec:
...
+            --log-file=/var/log/ovn/acl-audit-log.log \
+            -vFACILITY:"{{.OVNPolicyAuditSyslogFacility}}" \
             -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
-            -vconsole:"${OVN_LOG_LEVEL}"
+            -vconsole:"${OVN_LOG_LEVEL}" -vconsole:"acl_log:off" \
+            -vfile:"acl_log:${OVN_LOG_LEVEL}" -vsyslog:"acl_log:${OVN_LOG_LEVEL}"
```
squeed, line 61: I don't think this is right - OVN_LOG_LEVEL is user-defined, and could be disabled, theoretically. Shouldn't this be constant?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: Is there any way to turn off audit logging? What happens in that case?

This all is confusing. If audit logging is disabled, I would expect to see no configuration for it.

pkg/network/ovn_kubernetes.go
```diff
@@ -82,6 +82,10 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["LISTEN_DUAL_STACK"] = listenDualStack(bootstrapResult.OVN.MasterIPs[0])
 	data.Data["OVN_CERT_CN"] = OVN_CERT_CN
 	data.Data["OVN_NORTHD_PROBE_INTERVAL"] = os.Getenv("OVN_NORTHD_PROBE_INTERVAL")
+	data.Data["OVNPolicyAuditRateLimit"] = c.PolicyAuditConfig.RateLimit
```
squeed, line 85: This is unused.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: aha, I see now. audit logging is always enabled.

pkg/network/ovn_kubernetes.go
astoycos: I agree! However this is what I saw in the Docs that made me make it changesafe 

`You can only change the configuration for your default CNI network provider during cluster installation.` 

--> https://docs.openshift.com/container-platform/4.7/networking/cluster-network-operator.html#nw-operator-configuration-parameters-for-ovn-sdn_cluster-network-operator

astoycos: I can do whatever you want here though I think it should be configurable at runtime as well

pkg/network/ovn_kubernetes.go
astoycos: ACK Missed my one line change to `ovn-kube master` somehow, thanks for the catch!

bindata/network/ovn-kubernetes/ovnkube-node.yaml
astoycos: So I think for the `vfile` flag I will fix the level to `info` However the syslog flag I will keep fully configurable since it is off by default anyways

astoycos:
`ci/prow/e2e-metal-ipi-ovn-ipv6` Is broken and has been failing elsewhere consistently.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: So, OVN_LOG_LEVEL is the user-defined level, set by configmap, for what should show up in stdout -> `kubectl get logs`. It has nothing to do with ACL logging.

ACL logging should be at whatever causes the logs to show up. It can probably just be fixed to `info`

pkg/network/ovn_kubernetes.go
squeed: Yes, make it reconfigurable (and update the docs, please :).

pkg/network/ovn_kubernetes.go
squeed: That's probably fine, then.

pkg/network/ovn_kubernetes.go
```diff
@@ -275,6 +284,28 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
...
+	if sc.PolicyAuditConfig.RateLimit == nil {
+		var ratelimit uint32 = uint32(20)
+		sc.PolicyAuditConfig.RateLimit = &ratelimit
+	}
+	if sc.PolicyAuditConfig.MaxFileSize == nil {
+		var maxfilesize uint32 = uint32(50)
+		sc.PolicyAuditConfig.MaxFileSize = &maxfilesize
+	}
+	if sc.PolicyAuditConfig.Destination == "" {
+		var destination string = string("null")
```
squeed, line 301: no need for these two `string()` lines... can just do `sc.PolicyAuditConfig.Destination = "null"` (and a few lines down)

astoycos:
This should be good to do @squeed I will 
1. Add an e2e Test in ocp/origin
2. Fix docs for changesafe following the merge

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -80,12 +86,59 @@ spec:
...
+              file_size=`du -b ${LOGFILE} | tr -s '\t' ' ' | cut -d' ' -f1`
+            fi 
+            
+            if [ $file_size -gt $MAXFILESIZE ];then
+              echo "Rotating OVN ACL Log File"
+              timestamp=`date '+%Y-%m-%dT%H-%M-%S'`
+              mv ${LOGFILE} /var/log/ovn/acl-audit-log.$timestamp.log
+              ovs-appctl -t /run/ovn/ovn-controller.${CONTROLLERPID}.ctl vlog/reopen
+              CONTROLLERPID=$(cat /run/ovn/ovn-controller.pid)
+            fi
```
squeed, line 130: Sorry, one silly thing I just realized: this should sleep, even if for 15 seconds, so it's not spinning and using up the CPU.

squeed:
One silly little thing (that I should have noticed weeks ago) and this should be good to go.

astoycos:
ACK all good thanks! @squeed

astoycos:
All the failing tests are not required, and seem to be failing across all PR's pretty consistently

PR #999, opened 2021-03-03
Bug 1933269: Include LB members for Machines created on day-2 operation

MaysaMacedo:
The ports created for master VMs created on day-2 operation
have a different name pattern than the ones created on a
regular installation making it impossible for the CNO to
identify it and create a new member. This commit fixes the
issue by takeing into account the extra name pattern.


dulek:
I'm not 100% sure this is how we should handle it - shouldn't the fix in 4.8 be different than in 4.7…? Or do we handle 4.8 fix in another patch?

MaysaMacedo:
> I'm not 100% sure this is how we should handle it - shouldn't the fix in 4.8 be different than in 4.7…? Or do we handle 4.8 fix in another patch?
Isn't it required that the fix is first merged on 4.8 and only then backport to 4.7?
or your idea was to fix by removing the lb in 4.8 and on 4.7 propose this fix here?

MaysaMacedo:
My thinking was that the lb improvement would be included in another patch.

dulek:
/retest
/lgtm

Okay, makes sense to do it like this.

PR #1002, opened 2021-03-05
Bug 1935473: Include LB members for Machines created on day-2 operation

MaysaMacedo:
The ports created for master VMs created on day-2 operation
have a different name pattern than the ones created on a
regular installation making it impossible for the CNO to
identify it and create a new member. This commit fixes the
issue by takeing into account the extra name pattern.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -575,8 +575,12 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	// the API. With healthchecks enabled for the pool we'll get masters added automatically
 	// when they're up and ready.
 	log.Print("Creating OpenShift API loadbalancer pool members")
-	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port)$", clusterID))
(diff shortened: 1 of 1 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -575,8 +575,12 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	// the API. With healthchecks enabled for the pool we'll get masters added automatically
 	// when they're up and ready.
 	log.Print("Creating OpenShift API loadbalancer pool members")
-	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port)$", clusterID))
+	r, _ := regexp.Compile(fmt.Sprintf("^%s-(master-port-[0-9]+|bootstrap-port+|master-[0-9])$", clusterID))
```
crawford, line 578: This expression looks incorrect. I believe it should instead read: `^%s-(master-port-[0-9]+|bootstrap-port|master-[0-9]+)$`. As it's written, "bootstrap-portttttttttttttttttt" will match and "master-10" will not (though "master-port-10" will).

MaysaMacedo: Thanks again for spotting this issue. Withh CI cooperation the fix on main branch can be merged by tomorrow and be included in this backport.

PR #1006, opened 2021-03-08
Bug 1936710: OSD-6600 network-metrics missing priorityClass

dofinn:
Backing porting from master.


wking:
/bugzilla refresh

Needs the 4.7 #1007 to land and be verified first.

PR #1007, opened 2021-03-08
Bug 1936719: OSD-6600 network-metrics missing priorityClass

dofinn:
Back porting from master


wking:
Backporting from #994, which landed without a tracking bug.

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.20

PR #1008, opened 2021-03-09
WIP - openstack: Force OVN requirements

EmilienM:
In order to avoid using Amphora driver to create HTTPS LB pools,
let's switch to TCP protocol using SOURCE_IP LB method which is
supported by the OVN driver:

  https://docs.openstack.org/ovn-octavia-provider/latest/admin/driver.html#limitations-of-the-ovn-provider-driver

By avoiding the Amphora:

- Less cloud resources are consumed in the cloud
- Better performances for LB in general
- Traffic stays in the same Availability Zone (Octavia doesn't support
  AZs).

This changes:
- Switch LB pools to use TCP protocol instead of HTTPS and SOURCE_IP
  method.
- Stop creating Health Monitors, not supported in OVN
- Use TCP for LB Listeners instead of HTTPS

Signed-off-by: Emilien Macchi <emilien@redhat.com>


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -559,13 +558,6 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
 	}
 	log.Printf("OpenShift API loadbalancer pool %s present", poolId)
 
-	log.Print("Creating OpenShift API loadbalancer health monitor")
-	monitorId, err := ensureOpenStackLbMonitor(lbClient, "kuryr-api-loadbalancer-monitor", poolId)
```
EmilienM, line 563: I have concerns myself by removing that function and it makes me wonder if we should keep this around in case OVN provider isn't detected as available?
There is an assumption that OVN backend is enabled since the recent versions of OSP, but I don't think it makes it a safe assumption either a backward compatible change.

EmilienM:
cc @MaysaMacedo @luis5tb @dulek

EmilienM:
I haven't figured yet why ```ensureOpenStackLb``` creates an Amphora by default, while OVN is enabled, I wonder if the config used isn't compatible with OVN?

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -51,7 +51,6 @@ const (
 	machinesNamespace     = "openshift-machine-api"
 	// NOTE(ltomasbo): Amphora driver supports came on 2.11, but ovn-octavia only supports it after 2.13
 	MinOctaviaVersionWithMultipleListeners = "v2.13"
-	MinOctaviaVersionWithHTTPSMonitors     = "v2.10"
```
luis5tb, line 54: this is still needed for envs that do not support ovn-octavia, right?

luis5tb:
This needs to be updated to still support environments without ovn-octavia driver. So it needs to make use of the functions to discover if ovn-octavia is supported and enabled, and if that is the case, use it, otherwise it should still create the amphora lb. In addition, perhaps it should also handle the upgrade from one to another...

EmilienM:
I'm abandoning this one, because it overlaps with some efforts of moving out these blocks of code into Kuryr itself.
cc @MaysaMacedo

PR #1013, opened 2021-03-10
Bug 1937452: Fix golangci-lint violations

danielmellado:
This PR fixed golangci-lint violations in status_manager_test.go which
would make all PR fail on the current CI. I'm ignoring staticcheck on
that file rather than ignoring only the specific SA1019 as per this [1]
issue on golangci-lint.

[1] https://github.com/golangci/golangci-lint/issues/741


abhat:
What version of golangci-lint do we use in CNO? Rather than disabling the check altogether, does changing to a newer version of the linter help instead?

PR #1015, opened 2021-03-11
Fix possible name patterns on regex

MaysaMacedo:
Due to wrong place of usage of '+' the current regex that
handles Ports names allows the following name of ports:
'ostest-7g27t-bootstrap-porttt' and not
'ostest-7g27t-master-port-10', which is wrong as the
installer supports the latter, but not the former.
This commit fixes the issue by including '+' in the
correct place.


MaysaMacedo:
This PR addresses the [comment](https://github.com/openshift/cluster-network-operator/pull/1002#discussion_r591992487) at PR #1002

PR #992, opened 2021-02-25
Bug 1933159: The Multus daemonset should handle 10% maxUnavailable 

dougbtv:
Being that we have on-host OVS, this shouldn't cause connectivity issues for the nodes to go not ready while Multus upgrades on 10% of the nodes.

PR #1019, opened 2021-03-13
Bug 1933159: network-metrics-daemon should use 33% maxUnavailable

smarterclayton:
network-metrics-daemon has no impact on availability of the cluster,
but should rollout no more than one third of all pods at a time so
that an upgrade doesn't completely remove the function.

Consistent with direction expressed in openshift/origin#25928

### DIFF
--- a/bindata/network/network-metrics/001-daemonset.yaml
+++ b/bindata/network/network-metrics/001-daemonset.yaml
@@ -15,6 +15,8 @@ spec:
       app: network-metrics-daemon
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 33%
   template:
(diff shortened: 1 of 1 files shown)


wking:
#992 is already handling `network-metrics-daemon`, although at 10%.  I'm not all that opinionated between 10 and 33.  And it should be pretty easy to rebase #992 to just be about the `multus` DaemonSet if this one ends up landing first.  But splitting into two PRs might make this slightly more tedious to get back to 4.6.z, where we want it to speed up the eventual EUS updates.  :shrug:

smarterclayton:
/hold

I'm fine with 992 covering this, but i think 33% is more appropriate for NMD than 10% (will close once that merges)

PR #1020, opened 2021-03-15
Bug 1938920: Set maxUnavailable for ovs-node daemonset

vrutkovs:
Default settings are inefficient on larger clusters. This PR changes
~~* `ovnkube-master` maxUnavailable to 33% (one node at a time, third of control plane masters)~~
* `ovs-node` - 10% (as it runs on workers as well)

### DIFF
--- a/bindata/network/ovn-kubernetes/006-ovs-node.yaml
+++ b/bindata/network/ovn-kubernetes/006-ovs-node.yaml
@@ -13,6 +13,8 @@ spec:
       app: ovs-node
   updateStrategy:
     type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
   template:
     metadata:
       labels:


dcbw:
@vrutkovs for some reason, some of the CI jobs are still failing with:

```
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Mar 15 09:31:46.425: Daemonsets found that do not meet platform requirements for update strategy:
  expected daemonset openshift-ovn-kubernetes/ovnkube-master to have maxUnavailable 10% or 33% (see comment) instead of 1
```

vrutkovs:
I'm not too familiar with this code base - any ideas why the change doesn't apply?

dcbw:
> I'm not too familiar with this code base - any ideas why the change doesn't apply?

@vrutkovs I have no idea :( I think it should apply, but looking at the must-gather daemonsets it seems, for some reason, to still be `1`.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -33,7 +33,7 @@ spec:
       # by default, Deployments spin up the new pod before terminating the old one
       # but we don't want that - because ovsdb holds the lock.
       maxSurge: 0
-      maxUnavailable: 1
+      maxUnavailable: 33%
```
trozet, line 36: could this be 10%? I thought I remember reading that k8s always rounds up, so if we specify 10% it should still be 1/3. If for some reason we scaled to 7 masters, it will still be safer than doing 33%. Thoughts?

dcbw: @trozet it could be, but the tests check for 10% *or* 33%, so I don't see why this hsouldn't work. The CI failure message was:

```
Mar 15 09:37:25.922: FAIL: Daemonsets found that do not meet platform requirements for update strategy:
  expected daemonset openshift-ovn-kubernetes/ovnkube-master to have maxUnavailable 10% or 33% (see comment) instead of 1
```

smarterclayton: Control plane pods should probably still be maxUnavailable 1.  I will add something to the test that calls that out - there is really no harm updating masters 1 by 1 and that is intentional

smarterclayton:
/hold

See my comment about control plane pods - I didn't add a test expecting control plane pods to be maxUnavailable 1, but I probably should.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
smarterclayton: We will never support 10 masters so leaving this as 1 is probably clearer than 10%

dcbw: @smarterclayton can you link the PR you're doing to allow maxUnavailable=1 for control-plane pods when you post it?

dcbw: currently all our pods fall afoul of your test:

```
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Mar 14 15:24:03.940: Daemonsets found that do not meet platform requirements for update strategy:
  expected daemonset openshift-ovn-kubernetes/ovnkube-master to have maxUnavailable 10% or 33% (see comment) instead of 1
  expected daemonset openshift-ovn-kubernetes/ovs-node to have maxUnavailable 10% or 33% (see comment) instead of 1
```

dcbw:
/approve
/lgtm
/retest
gcp quota issues, aws-sdn flakes

dcbw:
We should be good once https://github.com/openshift/origin/pull/25962 trickles into CI.
/retest

PR #1025, opened 2021-03-16
Bug 1916809: mount node ovs db for multi-gather gather_network_logs

msherif1234:
Description
modify openvswitch mount volume so it will mount ovsdb inside ovn-kube container so it can be consumed by multi-gather tool

Type of change
Enhance debugging of openshift cluster.

How Has This Been Tested?
created Openshift cluster then run

drop in ovnkube node and check the mount is done correctly
[root@ci-ln-kzbqhr2-f76d1-mw92h-worker-d-hnqqd ~]# ls -l /etc/openvswitch/
total 200
-rw-r-----. 1 800 801 129157 Mar 23 14:49 conf.db
-rw-r--r--. 1 800 801    163 Mar 12 23:22 default.conf
-rw-r--r--. 1 800 801     37 Mar 23 12:49 system-id.conf

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -31,7 +31,7 @@ spec:
       priorityClassName: "system-node-critical"
       # volumes in all containers:
       # (container) -> (host)
-      # /etc/openvswitch -> /host/etc/openvswitch - ovsdb system id
+      # /etc/openvswitch -> /etc/openvswitch - ovsdb system id
       # /var/lib/openvswitch -> /var/lib/openvswitch/data - ovsdb data
       # /run/openvswitch -> tmpfs - ovsdb sockets
       # /env -> configmap env-overrides - debug overrides
@@ -298,7 +298,7 @@ spec:
           path: /var/lib/openvswitch/data
       - name: etc-openvswitch
         hostPath:
-          path: /host/etc/openvswitch
+          path: /etc/openvswitch
       - name: run-openvswitch
         hostPath:
           path: /var/run/openvswitch
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -298,7 +298,7 @@ spec:
           path: /var/lib/openvswitch/data
       - name: etc-openvswitch
         hostPath:
-          path: /var/lib/openvswitch/etc
+          path: /host/etc/openvswitch
       - name: run-openvswitch
         hostPath:
           path: /var/run/openvswitch
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -31,7 +31,7 @@ spec:
       priorityClassName: "system-node-critical"
       # volumes in all containers:
       # (container) -> (host)
-      # /etc/openvswitch -> /var/lib/openvswitch/etc - ovsdb system id
+      # /etc/openvswitch -> /host/etc/openvswitch - ovsdb system id
       # /var/lib/openvswitch -> /var/lib/openvswitch/data - ovsdb data
       # /run/openvswitch -> tmpfs - ovsdb sockets
       # /env -> configmap env-overrides - debug overrides


bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -298,7 +298,7 @@ spec:
           path: /var/lib/openvswitch/data
       - name: etc-openvswitch
         hostPath:
-          path: /var/lib/openvswitch/etc
```
astoycos, line 301: I'm not sure if we need this mount for the `ovsdb system id` as stated in the comment atop the daemonset yaml  @abhat Do you know if this mount is used anymore? (I didn't see anything there on the host last time I looked)

msherif1234: @astoycos  FWIW when I checked the folder in openshift cluster it was empty that is why I try to reuse it but yeah if I missed anything I can create new mount

astoycos: Yep I saw the same thing... @abhat Any Idea if we can repurpose this mount here?

danielmellado:
Just note that we're removing ovs from containers to host ovs, so not really sure about the scope of this PR into this at all. Otoh, it's a nit, but please consider best practices when writing commit messages! It'll greatly improve readability if to limit the subject line to 50 chars and then use the description to amend whatever you'd like to, in a slightly larger explanation.

msherif1234:
> Just note that we're removing ovs from containers to host ovs, so not really sure about the scope of this PR into this at all. Otoh, it's a nit, but please consider best practices when writing commit messages!
> 
> It'll greatly improve readability if you limit the subject line to 50 chars and then use the description to amend whatever you'd like to.

Thanks Daniel I will make sure my comments fit in online still updating to the new workflow I will ping you offline to better understand the ovs rework

astoycos:
@danielmellado ACK but I don't think this will be affected by that change, since we are just changing the mount for the ovnkube-node daemonset

bindata/network/ovn-kubernetes/ovnkube-node.yaml
dcbw: If you're on a cluster, you should be able to "ps ax | grep ovs" and see where ovsdb-server on the host puts its db. That's very likely going to be /etc/openvswitch. I think at some past point it used to be in /var/lib/openvswitch/etc but it's just /etc now, I'm 99% sure.

dcbw: But also, ovnkube node gets the system-id from `ovs-vsctl` these days, so I can't think why we'd need direct access to the OVS database. Instead we should just be talking to the dbserver socket.

msherif1234: > If you're on a cluster, you should be able to "ps ax | grep ovs" and see where ovsdb-server on the host puts its db. That's very likely going to be /etc/openvswitch. I think at some past point it used to be in /var/lib/openvswitch/etc but it's just /etc now, I'm 99% sure.

yes Dan if I login w/o chroot its under /host/etc/openvswicth if I ran as root and this what ovn node runs as its under /etc/openvswitch I did test it and I showed the content after modify the mount host path

squeed: what happens if we just remove this mount? is it even needed at all?

squeed: oh duh, nevermind. ignore this.

squeed:
If we're mounting `/host` already, can we just use that?

msherif1234:
> If we're mounting `/host` already, can we just use that?
its not under /host because ovn container its privileged the db is here 

root@ci-ln-kzbqhr2-f76d1-mw92h-worker-d-hnqqd ~]# ls -l /etc/openvswitch/
total 200
-rw-r-----. 1 800 801 129157 Mar 23 14:49 conf.db
-rw-r--r--. 1 800 801 163 Mar 12 23:22 default.conf
-rw-r--r--. 1 800 801 37 Mar 23 12:49 system-id.conf
I tested this w/home/mmahmoud/data/must-gather.local.5971192883354181951/quay-io-mmahmoud-must-gather-test-sha256-0cc86c9604173fe30cde33abcdcd0b71eebd9701a94b0635464edb61ed5d01e9/network_logs
[mmahmoud@mmahmoud network_logs]$ ls -l *conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 887425 Mar 24 07:44 ovnkube-node-7ntlv_conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 166369 Mar 24 07:44 ovnkube-node-gwzqh_conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 681968 Mar 24 07:44 ovnkube-node-jdzwm_conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 191094 Mar 24 07:44 ovnkube-node-kgsj8_conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 690742 Mar 24 07:44 ovnkube-node-mj7wm_conf.db
-rw-r--r--. 1 mmahmoud mmahmoud 527500 Mar 24 07:44 ovnkube-node-w9kjm_conf.db
hole thing e2e with must gather and I saw the collected ovs db

dcbw:
/retest
flakes

PR #1027, opened 2021-03-17
Bug 1939060: OVN Upgrade: fix upgrade order of node and master

fedepaol:
The current logic attempts to restore the old version of the daemonset if shouldUpdateOVNK returns false, but the replacement is done against a list that does not contain the ovnk rendered objects yet.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -141,6 +141,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to render manifests")
 	}
+	objs = append(objs, manifests...)
 
 	updateNode, updateMaster := shouldUpdateOVNK(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, os.Getenv("RELEASE_VERSION"))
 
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -141,6 +141,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to render manifests")
 	}
+	objs = append(objs, manifests...)
```
aojea, line 144: ic, so the things to replace are in manifests, not in objs

fedepaol: yep, manifests contain the rendered ovn-k objects

aojea:
/lgtm
That is a great catch, it took me a while, I think that the variables name are misleading or were the ones that confused me, at least

pkg/network/ovn_kubernetes.go
aojea: I think that the readibility will improve if we s/manifests/renderedObjs/ , but that is only a comment, not asking for changing

trozet:
@squeed FYI

/lgtm

dcbw:
gcp failures are due to capacity issues
bare-metal is due to lack of pod requests in BM operators (unrelated to this PR)

dcbw:
/override ci/prow/ewe-metal-ipi-ovn-ipv6
`: [sig-arch] Managed cluster should set requests but not limits` metal operators do not yet specify their resource requests

dcbw:
: [sig-arch] Managed cluster should set requests but not limits metal operators do not yet specify their resource requests

squeed:
doh, great catch!

PR #1028, opened 2021-03-18
Bug 1907353: Remove OVS daemonsets

jluhrsen:
In 4.6 we moved OVS to systemd, we needed to have a pod capable of
running OVS during the upgrade, but after that it became unnecessary.

In 4.7 we're guaranteed to launch ovs in systemd, both when installed
from scratch and when upgrading from 4.6, therefore there is no use for
the OVS pod other than being able to gather the logs from the apiserver,
we cannot even restart the OVS daemons by deleting the pod, wihch is
confusing.


jluhrsen:
this is an update to https://github.com/openshift/cluster-network-operator/pull/876 just to catch up on  the comment:
  https://github.com/openshift/cluster-network-operator/pull/876#issuecomment-800392932

dcbw:
@jluhrsen you'll need to update the alerts to remove NodeWithoutOVSPod too; that's why some CI jobs are failing.

dcbw:
Also, should we try to move the log mirroring from the ovs pod to the ovn-controller pod perhaps? Or at the very least, we gotta grab `journalctl -u openvswitch` in must-gather too, if we aren't already. I don't think that gets grabbed by the existing must-gather node journal bits. (I could be wrong though).

trozet:
/hold

I think this is doing more harm than good. The OVS logs provide extremely valuable debugging information. Unfortunately those are not collected on a must gather, and iiuc we would be relying on someone to provide journald logs. Even if we said ok, we will add it as part of network must gather to solve this, its still not great, because "network must-gather" is an optional thing that a lot of people forget to run when they create bzs or cases.

abhat:
If I understand it correctly, this needs to be fixed in must-gather. I think there is [provision to specify which services you want to mine host service logs for](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs#L8) and it's about adding `ovs-vswitchd` and `ovsdb-server` to that list. In general, we should try to get rid of this daemonset while mitigating the lack of logs once the PR lands. 

> /hold
> 
> I think this is doing more harm than good. The OVS logs provide extremely valuable debugging information. Unfortunately those are not collected on a must gather, and iiuc we would be relying on someone to provide journald logs. Even if we said ok, we will add it as part of network must gather to solve this, its still not great, because "network must-gather" is an optional thing that a lot of people forget to run when they create bzs or cases.

trozet:
nice @abhat we should also add ovs-configurations.service there

squeed:
Juan already added ovs logs to must-gather: https://github.com/openshift/must-gather/pull/199

jluhrsen:
> Juan already added ovs logs to must-gather: [openshift/must-gather#199](https://github.com/openshift/must-gather/pull/199)

Looks like those are just for the gather network part, but @trozet was noting that we  don't always get those so adding the extra stuff to the standard must-gather is needed.

jluhrsen:
> > Juan already added ovs logs to must-gather: [openshift/must-gather#199](https://github.com/openshift/must-gather/pull/199)
> 
> Looks like those are just for the gather network part, but @trozet was noting that we don't always get those so adding the extra stuff to the standard must-gather is needed.

@abhat , @trozet , I'm guessing it's not this simple or it would have already been done, but here is one way to get the ovs related logs in the standard MG:
https://github.com/openshift/must-gather/pull/225

just go yell at me there to get that right and then I'll address @dcbw up in https://github.com/openshift/cluster-network-operator/pull/1028#issuecomment-809502770 on this PR

jluhrsen:
> @jluhrsen: PR needs rebase.

@abhat , sorry for the delay. rebase is done.

jluhrsen:
/hold

some of the jobs are failing with an alert about NodeWithoutOVSPod. I need to track that failure down and figure out how to
fix it. remove the alert is what I'm guessing. not sure where that happens, but I'll figure it out.

jluhrsen:
@squeed , 5 failing jobs:

e2e-ovn-ipsec-step-registry
  it passed before so job wasn't affected by the alert issue that's now resolved. The latest failure is a rare infra issue that happens across other jobs as well, so unrelated. https://search.ci.openshift.org/?search=Evicted+The+node+was+low+on+resource%3A+memory&maxAge=48h&context=1&type=build-log&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job
  
  e2e-openstack-ovn
    permafails, but looks like it's getting healthier. single failure in last job and looks like https://bugzilla.redhat.com/show_bug.cgi?id=1950430 which is happening in other jobs as well, so unrelated to this PR
    
    e2e-gcp-ovn
    lot's of failures, but the periodic job is also in horrible shape. https://testgrid.k8s.io/redhat-openshift-ocp-release-4.8-informing#release-openshift-ocp-installer-e2e-gcp-ovn-4.8. this job should just be ignored for now
    
    e2e-gcp
    we don't even have this job as a periodic for anything newer than 4.6 to even compare to, but it's failing every time for different reasons. this was a job that caught the alert failure and it's now fixed.
    
    e2e-azure-ovn
    failures are unrelated and recently hitting "https://search.ci.openshift.org/?search=Managed+cluster+should+have+no+crashlooping+pods+in+core+namespaces+over+four+minutes&maxAge=48h&context=1&type=bug%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job" which you can see is happening in a lot of places

squeed:
test failures seem very uninteresting
/override ci/prow/e2e-gcp

PR #1033, opened 2021-03-22
Set the host-network-namespace parameter for ovn-kubernetes

abhat:
This commit puts the host-network-namespace setting in the config
map that ovn-kubernetes mounts to read in the Kubernetes config.


abhat:
Wait on the github.com/ovn-org/ovn-kubernetes/pull/2093 to merge downstream.

squeed:
Seems reasonable, but this is still pending upstream and downstream merges, right?

abhat:
we don't need to gate it on upstream PR merging. Worst case the downstream ovn-k continues to ignore the config map.

abhat:
/hold

Looks like the config file parsing is failing because the downstream PR hasn't landed.

danwinship:
the files need to be renamed to be more specific if they're going to be in a non-host-network-namespace-specific directory (eg, `openshift-sdn/ns.yaml` sounds like it's the openshift-sdn namespace, not the openshift-host-network namespace)

abhat:
@danwinship renamed the files PTAL.

trozet (approved):
/lgtm

abhat:
The downstream PR has merged. Removing the do-not-merge label.

/hold cancel

squeed:
Manually adding valid-bug; this PR was filed ages ago and got caught in CI heck.

squeed:
Great, it decided to retest everything (probably because something else merged in this repo)

PR #1034, opened 2021-03-24
automatic dual-stack conversion

aojea:
OVN kubernetes dual-stack conversion
    
    Add a new annotation to the OVN kubernetes daemonsets, ovnkube-master
    and ovnkube-node, indicating the IP family mode configured:
    - single-stack
    - dual-stack
    
    This annotations allow to automate the conversion of the cluster from
    single-stack to dual-stack and viceversa, since the change in the
    annotation triggers a rollout of the daemonset.
    
    Because the conversion requires to update the master daemonsets first,
    we add the same logic used to manage the versions upgrades to this
    process. The dual-stack conversion also blocks the upgrade, so both
    are exclusive.
    
### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -174,7 +174,24 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 	objs = append(objs, manifests...)
 
-	updateNode, updateMaster := shouldUpdateOVNK(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, os.Getenv("RELEASE_VERSION"))
+	// obtain the current IP family mode.
+	ipFamilyMode := names.IPFamilySingleStack
+	if len(conf.ServiceNetwork) == 2 {
+		ipFamilyMode = names.IPFamilyDualStack
+	}
+	// check if the IP family mode has changed and control the conversion process.
+	updateNode, updateMaster := shouldUpdateOVNKonIPFamilyChange(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, ipFamilyMode)
+	// annotate the daemonset and the daemonset template with the current IP family mode,
+	// this triggers a daemonset restart if there are changes.
+	err = setOVNDaemonsetAnnotation(objs, names.NetworkIPFamilyModeAnnotation, ipFamilyMode)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to set IP family %s annotation on daemonsets", ipFamilyMode)
+	}
+
+	// don't process upgrades if we are handling a dual-stack conversion.
+	if updateMaster && updateNode {
+		updateNode, updateMaster = shouldUpdateOVNKonUpgrade(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, os.Getenv("RELEASE_VERSION"))
+	}
 
 	// If we need to delay master or node daemonset rollout, then we'll replace the new one with the existing one
 	if !updateMaster {
@@ -495,12 +512,44 @@ func listenDualStack(masterIP string) string {
 	}
 }
 
-// shouldUpdateOVNK determines if we should roll out changes to
-// the master and node daemonsets.
-//
-// On upgrades, we roll out nodes first, then masters. Downgrades, we do the
-// opposite.
-func shouldUpdateOVNK(existingNode, existingMaster *appsv1.DaemonSet, releaseVersion string) (updateNode, updateMaster bool) {
+// shouldUpdateOVNKonIPFamilyChange determines if we should roll out changes to
+// the master and node daemonsets on IP family configuration changes.
+// We rollout changes on masters first when there is a configuration change.
+// Configuration changes take precedence over upgrades.
+func shouldUpdateOVNKonIPFamilyChange(existingNode, existingMaster *appsv1.DaemonSet, ipFamilyMode string) (updateNode, updateMaster bool) {
+	// Fresh cluster - full steam ahead!
+	if existingNode == nil || existingMaster == nil {
+		return true, true
+	}
+	// check current daemonsets IP family mode
+	nodeIPFamilyMode := existingNode.GetAnnotations()[names.NetworkIPFamilyModeAnnotation]
+	masterIPFamilyMode := existingMaster.GetAnnotations()[names.NetworkIPFamilyModeAnnotation]
+	// if there are no annotations this is a fresh cluster
+	if nodeIPFamilyMode == "" || masterIPFamilyMode == "" {
+		return true, true
+	}
+	// exit if there are no IP family mode changes
+	if nodeIPFamilyMode == ipFamilyMode && masterIPFamilyMode == ipFamilyMode {
+		return true, true
+	}
+	// If the master config has changed update only the master, the node will be updated later
+	if masterIPFamilyMode != ipFamilyMode {
+		klog.V(2).Infof("IP family mode change detected to %s, updating OVN-Kubernetes master", ipFamilyMode)
+		return false, true
+	}
+	// Don't rollout the changes on nodes until the master daemonset rollout has finished
+	if daemonSetProgressing(existingMaster, false) {
+		klog.V(2).Infof("Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node")
+		return false, true
+	}
+	klog.V(2).Infof("OVN-Kubernetes master daemonset rollout complete, updating IP family mode on node daemonset")
+	return true, true
+}
+
+// shouldUpdateOVNKonUpgrade determines if we should roll out changes to
+// the master and node daemonsets on upgrades. We roll out nodes first,
+// then masters. Downgrades, we do the opposite.
+func shouldUpdateOVNKonUpgrade(existingNode, existingMaster *appsv1.DaemonSet, releaseVersion string) (updateNode, updateMaster bool) {
 	// Fresh cluster - full steam ahead!
 	if existingNode == nil || existingMaster == nil {
 		return true, true
@@ -621,4 +670,32 @@ func daemonSetProgressing(ds *appsv1.DaemonSet, allowHung bool) bool {
 	}
 
 	return true
+}
+
+// setOVNDaemonsetAnnotation annotates the OVNkube master and node daemonset
+// it also annotated the template with the provided key and value to force the rollout
+func setOVNDaemonsetAnnotation(objs []*uns.Unstructured, key, value string) error {
+	for _, obj := range objs {
+		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" &&
+			(obj.GetName() == "ovnkube-master" || obj.GetName() == "ovnkube-node") {
+			// set daemonset annotation
+			anno := obj.GetAnnotations()
+			if anno == nil {
+				anno = map[string]string{}
+			}
+			anno[key] = value
+			obj.SetAnnotations(anno)
+
+			// set pod template annotation
+			anno, _, _ = uns.NestedStringMap(obj.Object, "spec", "template", "metadata", "annotations")
+			if anno == nil {
+				anno = map[string]string{}
+			}
+			anno[key] = value
+			if err := uns.SetNestedStringMap(obj.Object, anno, "spec", "template", "metadata", "annotations"); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
 }
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -37,6 +37,10 @@ const NonCriticalAnnotation = "networkoperator.openshift.io/non-critical"
 // that executing network migration (switching the default network type of the cluster) is allowed.
 const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migration"
 
+// NetworkIPFamilyAnnotation is an annotation on the OVN networks.operator.openshift.io daemonsets
+// to indicate the current IP Family mode of the cluster: "single-stack" or "dual-stack"
+const NetworkIPFamilyModeAnnotation = "networkoperator.openshift.io/ip-family-mode"
+
 // OVNRaftClusterInitiator is an annotation on the networks.operator.openshift.io CR to indicate
 // which node IP was the raft cluster initiator. The NB and SB DB will be initialized by the same member.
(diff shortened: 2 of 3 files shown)


aojea:
@dcbw @danwinship is it ok that the CNO restarts the OVN pods on configuration changes done by the operator?
or do we want more granularity here?

aojea:
/hold
I still need to figure out how to organize the configuration changes and the upgrades rollouts on daemonsets

aojea:
/hold
this is not working as expected, the daemonset are not holding until the rollout finish
@squeed should do they?

aojea:
> @squeed should do they?

nevermind, I was missing `daemonSetProgressing`

squeed:
hmm, why is this needed? shouldn't all config changes cause a restart already? we watch the configmap for changes...

aojea:
> hmm, why is this needed? shouldn't all config changes cause a restart already? we watch the configmap for changes...

https://github.com/ovn-org/ovn-kubernetes/pull/2133

ping @dcbw  @danwinship

danwinship:
@squeed we decided that watching the configmap and restarting was bad because it gives us too little control. In particular, for dual-stack conversion, we need to restart the masters before the nodes. (And for MTU changes, we also want to roll out the change in a very controlled way. Basically, it turns out we never want automatically-restart-on-config-change.)

squeed:
Neat.

Rather than using annotations, how about we version ConfigMaps?

aojea:
> Neat.
> 
> Rather than using annotations, how about we version ConfigMaps?

the annotations force the daemonset update ... other teams are doing a similar approach, as you can see this approach is very simple, and also provides a tool for users and support to verify their configurations

```
oc -n openshift-ovn-kubernetes get daemonset.apps/ovnkube-master \
-o jsonpath='{.spec.template.metadata.annotations.networkoperator\.openshift\.io\/configuration-md5sum}'
333c07b6cdc1cd9dcea624f073df0c05

oc -n openshift-ovn-kubernetes get cm ovnkube-config \
-o jsonpath='{.spec.template.data.ovnkube\.conf}' | md5sum
333c07b6cdc1cd9dcea624f073df0c05
````

I think that the configmap version is a good idea, but it is a project in itself that applies to the whole kubernetes project ... actually there is people interested in solve that upstream https://github.com/kubernetes/kubernetes/issues/94022

**Comment by aojea** [comment]
ok, this is working, ready for review
```
oc get pods -n openshift-ovn-kubernetes
NAME                   READY   STATUS        RESTARTS   AGE
ovnkube-master-2pbw7   6/6     Running       0          7m13s
ovnkube-master-flc26   6/6     Running       0          2m13s
ovnkube-master-l7cgn   6/6     Running       0          4m23s
ovnkube-node-2t8wr     4/4     Running       0          25s
ovnkube-node-2vrvf     4/4     Running       0          34m
ovnkube-node-4k7sh     4/4     Running       0          7s
ovnkube-node-cbxql     4/4     Terminating   0          26m
ovnkube-node-kr5wq     4/4     Running       0          34m
ovnkube-node-qksjb     4/4     Running       0          34m
ovs-node-ctqx6         1/1     Running       0          34m
ovs-node-lwjg7         1/1     Running       0          26m
ovs-node-rvgv2         1/1     Running       0          26m
ovs-node-wxxlx         1/1     Running       0          26m
ovs-node-x7q4k         1/1     Running       0          34m
ovs-node-zsspw         1/1     Running       0          34m

```

**Comment by squeed** [comment]
hang on, I think we're in trouble.

We'll be adding new configuration file directives in 4.8; this means we might crashloopbackoff if we instantaneously roll out configmap changes.

I think we should just make the config part of the DaemonSet and keep it at that.

Separately, it's really important that we always roll out nodes before masters. If, on a y-stream, there is both a config and version change, we need to still hit nodes first...

**Comment by aojea** [comment]
> hang on, I think we're in trouble.
> 
> We'll be adding new configuration file directives in 4.8; this means we might crashloopbackoff if we instantaneously roll out configmap changes.
> 

If we add new directives to the configmap, don't we need to update the pods? Actually, before they will restart automatically, we just changed it to make it controlled now, we don't instantly rollout changes, this is an equivalent of doing a rollout update, each pod is being restarted serially

> I think we should just make the config part of the DaemonSet and keep it at that.
> 
> Separately, it's really important that we always roll out nodes before masters. If, on a y-stream, there is both a config and version change, we need to still hit nodes first...

Upgrade and configuration changes are handled in this pr, if there is configuration changes the upgrade code is not executed

**Comment by danwinship** [comment]
> If, on a y-stream, there is both a config and version change, we need to still hit nodes first...

You are not allowed to change the configuration and upgrade at the same time.

In fact, we should enforce that:
  - CNO should reject all config changes that are made while it is Progressing
  - CNO should set `Upgradeable` to `False` whenever it is rolling out a config change

**Comment by danwinship** [comment]
> I think we should just make the config part of the DaemonSet and keep it at that.

You mean go back to command-line args? It's so ugly but that may actually be the right answer

**Comment by squeed** [comment]
> CNO should reject all config changes that are made while it is Progressing

This is about the config file,  which is an implementation detail and almost certainly going to change between releases

> You mean go back to command-line args? It's so ugly but that may actually be the right answer

Yup :-/Or find some way to embed it in the DaemonSet that doesn't hurt.

**Comment by danwinship** [comment]
> > CNO should reject all config changes that are made while it is Progressing
> 
> This is about the config file, which is an implementation detail and almost certainly going to change between releases

No, I don't care about the config file. We _do not support_ changing the cluster configuration and changing the cluster version at the same time. In particular, we do not support simultaneously upgrading to 4.8 and converting from single-stack to dual-stack. Thus, if an upgrade is in progress, CNO should not let you change the configuration until it is done, and if a config change is being rolled out, CNO should not let you upgrade until it is done.

**Comment by squeed** [comment]
> No, I don't care about the config file.

I agree, but the PR as written does.

**Comment by aojea** [comment]
so, what do you suggest then?

**Comment by squeed** [comment]
> so, what do you suggest then?

It's a bit ugly, but we could embed the contents of the config file in an init container. Or, just go back to cmdline flags; the cmdline and the config file are isomorphic.

**Comment by danwinship** [comment]
> > No, I don't care about the config file.
> 
> I agree, but the PR as written does.

I'm not talking about this PR specifically. I'm talking about configuration changes in general. You can't change the network configuration and upgrade the cluster at the same time, for _any_ value of "change the network configuration". It doesn't matter whether it is technically possible to pass the changed configuration to the operand at the same time as the operand is upgrading; I'm saying that whether or not we _can_ do it, _we don't want to do it_. It's hard enough to guarantee that a cluster where half the nodes are version N and half are version N+1 will still work right. We don't want to try to guarantee that a cluster works when 1/4 of the nodes are version N and old config, 1/4 are version N+1 and old config, 1/4 are version N and new config, and 1/4 are version N+1 and new config.

**Comment by aojea** [comment]
> I'm not talking about this PR specifically. I'm talking about configuration changes in general. You can't change the network configuration and upgrade the cluster at the same time, for any value of "change the network configuration". It doesn't matter whether it is technically possible to pass the changed configuration to the operand at the same time as the operand is upgrading; I'm saying that whether or not we can do it, we don't want to do it. It's hard enough to guarantee that a cluster where half the nodes are version N and half are version N+1 will still work right. We don't want to try to guarantee that a cluster works when 1/4 of the nodes are version N and old config, 1/4 are version N+1 and old config, 1/4 are version N and new config, and 1/4 are version N+1 and new config.

@danwinship @squeed IMHO the configuration change management is something that requires more work, ^^^ all those are great points and we shouldn't rush it an implement it in a drive-by PR.

I've created the scaffolding of what I think can serve as configuration change management, but I've only implemented  the dualstack conversion bits ... at this point of the release (code freeze is tomorrow) I think that is the only option
(I'm missing the testing _working on it_ but please let me know if this approach is not acceptable so I can find an alternative solution for tomorrow)

File: pkg/network/change_manager_test.go [hunk 1]
```diff
@@ -0,0 +1,143 @@
...
+					Type: operv1.NetworkTypeOVNKubernetes,
+				},
+			},
+			bootNode:   &appsv1.DaemonSet{},
+			bootMaster: &appsv1.DaemonSet{},
+			want:       true,
+			wantErr:    false,
+		},
+	}
+	for _, tt := range tests {
```
aojea, line 118: WIP

aojea:
/hold

the annotations on the daemonset template are cleared out and the daemonset rollout doesn't progress

aojea:
/retest
some jobs fail to acquire lease
other, with one of this "alter fired ... " cryptic tests

aojea:
This is ready to go, I've just tested in a real cluster and the rollout goes smoothly 

```


rollout starting

```
NAME                   READY   STATUS        RESTARTS   AGE
ovnkube-master-2lw5f   6/6     Running       4          34m
ovnkube-master-hgzp6   6/6     Terminating   6          34m
ovnkube-master-wwr9p   6/6     Running       0          117s
ovnkube-node-fsrpl     4/4     Running       0          34m
ovnkube-node-kvsjz     4/4     Running       0          34m
ovnkube-node-mctcq     4/4     Running       0          23m
ovnkube-node-nd2rx     4/4     Running       0          23m
ovnkube-node-nnd84     4/4     Running       0          34m
ovnkube-node-vlxzq     4/4     Running       0          24m
```

rollout in progress

```
oc -n openshift-ovn-kubernetes get pods
ovnkube-master-8jzzr   6/6     Running   0          13m
ovnkube-master-wtfh7   6/6     Running   0          10m
ovnkube-master-wwr9p   6/6     Running   0          16m
ovnkube-node-278xd     4/4     Running   0          8m7s
ovnkube-node-7x9lp     4/4     Running   0          7m33s
ovnkube-node-nhwzj     4/4     Running   0          8m26s
ovnkube-node-vphlz     4/4     Running   0          7m43s
ovnkube-node-x29zz     4/4     Running   0          8m34s
ovnkube-node-zdfgt     4/4     Running   0          7m53s
```


cluster-network-operator logs

oc -n openshift-network-operator logs network-operator-6cd6cbbdfd-kxvfd | grep famil
I0409 17:11:31.945113       1 ovn_kubernetes.go:537] IP family mode change detected to dual-stack, updating OVN-Kubernetes master
I0409 17:11:32.981477       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:11:45.830101       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:12:14.922225       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:14:43.181181       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:14:46.801593       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:17:47.811942       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:17:55.625593       1 ovn_kubernetes.go:542] Waiting for OVN-Kubernetes master daemonset IP family mode rollout before updating node
I0409 17:19:34.306974       1 ovn_kubernetes.go:545] OVN-Kubernetes master daemonset rollout complete, updating IP family mode on node daemonset
```

dcbw:
Known issue with 1.21 and dual-stack kube-proxy on Windows

pkg/network/ovn_kubernetes.go
```diff
@@ -174,7 +174,24 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	updateNode, updateMaster := shouldUpdateOVNKonIPFamilyChange(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, ipFamilyMode)
+	// annotate the daemonset and the daemonset template with the current IP family mode,
+	// this triggers a daemonset restart if there are changes.
+	err = setOVNDaemonsetAnnotation(objs, names.NetworkIPFamilyModeAnnotation, ipFamilyMode)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to set IP family %s annotation on daemonsets", ipFamilyMode)
+	}
+
+	// don't process upgrades if we are handling a dual-stack conversion.
+	if updateMaster && updateNode {
```
dcbw, line 192: @aojea do you mean `!updateMaster && !updateNode` here instead? 

If I read the code right, then at this point one or both of updateNode and updateMaster will be true when we need to dual-stack update. But in that case, we *don't* want to do the master/node version upgrades right?

dcbw: @aojea also any chance you can do a testcase that test both the version upgrade and the dual-stack conversion at the same time, so that we can ensure they work correctly together?

aojea: yeah, sure

aojea: It took me a while to understand at first too, updateMaster to true means ok, is an inverse logic, you hold the updates , so it means that you are doing something only when one is false

dcbw: Ok yeah, I see that now; I needed to look at more than just the diff.

abhat: @aojea do you want to add the test that @dcbw mentioned for conversion and upgrade at same time in this PR or do it later in a separate PR?

aojea: It should merge together, will add it later today or tomorrow

dcbw:
Still would like to see a testcase for dual-stack change + version upgrade to make sure they get ordered correctly; but that can be another PR as long as we don't forget. @aojea can you file an issue/bug/whatever for that when you get chance?

known flake `TestWMCO/create/Node_Metadata/ip-10-0-251-49.us-east-2.compute.internal_Validation_Tests/Kubelet_Version (0.00s)` that will settle out after 1.21 and dual-stack Win Proxy fixes propagate through

aojea:
@dcbw test case added

aojea:
/retest

: [sig-network] pods should successfully create sandboxes by writing network status

dcbw:
sdn-multi unrelated
```
fail [github.com/onsi/ginkgo@v4.7.0-origin.0+incompatible/internal/leafnodes/runner.go:113]: Apr 11 23:07:25.570: Unexpected alerts fired or pending after the test run:

alert ClusterOperatorDown fired for 24 seconds with labels: {endpoint="metrics", instance="10.0.182.195:9099", job="cluster-version-operator", name="authentication", namespace="openshift-cluster-version", pod="cluster-version-operator-f58c56c84-n2z5f", service="cluster-version-operator", severity="critical", version="4.8.0-0.ci.test-2021-04-11-220223-ci-op-wm0338ri"}
```

PR #1037, opened 2021-03-25
SDN-1249: Bump openshift/api vendoring to pull egressrouter CRD

tssurya:
Brings in the egress-router CRD changes.


squeed:
It's in the merge pool, should be in soon.

PR #1038, opened 2021-03-25
Bug 1905331: Set requests on multus components

smarterclayton:
All containers should make reasonable cpu and memory requests.
Correct a few places where they were still missing.

Supercedes #925 (some of which merged elsewhere)


smarterclayton:
Major tests have passed - ovn related ones are failing for unrelated issues.  Overriding failing tests.

PR #1039, opened 2021-03-27
ipsec: Always render IPsec DaemonSet

markdgray:
The IPsec pod is only used for IPsec control plane traffic
(i.e. IKE traffic for configured connections, and monitoring
the OVS Open_vSwitch and Interface table for changes). It
uses very minimal resources when not configured and only
marginally more when configured. The DaemonSet can always be rendered
in order to run the two required daemons and will be configured
when the 'ipsec' field is set to 'true' in the OVN NB database
table.

### DIFF
--- a/bindata/network/ovn-kubernetes/ipsec.yaml
+++ b/bindata/network/ovn-kubernetes/ipsec.yaml
@@ -1,4 +1,3 @@
-{{if .OVNIPsecDaemonsetEnable}}
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
@@ -274,4 +273,3 @@ spec:
           path: "{{.CNIConfDir}}"
       tolerations:
       - operator: "Exists"
-{{end}}
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -370,11 +370,8 @@ spec:
                   fi
                 fi
 
-                {{ if .OVNIPsecEnable }}
-                ${OVN_NB_CTL} set nb_global . ipsec=true
-                {{ else }}
-                ${OVN_NB_CTL} set nb_global . ipsec=false
-                {{ end }}
+                ${OVN_NB_CTL} set nb_global . ipsec={{ .OVNIPsecEnable }}
+
           preStop:
             exec:
               command:
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -111,29 +111,10 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVNHybridOverlayVXLANPort"] = ""
 	}
 
-	// If IPsec is enabled for the first time, we start the daemonset. If it is
-	// disabled after that, we do not stop the daemonset but only stop IPsec.
-	//
-	// TODO: We need to do this as, by default, we maintain IPsec state on the
-	// node in order to maintain encrypted connectivity in the case of upgrades.
-	// If we only unrender the IPsec daemonset, we will be unable to cleanup
-	// the IPsec state on the node and the traffic will continue to be
-	// encrypted.
 	if c.IPsecConfig != nil {
-		// IPsec is enabled
-		data.Data["OVNIPsecDaemonsetEnable"] = true
-		data.Data["OVNIPsecEnable"] = true
+		data.Data["OVNIPsecEnable"] = false
 	} else {
-		if bootstrapResult.OVN.ExistingIPsecDaemonset != nil {
-			// IPsec has previously started and
-			// now it has been requested to be disabled
-			data.Data["OVNIPsecDaemonsetEnable"] = true
-			data.Data["OVNIPsecEnable"] = false
-		} else {
-			// IPsec has never started
-			data.Data["OVNIPsecDaemonsetEnable"] = false
-			data.Data["OVNIPsecEnable"] = false
-		}
+		data.Data["OVNIPsecEnable"] = false
 	}
 
 	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network/ovn-kubernetes"), &data)
@@ -231,11 +212,6 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 	if pn.HybridOverlayConfig != nil {
 		if !reflect.DeepEqual(pn.HybridOverlayConfig, nn.HybridOverlayConfig) {
 			errs = append(errs, errors.Errorf("cannot edit a running hybrid overlay network"))
-		}
-	}
-	if pn.IPsecConfig != nil && nn.IPsecConfig != nil {
-		if !reflect.DeepEqual(pn.IPsecConfig, nn.IPsecConfig) {
-			errs = append(errs, errors.Errorf("cannot edit IPsec configuration at runtime"))
 		}
 	}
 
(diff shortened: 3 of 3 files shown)


markdgray:
/cc @squeed 
/cc @dcbw

I have been thinking about the best way to garbage collect the IPsec pods when IPsec is disabled and, in retrospect, I can't think of a good reason why I need to. I dont think any other pods can be unrendered by the CNO and I dont see why IPsec should follow that pattern now that it can be enabled and disabled. Especially because it will unnecessarily complicate how it get created, configured and enabled. What do you think?

As a separate point, I am wondering if the two containers should just form part of the ovnkube-node container (maybe not for this release)?

squeed:
> As a separate point, I am wondering if the two containers should just form part of the ovnkube-node container (maybe not for this release)?

@markdgray That's probably what we should have gone with all along; but rolling that change out after-the-fact takes a bit of planning. That's because there's no coordination between rolling out the new container and deleting the extra DaemonSet.

squeed:
Two questions

squeed:
If you're certain this won't have any chance of somehow affecting the node's networking, I'm okay with this in principle. We're doing this because rolling out the disablement of ipsec is disruptive otherwise?

Two questions:

1 - if we were to make it easy to delete the ipsec daemonset only when ipsec is fully disabled (and thus rolled out) across the cluster, would that be desirable? We now have similar logic for OVN upgrades.
2 - Can we reduce the memory reservation?

bindata/network/ovn-kubernetes/ipsec.yaml
```diff
@@ -1,4 +1,3 @@
-{{if .OVNIPsecDaemonsetEnable}}
```
squeed, line 1: Suggestion: add a log line to the console output in the container that indicates whether ipsec is enabled. I know it has no effect here but I foresee some curious / nervous administrators wondering why their nodes are suddenly running ipsec.

Also, the LivenessProbe is probably a bad idea... it will kill -9 the container if it fails, which is rarely desired. Every LivenessProbe I've ever added turned out to be a bad mistake.

markdgray:
> If you're certain this won't have any chance of somehow affecting the node's networking, I'm okay with this in principle. We're doing this because rolling out the disablement of ipsec is disruptive otherwise?

By design, if IPsec is disabled in the NB database, ovs-monitor-ipsec should not configure libreswan so IPsec should not be configured
> 
> Two questions:
> 
> 1 - if we were to make it easy to delete the ipsec daemonset only when ipsec is fully disabled (and thus rolled out) across the cluster, would that be desirable? We now have similar logic for OVN upgrades.

That is an issue. For example,
* The CNO can only apply kubernetes objects and, currently , cannot delete kubernetes objects.
* Just deleting an kubernetes object is probably not what we want to do with most of them as, I would expect, there would be two stages - 1. de-configure the service hosted by the obeject , 2. delete the actual kubernetes objects
* This becomes a whole lot more complicated as a) you are dealing with distributed services b) OVN itself works in an eventually consistent manner which can sometimes make it difficult to determine if a change has propagated.

This could be done with a specific approach for IPsec but I think my preference would be a more generic approach that could be applied to other features. Either way, it does require not insignificant effort which makes me wonder how important it is that we do this. If I was to think of the equivalent in a standalone server, many services could be running as processes and not configured in the same way that we may have pods running that are not configured.
> 2 - Can we reduce the memory reservation?

Its quite low as it is. I can benchmark again if that helps but I think it was 10s of MB.

bindata/network/ovn-kubernetes/ipsec.yaml
markdgray: > Suggestion: add a log line to the console output in the container that indicates whether ipsec is enabled. I know it has no effect here but I foresee some curious / nervous administrators wondering why their nodes are suddenly running ipsec.
> 

This is my only worry really. "Oh no, I have accidently configured IPsec". I guess the log would help.

> Also, the LivenessProbe is probably a bad idea... it will kill -9 the container if it fails, which is rarely desired. Every LivenessProbe I've ever added turned out to be a bad mistake.

The reason that is here is because we there was a random segmentation fault at initialization (due to a bug in libreswan) that would cause libreswan to die, this would catch this failure and restart the pod (and libreswan). I don't know if the libreswan change has propagated yet (but I haven't seen the segfault in the CI in a while) - s, I anm a bit reluctant to remove it right now.

markdgray:
> > As a separate point, I am wondering if the two containers should just form part of the ovnkube-node container (maybe not for this release)?
> 
> @markdgray That's probably what we should have gone with all along; but rolling that change out after-the-fact takes a bit of planning. That's because there's no coordination between rolling out the new container and deleting the extra DaemonSet.

Yes, I have been thinking about that in retrospect as well. It was hard to know at the time but the more I look at this the more I think it is the right approach. It could be done for 4.8 but maybe the next (or later) release for the reasons that you outlined.

squeed:
> > If you're certain this won't have any chance of somehow affecting the node's networking, I'm okay with this in principle. We're doing this because rolling out the disablement of ipsec is disruptive otherwise?
> 
> By design, if IPsec is disabled in the NB database, ovs-monitor-ipsec should not configure libreswan so IPsec should not be configured

Great, perfect. And I assume it will de-configure everything appropriately too?
> 
> > Two questions:
> > 1 - if we were to make it easy to delete the ipsec daemonset only when ipsec is fully disabled (and thus rolled out) across the cluster, would that be desirable? We now have similar logic for OVN upgrades.
> 
> That is an issue. For example,
> 
>     * The CNO can only apply kubernetes objects and, currently , cannot delete kubernetes objects.

It actually deletes objects that disappear from a previous rendering pass. So that works (though it doesn't help you in this case).

> 
>     * Just deleting an kubernetes object is probably not what we want to do with most of them as, I would expect, there would be two stages - 1. de-configure the service hosted by the obeject , 2. delete the actual kubernetes objects

Agreed, I think an extra idling pod is fine in this case.

The solution would be to annotate nodes with their ipsec status, and not actually delete the DaemonSet until the de-configuration had rolled out. But I agree, that's not worth the engineering effort right now.

> 
> > 2 - Can we reduce the memory reservation?
> 
> Its quite low as it is. I can benchmark again if that helps but I think it was 10s of MB.

Right, but the container requests 100 MiB unconditionally, which is thus forever reserved by the scheduler. Can we lower it to 50? (And bless you for having benchmarks!)

bindata/network/ovn-kubernetes/ipsec.yaml
squeed: Aha. That seems like one of the few cases when a LivenessProbe is actually useful.

markdgray: Yeah, I would potentially like to remove it in the future. I think RHEL9 is when I am sure that fix has been applied and even then, I would spend some time going through the CI logs to see if I saw it show up again.

markdgray:
> > > If you're certain this won't have any chance of somehow affecting the node's networking, I'm okay with this in principle. We're doing this because rolling out the disablement of ipsec is disruptive otherwise?
> > 
> > 
> > By design, if IPsec is disabled in the NB database, ovs-monitor-ipsec should not configure libreswan so IPsec should not be configured
> 
> Great, perfect. And I assume it will de-configure everything appropriately too?

Yes, when the NB change (i.e. 'ipsec=false') has propagated and been handled by "ovn-controller" by modifying the OVS tunnel ports, "ovs-monitor-ipsec" will handle this by removing the ipsec connection from Libreswan.
> 
> > > Two questions:
> > > 1 - if we were to make it easy to delete the ipsec daemonset only when ipsec is fully disabled (and thus rolled out) across the cluster, would that be desirable? We now have similar logic for OVN upgrades.
> > 
> > 
> > That is an issue. For example,
> > ```
> > * The CNO can only apply kubernetes objects and, currently , cannot delete kubernetes objects.
> > ```
> 
> It actually deletes objects that disappear from a previous rendering pass. So that works (though it doesn't help you in this case).

Oh, I looked through the code and didn't see that mechanism. Thats at least one less problem
> 
> > ```
> > * Just deleting an kubernetes object is probably not what we want to do with most of them as, I would expect, there would be two stages - 1. de-configure the service hosted by the obeject , 2. delete the actual kubernetes objects
> > ```
> 
> Agreed, I think an extra idling pod is fine in this case.
> 
> The solution would be to annotate nodes with their ipsec status, and not actually delete the DaemonSet until the de-configuration had rolled out. But I agree, that's not worth the engineering effort right now.

Yes, that was what you suggested and I looked into it. It would work something like:

* IPsec configuration is disabled.
* CNO disables IPsec by rendering an updated ovnkube-master with ipsec disabled
* ipsec pod monitors (polls) ipsec status and adds an annotation to the node stating that there are X IPsec connections or alternatively it annotates when there are 0 connections.
* CNO reconciliation loops un-renders ipsec pod when X=0 and IPsec has been disabled

> 
> > > 2 - Can we reduce the memory reservation?
> > 
> > 
> > Its quite low as it is. I can benchmark again if that helps but I think it was 10s of MB.
> 
> Right, but the container requests 100 MiB unconditionally, which is thus forever reserved by the scheduler. Can we lower it to 50? (And bless you for having benchmarks!)

I ran some tests to check it so I don't know if you could call them "benchmarks". I am also unsure how it scales as I dont have access to a 100 node cluster. I think 50 sounds reasonable but I will have to check again. I can't find my original results.

squeed:
> I think 50 sounds reasonable but I will have to check again. I can't find my original results.

Yup; this isn't a maximum; rather, it's a minimum. So if we can reduce it, that'd be good, especially since this is an idle DaemonSet for most.

markdgray:
I think having the ovn-ipsec pods on the system even when IPsec is not configured may be more confusing to a cluster administrator. Therefore, I think the current implementation is the best option for the moment. I suspect there should eventually be a broader rework of this (perhaps including tighter integration with the ovnkube-node Pod. As a result, I have added a Bugzilla Bz for this issue: https://bugzilla.redhat.com/show_bug.cgi?id=1945682

PR #1043, opened 2021-03-30
Allow Pods Network MTU to be configurable

MaysaMacedo:
When using a Provider Network for the Machines
the same MTU is attempted to be used for the
Tenant Networks of the Pods, which can be considered
too big by OpenStack Neutron. A new option is being included
to allow the user to set the desired MTU for the Pods
Network only before installing the cluster.

Depends-on https://github.com/openshift/api/pull/879


pkg/network/kuryr.go
```diff
@@ -113,7 +113,13 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
 	data.Data["WebhookKey"] = b.WebhookKey
 
 	// Nodes Network MTU
-	data.Data["NodesNetworkMTU"] = b.NodesNetworkMTU
+	mtu  := uint32(b.NodesNetworkMTU)
```
dulek, line 116: Please just change type of `KuryrBootstrapResults.NodesNetworkMTU` to uint32, it has more sense for sure.

pkg/network/kuryr.go
```diff
@@ -113,7 +113,13 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
...
+	if c.MTU != nil && *c.MTU != 0 {
+		if *c.MTU < mtu {
```
dulek, line 118: I think I'd rather report an error in that situation instead of trying to fix user's mistake.

dulek (changes requested):

pkg/network/kuryr.go
```diff
@@ -113,7 +113,15 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
 	data.Data["WebhookKey"] = b.WebhookKey
 
 	// Nodes Network MTU
-	data.Data["NodesNetworkMTU"] = b.NodesNetworkMTU
+	mtu := b.NodesNetworkMTU
+	if c.MTU != nil && *c.MTU != 0 {
+		if mtu >= *c.MTU {
+			mtu = *c.MTU
+		} else {
+			return nil, errors.Errorf("Provided MTU (%d) is too big.", *c.MTU)
```
dulek, line 121: That could be more specific, like:

```
return nil, errors.Errorf("Configured MTU (%d) is incompatible with OpenShift nodes network MTU (%d).", *c.MTU, mtu)
```

pkg/network/kuryr.go
```diff
@@ -113,7 +113,15 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
...
+		}
+	}
+	data.Data["NodesNetworkMTU"] = mtu
```
dulek, line 124: We could probably rename `NodesNetworkMTU` to `PodNetworkMTU`, but that's not a big deal.

pkg/network/kuryr.go
```diff
@@ -196,6 +204,12 @@ func validateKuryr(conf *operv1.NetworkSpec) []error {
 		}
 	}
 
+	if kc != nil {
+		if kc.MTU != nil && (*kc.MTU < 576 || *kc.MTU > 65536) {
```
dulek, line 208: We don't support `0` anymore? Then check on line 117 is not required, right?

dulek:
/lgtm

Minor suggestions in the comments, please check them out.

go.mod
```diff
@@ -32,13 +31,13 @@ require (
 	golang.org/x/net v0.0.0-20210226172049-e18ecbb05110 // indirect
 	golang.org/x/text v0.3.5 // indirect
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.20.4
+	k8s.io/api v0.21.0-rc.0
```
dulek, line 34: Just a pointer - this seems risky, but apparently newest API does exactly that too.

https://github.com/openshift/api/blob/47be53705a1362bfe78926a87dc5d12e0480accf/go.mod#L10-L12

pkg/network/kuryr.go
MaysaMacedo: indeed

pkg/network/kuryr.go
MaysaMacedo: right

squeed:
Do you also need to update `cluster_config.go` to set the correct ClusterNetworkMTU in the status? That would probably be a good idea.

dulek:
> Do you also need to update `cluster_config.go` to set the correct ClusterNetworkMTU in the status? That would probably be a good idea.

Whoa, that's an interesting question. So that option doesn't really control the MTU of the *cluster network* - this is independent of us and is set by installer or user on the OpenStack network they create. Now what the option does is to set the MTU for the networks Kuryr creates for the pods.

Previously we just discovered the MTU of the cluster network and used that for pods networks. The problem with that is that now the cluster network might be a provider (physical) network, meaning that the tenant (tunneled) networks created by Kuryr might need to use lower MTU - and that's why we expose the option to set it.

MaysaMacedo:
@dulek according to description of the `clusterNetworkMTU` field ` ClusterNetworkMTU is the MTU for inter-pod networking.`, I believe it makes sense to include it.

pkg/network/cluster_config.go
```diff
@@ -159,6 +159,8 @@ func StatusFromOperatorConfig(operConf *operv1.NetworkSpec, oldStatus *configv1.
 		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.OpenShiftSDNConfig.MTU)
 	case operv1.NetworkTypeOVNKubernetes:
 		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.OVNKubernetesConfig.MTU)
+	case operv1.NetworkTypeKuryr:
+		status.ClusterNetworkMTU = int(*operConf.DefaultNetwork.KuryrConfig.MTU)
```
dulek, line 163: This will explode on nil, wouldn't it? Checking it isn't necessary for other SDNs as both set a default [1][2]. We don't.

[1] https://github.com/openshift/cluster-network-operator/blob/b1abc131ba75de10deabb2b3e05bbfa74cbdb5ce/pkg/network/openshift_sdn.go#L200-L212
[2] https://github.com/openshift/cluster-network-operator/blob/6faec3b5d3615a37779f4825e9e0abd48db24204/pkg/network/ovn_kubernetes.go#L265-L273

dulek (changes requested):

pkg/network/cluster_config.go
MaysaMacedo: yes, it would. But, I made sure to fill the `KuryrConfig` with the MTU on the `renderKuryr`[1]. I needed to do there instead of fillDefaults cause we need the result of the Bootstrap.

[1] https://github.com/openshift/cluster-network-operator/pull/1043/commits/c6a86b08f3ef7a50e6a08a03cb5559e12ca7cf2b#diff-f66986e5dabe1ee056321bd3ff9bea98a698e49c8c25012f0af7327bfddac82eR123

dulek: Alright, makes sense, thanks!

dulek (approved):
/lgtm

MaysaMacedo:
/hold
need to verify one extra thing regarding changing the mtu value on a running cluster.

PR #1046, opened 2021-04-02
Bug 1943649: Add command for network-check-target instead of hello-openshift

yselkowitz:


smarterclayton:
Thank you - @openshift/networking if you can review so we can break the reference to hello-openshift which is being replaced with specific  use case solutions

cmd/cluster-network-check-target/main.go
```diff
@@ -0,0 +1,51 @@
...
+	"net"
+	"net/http"
+	"os"
+	"strconv"
+)
+
+func checktargetHandler(w http.ResponseWriter, r *http.Request) {
+	response := os.Getenv("RESPONSE")
+	if len(response) == 0 {
+		response = "Hello OpenShift!"
```
danwinship, line 14: we should improve this later but we can get it in as-is now

cmd/cluster-network-check-target/main.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	// Echo back the port the request was received on
+	// via a "request-port" header.
+	addr := r.Context().Value(http.LocalAddrContextKey).(net.Addr)
+	if tcpAddr, ok := addr.(*net.TCPAddr); ok {
+		w.Header().Set("x-request-port", strconv.Itoa(tcpAddr.Port))
+	}
```
danwinship, line 22: (I presume this was added for network-edge's purposes; we don't need this.)

danwinship (approved):
/lgtm

cmd/cluster-network-check-target/main.go
yselkowitz: Now that you mention it, you're right: https://github.com/openshift/origin/commit/938a8e99bcbd0d0ca33bf827dfd209859cf525be

danwinship:
/override ci/prow/e2e-gcp-ovn
that job is very red right now but it's irrelevant to this PR

PR #1048, opened 2021-04-06
copy the service network slice

aojea:
I really don't know if this is a real problem, but it is always better to copy this values instead of passing the references


squeed:
unlikeley

squeed:
unlikely to be a real problem, but can't hurt!

/lgtm
/approve

PR #1049, opened 2021-04-06
Bug 1949099: Fix up network-check-target

danwinship:
#1046 imported hello-openshift as cluster-network-check-target, but if we're going to have our own copy we should customize it a little...

(the gitignore change fixes spurious error messages where it thinks it's supposed to be ignoring `cmd/cluster-network-check-target/` rather than just the top-level `cluster-network-check-target` binary)


squeed:
needs gofmt, but lgtm

PR #1052, opened 2021-04-08
Bug 1936515: Use the election mechanism provided by library-go

fedepaol:
The current election mechanism does not work when the node the operator runs on becomes unreacheable. The old pod is stuck on terminating, hence the lock configmap is not garbage collected and the new instance of the operator can't become the leader. 
In the bug linked, the effect was that the operator wasn't reporting the status as degraded, but what happens is that the operator does not have control anymore.

Here we enable the leader election mechanism provided by library-go.


fedepaol:
/hold
Want to do one final validation

fedepaol:
Tested this with a custom image, scaled up the cno deployment to two and the second instance recognized the leader.
Then tried the scenario described in the BZ, clusteroperator network reports degraded.

PR #1053, opened 2021-04-10
Minor cleanups 

aojea:
- log formatting error
- is not good to swallow errors on the tests, it can give false positives


dcbw:
@aojea looks OK but it failed a unit test?

aojea:
> @aojea looks OK but it failed a unit test?

hmm 
```
TestWithShortWindow
```

maybe I'm not reading ok the logs, I've tested several times locally, let me retest
/retest

PR #1056, opened 2021-04-13
Bug 1939726: Enclose ApplyObject on RetryOnConflict

rcarrillocruz:
This prevents the operator to flip flop on Degraded state
if there is a concurrent update by other external actor.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -11,6 +11,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	k8sclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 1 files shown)


danwinship:
/cc @squeed 

In theory we don't do this because CNO will always try to reconcile it again later anyway. Though we should do that without going degraded at least...

I think this change should be fine?

rcarrillocruz:
Yeah, but with the patch I do not see the blip of going Degraded, then reconciling and changing status again, which is what @smarterclayton referred to in the BZ as I understand it.

pkg/apply/apply.go
```diff
@@ -34,32 +35,39 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
 
-	if err != nil && apierrors.IsNotFound(err) {
-		log.Printf("does not exist, creating %s", objDesc)
-		err := client.Create(ctx, obj)
+		if err != nil && apierrors.IsNotFound(err) {
+			log.Printf("does not exist, creating %s", objDesc)
+			err := client.Create(ctx, obj)
+			if err != nil {
+				log.Printf("create of %s was unsucessful", objDesc)
+				return err
```
squeed, line 46: Can you still wrap the error here?

squeed: And in all places.

squeed: oh crap nevermind. I see why.

pkg/apply/apply.go
```diff
@@ -34,32 +35,39 @@ func ApplyObject(ctx context.Context, client k8sclient.Client, obj *uns.Unstruct
...
+			} else {
+				log.Printf("update was successful")
+			}
+		}
+		return nil
+	})
 
-	return nil
+	return err
```
squeed, line 72: Can you wrap the error here with something a bit more descriptive?

squeed:
seems reasonable. There are probably a few places we update objects directly - it would be worthwhile to grep for `.Update(` and see if there are any other unwrapped ones.

PR #1057, opened 2021-04-14
Bug 1937594: Bump openshift/api

pliurh:
This allow us to use the latest API in networks.config and
networks.operator.


danwinship:
/lgtm
(I didn't mean it had to be a separate PR, just a separate commit within #763...)

PR #1058, opened 2021-04-14
Bug 1949677: The Multus daemonset should be separate other CNI binary installation

dougbtv:
This separates into two daemonsets, one for Multus itself, and one for all other CNI binary and configuration installation.

The goal is to improve the time it takes a node to become ready, as images are pulled serially.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -11,6 +11,7 @@ metadata:
 data:
   cnibincopy.sh: |-
     #!/bin/bash
+    set -e
 
     DESTINATION_DIRECTORY=/host/opt/cni/bin/
 
@@ -118,22 +119,47 @@ spec:
       tolerations:
       - operator: Exists
       serviceAccountName: multus
-      initContainers:
-      - name: multus-binary-copy
+      containers:
+      - name: kube-multus
         image: {{.MultusImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: [ "/bin/bash", "-ec", "--" ]
+        args:
+          - >
+            /entrypoint/cnibincopy.sh;
+            exec /entrypoint.sh
+            --multus-conf-file=auto
+            --multus-autoconfig-dir=/host/var/run/multus/cni/net.d
+            --multus-kubeconfig-file-host=/etc/kubernetes/cni/net.d/multus.d/multus.kubeconfig
+{{- if eq .DefaultNetworkType "OpenShiftSDN"}}
+            --readiness-indicator-file=/var/run/multus/cni/net.d/80-openshift-network.conf
+{{- else if eq .DefaultNetworkType "OVNKubernetes"}}
+            --readiness-indicator-file=/var/run/multus/cni/net.d/10-ovn-kubernetes.conf
+{{- end}}
+            --cleanup-config-on-exit=true
+            --namespace-isolation=true
+            --multus-log-level=verbose
+            --cni-version=0.3.1
+            --additional-bin-dir=/opt/multus/bin
+            --skip-multus-binary-copy=true
+            - "--global-namespaces=default,openshift-multus,openshift-sriov-network-operator"
         resources:
           requests:
             cpu: 10m
-            memory: 10Mi
+            memory: 65Mi
+        securityContext:
+          privileged: true
+        terminationGracePeriodSeconds: 10
         volumeMounts:
         - mountPath: /entrypoint
           name: cni-binary-copy
-        - mountPath: /host/opt/cni/bin
-          name: cnibin
         - mountPath: /host/etc/os-release
           name: os-release
-          readOnly: true
+        - name: system-cni-dir
+          mountPath: /host/etc/cni/net.d
+        - name: multus-cni-dir
+          mountPath: /host/var/run/multus/cni/net.d
+        - name: cnibin
+          mountPath: /host/opt/cni/bin
         env:
         - name: RHEL7_SOURCE_DIRECTORY
           value: "/usr/src/multus-cni/rhel7/bin/"
@@ -141,6 +167,62 @@ spec:
           value: "/usr/src/multus-cni/rhel8/bin/"
         - name: DEFAULT_SOURCE_DIRECTORY
           value: "/usr/src/multus-cni/bin/"
+        - name: KUBERNETES_SERVICE_PORT
+          value: "{{.KUBERNETES_SERVICE_PORT}}"
+        - name: KUBERNETES_SERVICE_HOST
+          value: "{{.KUBERNETES_SERVICE_HOST}}"
+      volumes:
+        - name: system-cni-dir
+          hostPath:
+            path: {{ .SystemCNIConfDir }}
+        - name: multus-cni-dir
+          hostPath:
+            path: {{ .MultusCNIConfDir }}
+        - name: cnibin
+          hostPath:
+            path: {{ .CNIBinDir }}
+        - name: os-release
+          hostPath:
+            path: /etc/os-release
+            type: File
+        - name: cni-binary-copy
+          configMap:
+            name: cni-binary-copy-script
+            defaultMode: 0744
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: multus-additional-cni-plugins
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon installs and configures auxiliary CNI plugins on each node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: multus-additional-cni-plugins
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 10%
+  template:
+    metadata:
+      labels:
+        app: multus-additional-cni-plugins
+        component: network
+        type: infra
+        openshift.io/component: network
+    spec:
+      hostNetwork: true
+      nodeSelector:
+        kubernetes.io/os: linux
+      priorityClassName: "system-node-critical"
+      tolerations:
+      - operator: Exists
+      serviceAccountName: multus
+      initContainers:
       - name: egress-router-binary-copy
         image: {{.EgressRouterImage}}
         command: ["/entrypoint/cnibincopy.sh"]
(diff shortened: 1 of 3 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -139,6 +139,96 @@ spec:
...
+            defaultMode: 0744
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: multus-additional-cni-plugins
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the Multus networking component on each node.
```
wking, line 206: Can we tweak this text so folks can distinguish from the description used for `openshift-multus`?

bindata/network/multus/multus.yaml
```diff
@@ -139,6 +139,96 @@ spec:
...
+        - "--readiness-indicator-file=/var/run/multus/cni/net.d/80-openshift-network.conf"
+{{- else if eq .DefaultNetworkType "OVNKubernetes"}}
+        - "--readiness-indicator-file=/var/run/multus/cni/net.d/10-ovn-kubernetes.conf"
+{{- end}}
+        - "--cleanup-config-on-exit=true"
+        - "--namespace-isolation=true"
+        - "--multus-log-level=verbose"
+        - "--cni-version=0.3.1"
+        - "--additional-bin-dir=/opt/multus/bin"
+        - "--skip-multus-binary-copy=true"
```
smarterclayton, line 160: So question - why do we need the init container to do the copy?  Why not simply remove the init container and have entrypoint do it here?  Does the init container get us anything?

smarterclayton: If it's just a different script, why not turn this into 

```
command:
- /bin/bash
- -c
-|
/entrypoint/cnibincopy.sh
exec /entrypoint.sh
```

?

smarterclayton: (if we can do this the advantage is we collapse two containers into one which reduces serialization cost and extra writes to the apiserver, improving the speed at which multus starts even more)

dougbtv: I like having the copy being common to all the CNI binaries, as we use the same method for all CNI binaries. So instead of packing it all into the entrypoint script, I've opted to remove the init container and run the two scripts in serial -- binary copy, and then the entrypoint.

bindata/network/multus/multus.yaml
dougbtv: Thanks for the eye, updated description.

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +116,109 @@ spec:
...
+            defaultMode: 0744
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: multus-additional-cni-plugins
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      This daemon installed and configures auxiliary CNI plugins on each node.
```
wking, line 196: nit: "installed" -> "installs"

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +116,109 @@ spec:
...
-      - name: multus-binary-copy
+      containers:
+      - name: kube-multus
         image: {{.MultusImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: [ "/bin/bash", "-c", "--" ]
+        args:
+          - >
+            /entrypoint/cnibincopy.sh;
+            /entrypoint.sh
```
wking, line 126: don't you want to `exec /entrypoint.sh ...` this, so you don't have to worry about Bash sitting in the middle and possibly complicating signal handling?

bindata/network/multus/multus.yaml
```diff
@@ -338,24 +418,10 @@ spec:
...
-        - "--namespace-isolation=true"
-        - "--multus-log-level=verbose"
-        - "--cni-version=0.3.1"
-        - "--additional-bin-dir=/opt/multus/bin"
-        - "--skip-multus-binary-copy=true"
+        command: [ "/bin/bash", "-c", "--" ]
+        args: [ "trap : TERM INT; sleep infinity & wait" ]
         resources:
           requests:
             cpu: 10m
```
wking, line 427: down below, now that this is just a sleep, it can probably have it's memory request dropped below `150Mi` ;)

bindata/network/multus/multus.yaml
dougbtv: So, this is why I had the init container originally, was that it simplified this process, that I want the installation of the binary, and the arguments for the configuration of Multus to be separate.... any further ideas on this one on how I might do it without the entrypoint?

bindata/network/multus/multus.yaml
dougbtv: Good call, got it at `10Mi`

bindata/network/multus/multus.yaml
smarterclayton: I would generally do:

```
command:
- /bin/bash
- -c
-|
  set -e
  <env> /entrypoint/cnibincopy.sh
  exec /entrypoint.sh $@
args:
- ...
```

That preserves the existing behavior of both.

dougbtv: gah, I was being dense. I added the bash command `exec` for running entrypoint.sh as suggested, thanks :+1:

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +116,109 @@ spec:
...
-      initContainers:
-      - name: multus-binary-copy
+      containers:
+      - name: kube-multus
         image: {{.MultusImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: [ "/bin/bash", "-c", "--" ]
+        args:
+          - >
+            /entrypoint/cnibincopy.sh;
```
smarterclayton, line 125: you still need `set -e`, or you can add `-ec` to the second command arg.  Otherwise if cnibincopy fails your scirpt keeps going.

dougbtv: I changed to run this commands with `&&` and added `set -e` to the cni binary copy script.

smarterclayton: set -e means you don't need && :)

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +117,109 @@ spec:
...
-      initContainers:
-      - name: multus-binary-copy
+      containers:
+      - name: kube-multus
         image: {{.MultusImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: [ "/bin/bash", "-c", "--" ]
+        args:
+          - >
+            /entrypoint/cnibincopy.sh &&
```
smarterclayton, line 126: If we add another flag this could get confusing.  I'd recommend switching `-c` above to `-ec` and then removing this &&

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +117,109 @@ spec:
...
+            --readiness-indicator-file=/var/run/multus/cni/net.d/80-openshift-network.conf
+{{- else if eq .DefaultNetworkType "OVNKubernetes"}}
+            --readiness-indicator-file=/var/run/multus/cni/net.d/10-ovn-kubernetes.conf
+{{- end}}
+            --cleanup-config-on-exit=true
+            --namespace-isolation=true
+            --multus-log-level=verbose
+            --cni-version=0.3.1
+            --additional-bin-dir=/opt/multus/bin
+            --skip-multus-binary-copy=true
```
smarterclayton, line 141: Not sure whether this was answered before, what's the difference between letting multus do the binary copy and /entrypoint/cnibincopy.sh above?  What does the custom script do that multus entrypoint does not do by default?

dougbtv: The binary copy here is an upstream option to let you copy the binary, but... it's "dumb" compared to how we do it in OCP. In OpenShift we want to copy the binary so that it's matched with the underlying host o/s. Which is why we have the mounted `cnibincopy.sh` which has the smarts to do this in the same fashion for all of the CNI binaries that we copy.

bindata/network/multus/multus.yaml
```diff
@@ -116,29 +117,109 @@ spec:
...
         resources:
           requests:
             cpu: 10m
-            memory: 10Mi
+            memory: 150Mi
```
smarterclayton, line 145: I forgot to ask, does multus actually use 150mi of memory?

smarterclayton: Trevor already asked, expected this to change

smarterclayton:
Not completely scientific, but it looks like here's what the cgroup thinks the max is for multus:

image

I would request 60Mi for the whole pod

bindata/network/multus/multus.yaml
dougbtv: Thanks, apparently I missed this. I set to 65Mi, using your 55 megs as a baseline and then added 15% and rounded up to the nearest 5 to give a little extra.

dougbtv:
seems to run OK with cluster-bot and `launch openshift/cluster-network-operator#1058`, so.... giving it a

/retest

smarterclayton:
/retest

It looks like this improves node restart times - I saw a 3m and a 2m30s full node upgrade, where normally they are 3m30s-4m.  So that's a 33% increase if it holds out!

s1061123 (approved):
It looks good to me, thanks!

dougbtv:
/test e2e-aws-sdn-multi
--


<br class="Apple-interchange-newline">

smarterclayton:
The last passing aws-sdn-multi was https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1078/pull-ci-openshift-cluster-network-operator-master-e2e-aws-sdn-multi/1391879544251617280, on the 10th (2 weeks ago)

smarterclayton:
Last passing ovn-windows was 5/11 in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1059/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-windows/1392183660664328192

smarterclayton:
I think this looks like it is passing on all platforms where the job isn't perma broken.

PR #1059, opened 2021-04-15
Bug 1825219: drop icmp frag needed received from other nodes in the cluster

mccv1r0:
drop icmp frag needed received from other nodes in the cluster

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -235,6 +235,63 @@ spec:
         - name: sdn-metrics-certs
           mountPath: /etc/pki/tls/metrics-certs
           readOnly: True
+      {{- if .SDNPlatformAzure}}
+      - name: drop-icmp
+        image: {{.SDNImage}}
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -xe
+
+          touch /var/run/add_iptables.sh
+          chmod 0755 /var/run/add_iptables.sh
+          cat <<'EOF' > /var/run/add_iptables.sh
+          #!/bin/sh
+          if [ -z "$3" ]
+          then
+               echo "Called with host address missing, ignore"
+               exit 0
+          fi
+          echo "Adding ICMP drop rule for '$3' "
+          if iptables -C CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+          then
+               echo "iptables already set for $3"
+          else
+               iptables -A CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+          fi
+          EOF
+
+          echo "I$(date "+%m%d %H:%M:%S.%N") - drop-icmp - start drop-icmp ${K8S_NODE}"
+          iptables -X CHECK_ICMP_SOURCE || true
+          iptables -N CHECK_ICMP_SOURCE || true
+          iptables -F CHECK_ICMP_SOURCE
+          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE || true
+          iptables -I INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE
+          iptables -N ICMP_ACTION || true
+          iptables -F ICMP_ACTION
+          iptables -A ICMP_ACTION -j LOG
+          iptables -A ICMP_ACTION -j DROP
+          oc observe pods -n openshift-sdn -l app=sdn -a '{ .status.hostIP }' -- /var/run/add_iptables.sh
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/bin/bash", "-c", "echo drop-icmp done"]
+        securityContext:
+          privileged: true
+        volumeMounts:
+        - mountPath: /
+          name: host-slash
+        resources:
+          requests:
+            cpu: 5m
+            memory: 20Mi
+        env:
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+      {{- end}}
       nodeSelector:
         kubernetes.io/os: linux
       volumes:
--- a/pkg/network/openshift_sdn.go
+++ b/pkg/network/openshift_sdn.go
@@ -1,6 +1,8 @@
 package network
 
 import (
+	"context"
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
@@ -9,9 +11,15 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 
+	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
+	types "k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
+	configv1 "github.com/openshift/api/config/v1"
 	netv1 "github.com/openshift/api/network/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/render"
@@ -24,7 +32,7 @@ import (
 // - the sdn daemonset
 // - the openvswitch daemonset
 // and some other small things.
-func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstructured, error) {
+func renderOpenShiftSDN(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult, manifestDir string) ([]*uns.Unstructured, error) {
 	c := conf.DefaultNetwork.OpenShiftSDNConfig
 
 	objs := []*uns.Unstructured{}
@@ -39,6 +47,11 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	data.Data["Mode"] = c.Mode
 	data.Data["CNIConfDir"] = pluginCNIConfDir(conf)
 	data.Data["CNIBinDir"] = CNIBinDir
+	if bootstrapResult.SDN.Platform == configv1.AzurePlatformType {
+		data.Data["SDNPlatformAzure"] = true
+	} else {
+		data.Data["SDNPlatformAzure"] = false
+	}
 
 	clusterNetwork, err := clusterNetwork(conf)
 	if err != nil {
@@ -80,6 +93,7 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	}
 
 	objs = append(objs, manifests...)
+
 	return objs, nil
 }
 
@@ -265,4 +279,28 @@ func clusterNetwork(conf *operv1.NetworkSpec) (string, error) {
 	}
 
 	return string(cnBuf), nil
+}
+
+func bootstrapSDN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
+
+	var platformType configv1.PlatformType
+
+	infraConfig := &configv1.Infrastructure{}
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+		return nil, fmt.Errorf("failed to get infrastructure 'config': %v", err)
+	}
+
+	if infraConfig != nil {
+		klog.V(2).Infof("Openshift-SDN: Bootstrap SDN infraConfig Platform: %v ", infraConfig.Status.PlatformStatus.Type)
+		if infraConfig.Status.PlatformStatus.Type != "" {
+			platformType = infraConfig.Status.PlatformStatus.Type
+		}
+	}
+
+	res := bootstrap.BootstrapResult{
+		SDN: bootstrap.SDNBootstrapResult{
+			Platform: platformType,
+		},
+	}
+	return &res, nil
 }
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -3,6 +3,7 @@ package bootstrap
 import (
 	"github.com/gophercloud/utils/openstack/clientconfig"
 
+	configv1 "github.com/openshift/api/config/v1"
 	appsv1 "k8s.io/api/apps/v1"
 )
 
@@ -33,9 +34,15 @@ type OVNBootstrapResult struct {
 	ExistingNodeDaemonset   *appsv1.DaemonSet
 	ExistingIPsecDaemonset  *appsv1.DaemonSet
 	GatewayMode             string
+	Platform                configv1.PlatformType
 }
 
 type BootstrapResult struct {
 	Kuryr KuryrBootstrapResult
 	OVN   OVNBootstrapResult
+	SDN   SDNBootstrapResult
+}
+
+type SDNBootstrapResult struct {
+	Platform configv1.PlatformType
 }
--- a/pkg/network/bootstrap.go
+++ b/pkg/network/bootstrap.go
@@ -14,7 +14,7 @@ func Bootstrap(conf *operv1.Network, client client.Client) (*bootstrap.Bootstrap
 	case operv1.NetworkTypeKuryr:
 		return openstack.BootstrapKuryr(&conf.Spec, client)
 	case operv1.NetworkTypeOpenShiftSDN:
-		return nil, nil
+		return bootstrapSDN(conf, client)
 	case operv1.NetworkTypeOVNKubernetes:
 		return bootstrapOVN(conf, client)
 	}
--- a/pkg/network/openshift_sdn_test.go
+++ b/pkg/network/openshift_sdn_test.go
@@ -7,6 +7,7 @@ import (
 
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/apply"
+	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
@@ -44,11 +45,15 @@ func TestRenderOpenShiftSDN(t *testing.T) {
 	crd := OpenShiftSDNConfig.DeepCopy()
 	config := &crd.Spec
 
+	bootstrapResult := &bootstrap.BootstrapResult{
+		SDN: bootstrap.SDNBootstrapResult{},
+	}
+
 	errs := validateOpenShiftSDN(config)
 	g.Expect(errs).To(HaveLen(0))
 	FillDefaults(config, nil)
 
-	objs, err := renderOpenShiftSDN(config, manifestDir)
+	objs, err := renderOpenShiftSDN(config, bootstrapResult, manifestDir)
 	g.Expect(err).NotTo(HaveOccurred())
 
 	// It's important that the namespace is first
@@ -189,6 +194,10 @@ func TestProxyArgs(t *testing.T) {
 	config := &crd.Spec
 	FillDefaults(config, nil)
 
+	bootstrapResult := &bootstrap.BootstrapResult{
+		SDN: bootstrap.SDNBootstrapResult{},
+	}
+
 	// iter through all objects, finding the kube-proxy config map
 	getProxyConfigFile := func(objs []*uns.Unstructured) *uns.Unstructured {
 		for _, obj := range objs {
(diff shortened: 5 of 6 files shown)


bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+          if [[ -f "/env/_master" ]]; then
+            set -o allexport
+            source "/env/_master"
+            set +o allexport
+          fi
+
+          echo "I$(date "+%m%d %H:%M:%S.%N") - drop-icmp - start drop-icmp ${K8S_NODE}"
+          iptables -X CHECK_ICMP_SOURCE || true
+          iptables -N CHECK_ICMP_SOURCE || true
+          iptables -F CHECK_ICMP_SOURCE
```
Miciah, line 70: Is (1) deleting and creating and then (2) flushing redundant?  Seems like either (1) or (2) alone would suffice.  You don't delete and recreate the `ICMP_ACTION` chain below.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE || true
+          iptables -I INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE
+          iptables -N ICMP_ACTION || true
+          iptables -F ICMP_ACTION
+          iptables -A ICMP_ACTION -j LOG
+          iptables -A ICMP_ACTION -j DROP
+          oc observe nodes -a '{ .status.addresses[1].address }' -- /tmp/add_iptables.sh
+          tail -F /dev/null
```
Miciah, line 78: If `oc observe` exits, what should happen?  Running `tail -F /dev/null` means the continue will keep running but doing nothing, but it seems like you'd want the container to restart.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+        volumeMounts:
+        - mountPath: /
+          name: host-slash
+        - name: add-iptables
+          mountPath: /tmp/add_iptables.sh
+          subPath: add_iptables.sh
+        resources:
+          requests:
+            cpu: 10m
+            memory: 300Mi
```
Miciah, line 94: This seems really high, especially the memory request.  I'd have expected something more like 5m CPU and 20Mi memory.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+        env:
+        - name: OVN_KUBE_LOG_LEVEL
```
Miciah, line 96: Does this have any effect?

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+          value: "4"
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: K8S_NODE_IP
```
Miciah, line 102: Is this needed?

pkg/network/openshift_sdn.go
```diff
@@ -9,11 +11,17 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
+	types "k8s.io/apimachinery/pkg/types"
```
Miciah, line 16: This grouping is weird.  `operv1 "github.com/openshift/api/operator/v1"` should be grouped with the other openshift/api imports as it was before, `"github.com/openshift/cluster-network-operator/pkg/bootstrap"` and `"k8s.io/klog/v2"` probably should be in separate groups, and `types "k8s.io/apimachinery/pkg/types"` can be grouped with the other apimachinery imports.

pkg/network/openshift_sdn.go
```diff
@@ -266,3 +286,20 @@ func clusterNetwork(conf *operv1.NetworkSpec) (string, error) {
...
 }
+
+func bootstrapSDN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.BootstrapResult, error) {
+
+	infraConfig := &configv1.Infrastructure{}
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+		return nil, fmt.Errorf("failed to get infrastructure 'config': %v", err)
+	}
+
+	klog.V(2).Infof("Openshift-SDN: Bootstrap SDN infraConfig Platform: %v ", infraConfig.Status.PlatformStatus.Type)
```
Miciah, line 297: Check for nil `infraConfig.Status.PlatformStatus` (it could be nil if the cluster was installed with OpenShift 4.1 and upgraded).

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+      labels:
+        app: drop-icmp
+        component: network
+        type: infra
+        openshift.io/component: network
+        kubernetes.io/os: "linux"
+    spec:
+      serviceAccountName: sdn-controller
+      hostNetwork: true
+      hostPID: true
```
Miciah, line 50: Why is `hostPID: true` needed?

pkg/network/openshift_sdn.go
```diff
@@ -80,6 +88,18 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
 	}
 
 	objs = append(objs, manifests...)
+
+	klog.V(2).Infof("Openshift-SDN: Bootstrap Platform found %v  Azure platform value: %v", bootstrapResult.SDN.Platform, configv1.AzurePlatformType)
+	// Add Azure specific yaml needed to work around Azure specific issues
+	if bootstrapResult.SDN.Platform == configv1.AzurePlatformType {
+		manifestsAzure, err := render.RenderDir(filepath.Join(manifestDir, "network/openshift-sdn-azure"), &data)
```
Miciah, line 95:
```suggestion
		manifestsAzure, err := render.RenderDir(filepath.Join(filepath.Join(manifestDir, "network"), "openshift-sdn-azure"), &data)
```

pkg/network/openshift_sdn.go
```diff
@@ -80,6 +88,18 @@ func renderOpenShiftSDN(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Un
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to render azure specific manifests")
+		}
+		klog.V(2).Infof("Openshift-SDN: Starting Azure Append")
+		objs = append(objs, manifestsAzure...)
+		klog.V(2).Infof("Openshift-SDN: Azure Append Completed")
```
Miciah, line 101: This is like logging a malloc—seems redundant.  Probably just one log message (if even that) would suffice: 
```suggestion
		klog.V(2).Infof("Openshift-SDN: Including Azure-specific manifests")
		objs = append(objs, manifestsAzure...)
```

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+      priorityClassName: "system-cluster-critical"
+      containers:
+      # ovnkube master: convert kubernetes objects in to nbdb logical network components
+      - name: drop-icmp
+        image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:bcc1fb20f06f00829727cb46ff21e22103fd4c737fdcbbf2fab13121f31ebcbd
```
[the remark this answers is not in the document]
mccv1r0: OvnImage?  Or SdnImage?

pkg/network/openshift_sdn.go
mccv1r0: ack s/nil/""/

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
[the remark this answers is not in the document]
mccv1r0: I stole this image from ARO.  Newer images don't contain iptables binary and we didn't want to re-write solution that was tested, working and accepted by customers in the field.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+          iptables -X CHECK_ICMP_SOURCE || true
+          iptables -N CHECK_ICMP_SOURCE || true
+          iptables -F CHECK_ICMP_SOURCE
+          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE || true
+          iptables -I INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE
+          iptables -N ICMP_ACTION || true
+          iptables -F ICMP_ACTION
+          iptables -A ICMP_ACTION -j LOG
+          iptables -A ICMP_ACTION -j DROP
+          oc observe nodes -a '{ .status.addresses[1].address }' -- /tmp/add_iptables.sh
```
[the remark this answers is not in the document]
mccv1r0: We only want the primary external (aka "eth0")

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: add-iptables
+  namespace: openshift-sdn
+data:
+  add_iptables.sh: |
+   #!/bin/sh
+   echo "Adding ICMP drop rule for '$2' "
+   #iptables -C CHECK_ICMP_SOURCE -p icmp -s $2 -j ICMP_ACTION || iptables -A CHECK_ICMP_SOURCE -p icmp -s $2 -j ICMP_ACTION
```
mccv1r0, line 10: Removed this

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,126 @@
...
+   if iptables -C CHECK_ICMP_SOURCE -p icmp -s $2 -j ICMP_ACTION
+   then
+        echo "iptables already set for $2"
+   else
+        iptables -A CHECK_ICMP_SOURCE -p icmp -s $2 -j ICMP_ACTION
+   fi
+   #iptables -nvL
```
mccv1r0, line 17: Commented code removed

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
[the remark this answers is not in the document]
mccv1r0: Using .SDNImage, and (thankfully) iptables is still part of it.  This postpones the need to move to nft now.

mccv1r0:
Fixes https://github.com/openshift/cluster-network-operator/issues/1086

pkg/network/openshift_sdn.go
Miciah: `infraConfig.Status.PlatformStatus` itself could be nil.  The API field was added in OpenShift 4.2, and a cluster that was installed on OpenShift 4.1 and upgraded 4.1 → 4.2 → ... → 4.8 might not have this field set.  We've had BZs in the past when we forgot to check `Status.PlatformStatus` for nil (e.g., https://bugzilla.redhat.com/show_bug.cgi?id=1734192, https://bugzilla.redhat.com/show_bug.cgi?id=1834194).  We should try not to repeat that mistake yet another time.  * grin *.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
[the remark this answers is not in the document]
Miciah: Is it safe to assume that `.status.addresses[1].address` is always the primary external?  Sorry if this is a silly question, but I want to make sure we aren't overlooking if there's a possibility that this assumption may be invalid.

Miciah:
GitHub is being wonky, so I don't know if my last review posted, but there are two outstanding issues:
* Is it a safe to assume that `.status.addresses[1].address` is always the primary external address? https://github.com/openshift/cluster-network-operator/pull/1059#discussion_r629729421
* We need to check whether `infraConfig.Status.PlatformStatus` is nil.  https://github.com/openshift/cluster-network-operator/pull/1059#discussion_r629727464

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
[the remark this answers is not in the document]
aojea: external is not eth0, kubernetes node addresses are ... complicated , @danwinship is the expert https://github.com/kubernetes/kubernetes/pull/95768

if you want to get eth0 IP in a cloud environment, it uses to be the HostIP , so I suggest you to use the downward API to get it from an environment variable the same you are doing with the nodename:
```yaml
        env:
        - name: HOST_IP
          valueFrom:
            fieldRef:
              fieldPath: status.hostIP
```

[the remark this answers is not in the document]
danwinship: Yeah, you can't assume `.status.addresses[1].address` is the one you want.
Easiest thing would be to `oc observe pods -n openshift-sdn -l app=drop-icmp -a '{ .status.hostIP }'` (not tested). ie, watch the hostIPs of the other pods in the drop-icmp DaemonSet.

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: drop-icmp
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      # by default, Deployments spin up the new pod before terminating the old one
+      # but we don't want that - because ovsdb holds the lock.
```
danwinship, line 34: this comment does not apply to this DaemonSet, which may mean that setting `maxSurge` doesn't apply

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+        echo "iptables already set for $2"
+   else
+        iptables -A CHECK_ICMP_SOURCE -p icmp -s $2 -j ICMP_ACTION
+   fi
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: drop-icmp
+  namespace: openshift-sdn
```
danwinship, line 21: rather than being its own DaemonSet, it could just be another container in the sdn DaemonSet, conditionally included only on Azure. ie, instead of conditionally calling `render.RenderDir` again, just set another variable in `data` and then have

    {{- if .PlatformAzure}}
    - name: drop-icmp
      ...
    {{- end}}

in the openshift-sdn containers

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+        command:
+        - /bin/bash
+        - -c
+        - |
+          set -xe
+          if [[ -f "/env/_master" ]]; then
+            set -o allexport
+            source "/env/_master"
+            set +o allexport
+          fi
```
danwinship, line 62: this is to set environment variables in the sdn master and doesn't make sense here. (You're not even mounting the `/env` configmap anyway)

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: add-iptables
```
danwinship, line 4: It's weird to write this out as its own separate object when the script is so dependent on the specifics of the script in the DaemonSet. I feel like it would be better to just have the DS `cat` this out to a file at startup...

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+          iptables -F ICMP_ACTION
+          iptables -A ICMP_ACTION -j LOG
+          iptables -A ICMP_ACTION -j DROP
+          oc observe nodes -a '{ .status.addresses[1].address }' -- /tmp/add_iptables.sh
+        lifecycle:
+          preStop:
+            exec:
+              command: ["/bin/bash", "-c", "echo drop-icmp done"]
+        securityContext:
+          privileged: true
```
danwinship, line 80: in the interest of having fewer privileged pods, I think this only needs to be `CAP_NET_ADMIN`, not fully privileged?

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+        volumeMounts:
+        - mountPath: /
+          name: host-slash
+        - name: add-iptables
+          mountPath: /tmp/add_iptables.sh
```
danwinship, line 85: don't use `/tmp`... use `/var/run/SOMETHING` or something

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,116 @@
...
+          requests:
+            cpu: 5m
+            memory: 20Mi
+        env:
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+      nodeSelector:
+        ### node-role.kubernetes.io/master: ""
```
danwinship, line 97: kill that

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -293,3 +293,70 @@ spec:
...
+          cat <<'EOF' > /var/run/add_iptables.sh
+#!/bin/sh
+echo "Adding ICMP drop rule for '$3' "
+if iptables -C CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+then
+     echo "iptables already set for $3"
+else
+     iptables -A CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+fi
+EOF
```
Miciah, line 314: Try indenting this level with the `cat` command.

mccv1r0:
test /test e2e-agnostic-upgrade
/test e2e-aws-ovn-windows
/test e2e-aws-sdn-multi
/test e2e-gcp-ovn

bindata/network/openshift-sdn-azure/azure-drop-icmp-daemonset-sdn.yaml
```diff
@@ -0,0 +1,110 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: add-iptables
```
danwinship, line 4: you need to remove the `azure-drop-icmp-daemonset-sdn.yaml` file from the PR since it's no longer used (right?)

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -235,6 +235,62 @@ spec:
...
+          echo "Adding ICMP drop rule for '$3' "
+          if iptables -C CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+          then
+               echo "iptables already set for $3"
+          else
+               iptables -A CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+          fi
+          EOF
+
+          chmod +x /var/run/add_iptables.sh
```
danwinship, line 263: So if we're mounting `host-slash` on `/` then this is the root `/var/run` right? so we should be a bit more paranoid and `chmod` it to a specific value (0755?). Actually, maybe do

    touch /var/run/add_iptables.sh
    chmod 0755 /var/run/add_iptables.sh
    cat <<'EOF' > /var/run/add_iptables.sh
    ...

or else

    mkdir -m 0700 /var/run/drop-icmp
    cat <<'EOF' > /var/run/add_iptables.sh
    ...

knobunc:

Overriding these jobs since they are known failing and the work is being tracked to fix them elsewhere.

PR #1066, opened 2021-04-21
Bug 1948582: Allow for local gateway mode to be configured

jluhrsen:
- if a configMap openshift-ovn-kubernetes/gateway-mode-config
  has a value of "local" or "shared", it will be honored. Otherwise
  the default (currently "shared") is used.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -795,13 +795,13 @@ spec:
             set +o allexport
           fi
 
-          gateway_mode_flags=
-          # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
-          # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
+          if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
             gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+          elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
+            gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
-            gateway_mode_flags="--gateway-mode local --gateway-interface none"
+            echo "Invalid OVN_GATEWAY_MODE: \"{{.OVN_GATEWAY_MODE}}\". Must be \"local\" or \"shared\"."
+            exit 1
           fi
 
           # start nbctl daemon for caching
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -233,13 +233,13 @@ spec:
 
           echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node db_ip ${db_ip}"
 
-          gateway_mode_flags=
-          # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
-          # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
+          if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
             gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+          elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
+            gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
           else
-            gateway_mode_flags="--gateway-mode local --gateway-interface none"
+            echo "Invalid OVN_GATEWAY_MODE: \"{{.OVN_GATEWAY_MODE}}\". Must be \"local\" or \"shared\"."
+            exit 1
           fi
 
           export_network_flows_flags=
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -67,6 +67,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["GenevePort"] = c.GenevePort
 	data.Data["CNIConfDir"] = pluginCNIConfDir(conf)
 	data.Data["CNIBinDir"] = CNIBinDir
+	data.Data["OVN_GATEWAY_MODE"] = bootstrapResult.OVN.GatewayMode
 	data.Data["OVN_NB_PORT"] = OVN_NB_PORT
 	data.Data["OVN_SB_PORT"] = OVN_SB_PORT
 	data.Data["OVN_NB_RAFT_PORT"] = OVN_NB_RAFT_PORT
@@ -210,6 +211,30 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 
 	return objs, nil
+}
+
+// returns the value of mode found in the openshift-ovn-kubernetes/gateway-mode-config configMap
+// if it exists, otherwise returns whatever the global OVN_GATEWAY_MODE is set to (shared)
+func GetGatewayMode(kubeClient client.Client) (string, error) {
+	defaultGatewayMode := "shared"
+	cm := &corev1.ConfigMap{}
+	nsn := types.NamespacedName{Namespace: "openshift-network-operator", Name: "gateway-mode-config"}
+	err := kubeClient.Get(context.TODO(), nsn, cm)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Warningf("Did not find gateway-mode-config. Using default mode: \"%s\"", defaultGatewayMode)
+			return defaultGatewayMode, nil
+		} else {
+			return "", fmt.Errorf("Could not determine gateway mode: %w", err)
+		}
+	}
+	if cm.Data["mode"] != "shared" && cm.Data["mode"] != "local" {
+		klog.Warningf("Ignoring gateway-mode-config %s. Does not match \"shared\" or \"local\"", cm.Data["mode"])
+		return defaultGatewayMode, nil
+	}
+	klog.Infof("Overriding OVN gateway mode to %s", cm.Data["mode"])
+	return cm.Data["mode"], nil
 }
 
 // validateOVNKubernetes checks that the ovn-kubernetes specific configuration
@@ -369,11 +394,16 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 		return nil, fmt.Errorf("Unable to bootstrap OVN, unable to unmarshal install-config: %s", err)
 	}
 
+	gatewayMode, err := GetGatewayMode(kubeClient)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to bootstrap OVN, undetermined gateway-mode: '%s'", err)
+	}
+
 	controlPlaneReplicaCount, _ := strconv.Atoi(rcD.ControlPlane.Replicas)
 
 	var heartBeat int
 
-	err := wait.PollImmediate(OVN_MASTER_DISCOVERY_POLL*time.Second, time.Duration(OVN_MASTER_DISCOVERY_TIMEOUT)*time.Second, func() (bool, error) {
+	err = wait.PollImmediate(OVN_MASTER_DISCOVERY_POLL*time.Second, time.Duration(OVN_MASTER_DISCOVERY_TIMEOUT)*time.Second, func() (bool, error) {
 		matchingLabels := &client.MatchingLabels{"node-role.kubernetes.io/master": ""}
 		if err := kubeClient.List(context.TODO(), masterNodeList, matchingLabels); err != nil {
 			return false, err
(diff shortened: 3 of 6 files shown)


jluhrsen:
I know kind+cno is supposed to be broken, but using some work from @astoycos I was able to stand it up well enough to locally test this PR. this work:
https://github.com/astoycos/cluster-network-operator/commit/9735d45debb7945ee059de93d9e295d2cdde5c8d

hack/ovn-kind-cno.sh
```diff
@@ -175,6 +175,20 @@ data:
       replicas: ${NUM_MASTER_NODES}
 EOF
 
+# To test the hack that will allow an installation to change the gateway mode
```
astoycos, line 178: nit: Not sure if we wan't to add this here, maybe just add a quick note to the KIND README?

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+		TypeMeta:   metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
+		ObjectMeta: metav1.ObjectMeta{Name: "gateway-mode-config"},
+	}
+	nsn := types.NamespacedName{Namespace: "openshift-ovn-kubernetes", Name: "gateway-mode-config"}
+	err := kubeClient.Get(context.TODO(), nsn, cm)
+	if err != nil {
+		klog.V(2).Infof("Did not find gateway-mode-config. Using default mode: %s", OVN_GATEWAY_MODE)
+	}
+	if cm.Data["mode"] != "shared" && cm.Data["mode"] != "local" {
+		klog.V(2).Infof("Ignoring gateway-mode-config %s. Does not match \"shared\" or \"local\"", cm.Data["mode"])
```
astoycos, line 231: nit: Should this be a `klog.Warningf` rather than just a log?

trozet:
/retitle Bug 1948582: Allow for local gateway mode to be configured

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -795,13 +795,13 @@ spec:
...
 
-          gateway_mode_flags=
-          # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
-          # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
+          if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
             gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
-          else
+          elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
             gateway_mode_flags="--gateway-mode local --gateway-interface none"
```
trozet, line 801: this needs to say --gateway-interface br-ex

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -233,13 +233,13 @@ spec:
...
 
-          gateway_mode_flags=
-          # Check to see if ovs is provided by the node. This is only for upgrade from 4.5->4.6 or
-          # openshift-sdn to ovn-kube conversion
-          if grep -q OVNKubernetes /etc/systemd/system/ovs-configuration.service ; then
+          if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
             gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
-          else
+          elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
             gateway_mode_flags="--gateway-mode local --gateway-interface none"
```
trozet, line 239: ditto

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 }
 
+// returns the value of mode found in the openshift-ovn-kubernetes/gateway-mode-config configMap
+// if it exists, otherwise returns whatever the global OVN_GATEWAY_MODE is set to (shared)
+func GetGatewayMode(kubeClient client.Client) string {
+	cm := &corev1.ConfigMap{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
+		ObjectMeta: metav1.ObjectMeta{Name: "gateway-mode-config"},
+	}
+	nsn := types.NamespacedName{Namespace: "openshift-ovn-kubernetes", Name: "gateway-mode-config"}
```
trozet, line 225: would this namespace exist at cluster create time? Plus your config map doesn't have this namespace in the hack example

trozet (changes requested):

hack/ovn-kind-cno.sh
jluhrsen: done. see if you like how it looks now. just a few lines in the readme and points to the hack/gateway-mode.yaml file as
the example to use.

pkg/network/ovn_kubernetes.go
jluhrsen: done

bindata/network/ovn-kubernetes/ovnkube-master.yaml
jluhrsen: done

bindata/network/ovn-kubernetes/ovnkube-node.yaml
jluhrsen: done

pkg/network/ovn_kubernetes.go
jluhrsen: I don't know if the namespace will exist at this point or not. I'll have to test to check.

I fixed the hack/gateway-mode.yaml file to include the namespace.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
astoycos: This might be why ovn-k wasn't working on the kind setup @jamoo

pkg/network/ovn_kubernetes.go
astoycos: [This](https://github.com/openshift/cluster-network-operator/blob/master/docs/operands.md#general-notes) makes me think it'll be fine... specifically 

```
Most operand directories start with a 000-ns.yaml object to create the operand namespace before anything else is created.
```
Which makes me thing the ns should be there before `ovn-kubernetes.go` is ever touched

jluhrsen:
@trozet @danwinship @squeed, can we get review/approval on this?

there are 4 failing checks:

- e2e-gcp and e2e-gcp-ovn
  permafailing, but failures seem unrelated and mostly due to https://bugzilla.redhat.com/show_bug.cgi?id=1943442

- e2e-gcp-ovn-upgrade
  permafailing. seems just a waste of time and resources to run this at this point. and looks just as
  ugly as we have in a real upgrade job 4.7->4.8:
  https://testgrid.k8s.io/redhat-openshift-ocp-release-4.8-informing#periodic-ci-openshift-release-master-ci-4.8-upgrade-from-stable-4.7-e2e-gcp-ovn-upgrade

- e2e-openstack-ovn
  this job has no periodic/release version so only running on PRs. Looks like it permafails:
  https://prow.ci.openshift.org/?job=*e2e-openstack-ovn*
  last three times it ran on this PR it had the same exact 17 test case failures.

trozet (approved):
/lgtm

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	return objs, nil
 }
 
+// returns the value of mode found in the openshift-ovn-kubernetes/gateway-mode-config configMap
+// if it exists, otherwise returns whatever the global OVN_GATEWAY_MODE is set to (shared)
+func GetGatewayMode(kubeClient client.Client) string {
+	cm := &corev1.ConfigMap{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
```
squeed, line 222: FYI: you don't need to set the TypeMeta (though it doesn't hurt anything)

squeed: In fact, this can be a zero-valued object (there were a few exceptions due to bugs with CRD and type registration, but they're all gone now).

pkg/network/ovn_kubernetes.go
```diff
@@ -41,6 +42,7 @@ const OVN_CERT_CN = "ovn"
 const OVN_MASTER_DISCOVERY_POLL = 5
 const OVN_MASTER_DISCOVERY_BACKOFF = 120
 
+var OVN_GATEWAY_MODE = "shared"
```
squeed, line 45: This should not be a global.

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	err := kubeClient.Get(context.TODO(), nsn, cm)
+	if err != nil {
+		klog.Warningf("Did not find gateway-mode-config. Using default mode: %s", OVN_GATEWAY_MODE)
+	}
+	if cm.Data["mode"] != "shared" && cm.Data["mode"] != "local" {
+		klog.Warningf("Ignoring gateway-mode-config %s. Does not match \"shared\" or \"local\"", cm.Data["mode"])
+	} else {
+		OVN_GATEWAY_MODE = cm.Data["mode"]
+	}
+	return OVN_GATEWAY_MODE
```
squeed, line 235: Since you're returning the value, this should not be a global.

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+		ObjectMeta: metav1.ObjectMeta{Name: "gateway-mode-config"},
+	}
+	nsn := types.NamespacedName{Namespace: "openshift-network-operator", Name: "gateway-mode-config"}
+	err := kubeClient.Get(context.TODO(), nsn, cm)
+	if err != nil {
+		klog.Warningf("Did not find gateway-mode-config. Using default mode: %s", OVN_GATEWAY_MODE)
+	}
+	if cm.Data["mode"] != "shared" && cm.Data["mode"] != "local" {
+		klog.Warningf("Ignoring gateway-mode-config %s. Does not match \"shared\" or \"local\"", cm.Data["mode"])
+	} else {
```
squeed, line 232: I would log a message here - something like "overriding OVN gateway-mode to %s"

squeed:
/lgtm cancel
one quick bug to squash.

pkg/network/ovn_kubernetes.go
```diff
@@ -212,6 +215,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+// returns the value of mode found in the openshift-ovn-kubernetes/gateway-mode-config configMap
+// if it exists, otherwise returns whatever the global OVN_GATEWAY_MODE is set to (shared)
+func GetGatewayMode(kubeClient client.Client) string {
+	cm := &corev1.ConfigMap{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "v1", Kind: "ConfigMap"},
+		ObjectMeta: metav1.ObjectMeta{Name: "gateway-mode-config"},
+	}
+	nsn := types.NamespacedName{Namespace: "openshift-network-operator", Name: "gateway-mode-config"}
+	err := kubeClient.Get(context.TODO(), nsn, cm)
+	if err != nil {
```
squeed, line 227: You have to check to see if this is a not-found error. Only then can you safely return the default. Otherwise you have to return `err` and stop rendering.

Otherwise we could flap gateway modes (bad) or do the wrong thing if the apiserver blips.

squeed: something like
```
import (
    apierrors "k8s.io/apimachinery/pkg/api/errors"
)

...
if apierrors.IsNotFound(err) {
    gatewayMode = DEFAULT_GATEWAY_MODE
} else {
    return "", fmt.Errorf("could not determine gateway mode: %w", err)
}

**Comment by squeed** [comment]
@jluhrsen is it safe to change gateway modes on a running cluster?

File: pkg/network/ovn_kubernetes.go [hunk 11]
```diff
@@ -369,6 +394,8 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 		return nil, fmt.Errorf("Unable to bootstrap OVN, unable to unmarshal install-config: %s", err)
 	}
 
+	gatewayMode, _ := GetGatewayMode(kubeClient)
```
squeed, line 397: You need to check for `err` here and bail appropriately.

jluhrsen:
> @jluhrsen is it safe to change gateway modes on a running cluster?

I had assumed it was not safe, but I actually don't really know. @trozet, can you chime in?

Otherwise @squeed, I think I've fixed up everything you pointed out. Thanks for looking things over and making it  better. Let me know if you see anything else. I'm going to manually run some tests with this PR as it sits now to make sure it's working as expected. One thing I'm not sure how to manually test would be the error case when the error we get checking the configMap is not just apierrors.IsNotFound(). Any tips?

jluhrsen:
I think this is ready now. I tested four scenarios with this PR:

1. configmap in manifests/ with local gateway-mode and all ovnkube pods get local
2. configmap with garbage for gateway-mode and shared gets used on the pods
3. explicitly putting "shared" in the configmap and shared is used
4. not having a configmap at all and shared is used

Here is a [doc with some pod logs](https://docs.google.com/document/d/1MwBs-GT_OWmwsaBbTmjBm5_LSYjuLgP-u40hh58zszA/edit?usp=sharing) showing things working as expected.

PR #1070, opened 2021-04-21
Bug 1932114: Detect Nodes Network MTU

MaysaMacedo:
same MTU as the host.


JAORMX (approved):

JAORMX (approved):

derekwaynecarr:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

approved

PR #1071, opened 2021-04-22
Bug 1951007: Wait for NB and SB servers to be in active state before s…

msherif1234:
Description
Its not always guaranteed to have NN and SB servers in running state when ovnkube-master starts.

Type of change
Add a loop to check both NB and SB servers are active before starting ovnkube master

How Has This Been Tested?
verified the bash changes on local gcp cluster

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -818,6 +818,27 @@ spec:
            # REMOVEME once OVN path for control socket is fixed (right now uses /var/run/openvswitch)
           ln -sf $OVN_NB_DAEMON /var/run/ovn/ || true
 
+          # Check both nbServer and sbServer to make sure both are active before starting ovnkube-master
+          OVN_SSL_DB_CHK="--no-leader-only -t 5 -p /ovn-cert/tls.key -c /ovn-cert/tls.crt \
+                          -C /ovn-ca/ca-bundle.crt --db "{{.OVN_NB_DB_LIST}}" find SSL"
+          retries=0
+          while ! ovn-nbctl ${OVN_SSL_DB_CHK}; do
+             (( retries += 1 ))
+            if [[ "${retries}" -gt 10 ]]; then
+              echo "Northbound DB Server status not active for longtime. Continuing..."
+            fi
+            sleep 1
+          done
+
+          retries=0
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -818,6 +818,25 @@ spec:
...
           ln -sf $OVN_NB_DAEMON /var/run/ovn/ || true
 
+          # Check both nbServer and sbServer to make sure both are active before starting ovnkube-master
+          retries=0
+          while [ "$(ovn-appctl -t /var/run/ovn/ovnnb_db.ctl  ovsdb-server/sync-status | awk -F': ' '{print $2}')" != "active" ]; do
+             (( retries += 1 ))
+            if [[ "${retries}" -gt 10 ]]; then
+              echo "Northbound DB Server status not active for longtime. Continuing..."
+            fi
+            sleep 0.001
```
dcbw, line 828: @msherif1234 latency isn't critical at pod startup here since we only do it once in a while (hopefully!); I'd say 0.1 is a better balance between latency and hammering the db server.

I think this is OK, however keep in mind that this is checking the *local* DBs but ovnkube master is given the IPs of all databases and picks one from that list; it may not be the local DB. Also, just because the local DB is serving on its unix socket doesn't mean it's serving on TCP/SSL yet; that's a postStart for the DB pods:

https://github.com/openshift/cluster-network-operator/blob/596044d42085aa3b62b2571fa046de77600f055b/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L262

What I'd do here instead is to use `ovn-nbctl find SSL` and `ovn-sbctl find SSL` along with the --db and SSL args to ensure that we can talk to at least one DB over SSL before declaring them "up" from onvkube-master's perspective.

abhat: I thought you said it takes 8s between master starting and db starting? The math won't have the desired effect?

msherif1234: that was what the log showed will make it 10 iteration 1sec apart

msherif1234:
not needed as when ivnkubemaster restart it will connect so it seems its fine for ov-kubemaster to restart to reconnect of DB servers not ready

PR #1072, opened 2021-04-26
Bug 1952665: Update multi-networkpolicy memory limits/requests

s1061123:


dcbw:
only question I have is, are limits high enough? Does the mem usage scale with cluster size? We removed the limits on most networking components because we can't accurately predict what their consumption will be at scale, especially in large/complex clusters. IIRC we kept requests though.

s1061123:
@dcbw make cense. I could not test high scale environment and it might consume than the limits. So I will keep requests but will remove limits.

PR #1073, opened 2021-04-26
Local mode

jluhrsen:


jluhrsen:
using this to test that local mode is deployed like we expect if a manifest exists with the proper configMap

PR #1075, opened 2021-04-26
Borked mode

jluhrsen:


jluhrsen:
using this to test that shared mode is still deployed like we expect if a manifest exists with the borked configMap

PR #1076, opened 2021-04-28
Bug 1954302: Remove OVS daemonsets

jluhrsen:
In 4.6 we moved OVS to systemd, we needed to have a pod capable of
running OVS during the upgrade, but after that it became unnecessary.

In 4.7 we're guaranteed to launch ovs in systemd, both when installed
from scratch and when upgrading from 4.6, therefore there is no use for
the OVS pod other than being able to gather the logs from the apiserver,
we cannot even restart the OVS daemons by deleting the pod, wihch is
confusing.


jluhrsen:
@abhat , one check still failing with random test cases. can you lgtm?

knobunc:

We keep flaking on infrastructure problems (image pull failures, etc.) and have previously passed that job with this PR.

PR #1077, opened 2021-04-28
Bug 1952001: Updating api to reduce the  number of watch requests

JacobTanenbaum:
the first commit updates the kube-components to tag v0.21.

second commit takes in the fix for the kube/apiserver from https://github.com/kubernetes/kubernetes/pull/101102


JacobTanenbaum:
/retest 

lease not found

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openshift/cluster-network-operator
 
-go 1.15
+go 1.16
```
danwinship, line 3: This requires a change to Dockerfile too. (Oddly, Dockerfile.rhel7 already says 1.16)

go.mod
```diff
@@ -31,49 +31,23 @@ require (
...
+	k8s.io/apiextensions-apiserver v0.21.0
 	k8s.io/apimachinery v0.21.0
-	k8s.io/client-go v0.20.0
-	k8s.io/code-generator v0.21.0-rc.0
-	k8s.io/component-base v0.20.0
+	k8s.io/client-go v0.21.0
+	k8s.io/code-generator v0.21.0
+	k8s.io/component-base v0.21.0
 	k8s.io/klog/v2 v2.8.0
 	k8s.io/kube-proxy v0.20.0
```
danwinship, line 40: bump kube-proxy to `v0.21.0` too

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -58,7 +58,6 @@ func NewNetworkConnectivityCheckController(
 			operatorcontrolplaneClient,
 			apiextensionsClient,
 			apiextensionsInformers,
-			configInformers,
```
danwinship, line 61: hm... how come?

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -126,7 +126,7 @@ func (f *fakeRESTMapper) ResourceSingularizer(resource string) (singular string,
 }
 
 func TestStatusManager_set(t *testing.T) {
-	client := fake.NewFakeClient()
+	client := fake.NewClientBuilder().WithRuntimeObjects().Build()
```
danwinship, line 129: weird... did you copy this from somewhere? I would have expected `NewFakeClient()` to still work...

JacobTanenbaum: when compiling a message prints from kube that `NewFakeClient()` is deprecated.

go.mod
JacobTanenbaum: done

pkg/controller/connectivitycheck/connectivity_check_controller.go
JacobTanenbaum: fixed. This was not required, it was a mixup after removing some of the replace in the go.mod

pkg/controller/statusmanager/status_manager_test.go
danwinship: ah. ok, but it looks like the `WithRuntimeObjects` part isn't needed if you aren't actually passing any objects. So just:

    client := fake.NewClientBuilder().Build()

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.15 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.16 AS builder
```
danwinship, line 1: wrong commit; this should go in the "updating to use tag v0.21.0 for kubernetes repos like the rest of openshift" commit, not "Updating api to reduce the number of watch requests "

JacobTanenbaum: done

dcbw:
Metal jobs are broken: `INFO[2021-05-03T21:27:21Z] level=error msg=Error: could not contact Ironic API: context deadline exceeded`
/lgtm

PR #1082, opened 2021-04-30
Bug 1956352: [4.7] OVN kubernetes dual-stack conversion

dcbw:
Manual cherry-pick of https://github.com/openshift/cluster-network-operator/pull/1034
We backported the ovnkube bits to 4.7 as part of https://github.com/openshift/ovn-kubernetes/pull/460

Add a new annotation to the OVN kubernetes daemonsets, ovnkube-master
and ovnkube-node, indicating the IP family mode configured:
- single-stack
- dual-stack

This annotations allow to automate the conversion of one cluster from
single-stack to dual-stack and viceversa, since the change in the
annotation triggers a rollout of the daemonset.

Because the conversion requires to update the master daemonsets first,
we add the same logic used to manage the versions upgrades to this
process. The dual-stack conversion also blocks the upgrade, so both
are exclusive.

### DIFF
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -37,6 +37,10 @@ const NonCriticalAnnotation = "networkoperator.openshift.io/non-critical"
 // that executing network migration (switching the default network type of the cluster) is allowed.
 const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migration"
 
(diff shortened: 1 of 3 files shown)


aojea:
It still need to switch the feature gate in the apiserver

PR #1089, opened 2021-05-10
Revert "[release-4.6] Bug 1950407: Fix kube-rbac-proxy startup scripts"

abhat:
Reverts openshift/cluster-network-operator#1061


jluhrsen:
@abhat, we don't need this now. This [one](https://github.com/openshift/cluster-network-operator/pull/1096) landed and gets us working again.

PR #1092, opened 2021-05-11
set ovn inactivity_probe to 15s

aojea:
From ovn-sb man

inactivity_probe: optional integer
    Maximum number of milliseconds of idle time on connection
    to the client before sending an inactivity probe message.
    If Open vSwitch does not communicate with the client for
    the specified number of seconds, it will send a probe. If
    a response is not received for the same additional amount
    of time, Open vSwitch assumes the connection has been
    broken and attempts to reconnect. Default is
    implementation-specific. A value of 0 disables inactivity
    probes.

ovsdb-server will clear the lock once it clears the connection.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -259,7 +259,7 @@ spec:
 
                   # set the connection and disable inactivity probe
                   retries=0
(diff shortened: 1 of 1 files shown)


squeed:
@aojea what would this be for? We'll need a bug now to get this in.

aojea:
> @aojea what would this be for? We'll need a bug now to get this in.

It was just testing

PR #1064, opened 2021-04-19
Bug 1951158: Fix egress-router-cni CRD parsing

danielmellado:
This PR fixes an issue where parsing from the openshift/api egressrouter
would use an IP instead of a CIDR, and was non-properly detected by the
CNO controller.

PR #1094, opened 2021-05-14
Bug 1951158: Addressing original PR 1064 review comments.

msherif1234:
follow up to  PR [1064](https://github.com/openshift/cluster-network-operator/pull/1064)
locations of the fields in AllowedDestinations need to be in very specific order to match the implmentation in  [macvlan logic](https://github.com/egress-router-cni/blob/master/pkg/macvlan/macvlan.go)

based on [egress-router-cni](https://github.com/msherif1234/egress-router-cni) doc its assumed that addresses are /24 and destinations are /32 to set the IPTables rules


bindata/egress-router/000-nad.yaml
```diff
@@ -13,11 +13,27 @@ spec:
     "name": "egress-router-cni-nad",
     "ip": {
       "addresses": [
-        "{{.Addresses}}"
+        "{{.Addresses}}/24"
```
msherif1234, line 16: egress-router-cni is expecting a CIDR that is why the additional /24
[macvlan.go#L304](https://github.com/msherif1234/egress-router-cni/blob/master/pkg/macvlan/macvlan.go#L304)

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -173,13 +173,14 @@ func (r *EgressRouterReconciler) ensureEgressRouter(manifestDir string, namespac
 	data := render.MakeRenderData()
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
 	data.Data["EgressRouterNamespace"] = namespace
-	if isItValidCidr(string(router.Spec.Addresses[0].IP)) {
+	if isItValidIPAddress(string(router.Spec.Addresses[0].IP)) {
```
dcbw, line 176: @msherif1234 I have zero context here, just doing a drive-by. What sets the EgressRouter's Spec.Addresses array? If it used to be a CIDR, I don't think we can just change it over to a plain IP without breaking existing EgressRouter objects, right?

If that's true, at the very least we should warn if we find a CIDR, then ParseCIDR() it to pull off the IP and then just use that with the /24.

bindata/egress-router/000-nad.yaml
dcbw: @msherif1234 but previously, if you did put a CIDR into the Addresses[x].IP field, that would pass through all the way to the macvlan interface (right?). But now you're limited to a /24, is that what we expect?

(again, I don't know the code much and maybe it already needs to be a /24...)

msherif1234: the CIDR define this field as the following 
```
IP string `json:"ip" protobuf:"bytes,1,opt,name=ip"`
```
and validate as follows 
```
ip:
                      anyOf:
                      - format: ipv4
                      - format: ipv6
                      description: IP is the address to configure on the router's
                        interface. Can be IPv4 or IPv6.
                      type: string
```
so passing /xx will fail the CRD validation 
changing CIRD to something like this will help
```
CIDR:
                  description: CIDR defines the total range of a cluster networks
                    address space.
                  type: string
                  pattern: ^(([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[0-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])/([0-9]|[12][0-9]|3[0-2])$
                hostSubnetLength:
                  description: HostSubnetLength is the number of bits of the accompanying
                    CIDR address to allocate to each node. eg, 8 would mean that each
                    node would have a /24 slice of the overlay network for its pods.
                  type: integer
                  format: int32
                  maximum: 30
                  minimum: 2
```
I can change the API no prob if APIs folks are ok ?

pkg/controller/egress_router/egress_router_controller.go
msherif1234: @dcbw me too :) but thanks for driving by!! 
its not CIDR its just array of strings "ips" I am with u we should let customer configure the prefix length and not hardcode it seems like a kludge I will see if @tssurya and @squeed  ok with me changing the API

squeed:
Right now, the validation for the CRD doesn't allow a CIDR, only a bare IP address: https://github.com/openshift/api/blob/master/networkoperator/v1/001-egressrouter.crd.yaml#L74

You'll have to fix this here if you want to allow a CIDR: https://github.com/openshift/api/blob/master/networkoperator/v1/001-egressrouter.crd.yaml-patch

bindata/egress-router/000-nad.yaml
```diff
@@ -13,11 +13,27 @@ spec:
...
         ],
       "destinations": [
-        "{{.AllowedDestinations}}"
+        {{ range $i, $val := .AllowedDestinations }}
```
squeed, line 19: I would rather see this done in go rather than the templates. It's a lot easier to test.

msherif1234:
> Right now, the validation for the CRD doesn't allow a CIDR, only a bare IP address: https://github.com/openshift/api/blob/master/networkoperator/v1/001-egressrouter.crd.yaml#L74
> 
> You'll have to fix this here if you want to allow a CIDR: https://github.com/openshift/api/blob/master/networkoperator/v1/001-egressrouter.crd.yaml-patch

API has been modified to allow CIDR fmt [924](https://github.com/openshift/api/pull/924) once its approved  and merged, I will rework this PR

knobunc:

This has been failing, but the fix is posted and under review.  Overriding in the mean time.

PR #1095, opened 2021-05-14
Bug 1954773: update the cno to use the egressfirewall flag

JacobTanenbaum:
ovn-kubernetes has changed to use a flag to enable egressFirewall, the cno needs to use that flag

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -23,6 +23,7 @@ data:
  
     [ovnkubernetesfeature]
     enable-egress-ip=true
+    enable-egress-firewall=true
 
     [gateway]
     mode=local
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu (approved):
/lgtm 

but also

/hold 

until https://github.com/openshift/ovn-kubernetes/pull/541 is in

dcbw:
@alexanderConstantinescu don't we want to merge this first, and *then* the ovnkube bits? becuase otherwise ovnkube won't have egress firewall enabled until the CNO PR merges, and presumably we test that in CI and thus it would break?

alexanderConstantinescu:
> @alexanderConstantinescu don't we want to merge this first, and _then_ the ovnkube bits? becuase otherwise ovnkube won't have egress firewall enabled until the CNO PR merges, and presumably we test that in CI and thus it would break?

We don't test egress firewall in CI. So in any case we have nothing to worry about :) 

The reason I put a hold on it was because I thought ovnkube errors if you provide it a flag it doesn't recognize (my memory might not serve me right on this one though...), which would then make it CrashLoop and block CI.

sdodson:
e2e-aws-ovn-windows should no longer be run
/refresh
/skip

squeed:
doesn't test ovn

PR #1100, opened 2021-05-20
Bug 1955548: Kuryr: Do not expand services subnet with OVN

dulek:
When creating a Services subnet, in order to accommodate additional VRRP
ports of Octavia Amphora VMs, Kuryr was creating a subnet for services
in OpenStack of twice the size requested by user in ServiceNetwork. To
make sure a VRRP port will never land in the ServiceNetwork range,
allocation ranges were used to point to the _other_ half of the subnet.

This is no longer required when using OVN-Octavia, so this commit gets
rid of it, making sure that the change will only affect existing
installations (as we cannot change the CIDR of the existing subnet in
OpenStack).

The services subnet requires a gateway IP to be put somewhere in the
range and we normally put it somewhere on the range reserved from VRRP
so K8s would never choose it for a Service. As there's no way to inform
K8s about a reserved IP, we now put the gateway on the last usable IP of
the services range and create a dummy Service with that IP to make sure
it's not reserved by K8s itself.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -49,10 +49,7 @@ const (
 	machinesNamespace     = "openshift-machine-api"
 	// NOTE(ltomasbo): Amphora driver supports came on 2.11, but ovn-octavia only supports it after 2.13
 	MinOctaviaVersionWithMultipleListeners = "v2.13"
-	MinOctaviaVersionWithHTTPSMonitors     = "v2.10"
 	MinOctaviaVersionWithProviders         = "v2.6"
-	MinOctaviaVersionWithTagSupport        = "v2.5"
-	MinOctaviaVersionWithTimeouts          = "v2.1"
 	KuryrNamespace                         = "openshift-kuryr"
 	KuryrConfigMapName                     = "kuryr-config"
 	DNSNamespace                           = "openshift-dns"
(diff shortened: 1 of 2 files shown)


dulek:
/hold cancel

This should be okay now.

dulek:
@MaysaMacedo

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -49,10 +49,7 @@ const (
 	machinesNamespace     = "openshift-machine-api"
 	// NOTE(ltomasbo): Amphora driver supports came on 2.11, but ovn-octavia only supports it after 2.13
 	MinOctaviaVersionWithMultipleListeners = "v2.13"
-	MinOctaviaVersionWithHTTPSMonitors     = "v2.10"
```
MaysaMacedo, line 52: thanks :)

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -206,6 +203,48 @@ func getMasterMachines(kubeClient client.Client) (*machineapi.MachineList, error
 	return machineList, nil
 }
 
+func createSubnetGatewayGuardSvc(kubeClient client.Client, requestedIP string) (clusterIP net.IP, err error) {
```
MaysaMacedo, line 206: perhaps it can be renamed to ensureSubnetGatewayGuardSvc as it's finding and creating if not Found?

MaysaMacedo (commented):
It looks good to me. I just have one minor suggestion.

pkg/platform/openstack/kuryr_bootstrap.go
dulek: Makes sense, done.

PR #1101, opened 2021-05-21
Bug 1962951: enable ovs column diffs feature

msherif1234:
enable ovs column diffs feature for release 4.9
Note: after this commit is merged we no longer can downgrade to older releases.
New ovsdb-server will be able to read old database format, but older ovsdb-server will fail to read database created by this new OVS version.

changes in ovs2.15: https://github.com/openvswitch/ovs/blob/0b3ff31d35f56c9401d868613581b4d35bb42724/NEWS#L17
package details : https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=1584628

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -180,15 +180,6 @@ spec:
             --ovn-nb-db-ssl-cert=/ovn-cert/tls.crt \
             --ovn-nb-db-ssl-ca-cert=/ovn-ca/ca-bundle.crt"
 
-          # if ovsdb supports column diffs, disable them, since they can cause upgrade issues
-          set +e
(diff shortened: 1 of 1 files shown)


squeed:
@dcbw is this going to work? Did 4.7 ever take an OVS version that supports column diffs?

msherif1234:
> @dcbw is this going to work? Did 4.7 ever take an OVS version that supports column diffs?

I have this PR to update 4.7
https://github.com/openshift/ovn-kubernetes/pull/549

squeed:
@msherif1234 sounds reasonable; we need to block this PR until that one lands

/hold

so we don't accidentally break upgrades.

dcbw:
4.7 won't get 2.15, but 4.8 has it.
Thus, we can merge this PR for 4.9 as at least one release before 4.9 has 2.15 + column diffs.

/approve
/lgtm

PR #1104, opened 2021-05-24
Bug 1954870: Add priorityClass to network-check-source/target 

bpickard22:


bindata/network-diagnostics/network-check-source.yaml
```diff
@@ -24,6 +24,7 @@ spec:
         app: network-check-source
         kubernetes.io/os: "linux"
     spec:
+      priorityClassName: system-cluster-critical
```
tssurya, line 27: I think we can use the `openshift-user-critical` priority for these daemonsets since these daemon sets are there only to check endpoints connectivity for debugging right ? They are not really more critical than the user-workloads. So we should be ok preempting them to accommodate the user load ?

tssurya (commented):
Make sure to add the Bug ID to the PR Title so that it gets linked to the bug and also fill in a description of the PR saying what it does so that it helps the reviewers.

bindata/network-diagnostics/network-check-source.yaml
bpickard22: I believe that you are correct. Will make that change then open the pr

PR #1107, opened 2021-05-24
Bug 1825219: Fix nil checks in bootstrapSDN

Miciah:
Follow-up to #1059.

* `pkg/network/openshift_sdn.go` (`bootstrapSDN`): Delete an unnecessary check for nil `infraConfig`.  Check if `infraConfig.Status.PlatformStatus` is nil, which is possible if the cluster was originally installed using OpenShift 4.1 and then upgraded.


knobunc:

This failure is being tracked and addressed in a different bug.

knobunc:

This failure is known and being addressed elsewhere.

PR #1108, opened 2021-05-25
Bug 1960711: Revert "ipsec: Allow enablement/disablement at runtime"

markdgray:
Due to https://bugzilla.redhat.com/1960711, IPsec cannot be enabled
at runtime. IPsec requires an additional 46B overhead for packet headers,
therefore we will need to be able to dynamically adjust the MTU at
runtime before enabling this feature.

This reverts commit 59a50f93ea671564412334ee181c0404588f2cb5.


markdgray:
All failures are due to "cloneref"  which looks unrelated

PR #1111, opened 2021-05-28
Bug 1970453: Replace to correct OpenFlow version when replacing flows

alexanderConstantinescu:
The ovs-save script restoring flows when running OVS in a container on 4.6 defaults to OpenFlow14, which is not supported by openshift-sdn. This leads to flow restore errors on upgrades from 4.5 and hence a larger downtime window. See: https://github.com/openshift/sdn/pull/306#issuecomment-849773314

Force the script to use OpenFlow13, which is the only version of OpenFlow openshift-sdn supports.

### DIFF
--- a/bindata/network/openshift-sdn/sdn-ovs.yaml
+++ b/bindata/network/openshift-sdn/sdn-ovs.yaml
@@ -127,6 +127,11 @@ spec:
             # Load any flows that we saved
             echo "$(date -u "+%Y-%m-%d %H:%M:%S") info: Loading previous flows ..." 2>&1
             if [[ -f /var/run/openvswitch/flows.sh ]]; then
+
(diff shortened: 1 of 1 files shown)


alexanderConstantinescu:
/retitle Bug 1970453: Replace to correct OpenFlow version when replacing flows
/auto-cc

danwinship:
/lgtm
so... did this never actually work?

alexanderConstantinescu:
> /lgtm
> so... did this never actually work?

Well, I think it must have at one point...my suspicion is that https://github.com/openshift/sdn/pull/238 broke it.

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.20

approved

PR #1112, opened 2021-05-28
Cloud network config controller - CNO deployment

alexanderConstantinescu:
This PR implements the CNO required dependencies for deploying the cloud-network-config-controller. It has a dependency on https://github.com/openshift/cloud-network-config-controller/pull/6 which means

/hold

In short, for people not aware of this component: this is a component which is aimed at configuring cloud-level network configurations. In the short-term this will mean supporting egress IP on public clouds. This means that this component should only be deployed on Azure, AWS and GCP. 

This component does not run `hostNetwork: true` and hence, is not expected to be ready when all other networking components are. The CNO is made aware of this by the annotation: `networkoperator.openshift.io/non-critical: ""`. This component will only ever finish deploying once the [cloud-credentials-operator](https://github.com/openshift/cloud-credential-operator) and [service-ca-operator](https://github.com/openshift/service-ca-operator) have finished deploying, processed and provided the resources required for the cloud-network-config-controller to run, that being: cloud-credentials and injected TLS certificates for the admission controller.


bindata/cloud-network-config-controller/000-ns.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Namespace
+metadata:
+  name: openshift-cloud-network-config-controller
+  labels:
+    openshift.io/run-level: "0"
+    openshift.io/cluster-monitoring: "true"
+  annotations:
+    openshift.io/node-selector: ""
+    openshift.io/description: "OpenShift cloud network config controller namespace - a controller used to manage cloud-level network configuration"
+    workload.openshift.io/allowed: "management"
```
flavio-fernandes, line 11: nit : end of line

bindata/cloud-network-config-controller/005-credentialrequest.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  secretRef:
+    name: cloud-credentials
+    namespace: openshift-cloud-network-config-controller
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: AzureProviderSpec
+    roleBindings:
+      - role: Contributor 
+{{end}}
```
flavio-fernandes, line 57: nit : end of line

pkg/network/render.go
```diff
@@ -615,3 +629,23 @@ func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*
...
+
+// renderCloudNetworkConfigController renders the cloud network config controller
+func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, cloudBootstrapResult bootstrap.CloudBootstrapResult, manifestDir string) ([]*uns.Unstructured, error) {
+	data := render.MakeRenderData()
+	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+	data.Data["CloudProvider"] = cloudBootstrapResult.Platform
+	data.Data["CloudRegion"] = cloudBootstrapResult.Region
+	data.Data["CloudProviderAWS"] = v1.AWSPlatformType
+	data.Data["CloudProviderAzure"] = v1.AzurePlatformType
+	data.Data["CloudProviderGCP"] = v1.GCPPlatformType
```
msherif1234, line 641: question CloudProvider should be sufficient to capture different public cloud types so I am not sure why we need the other fields CloudProviderAWS/CloudProviderAzure/CloudProviderGCP , seems like redundancy to me no ?

manifests/image-references
```diff
@@ -66,3 +66,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-network-tools:latest
+  - name: cloud-network-config-controller
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-cloud-network-config-controller:latest
```
martinkennelly, line 72: nit: newline

pkg/network/cloud_network.go
```diff
@@ -0,0 +1,39 @@
...
+func bootstrapCloud(conf *operv1.Network, kubeClient client.Client) (*bootstrap.CloudBootstrapResult, error) {
+	var platformType configv1.PlatformType
+	var region string
+
+	infraConfig := &configv1.Infrastructure{}
+	if err := kubeClient.Get(context.TODO(), types.NamespacedName{Name: "cluster"}, infraConfig); err != nil {
+		return nil, fmt.Errorf("failed to get infrastructure 'config': %v", err)
+	}
+
+	if infraConfig != nil {
```
martinkennelly, line 23: Its impossible for `infraConfig` to be nil.

pkg/network/cloud_network.go
```diff
@@ -0,0 +1,39 @@
...
+		if infraConfig.Status.PlatformStatus.Type != "" {
+			platformType = infraConfig.Status.PlatformStatus.Type
+		}
+		if platformType == configv1.AWSPlatformType {
+			region = infraConfig.Status.PlatformStatus.AWS.Region
+		} else if platformType == configv1.GCPPlatformType {
+			region = infraConfig.Status.PlatformStatus.GCP.Region
```
martinkennelly, line 30: No need to get Azure's region?

bindata/cloud-network-config-controller/controller.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: apps/v1
+kind: Deployment
```
martinkennelly, line 2: no replicas here in this deployment but you have leader election code in https://github.com/openshift/cloud-network-config-controller/pull/6

Why have LE if theres only going to be one instance of ```cloud-network-config-controller``` ?

pkg/network/render.go
alexanderConstantinescu: Not really: the cloud-network-config-controller needs credentials to be able to talk to the cloud API, the component generating the credentials is the cloud-credentials-operator. That component watches a CRD for which the CNO will write a CR, defined here: https://github.com/openshift/cluster-network-operator/blob/5617e47380541826e9496af905c2b0e3effba70a/bindata/cloud-network-config-controller/005-credentialrequest.yaml . As you can see the spec for that CR changes depending on the cloud provider, so we will need to determine which cloud we're deploying in as to write the correct CR, these are just helper variables to make that determination.

bindata/cloud-network-config-controller/admission-controller.yaml
```diff
@@ -0,0 +1,69 @@
...
+metadata:
+  name: cloud-network-config-admission-controller
+  namespace: openshift-cloud-network-config-controller
+  labels:
+    app: cloud-network-config-admission-controller
+  annotations:
+    kubernetes.io/description: |
+      This daemon set launches the cloud network config admission controller component on each master node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+    networkoperator.openshift.io/non-critical: ""
```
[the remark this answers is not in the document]
alexanderConstantinescu: `networkoperator.openshift.io/non-critical: ""` is a CNO specific annotation which just tells the CNO to ignore this resource when computing the status. The cloud-network-config-controller is not host networked (because it doesn't need to be) and will moreover depend on the cloud-credentials-operator for the cloud credentials it requires to talk to the cloud API. These components will not start until the network is up and running, but if the CNO doesn't skip these components it will never be ready and the cluster deployment will never reach the point where the CCO is deployed.

bindata/cloud-network-config-controller/controller.yaml
```diff
@@ -0,0 +1,70 @@
...
+        openshift.io/component: network
+    spec:
+      serviceAccountName: cloud-network-config-controller
+      priorityClassName: system-cluster-critical
+      containers:
+      - name: controller
+        image: {{.CloudNetworkConfigControllerImage}}
+        imagePullPolicy: IfNotPresent
+        command: ["/usr/bin/cloud-network-config-controller"]
+        args: ["-cloud-provider", "{{.CloudProvider}}", "-cloud-region", "{{.CloudRegion}}", "-secret-name", "cloud-credentials"]
```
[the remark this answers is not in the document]
alexanderConstantinescu: No, on Azure it will be an empty string, but that's not a problem: the cloud region is only required on AWS for initiating a client when communicating with the cloud API. It can be left empty on Azure.

pkg/network/cloud_network.go
alexanderConstantinescu: There is no "region" on Azure, at least not in `infraConfig.Status.PlatformStatus.Azure.Region`, and as mentioned above: we don't need it on Azure.

bindata/cloud-network-config-controller/controller.yaml
alexanderConstantinescu: For the same reason most operators have a leader election defined: on upgrades in certain corner cases there can be situations where one replica of a deployment is started on one node while the previous replica on the old node (the one being rebooted) still runs, it's thus just used as redundancy mechanism to avoid running two instances of the same controller at any point in time.

alexanderConstantinescu:
Everything has merged, this should be good to go

/hold cancel

trozet:
even though jobs are passing, the cloud network config controller is in crashloopbackoff

alexanderConstantinescu:
> even though jobs are passing, the cloud network config controller is in crashloopbackoff

Those were old jobs which had run before openshift/cloud-network-config-controller#6 had merged. Every job now passes or flakes, and no jobs have the CNCC crashlooping anymore.

alexanderConstantinescu:
- `e2e-aws-sdn-multi` is perma failing: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-aws-sdn-multi 
- `e2e-aws-ovn-windows` is failing on: `FAIL: TestWMCO/network/Pod_DNS_Resolution/Valid_URL (10.88s)`, that seems to be a pod running on a windows node and trying to resolve a valid URL, [but failing because it's unable to resolve openshift.com](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1112/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-windows/1471417098197536768/artifacts/e2e-aws-ovn-windows/gather-extra/artifacts/pods/wmco-test_win-dns-tester-valid-url-job-xnsfl--1-88kdb_win-dns-tester-valid-url.log) ...this is completely unrelated to my PR :shrug: 

I think both can be overriden

abhat:
/override e2e-aws-sdn-multi

The e2e-aws-sdn-multi job started failing more since yesterday Dec 15th. Did we introduce a regression through another PR?

The windows test failure is interesting. It does fail on that particular test in some other PRs too.

/override e2e-aws-ovn-windows

trozet:
only failure for e2e-agnostic-upgrade was:
[sig-network] pods should successfully create sandboxes by not timing out

ns/openshift-kube-controller-manager pod/kube-controller-manager-ci-op-n6vb4170-91326-7tpln-master-1 node/ci-op-n6vb4170-91326-7tpln-master-1 - 1951.15 seconds after deletion - reason/FailedCreatePodSandBox Failed to create pod sandbox: rpc error: code = Unknown desc = error creating pod sandbox with name "k8s_kube-controller-manager-ci-op-n6vb4170-91326-7tpln-master-1_openshift-kube-controller-manager_b2c73eb6-e948-44c1-bf3d-118b073fa5d3_0": initializing source docker://registry.build02.ci.openshift.org/ci-op-n6vb4170/stable@sha256:9b17b09283da8459d6bbcbc18bde0a4b38d7ac301c2d2c0fadf24b0cfc36aeb3: pinging container registry registry.build02.ci.openshift.org: Get "https://registry.build02.ci.openshift.org/v2/": dial tcp 34.74.144.21:443: i/o timeout

Has nothing to do with this PR

PR #1115, opened 2021-05-31
Bug 1963964: Escape characters on ini file

MaysaMacedo:
The configuration file supports variable substituion with oslo[1].
However, for the Kuryr scenario variables substituion wouldn't
happen as we expect the real value from the Secret. This commit
makes sure to escape the '$' caracter to avoid breaking oslo
parsing for the configuration file.

[1] https://github.com/openstack/oslo.config/blob/master/doc/source/configuration/format.rst#substitution


pkg/render/funcs.go
```diff
@@ -28,3 +32,9 @@ func isSet(m map[string]interface{}, key string) interface{} {
 	}
 	return val
 }
+
+// escapeCharacters returns the given string with any
+// possible reference of '$' escaped.
+func escapeCharacters(text string) string {
```
dulek, line 38: I'd make it e.g. `iniEscapeCharacters`, various formats have various special chars.

pkg/render/funcs.go
```diff
@@ -28,3 +32,9 @@ func isSet(m map[string]interface{}, key string) interface{} {
...
+	return strings.ReplaceAll(text, "$", "\\$")
```
dulek, line 39: If it's so simple, you could probably just use sprig function [1]. But maybe it's better to leave it as a separate function as we might discover more special cases. 

[1] https://masterminds.github.io/sprig/strings.html#replace

bindata/network/kuryr/003-config.yaml
```diff
@@ -70,7 +70,7 @@ data:
     auth_url = {{ $AuthInfo.AuthURL }}
     insecure = {{ .OpenStackInsecureAPI }}
     token = {{ default "\"\"" $AuthInfo.Token }}
-    password = {{ default "\"\"" $AuthInfo.Password }}
+    password = {{ default "\"\"" (escapeCharacters $AuthInfo.Password) }}
```
dulek, line 73: I believe we should escape other vars too, in particular: `token`, `username`, `project_domain_name`, `project_name`, `user_domain_name`, `region_name`.

dulek (changes requested):

bindata/network/kuryr/003-config.yaml
MaysaMacedo: Indeed. When posting directly to the API it's possible to set those values with '$'

pkg/render/funcs.go
MaysaMacedo: yup

MaysaMacedo:
/hold

need to try one installation with it first

bindata/network/kuryr/003-config.yaml
```diff
@@ -69,16 +69,16 @@ data:
...
-    project_domain_name = {{ default "\"\"" $AuthInfo.ProjectDomainName }}
+    token = {{ default "\"\"" (iniEscapeCharacters $AuthInfo.Token) }}
+    password = {{ default "\"\"" (iniEscapeCharacters $AuthInfo.Password) }}
+    username = {{ default "\"\"" (iniEscapeCharacters $AuthInfo.Username) }}
+    project_domain_name = {{ default "\"\"" (iniEscapeCharacters $AuthInfo.ProjectDomainName) }}
     project_domain_id = {{ default "\"\"" $AuthInfo.ProjectDomainID }}
     project_id = {{ default "\"\"" $AuthInfo.ProjectID }}
-    project_name = {{ default "\"\"" $AuthInfo.ProjectName }}
+    project_name = {{ default "\"\"" (iniEscapeCharacters $AuthInfo.ProjectName) }}
     user_domain_name = {{ default "\"\"" $AuthInfo.UserDomainName }}
```
dulek, line 79: User domain name might probably have it as well?

dulek (commented):
Looks good, just minor thing inline.

PR #1123, opened 2021-06-07
Bug 1968565: Set check-source update strategy

bpickard22:
set the update strategy of network-check-source to rolling update with a maxUnavailable of 10% to match network-check-target. This was done to meet the platform requirement that cluster Daemonsets have a rolling update with max unavailable of 10%


knobunc:
/hold
Holding until we understand whether we need a fix or if the conformance test needs to be changed.

PR #1124, opened 2021-06-07
Bug 1914398: Changed pod user to non-root

nicklesimba:
Changed pod user to non-root based on [Bug 1914398](https://bugzilla.redhat.com/show_bug.cgi?id=1914398)


nicklesimba:
/retitle Bug 1914398: Changed pod user to non-root

PR #1126, opened 2021-06-09
Bug 1967933: Add network-tools to image-stream

tssurya:
We need to reference the network-tools imagestream from an
operator for it to be a part of the payload.

### DIFF
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -62,3 +62,7 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-network-metrics-daemon:latest
+  - name: network-tools
+    from:
+      kind: DockerImage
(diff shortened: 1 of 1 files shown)


alexanderConstantinescu:
@tssurya : this is needed for 4.9 right? So we should hold it until after code freeze if I understood correctly?

tssurya:
> @tssurya : this is needed for 4.9 right? So we should hold it until after code freeze if I understood correctly?

@alexanderConstantinescu : I  think we should get it in now before Friday. It determines if the image gets into the payload for 4.8. In OCP build we already have the `payload:true` payload flag set to true for 4.8. [https://coreos.slack.com/archives/CJARLA942/p1623164057338300?thread_ts=1623078940.320700&cid=CJARLA942]

PR #1128, opened 2021-06-14
Bug 1896533: Nonexistent Namespaces Degradation logging message

nicklesimba:
When nonexistent namespaces cause errors, provided users with additional logging to clarify possible next steps:

"Consider possible solutions:
(1) Edit config files to include existing namespace
(2) Create non-existent namespace
(3) Delete erroneous network-attachment-definition"

Based on [Bugzilla Bug #1896533](https://bugzilla.redhat.com/show_bug.cgi?id=1896533)

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -299,6 +300,12 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		// Open question: should an error here indicate we will never retry?
 		if err := apply.ApplyObject(ctx, r.client, obj); err != nil {
 			err = errors.Wrapf(err, "could not apply (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
+
+			// If error comes from nonexistent namespace print out a help message.
+			if obj.GroupVersionKind().Kind == "NetworkAttachmentDefinition" && strings.Contains(err.Error(), "namespaces") {
+				err = errors.Wrapf(err, "could not apply (%s) %s/%s; Namespace error for networkattachment definition, consider possible solutions: (1) Edit config files to include existing namespace (2) Create non-existent namespace (3) Delete erroneous network-attachment-definition", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
+			}
+
 			log.Println(err)
 
 			// Ignore errors if we've asked to do so.


bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -74,6 +74,9 @@ spec:
         - name: webhook-certs
           mountPath: /etc/webhook
           readOnly: True
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
```
dougbtv, line 79: Hey Nikhil, I think these code is for a different PR?

bindata/network/network-metrics/001-daemonset.yaml
```diff
@@ -69,6 +69,9 @@ spec:
             - name: metrics-certs
               mountPath: /etc/metrics
               readOnly: True
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534
```
dougbtv, line 74: And this is for another PR.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -301,6 +302,16 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 			err = errors.Wrapf(err, "could not apply (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
 			log.Println(err)
 
+			// If error comes from nonexistent namespace print out a help message.
+			if obj.GroupVersionKind().Kind == "NetworkAttachmentDefinition" && strings.Contains(err.Error(), "namespaces") {
+				log.Println(" ")
```
dougbtv, line 307: Hey Nikhil, this error message probably looks really nice in a terminal, and I'm all about that. However, I think by convention, the go error

Thanks regardless for making this actionable output, I really like that it gives the users steps to take.

But I think we want to condense this into a single line, even if it's a big long. Also, maybe prefix it with some more information about the context of the error, for example:

```
namespace error for networkattachment definition, consider these possible solutions: ...etc
```

dougbtv (changes requested):

dougbtv (approved):
/lgtm

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -301,6 +302,11 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 			err = errors.Wrapf(err, "could not apply (%s) %s/%s", obj.GroupVersionKind(), obj.GetNamespace(), obj.GetName())
 			log.Println(err)
 
+			// If error comes from nonexistent namespace print out a help message.
+			if obj.GroupVersionKind().Kind == "NetworkAttachmentDefinition" && strings.Contains(err.Error(), "namespaces") {
+				log.Println("Namespace error for networkattachment definition, consider possible solutions: (1) Edit config files to include existing namespace (2) Create non-existent namespace (3) Delete erroneous network-attachment-definition")
```
danwinship, line 307: Just logging this to stdout isn't likely to get noticed. The suggestion in bz1896533 was that if CNO is going to become Degraded because of this, then it needs to explain why _in the Degraded status message_. (In this context, that means you want to change the value of `err`, so as to change what gets passed to `r.status.SetDegraded` below.)

But I also think that the reporters are correct and there are good reasons to not consider this case to be an error, so I'm not sure this patch is right anyway.

danwinship (changes requested):

PR #1131, opened 2021-06-15
Bug 1973770: Reduce SSL sec level for ovsdb connections 

jcaamano:
When upgrading from 4.5 to 4.6 the more recent openssl defaults to
stricter security requirements and rejects the use of 1024 bit dh params
when ovn-controller 4.6 tries to connect to sbdb 4.5, blocking the
upgrade.

This patch softens the security requirement back to allow 1024 bit dh
params.

Note that this should not affect once the upgrade is completed and these
connections are among 4.6 components with a recent openssl version on
both sides as this will cause to negotiate a cipher that does not use
fixed dh params.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1973770


asood-rh:
/lgtm

The PR fixes the networking component issue, though the upgrade to 4.5->4.6 fails.

deads2k:
(patch manager)

high bz, 1.5

PR #1138, opened 2021-06-22
ovn offload: enable ovnkube-node daemonset to run in different modes

zshi-redhat:
Enable Data Processing Unit(DPU) modes for OVNKubernetes

With ovn-kubernetes PR: https://github.com/ovn-org/ovn-kubernetes/pull/2005
ovnkube-node pod is able to run in one of the three modes:

1. full: default mode, ovnkube-node runs on normal hosts, namely ovnkube-node
2. dpu: ovnkube-node runs on DPUs, namely ovnkube-node-dpu
3. dpu-host: ovnkube-node runs on DPU hosts, namely ovnkube-node-dpu-host

Since DPUs and DPU hosts are supposed to be managed by separate OpenShift
clusters, the following mode combinations can be used in each cluster:

1. Infra cluster (DPU cluster): full and dpu modes
2. Tenant cluster (DPU host cluster): full and dpu-host modes

This commit introduces "network.operator.openshift.io/dpu-host" label to:

1. Allow scheduling of ovnkube-node daemonset pods on normal(non-DPU) hosts in Tenant cluster
2. Allow scheduling of ovnkube-node-dpu-host daemonset pods on DPU hosts in Tenant cluster

ovnkube-node-dpu in "dpu" mode is not enabled for Infra cluster

Example configmap "hack/dpu-mode.yaml" can be used to switch DPU mode
in Tenant cluster. Currently only "dpu-host" mode is supported.


zshi-redhat:
@squeed I removed the empty daemonset pod_status check as I'm adding a demonset which can be empty. (I see your note in the code for removing the check).

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -267,6 +274,7 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            --ovnkube-node-mode full \
```
squeed, line 277: Shouldn't the default for this argument be `full`? Otherwise everyone has to update their daemonsets.

zshi-redhat: Yes, the default ovnkube-node-mode is full if not specified. Do you prefer to remove it? I was thinking it may cause trouble if the default mode will be changed, but that is unlikely to happen as far as I can see.

bindata/network/ovn-kubernetes/ovnkube-node-smartnic-host.yaml
```diff
@@ -0,0 +1,315 @@
+---
```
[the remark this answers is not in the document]
zshi-redhat: > Ugh. So much duplication between this file and `ovnkube-node.yaml`. Which also makes it hard to see what is actually different between the two...
> 
> Is there some way we could make the go code process `ovnkube-node.yaml` twice with different variables? 

I think yes, we want to process the `ovnkube-node.yaml` twice when `network.operator.openshift.io/ovn-offload-host` is labeled to any of the nodes in the cluster. perhaps we can list all nodes and detect the node labels, return a `smart-nic-enabled` flag in bootstrapOVN, which indicates that `renderOVNKubernetes` shall render the `ovnkube-node.yaml` twice, one in the `full` mode, the other in the `smart-nic-host` mode. wdyt?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -97,9 +97,6 @@ func (status *StatusManager) SetFromPods() {
 			if !isNonCritical(ds) {
 				hung = append(hung, status.CheckCrashLoopBackOffPods(dsName, ds.Spec.Selector.MatchLabels, "DaemonSet")...)
 			}
-		} else if ds.Status.NumberAvailable == 0 { // NOTE: update this if we ever expect empty (unscheduled) daemonsets ~cdc
```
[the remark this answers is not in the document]
zshi-redhat: By merging `ovnkube-node.yaml` and ovnkube-node-smart-nic-host.yaml` into one yaml file `ovnkube-node.yaml`, this check won't need to be removed (we only render the ovnkube-node-smart-nic daemon when at least one node is configured to offload).

bindata/network/ovn-kubernetes/ipsec.yaml
```diff
@@ -27,6 +27,13 @@ spec:
         openshift.io/component: network
         kubernetes.io/os: "linux"
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: network.operator.openshift.io/ovn-offload-host
```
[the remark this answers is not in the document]
zshi-redhat: Not yet, I was thinking we can bikeshed the name in this PR.
1. not necessary, it is based on the fact only ovn-k8s supports smartnic offload as of writing and ovn would help differentiate the cx-5 ovs flow offloading (ovs hw offload).
2. I learned that smart-nic usually refers to the embedded NIC on the DPU, for example, the cx-6 dx. DPU may be a better user/admin facing name, although we use `smart-nic` in ovn-k8s project.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
zshi-redhat: I will remove the `--ovnkube-node-mode full` from the `ovnkube-node.yaml`.

bindata/network/ovn-kubernetes/ipsec.yaml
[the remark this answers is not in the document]
danwinship: Re "ovn" or not, I guess I was saying, if we imagine that some third-party plugin eventually implements offloading the network plugin to the BF-2, would we expect them to be able to use the same annotation we do (and thus maybe automatically cause some other parts of OCP to do the right thing with their cluster) or would they need to use their own "calico-offload-host", etc, annotation? (I'm not sure what the right answer here is...)

For offload/smart-nic/dpu, I was saying, depending on how you look at the feature, the OVN offload part may not be the most interesting part of it. Some customers might care more about running other software on the DPU than they care about offloading packet processing...

bindata/network/ovn-kubernetes/ovnkube-node-smartnic-host.yaml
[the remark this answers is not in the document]
danwinship: > I think yes, we want to process the `ovnkube-node.yaml` twice when `network.operator.openshift.io/ovn-offload-host` is labeled to any of the nodes in the cluster.

Hm... it seems weird to me to generate different outputs based on how nodes are labeled... I guess it's not exactly _forbidden_, but in general, we only vary our outputs based on config objects, not based on annotations on random other objects in the cluster... (And especially, this means that when CNO first starts up, it would only generate the "mode=full" configuration, because none of the masters would be labeled for offload, and then later when the workers got added to the cluster, it would have to generate the "mode=smart-nic-host" config as well...

OTOH, maybe the answer is just that there should be some flag in some config object that has to be enabled if you're going to use bf-2s...

bindata/network/ovn-kubernetes/ipsec.yaml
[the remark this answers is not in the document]
zshi-redhat: > Re "ovn" or not, I guess I was saying, if we imagine that some third-party plugin eventually implements offloading the network plugin to the BF-2, would we expect them to be able to use the same annotation we do (and thus maybe automatically cause some other parts of OCP to do the right thing with their cluster) or would they need to use their own "calico-offload-host", etc, annotation? (I'm not sure what the right answer here is...)

Good point, I didn't think of a third-party network plugin for BF2 offload.
I think third-party plugin may or may not use ovn. We shall expect to use the same label for the dpu offload feature. 

> 
> For offload/smart-nic/dpu, I was saying, depending on how you look at the feature, the OVN offload part may not be the most interesting part of it. Some customers might care more about running other software on the DPU than they care about offloading packet processing...

I learned that some vendors may not have the smart-nic function (packet processing offload) on DPU, and DPU is currently used by Nvidia, not an industry standard term.  I presume we want to include the DPU equivalent term here, something like `network.operator.openshift.io/dpu-host` for host that has DPU card installed and `network.operator.openshift.io/dpu` for DPU node (`dpu` may be replaced by a equivalent neutral term in the future).

bindata/network/ovn-kubernetes/ovnkube-node-smartnic-host.yaml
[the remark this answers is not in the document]
zshi-redhat: > > I think yes, we want to process the `ovnkube-node.yaml` twice when `network.operator.openshift.io/ovn-offload-host` is labeled to any of the nodes in the cluster.
> 
> Hm... it seems weird to me to generate different outputs based on how nodes are labeled... I guess it's not exactly _forbidden_, but in general, we only vary our outputs based on config objects, not based on annotations on random other objects in the cluster... (And especially, this means that when CNO first starts up, it would only generate the "mode=full" configuration, because none of the masters would be labeled for offload, and then later when the workers got added to the cluster, it would have to generate the "mode=smart-nic-host" config as well...
> 
> OTOH, maybe the answer is just that there should be some flag in some config object that has to be enabled if you're going to use bf-2s...


Agree. I think this is the first time we need to distinguish the nodes (config) in CNO because not all nodes in the cluster will be enabled with DPU. This means a single boolean flag in the CNO operator config will not be sufficient, we also need to support specifying the node group (e.g. MCP) which has DPU installed. Do you see we are adding this node group related config in CNO operator?

danwinship:
So I think at this point, the expectation is:

- In the infra cluster, we will deploy
  - ovn-kubernetes in `full` mode to masters/workers
  - ovn-kubernetes in `smart-nic` mode to DPUs
- In the tenant cluster, we will deploy
  - ovn-kubernetes in `full` mode on hosts with no DPU
  - ovn-kubernetes in `smart-nic-host` mode on hosts with DPUs

So that sounds like

- In all cases, we generate the `full`-mode DaemonSet. (Maybe we conditionalize the anti-affinity clause and don't bother generating that in non-DPU clusters.)
- If there is some config flag indicating that this is a DPU tenant cluster, we generate the `smart-nic-host`-mode DaemonSet
- If there is some config flag indicating that this is a DPU infra cluster, we generate the `smart-nic`-mode DaemonSet

For now, until we have a real API, maybe we could just reuse the [`gateway-mode-config` ConfigMap hack](https://github.com/openshift/cluster-network-operator/blob/1b680a85/pkg/network/ovn_kubernetes.go#L246..L264) to configure this? Just check for a few more keys in that ConfigMap and set some variables in the template data, and then check them from the templates

pkg/controller/statusmanager/pod_status.go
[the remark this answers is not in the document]
zshi-redhat: Done

bindata/network/ovn-kubernetes/ipsec.yaml
```diff
@@ -27,6 +27,13 @@ spec:
         openshift.io/component: network
         kubernetes.io/os: "linux"
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: network.operator.openshift.io/dpu-host
```
squeed, line 35: Does `dpu-host` mean any host with a DPU, or the literal DPU itself?

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -2,7 +2,11 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
+  {{ if eq .OVN_DPU_MODE "dpu-host" }}
```
squeed, line 5: What happens if this switch is flipped? Do we support enabling this in running clusters? Is every node expected to have a DPU?

hack/dpu-mode.yaml
```diff
@@ -0,0 +1,9 @@
+# Example ConfigMap to enable dpu-host mode with OVNKubernetes
```
squeed, line 1: Why is this a separate configmap and not part of the CNO type?

bindata/network/ovn-kubernetes/ipsec.yaml
zshi-redhat: The former, any host with a DPU.

squeed: aha, makes sense.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
zshi-redhat: Yes, current idea is to enable this in running cluster, not changing CNO API.
We'd start by deploying the full ovn-k8s CNI on the dpu host, then configure the dpu host to run in the dpu-host mode.

If the switch is flipped,  full-mode (default) ovnkube-node will be re-deployed to the dpu host,  which may conflict with the ovnkube-node running on dpu.

hack/dpu-mode.yaml
zshi-redhat: The initial target is to make it dev preview in 4.10. we don't want to risk changing the API until the overall design is properly reviewed and tested.

squeed:
A question around why this needs a config map, but otherwise seems reasonable.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
squeed: I see. How much do you care if the change is disruptive?

So, one problem is that renaming a daemonset is the same as a delete and add, which is unsynchronized. This means you could have two ovnkube-nodes running on the same host. It would be better if you could do this without renaming the daemonset.

Is there some way for ovn-kube node to autodetect that it should be in dpu mode? Maybe by reading the labels on the Node object itself? Then you don't have to do any work when you roll out the changes.

danwinship: > So, one problem is that renaming a daemonset

We're not renaming it; see the corresponding code change. When DPU offload is enabled, CNO will render this particular template file twice, once with `OVN_DPU_MODE=full` to generate the `ovnkube-node` DaemonSet for the "dumb" nodes, and once with `OVN_DPU_MODE=dpu-host` to generate the `ovnkube-node-dpu-host` DaemonSet for the "smart" nodes.

> This means you could have two ovnkube-nodes running on the same host.

This part is still a problem though.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -21,12 +29,27 @@ spec:
...
         app: ovnkube-node
+        {{ end }}
         component: network
         type: infra
         openshift.io/component: network
         kubernetes.io/os: "linux"
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
```
danwinship, line 44: note that (as the overly-long name indicates), this only affects the initial scheduling; it will not cause existing `ovnkube-node` pods on `dpu-host` nodes to be killed

zshi-redhat: @danwinship  Yes, I think we shall change to use requiredDuringExecution when it is supported in k8s.
From my testing, removing the label `dpu-host` also triggers the removal of `ovnkube-node-dpu-host` pods. Could it be due to the reconciling of daemonset changes the resourceVersion which is considered as a new scheduling event?

zshi-redhat:
pushed a new version to resolve conflict.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -274,6 +299,9 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            {{ if eq .OVN_DPU_MODE "dpu-host" }}
```
Billy99, line 302:
```suggestion
            {{- if eq .OVN_DPU_MODE "dpu-host" }}
```
White line was being added and cause script to break.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -274,6 +299,9 @@ spec:
...
+            --ovnkube-node-mode dpu-host \
+            {{ end }}
```
Billy99, line 304:
```suggestion
            {{- end }}
```
White line was being added and cause script to break.

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -274,6 +299,9 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            {{ if eq .OVN_DPU_MODE "dpu-host" }}
+            --ovnkube-node-mode dpu-host \
```
Billy99, line 303:
```suggestion
            --ovnkube-node-mode smart-nic-host \
```
Needed until https://github.com/ovn-org/ovn-kubernetes/pull/2409/files merges

Billy99: Need this before `exec /usr/bin/ovnkube ... ` is called:
```
          node_mgmt_port_netdev_flags=
          if [[ -n "${OVNKUBE_NODE_MGMT_PORT_NETDEV}" ]] ; then
            node_mgmt_port_netdev_flags="--ovnkube-node-mgmt-port-netdev ${OVNKUBE_NODE_MGMT_PORT_NETDEV}"
          fi
```
And then this here:
```suggestion
            --ovnkube-node-mode dpu-host \
            ${node_mgmt_port_netdev_flags} \
```

Billy99 (commented):
Changes discovered when running on Cluster-2.

zshi-redhat:
@danwinship  Could you take another look?

zshi-redhat:
Superseded by https://github.com/openshift/cluster-network-operator/pull/1220

PR #1141, opened 2021-06-28
Bug 1970985: SDN-1955: Add pre-puller ds to reduce upgrade downtime

tssurya:
According to https://bugzilla.redhat.com/show_bug.cgi?id=1943334#c0 it takes roughly a minute during upgrades for old pods to get killed and new pods to get created.

It is predicted that a major chunk of this time is spent in pulling the new image into the node. (@squeed : do we have data to back up this claim that I can point to?)

This PR adds a new prepuller daemonset that is basically a no-op which simply assists in pulling the new image onto the nodes before the new pods get created so that it cuts down on the downtime.

Idea, Co-Authored By: https://github.com/squeed/openshift-cluster-network-operator/commit/42c0d1db5576a2e4e6b16b115e657477dbc33073

### DIFF
--- a/bindata/network/ovn-kubernetes/pre-puller.yaml
+++ b/bindata/network/ovn-kubernetes/pre-puller.yaml
@@ -1,0 +1,40 @@
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: ovnkube-upgrades-prepuller
+  namespace: openshift-ovn-kubernetes
+  annotations:
+    kubernetes.io/description: |
+      This daemonset launches the pre-puller component during upgrades that pulls the image onto the node.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+spec:
+  selector:
+    matchLabels:
+      app: ovnkube-upgrades-prepuller
+  template:
+    metadata:
+      labels:
+        app: ovnkube-upgrades-prepuller
+        component: network
+        type: infra
+        openshift.io/component: network
+        kubernetes.io/os: "linux"
+    spec:
+      serviceAccountName: ovn-kubernetes-node
+      hostNetwork: true
+      priorityClassName: "system-node-critical"
+      containers:
+        # ovnkube-upgrades-prepuller: no-op container that simply pulls the new image during upgrades
+      - name: ovnkube-upgrades-prepuller
+        image: "{{.OvnImage}}"
+        imagePullPolicy: Always
+        command:
+        - /bin/bash
+        - -c
+        - |
+          echo "$(date -Iseconds) - finished pulling ovnkube-node image."
+          sleep infinity
+      nodeSelector:
+        beta.kubernetes.io/os: "linux"
+      tolerations:
+      - operator: "Exists"
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -179,6 +179,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		updateNode, updateMaster = shouldUpdateOVNKonUpgrade(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.ExistingMasterDaemonset, os.Getenv("RELEASE_VERSION"))
 	}
 
+	renderPrePull := false
+	if updateNode {
+		updateNode, renderPrePull = shouldUpdateOVNKonPrepull(bootstrapResult.OVN.ExistingNodeDaemonset, bootstrapResult.OVN.PrePullerDaemonset, os.Getenv("RELEASE_VERSION"))
+	}
+
 	// If we need to delay master or node daemonset rollout, then we'll replace the new one with the existing one
 	if !updateMaster {
 		us, err := k8s.ToUnstructured(bootstrapResult.OVN.ExistingMasterDaemonset)
@@ -193,6 +198,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 			return nil, errors.Wrap(err, "failed to transmute existing node daemonset")
 		}
 		objs = k8s.ReplaceObj(objs, us)
+	}
+
+	if !renderPrePull {
+		// remove prepull from the list of objects to render.
+		objs = k8s.RemoveObjByGroupKindName(objs, "apps", "DaemonSet", "openshift-ovn-kubernetes", "ovnkube-upgrades-prepuller")
 	}
 
 	return objs, nil
@@ -493,6 +503,16 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 		}
 	}
 
+	prePullerDS := &appsv1.DaemonSet{}
+	nsn = types.NamespacedName{Namespace: "openshift-ovn-kubernetes", Name: "ovnkube-upgrades-prepuller"}
+	if err := kubeClient.Get(context.TODO(), nsn, prePullerDS); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("Failed to retrieve existing prepuller DaemonSet: %w", err)
+		} else {
+			prePullerDS = nil
+		}
+	}
+
 	res := bootstrap.BootstrapResult{
 		OVN: bootstrap.OVNBootstrapResult{
 			MasterIPs:               ovnMasterIPs,
@@ -500,6 +520,7 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 			ExistingMasterDaemonset: masterDS,
 			ExistingNodeDaemonset:   nodeDS,
 			OVNKubernetesConfig:     ovnConfigResult,
+			PrePullerDaemonset:      prePullerDS,
 		},
 	}
 	return &res, nil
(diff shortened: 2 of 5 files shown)


bindata/network/ovn-kubernetes/pre-puller.yaml
```diff
@@ -0,0 +1,44 @@
...
+        component: network
+        type: infra
+        openshift.io/component: network
+        kubernetes.io/os: "linux"
+    spec:
+      serviceAccountName: ovn-kubernetes-node
+      hostNetwork: true
+      hostPID: true
+      priorityClassName: "system-node-critical"
+      #restartPolicy: OnFailure
```
tssurya, line 27: So apparently `Always` is the only `restartPolicy` supported for a ds, due to which this ds keeps crashing. I wonder how folks move the daemonset pods to `succeeded` state and not crashloop after completing??

pkg/network/ovn_kubernetes.go
```diff
@@ -477,13 +491,24 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 		}
 	}
 
+	prePullerDS := &appsv1.DaemonSet{}
```
tssurya, line 494: We should probably add a wait loop to make sure the daemonset gets created before we remove it from the loop? Like stated in the previous comment the startup situation is a bit confusing.

bindata/network/ovn-kubernetes/pre-puller.yaml
```diff
@@ -0,0 +1,44 @@
...
+      containers:
+        # ovnkube-upgrades-prepuller: programs the vswitch with flows from the sbdb
```
squeed, line 29: this comment is wrong :-)

bindata/network/ovn-kubernetes/pre-puller.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: ovnkube-upgrades-prepuller
+        image: "{{.OvnImage}}"
+        imagePullPolicy: Always
+        command:
+        - /bin/bash
+        - -c
+        - |
+          echo "$(date -Iseconds) - finished pulling ovnkube-node image."
+          exit 0
+        securityContext:
```
squeed, line 39: This doesn't need to be privileged.

bindata/network/ovn-kubernetes/pre-puller.yaml
```diff
@@ -0,0 +1,44 @@
...
+      labels:
+        app: ovnkube-upgrades-prepuller
+        component: network
+        type: infra
+        openshift.io/component: network
+        kubernetes.io/os: "linux"
+    spec:
+      serviceAccountName: ovn-kubernetes-node
+      hostNetwork: true
+      hostPID: true
```
squeed, line 25: This doesn't need to be hostPID

pkg/network/ovn_kubernetes.go
squeed: We don't need a wait loop if our logic is level-triggered. In other words, if we decide we want the pre-pull, it doesn't matter how many times CNO triggers a re-rendering pass, the output *should* be stable.

pkg/network/ovn_kubernetes.go
```diff
@@ -550,6 +575,38 @@ func shouldUpdateOVNKonIPFamilyChange(existingNode, existingMaster *appsv1.Daemo
...
+
+	// if node is already upgraded, then no need to pre-pull
+	existingVersion := existingNode.GetAnnotations()["release.openshift.io/version"]
+	if existingVersion == releaseVersion {
+		return true, false
+	}
+
+	// at this point, we've determined we need an upgrade
+	if prePuller == nil {
+		klog.Infof("Rolling out the no-op prepuller daemonset...")
```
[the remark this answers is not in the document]
squeed: that's correct, there is additional logic that deletes any previously-rendered (but now missing) objects.

pkg/network/ovn_kubernetes.go
```diff
@@ -191,6 +196,15 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		objs = k8s.ReplaceObj(objs, us)
 	}
 
+	if !renderPrePull && bootstrapResult.OVN.PrePullerDaemonset != nil {
```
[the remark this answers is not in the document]
squeed: We handle this situation by not rendering the prepul daemonset if there is no node daemonset. I think this condition can simply be
```go
if !renderPrePull {
    // ...
```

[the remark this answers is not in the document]
squeed: This does mean we will need to hard-code the name of the pre-puller daemonset, but that's fine.

bindata/network/ovn-kubernetes/pre-puller.yaml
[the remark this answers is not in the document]
tssurya: done.

bindata/network/ovn-kubernetes/pre-puller.yaml
tssurya: oops, copy paste typos. good eye.

bindata/network/ovn-kubernetes/pre-puller.yaml
tssurya: oh yea true, why do we need access to the process namespace of the host? removing this...

pkg/network/ovn_kubernetes.go
tssurya: hmm yea, I am unraveling the magic of level-triggering..

pkg/network/ovn_kubernetes.go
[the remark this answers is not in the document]
tssurya: All ds names are hardcoded, so I think that's fine like you are saying.
The problem with the

```
if !renderPrePull {
    // ...
```

is that let's say renderPrePull is false which it is most of the time... now it goes into this logic and tries to remove the daemonset which actually doesn't exist in the first place to begin with. The first time we actually create that ds is only when this flag becomes true which would be during an image version change/update/upgrade. So end up trying to remove a nil ds:

```
I0701 11:05:55.216099 1000141 log.go:184] Failed to render: failed to transmute existing prepuller daemonset: failed to convert to unstructured (unmarshal): Object 'Kind' is missing in 'null'
```

pkg/network/ovn_kubernetes.go
```diff
@@ -550,6 +585,38 @@ func shouldUpdateOVNKonIPFamilyChange(existingNode, existingMaster *appsv1.Daemo
...
+	if existingVersion == releaseVersion {
+		return true, false
+	}
+
+	// at this point, we've determined we need an upgrade
+	if prePuller == nil {
+		klog.Infof("Rolling out the no-op prepuller daemonset...")
+		return false, true
+	}
+
```
squeed, line 610: We should also check to see that the version of the pre-puller matches the expectedVersion. Unlikely, but it could happen if we upgrade and downgrade.

tssurya: Yea I have added this in the new PR. PTAL!

pkg/util/k8s/unstructured.go
```diff
@@ -67,3 +67,27 @@ func ReplaceObj(objs []*uns.Unstructured, new *uns.Unstructured) []*uns.Unstruct
 
 	return out
 }
+
+// RemoveObjByKindName will remove a given object from the list of objects.
+// It will match on the object's kind, namespace and name. If the object isn't found
+// in the list, it will not be removed.
+func RemoveObjByKindName(objs []*uns.Unstructured, kind, namespace, name string) []*uns.Unstructured {
```
squeed, line 74: This needs to take the Group (e.g.  `apps`) as well, since there are sometimes collisions between groups.

tssurya: oh yea totally forgot that!

squeed:
Would also like to see e2e-gcp-ovn-upgrade pass (or otherwise look good, logs-wise)

jluhrsen:
@tssurya , is this ready now besides hopefully getting some of these failing checks to pass with a
/retest

jluhrsen:
there is little hope that the gcp-ovn-upgrade or openstack job will pass, but both [[1](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/cluster-network-operator/openshift-cluster-network-operator-master-presubmits.yaml#L467)][[2](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/cluster-network-operator/openshift-cluster-network-operator-master-presubmits.yaml#L732)] are optional so shouldn't
block this from being merged.

squeed:
I just checked the Loki logs, and everything is as expected: https://grafana-loki.ci.openshift.org/explore?orgId=1&left=%5B%221626206400000%22,%221626210000000%22,%22Grafana%20Cloud%22,%7B%22expr%22:%22%7Binvoker%3D%5C%22openshift-internal-ci%2Fpull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-upgrade%2F1415041102456557568%5C%22%7D%20%7C%20unpack%20%7C%20namespace%3D%5C%22openshift-network-operator%5C%22%20%7C%3D%20puller%5Cn%22%7D%5D

/approve
/lgtm
woohoo

jluhrsen:
> /retest-required
> 
> Please review the [full test history](https://prow.ci.openshift.org/pr-history/?org=openshift&repo=cluster-network-operator&pr=1141) for this PR and help us cut down flakes.

just for my learning, can someone explain why this happened?

jluhrsen:
> > /retest-required
> > Please review the [full test history](https://prow.ci.openshift.org/pr-history/?org=openshift&repo=cluster-network-operator&pr=1141) for this PR and help us cut down flakes.
> 
> just for my learning, can someone explain why this happened?

n'mind. I get it now after reading the email I just received about this. it only reruns the required tests:
/retest-required

tssurya:
Measuring the time gained from after the PR and before the PR using the `gcp-ovn-upgrade` job stats:

Based on https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1141/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-upgrade/1415041102456557568/build-log.txt :

```
Jul 13 21:33:15.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-9xjn7 node/ci-op-87flfx9t-8aaeb-9snjd-master-0 container/ovnkube-node reason/Killing
Jul 13 21:33:24.476 I ns/openshift-ovn-kubernetes pod/ovnkube-node-9gfkk node/ci-op-87flfx9t-8aaeb-9snjd-master-0 container/ovnkube-node reason/ContainerStart duration/2.00s
```

Takes approximately ~9.4seconds [downtime]

versus jobs before the PR:

```
Jul 08 15:26:51.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-f28sm node/ci-op-t1r1b67d-8aaeb-46xvl-worker-a-fn9px container/kube-rbac-proxy reason/Killing
Jul 08 15:27:06.018 I ns/openshift-ovn-kubernetes pod/ovnkube-node-t8p2l node/ci-op-t1r1b67d-8aaeb-46xvl-worker-a-fn9px container/ovnkube-node reason/ContainerStart duration/2.00s
```

Takes approximately ~15seconds [downtime]

Another run:

```
Jul 12 19:18:30.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-bnsk2 node/ci-op-dvqpcyxi-8aaeb-kw8n9-master-0 container/ovnkube-node reason/Killing
Jul 12 19:18:39.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-8cqjj node/ci-op-dvqpcyxi-8aaeb-kw8n9-master-0 container/ovnkube-node reason/Started
```
Takes approximately 9seconds [downtime]


Since this number varies, I think the true gain is in the time it takes for pulling:

```
Jul 12 19:18:38.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-8cqjj node/ci-op-dvqpcyxi-8aaeb-kw8n9-master-0 container/ovnkube-node reason/Pulled duration/1.187s image/registry.build01.ci.openshift.org/ci-op-dvqpcyxi/stable@sha256:e015efd1c9c17ad497e29ea4615e73c4fd6d149de2084596166716130f2fc1ba
Jul 08 15:27:04.000 I ns/openshift-ovn-kubernetes pod/ovnkube-node-t8p2l node/ci-op-t1r1b67d-8aaeb-46xvl-worker-a-fn9px container/ovnkube-node reason/Pulled duration/1.201s image/registry.build01.ci.openshift.org/ci-op-t1r1b67d/stable@sha256:2a7401b16ef332f96902a3b5b5494b165caf9f94968281e79515f5f80400fe73
```

which is like 1-2 seconds per pod (I don't know if the CI pulls reflect the real scenario pulls). That multiplied by each node on a default cluster of 6 nodes would give us a gain of ~6seconds at least in total \o/. That much of downtime is what we save. 

The prepuller ds itself takes ~5 seconds for creation+pulling totally. Considering the goal is to reduce the ovn-k pods downtime its not much plus there is no change in the total upgrade time due to this change.

I'll work with Jamo to see if the time difference can be measured in a better manner.

tssurya:
> @tssurya: tssurya unauthorized: /override is restricted to Repo administrators, approvers in top level OWNERS file.

I think we should override gcp-ovn job. Failures are unrelated and this job had passed previously for this PR multiple times, now its failing with some weird unrelated issues.

tssurya:
/hold

noooooooooooo 
Conflict due to https://github.com/openshift/cluster-network-operator/pull/1145 in the last minute.

Seems like other PRs overrode the gcp-ovn job

tssurya:
@squeed : Can you leave the lgtm again? [https://github.com/openshift/cluster-network-operator/pull/1145#issuecomment-881946269 got in first so got a conflict]. Let's also override the e2e-gcp-ovn job if it doesn't pass, has been failing for the whole weekend.

PR #1144, opened 2021-06-30
WIP: test SSL invalidity on OVN cluster

alexanderConstantinescu:
/cc

### DIFF
--- a/pkg/controller/pki/pki_controller.go
+++ b/pkg/controller/pki/pki_controller.go
@@ -210,7 +210,7 @@ func newPKI(config *netopv1.OperatorPKI, clientset *kubernetes.Clientset, mgr ma
 			Namespace: config.Namespace,
 			Name:      config.Name + "-cert",
 			Validity:  OneYear / 2,
-			Refresh:   OneYear / 4,
+			Refresh:   20 * time.Minute,
 			CertCreator: &certrotation.ServingRotation{
 				Hostnames: func() []string { return []string{spec.TargetCert.CommonName} },
 


alexanderConstantinescu:
Notes here for future reference:

Background: I wanted to do this as to test the outcome of using expired certificates on an OVN cluster. 

Outcome: 

- library-go: library-go updates the secret cert when it expires, even if the refresh period is longer than the validity. 
- pods: pods mount the secret cert created by library-go, when the secret is updated by library-go: the secret data in the pod is updated. It is thus up to the process in the pod to re-initialize the TLS / SSL credentials
- nbdb/sbdb: this is fine. ovsdb-server will dynamically re-initialize its TLS data when it sees that the certificate is expired. It will thus log something like this, until it picks up the refreshed data on disk in the pod. 
```
2021-06-30T11:17:50Z|00092|reconnect|WARN|ssl:10.0.0.3:36084: connection dropped (Protocol error)
2021-06-30T11:17:56Z|00093|stream_ssl|WARN|SSL_accept: error:14094415:SSL routines:ssl3_read_bytes:sslv3 alert certificate expired
2021-06-30T11:17:56Z|00094|jsonrpc|WARN|ssl:10.0.0.3:36304: receive error: Protocol error
``` 
- ovnkube-master: most exec'ed ovn-nbctl commands are fine, this is because the binary is exec'ed for every call. The calls between the time of expiration and data reload in the pod might fail though (I didn't see any however when using this PR). go-ovn will have the same behavior, since we have https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/util/go_ovn.go#L110 
- northd: does not seem to indicate any problems when using an expired cert...it either doesn't log it, or doesn't care about it. https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/ovnkube-master.yaml#L96 seems to indicate that it doesn't care about it
- ovn-controller: doesn't log anything either except
```
2021-06-30T11:52:46Z|01659|poll_loop|INFO|Dropped 28 log messages in last 31 seconds (most recently, 28 seconds ago) due to excessive rate
2021-06-30T11:52:46Z|01660|poll_loop|INFO|wakeup due to [POLLIN] on fd 20 (<->/var/run/openvswitch/br-int.mgmt) at lib/stream-fd.c:157 (53% CPU usage)
2021-06-30T11:52:47Z|01661|poll_loop|INFO|wakeup due to [POLLIN] on fd 19 (10.0.128.2:41502<->10.0.0.5:9642) at lib/stream-ssl.c:832 (53% CPU usage)
2021-06-30T11:52:47Z|01662|poll_loop|INFO|wakeup due to [POLLIN] on fd 20 (<->/var/run/openvswitch/br-int.mgmt) at lib/stream-fd.c:157 (53% CPU usage)
2021-06-30T11:52:47Z|01663|poll_loop|INFO|wakeup due to [POLLIN] on fd 19 (10.0.128.2:41502<->10.0.0.5:9642) at lib/stream-ssl.c:832 (53% CPU usage)
2021-06-30T11:52:47Z|01664|poll_loop|INFO|wakeup due to [POLLIN] on fd 21 (FIFO pipe:[42783]) at controller/pinctrl.c:3883 (53% CPU usage)
```
but that is not because the certificate has expired, because using this PR the certificates had not expired at that moment. The validity period was:
```
    auth.openshift.io/certificate-not-after: "2021-06-30T11:56:47Z"
    auth.openshift.io/certificate-not-before: "2021-06-30T11:46:46Z"
```
so maybe it doesn't use the certificate either?
- ovnkube-node: does not error, but does not reload the cert either........ :shrug: 

The assumption I have is that ovnkube-node is mainly executing ovs-vsctl commands and talking to OVS, which does not have SSL configured...? Maybe that's why it isn't experiencing errors?  But it does execute some `ovn-sbctl` commands, but that should work for the same reasons `ovn-nbctl` commands should, the binary is exec'ed. 

```
I0630 10:56:19.113504    3337 config.go:1081] exec: /usr/bin/ovs-vsctl --timeout=15 del-ssl
I0630 10:56:19.124041    3337 config.go:1081] exec: /usr/bin/ovs-vsctl --timeout=15 set-ssl /ovn-cert/tls.key /ovn-cert/tls.crt /ovn-ca/ca-bundle.crt
I0630 10:56:19.136589    3337 config.go:1081] exec: /usr/bin/ovs-vsctl --timeout=15 set Open_vSwitch . external_ids:ovn-remote="ssl:10.0.0.3:9642,ssl:10.0.0.4:9642,ssl:10.0.0.5:9642"

```

So expired certs are fine. What is not is when they are not expired, but all members do not use the same TLS/SSL config. ovsdb-server will experience problems syncing its raft state, because it will complain that it can't verify the certificate some member is using - even if that member is itself:

```
$ omg get pod -n openshift-ovn-kubernetes -owide | head -n 4
NAME                  READY  STATUS   RESTARTS  AGE    IP           NODE
ovnkube-master-f8chn  5/6    Running  24        1h39m  172.18.3.12  ynprdocpm003.yanocp2.turkcell.tgc
ovnkube-master-g8tns  6/6    Running  0         175d   172.18.3.11  ynprdocpm002.yanocp2.turkcell.tgc
ovnkube-master-pl5d2  6/6    Running  0         2h31m  172.18.3.10  ynprdocpm001.yanocp2.turkcell.tgc
$ omg logs -c sbdb ovnkube-master-g8tns -n openshift-ovn-kubernetes | less
2021-06-25T14:54:15.585287033Z 2021-06-25T14:54:15Z|16627|jsonrpc|WARN|ssl:172.18.3.11:59606: receive error: Protocol error
2021-06-25T14:54:15.585451897Z 2021-06-25T14:54:15Z|16628|reconnect|WARN|ssl:172.18.3.11:59606: connection dropped (Protocol error)
2021-06-25T14:54:36.093696179Z 2021-06-25T14:54:36Z|16629|stream_ssl|WARN|SSL_accept: error:1417C086:SSL routines:tls_process_client_certificate:certificate verify failed
```

It however will not proceed to reloading its own data in event that it might itself be using old data.

PR #1145, opened 2021-06-30
Make egress IP and ICNI mutually exclusive when bootstrapping OVN-kube

alexanderConstantinescu:
OVN-Kubernetes cannot support having egress IP and `--disable-snat-multiple-gws` (used for ICNI traffic) defined at the same time. The reason is how both features SNAT on the GW router. Egress IP SNATs to the egress IP, whereas with `--disable-snat-multiple-gws` we will force all pod IPs to SNAT individually to the node IP. Having both configured at the same time will scramble the egress IP configuration. Many OCP customers want the egress IP feature, one OCP customer wants the ICNI feature...hence, have that one customer use a dedicated field in the config map they are already using, to specify this.

/assign @trozet

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -40,6 +40,8 @@ const CLUSTER_CONFIG_NAMESPACE = "kube-system"
 const OVN_CERT_CN = "ovn"
 const OVN_MASTER_DISCOVERY_POLL = 5
 const OVN_MASTER_DISCOVERY_BACKOFF = 120
+const OVN_LOCAL_GW_MODE = "local"
+const OVN_SHARED_GW_MODE = "shared"
 
 var OVN_MASTER_DISCOVERY_TIMEOUT = 250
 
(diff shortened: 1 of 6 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -194,28 +198,40 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
-		klog.Warningf("Ignoring gateway-mode-config %s. Does not match \"shared\" or \"local\"", cm.Data["mode"])
-		return defaultGatewayMode, nil
+	modeOverride := cm.Data["mode"]
+	_, disableSNATMultipleGWsOverride := cm.Data["disable-snat-multiple-gws"]
+	if modeOverride != OVN_SHARED_GW_MODE && modeOverride != OVN_LOCAL_GW_MODE {
+		klog.Warningf("gateway-mode-config does not match %q or %q, is: %q. Using default OVN configuration: %+v", OVN_LOCAL_GW_MODE, OVN_SHARED_GW_MODE, modeOverride, ovnConfigResult)
+		return ovnConfigResult, nil
+	}
+	ovnConfigResult.GatewayMode = modeOverride
+	if disableSNATMultipleGWsOverride {
```
trozet, line 229: disableSNATMultipleGWs with local gateway mode doesn't do anything, it is assumed to always be enabled for that mode. That being said I dont think it hurts to pass it as enabled to ovnkube with local gw mode +  I dont see why anyone would set this with local gateway mode so safe to ignore checking gateway mode here.

trozet (approved):
/lgtm

PR #1148, opened 2021-07-06
Add alerts for issues with load balancers/ports.

gryf:
Sometimes, we could experience issues regarding unrecoverable state for
either Neutron ports and/or Octavia load balancers being stuck forever.
Currently there is nothing we could do from Kuryr-Kubernetes side except
alerting.

### DIFF
--- a/bindata/network/kuryr/alert-rules.yaml
+++ b/bindata/network/kuryr/alert-rules.yaml
@@ -121,3 +121,17 @@ spec:
       for: 5m
       labels:
         severity: critical
+    - alert: KuryrLoadBalancerNotReady
+      annotations:
+        message: Kuryr noticed an Octavia load balancer unexpectedly stuck in PENDING_* state in the last 20 minutes. Most likely this indicates an issue with OpenStack Octavia.
+      expr: |
+        changes(kuryr_load_balancer_readiness_total[20m]) > 0
+      labels:
+        severity: critical
+    - alert: KuryrPortNotReady
+      annotations:
+        message: Kuryr noticed that a Neutron port was in unexpected DOWN state in the last 20 minutes. Most likely this indicates an issue with OpenStack Neutron.
+      expr: |
+        changes(kuryr_port_readiness_total[20m]) > 0
+      labels:
+        severity: critical


bindata/network/kuryr/alert-rules.yaml
```diff
@@ -89,3 +89,17 @@ spec:
         histogram_quantile(0.95, rate(kuryr_cni_request_duration_seconds_bucket[10m])) > 30
       labels:
         severity: warning
+    - alert: KuryrLoadBalancerNotReady
+      annotations:
+        message: One of the Octavia Load Balancers has timed out with either PENDING_UPDATE/PENDING_CREATE/PENDIN_DELETE state
```
dulek, line 94: Hm, let's try to make sure we're blaming Octavia properly. ;)

> Kuryr noticed an Octavia load balancer unexpectedly stuck in PENDING_* state in the last 20 minutes. Most likely this indicates an issue with OpenStack Octavia.

bindata/network/kuryr/alert-rules.yaml
```diff
@@ -89,3 +89,17 @@ spec:
...
+      expr: |
+        changes(kuryr_load_balancer_readiness[20m]) > 0
+      labels:
+        severity: critical
+    - alert: KuryrPortNotReady
+      annotations:
+        message: One of the Neutron port has timed out with making it active
```
dulek, line 101: And Neutron:

> Kuryr noticed that a Neutron port was in unexpected DOWN state in the last 20 minutes. Most likely this indicates an issue with OpenStack Neutron.

dulek (commented):
Awesome - simple but helpful. I left some wording propositions in the comments, again I'm not sure if they're significantly better.

bindata/network/kuryr/alert-rules.yaml
```diff
@@ -121,3 +121,17 @@ spec:
       for: 5m
       labels:
         severity: critical
+    - alert: KuryrLoadBalancerNotReady
+      annotations:
+        message: Kuryr noticed an Octavia load balancer unexpectedly stuck in PENDING_* state in the last 20 minutes. Most likely this indicates an issue with OpenStack Octavia.
```
MaysaMacedo, line 126: Will it only start alerting after 20 min? It might be too long, maybe 10 mins fits better?

bindata/network/kuryr/alert-rules.yaml
```diff
@@ -121,3 +121,17 @@ spec:
...
+      expr: |
+        changes(kuryr_load_balancer_readiness_total[20m]) > 0
```
MaysaMacedo, line 128: The name of the metric differs from what is registered kuryr_load_balancer_readiness

bindata/network/kuryr/alert-rules.yaml
```diff
@@ -121,3 +121,17 @@ spec:
...
+      labels:
+        severity: critical
+    - alert: KuryrPortNotReady
+      annotations:
+        message: Kuryr noticed that a Neutron port was in unexpected DOWN state in the last 20 minutes. Most likely this indicates an issue with OpenStack Neutron.
+      expr: |
+        changes(kuryr_port_readiness_total[20m]) > 0
```
MaysaMacedo, line 135: ditto

bindata/network/kuryr/alert-rules.yaml
gryf: It works like that: if there is any change on the counter within 20 minutes timeframe, then alarm will be triggered. Given the fact, we do the reconciliation, and retry loop is taken around 10-15 minutes, then we make sure that alarm will remain because we will re-trigger it within 20 min block. Alarm will go if counter stops increase itself. Initially @dulek proposed 30 minutes for the window, but IMHO it's a little bit to long.

bindata/network/kuryr/alert-rules.yaml
gryf: Yes, I know :) what is weird, prometheus client produces two child samples for type counter - one with suffix "_total" and one with suffix "_created"[1].

[1] https://github.com/prometheus/client_python/blob/9a24236695c9ad47f9dc537a922a6d1333d8d093/prometheus_client/metrics.py#L276-L280

bindata/network/kuryr/alert-rules.yaml
MaysaMacedo: @dulek do you still have a preference for 30 min?

dulek: Not at all, @gryf convinced me to do 20 minutes.

PR #1152, opened 2021-07-07
OVNKube: check if br-ex1 is available and pass it as a parameter

fedepaol:
If br-ex1 is available on the host, it means that the user wants to use
it as the ovs bridge for external gateway traffic, so we set the related
parameter in ovnkube-node


trozet (approved):
/lgtm

PR #1154, opened 2021-07-15
Bug 1981055: ovnkube-master handle 60 seconds downtime of API server gracefully in SNO

creydr:
Currently ovnkube-master leader election has the following timers [(ovn-org/ovn-kubernetes config.go)](https://github.com/ovn-org/ovn-kubernetes/blob/f5ddbe324fc561c58629db649524cd4b44075255/go-controller/pkg/config/config.go#L119-L123):
* lease duration: 60 seconds
* renew deadline: 30 seconds
* retry period: 20 seconds
This causes it to go through leader election and restart during an kube-apiserver rollout in a SNO (as this currently takes longer than 60 seconds).

This PR increases the ovnkube-master leader election timers for SNO clusters to its [recommended](https://github.com/openshift/enhancements/pull/832/files#diff-2e28754e69aa417e5b6d89e99e42f05bfb6330800fa823753383db1d170fbc2fR183) values to handle a apiserver disruption:
* lease duration: 137 seconds
* renew deadline: 107 seconds
* retry period: 26 seconds

How to reproduce / validate:
1. Start a SNO cluster with this patch
2. Trigger a apiserver rollout (e.g. via `k patch kubeapiserver/cluster --type merge -p '{"spec":{"forceRedeploymentReason":"something"}}'`)
3. Check the logs from ovnkube-master pod (ovnkube-master container) during the update
4. Validate the pod did not get restarted cause it lost the lease (e.g. check no previous logs are available / don't contain something like the following:
```
$ k logs ovnkube-master-5wwhv -c ovnkube-master --previous
...
I0720 10:42:28.610244       1 leaderelection.go:278] failed to renew lease openshift-ovn-kubernetes/ovn-kubernetes-master: timed out waiting for the condition
E0720 10:42:28.610348       1 leaderelection.go:301] Failed to release lock: resource name may not be empty
I0720 10:42:28.610356       1 master.go:107] No longer leader; exiting
```

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -32,7 +32,8 @@ data:
     [gateway]
     mode={{.OVN_GATEWAY_MODE}}
     nodeport=true
-{{ if .OVNHybridOverlayEnable }}
+{{- if .OVNHybridOverlayEnable }}
+
     [hybridoverlay]
     enabled=true
     {{- if .OVNHybridOverlayNetCIDR }}
@@ -41,4 +42,17 @@ data:
     {{- if .OVNHybridOverlayVXLANPort}}
     hybrid-overlay-vxlan-port="{{.OVNHybridOverlayVXLANPort}}"
     {{- end }}
+{{- end  }}
+{{- if .IsSNO }}
+
+    [masterha]
+    {{- /* 
+    Even in case of SNO there will be only one ovn-master, we set dedicated values for leader election 
+    durations in SNO, as disabling it can cause issues on scaling out SNO again. 
+    The whole discussion can be found at https://coreos.slack.com/archives/CDCP2LA9L/p1627402405090600. 
+    Recommended values at https://github.com/openshift/enhancements/blame/84e894ead7b188a1013556e0ba6973b8463995f1/CONVENTIONS.md#L183
+    */}}
+    election-lease-duration=137
+    election-renew-deadline=107
+    election-retry-period=26
 {{- end  }}
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -154,6 +154,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		}
 	}
 
+	if len(bootstrapResult.OVN.MasterIPs) == 1 {
+		data.Data["IsSNO"] = true
+	} else {
+		data.Data["IsSNO"] = false
+	}
+
 	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network/ovn-kubernetes"), &data)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to render manifests")
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -168,6 +168,7 @@ func TestRenderedOVNKubernetesConfig(t *testing.T) {
 		desc                string
 		expected            string
 		hybridOverlayConfig *operv1.HybridOverlayConfig
+		masterIPs           []string
 	}
 	testcases := []testcase{
 		{
@@ -193,8 +194,8 @@ enable-egress-firewall=true
 [gateway]
 mode=shared
 nodeport=true`,
+			masterIPs: []string{"1.2.3.4", "2.3.4.5"},
 		},
-
 		{
 			desc: "HybridOverlay",
 			expected: `
@@ -228,6 +229,7 @@ cluster-subnets="10.132.0.0/14"`,
 					{CIDR: "10.132.0.0/14", HostPrefix: 23},
 				},
 			},
+			masterIPs: []string{"1.2.3.4", "2.3.4.5"},
 		},
 		{
 			desc: "HybridOverlay with custom VXLAN port",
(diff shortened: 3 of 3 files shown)


creydr:
@rcarrillocruz, @tssurya could you review this PR please?

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -30,7 +30,8 @@ data:
     [gateway]
     mode={{.OVN_GATEWAY_MODE}}
     nodeport=true
-{{ if .OVNHybridOverlayEnable }}
+{{- if .OVNHybridOverlayEnable }}
+
```
tssurya, line 34: why the extra line, it doesn't work without this?

pkg/network/ovn_kubernetes.go
```diff
@@ -154,6 +154,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		}
 	}
 
+	if len(bootstrapResult.OVN.MasterIPs) == 1 {
```
tssurya, line 157: looks good, leverages the existing field in the bootstrap (https://github.com/openshift/cluster-network-operator/blob/7498fab6bdc2b872c2f09e9a35eb19794a4c0fd7/pkg/bootstrap/types.go#L37)

Q: Does single master = SNO always? Aren't the number of master nodes in OpenShift configurable in the sense, maybe we have multiple nodes but single master? I am not sure, hence asking. Is this the standard way of recognizing if a cluster is SNO or should we actually check if there is only one node in the cluster?

tssurya (commented):
Just a small doubt regarding the logic of determining if the cluster is SNO.

bindata/network/ovn-kubernetes/004-config.yaml
creydr: It's only for cosmetic. I just wanted to have like for the others, e.g. `[kubernetes]` has an empty line above as well. If I would have added the empty line before the `if`, it would might end up in multiple empty lines rendered.

pkg/network/ovn_kubernetes.go
creydr: Good question. My understanding was that Openshift has >= 3 masters and just in case of SNO only one. But I am not sure as you're asking. Asked it on [slack](https://coreos.slack.com/archives/C018KQE33MF/p1627304315423500) Any idea who can answer this as well?

creydr: @romfreiman: fyi

tssurya: Also on second thoughts, why do we do leader election in SNO? Shouldn't we just remove it all together?

bindata/network/ovn-kubernetes/004-config.yaml
tssurya: ack, my eyes just got caught on the unrelated change on that line.

pkg/network/ovn_kubernetes.go
romfreiman: @tssurya yes we should :) But we dont know how every component behaves without leader and we cannot enforce it at this point, so we dont provide such guideline. If you can/know how to remove it, feel free. IMO it's even better.

cc: @chaitanyaenr

tssurya: @romfreiman : ack yea I understand. Not sure about how other components do it, but we can see if it can be removed in ovn-k for SNOs.

romfreiman: @tssurya sounds great! btw, we have a sno weekly today if you would like to participate.

creydr: have a pr for this open in ovn-kubernetes: https://github.com/ovn-org/ovn-kubernetes/pull/2364

creydr: as pointed out by @aojea in [slack](https://coreos.slack.com/archives/CDCP2LA9L/p1627412484098900?thread_ts=1627402405.090600&cid=CDCP2LA9L) there can be scale out cases and then not having leader election will create a mess.

creydr: After the discussion on [slack](https://coreos.slack.com/archives/CDCP2LA9L/p1627402405090600), we will increase the timeouts. I will add a code comment with a link to the slack discussion

bindata/network/ovn-kubernetes/004-config.yaml
```diff
@@ -42,3 +43,16 @@ data:
     hybrid-overlay-vxlan-port="{{.OVNHybridOverlayVXLANPort}}"
     {{- end }}
 {{- end  }}
+{{- if .IsSNO }}
+
+    [masterha]
+    {{- /* 
+    Even in case of SNO there will be only one ovn-master, we set dedicated values for leader election
```
tssurya, line 50: According the [slack conversation](https://coreos.slack.com/archives/CDCP2LA9L/p1627477771126100?thread_ts=1627402405.090600&cid=CDCP2LA9L) there are concerns around increasing the leader election times as well, but going ahead with this option as the current way forward.

tssurya (commented):
/lgtm

tssurya:
> /assign @rcarrillocruz

Ricky has left the team.

creydr:
> > /assign @rcarrillocruz
> 
> Ricky has left the team. Every PR can be reviewed by all team members. The assignment is just to do a round robin.

Ah ok. Thanks.

/assign @abhat

flavio-fernandes:
Bug 1984049: Slow OVN Recovery on SNO depends on this and is needed on 4.8.
So:
/cherry-pick release-4.8

PR #1155, opened 2021-07-16
Bug 1981975: Update service network status to reflect dual stack entries

lmzuccarelli:
This addresses the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1981975

As mentioned in the BZ, if this fix is approved we will need to evaluate to backport the fix

### DIFF
--- a/pkg/util/proxyconfig/no_proxy_test.go
+++ b/pkg/util/proxyconfig/no_proxy_test.go
@@ -66,9 +66,8 @@ func infraConfig(platform configv1.PlatformType, domain, region string) *configv
 	}
 }
 
-func netConfig(cluster, svc string) *configv1.Network {
+func netConfig(cluster string, svcNet []string) *configv1.Network {
 	clusterNet := configv1.ClusterNetworkEntry{CIDR: cluster}
-	svcNet := []string{svc}
 	return &configv1.Network{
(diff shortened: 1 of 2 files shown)


pkg/util/proxyconfig/no_proxy_test.go
```diff
@@ -176,7 +197,7 @@ func TestMergeUserSystemNoProxy(t *testing.T) {
 			args: args{
 				proxy:   proxyConfigWithNoProxy("172.30.0.1."),
 				infra:   infraConfig(configv1.AWSPlatformType, "^&", "us-west-2"),
-				network: netConfig("10.128.0.0/14", ""),
+				network: netConfig("10.128.0.0/14", []string{""}),
```
frobware, line 200: Any reason to explicitly test with `[]string{""}` instead of `nil`?

lmzuccarelli: No reason, I will change it

Miciah:
/retitle Bug 1981975: Update service network status to reflect dual stack entries
Looks good, but please change the base branch from "release-4.9" to "master".  See https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request for instructions on how to do that.

PR #1158, opened 2021-07-20
Bug 1987019: Support external control plane topology 

csrwng:
When Infrastructure.Status.ControlPlaneTopology is External, the
OpenShift cluster is running with no masters. Workloads created by
the cluster network operator need to be scheduled on non-master nodes.


danwinship:
/hold

As discussed in https://coreos.slack.com/archives/CDCP2LA9L/p1626797891463000, this is incomplete because CNO still has other code that assumes ovnkube-master gets deployed to the master nodes, because it needs to tell the OVN databases (OVN, not ovn-kubernetes) what the IPs of the other db replicas are, because core OVN doesn't know how to ask Kubernetes for that information.

In general, it also feels wrong to be deploying the ovnkube-masters to worker nodes anyway. If you kill the workers that host the ovnkube-master Deployment, then _all_ of the workers in the cluster will go down. That seems contrary to the definition of "worker node"... it seems like we really need to have "infra nodes" in HyperShift. (The ovnkube-node masters may also have CPU/RAM requirements that exceed ordinary worker node requirements.)

csrwng:
@danwinship I have reverted the change to ovn-kube. It's something we'll need to resolve separately, but it'd be good if we don't block the fix for OpenShiftSDN.

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -921,7 +921,9 @@ spec:
           value: "4"
         terminationMessagePolicy: FallbackToLogsOnError
       nodeSelector:
+{{- if not .ExternalControlPlane }}
         node-role.kubernetes.io/master: ""
+{{- end }}
```
danwinship, line 926: I think it would be clearer to revert this change and the `pkg/network/ovn_kubernetes.go` change entirely, given that the DaemonSet won't actually start if you do set `ExternalControlPlane: true`.

Or maybe better, revert the `ovnkube-master.yaml` change, make `ovn_kubernetes.go` detect `ExternalControlPlane` correctly again, and make `renderOVNKubernetes` return an error if it's true, since it can't properly render the output in that case.

csrwng: Thx, reverted ovnkube-master.yaml and now return an error from `renderOVNKubernetes`

csrwng:
ci/prow/e2e-metal-ipi-ovn-ipv6

danwinship:
> I have reverted the change to ovn-kube. It's something we'll need to resolve separately, but it'd be good if we don't block the fix for OpenShiftSDN.

FTR, there is now https://issues.redhat.com/browse/SDN-2119 "change ovnkube-master DB startup to be compatible with HyperShift"

PR #1159, opened 2021-07-20
Bug 1984049: Slow OVN Recovery on SNO

flavio-fernandes:
Adding template OVN_DB_PROBE_INITIAL_DELAY parameter, which will set to 0 when there is only one master node.

Signed-off-by: Flavio Fernandes <flaviof@redhat.com>

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -374,7 +374,9 @@ spec:
                 - /var/run/ovn/ovnnb_db.ctl
                 - exit
         readinessProbe:
+{{ if not .IsSNO }}
           initialDelaySeconds: 90
+{{ end }}
           timeoutSeconds: 5
           exec:
             command:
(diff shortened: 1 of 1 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -82,6 +82,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVN_NB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_NB_PORT)
 	data.Data["OVN_SB_DB_LIST"] = dbList(bootstrapResult.OVN.MasterIPs, OVN_SB_PORT)
 	data.Data["OVN_DB_CLUSTER_INITIATOR"] = bootstrapResult.OVN.ClusterInitiator
+	if len(bootstrapResult.OVN.MasterIPs) == 1 {
```
flavio-fernandes, line 85: As pointed out in the bug dispatch meeting, we seem to be doing this "len(bootstrapResult.OVN.MasterIPs)" call multiple times in the function. Ref: https://github.com/openshift/cluster-network-operator/pull/1154/commits 
Maybe we should convert it ti a local variable?

flavio-fernandes:
Rebasing to [PR 1154](https://github.com/openshift/cluster-network-operator/pull/1154) to remove duplicate logic in determining the number of masters.

PR #1160, opened 2021-07-21
Bug 1984449: Change to use mountPath: /host

mccv1r0:
Change to use mountPath: /host


mffiedler:
1.  Reproduced on 4.7.latest on Azure.   Did not even have to reboot - ssh hung straight away
2. Built a cluster-bot cluster from this PR.   SSH worked fine before and after reboot

/label qe-approved

PR #1161, opened 2021-07-23
Bug 1976399: Raft election timer: move the logic to ovndbchecker

fedepaol:
The current logic checks the change of election timer in the postStart
hook, checking if the current node is the leader and in case applying
the desired change. When rolling out an upgrade, each ovnkube-master pod
is updated one at the time, with the result that when postStart
executes, the current node is never the leader. Here, we move the logic
to ovndbchecker (changed in https://github.com/ovn-org/ovn-kubernetes/pull/2356) to
reconcile the election timer asynchronously.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -308,61 +308,6 @@ spec:
                   done
                 fi
 
-                #configure NB RAFT election timers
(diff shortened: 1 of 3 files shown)


fedepaol:
/hold
This depends on when https://github.com/ovn-org/ovn-kubernetes/pull/2356 lands d/s

fedepaol:
U/s merged, depends on https://github.com/openshift/ovn-kubernetes/pull/647

PR #1169, opened 2021-07-30
Bug 1988425: Change to use mountPath: /host

mccv1r0:
Change iptables chain names to not conflict with existing workaround(s)

(cherry picked from commit 32d72e5794ecd5e5f742c8708a2e4a6c183b0e4b)

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -254,25 +254,25 @@ spec:
                exit 0
           fi
           echo "Adding ICMP drop rule for '$3' "
-          if iptables -C CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+          if iptables -C AZURE_CHECK_ICMP_SOURCE -p icmp -s $3 -j AZURE_ICMP_ACTION
           then
                echo "iptables already set for $3"
           else
-               iptables -A CHECK_ICMP_SOURCE -p icmp -s $3 -j ICMP_ACTION
+               iptables -A AZURE_CHECK_ICMP_SOURCE -p icmp -s $3 -j AZURE_ICMP_ACTION
           fi
           EOF
 
           echo "I$(date "+%m%d %H:%M:%S.%N") - drop-icmp - start drop-icmp ${K8S_NODE}"
-          iptables -X CHECK_ICMP_SOURCE || true
-          iptables -N CHECK_ICMP_SOURCE || true
-          iptables -F CHECK_ICMP_SOURCE
-          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE || true
-          iptables -I INPUT -p icmp --icmp-type fragmentation-needed -j CHECK_ICMP_SOURCE
-          iptables -N ICMP_ACTION || true
-          iptables -F ICMP_ACTION
-          iptables -A ICMP_ACTION -j LOG
-          iptables -A ICMP_ACTION -j DROP
-          oc observe pods -n openshift-sdn -l app=sdn -a '{ .status.hostIP }' -- /var/run/add_iptables.sh
+          iptables -X AZURE_CHECK_ICMP_SOURCE || true
+          iptables -N AZURE_CHECK_ICMP_SOURCE || true
+          iptables -F AZURE_CHECK_ICMP_SOURCE
+          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j AZURE_CHECK_ICMP_SOURCE || true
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -254,33 +254,33 @@ spec:
...
+          iptables -X AZURE_CHECK_ICMP_SOURCE || true
+          iptables -N AZURE_CHECK_ICMP_SOURCE || true
+          iptables -F AZURE_CHECK_ICMP_SOURCE
+          iptables -D INPUT -p icmp --icmp-type fragmentation-needed -j AZURE_CHECK_ICMP_SOURCE || true
+          iptables -I INPUT -p icmp --icmp-type fragmentation-needed -j AZURE_CHECK_ICMP_SOURCE
+          iptables -N AZURE_ICMP_ACTION || true
+          iptables -F AZURE_ICMP_ACTION
+          iptables -A AZURE_ICMP_ACTION -j LOG
+          iptables -A AZURE_ICMP_ACTION -j DROP
+          /host/usr/bin/oc observe pods -n openshift-sdn -l app=sdn -a '{ .status.hostIP }' -- /var/run/add_iptables.sh
```
danwinship, line 275: This seems not safe; is `oc` statically linked, and is it guaranteed to _always_ be statically linked? I think you have to do `chroot /host oc ...` though of course then you also have to write the script out to `/home/var/run` rather than `/var/run`

danwinship (commented):
It's confusing that the commit message summary only describes one of the two things the commit does. Either split it into two commits, or else make the summary more generic ("fixes to azure iptables setup script" etc) and describe both fixes in the commit message body, rather than one in the summary and one in the body

bindata/network/openshift-sdn/sdn.yaml
cgwalters: I think we've glossed over this so far but due to FIPS needing openssl for Go, we can't rely on these things being fully static.

cgwalters: The other pattern is to use e.g. `systemd-run -qPG -- oc observe pods -n openshift-sdn -l app=sdn -a '{ .status.hostIP }' -- /var/run/add_iptables.sh`.

mccv1r0: @mjudeikis fyi

mccv1r0: The commit message already describes: 

1. Change to use mountPath: /host
2. Change iptables chain names to not conflict with existing workaround(s)

For SDN, I'll probably need to change upstream to use systemd-run (or whatever ends up being used) and back port.

The OVN version of this fix will start with whatever consensus decides is best.

mccv1r0:
@knobunc Can you override this required gcp test?  As you know this PR changes yaml that runs on Azure only.

brenton:
Is this PR opened against the correct branch?  I would have expected it to merge into release-4.8.  When I look now the master, release-4.10 and release-4.9 branches all already have the fix from https://bugzilla.redhat.com/show_bug.cgi?id=1984449.

abhat:
@mccv1r0 since this is a master branch PR, the bz needs retitling?

mccv1r0:
> @mccv1r0 since this is a master branch PR, the bz needs retitling?

The PR needs to be against 4.8 <g>

dcbw:
It appears that ARO already uses /host and it works there, so for now:
/lgtm
/approve

mffiedler:
Verified as with #1160.   4.8.z cluster-bot cluster built from this PR.   ssh into node works fine both before and after boot.   Without this PR ssh just hangs.

/lgtm
/label qe-approved

dcbw:
/override ci/prow/e2e-azure-ovn-dualstack

Error on cloud setup:
`level=error msg=Error: updating Load Balancer "ci-op-5hhdw15j-f5292-k6tnt" (Resource Group "ci-op-5hhdw15j-f5292-k6tnt-rg") for Probe "api-internal-probe": network.LoadBalancersClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="PrivateIPAddressVersionCannotBeModified" Message="Load balancer frontend IP configuration /subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-5hhdw15j-f5292-k6tnt-rg/providers/Microsoft.Network/loadBalancers/ci-op-5hhdw15j-f5292-k6tnt/frontendIPConfigurations/public-lb-ip-v6 has an existing private IP address version IPv6 and the request has private IP address version IPv4, which do not match. Private IP address version cannot be modified once the resource is created." Details=[]`

PR #1172, opened 2021-08-02
openshift-sdn/daemonset: Mount /host/opt/cni/bin at /host-cni-bin

cgwalters:
For some as yet undebugged reason, the theory is this mount setup seems to
not work with the CentOS 9 Stream kernel at least.
https://bugzilla.redhat.com/show_bug.cgi?id=1988520

Here we're avoiding "overmounting" on `/host` inside the container;
instead create `/host-cni-bin` which contains the in-container
view of the host's `/opt/cni/bin` (which is `/var/opt/cni/bin` on ostree
systems).

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -85,7 +85,7 @@ spec:
 
           # Take over network functions on the node
           rm -f /etc/cni/net.d/80-openshift-network.conf
-          cp -f /opt/cni/bin/openshift-sdn /host/opt/cni/bin/
+          cp -f /opt/cni/bin/openshift-sdn /host-cni-bin/
 
           # Launch the network process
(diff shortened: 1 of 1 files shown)


cgwalters:
I'm testing this manually now with the current c9s kernel.

cgwalters:
Confirmed this fixes the issue, see https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-aws/1422225736164446208 which at least has the sdn daemonset starting.

cgwalters:
> hang on, have a thought here.

:thinking: ?

ashcrow:
Any progress on this?

danwinship:
/retitle openshift-sdn/daemonset: Mount /host/opt/cni/bin at /host-cni-bin
/lgtm

PR #1173, opened 2021-08-02
Bug 1985486: Use proxy to connect to OSP cloud

MaysaMacedo:
When using Proxy and the installer is configured
on a restricted installation, Kuryr bootstrap
can't progress as the http client is not
configured to use Proxy. This commit ensures
proxy is used by relying on the cluster proxy
resource instead of relying on cvo to injecting
the proxy env vars.

### DIFF
--- a/pkg/platform/openstack/kuryr_bootstrap.go
+++ b/pkg/platform/openstack/kuryr_bootstrap.go
@@ -6,25 +6,28 @@ import (
 	"crypto/x509"
 	b64 "encoding/base64"
 	"fmt"
+	"golang.org/x/net/http/httpproxy"
 	"log"
 	"net"
 	"net/http"
+	"net/url"
+	"os"
 
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/Masterminds/semver"
-	"github.com/pkg/errors"
-	"gopkg.in/yaml.v2"
-
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/utils/openstack/clientconfig"
+	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/platform/openstack/util/cert"
+	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -283,6 +286,21 @@ func getWorkersSubnetFromMasters(client *gophercloud.ServiceClient, kubeClient c
...
+	auth_url, err := url.Parse(auth)
+	if err != nil {
+		return false, err
+	}
+	no_proxy_addresses := strings.Split(no_proxy, ",")
+	for _, address := range no_proxy_addresses {
+		if auth_url.Hostname() == address {
+			return false, nil
+		}
+	}
```
dulek, line 300: Uhm, the logic might need to be more complicated. Something like `.openshift.org` should probably match `foo.openshift.org`.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -330,17 +348,41 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
-					},
-				},
+			transport.TLSClientConfig = &tls.Config{RootCAs: certPool}
+		}
+		client := http.Client{Transport: &transport}
+		provider.HTTPClient = client
+	}
+
+	// We cannot rely on the inject-proxy annotation because the CVO, which is
+	// responsible to inject the proxy env vars, runs on Pod Network.
```
dulek, line 363: You're sure cluster-version-operator doesn't run on host networking? It's the first operator there is.

dulek (changes requested):

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: yeah, it hanged on containerCreating and the ip was not one of the hosts

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: true, thanks

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -331,19 +334,40 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
-				Transport: &http.Transport{
-					TLSClientConfig: &tls.Config{
-						RootCAs: certPool,
-					},
-				},
-			}
-			provider.HTTPClient = client
+			transport.TLSClientConfig = &tls.Config{RootCAs: certPool}
 		}
+		provider.HTTPClient = http.Client{Transport: &transport}
```
dulek, line 370: Does it mean that now we can either have proxy or certs? Is it desired?

MaysaMacedo: Yes, we can have either of them, e.g. Cloud configured without proxy and without self signed certs, or proxy with only http and not https or ssc and proxy.
Note that when the code arrives here the Transport will be overwritten to contain both proxy and certs or just certs

MaysaMacedo: I just noticed this line should be at 368

dulek: This is true, thanks!

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -331,18 +334,54 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse cluster-wide proxy https URL")
+		}
+
+		if proxyUrl.Scheme == "https" {
+			log.Printf("Kuryr only supports cluster-wide http scheme.")
+			if len(httpProxy) > 0 {
+				log.Printf("Defaulting cluster-wide proxy to %s", httpProxy)
+				httpsProxy = httpProxy
+			}
```
dulek, line 375: Should we error out on `else` here?

MaysaMacedo: yeah, good point

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -331,18 +334,56 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+		proxyUrl, err := url.Parse(httpsProxy)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to parse cluster-wide proxy https URL")
+		}
+
+		if proxyUrl.Scheme == "https" {
+			log.Printf("Kuryr only supports cluster-wide http scheme.")
+			if len(httpProxy) > 0 {
+				log.Printf("Defaulting cluster-wide proxy to %s", httpProxy)
+				httpsProxy = httpProxy
```
danwinship, line 374: `HTTP_PROXY` can be `https` too... in fact, `HTTP_PROXY` and `HTTPS_PROXY` are almost always going to be exactly the same, so this check is probably useless.

pkg/platform/openstack/kuryr_bootstrap.go
```diff
@@ -331,18 +334,56 @@ func BootstrapKuryr(conf *operv1.NetworkSpec, kubeClient client.Client) (*bootst
...
+			} else {
+				return nil, errors.New("failed to configure cluster-wide proxy for Kuryr. No httpProxy provided.")
```
danwinship, line 376: which makes this error wrong/confusing. You should say that Kuryr currently requires that the proxy be "http" rather than "https". (Is this because of the urllib3 bug above? In which case it should hopefully eventually be fixed?)

I assume returning an error here will cause CNO to become Degraded and thus cause the install to fail? Because if they configured a proxy and you can't support it, that's what you need to do.

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: It seems CNO is only allowing `http` scheme at `HTTP_PROXY`
https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/proxyconfig/validation.go#L42-L47

MaysaMacedo: And the validation of install-config produces the following:
`"Install Config": invalid "install-config.yaml" file: proxy.httpProxy: Unsupported value: "https": supported values: "http"`

pkg/platform/openstack/kuryr_bootstrap.go
MaysaMacedo: Yes, it's because of urllib3 issue. The fix was merged, but there is no release with it yet.
CNO would become degraded once this error arises.

pkg/platform/openstack/kuryr_bootstrap.go
danwinship: huh, ok, that's interesting. In that case, I guess this code is reasonable

PR #1174, opened 2021-08-03
Bug 1989734: Whereabouts should have RBAC for leases

dougbtv:
Newer feature in Whereabouts uses leases for coordination of leaders as to not cause a race condition in ip allocations.

### DIFF
--- a/bindata/network/multus/002-rbac.yaml
+++ b/bindata/network/multus/002-rbac.yaml
@@ -79,6 +79,21 @@ subjects:
   namespace: openshift-multus
 
 ---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: multus-whereabouts
+  namespace: openshift-multus
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: whereabouts-cni
+subjects:
+- kind: ServiceAccount
+  name: multus
+  namespace: openshift-multus
+
+---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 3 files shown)


dougbtv:
/title Bug 1989734: Whereabouts should have RBAC for leases

dougbtv:
/retitle Bug 1989734: Whereabouts should have RBAC for leases

bindata/network/multus/002-rbac.yaml
```diff
@@ -97,3 +97,10 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
```
squeed, line 100: Just to confirm, this is a ClusterRole - you want access to the whole cluster?

dougbtv: Great eye! Thanks. Upstream, the namespace where these live is configurable, so it's convenient to have it that way upstream, but... We pin the namespace for leases in openshift, so we can use a role for this binding.

Really appreciate that. New commit incoming with a `Role` :+1:

dougbtv: Also, fwiw... we DO want a `clusterrole` for the other permissions, as they can live in differing namespaces.

dougbtv:
Sorry for the slew of commits :/ PEBKAC, I had an out-of-sync fork and I didn't trust my instruments (*cough* github merge robot *cough*) and just kept doing dumb things. This should be the one after I had a proper rebase :+1:

dougbtv:
/test e2e-gcp-ovn

unsure if this is expected to be busted, but it's the only failing required test.

knobunc:
/approve

@squeed please can you take another look and make sure your concerns are addressed

PR #1177, opened 2021-08-09
Bug 1991551: allow sdn (and others) to use new events.k8s.io API

danwinship:
Events are slowly being moved from `core/v1` to `events.k8s.io/v1`, so update everything to be able to use both namespaces for now.

Also, while we're here, drop the rules for the old `extensions/v1beta1` NetworkPolicy API.

/hold
need to test manually as the bug is not currently caught by CI


danwinship:
/hold cancel
this is necessary, but so is https://github.com/openshift/sdn/pull/332

PR #1197, opened 2021-09-14
[WIP] Bug 2003676: Restrict serving metrics to loopback only

martinkennelly:
Previously to this change, we served on all
interfaces.

/cc martinkennelly


alexanderConstantinescu (approved):
/lgtm

@danwinship : could you have a look? Do you happen to know if there was a specific reason it was done like this, or if it's simply a bug we should fix?

danwinship:
> @danwinship : could you have a look? Do you happen to know if there was a specific reason it was done like this, or if it's simply a bug we should fix?

It's just a bug; all the metrics used to be exposed without auth, and when we changed that for sdn (#574) we messed up and added an authenticated endpoint without removing the unauthenticated one :man_facepalming: 

/approve

PR #1204, opened 2021-10-05
NETOBSERV-31: Expose CNI type features as a config-map

mariomac:
Related enhancement: [Expose Network features](https://github.com/openshift/enhancements/blob/master/enhancements/console/expose-network-features.md)

The console requires to know the network type capabilities to show/hide some Network Policy form fields.

As agreed in the linked enhancement document, the required network capabilities will be added to an `sdn-public` `ConfigMap`, created by the CNO, readable by any `system:authenticated` user.

This PR is also the placeholder to add future network capabilities, as long as our Console components require more knowledge about the underlying CNI.

We provide the known values to `OVNKubernetes` and `OpenShiftSDN` CNI types. If the CNI type is `Kuryr` or any other, we will take default actions (e.g. show the related form field with a warning message).

We restrict the access control for the new Role as much as possible. That means that a non-administrative user won't be able to read the configmap with the `oc describe configmap sdn-public` as long as  this command requires access to the `events` resource type and we don't grant it. Also, the created ConfigMap won't be listed in the non-admin user UI (but visible if you paste the URL in the browser bar)

Direct access by means of the REST API is possible, as it is the only required form of accessing this map.

### DIFF
--- a/bindata/network/openshift-sdn/cni-features.yaml
+++ b/bindata/network/openshift-sdn/cni-features.yaml
@@ -1,0 +1,12 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openshift-network-features
+  namespace: openshift-config-managed
+  annotations:
+    openshift.io/description: |
+      Exposes available network features as required by the Console in order to show or hide some form fields.
+      If the map or a given property is undefined, the Console won't throw error and will take a default action (show, hide, show with a warning message...).
+data:
+  policy_egress: "false"
+  policy_peer_ipblock_exceptions: "false"
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -72,6 +72,12 @@ func Render(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult
 	}
 	objs = append(objs, o...)
 
+	o, err = renderNetworkPublic(manifestDir)
+	if err != nil {
+		return nil, err
+	}
+	objs = append(objs, o...)
+
 	log.Printf("Render phase done, rendered %d objects", len(objs))
 	return objs, nil
 }
@@ -613,5 +619,16 @@ func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*
 		return nil, errors.Wrap(err, "failed to render network-diagnostics manifests")
 	}
 
+	return manifests, nil
+}
+
+// renderNetworkPublic renders the common objects related to the openshift-network-features configmap
+func renderNetworkPublic(manifestDir string) ([]*uns.Unstructured, error) {
+	data := render.MakeRenderData()
+
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network", "public"), &data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render network/public manifests")
+	}
 	return manifests, nil
 }
--- a/bindata/network/ovn-kubernetes/cni-features.yaml
+++ b/bindata/network/ovn-kubernetes/cni-features.yaml
@@ -1,0 +1,12 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openshift-network-features
+  namespace: openshift-config-managed
+  annotations:
+    openshift.io/description: |
+      Exposes available network features as required by the Console in order to show or hide some form fields.
+      If the map or a given property is undefined, the Console won't throw error and will take a default action (show, hide, show with a warning message...).
+data:
+  policy_egress: "true"
+  policy_peer_ipblock_exceptions: "true"
--- a/bindata/network/public/000-roles.yaml
+++ b/bindata/network/public/000-roles.yaml
@@ -1,0 +1,31 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: openshift-network-public-role
+  namespace: openshift-config-managed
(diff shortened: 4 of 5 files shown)


bindata/network/openshift-sdn/sdn-public.yaml
```diff
@@ -0,0 +1,13 @@
+kind: ConfigMap
+apiVersion: v1
+immutable: true
```
martinkennelly, line 3: Everything in this configmap will be generated once and left alone for the life time of the cluster? With this feature, you cannot edit the configmap. What happens if a CNI gets updated with a new feature that contradicts the existing configmap and we have to update this configmap? You can delete it and recreate it I guess.

bindata/network/openshift-sdn/sdn-public.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: sdn-public
```
martinkennelly, line 5: I am not a fan of the name. ```openshift-sdn-features``` maybe? WDYT?

bindata/network/ovn-kubernetes/sdn-public.yaml
```diff
@@ -0,0 +1,13 @@
+kind: ConfigMap
+apiVersion: v1
+immutable: true
+metadata:
+  name: sdn-public
```
martinkennelly, line 5: To us, SDN means Openshift-SDN. It may confuse us and uses to name the configmap this name.

bindata/network/sdn-public/000-roles.yaml
```diff
@@ -0,0 +1,31 @@
...
+  annotations:
+    openshift.io/description: "Grants access from any authenticated user to the sdn-public ConfigMap"
+subjects:
+  - kind: Group
+    apiGroup: rbac.authorization.k8s.io
+    name: 'system:authenticated'
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: sdn-public-role
```
martinkennelly, line 31: Missing newline

pkg/network/render.go
```diff
@@ -615,3 +621,15 @@ func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*
 
 	return manifests, nil
 }
+
+// renderSdnPublic renders the common objects related to the sdn-public namespace
+func renderSdnPublic(manifestDir string) ([]*uns.Unstructured, error) {
+	data := render.MakeRenderData()
+	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
```
martinkennelly, line 628: May I ask why you need to do this?

pkg/network/render.go
```diff
@@ -615,3 +621,15 @@ func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*
...
+
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network", "sdn-public"), &data)
```
martinkennelly, line 630: Not a fan of the "sdn-public" dir name because of associations with OpenShift SDN when there are other CNIs.

astoycos:
@mariomac  Can you please squash your commits here? (No need to show the responses to Martin's comments :)

mariomac:
@astoycos squashed. Thanks!

pkg/network/render.go
```diff
@@ -615,3 +621,14 @@ func renderNetworkDiagnostics(conf *operv1.NetworkSpec, manifestDir string) ([]*
 
 	return manifests, nil
 }
+
+// renderSdnPublic renders the common objects related to the openshift-network-features configmap
+func renderSdnPublic(manifestDir string) ([]*uns.Unstructured, error) {
```
abhat, line 626: nit: `s/renderSdnPublic/renderNetworkPublic`

bindata/network/openshift-sdn/cni-features.yaml
```diff
@@ -0,0 +1,13 @@
...
+immutable: true
+metadata:
+  name: openshift-network-features
+  namespace: openshift-config-managed
+  annotations:
+    openshift.io/description: |
+      Exposes available network features as required by the Console in order to show or hide some form fields.
+      If the map or a given property is undefined, the Console won't throw error and will take a default action (show, hide, show with a warning message...).
+data:
+  policy_egress: "false"
```
abhat, line 12: What does this data correspond to? Is this just free-form or backed by an enumeration for network features?

mariomac: The names have been chosen freely, though they correspond to real network features. In the current entries:
* `policy_egress`: whether the CNI type supports egress policies.
* `policy_peer_ipblock_exceptions`: whether the CNI type supports setting exceptions to a policy based on peer IP block.

The network features are for internal use (currently, the [Console network policy forms](https://github.com/openshift/console/blob/a28af190fe9bca902b4398b50ee38e89fbdc0f2f/frontend/public/module/k8s/network.ts#L34-L46), but any other naming that fits better your team conventions is accepted.

danwinship:
@mariomac Doh. I hadn't know about this PR. As of a few hours ago, this is obsolete; openshift-sdn also implements both egress rules and ipBlock exceptions.

mariomac:
@danwinship thank you, we will update our maps ASAP

mariomac:
Updating network capabilities here: https://github.com/openshift/cluster-network-operator/pull/1251

PR #1206, opened 2021-10-12
Specific SDN controller alert

martinkennelly:
Previous to this change, we are overloading the meaning of "SDN pod" to include SDN controller. This PR removes SDN controller and adds a separate alert for SDN controller.

~~Looking for a better metric than ```cluster:master_nodes``` to get the total number of control plane nodes. It comes from container ```kube-rbac-proxy-main```..~~

edit: Got advice from monitoring team to use kube_node_role metric.


martinkennelly:
@danwinship We talked about this previously. If you've any time to review, please do.

PR #1207, opened 2021-10-13
Bug 1914053: whereabouts: add ip-reconciler cronjob

maiqueb:
Signed-off-by: Miguel Duarte Barroso <mdbarroso@redhat.com>

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -455,3 +455,41 @@ spec:
           configMap:
             name: cni-binary-copy-script
             defaultMode: 0744
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: ip-reconciler
+  namespace: openshift-multus
+  labels:
+    tier: node
+    app: whereabouts
+spec:
+  # reconcile loop every 5 minutes, starting at the top of the hour
+  schedule: "*/5 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          serviceAccountName: multus
+          containers:
+            - name: whereabouts
+              image: {{.WhereaboutsImage}}
+              resources:
+                requests:
+                  cpu: "100m"
+                  memory: "50Mi"
+                limits:
+                  cpu: "100m"
+                  memory: "50Mi"
+              command:
+              - /ip-reconciler
+              - -kubeconfig=/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
+              volumeMounts:
+                - name: cni-net-dir
+                  mountPath: /host/etc/cni/net.d
+          volumes:
+            - name: cni-net-dir
+              hostPath:
+                path: {{ .SystemCNIConfDir }}
+          restartPolicy: OnFailure
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -471,20 +471,18 @@ spec:
     spec:
       template:
         spec:
+          priorityClassName: "system-cluster-critical"
           serviceAccountName: multus
           containers:
             - name: whereabouts
               image: {{.WhereaboutsImage}}
               resources:
                 requests:
-                  cpu: "100m"
-                  memory: "50Mi"
-                limits:
-                  cpu: "100m"
-                  memory: "50Mi"
+                  cpu: "25m"
+                  memory: "25Mi"
               command:
               - /ip-reconciler
-              - -kubeconfig=/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
+              - -log-level=verbose
               volumeMounts:
                 - name: cni-net-dir
                   mountPath: /host/etc/cni/net.d
--- a/bindata/network/multus/002-rbac.yaml
+++ b/bindata/network/multus/002-rbac.yaml
@@ -112,6 +112,11 @@ rules:
   - update
   - patch
   - delete
+- apiGroups: [""]
+  resources:
+  - pods
+  verbs:
+  - list
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/network/multus_test.go
+++ b/pkg/network/multus_test.go
@@ -50,13 +50,14 @@ func TestRenderMultus(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-multus", "multus")))
 
 	// It's important that the namespace is first
-	g.Expect(len(objs)).To(Equal(22))
+	g.Expect(len(objs)).To(Equal(23), "Expected 23 multus related objects")
 	g.Expect(objs[0]).To(HaveKubernetesID("CustomResourceDefinition", "", "network-attachment-definitions.k8s.cni.cncf.io"))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("Namespace", "", "openshift-multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRole", "", "multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ServiceAccount", "openshift-multus", "multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("ClusterRoleBinding", "", "multus")))
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("DaemonSet", "openshift-multus", "multus")))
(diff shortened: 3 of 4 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,142 @@ spec:
...
+
+                # ---------------------- end Generate a "kube-config".
+
+                # Unless told otherwise, sleep forever.
+                # This prevents Kubernetes from restarting the pod repeatedly.
+                should_sleep=${SLEEP:-"true"}
+                echo "Done configuring CNI.  Sleep=$should_sleep"
+                while [ "$should_sleep" == "true"  ]; do
+                    sleep 1000000000000
+                done
```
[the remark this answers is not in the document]
maiqueb: Ah! I found it available on the correct path, which is actually referenced by the `SystemCNIConfDir` variable:
https://github.com/openshift/cluster-network-operator/blob/87d4207091c6a6aed36ca9cf6f190c79c9a8d456/pkg/network/multus.go#L14

But, the file is available on host with r/w permissions only to the root user:
```
sh-4.4# ls -la /etc/kubernetes/cni/net.d/whereabouts.d/
total 16
drwxr-xr-x. 2 root root    60 Oct 14 09:14 .
drwxr-xr-x. 4 root root    65 Oct 14 09:30 ..
-rw-r--r--. 1 root root   162 Oct 14 09:29 whereabouts.conf
-rw-------. 1 root root 11559 Oct 14 09:29 whereabouts.kubeconfig

```
Which causes our humble pod to not be able to access it:
```
oc logs -nopenshift-multus ip-reconciler-27236755--1-zzwlw
2021-10-14T09:55:44Z [debug] NewReconcileLooper - Kubernetes config file located at: /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
2021-10-14T09:55:44Z [error] failed to instantiate the Kubernetes client: Error loading config file "/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig": open /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig: permission denied
2021-10-14T09:55:44Z [error] failed to create the reconcile looper: failed to instantiate the Kubernetes client: Error loading config file "/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig": open /host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig: permission denied

```

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,142 @@ spec:
...
+
+                /ip-reconciler -kubeconfig=/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
```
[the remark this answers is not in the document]
maiqueb: You mean I do not need to mount the entire cni config folder, right ? Just the particular k8s config file. 

That's a good point. Once we decide how we want to do this, I'll update it to point at this particular file.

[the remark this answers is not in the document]
dougbtv: +1, you can use just the file.

bindata/network/multus/multus.yaml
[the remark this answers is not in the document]
dougbtv: Hrmmm, I'm a little torn. So, all the other CNI components herein are privileged. So it's not... weird. 

Maybe it's even appropriate. Because if we make that file more easily readable, then, that service account credentials are exposed.

dougbtv:
/title Bug 1914053: whereabouts: add ip-reconciler cronjob

dougbtv:
/retitle Bug 1914053: whereabouts: add ip-reconciler cronjob

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,42 @@ spec:
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: ip-reconciler
+  namespace: openshift-multus
+  labels:
+    tier: node
+    app: whereabouts
+spec:
+  schedule: "*/5 * * * *"
```
danwinship, line 468: Can you add a comment before this explaining what it means? I always forget which cronjob field is which...

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,42 @@ spec:
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          priorityClassName: "system-node-critical"
```
danwinship, line 473: This seems maybe incorrect for this job? I forget exactly what this field does, but it seems like a periodic cleanup job is not "system critical".

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,42 @@ spec:
...
+          securityContext:
+            privileged: true
```
danwinship, line 475: `privileged` is only defined in the container-level `securityContext`, not the pod-level. So this has no effect

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,42 @@ spec:
...
+                  cpu: "25m"
+                  memory: "25Mi"
+              command:
+              - /ip-reconciler
+              - -kubeconfig=/host/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig
+              volumeMounts:
+                - name: cni-net-dir
+                  mountPath: /host/etc/cni/net.d
+              securityContext:
+                privileged: true
```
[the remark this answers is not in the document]
maiqueb: The reasons for it are described in https://github.com/openshift/cluster-network-operator/pull/1207/commits/fdf46947684e41f7b3ae97dcda377d2924a8ca8c#r728837192 . 

This pod does **not** require to be privileged - it could regenerate the kubeconfig file instead. 

Your call.

bindata/network/multus/multus.yaml
danwinship: As I said, I don't remember where the exact definition of how to deal with this field is... (maybe something in openshift/enhancements/dev-guides says? If not, probably there _should_ be something there).

But, for example, if a node is at max capacity, and it's time to run this job, do you want kubernetes to kill user workload pods to make room for this pod? I'm guessing no, in which case this probably needs to have different priority characteristics than most other network-related pods.

danwinship:
You don't need an actual kubeconfig when you're running as a pod; call `rest.InClusterConfig()` from `k8s.io/client-go/rest` and it will use `KUBERNETES_SERVICE_HOST` and the service token, etc, to generate an appropriate kubeconfig for you. (You can just make the existing `kubernetes.NewClient` in whereabouts try both ways; `InClusterConfig` will return a specific error you can check for if you call it from outside a pod.)

dougbtv:
/hold

waiting for https://github.com/openshift/whereabouts-cni/pull/73 to be merged so we have the downstream integration

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,39 @@ spec:
...
+kind: CronJob
+metadata:
+  name: ip-reconciler
+  namespace: openshift-multus
+  labels:
+    tier: node
+    app: whereabouts
+spec:
+  # reconcile loop every 5 minutes, starting at the top of the hour
+  schedule: "*/5 * * * *"
```
danwinship, line 469: Oh, huh, I was thinking it meant "run 5 minutes after the hour", not "run every 5 minutes".

In the long term we should think about whether it makes more sense to just have some existing long-running pod handle this, rather than spinning up a pod every 5 minutes... (Maybe CNO could do it?)

bindata/network/multus/multus.yaml
```diff
@@ -455,3 +455,39 @@ spec:
...
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          priorityClassName: "system-cluster-critical"
```
danwinship, line 474: Did you actually find any documentation about this field and how to pick a value for it? (We should probably link to that from the wiki so that next time we add something we know where to look...)

danwinship (approved):
/lgtm
though I'd still like more info on `priorityClassName`...

bindata/network/multus/multus.yaml
maiqueb: Turning whereabouts into a thick plugin (long running pod) and migrate the ip cleanup reconcile loop there is one of our goals. I can't speak for **when** though... But it's on our radar.

bindata/network/multus/multus.yaml
maiqueb: Unfortunately I didn't. 

I went with the intermediate value. Having said that, I agree that it would be very helpful for the next time it is needed.

PR #1208, opened 2021-10-21
fix a typo in a field name

danwinship:
(It's spelled correctly in the bindata templates)

### DIFF
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -33,7 +33,7 @@ type KuryrBootstrapResult struct {
 type OVNConfigBoostrapResult struct {
 	GatewayMode            string
 	EnableEgressIP         bool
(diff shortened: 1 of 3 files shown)


pkg/bootstrap/types.go
```diff
@@ -33,7 +33,7 @@ type KuryrBootstrapResult struct {
 type OVNConfigBoostrapResult struct {
 	GatewayMode            string
 	EnableEgressIP         bool
-	DisableSNATMutlipleGWs bool
```
tssurya, line 36: wow! I had to stare at it for a good 5 seconds before realizing what the error was. Nice catch, at least they were consistently wrong and hence worked fine (I had spent sometime this week testing this out and this never caught my eye!) nice catch Dan.

PR #1209, opened 2021-10-25
SDN 2316: Use GatewayConfig in OVN-K to set gateway modes

tssurya:
Depends On https://github.com/openshift/api/pull/1033 and https://github.com/openshift/cluster-network-operator/pull/1242

In this PR, we are doing two things:

- Removing the exclusiveness between EgressIPs and ICNI
- Removing the hidden config-map and using the new api `routingViaHost` to configure gateway modes.

We also handle the upgrade logic of populating the new API field in the cluster CRD based on the `gateway-mode-config` map values. Also note that this enables users to migrate between the two gateway modes at runtime.


tssurya:
I have tested this PR against a cluster and it works as expected.
```
spec:
  clusterNetwork:
  - cidr: 10.128.0.0/14
    hostPrefix: 23
  defaultNetwork:
    ovnKubernetesConfig:
      gatewayConfig:
        routingViaHost: true
      genevePort: 6081
      mtu: 8901
      policyAuditConfig:
        destination: "null"
        maxFileSize: 50
        rateLimit: 20
        syslogFacility: local0
    type: OVNKubernetes

```

It rolls out local gateway as soon as we change the gatewayConfig and add the `routingViaHost` config parameter.

pkg/network/ovn_kubernetes.go
```diff
@@ -473,6 +458,36 @@ type replicaCountDecoder struct {
...
+		if modeOverride != OVN_SHARED_GW_MODE && modeOverride != OVN_LOCAL_GW_MODE {
+			klog.Warningf("gateway-mode-config does not match %q or %q, is: %q. Using default gateway mode: %s",
+				OVN_LOCAL_GW_MODE, OVN_SHARED_GW_MODE, modeOverride, OVN_SHARED_GW_MODE)
+			modeOverride = OVN_SHARED_GW_MODE
+		}
+	}
+	if modeOverride == OVN_LOCAL_GW_MODE {
+		routeViaHost = true
+	}
+	conf.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig = &operv1.GatewayConfig{
```
trozet, line 485: This doesn't actually update the config map right? Do we need to update the config map from here so that when we remove the ability to use the hidden configmap we dont break users? Or should we print some warning or event to tell the user they need to update the gatewayConfig in their configmap?

tssurya: We do update the configmap when this function returns. Let me explain:

1. https://github.com/openshift/cluster-network-operator/blob/38b07c3f025b0ab7dad6b5480605b20976fe1e47/pkg/network/bootstrap.go#L18
2. https://github.com/openshift/cluster-network-operator/blob/38b07c3f025b0ab7dad6b5480605b20976fe1e47/pkg/network/ovn_kubernetes.go#L485
3. https://github.com/openshift/cluster-network-operator/blob/38b07c3f025b0ab7dad6b5480605b20976fe1e47/pkg/network/ovn_kubernetes.go#L284
4. Within `bootstrapOVNConfig` we do our thing. 
5. Then we return all the way back to the bootstrap call https://github.com/openshift/cluster-network-operator/blob/38b07c3f025b0ab7dad6b5480605b20976fe1e47/pkg/controller/operconfig/operconfig_controller.go#L192
6.  Aaaand this is where we update the CRD: https://github.com/openshift/cluster-network-operator/blob/38b07c3f025b0ab7dad6b5480605b20976fe1e47/pkg/controller/operconfig/operconfig_controller.go#L200-L201

Hence any changes done to the CRD during bootstrap gets written into the CRD. After all that the rendering starts and it regards the updated CRD the source of truth.


I tested it using the steps mentioned in https://docs.google.com/document/d/1SYaNQNW4sWVTE9XgHG6gKsS4ZActK8_Z7wILXU5YuoU/edit#heading=h.d5s838mpuww0 in a cluster-bot cluster and it works well. The CRD does get updated to something like:
```
Gateway Config:{
    Routing Via Host: true
}
```

trozet (approved):
/lgtm

PR #1212, opened 2021-11-02
Remove NetworkPodsCrashLooping alert for ovn-kubernetes

npinaeva:
use existing KubePodCrashLooping (https://github.com/openshift/cluster-monitoring-operator/blob/master/assets/control-plane/prometheus-rule.yaml#L15) alert from cluster-network-operator instead.
NetworkPodsCrashLooping not only duplicates existing KubePodCrashLooping alert, but actually alerts on restarts instead of crashlooping events


wking (approved):
I'm not an approver, but:

/lgtm
/retest

trozet (approved):
/lgtm

PR #1214, opened 2021-11-04
Set upgrade strategy on kube-proxy

bpickard22:
All daemonsets in OCP must have a max unavailable value set to 10% or
33%. Fixed kube-proxy daemonset to be in accordance with this rule


bpickard22:
cherry-pick release-4.9

bpickard22:
cherry-pick release-4.8

PR #1215, opened 2021-11-04
Add command to change running pods MTU

jcaamano:
New command to change MTU of running pods. It is used in pods that are to be deployed on every node as part of adding support to CNO to change the network MTU.


jcaamano:
Closing in favor of https://github.com/openshift/enhancements/blob/master/enhancements/network/allow-mtu-changes.md

PR #1220, opened 2021-11-08
Add CNI to DPU and enable Kube-Proxy on DPU

Billy99:
In the Infra-Cluster (Cluster with DPUs), OVN-Kubernetes on DPU is managing the networking for the Tenant-Cluster (Cluster with x86-hosts). Currently, only host network based pods are supported on the DPU. This PR adds an Infra-DPU-CNI to the DPU, which is just a script that return an error to any CNI cmdAdd() calls. Multus requires a "readiness-indicator-file", which is a conf file with the CNI Plugin specifics. The presence of this file indicates to Multus that the CNI is ready. This PR also setups the "readiness-indicator-file".

Since OVN-Kubernetes running on the DPU is serving the Tenant-Cluster (x86 host), enable KubeProxy to run on the DPU to  enable networking on the DPUs in the Infra-Cluster.

### DIFF
--- a/bindata/kube-proxy/kube-proxy.yaml
+++ b/bindata/kube-proxy/kube-proxy.yaml
@@ -163,6 +163,9 @@ spec:
       - operator: Exists
       nodeSelector:
         kubernetes.io/os: linux
+{{- if .KUBE_PROXY_NODE_SELECTOR }}
+        {{ .KUBE_PROXY_NODE_SELECTOR }}
+{{- end }}
       volumes:
       - name: host-slash
         hostPath:
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -50,6 +50,8 @@ spec:
                 {{ else }}
                 operator: DoesNotExist
                 {{ end }}
+              - key: network.operator.openshift.io/dpu
+                operator: DoesNotExist
       serviceAccountName: ovn-kubernetes-node
       hostNetwork: true
       hostPID: true
--- a/bindata/network/ovn-kubernetes/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-node.yaml
@@ -2,7 +2,11 @@
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
+  {{ if eq .OVN_NODE_MODE "dpu-host" }}
+  name: ovnkube-node-dpu-host
+  {{ else }}
   name: ovnkube-node
+  {{ end }}
   namespace: openshift-ovn-kubernetes
   annotations:
     kubernetes.io/description: |
@@ -11,7 +15,11 @@ metadata:
 spec:
   selector:
     matchLabels:
+      {{ if eq .OVN_NODE_MODE "dpu-host" }}
+      app: ovnkube-node-dpu-host
+      {{ else }}
       app: ovnkube-node
+      {{ end }}
   updateStrategy:
     type: RollingUpdate
     rollingUpdate:
@@ -21,12 +29,27 @@ spec:
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
       labels:
+        {{ if eq .OVN_NODE_MODE "dpu-host" }}
+        app: ovnkube-node-dpu-host
+        {{ else }}
         app: ovnkube-node
+        {{ end }}
         component: network
         type: infra
         openshift.io/component: network
         kubernetes.io/os: "linux"
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: network.operator.openshift.io/dpu-host
+                {{ if eq .OVN_NODE_MODE "dpu-host" }}
+                operator: Exists
+                {{ else }}
+                operator: DoesNotExist
+                {{ end }}
       serviceAccountName: ovn-kubernetes-node
       hostNetwork: true
       hostPID: true
@@ -38,6 +61,7 @@ spec:
       # /run/openvswitch -> tmpfs - ovsdb sockets
       # /env -> configmap env-overrides - debug overrides
       containers:
+      {{ if eq .OVN_NODE_MODE "full" }}
       # ovn-controller: programs the vswitch with flows from the sbdb
       - name: ovn-controller
         image: "{{.OvnImage}}"
@@ -145,6 +169,7 @@ spec:
           name: node-log
         - mountPath: /run/ovn/
           name: run-ovn
+      {{ end }}
       - name: kube-rbac-proxy
         image: {{.KubeRBACProxyImage}}
         command:
@@ -261,6 +286,11 @@ spec:
             gw_interface_flag="--exgw-interface=br-ex1"
           fi
 
+          node_mgmt_port_netdev_flags=
+          if [[ -n "${OVNKUBE_NODE_MGMT_PORT_NETDEV}" ]] ; then
+            node_mgmt_port_netdev_flags="--ovnkube-node-mgmt-port-netdev ${OVNKUBE_NODE_MGMT_PORT_NETDEV}"
+          fi
+
           exec /usr/bin/ovnkube --init-node "${K8S_NODE}" \
             --nb-address "{{.OVN_NB_DB_LIST}}" \
             --sb-address "{{.OVN_SB_DB_LIST}}" \
@@ -276,6 +306,10 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            {{- if eq .OVN_NODE_MODE "dpu-host" }}
+            --ovnkube-node-mode dpu-host \
+            ${node_mgmt_port_netdev_flags} \
+            {{- end }}
             --metrics-bind-address "127.0.0.1:29103" \
             --metrics-enable-pprof \
             {{- if .OVN_DISABLE_SNAT_MULTIPLE_GWS }}
@@ -467,6 +501,7 @@ spec:
       - name: run-ovn
         hostPath:
           path: /var/run/ovn
+      {{ if eq .OVN_NODE_MODE "full" }}
       # Used for placement of ACL audit logs 
       - name: node-log
         hostPath: 
@@ -474,6 +509,7 @@ spec:
       - name: log-socket
         hostPath: 
           path: /dev/log
+      {{ end }}
       # For CNI server
       - name: host-run-ovn-kubernetes
         hostPath:
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -44,6 +44,8 @@ const OVN_MASTER_DISCOVERY_BACKOFF = 120
 const OVN_LOCAL_GW_MODE = "local"
 const OVN_SHARED_GW_MODE = "shared"
 const OVN_LOG_PATTERN_CONSOLE = "%D{%Y-%m-%dT%H:%M:%S.###Z}|%05N|%c%T|%p|%m"
+const OVN_NODE_MODE_FULL = "full"
+const OVN_NODE_MODE_DPU_HOST = "dpu-host"
 
 var OVN_MASTER_DISCOVERY_TIMEOUT = 250
 
@@ -79,6 +81,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["CNIConfDir"] = pluginCNIConfDir(conf)
 	data.Data["CNIBinDir"] = CNIBinDir
 	data.Data["OVN_GATEWAY_MODE"] = bootstrapResult.OVN.OVNKubernetesConfig.GatewayMode
+	data.Data["OVN_NODE_MODE"] = OVN_NODE_MODE_FULL
 	data.Data["OVN_ENABLE_EGRESS_IP"] = bootstrapResult.OVN.OVNKubernetesConfig.EnableEgressIP
 	data.Data["OVN_DISABLE_SNAT_MULTIPLE_GWS"] = bootstrapResult.OVN.OVNKubernetesConfig.DisableSNATMultipleGWs
 	data.Data["OVN_NB_PORT"] = OVN_NB_PORT
@@ -193,6 +196,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 	objs = append(objs, manifests...)
 
+	nodeMode := bootstrapResult.OVN.OVNKubernetesConfig.NodeMode
+	if nodeMode != OVN_NODE_MODE_FULL {
+		data.Data["OVN_NODE_MODE"] = nodeMode
+		manifests, err = render.RenderTemplate(filepath.Join(manifestDir, "network/ovn-kubernetes/ovnkube-node.yaml"), &data)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to render manifests")
+		}
+		objs = append(objs, manifests...)
+	}
+
 	// obtain the current IP family mode.
 	ipFamilyMode := names.IPFamilySingleStack
 	if len(conf.ServiceNetwork) == 2 {
@@ -246,6 +259,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 func bootstrapOVNConfig(kubeClient client.Client) (*bootstrap.OVNConfigBoostrapResult, error) {
 	ovnConfigResult := &bootstrap.OVNConfigBoostrapResult{
 		GatewayMode:            OVN_SHARED_GW_MODE,
+		NodeMode:               OVN_NODE_MODE_FULL,
 		EnableEgressIP:         true,
 		DisableSNATMultipleGWs: false,
 	}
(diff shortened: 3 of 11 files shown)


Billy99:
This PR builds on top of #1138. Once that PR merges, the first 5 commits will be removed.

Billy99:
@zshi-redhat PTAL

zshi-redhat:
`pkg/network/kube_proxy_test.go:363:40: not enough arguments in call to renderStandaloneKubeProxy` , `bootstrapResult` is missing in the test.

I rebased the [138 PR](https://github.com/openshift/cluster-network-operator/pull/1138) to fix a typo in: https://github.com/openshift/cluster-network-operator/pull/1138/files#diff-3a72fb129233dbf79f270cfcc408ec08f67a3b21e8ecf4fba9ae8d8dd849a83eR287.  (`ovnConfigResult.DisableSNATMutlipleGWs` -> `ovnConfigResult.DisableSNATMultipleGWs`), you might want to rebase this PR as well to fix CI errors.

Billy99:
Design Considerations:

1) Deploying Infra-DPU-CNI as a DaemonSet and in its own Namespace:
* Original thought from DanW was to run as initContainer in Multus (multus-additional-cni-plugins DaemonSet has a handful of initContainers this would fit nicely with). Issue is that it requires a separate instance (multus-additional-cni-plugins and multus-additional-cni-plugins-dpu) as done in #1138 for OVN-Kubernetes. But didn't want to taint Multus yaml with the DPU.
* Ran Infra-DPU-CNI as a pod launched from a DaemonSet. Tried as an initContainer in the DaemonSet, but DaemonSets require a running pod (not just one initiContainer). Just running a script in bash of an existing image, using OVN-Kubernetes Image but could use any image, that sleeps forever after copying the files.
* Created a new Namespace, openshift-infra-dpu-cni. Could run under openshift-ovn-kubernetes, but wasn't sure if it was preferred in it's own namespace, so started there. Did not add rbac resources in the namespace. Running under the openshift-ovn-kubernetes could leverage the rbac from OVN-Kubernetes. Otherwise should rbac be added to the namespace? Not real familiar with rbac.

2) Enabling KubeProxy for DPU:
* Currently, enabling KubeProxy from within the OVN-Kubernetes logic by setting deploy flag. The downside to this approach is it won't be possible to launch KubeProxy Standalone on the Infra-Cluster (Master and non-DPU Workers). OVN-Kubernetes is being run on those nodes and it is not recommended to run KubeProxy in that deployment, so probably not a big deal. The alternative would be to created a separate instance (openshift-kube-proxy and openshift-kube-proxy-dpu) as done in #1138 for OVN-Kubernetes, and render the DPU specific yaml from the OVN-Kubernetes logic. This approach may require more changes in the KubeProxy logic so it can be rendered a second time, mainly making sure the defaults are setup proper on the second render call.

zshi-redhat:
> 
>     1. Enabling KubeProxy for DPU:
> 
> 
>     * Currently, enabling KubeProxy from within the OVN-Kubernetes logic by setting deploy flag. The downside to this approach is it won't be possible to launch KubeProxy Standalone on the Infra-Cluster (Master and non-DPU Workers). OVN-Kubernetes is being run on those nodes and it is not recommended to run KubeProxy in that deployment, so probably not a big deal. 

agree, I think both openshift-sdn and ovn-kubernetes are not supposed to be enabled with kube-proxy by default.

pkg/network/ovn_kubernetes.go
```diff
@@ -300,7 +307,7 @@ func bootstrapOVNConfig(kubeClient client.Client) (*bootstrap.OVNConfigBoostrapR
 		}
 	} else {
 		dpuModeOverride := cm.Data["mode"]
-		if dpuModeOverride != OVN_DPU_MODE_DPU_HOST {
+		if dpuModeOverride != OVN_DPU_MODE_DPU_HOST && dpuModeOverride != OVN_DPU_MODE_DPU {
 			klog.Warningf("dpu-mode-config does not match %q, is: %q. Using OVN configuration: %+v", OVN_DPU_MODE_DPU_HOST, dpuModeOverride, ovnConfigResult)
```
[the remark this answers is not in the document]
Billy99: Done.

Billy99:
> Two unit test failures:
> 
> `pkg/network/kube_proxy_test.go:363:40: not enough arguments in call to renderStandaloneKubeProxy` , `bootstrapResult` argument is missing in the test when calling renderStandaloneKubeProxy.
> 
> I rebased the [138 PR](https://github.com/openshift/cluster-network-operator/pull/1138) to fix a typo in: https://github.com/openshift/cluster-network-operator/pull/1138/files#diff-3a72fb129233dbf79f270cfcc408ec08f67a3b21e8ecf4fba9ae8d8dd849a83eR287. (`ovnConfigResult.DisableSNATMutlipleGWs` -> `ovnConfigResult.DisableSNATMultipleGWs`), you might want to rebase this PR as well to fix CI errors.

Done. Both fixed.

Billy99:
@danwinship PTAL at my "[Design Considerations](https://github.com/openshift/cluster-network-operator/pull/1220#issuecomment-964285423)" comment above and let me know if you would prefer a different approach (i.e. - use the OVN-Kubernetes namespace, use a different name for my CNI, etc). If it's fine, I'll remove the WIP, otherwise I'll rework to your suggestion and then remove it. Thanks.

bindata/network/infra-dpu-cni/000-ns.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-infra-dpu-cni
```
danwinship, line 4: I don't think we need a new namespace for this. Just put it in openshift-multus with the other misc CNI stuff

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
```diff
@@ -0,0 +1,163 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: infra-dpu-cni-script
```
danwinship, line 5: There's nothing DPU-specific about the script itself...

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
```diff
@@ -0,0 +1,163 @@
...
+  annotations:
+    kubernetes.io/description: |
+      This is a script that serves as the CNI on DPU. Only host network supported, so returns error to all calls.
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+data:
+  infra-dpu-cni.sh: |-
+    #!/bin/bash
+    set -e
+
+    exec 3>&1 # make stdout available as fd 3 for the result
```
danwinship, line 16: Why? Is this leftover from an earlier attempt to make fd 1 _not_ be stdout?

note that you redirect the ADD output to 3, but leave the VERSION output on 1...

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
```diff
@@ -0,0 +1,163 @@
...
+
+    log=/var/log/infra-dpu-cni.log
```
danwinship, line 18: not sure logging is really necessary...

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
```diff
@@ -0,0 +1,163 @@
...
+data:
+  08-infra-dpu-cni.conf: |-
+    {"cniVersion":"0.4.0","name":"infra-dpu-cni","type":"infra-dpu-cni"}
+
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: infra-dpu-cni-plugin
+  namespace: openshift-infra-dpu-cni
```
danwinship, line 84: Can this be squashed into the existing multus daemonset somehow? This is adding another set of pods that will be created, do something at startup, and then sit around consuming some resources and cluttering up `oc get` output forever after

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
```diff
@@ -0,0 +1,163 @@
...
+          name: infra-dpu-cni-conf-config-map
+        - mountPath: /cni-bin-dir
+          name: host-cni-bin
+        - mountPath: /etc/cni/net.d
+          name: host-cni-netd
+      volumes:
+        - name: infra-dpu-cni-script-config-map
+          configMap:
+            name: infra-dpu-cni-script
+            defaultMode: 0744
```
danwinship, line 152: 0755. or 0700. 0744 is just weird

bindata/kube-proxy/kube-proxy.yaml
```diff
@@ -161,6 +161,9 @@ spec:
       - operator: Exists
       nodeSelector:
         kubernetes.io/os: linux
+{{ if eq .OVN_DPU_MODE "dpu" }}
+        network.operator.openshift.io/dpu: ''
+{{ end }}
```
danwinship, line 166: Maybe

```
{{- if .KUBE_PROXY_NODE_SELECTOR }}
        {{ .KUBE_PROXY_NODE_SELECTOR }}
{{- end }}
```

and update the go code accordingly? Rather than putting DPU-specific stuff in this generic template...

danwinship (commented):
Can you squash the fixups?

bindata/network/infra-dpu-cni/infra-dpu-cni.yaml
Billy99: Nothing DPU specific other than it is intended to run only on the DPU. There is a [noop CNI](https://github.com/containernetworking/cni/tree/master/plugins/test/noop), so didn't want to confuse with that one. Is there a better name?

bindata/kube-proxy/kube-proxy.yaml
Billy99: Done.

pkg/network/ovn_kubernetes.go
```diff
@@ -197,13 +202,31 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+		objs = append(objs, manifests...)
+
+		// Run KubeProxy on DPU
+		if conf.DeployKubeProxy == nil {
+			v := true
+			conf.DeployKubeProxy = &v
+		} else {
+			*conf.DeployKubeProxy = true
+		}
+		fillKubeProxyDefaults(conf, nil)
```
danwinship, line 229: This relies on the fact that `renderStandaloneKubeProxy` happens to get called after `renderDefaultNetwork`... you should set the kube proxy config from `fillOVNKubernetesDefaults`.

pkg/network/ovn_kubernetes.go
```diff
@@ -46,6 +46,11 @@ const OVN_SHARED_GW_MODE = "shared"
 const OVN_LOG_PATTERN_CONSOLE = "%D{%Y-%m-%dT%H:%M:%S.###Z}|%05N|%c%T|%p|%m"
 const OVN_DPU_MODE_FULL = "full"
 const OVN_DPU_MODE_DPU_HOST = "dpu-host"
+const OVN_DPU_MODE_DPU = "dpu"
+const OVN_NODE_LABEL_DPU = "network.operator.openshift.io/dpu"
+const OVN_NODE_SELECTOR_DPU = "network.operator.openshift.io/dpu: ''"
+const OVN_NODE_LABEL_DPU_HOST = "network.operator.openshift.io/dpu-host"
+const OVN_NODE_SELECTOR_DPU_HOST = "network.operator.openshift.io/dpu-host: ''"
```
danwinship, line 53: you don't use the `OVN_NODE_LABEL_xxx` consts, and you don't use `OVN_NODE_SELECTOR_xxx` until the next commit so it probably should be added there, not in the error-cni commit.

pkg/network/ovn_kubernetes.go
Billy99: @danwinship When I originally started coding this up, I started coding it up in `fillKubeProxyDefaults()`. I moved it to it's current location because `FillDefaults()` doesn't have `bootstrapResult` computed yet. None of the defaults are derived from configuration data. I can rearrange the logic where `FillDefaults()` is called, the `bootstrapResult` is generated right after it. Just wasn't sure if that was better or not. If I am doing that, should it go into `fillKubeProxyDefaults()` or `fillDefaultNetworkDefaults()`?

`FillDefaults()` -> https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/render.go#L183
 Where `FillDefaults()` is called -> https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/operconfig/operconfig_controller.go#L173

danwinship: oh... yeah, we can't change `conf` based on `bootstrapResult`... The bootstrap data is only allowed to affect the rendering output. And actually, `fillKubeProxyDefaults` won't have filled in all of the defaults if `conf.DeployKubeProxy` had been `false` when it was called, so even the way the code is now, it's not getting entirely the right answer.

hm... and I don't really like the changes to `kube_proxy.go` either... it shouldn't be looking in someone else's bootstrap data... the ovn-kube rendering code ought to be able to tell the kube-proxy-rendering code what selector it wants to use. Somehow.

Really though, the answer here is just that we've hit the breaking point of our current abstractions and we need to redesign it a little... but not now.

OK, so eventually the dpu mode will actually be in the network config, right? So we'd have access to it at `fillOVNKubernetesDefaults` time... so I think the best answer that doesn't involve totally rewriting CNO rendering is to move the dpu-configmap-reading code from `bootstrapOVN` into a separate helper function, and then have `fillOVNKubernetesDefaults` and `bootstrapOVN` both call that helper function, and so then we can adjust the defaults based on the dpu mode.

Billy99: Per Discussion with @danwinship, leaving as is. Not ideal, but we are limited because the DPU support is DevPreview and we don't want to change the API until the feature is more stable. Enabling KubeProxy should be done in KubeProxy logic, but when setting up defaults, the code doesn't have access to the DPU Mode field. Currently it is stored in `bootstrapResult.OVN.OVNKubernetesConfig`, but longer term it will be stored in conf (operv1.NetworkSpec), which `FillDefaults` does have access.

Added comments with the tag `DPU_DEV_PREVIEW` in:
* rendor.go, where it is assumed that `renderStandaloneKubeProxy()` is called after `renderDefaultNetwork()`
* kube_proxy.go where it is accessing `bootstrapResult.OVN.OVNKubernetesConfig.DPUMode`
* ovn_kubernetes.go where it enables KubeProxy, which should eventually be moved to KubeProxy logic.

pkg/network/ovn_kubernetes.go
Billy99: Moved `OVN_NODE_SELECTOR_DPU` to next commit and removed the rest. My plan was to replace all labels in the yaml files with these constants, but after doing so, thought it was too much abstraction and backed it out. Just using the one constant now.
Done.

Billy99: Moved `OVN_NODE_SELECTOR_DPU` to the next commit and removed the rest of the constants.
Original plan was to replace all the DPU base labels with these constants, but once implemented I decided it was too much abstract and back it out.

Billy99:
@danwinship PTAL

bindata/network/ovn-kubernetes/ovnkube-node.yaml
```diff
@@ -261,6 +286,11 @@ spec:
             gw_interface_flag="--exgw-interface=br-ex1"
           fi
 
+          node_mgmt_port_netdev_flags=
+          if [[ -n "${OVNKUBE_NODE_MGMT_PORT_NETDEV}" ]] ; then
```
danwinship, line 292: nothing seems to ever set this? Is this part not implemented yet?

pkg/network/ovn_kubernetes.go
```diff
@@ -246,6 +259,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 func bootstrapOVNConfig(kubeClient client.Client) (*bootstrap.OVNConfigBoostrapResult, error) {
 	ovnConfigResult := &bootstrap.OVNConfigBoostrapResult{
 		GatewayMode:            OVN_SHARED_GW_MODE,
+		DPUMode:                OVN_DPU_MODE_FULL,
```
danwinship, line 287: This makes it sound like the default is "full DPU usage" rather than "full ovnkube-node functionality". Can you change "dpu mode" to "node mode" everywhere? (template variable names, golang variable names, struct fields, const names)

danwinship (commented):
mostly good

pkg/network/ovn_kubernetes.go
Billy99: Done

bindata/network/ovn-kubernetes/ovnkube-node.yaml
Billy99: We set it via [configmap-override-cno-env.yaml](https://github.com/zshi-redhat/bluefield-demo/blob/a7cf4aff5c976fea8d9ce6ed1da3468805b9f3ea/manifests/configmap-override-cno-env.yaml), which is documented in the [Order of Installation Operation](https://docs.google.com/document/d/1m3xKFERxI_teUbln3dMVPY_xJ5hPLT3xM6JG_hFNZmI/edit#heading=h.7axwjok4nzxb) section, Step 8, of the "Pod Networking in a DPU Cluster" document. This is probably something that can automated later. In early development, the VF name changed depending on kernel version loaded.

hack/node-mode.yaml
```diff
@@ -0,0 +1,9 @@
+# Example ConfigMap to enable dpu-host mode with OVNKubernetes
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: node-mode-config
```
danwinship, line 5: oh, I had meant to leave the configmap as "dpu-mode-config" still... or _something_ with "dpu" in the name anyway...

danwinship:
/approve
your call if you want to re-rename the configmap

hack/node-mode.yaml
Billy99: OK, changed back.

pkg/network/ovn_kubernetes.go
```diff
@@ -256,24 +295,44 @@ func bootstrapOVNConfig(kubeClient client.Client) (*bootstrap.OVNConfigBoostrapR
...
+	err = kubeClient.Get(context.TODO(), dmc, cm)
+
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Did not find dpu-mode-config")
+		} else {
+			return nil, fmt.Errorf("Could not determine Node Mode: %w", err)
+		}
+	} else {
+		nodeModeOverride := cm.Data["mode"]
```
tssurya, line 327: I think this is supposed to be `nodeModeOverride := dmc.Data["mode"]` ? Else this value will be local or shared gateway mode :) and below condition will always be hit.
Note that we are removing the hidden gateway-config-mode this release cycle, so hit this code conflict as I was trying to fix my PR - I can fix this in my PR, but I am surprised none of the testing caught this?
We should expose this as a proper API once this feature is stable since we are going to be overloading with config-maps for every config otherwise if we follow this pattern.

tssurya (commented):
cc @Billy99 @zshi-redhat

pkg/network/ovn_kubernetes.go
tssurya: oops nvm ignore my comment, my eyes read the object name differently...

zshi-redhat: @tssurya Thanks for the comments and review!
Yes, we followed the same pattern as modeOverride (local/shared gateway mode) for DPU mode.
We will consider the API change for DPU node mode once the bluefield-2 design get mature and stable.

PR #1226, opened 2021-11-12
Bug 2021191: Project admins should be able to list net-attach-defs in their namespaces

dougbtv:
Non-privileged / normal project admin users should be able to list available NetworkAttachmentDefinitions for those created in their project.


bindata/network/multus/002-rbac.yaml
```diff
@@ -132,3 +132,14 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: net-attach-def-project
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups: ["k8s.cni.cncf.io"]
+  resources: ["network-attachment-definitions"]
+  verbs: ["watch", "list"]
```
danwinship, line 145: add `"get"` too? It's not technically _required_, but it doesn't make a lot of sense to let people do `oc get network-attachment-definitions` but _not_ do `oc get network-attachment-definition foo`

yaacov: @danwinship hi, customer request was that a project admin ( a user that can do anything in a namespace ) can look for a `network-attachment-definition` name  to use for creating a network interface for a virtual machine, so we only need to list the names, I have no opinion about  allowing to see the internals of a `network-attachment-definition` but it's not required for this issue, currently they need to manually type the NAD name they need to get from a cluster admin (by mail?).

danwinship: >  I have no opinion about allowing to see the internals of a `network-attachment-definition`

That's implied by "list" though; when you do a list, you get the full objects back, not just their names. So giving someone "list" but not "get" means they can see the internals of any of the objects, but they have to request all of them in order to do it...

dougbtv: Thanks for the eye that list does indeed add pull the entire object. I added get, as you're right -- it's goofy if they can already seem them but not get individual.

...This one is moving slowly as we'd ideally like to get some more opinions on if this makes sense to entirely expose these objects in this way, it kind of changes the paradigm for who has access (e.g. previously, it's just cluster admin).

yaacov: > ...This one is moving slowly as we'd ideally like to get some more opinions on if this makes sense to entirely expose these objects in this way, it kind of changes the paradigm for who has access (e.g. previously, it's just cluster admin).

@phoracek hi, can you please look into this ?

danwinship:
/approve
someone else can lgtm after you've decided whether or not the change is ok...

PR #1231, opened 2021-11-16
Bug 2034517: watch and apply changes of the ovs-flows-config configmap

mariomac:
As described in the [Related RFE](https://github.com/openshift/enhancements/blob/master/enhancements/network/ovs-flow-export-configuration.md), the CNO needs to listen for any change in the `ovs-flows-config` ConfigMap that is created after a `FlowCollector` object, so the CNO can properly configure and tune the OVN-Kubernetes' OpenVSwitch instance to export IPFIX flows.

This PR adds the following functionalities:

1. Adds a watcher for the `ovs-flows-config` configmap and retriggers the `operconfig_controller` reconciliation process on any change in that map.
2. On reconciliation, bootstraps the `ovs-flows-config` configmap information and renders it into the `ovnkube-node.yaml` to properly reconfigure`ovn-kubernetes`.
3. Defines the RBAC assets according to the [CNO<->NOO RBAC configuration document](https://docs.google.com/document/d/1AeUgETCtsVsYPuvpu0HuDfeuhTaWGRd3Qpkf38bu81U)

The following PR for the `ovn-kubernetes` repository must be merged before this PR: https://github.com/ovn-org/ovn-kubernetes/pull/2649

Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2034517

### DIFF
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -2,7 +2,6 @@ package bootstrap
 
 import (
 	"github.com/gophercloud/utils/openstack/clientconfig"
-
 	configv1 "github.com/openshift/api/config/v1"
 	appsv1 "k8s.io/api/apps/v1"
 )
@@ -42,6 +41,7 @@ type OVNBootstrapResult struct {
 	ExistingNodeDaemonset   *appsv1.DaemonSet
 	OVNKubernetesConfig     *OVNConfigBoostrapResult
 	PrePullerDaemonset      *appsv1.DaemonSet
+	FlowsConfig             *FlowsConfig
 }
 
 type BootstrapResult struct {
@@ -55,4 +55,18 @@ type BootstrapResult struct {
 type CloudBootstrapResult struct {
 	PlatformType   configv1.PlatformType
 	PlatformRegion string
+}
+
+type FlowsConfig struct {
+	// Target IP:port of the flow collector
+	Target string
+
+	// CacheActiveTimeout is the max period, in seconds, during which the reporter will aggregate flows before sending
+	CacheActiveTimeout *uint
+
+	// CacheMaxFlows is the max number of flows in an aggregate; when reached, the reporter sends the flows
+	CacheMaxFlows *uint
+
+	// Sampling is the sampling rate on the reporter. 100 means one flow on 100 is sent. 0 means disabled.
+	Sampling *uint
 }
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/network"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -28,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -69,6 +71,14 @@ func add(mgr manager.Manager, r *ReconcileOperConfig) error {
 	// Watch for changes to primary resource Network
 	err = c.Watch(&source.Kind{Type: &operv1.Network{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
+		return err
+	}
+
+	// watch for changes in the ovs-flows-config map
+	if err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}},
+		handler.EnqueueRequestsFromMapFunc(reconcileOvsFlowsConfig),
+		predicate.ResourceVersionChangedPredicate{},
+	); err != nil {
 		return err
 	}
 
@@ -348,4 +358,19 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	// All was successful. Request that this be re-triggered after ResyncPeriod,
 	// so we can reconcile state again.
 	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
+}
+
+// reconcileOvsFlowsConfig filters non-ovs-flows-config events and forwards a request to the
+// openshift-network-operator/cluster operator
+func reconcileOvsFlowsConfig(object client.Object) []reconcile.Request {
+	n := object.GetName()
+	ns := object.GetNamespace()
+	if n != network.OVSFlowsConfigMapName || ns != network.OVSFlowsConfigNamespace {
+		return nil
+	}
+	log.Println(network.OVSFlowsConfigMapName + ": enqueuing operator reconcile request from configmap")
+	return []reconcile.Request{{NamespacedName: types.NamespacedName{
+		Name:      names.OPERATOR_CONFIG,
+		Namespace: names.APPLIED_NAMESPACE,
+	}}}
 }
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -51,6 +51,11 @@ const OVN_NODE_SELECTOR_DPU = "network.operator.openshift.io/dpu: ''"
 
 var OVN_MASTER_DISCOVERY_TIMEOUT = 250
 
+const (
+	OVSFlowsConfigMapName   = "ovs-flows-config"
+	OVSFlowsConfigNamespace = names.APPLIED_NAMESPACE
+)
+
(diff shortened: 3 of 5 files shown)


pkg/controller/flowsconfig/flowsconfig_controller.go
```diff
@@ -0,0 +1,135 @@
+package flowsconfig
```
astoycos, line 1: I don't think this is 100% correct implementation wise... If you were trying to make a new controller, it would need to follow the same schema all the other controllers follow.. 

For example, the `egress_router_controller` defines an `Add` function which is added to the list of controller create functions [here](https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/add_networkconfig.go#L16)

Really all we're doing here is providing another input to the `operconfig` controller

astoycos: I would probably move these files/ helper functions to the `operconfig/` directory

pkg/network/ovn_kubernetes.go
```diff
@@ -568,6 +600,7 @@ func bootstrapOVN(conf *operv1.Network, kubeClient client.Client) (*bootstrap.Bo
 			OVNKubernetesConfig:     ovnConfigResult,
 			PrePullerDaemonset:      prePullerDS,
 			Platform:                platformType,
+			FlowsConfig:             flowsconfig.Bootstrap(kubeClient),
```
astoycos, line 603: Isn't this always going to be nil at startup? (I'm assuming the NOO starts up after the CNO)

pkg/controller/flowsconfig/flowsconfig_controller.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil
+	}
+	log.Println(ovsFlowsConfigMapName + ": enqueuing operator reconcile request from configmap")
+	return []reconcile.Request{{NamespacedName: types.NamespacedName{
+		Name:      names.OPERATOR_CONFIG,
+		Namespace: names.APPLIED_NAMESPACE,
+	}}}
+}
+
+// Bootstrap looks for the openshift-network-operator/ovs-flows-config configmap, and
```
astoycos, line 71: Shouldn't we default to _something_ for these values if the configmap does not exist?  ((Especially during bootstrap since the configmap most likely won't be there yet if the NOO hasn't started)

bindata/network/ovn-kubernetes/008-ovs-flows-config.yaml
```diff
@@ -0,0 +1,54 @@
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: ovs-flows-config-publisher
```
astoycos, line 4: So let me preface this by saying i'm not an RBAC expert...... However, I've read though https://docs.google.com/document/d/1AeUgETCtsVsYPuvpu0HuDfeuhTaWGRd3Qpkf38bu81U/edit#heading=h.1vsfycjyceaq and tend to think that the CNO should only be defining RBAC roles that it intends to use (for resources it doesn't own), or delegate out(for resources it does own) 

In this case we're defining an a role in the CNO for an resource it doesn't create, nor will it ever use the role.  

I _think_  we should instead have the NOO make/ bind the `ovs-flows-config-publisher` role

astoycos (commented):
Some questions/ recs, overall it looks good!

pkg/controller/flowsconfig/flowsconfig_controller.go
martinkennelly: "Bootstrap" can be executed after any reconcile event and not just on startup.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -72,6 +74,12 @@ func add(mgr manager.Manager, r *ReconcileOperConfig) error {
 		return err
 	}
 
+	// watch for changes in the ovs-flows-config map
+	err = flowsconfig.WatchForConfigMap(c)
```
martinkennelly, line 78: super small nit: if err = flowsconfig... {
}
golang idiomatic

martinkennelly (commented):
Looks good overall, just a few nits and recommend move the helper code into operconfig dir.

pkg/network/ovn_kubernetes.go
mariomac: This is expected to be nil most of the times, as it will be only not nil when the flow tracing is enabled. If it  is nil, the `renderOVNFlowsConfig` function just ignores it and then the `ovnkube-node` instances will be rendered with the usual configuration.

pkg/controller/flowsconfig/flowsconfig_controller.go
mariomac: As explained in the comment before, if the configmap does not exist (which will be the usual case), then it will just proceed as now: the `ovnkube-nodes` will be configured normally without any flow tracing capability.

pkg/controller/flowsconfig/flowsconfig_controller.go
mariomac: In the previous commit I needed to invoke it from the operconfig controller, instead of creating a new one, because the OVN-K configuration is rendered together with the rest of the network configuration (which is invoked from the operconfig).

My initial decision of having everything in a package apart was to put together the bootstrapping/rendering and the reconciliation of the ConfigMap option, but it's true that could give the false impression of being a standalone controller. In the following commit I'll do:

- move the flowsconfig bootstrap function to the `ovn_kubernetes.go` file, to avoid import cycles.
- move watch and reconcile to the `operconfig` package.

pkg/controller/flowsconfig/flowsconfig_controller.go
```diff
@@ -0,0 +1,135 @@
...
+	ovsFlowsConfigMapName   = "ovs-flows-config"
+	ovsFlowsConfigNamespace = names.APPLIED_NAMESPACE
+)
+
+type FlowsConfig struct {
+	// Target IP:port of the flow collector
+	Target string
+
+	// CacheActiveTimeout is the max period during which the reporter will aggregate flows before sending
+	CacheActiveTimeout *uint
```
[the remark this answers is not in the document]
mariomac: The pointers are used to detect if a feature is enabled or not, as they are optional configuration values. Another options could be:
- using the zero value (if `0 => disabled`) but it might be not so clear for some properties where zero is a possible value.
- adding some kind of extra boolean property, like a boolean saying if it is enabled or not, but it could maybe dirty the code.

bindata/network/ovn-kubernetes/008-ovs-flows-config.yaml
mariomac: Ok, sounds reasonable. I'll remove this

pkg/controller/operconfig/operconfig_controller.go
mariomac: Done. (I did it that way to be consistent with the style of the rest of the function)

mariomac:
Thank you for your review @astoycos and @martinkennelly . I addressed all your points. You can find some explanations after your comments.

pkg/controller/flowsconfig/flowsconfig_controller.go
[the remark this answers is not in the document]
martinkennelly: Its fine. If anyone can think of a cleaner idea, I am all ears but its fine as it is.

martinkennelly:
/lgtm 
Thanks for the changes Mario.

mariomac:
/hold
This another PR would need to be merged first: https://github.com/ovn-org/ovn-kubernetes/pull/2649

astoycos:
Hey @mariomac sorry for the NIT but can you please squash your commits :)

mariomac:
/unhold
The ovn-kubernetes PR that this PR depends on has been already merged

astoycos:
`e2e-aws-sdn-multi` seems to be perma-failing 

/lgtm

pkg/bootstrap/types.go
```diff
@@ -56,3 +56,17 @@ type CloudBootstrapResult struct {
 	PlatformType   configv1.PlatformType
 	PlatformRegion string
 }
+
+type FlowsConfig struct {
+	// Target IP:port of the flow collector
+	Target string
+
+	// CacheActiveTimeout is the max period, in seconds, during which the reporter will aggregate flows before sending
+	CacheActiveTimeout *uint
```
abhat, line 65: why not just have them as uint or int? Size of uint or int is the same as size of *uint on 64-bit systems. This looks a very C-like syntax.

mariomac: The main reason was that all the properties here are optional and 0 are meaningful values which might not coincide with the defaults.

abhat: Ok makes sense.

pkg/bootstrap/types.go
```diff
@@ -56,3 +56,17 @@ type CloudBootstrapResult struct {
...
+
+	// CacheMaxFlows is the max number of flows in an aggregate; when reached, the reporter sends the flows
```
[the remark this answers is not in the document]
mariomac: That's whayt I understand from the openvswitch documentation: https://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.html

> cache_max_flows: optional integer, in range 0 to 4,294,967,295
> The maximum number of IPFIX flow records that can be cached at a
> time.  If  not  specified,  defaults to 0. If 0, caching is dis‐
> abled.

dcbw:
Permafailing; we'll look into it. This change does not affect SDN.

PR #1232, opened 2021-11-16
Add bare pods to status manager

jcaamano:
As part of the ongoing effort to allow MTU changes which would include
jobs deployed as bare pods, add support in status manager to track their
status assuming that they run to completion

ref: https://github.com/openshift/cluster-network-operator/pull/1215


jcaamano:
Closing in favor of https://github.com/openshift/enhancements/blob/master/enhancements/network/allow-mtu-changes.md

PR #1237, opened 2021-11-23
OVN-K alerts: Fix incorrect metric name reference

martinkennelly:
Fix incorrect metric name reference. 
As you can see upstream [1], we register with the default register, and this means the metric starts with "ovnkube_master".

[1] https://github.com/ovn-org/ovn-kubernetes/blob/b1b54eddff1490d6c93a5aceba6bd2029b6b31cf/go-controller/pkg/metrics/master.go#L201


martinkennelly:
Alex noted this maybe not needed as we depend on Leader Election code.

npinaeva:
lgtm

PR #1238, opened 2021-11-24
The ip-reconciler should not restart on failures.

dougbtv:
When failures occur, a number of failed cronjobs pile up. Since another one will start in 5 minutes, we should just wait until the next instance in order to start another pod.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -483,12 +483,15 @@ metadata:
     tier: node
     app: whereabouts
 spec:
-  # reconcile loop every 5 minutes, starting at the top of the hour
-  schedule: "*/5 * * * *"
+  # reconcile loop every 15 minutes, starting at the top of the hour
+  schedule: "*/15 * * * *"
+  concurrencyPolicy: Replace
+  successfulJobsHistoryLimit: 0
   jobTemplate:
     spec:
       template:
         spec:
+          backoffLimit: 0
           priorityClassName: "system-cluster-critical"
           serviceAccountName: multus
           containers:
@@ -508,4 +511,4 @@ spec:
             - name: cni-net-dir
               hostPath:
                 path: {{ .SystemCNIConfDir }}
-          restartPolicy: OnFailure
+          restartPolicy: Never


maiqueb (changes requested):
I do not think this is enough, since the job gets restarted plenty of times (6 by default) before it is considered failed.

I've pushed https://github.com/k8snetworkplumbingwg/whereabouts/pull/167 upstream with a more conservative configuration, that will ensure that you run exactly **one** time each period.

It should also remove the pods hanging around on failure / success.

maiqueb (approved):

bindata/network/multus/multus.yaml
```diff
@@ -483,14 +483,18 @@ metadata:
...
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 0
   jobTemplate:
     spec:
       template:
         spec:
           priorityClassName: "system-cluster-critical"
           serviceAccountName: multus
+          concurrencyPolicy: Forbid
+          successfulJobsHistoryLimit: 0
```
Billy99, line 497: I don't think these are needed. After applying, when the yaml is inspected, these fields are dropped:
```
# oc get cronjob -n openshift-multus ip-reconciler -o yaml
apiVersion: batch/v1
kind: CronJob
metadata:
:
spec:
  concurrencyPolicy: Replace
  failedJobsHistoryLimit: 1
  jobTemplate:
    metadata:
      creationTimestamp: null
    spec:
      template:
        metadata:
          creationTimestamp: null
        spec:
:
          dnsPolicy: ClusterFirst
          priorityClassName: system-cluster-critical
          restartPolicy: Never
          schedulerName: default-scheduler
          securityContext: {}
          serviceAccount: multus
          serviceAccountName: multus
          terminationGracePeriodSeconds: 30
          volumes:
:
```

bindata/network/multus/multus.yaml
```diff
@@ -483,14 +483,18 @@ metadata:
     tier: node
     app: whereabouts
 spec:
-  # reconcile loop every 5 minutes, starting at the top of the hour
-  schedule: "*/5 * * * *"
+  # reconcile loop every 15 minutes, starting at the top of the hour
+  schedule: "*/15 * * * *"
+  concurrencyPolicy: Forbid
```
Billy99, line 488: Would `Replace` be better here? On one of my test setups, the `ip-reconciler-xxx` pod isn't starting (expected). One pods runs forever trying to start. With `Replace`, it is at least restarted every time period. Not sure which is better.

Billy99 (commented):
Just some observations from testing.

bindata/network/multus/multus.yaml
dougbtv: I think I like replace in that case, as well.

dougbtv:
Additionally from Billy's findings:

> On my cluster, ip-reconciler was passing. So if you watched all the Multus pods, ip-reconciler would periodically pop in, run, then disappear.

> On [another] cluster, where ip-reconciler can't start because it is being scheduled on a [machine that can't run this process], there is only one instance in the ContainerCreating state. I still have it set to run every 5 mins (instead of 15 mins like in the patch):

```
# oc get pods -n openshift-multus
NAME                                  READY   STATUS              RESTARTS   AGE
ip-reconciler-27304970-mwrlp          0/1     ContainerCreating   0          16m
```

> So this is better than I found it, with 2000 instances building up over the Thanksgiving break.

bindata/network/multus/multus.yaml
dougbtv: Ooops!! Yeah that shouldn't be there at all, that's a copy-paste error from me, in the wrong place. Thanks for the eye.

PR #1240, opened 2021-12-02
Allow any master to upgrade db schema

trozet:
This will allow upgrades for libovsdb to work. Libovsdb has an issue
where if it tries to connect to a db that does not match the client's
schema it will fail. So for example, attempting to connect from a client
with a higher schema revision to an older database would fail, because
the tables that the client expects would be missing. In the upgrade
scenario, we dont know which master will be upgraded first and in that
case we need to ensure that the db schema will be upgraded so libovsdb
will be compatible.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -244,6 +244,22 @@ spec:
               - -c
               - |
                 set -x
+                # Upgrade the db if required.
+                DB_SCHEMA="/usr/share/ovn/ovn-nb.ovsschema"
+                DB_SERVER="unix:/var/run/ovn/ovnnb_db.sock"
+                schema_name=$(ovsdb-tool schema-name $DB_SCHEMA)
+                db_version=$(ovsdb-client -t 10 get-schema-version "$DB_SERVER" "$schema_name")
+                target_version=$(ovsdb-tool schema-version "$DB_SCHEMA")
+
+                if ovsdb-tool compare-versions "$db_version" == "$target_version"; then
+                  :
+                elif ovsdb-tool compare-versions "$db_version" ">" "$target_version"; then
+                    echo "Database $schema_name has newer schema version ($db_version) than our local schema ($target_version), possibly an upgrade is partially complete?"
+                else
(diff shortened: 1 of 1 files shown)


trozet:
@dcbw @numansiddique @dceara @tssurya

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -244,6 +244,22 @@ spec:
               - -c
               - |
                 set -x
+                # Upgrade the db if required.
```
tssurya, line 247: I can't find anything specifically wrong in moving this out of the loop and providing the freedom of doing an upgrade on any control-plane node. Having said that I'd be curious to see why this code was restricted to the initiator in the first place, maybe @numansiddique (https://github.com/openshift/cluster-network-operator/pull/755/files#diff-58be2f354a42fe685cf7eadff055a78e1f9be499adf74f786cde1a26d509cc7bR155) can say why it used to be done only for `MASTERIP` which later became `INITIATOR`?

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -258,22 +274,6 @@ spec:
                   fi
                   sleep 2
                   done
-
```
tssurya, line 261: Might be a stupid question, but is it necessary to do `set-connection` before executing the ovs-db client commands for upgrade? Seems like the error would be caught after retried and cleanly exited (`exit 1` on L257) in case it failed and we wouldn't proceed with upgrading the db? Or is it only for the nbctl commands that follow this block?

tssurya (commented):
Would be interesting to also know about the question @trozet asked https://coreos.slack.com/archives/C01G7T6SYSD/p1638459666280600?thread_ts=1638457589.269400&cid=C01G7T6SYSD.

trozet:
Closing this and moving it into ovnkube upstream:
https://github.com/ovn-org/ovn-kubernetes/pull/2694

bindata/network/ovn-kubernetes/ovnkube-master.yaml
trozet: Spoke with Numan and he said there was no real reason only the initiator was originally doing it

PR #1241, opened 2021-12-03
Add MTU migration support for OVNKubernetes and OpenshiftSDN

Kyrtapz:
Add MTU migration support for OVNKubernetes and OpenshiftSDN

Enhancement: https://github.com/openshift/enhancements/blob/master/enhancements/network/allow-mtu-changes.md

Depends on https://github.com/openshift/sdn/pull/381, should be merged after that

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -6,6 +6,15 @@ metadata:
 data:
   kube-proxy-config.yaml: |-
 {{.KubeProxyConfig | indent 4}}
+{{- if or .MTU .RoutableMTU }}
+  mtu.yaml: |-
+  {{- if .MTU }}
+    mtu: {{.MTU}}
+  {{- end}}
+  {{- if .RoutableMTU}}
+    routable-mtu: {{.RoutableMTU}}
+  {{- end}}
+{{- end}}
 ---
 kind: DaemonSet
 apiVersion: apps/v1
@@ -91,10 +100,16 @@ spec:
           rm -f /etc/cni/net.d/80-openshift-network.conf
           cp -f /opt/cni/bin/openshift-sdn /host-cni-bin/
 
+          mtu_override_flag=
+          if [[ -f /config/mtu.yaml ]]; then
+            mtu_override_flag="--mtu-override /config/mtu.yaml"
+          fi
+
           # Launch the network process
           exec /usr/bin/openshift-sdn-node \
             --node-name ${K8S_NODE_NAME} --node-ip ${K8S_NODE_IP} \
             --proxy-config /config/kube-proxy-config.yaml \
+            ${mtu_override_flag} \
             --v ${OPENSHIFT_SDN_LOG_LEVEL:-2}
         securityContext:
           privileged: true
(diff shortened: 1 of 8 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -75,6 +75,21 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	if conf.Migration != nil && conf.Migration.MTU != nil {
+		if *conf.Migration.MTU.Network.From > *conf.Migration.MTU.Network.To {
+			data.Data["MTU"] = conf.Migration.MTU.Network.From
+			data.Data["RoutableMTU"] = conf.Migration.MTU.Network.To
+		} else {
+			data.Data["MTU"] = conf.Migration.MTU.Network.To
+			data.Data["RoutableMTU"] = conf.Migration.MTU.Network.From
+		}
+
+		// c.MTU is used to set the applied network configuration MTU
```
jcaamano, line 89: Maybe we could add a comment block here explaining the procedure:
```
MTU migration procedure
1. User sets the MTU it wants to migrate to
2. CNO sets the MTU as applied
3. User can then set the MTU as configured
```

pkg/network/ovn_kubernetes.go
```diff
@@ -334,9 +359,26 @@ func isOVNKubernetesChangeSafe(prev, next *operv1.NetworkSpec) []error {
 	nn := next.DefaultNetwork.OVNKubernetesConfig
 	errs := []error{}
 
-	if !reflect.DeepEqual(pn.MTU, nn.MTU) {
-		errs = append(errs, errors.Errorf("cannot change ovn-kubernetes MTU"))
+	if next.Migration != nil && next.Migration.MTU != nil {
+		mtuNet := next.Migration.MTU.Network
+		mtuMach := next.Migration.MTU.Machine
+
+		// MTUMigrationValues.To is required
```
jcaamano, line 366: Similar here, comment block
```
For MTU values provided for migration, verify that:
* The current and target MTUs for the CNI are provided
* The machine target MTU is provided
* The current MTU actually matches the MTU known as current
* The machine target MTU has a valid overhead with the CNI target MTU 

```

Kyrtapz:
`/ci/prow/verify` fails due to: https://github.com/openshift/build-machinery-go/issues/56

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -91,10 +97,21 @@ spec:
           rm -f /etc/cni/net.d/80-openshift-network.conf
           cp -f /opt/cni/bin/openshift-sdn /host-cni-bin/
 
+          mtu_flag=
+          if [[ -f /config/mtu ]]; then
+            mtu_flag="--mtu $(cat /config/mtu)"
```
danwinship, line 102: There's no reason to put it into the config file and then read it out of it here; you can just do
```
mtu_flag={{if .MTU}}--mtu {{.MTU}}{{end}}
routable_mtu_flag={{if .RoutableMTU}}--routable-mtu {{.RoutableMTU}}{{end}}
```

alternatively, put the MTU and RoutableMTU into an `mtu.yaml` file in the config, and then pass `--mtu-override /config/mtu.yaml` to sdn

Kyrtapz: I did it this way to avoid the rollout of sdn daemonset.
I will use the `mtu.yaml` approach

danwinship:
/approve
I only skimmed but it looks good.
It would be nice to share more code between ovn and sdn. eg, the validation stuff, and deciding which value is MTU and which is RoutableMTU

danwinship:
> Depends on [openshift/sdn#381](https://github.com/openshift/sdn/pull/381), should be merged after that

It doesn't actually matter; it only uses the new command-line flag when you're doing migration, so merging this first won't break anything in the normal case. You just won't be able to test out migration yet.

trozet:
windows job was failing FAIL: TestWMCO/network/Pod_DNS_Resolution (21.94s)

but passes all other hybrid tests. I dont think the DNS lookup issue is related to this PR

jcaamano:
@trozet the override of the sdn-multi job was lost 🤦‍♂️

abhat:
@jcaamano when the PR gets into the merge queue batch, the overrides don't matter. There is some esoteric logic that the overrides apply only to getting into the merge batch, but not after it's in the batch.

PR #1251, opened 2021-12-13
Bug 2035459: modify cluster-network-features for OpenshiftSDN

mariomac:
According to the recent comment of Dan Winship in a recently merged
PR: https://github.com/openshift/cluster-network-operator/pull/1204#issuecomment-990201372

> openshift-sdn also implements both egress rules and ipBlock exceptions.

This PR updates the reported capabilities for OpenshiftSDN so the Console
will accordingly show the proper sections in the Network Policy forms.

### DIFF
--- a/bindata/network/openshift-sdn/cni-features.yaml
+++ b/bindata/network/openshift-sdn/cni-features.yaml
@@ -8,5 +8,5 @@ metadata:
       Exposes available network features as required by the Console in order to show or hide some form fields.
       If the map or a given property is undefined, the Console won't throw error and will take a default action (show, hide, show with a warning message...).
 data:
-  policy_egress: "false"
-  policy_peer_ipblock_exceptions: "false"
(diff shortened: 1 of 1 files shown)


jotak:
@martinkennelly looks like the failing tests are unrelated to this change, right? Are they frequently failing randomly? Is it possible to bypass them?

martinkennelly:
> @martinkennelly looks like the failing tests are unrelated to this change, right? Are they frequently failing randomly? Is it possible to bypass them?

Theres an issue with dual stack job. Ask Tim/aniket for an override if that job is blocking you.
I'm away from office.

mariomac:
@trozet @abhat this is the bug that the tests are preventing us to complete (we double-checked that the failing tests are unrelated to our small change): https://bugzilla.redhat.com/show_bug.cgi?id=2035459

¿Would it be possible to override tests? Thanks

PR #1266, opened 2021-12-22
Bug 2034355: Whereabouts IP Reconciliaton [backport 4.7]

dougbtv:
This implements the Whereabouts IP reconciliation cron job

NOTE: This pull request should be merged AFTER https://github.com/openshift/whereabouts-cni/pull/79


dougbtv:
I pushed a new commit to address some issues with apiversion for the ippools CRD as well as the cronjob apiversion. 

Additionally, it's worth noting that this won't pass CI until this PR passes: https://github.com/openshift/whereabouts-cni/pull/79

PR #1268, opened 2021-12-22
Bug 2035093: Cloud network config controller: Fix for Hypershift

alvaroaleman:
Two changes are needed:
* Hypershift doesn't have master nodes, so the node selector must be absent
* Hypershift uses STS to auth with AWS, so a SA token with an `openshift` audience needs to be mounted. This also requires https://github.com/openshift/cloud-network-config-controller/pull/13 to work

/assign @alexanderConstantinescu

### DIFF
--- a/bindata/cloud-network-config-controller/controller.yaml
+++ b/bindata/cloud-network-config-controller/controller.yaml
@@ -55,8 +55,10 @@ spec:
           mountPath: "/etc/secret/cloudprovider"
           readOnly: true
         terminationMessagePolicy: FallbackToLogsOnError
+{{- if not .ExternalControlPlane }}
       nodeSelector:
         node-role.kubernetes.io/master: ""
+{{- end }}
       tolerations:
       - key: "node-role.kubernetes.io/master"
         operator: "Exists"
--- a/bindata/cloud-network-config-controller/controller.yaml
+++ b/bindata/cloud-network-config-controller/controller.yaml
@@ -54,6 +54,9 @@ spec:
         - name: cloud-provider-secret
           mountPath: "/etc/secret/cloudprovider"
           readOnly: true
+        - name: bound-sa-token
+          mountPath: /var/run/secrets/openshift/serviceaccount
+          readOnly: true
         terminationMessagePolicy: FallbackToLogsOnError
 {{- if not .ExternalControlPlane }}
       nodeSelector:
@@ -70,3 +73,11 @@ spec:
       - name: cloud-provider-secret
         secret:
           secretName: cloud-credentials
+      # This service account token can be used to provide identity outside the cluster.
+      # For example, this token can be used with AssumeRoleWithWebIdentity to authenticate with AWS using IAM OIDC provider and STS.
+      - name: bound-sa-token
+        projected:
+          sources:
+          - serviceAccountToken:
+              path: token
+              audience: openshift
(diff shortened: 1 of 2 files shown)


alexanderConstantinescu (approved):
Feels like we need a Hypershift CI job...

I am actually not even sure if we should be deploying the cloud-network-config-controller on Hypershift. The genesis of the component is: being able to create egress IPs on public clouds. I am not sure if egress IP and Hypershift even make logical sense as compatible "features", I know way too little about what Hypershift is/does/works to be able to fully tell. I will need to talk to product management about this. We can however deal with that later, so for now: 

/lgtm

alvaroaleman:
Hypershift means that there are no master nodes and most of what would run on them instead runs as pods in a different cluster.

I suppose the underlying need for the egress ips is to communicate with endpoints that require to allowlist source IPs? 

/assign @dcbw

alexanderConstantinescu:
> Hypershift means that there are no master nodes and most of what would run on them instead runs as pods in a different cluster.

Can this cloud-network-config-controller (which runs on this different cluster) access/configure the host on other clusters? I guess the roles added by https://github.com/openshift/hypershift/pull/804 allows for that. I am also seeing that the changes you're proposing only apply to AWS, so the "Hypershift control plane" is on AWS? Can/is it intended to configure "other clusters" on other providers?   

> I suppose the underlying need for the egress ips is to communicate with endpoints that require to allowlist source IPs?

Yes. In the classical cluster model: customers chose nodes in their cluster (by labeling them) which should act as "egress nodes" and service this egress IP intended traffic. The network plugin will then, with the cloud-network-config-controller, configure what is required on the cloud provider's and cluster networking's end to be able assign/remove the tuple (egress IP - to - egress node). If this Hypershift model is having a "control plane cluster" running on AWS and configuring other "data plane clusters" on GCP / Azure, then this won't work...but that might not be the case?

alvaroaleman:
> Can this cloud-network-config-controller (which runs on this different cluster) access/configure the host on other clusters?

It gets the same role as on self-hosted OCP which allows describing and managing network interfaces for all ec2 instances in the account.

>  I am also seeing that the changes you're proposing only apply to AWS, so the "Hypershift control plane" is on AWS? Can/is it intended to configure "other clusters" on other providers?

So today we only support AWS (or no cloud provider at all). This is intended to be extended in the future. Note that it technically doesn't matter where the management cluster that hosts the controlplane pods for the guest cluster runs, the aws-only limitation only applies to guest clusters.

> If this Hypershift model is having a "control plane cluster" running on AWS and configuring other "data plane clusters" on GCP / Azure, then this won't work...but that might not be the case?

Why would this not work? From the perspective of the cluster user (and the new controller for that matter), a hypershift cluster is like any other OCP cluster, it just happens to not have master nodes.

alexanderConstantinescu:
> Why would this not work? From the perspective of the cluster user (and the new controller for that matter), a hypershift cluster is like any other OCP cluster, it just happens to not have master nodes.

I didn't mean "it can't logically work", I meant "it doesn't work today with the existing code". This is to say: currently, the CNCC can only manage IP address for nodes in the same cluster and cloud provider as it itself is deployed on. This is not to say that we can't change that in the future as to have the CNCC deployed on a AWS Hypershift control plane and have it manage IP addresses for nodes in a Azure guest cluster. My concern was that this is the current model for Hypersift, in which case we should not have proceed with this PR and should not have had the CNCC be deployed at all on Hypershift. However, this is not the case currently, so we are fine.

alexanderConstantinescu (approved):
/lgtm

squeed:
/approve
seems that everyone is doing this now

alexanderConstantinescu:
@squeed : both [e2e-aws-sdn-multi](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-aws-sdn-multi) and [e2e-agnostic-upgrade](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-e2e-agnostic-upgrade) are perma failing. Could you please override?

PR #1269, opened 2021-12-24
Bug 2018093: Kuryr: Add resource requests for pods

dulek:


dulek:
Looks good, I tried a deployement with this and it works just fine.

PR #1277, opened 2022-01-12
BUG 2034413: cncc: create Cloud CredentialsRequest in /manifests

squeed:
It turns out, that thanks to https://github.com/openshift/enhancements/blob/master/enhancements/installer/credentials-management-outside-openshift-cluster.md, we need to be embedding all CredentialsRequests in `manifests` instead of dynamically in the CNO itself. So, do that. It's a bit awkward but doesn't actually break anything.

Also,
- tighten up RBAC for the CNCC. We accidentally added access to all Secrets, oops.
- don't blanket tolerate NotReady -- we're not critical to startup, and this keeps us from getting Pod creation errors

### DIFF
--- a/bindata/cloud-network-config-controller/002-rbac.yaml
+++ b/bindata/cloud-network-config-controller/002-rbac.yaml
@@ -31,6 +31,14 @@ rules:
   - list
   - watch
   - update
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: cloud-network-config-controller
+  namespace: openshift-cloud-network-config-controller
+rules:
 - apiGroups: 
   - ""
   resources:
@@ -39,6 +47,7 @@ rules:
   - get
   - create
   - update
+  - patch
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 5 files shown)


squeed:
/hold
until #1275 goes in

squeed:
/hold cancel
@alexanderConstantinescu this is ready for a final review, thanks!

squeed:
Test failures seem unrelated; the cluster came up without issues.
/retest-required

bindata/cloud-network-config-controller/002-rbac.yaml
```diff
@@ -39,6 +47,7 @@ rules:
   - get
   - create
   - update
+  - patch
```
alexanderConstantinescu, line 50: Why was patch added? It's not "tightening up" and is not needed (presumably since it has worked without it)

alexanderConstantinescu (approved):
I have a small nit, but since this is quite urgent to get in and I am not _that_ much against it:

/lgtm

squeed:
> On Jan 13, 2022, at 14:02, Alexander Constantinescu ***@***.***> wrote:
> 
> ﻿
> @alexanderConstantinescu approved this pull request.
> 
> I have a small nit, but since this is quite urgent to get in and I am not that much against it:
> 
> /lgtm
> 
> In bindata/cloud-network-config-controller/002-rbac.yaml:
> 
> > @@ -39,6 +47,7 @@ rules:
>    - get
>    - create
>    - update
> +  - patch
> Why was patch added? It's not "tightening up" and is not needed (presumably since it has worked without it)
> 

I added patch just because it’s implied with update. I’ve had the experience of bumping a library that internally switched from update to patch, and then something broke.

PR #1271, opened 2022-01-05
Bug 2037500: Bump openshift/build-machinery-go

MaysaMacedo:
Bumps openshift/build-machinery-go to fix
openshift/build-machinery-go#56.

PR #1278, opened 2022-01-13
[release-4.8] Bug 2021221: ovnkube: set ovn-controller lflow cache limit to 1GB

jcaamano:
By default the lflow cache is unlimited, but at scale this
makes ovn-controller consume large amounts of RSS and cause OOM
situations. Limit the cache size to prevent unbounded memory
usage while still providing enough for most setups.

The lflow cache is a tradeoff between CPU and memory, but we'd
rather use a bit more CPU than OOM a box.

Pick a 1GB limit based on data gathered from various scale runs,
which seems to be the upper-limit (for now) for 120-node
density runs.

(cherry picked from commit 4d1056a590678ddacd1a884242a0f606bbac93c2)

### DIFF
--- a/bindata/network/ovn-kubernetes/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/004-config.yaml
@@ -11,6 +11,8 @@ data:
     mtu="{{.MTU}}"
     cluster-subnets="{{.OVN_cidr}}"
     encap-port="{{.GenevePort}}"
+    enable-lflow-cache=true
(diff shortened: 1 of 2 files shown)


jcaamano:
we need https://github.com/openshift/cluster-network-operator/pull/1271

jcaamano:
@anuragthehatter is QE in charge now of setting the cherrypick-approved label?

PR #1280, opened 2022-01-14
Bug 2041989: no CredentialsRequests in ibm-cloud-managed

joelddiaz:
The ibm-cloud-managed profile doesn't use the cloud-credential-operator.
The CredentialsRequest CRs should not be installed in that environment.

Unmark the existing CredentialsRequests so they are no longer installed
in ibm-cloud-managed.

Create a list of previously-installed CredentialsRequests so that they
are cleaned up by the CVO only on the ibm-cloud-managed profile.

xref: https://issues.redhat.com/browse/CCO-177

### DIFF
--- a/manifests/0000_70_cluster-network-operator_01_credentialsrequest.yaml
+++ b/manifests/0000_70_cluster-network-operator_01_credentialsrequest.yaml
@@ -7,7 +7,6 @@ metadata:
   namespace: openshift-cloud-credential-operator
   annotations:
     include.release.openshift.io/self-managed-high-availability: "true"
-    include.release.openshift.io/ibm-cloud-managed: "true"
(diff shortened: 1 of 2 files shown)


squeed:
@joelddiaz do we need to add the deletions? This was never shipped.

joelddiaz:
@squeed If this PR will make it in before 4.10, then the deletions aren't needed. I can update it to just remove the annotations if we think this PR has a chance for 4.10.

squeed:
@joelddiaz definietly, this will make it in in the next few days.

joelddiaz:
@squeed ok, i've remove the list of resources marked for deletion

squeed:
@joelddiaz this will need a bz to merge.

PR #1225, opened 2021-11-12
Bug 2022747: Allow to use proxy to connect to OSP cloud

MaysaMacedo:
When using Proxy and the installer is configured
on a restricted installation, Kuryr bootstrap
can't progress as the http client is not
configured to use Proxy. This commit ensures
proxy is used by relying on the cluster proxy
resource instead of relying on cvo to injecting
the proxy env vars.

PR #1284, opened 2022-01-18
[release-4.7] Bug 2041893: Allow to use proxy to connect to OSP cloud

MaysaMacedo:
When using Proxy and the installer is configured
on a restricted installation, Kuryr bootstrap
can't progress as the http client is not
configured to use Proxy. This commit ensures
proxy is used by relying on the cluster proxy
resource instead of relying on cvo to injecting
the proxy env vars.


MaysaMacedo:
Manual backport of https://github.com/openshift/cluster-network-operator/pull/1225

MaysaMacedo:
/cc @trozet 
Can you evaluate the backport-risk-assessed?

MaysaMacedo:
@anuragthehatter if you see fit, can you apply the cherry-pick-approved label here?
Thanks

PR #1297, opened 2022-01-31
The Whereabouts ip-reconciler should use the internal load balancer

dougbtv:
This gives it connectivity during cluster lifecycle events where the SDN is in a transitioning state.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -507,6 +507,11 @@ spec:
               volumeMounts:
                 - name: cni-net-dir
                   mountPath: /host/etc/cni/net.d
+              env:
+              - name: KUBERNETES_SERVICE_PORT
+                value: "{{.KUBERNETES_SERVICE_PORT}}"
+              - name: KUBERNETES_SERVICE_HOST
+                value: "{{.KUBERNETES_SERVICE_HOST}}"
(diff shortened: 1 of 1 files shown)


dougbtv:
/retitle Bug 2048575: 
The Whereabouts ip-reconciler should use the internal api load balancer

dougbtv:
/retitle Bug 2048575: The Whereabouts ip-reconciler should use the internal api load balancer

deads2k:
@dougbtv I think your CI namespace is dirty.  Try opening another PR with a different SHA (update the commit comment or something)

deads2k:
> @deads2k am I making any sense ? Isn't this PR just re-setting already set variables (to the same values ...) ?

it is not.  Have a look at the actual results.

```
    env:
    - name: KUBERNETES_SERVICE_PORT
      value: "6443"
    - name: KUBERNETES_SERVICE_HOST
      value: api-int.ci-op-rj7rtjc2-9a01a.*********************************
```

from the ip-reconciler pod.  That names doesn't look quite right to me.  Bad substitution?  It's the same one in the multus pods, but I'm guessing those are doing a substitution in the pod?

I think the alternative is dropping the not-ready toleration.

dougbtv:
Closing in favor of #1302

PR #1298, opened 2022-02-01
Bug 2048413: Fix bond cni source directory path

mmirecki:
The bond cni source directory path is incorrect. It is missing a trailing
directory separator. Without the separator the cnicopy.sh command will not
copy the contents of the directory.


cgoncalves:
/bugzilla refresh

@mmirecki I've triaged a bit the 4.10 BZ. We still need a 4.11 BZ to please the openshift-ci bot.

SchSeba:
@mmirecki I am not sure you can mark the BZ as 4.10.0

PR #1299, opened 2022-02-01
Bug 2049613: Use a separate configmap for mtu migration config to avoid pod restart

kyrtapz:
SDN pods get restarted once `sdn-config` configmap changes, this caused transient issues with pods being deployed with wrong MTU during MTU migration.

Use a separate, optional config map that is used for mtu migration.


danwinship:
/lgtm
though I'm increasingly thinking that the right answer is that we never want to auto-restart pods on config change

jcaamano:
e2e-agnostic-upgrade is permafailing

danwinship:
job is busted and this PR should not break upgrades if it does not break non-upgrades

PR #1302, opened 2022-02-07
Bug 2048575: The Whereabouts ip-reconciler should use api-int load balancer

dougbtv:
It should also use the internal load balancer.

This gives it connectivity during cluster lifecycle events where the SDN is in a transitioning state.

PR #1305, opened 2022-02-09
Bug 2051639: The Whereabouts ip-reconciler should use the internal load balancer and host network [backport 4.10]

dougbtv:
This gives it connectivity during cluster lifecycle events where the SDN is in a transitioning state.

backport of #1302


omertuc:
The alert still seems to be raised on SNO, and [also](https://bugzilla.redhat.com/show_bug.cgi?id=2050409) on non-SNO clusters now (not sure if it's related to this PR or not)

PR #1308, opened 2022-02-10
Bug 2049590: Limit reconcile loop for cm controller

martinkennelly:
If a user erroneously applies label
config.openshift.io/inject-trusted-cabundle=true and
annotation service.beta.openshift.io/inject-cabundle=true to
a config map, it causes the reconcile loop to be called
at a regular short interval due to updates to the configmap
from the service CA operator which owns the annotation.
This causes repeated requests to the API server and
doesn't inform the user of the issue.
This change will limit the API requests and inform the admin of
the issue via operator degraded status message.

### DIFF
--- a/pkg/controller/configmap_ca_injector/controller.go
+++ b/pkg/controller/configmap_ca_injector/controller.go
@@ -163,6 +163,15 @@ func (r *ReconcileConfigMapInjector) Reconcile(ctx context.Context, request reco
 				log.Println(err)
 				return err
 			}
+			// An infinitely occurring reconcile loop may occur if the user incorrectly annotates a config map with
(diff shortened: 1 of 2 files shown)


trozet:
/hold 

wait for @squeed input and potential alternative

squeed:
Oh nuts, do they both set the same key? `ca-bundle.crt`?

squeed:
No, they set different keys. This *should* merge the Data field, and thus cause a noop update. Is the service-ca-operator overwriting our key?

martinkennelly:
I see now. Closing this and will get them to fix their cm reconcile. Thanks Casey.

PR #1313, opened 2022-02-22
Probe MTU from a Job, rather than directly in the CNO

squeed:
This creates, if needed, a Job that writes the cluster's MTU to a configmap. it waits for that configmap to be created.

As an absolute emergency fallback, it will probe the node's MTU, but this is paranoia rather than expected to be executed.


squeed:
@kyrtapz this is ready for review - it Works On My Machine :-)

squeed:
/hold cancel
this should be ready for review and merging.

squeed:
Okay, CI is looking good now. This should be OK to merge (after review).

squeed:
@kyrtapz can you take a look?

cmd/cluster-network-operator/mtu-probe.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		mtu, err := network.GetDefaultMTU()
+		if err != nil {
+			return err
+		}
+		fmt.Println("Detected node MTU:", mtu)
+
+		cm := v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
```
kyrtapz, line 62: Please correct me if I am wrong but I am concerned that we are not setting the `ownerReference`, would that mean that this is not gathered when we do a must-gather?

cmd/cluster-network-operator/mtu-probe.go
```diff
@@ -0,0 +1,89 @@
...
+				Namespace: namespace,
+				Name:      name,
+			},
+			Data: map[string]string{
+				"mtu": strconv.Itoa(mtu),
+			},
+		}
+
+		// Write the CM in the apiserver, retrying as needed.
+		for tries := 0; tries < 10; tries++ {
```
kyrtapz, line 72: Can we use `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate` instead of the for loop?

bindata/network/mtu-probe/02-job.yaml
```diff
@@ -0,0 +1,47 @@
...
+  selector: {} #need this to make the apiserver happy :/
+  template:
+    spec:
+      containers:
+      - name: prober
+        image: {{.CNOImage}}
+        command:
+        - /usr/bin/cluster-network-operator
+        - probe-mtu
+        - --namespace=openshift-network-operator
```
kyrtapz, line 20: `namespace` and `name` values should templated

pkg/controller/operconfig/mtu_probe.go
```diff
@@ -0,0 +1,143 @@
...
+	// cm doesn't exist, create Job
+	err = deployMTUProber(ctx, c)
+	if err != nil {
+		return 0, fmt.Errorf("failed to deploy mtu prober: %w", err)
+	}
+
+	log.Println("MTU prober deployed, waiting for result ConfigMap")
+
+	// wait up to 100 seconds for Job to report back.
+	tries := 10
```
kyrtapz, line 50: Can we use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate instead of the for loop?

pkg/controller/operconfig/mtu_probe.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return mtu, nil
+	}
+
+	return 0, fmt.Errorf("timed out getting result from MTU prober %v", err)
+
+}
+
+func readMTUConfigMap(ctx context.Context, c crclient.Client) (int, error) {
+	log.Printf("Looking for ConfigMap %s/%s", cmNamespace, cmName)
```
kyrtapz, line 75: Should that be changed to `klog.Infof`?

pkg/network/openshift_sdn.go
```diff
@@ -260,12 +261,23 @@ func fillOpenShiftSDNDefaults(conf, previous *operv1.NetworkSpec, hostMTU int) {
...
 	if sc.MTU == nil {
-		var mtu uint32 = uint32(hostMTU) - 50 // 50 byte VXLAN header
+		var mtu uint32
 		if previous != nil &&
 			previous.DefaultNetwork.Type == operv1.NetworkTypeOpenShiftSDN &&
 			previous.DefaultNetwork.OpenShiftSDNConfig != nil &&
 			previous.DefaultNetwork.OpenShiftSDNConfig.MTU != nil {
 			mtu = *previous.DefaultNetwork.OpenShiftSDNConfig.MTU
+		} else {
+			// utter paranoia
```
kyrtapz, line 271: We did, it was previously done in `FillDefaults`, this changed in this PR

cmd/cluster-network-operator/mtu-probe.go
squeed: Fair enough; given that this is very quickly deleted, I hadn't bothered. But, you're right - we should.

bindata/network/mtu-probe/02-job.yaml
squeed: agreed.

pkg/controller/operconfig/mtu_probe.go
squeed: derp, yes. Switched everything to `klog`

pkg/network/openshift_sdn.go
squeed: The logic that determines whether or not we should bother probing MTU is slightly complex, so I kept this in to ensure we don't accidentally break the CNO. Hopefully this never happens.

cmd/cluster-network-operator/mtu-probe.go
squeed: good point, yes.

pkg/controller/operconfig/mtu_probe.go
squeed: yup

cmd/cluster-network-operator/mtu-probe.go
squeed: Actually, no, that has the wrong semantics: if an error occurs, we want to keep trying, not error out.

squeed:
@kyrtapz updated, thanks for the review.

cmd/cluster-network-operator/mtu-probe.go
kyrtapz: If we want to retry we can just return `false, nil` from `ConditionFunc` instead of returning the error.

squeed: right; ultimately, I want to retry a set number of times; `wait.Poll()` families are more intended for retrying a number of times before a deadline.

cmd/cluster-network-operator/mtu-probe.go
squeed: AIUI because this is in the `openshift-network-operator` namespace, `must-gather` is going to gather it regardless. So there's no need.

(I did set the OwnerReference for everything else, fwiw)

PR #1315, opened 2022-02-23
ovn: stop spawning the ovn-nbctl daemon

dcbw:
All normal operation has been converted over to libovsdb; we no longer
use ovn-nbctl for anything other than one-time checks and properties at
startup.

As such we no longer need the nbctl daemon keeping another whole copy
of the NB database in memory. Let's just use libovsdb.

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -751,19 +751,6 @@ spec:
             exit 1
           fi
 
-          # start nbctl daemon for caching
-          echo "I$(date "+%m%d %H:%M:%S.%N") - ovnkube-master - start nbctl daemon for caching"
-          export OVN_NB_DAEMON=$(ovn-nbctl --pidfile=/var/run/ovn/ovn-nbctl.pid \
-            --detach \
-            -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt \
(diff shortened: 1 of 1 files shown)


dcbw:
/retest
AWS setup issues

jcaamano:
/lgtm

@dcbw cross-checking a couple of things:
* I guess we should not worry about this op
https://github.com/ovn-org/ovn-kubernetes/blob/bd4a607ca4ae304a624db858f90ad9c1c19db55d/go-controller/hybrid-overlay/pkg/controller/node_linux.go#L250
* Does running nbctl a single time in non-damon mode still cache all the DB even if ephemerally? If so, is there an arg to it to avoid it?

dcbw:
> * I guess we should not worry about this op
>   https://github.com/ovn-org/ovn-kubernetes/blob/bd4a607ca4ae304a624db858f90ad9c1c19db55d/go-controller/hybrid-overlay/pkg/controller/node_linux.go#L250
> 
> * Does running nbctl a single time in non-damon mode still cache all the DB even if ephemerally? If so, is there an arg to it to avoid it?

@jcaamano for the first, that code actually never runs and has gotten stale. We should either kill the Linux bits here or rewrite that part to pull the subnet from the node annotation instead of OVN NB.

For the second question, no, nbctl won't download the full DB anymore as of OVN 21.09 and later. It will only download the tables it needs for a given command. eg for "ls-list" it'll only grab the Logical_Switch table, not everything.

PR #1319, opened 2022-02-24
Multi cluster support in CNO

kyrtapz:


cmd/cluster-network-operator/main.go
```diff
@@ -78,13 +79,15 @@ which is a kubeconfig from which to take just the URL to the apiserver`,
...
+	var managedClusters *map[string]string
+	cmdcfg := controllercmd.NewControllerCommandConfig("network-operator", version.Get(), func(ctx context.Context, controllerConfig *controllercmd.ControllerContext) error {
+		return operator.RunOperator(ctx, controllerConfig, *managedClusters)
+	})
 
 	cmd2 := cmdcfg.NewCommand()
 	cmd2.Use = "start"
 	cmd2.Short = "Start the cluster network operator"
-
+	managedClusters = cmd2.Flags().StringToString("managed-clusters", nil, "managed clusters, pairs of cluster name and kubeconfig path")
```
squeed, line 90: StringToString is cute :)

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -320,6 +322,14 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 				return reconcile.Result{}, err
 			}
 		}
+		anno := obj.GetAnnotations()
```
squeed, line 325: I think I'd rather this live in `pkg/apply`, since the need to copy objects may not be limited to the operconfig controller.

squeed: Accordingly, you'll have to pass the full set of Clients to all the controllers. But that was inevitable anyways.

pkg/apply/apply.go
```diff
@@ -30,8 +31,12 @@ type Object interface {
 func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subcontroller string) error {
 	name := obj.GetName()
 	namespace := obj.GetNamespace()
+	clusterClient := client.ClientFor(obj.GetClusterName())
+	if clusterClient == nil {
+		return fmt.Errorf("client not defined for %v", obj)
```
squeed, line 36: wording suggestion: "Object %s/%s specifies unknown cluster %s", namespace, name, ...

pkg/apply/apply.go
```diff
@@ -51,15 +56,26 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
 	// It isn't allowed to send ManagedFields in a Patch.
 	obj.SetManagedFields(nil)
 
+	anno := obj.GetAnnotations()
+	if anno != nil {
```
squeed, line 60: I _think_ you don't need this nil check (see the create-only code below).

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
 	}
-	if _, err := client.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions); err != nil {
+	if _, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions); err != nil {
 		return fmt.Errorf("failed to apply / update %s: %w", objDesc, err)
 	}
 	log.Printf("Apply / Create of %s was successful", objDesc)
 	return nil
 }
+
+func getObjectCopy(ctx context.Context, obj Object, client *cnoclient.Client) (Object, error) {
```
squeed, line 112: two things:
1. a brief docblock please
2. naming: getCopySource or something?

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	anno := obj.GetAnnotations()
+	if anno == nil {
+		return nil, fmt.Errorf("no annotations defined")
+	}
+
+	parts := strings.Split(anno[names.CopyFromAnnotation], "/")
```
squeed, line 118: Will this work for non-namespaced objects? Is it just `defaut//CoolRoleBinding` ?

pkg/apply/apply.go
```diff
@@ -51,15 +56,26 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
 	// It isn't allowed to send ManagedFields in a Patch.
 	obj.SetManagedFields(nil)
 
+	anno := obj.GetAnnotations()
+	if anno != nil {
+		if _, exists := anno[names.CopyFromAnnotation]; exists {
+			var err error
+			obj, err = getObjectCopy(ctx, obj, client)
+			if err != nil {
+				return err
```
squeed, line 65: Wrap this error with some useful text: `return fmt.Errorf("failed to retrieve copy-from object: %w", err)`

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	parts := strings.Split(anno[names.CopyFromAnnotation], "/")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("'%s' annotation is invalid, expected: ClusterName/Namespace/ResourceName, got: %s", names.CopyFromAnnotation, anno[names.CopyFromAnnotation])
+	}
+
+	clusterName, namespace, name := parts[0], parts[1], parts[2]
+
+	cli := client.ClientFor(clusterName)
+	if cli == nil {
+		return nil, fmt.Errorf("failed to find the client for %s", clusterName)
```
squeed, line 127: `"cluster %s is unknown"`

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+
+	err := cli.CRClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, ret)
+	if err != nil {
+		return nil, err
+	}
+
+	ret.SetNamespace(obj.GetNamespace())
+	ret.SetName(obj.GetName())
+	ret.SetClusterName(obj.GetClusterName())
+	ret.SetAnnotations(obj.GetAnnotations())
```
squeed, line 141: Should we overwrite or merge annotations?

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	ret.SetLabels(obj.GetLabels())
+	ret.SetOwnerReferences(obj.GetOwnerReferences())
+
+	ret.SetCreationTimestamp(obj.GetCreationTimestamp())
```
squeed, line 145: timestamp, link, generation, uid, resourceVersion are read-only (provided by the apiserver)

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	ret.SetSelfLink(obj.GetSelfLink())
+	ret.SetGeneration(obj.GetGeneration())
+	ret.SetUID(obj.GetUID())
+	ret.SetResourceVersion(obj.GetResourceVersion())
+	ret.SetManagedFields(obj.GetManagedFields())
```
squeed, line 150: We cant touch managed fields.

pkg/apply/apply.go
squeed: (Update: let's merge annotations)

pkg/apply/apply.go
```diff
@@ -86,9 +102,53 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	err := cli.CRClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, ret)
+	if err != nil {
+		return nil, err
+	}
+
+	ret.SetNamespace(obj.GetNamespace())
+	ret.SetName(obj.GetName())
+	ret.SetClusterName(obj.GetClusterName())
+	ret.SetAnnotations(obj.GetAnnotations())
+	ret.SetLabels(obj.GetLabels())
```
squeed, line 142: Let's not do this - labels shouldn't be copied over.

kyrtapz: they are not, `ret.SetLabels(obj.GetLabels())` means that we are not copying in the annotations but using what is there in obj(`ret` is the copy and it is being applied)

pkg/apply/apply.go
squeed: Aha, this is subtle. Fun. We need to be careful about all of the ObjectMeta fields.
- namespace, name: from the requesting object
- annotations: merge from requesting and source objects?
- labels: from the requesting object?
- owners: from the requesting object
- finalizers: from the requesting object (or just don't support setting them)

all other ObjectMeta fields should be empty.

squeed: Oh, ResourceVersion should be from the requesting object too.

pkg/apply/apply.go
squeed: My comment was wrong re: resourceVersion.

pkg/apply/apply.go
```diff
@@ -86,9 +98,72 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+	if cli == nil {
+		return nil, fmt.Errorf("cluster %s is unknown", clusterName)
+	}
+
+	ret := &unstructured.Unstructured{}
+	ret.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+
+	err := cli.CRClient().Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, ret)
+	if err != nil {
+		return nil, err
```
squeed, line 141: `fmt.Errorf("get failed (%s %s) %s/%s: %w", group, kind, namespace, name, err)`

pkg/apply/apply.go
```diff
@@ -86,9 +98,72 @@ func ApplyObject(ctx context.Context, client *cnoclient.Client, obj Object, subc
...
+// Annotations are merged, when there is a conflict obj's annotation is used.
+func getCopySource(ctx context.Context, obj Object, client *cnoclient.Client) (Object, error) {
+	anno, exists := obj.GetAnnotations()[names.CopyFromAnnotation]
+	if !exists {
+		return nil, fmt.Errorf("%s annotation not specified", names.CopyFromAnnotation)
+	}
+
+	parts := strings.Split(anno, "/")
+	if len(parts) != 3 {
+		return nil, fmt.Errorf("'%s' annotation is invalid, expected: ClusterName/Namespace/ResourceName, got: %s", names.CopyFromAnnotation, anno)
```
squeed, line 126: s/ResourceName/name/
"resource" in Kubernetes has a specific meaning.

pkg/apply/apply.go
kyrtapz: yes, I was able to verify that it works

squeed:
Update: we will move mocking in to a new PR.

squeed:
good test

squeed:
clearly broken

PR #1320, opened 2022-02-24
Bug 2058368: Move memory-trimming-on-compaction out of dbchecker to nbdb/sbdb

msherif1234:
In order to avoid any potential bringup race conditions between dbchecker container
and sbdb/nbdb we move the memory trimming check and enable to start of each db.
upstream PR#[2830](https://github.com/ovn-org/ovn-kubernetes/pull/2830) will remove
enable memory trimming from dbchecker container


trozet (approved):
/lgtm

PR #1328, opened 2022-03-02
raise the alert NoOvnMasterLeader to critical and add the runbook url

bmeng:
raise the severity level for the alert and add the runbook

should be merged after this https://github.com/openshift/runbooks/pull/41


martinkennelly:
@bmeng The url doesnt work for me.

PR #1335, opened 2022-03-09
[wip] render: don't get apiserver from the environment, pass explicitly

squeed:
A tiny mechanical change that touched tons and tons of code :(

/cc @kyrtapz


squeed:
/hold
this is based on another pending PR

PR #1338, opened 2022-03-11
Fix MTU detection for multi path default routes

kyrtapz:
Multi path routes do not have a link index attribute set on the route, instead each path has it's own link index.

### DIFF
--- a/pkg/network/mtu.go
+++ b/pkg/network/mtu.go
@@ -4,6 +4,7 @@
 package network
 
 import (
+	"fmt"
 	"github.com/pkg/errors"
 	"github.com/vishvananda/netlink"
 )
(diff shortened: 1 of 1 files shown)


squeed:
Seems reasonable. While we're in here touching this code, how about we add a fallback mode that picks the first up non-veth with a L3 address?

kyrtapz:
> Seems reasonable. While we're in here touching this code, how about we add a fallback mode that picks the first up non-veth with a L3 address?

Don't we have a requirement that the default route is defined?
For ovn-k we check it in `configure-ovs` https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml#L510

kyrtapz:
@jcaamano PTAL

squeed:
> Don't we have a requirement that the default route has to be defined?

Sure. Was just thinking of extra paranoia to keep the progress moving forward
/approve
/lgtm
/retest-required

PR #1341, opened 2022-03-17
[WIP] hypershift route status

zshi-redhat:


zshi-redhat:
fyi @squeed @kyrtapz

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -335,6 +335,17 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	allResources = append(allResources, deployments...)
 	r.podReconciler.SetResources(allResources)
 
+	// Remove ovnkube-node object if ovn db route doesn't exist in hypershift management cluster
+	if bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Enabled && bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Route == "" {
+		log.Printf("Removing ovnkube-node objects from rendered objs")
+		for i, obj := range objs {
+			if obj.GetName() == "ovnkube-node" && obj.GetKind() == "DaemonSet" {
+				objs = append(objs[:i], objs[i+1:]...)
+				r.status.SetDegraded(statusmanager.OperatorConfig, "ApplyOperatorConfig", "Waiting ovn db route to be created")
```
squeed: Suggestion: add a `status.SetProgressing` function that does the same thing.

zshi-redhat: added status.SetProgressing

pkg/apply/apply.go
```diff
@@ -24,6 +24,15 @@ type Object interface {
 	runtime.Object
 }
 
+func RemoveObject(objs []*unstructured.Unstructured, name string, kind string) []*unstructured.Unstructured {
```
squeed, line 27: There's already a function in `pkg/util/k8s` called `RemoveObjectByKindGroupName` that does this.
`

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -335,6 +335,13 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	allResources = append(allResources, deployments...)
 	r.podReconciler.SetResources(allResources)
 
+	// Remove ovnkube-node object if ovn db route doesn't exist in hypershift management cluster
+	if bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Enabled && bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Route == "" {
```
squeed, line 339: Do we ever need to worry about the route changing or going away? If that happens, what should / could we do?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -335,6 +335,13 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+		log.Printf("Removing ovnkube-node objects from rendered objs")
+		objs = apply.RemoveObject(objs, "ovnkube-node", "DaemonSet")
```
squeed, line 341: This logic should be in `pkg/render` IMHO. Then we need to return an additional "Progressing" status from `Render()`. Which we should have been doing all along, especially with the logic to wait for node rollout on upgrades...

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -397,7 +404,9 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
-	r.status.SetNotDegraded(statusmanager.OperatorConfig)
+	if !bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Enabled || bootstrapResult.OVN.OVNKubernetesConfig.HypershiftConfig.Route != "" {
+		r.status.SetNotDegraded(statusmanager.OperatorConfig)
```
squeed, line 408: Presumably this should be SetNotProgressing...

zshi-redhat: I was thinking `SetNotProgressing` might sound that the operator config is not progressing.
By SetNotDegraded, it will override the progressing operator config, right?

kyrtapz: Is `SetProgressing/UnsetProgressing` more clear in that context?

pkg/apply/apply.go
zshi-redhat: Moved this logic to `pkg/network/ovn_kubernetes.go` and use `k8s.UpdateObjByGroupKindName`.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: I updated the PR with the following changes:
1) add create-wait annotation (similar to create-only) to indicate that the ovnkube-node object needs to wait
2) return a progressing boolean value in render* all the way back to operator-controller
3) set/unset the progressing condition based on the returned value from render*.

Is this what you have in mind?

I wonder if we can just check the existense of create-wait annotation to set/unset the progressing status.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: changed to `SetProgressing/UnsetProgressing`.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: Maybe mark the operator status to be degraded=true if route goes away for certian period of time?
If route changes are detected, ovnk-node will be reconciled, we do nothing?

squeed:
Seems reasonable! This looks great.

Do we also want to create the Route in this PR?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -312,6 +312,52 @@ func (status *StatusManager) SetNotDegraded(statusLevel StatusLevel) {
 	status.setNotDegraded(statusLevel)
 }
 
+// syncProgressing syncs the current Progressing status
+func (status *StatusManager) syncProgressing() {
+	for _, c := range status.failing {
```
squeed, line 317: I think you need to check here if the status.Type is Progressing.

squeed:
/lgtm cancel
one small thing I just thought of

pkg/controller/statusmanager/status_manager.go
zshi-redhat: fixed

pkg/network/ovn_kubernetes.go
```diff
@@ -341,7 +342,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		objs = k8s.RemoveObjByGroupKindName(objs, "apps", "DaemonSet", "openshift-ovn-kubernetes", "ovnkube-upgrades-prepuller")
 	}
 
-	return objs, nil
+	if bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.Enabled && bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.Route == "" {
+		k8s.UpdateObjByGroupKindName(objs, "apps", "DaemonSet", "openshift-ovn-kubernetes", "ovnkube-node", func(o *uns.Unstructured) {
+			anno := o.GetAnnotations()
```
kyrtapz, line 347: I think `GetAnnatations` could return `nil`, we might need to add a check if it does

zshi-redhat: Added a conditional check on `anno == nil`.

zshi-redhat:
Rebased to resolve conflict.

abhat:
perma-failing job.

squeed:
/override ci/prow/e2e-gcp-ovn 
stop the bleeding; this PR has been green many times over

PR #1342, opened 2022-03-17
Add an option to define the client name for in-cluster config

kyrtapz:
For usual deployments the in-cluster config is the one for the `default` clusterClient, in hypershift the in-cluster config will be for the `management` clusterClient instead.
This change allows us to define which client is tied to the in-cluster config.

I initially tried to create an additional kubeconfig from the ServiceAccount token but it seems that it is not the best idea since the token can expire: https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/1205-bound-service-account-tokens/README.md#safe-rollout-of-time-bound-token.
Currently it is not a huge deal since the expiration is set to 1year, but we do get the following audit log in kube-apiserver:
```
{"kind":"Event",...,"annotations":{"authentication.k8s.io/stale-token":"subject: system:serviceaccount:clusters-ovn-hypershift:cluster-network-operator, seconds after warning threshold: 3234","authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":"RBAC: allowed by RoleBinding \"cluster-network-operator/clusters-ovn-hypershift\" of Role \"cluster-network-operator\" to ServiceAccount \"cluster-network-operator/clusters-ovn-hypershift\""}}
```

### DIFF
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"log"
 	"math/rand"
 	"net/url"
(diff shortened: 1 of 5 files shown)


cmd/cluster-network-operator/main.go
```diff
@@ -80,14 +81,16 @@ which is a kubeconfig from which to take just the URL to the apiserver`,
...
 	cmdcfg := controllercmd.NewControllerCommandConfig("network-operator", version.Get(), func(ctx context.Context, controllerConfig *controllercmd.ControllerContext) error {
-		return operator.RunOperator(ctx, controllerConfig, *extraClusters)
+		return operator.RunOperator(ctx, controllerConfig, *inClusterClient, *extraClusters)
 	})
 
 	cmd2 := cmdcfg.NewCommand()
 	cmd2.Use = "start"
 	cmd2.Short = "Start the cluster network operator"
 	extraClusters = cmd2.Flags().StringToString("extra-clusters", nil, "extra clusters, pairs of cluster name and kubeconfig path")
+	inClusterClient = cmd2.Flags().String("in-cluster-client", cnoclient.DefaultClusterName, "client name for in-cluster config(service account or kubeconfig)")
```
squeed, line 93: nit: `in-cluster-client-name`

squeed:
/approve
one small nit, then we can merge this.

squeed:
something funny is going on, it says it's waiting for a job that has succeeded (aws-multi). I'm gonna push some buttons.

squeed:
aha, CI crashed sometime last night.

PR #1343, opened 2022-03-17
OCPBUGS-9174: The cluster-readers group should be able to get net-attach-defs

dougbtv:
This adds a new clusterrolebinding for the cluster-readers group


zshi-redhat:
/retitle OCPBUGS-9174: The cluster-readers group should be able to get net-attach-defs

PR #1345, opened 2022-03-23
Add statefulset in status manager

zshi-redhat:


pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -264,15 +268,22 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
 	relatedObjects := []configv1.ObjectReference{}
 	for _, obj := range objs {
 		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
 			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
 			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "StatefulSet" {
+			statefulSets = append(statefulSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+			continue
+		} else if obj.GetClusterName() == cnoclient.ManagementClusterName {
```
kyrtapz, line 284: should that be changed to `if obj.GetClusterName() == cnoclient.ManagementClusterName {`?
If there is a daemonset or deployment for management cluster it won't hit that if

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -264,15 +268,22 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+	statefulSets := []types.NamespacedName{}
 	relatedObjects := []configv1.ObjectReference{}
 	for _, obj := range objs {
 		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
 			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
 			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "StatefulSet" {
+			statefulSets = append(statefulSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+			continue
```
kyrtapz, line 283: is that `continue` intended? I am not sure why we need that here

pkg/controller/statusmanager/status_manager.go
```diff
@@ -203,12 +214,86 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
...
 
+	if status.hypershiftConfig.Enabled {
+		err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			hcp := &hyperv1.HostedControlPlane{ObjectMeta: metav1.ObjectMeta{Name: status.hypershiftConfig.Name}}
+			gvr := schema.GroupVersionResource{
+				Group:    "hypershift.openshift.io",
+				Version:  "v1alpha1",
+				Resource: "hostedcontrolplanes",
+			}
+			hcpObj, err := status.client.ClientFor(cnoclient.ManagementClusterName).Dynamic().Resource(gvr).Namespace(
```
kyrtapz, line 225: is the dynamic client needed here? could we just use `CRClient` instead?

zshi-redhat: let me try CRClient, I remembered it didn't work with the route PR.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: you're right, daemonset/deployment also need to be added in relatedObjects.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: The `deleteRelatedObjectsNotRendered` in status manager doesn't support deleting resources in mgmt, thus I skip adding it to relatedObjects. I will fix it in the next push.

kyrtapz: I mean that the `continue` skips when an object is a `StatefulSet`, it shouldn't just based on that, right?

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: Fixed.
use separate `if` for management cluster objects.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: removed `continue`.

pkg/controller/statusmanager/status_manager.go
zshi-redhat: `CRClient` works.
Removed dynamic client.

zshi-redhat:
@kyrtapz This PR uses another env var passed from hypershift cno deployment `HOSTED_CLUSTER_NAME` to get and update the hostedControlPlane status: https://github.com/openshift/cluster-network-operator/pull/1345/commits/0c25effb0f9a1f4c6f1657747537bb06770aa201#diff-e071dcaa05eb3b54b619c7bf7a70844766015e2641f9298cd7c7dc1e944f21c4R310

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -264,21 +268,38 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
 			deployments = append(deployments, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "StatefulSet" {
+			statefulSets = append(statefulSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
 		}
 		restMapping, err := r.mapper.RESTMapping(obj.GroupVersionKind().GroupKind())
 		if err != nil {
 			log.Printf("Failed to get REST mapping for storing related object: %v", err)
 			continue
 		}
+		if obj.GetClusterName() == cnoclient.ManagementClusterName {
```
kyrtapz, line 290: Shouldn't that be `if obj.GetClusterName() != cnoclient.DefaultClusterName`?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -57,23 +59,31 @@ type deploymentState struct {
...
 	progressing := []string{}
 	hung := []string{}
 
-	daemonsetStates, deploymentStates := status.getLastPodState()
+	daemonsetStates, deploymentStates, statefulsetStates := status.getLastPodState()
 
 	for _, dsName := range status.daemonSets {
 		ds := &appsv1.DaemonSet{}
-		if err := status.client.Get(context.TODO(), dsName, ds); err != nil {
+		if err := status.client.Default().CRClient().Get(context.TODO(), dsName, ds); err != nil {
```
kyrtapz, line 86: Can we use `status.client.ClientFor(...` instead? if `clusterName` is not set it will pick the default client

pkg/controller/statusmanager/pod_status.go
```diff
@@ -135,9 +145,70 @@ func (status *StatusManager) SetFromPods() {
 		}
 	}
 
+	for _, ssName := range status.statefulSets {
+		ss := &appsv1.StatefulSet{}
+		if err := status.client.ClientFor(cnoclient.ManagementClusterName).CRClient().Get(context.TODO(), ssName, ss); err != nil {
```
kyrtapz, line 150: why not use `status.client.ClientFor(serviceItemFromRange.GetClusterName())`

kyrtapz: Also we might need to check if the result of `ClientFor` is not nil, just to be safe

pkg/controller/statusmanager/pod_status.go
```diff
@@ -135,9 +145,70 @@ func (status *StatusManager) SetFromPods() {
...
+		}
+		if err := status.setSSAnnotation(ss, names.RolloutHungAnnotation, ssHung); err != nil {
+			log.Printf("Error setting StatefulSet %q annotation: %v", ssName, err)
+		}
+	}
+
 	for _, depName := range status.deployments {
 		dep := &appsv1.Deployment{}
-		if err := status.client.Get(context.TODO(), depName, dep); err != nil {
+		if err := status.client.Default().CRClient().Get(context.TODO(), depName, dep); err != nil {
```
kyrtapz, line 211: Use `status.client.ClientFor` instead?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -156,7 +227,7 @@ func (status *StatusManager) SetFromPods() {
 			depProgressing = true
 			// Check for any pods in CrashLoopBackOff state and mark the operator as degraded if so.
 			if !isNonCritical(dep) {
-				hung = append(hung, status.CheckCrashLoopBackOffPods(depName, dep.Spec.Selector.MatchLabels, "Deployment")...)
+				hung = append(hung, status.CheckCrashLoopBackOffPods(depName, dep.Spec.Selector.MatchLabels, "Deployment", cnoclient.DefaultClusterName)...)
```
kyrtapz, line 230: To make it more future proof and generic we should use the `clusterName` here and pick the `DefaultClusterName` if `clusterName` is empty

pkg/controller/statusmanager/pod_status.go
```diff
@@ -240,33 +311,34 @@ func (status *StatusManager) SetFromPods() {
...
+func (status *StatusManager) getLastPodState() (map[types.NamespacedName]daemonsetState, map[types.NamespacedName]deploymentState, map[types.NamespacedName]statefulsetState) {
 	// with maps allocated
 	daemonsetStates := map[types.NamespacedName]daemonsetState{}
 	deploymentStates := map[types.NamespacedName]deploymentState{}
+	statefulsetStates := map[types.NamespacedName]statefulsetState{}
 
 	// Load the last-seen snapshot from our annotation
 	co := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
-	err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
+	err := status.client.Default().CRClient().Get(context.TODO(), types.NamespacedName{Name: status.name}, co)
```
kyrtapz, line 325: Use status.client.ClientFor instead?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -299,14 +377,19 @@ func (status *StatusManager) setLastPodState(
...
+
 	lsbytes, err := json.Marshal(ps)
 	if err != nil {
 		return err
 	}
 
 	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		oldStatus := &configv1.ClusterOperator{ObjectMeta: metav1.ObjectMeta{Name: status.name}}
-		err := status.client.Get(context.TODO(), types.NamespacedName{Name: status.name}, oldStatus)
+		err := status.client.Default().CRClient().Get(context.TODO(), types.NamespacedName{Name: status.name}, oldStatus)
```
kyrtapz, line 392: Use status.client.ClientFor instead?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -48,20 +52,77 @@ const (
...
+func New(client cnoclient.Client, name string) *StatusManager {
+	return &StatusManager{
+		client:           client,
+		name:             name,
+		hypershiftConfig: network.NewHypershiftConfig(),
+	}
+}
+
+// setCOAnnotation sets an annotation on the clusterOperator network object
+func (status *StatusManager) setCOAnnotation(obj *configv1.ClusterOperator) error {
```
kyrtapz, line 78: Can we rename this to `setClusterOperAnnotation` or similar? CO is a little confusing for me

pkg/controller/statusmanager/pod_status.go
zshi-redhat: @kyrtapz I just realized this `status.statefulSets` is `types.NamespacedName` which doesn't have GetClusterName() interface.

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: This is intended, it puts the management cluster objects in a new relatedClusterObjects struct and track them separately.
management cluster objects will be updated in clusterOperator annotation, guest related objects are updated in clusterOperator status.

pkg/controller/statusmanager/pod_status.go
kyrtapz: With the change as is we make an implicit assumption that statefulsets can only be deployed on the management cluster.
I am not sure if we should do that, since we are in control of `statefulSets` and the rest for watching objects maybe we could extend them to support multiple clusters? This could be another commit

pkg/names/names.go
```diff
@@ -64,6 +64,10 @@ const KuryrOctaviaVersionAnnotation = "networkoperator.openshift.io/kuryr-octavi
 // value format: cluster/namespace/name
 const CopyFromAnnotation = "network.operator.openshift.io/copy-from"
 
+// CopyFromAnnotation is an annotation that allows copying resources from specified clusters
```
kyrtapz, line 67: Fix docblock

pkg/network/hypershift.go
```diff
@@ -0,0 +1,50 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+var (
+	enabled   = os.Getenv("HYPERSHIFT")
+	name      = os.Getenv("HOSTED_CLUSTER_NAME")
+	namespace = os.Getenv("HOSTED_CLUSTER_NAMESPACE")
+
+	// NetworkOperatorStatusTypeProgressing indicates Degraded condition in hostedControlPlane status
```
kyrtapz, line 15: The two following variables can be const

pkg/controller/statusmanager/pod_status.go
zshi-redhat: @kyrtapz I agree that `serviceItemFromRange.GetClusterName()` is better than making assumption on resource location. I will make another PR to fix it.

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Done

pkg/controller/statusmanager/pod_status.go
zshi-redhat: will fix it when we use obj that supports clusterName.

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Done

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Done

pkg/controller/statusmanager/status_manager.go
zshi-redhat: Done

pkg/names/names.go
zshi-redhat: Fixed

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Fixed.

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Done

pkg/controller/statusmanager/pod_status.go
zshi-redhat: Done

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -271,21 +275,38 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+	// Set up the Pod reconciler before we start creating DaemonSets/Deployments/StatefulSets
+	daemonSets := []statusmanager.ClusteredName{}
+	deployments := []statusmanager.ClusteredName{}
+	statefulSets := []statusmanager.ClusteredName{}
 	relatedObjects := []configv1.ObjectReference{}
+	relatedClusterObjects := []network.RelatedObject{}
 	for _, obj := range objs {
 		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
-			daemonSets = append(daemonSets, types.NamespacedName{Namespace: obj.GetNamespace(), Name: obj.GetName()})
+			daemonSets = append(daemonSets, statusmanager.ClusteredName{ClusterName: obj.GetClusterName(), Namespace: obj.GetNamespace(), Name: obj.GetName()})
```
squeed, line 286: Will the cluster name be `""` or `"default"` for most objects? And does the rest of the code reflect that?

squeed: aha, ClusterClient treats `""` as default.

pkg/controller/operconfig/operconfig_controller.go
squeed: But isn't that true for all non-default-cluster objects? Why would the management cluster be special?

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: I think It will be the default value `""` for most objects. and `""` is considered as `default`. The rest of code use `ClientFor(cluster name)` to access the right client.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -203,12 +320,80 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 		log.Printf("Failed to set operator status: %v", err)
 	}
 
+	if status.hyperShiftConfig.Enabled {
```
squeed, line 323: Can you add a comment indicating what this is doing? (Namely, mirroring in to the HostedControlPlane type). And consider breaking it out in to its own function -- `WriteHypershiftStatus()` or something like that.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -249,13 +441,13 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 		}
 
 		if isNotFound {
-			if err := status.client.Create(context.TODO(), co); err != nil {
+			if err := status.client.ClientFor(cnoclient.DefaultClusterName).CRClient().Create(context.TODO(), co); err != nil {
```
squeed, line 444: You can just write `status.client.Default()` in this case.

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -396,26 +356,26 @@ func TestStatusManagerSetFromDaemonSets(t *testing.T) {
 
 	// Create minimal DaemonSets
 	dsA := &appsv1.DaemonSet{
-		ObjectMeta: metav1.ObjectMeta{Namespace: "one", Name: "alpha", Generation: 1},
+		ObjectMeta: metav1.ObjectMeta{Namespace: "one", Name: "alpha", Generation: 1, ClusterName: "default"},
```
squeed, line 359: We have a bit of a problem: we consider cluster names "" and "default" to be identical. It would be good if we normalized everything, probably to "", for the purposes of status reporting.

squeed:
A few comments, but otherwise looks good.
/approve

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: @squeed I remembered we have this summary for RelatedObjects that: we should not be adding management cluster objects to RelatedObjects. We might need to add them as an annotation on our ClusterOperator object for internal reasons (cleanup). This is why the management objects are tracked in a different struct in status manager and be updated later in the network.operator annotation.

pkg/controller/statusmanager/status_manager_test.go
zshi-redhat: Will update to use `""`.

cc @kyrtapz

pkg/controller/statusmanager/status_manager.go
zshi-redhat: will update

pkg/controller/statusmanager/status_manager.go
zshi-redhat: will update to use `""`

pkg/controller/operconfig/operconfig_controller.go
squeed: Right, what I'm saying is that this conditional should probably be `obj.GetClusterName() != ""`

pkg/controller/statusmanager/status_manager_test.go
zshi-redhat: Done.

pkg/controller/statusmanager/status_manager.go
zshi-redhat: Updated

pkg/controller/statusmanager/status_manager.go
zshi-redhat: Updated

pkg/controller/operconfig/operconfig_controller.go
zshi-redhat: Done

zshi-redhat:
> @zshi-redhat: PR needs rebase.

rebased

PR #1347, opened 2022-03-24
Add tuning cni sysctl allowlist to nodes

mmirecki:
The tuning cni currently allows to set any sysctl in the system, instead of just the sysctls for the specific interface on which it is run. In fact, until now it was not even possible to set the interface level sysctl, only global sysctls.
This poses security risks. It also makes the feature of limited pod level sysctls ([here](https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls)) void, as it can be easily circumvented with an appropriate tuning-cni nad.

For this reason two new features were added to the tuning-cni:
- interface level sysctls - allowing to set a sysctl for the interface on which the cni is chained
- sysctl whitelist - allowing to limit the sysctls that can be set by the tuning-cni.

The sysctl whitelist requires a file (/etc/cni/tuning/allowlist.conf) to be placed on each node.
This PR creates such a file with a predefined list of safe sysctls (RFE-2038). The list of the sysctl is a minimal list, additional sysctl's will be added when approved.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: cni-binary-copy-script
+  name: cni-copy-resources
   namespace: openshift-multus
   annotations:
     kubernetes.io/description: |
@@ -84,6 +84,16 @@ data:
       echo "Failed to copy files in ${sourcedir} to $DESTINATION_DIRECTORY"
       exit 1
     fi
+  allowlist.conf: |-
+    ^net.ipv4.conf.IFNAME.accept_redirects$
+    ^net.ipv4.conf.IFNAME.accept_source_route$
+    ^net.ipv4.conf.IFNAME.disable_policy$
+    ^net.ipv4.conf.IFNAME.secure_redirects$
+    ^net.ipv4.conf.IFNAME.send_redirects$
+    ^net.ipv6.conf.IFNAME.accept_redirects$
+    ^net.ipv6.conf.IFNAME.accept_source_route$
+    ^net.ipv6.neigh.IFNAME.base_reachable_time_ms$
+    ^net.ipv6.neigh.IFNAME.retrans_time_ms$
 ---
 kind: DaemonSet
 apiVersion: apps/v1
@@ -187,7 +197,7 @@ spec:
             type: File
         - name: cni-binary-copy
           configMap:
-            name: cni-binary-copy-script
+            name: cni-copy-resources
             defaultMode: 0744
 ---
 kind: DaemonSet
@@ -245,7 +255,7 @@ spec:
           value: "/usr/src/egress-router-cni/bin/"
       - name: cni-plugins
         image: {{.CNIPluginsImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: ["/bin/bash", "-c", "/entrypoint/cnibincopy.sh && cp /entrypoint/allowlist.conf /host/etc/cni/tuning/"]
         volumeMounts:
         - mountPath: /entrypoint
           name: cni-binary-copy
@@ -254,6 +264,9 @@ spec:
         - mountPath: /host/etc/os-release
           name: os-release
           readOnly: true
+        - mountPath: /host/etc/cni/tuning/
+          name: tuning-conf-dir
+          readOnly: false
         env:
         - name: RHEL7_SOURCE_DIRECTORY
           value: "/usr/src/plugins/rhel7/bin/"
@@ -471,8 +484,12 @@ spec:
             type: File
         - name: cni-binary-copy
           configMap:
-            name: cni-binary-copy-script
+            name: cni-copy-resources
             defaultMode: 0744
+        - hostPath:
+            path: /etc/cni/tuning/
+            type: DirectoryOrCreate
+          name: tuning-conf-dir
 ---
 apiVersion: batch/v1
 kind: CronJob


bindata/network/multus/multus.yaml
```diff
@@ -84,6 +84,24 @@ data:
       echo "Failed to copy files in ${sourcedir} to $DESTINATION_DIRECTORY"
       exit 1
     fi
+  allowlistcopy.sh: |-
```
yuvalk, line 87: we better collect this file in must gather

yuvalk:
what about test coverage?

bindata/network/multus/multus.yaml
```diff
@@ -84,6 +84,24 @@ data:
...
+
+    if [ $? -eq 0 ]; then
+      echo "Successfully copied tuning-cni config $ALLOWLIST_CONF to $DESTINATION_DIRECTORY"
+    else
+      echo "Failed to copy tuning-cni config $ALLOWLIST_CONFto $DESTINATION_DIRECTORY"
+      exit 1
+    fi
+  allowlist.conf: |-
+    ^net\.ipv4\.conf\.IFNAME\.[a-z_]*$
+    ^net\.ipv6\.conf\.IFNAME\.[a-z_]*$
```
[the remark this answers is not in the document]
mmirecki: Replaced by an explicit sysctl list as specified in RFE-2038

bindata/network/multus/multus.yaml
mmirecki: Will add a PR to must gather for this one

mmirecki:
> what about test coverage?
Tests are added in the origin repo [(PR26809)](https://github.com/openshift/origin/pull/26809) + more to come

yuvalk:
> > what about test coverage?
> > Tests for both pod and interface level sysctl's are added in the origin repo ([PR26809](https://github.com/openshift/origin/pull/26809) merged + more to come)

Was thinking more about CNO tests, for this list

mmirecki:
> > > what about test coverage?
> > > Tests for both pod and interface level sysctl's are added in the origin repo ([PR26809](https://github.com/openshift/origin/pull/26809) merged + more to come)
> 
> Was thinking more about CNO tests, for this list

Unit tests for this are not really doable. [origin PR26953](https://github.com/openshift/origin/pull/26953) does a e2e tests for this and the new tuning-cni features (interface level sysctls and allowlist)

bindata/network/multus/multus.yaml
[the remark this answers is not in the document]
yuvalk: if allowlist.conf allow for comments in it, 
let's add that comment in there, as instructions of how to add more sysclts

taken from kubernetes code:
```
// SafeSysctlWhitelist returns the whitelist of safe sysctls and safe sysctl patterns (ending in *).
//
// A sysctl is called safe iff
// - it is namespaced in the container or the pod
// - it is isolated, i.e. has no influence on any other pod on the same node.
```

bindata/network/multus/multus.yaml
```diff
@@ -84,6 +84,38 @@ data:
...
+    ^net.ipv4.conf.IFNAME.log_martians$
+    ^net.ipv4.conf.IFNAME.rp_filter$
+    ^net.ipv6.conf.IFNAME.forwarding$
+    ^net.ipv4.conf.IFNAME.arp_notify$
+    ^net.ipv4.conf.IFNAME.arp_accept$
+    ^net.ipv4.neigh.IFNAME.base_reachable_time_ms$
+    ^net.ipv6.neigh.IFNAME.base_reachable_time_ms$
+    ^net.ipv6.neigh.IFNAME.retrans_time_ms$
+    ^net.ipv4.conf.IFNAME.disable_policy$
+    ^net.ipv6.conf.IFNAME.disable_policy$
```
maiqueb, line 118: would it make sense to order this somehow ? at least have all the ipv4 knobs listed together, then the ipv6 ones.

mmirecki: np, done

yuvalk: I actually like it better that each sysctl pair is togther
maybe even move to regex `"net.ipv[4|6].conf.IFNMAE..."`
to show that we are adding both variants

maiqueb: But what about the odd one out - i.e. the ones without matches ? Like `send_redirects` , `log_martians`, `rp_filter`, etc ?...

yuvalk: externally to this PR, I think we should complete the work on their safety and add them
but meanwhile,
with my ordering suggestion it will be much easier to call them out (ie- see they are different)

maiqueb: ... and listing them by IP family allows the reader to easily understand what knobs are available for each family - as I see it, it nearly becomes documentation. 

I won't fight for this, it can happen either way.

mmirecki: .. meanwhile, until all interested parties reach an agreement on the order, I'm leaving the sorting as is :)
The list was further reduced. More sysctls will be added when certified as safe.

maiqueb (approved):
Thanks. 

Just want to point out that I'm OK either way with the sorting of the knobs on the allowlist.

bindata/network/multus/multus.yaml
```diff
@@ -84,6 +84,31 @@ data:
...
+
+    ALLOWLIST_CONF=/entrypoint/allowlist.conf
+    DESTINATION_DIRECTORY=/host/etc/cni/tuning/
+
+    cp -f $ALLOWLIST_CONF $DESTINATION_DIRECTORY
+
+    if [ $? -eq 0 ]; then
+      echo "Successfully copied tuning-cni config $ALLOWLIST_CONF to $DESTINATION_DIRECTORY"
+    else
+      echo "Failed to copy tuning-cni config $ALLOWLIST_CONFto $DESTINATION_DIRECTORY"
```
danwinship, line 99:
```suggestion
      echo "Failed to copy tuning-cni config $ALLOWLIST_CONF to $DESTINATION_DIRECTORY"
```

bindata/network/multus/multus.yaml
```diff
@@ -84,6 +84,31 @@ data:
...
+      exit 1
+    fi
+  allowlist.conf: |-
```
danwinship, line 102: should update the name of the configmap to not be `cni-binary-copy-script` since it's no longer just that

bindata/network/multus/multus.yaml
```diff
@@ -245,7 +270,7 @@ spec:
           value: "/usr/src/egress-router-cni/bin/"
       - name: cni-plugins
         image: {{.CNIPluginsImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: ["/bin/bash", "-c", "/entrypoint/cnibincopy.sh && /entrypoint/allowlistcopy.sh"]
```
danwinship, line 273: is `allowlistcopy.sh` really that useful? Couldn't you just `cp /entrypoint/allowlist.conf /host/etc/cni/tuning/`?

bindata/network/multus/multus.yaml
mmirecki: Updated to: "cni-copy-resources"

bindata/network/multus/multus.yaml
```diff
@@ -245,7 +255,7 @@ spec:
           value: "/usr/src/egress-router-cni/bin/"
       - name: cni-plugins
         image: {{.CNIPluginsImage}}
-        command: ["/entrypoint/cnibincopy.sh"]
+        command: ["/bin/bash", "-c", "/entrypoint/cnibincopy.sh && cp /entrypoint/allowlist.conf /host/etc/cni/tuning/"]
```
danwinship, line 258: maybe do this as a multi-line script? (eg like the multus container)

danwinship (approved):
one suggestion but it's not a requirement

bindata/network/multus/multus.yaml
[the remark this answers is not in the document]
yuvalk: @mmirecki 
can we add comments to the `allowlist.conf`?

PR #1350, opened 2022-03-25
Bug 2063123: Drop Node update permission for sdn-node

danwinship:
https://github.com/openshift/sdn/pull/417 makes openshift-sdn-node no longer make use of its Node modifying permissions, so this drops it from its service account.

/hold


danwinship:
/hold cancel
@jcaamano or @msherif1234 want to lgtm this too? (It removes the permissions we no longer need post-https://github.com/openshift/sdn/pull/417.)

PR #1359, opened 2022-03-30
Bug 2070047: Bump max value of hist quantile for kuryr_cni_request_duration

MaysaMacedo:
With Kuryr the CNI requests can take a considerable
time given that it has to wait for a VIF from Neutron.
We've seen warnings being raised for KuryrCNISlow, so let's
increase it's max acceptable value to 60 which is the value
hit in most of the observations.


dulek:
/test e2e-openstack-kuryr

I would say this should include a release note BTW, but that's to be done on the bug.

PR #1361, opened 2022-03-31
Run ovnkube-master statefulset pods in parallel

zshi-redhat:


zshi-redhat:
@squeed @kyrtapz

PR #1363, opened 2022-04-01
Remove Kuryr mutating DNS webhook

dulek:
This mutating webhook was a hack to allow running OpenShift with Kuryr
on older OpenStack versions that were not supporting specifying multiple
listeners for an Octavia LB. This meant that the DNS LB was only
listening for TCP requests and all of the pods got to have the DNS
options in the spec adjusted for that. The webhook injected them.

Obviously a terrible thing, but so is running old OpenStack. Anyway
4.10 (or even 4.9?) dropped support for OSP13, meaning that all the
Octavia versions we'll run against will support multiple listeners in an
LB and this dangerous hack can be rightfully removed.


dulek:
/hold

I'd like to wait for a confirmation from @MaysaMacedo that this is sane.

dulek:
/hold cancel

CI is fine now.

squeed:
clearly unrelated...

PR #1368, opened 2022-04-05
OVN: remove detecing db_ip via kapi

trozet:
This was some legacy code that was used to pass the db address when
starting ovnkube binary. It's no longer needed.


tssurya:
@trozet : lgtm, but just for my understanding why did we require it before I wonder?

squeed:
We needed it just to enforce startup ordering -- some components were more touchy about reconnecting.

/lgtm
/override ci/prow/e2e-gcp

squeed:
(Also, there was a lot of cargo-culting in the old ovn scripts, not all of which was exorcized)

PR #1370, opened 2022-04-06
Use the hosted cluster token explicitly

kyrtapz:
Instead of minting the token to the default service account mount use an alternative path and pass it as an argument to ovnkube-master and ovn-dbchecker.
This allows us to use the default service account token to talk to the managment cluster, it is required for kube-rbac-proxy.


squeed:
@zshi-redhat do we still need this, given that we're getting rid of kube-rbac-proxy in #1374?

zshi-redhat:
> @zshi-redhat do we still need this, given that we're getting rid of kube-rbac-proxy in #1374?

@squeed It is not strictly needed given kube-rbac-proxy is not used.
I talked with @kyrtapz on this, we believe it may still be a clean way to mount the generated token in a different path. wdyt?

PR #1372, opened 2022-04-06
hypershift: enable publishNotReadyAddress explicitly for ovnk-master service

zshi-redhat:


squeed:
@zshi-redhat If we do this, then I'm pretty sure we need a second Service, with `PublishNotReadyAddresses: false`, to use for the Route. Otherwise we will have connection failures on upgrade.

squeed:
Summarizing an in-person discussion:
1. we need two services, an "internal" service with `publishNotReadyAddresses`, and an "external" one to use for the Route
2. the "internal" service can be used for metrics / ServiceMonitor too
3. the "internal" service can be headless.

zshi-redhat:
Rebased due to merge conflict.

PR #1374, opened 2022-04-07
hypershift: enable ovnk-master metrics in management cluster

zshi-redhat:
Kube-rbac-proxy is not used since it requires ClusterRoleBinding
to tokenreviews and subjectaccessreviews and we want to avoid
creating ClusterRoleBinding for individual control planes where
possible.

By getting rid of kube-rbac-proxy, we expose all metrics to the
entire management cluster. This is OK since workloads there are
trusted.

Depends on tls support for metrics in ovnk:
https://github.com/ovn-org/ovn-kubernetes/pull/2900


zshi-redhat:
Ovnk tls support for metrics is being added here: https://github.com/ovn-org/ovn-kubernetes/pull/2900
Updated this PR to support tls and depend on the ovnk change.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
```diff
@@ -772,7 +725,7 @@ spec:
             --config-file=/run/ovnkube-config/ovnkube.conf \
             --ovn-empty-lb-events \
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
-            --metrics-bind-address "127.0.0.1:29102" \
+            --metrics-bind-address "${K8S_POD_IP}:9102" \
```
squeed, line 728: why not just `0.0.0.0`?

squeed:
By getting rid of kube-rbac-proxy, we expose all metrics to the entire management cluster. I think this is OK, since workloads there are trusted, but it's important to call out.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
zshi-redhat: Fixed

bindata/network/ovn-kubernetes/managed/monitor.yaml
```diff
@@ -0,0 +1,160 @@
...
+      serverName: ovn-kubernetes-node.openshift-ovn-kubernetes.svc
+  jobLabel: app
+  namespaceSelector:
+    matchNames:
+    - openshift-ovn-kubernetes
+  selector:
+    matchLabels:
+      app: ovnkube-node
+---
+apiVersion: v1
```
squeed, line 105: Does it make sense to move the ovn-kube node monitoring bits in to `/shared`? Or do we expect them to be different?

zshi-redhat: sure, the ovnkube-node monitoring manifest are shared, not difference.
Moved the ovnkube-node service and serviceMonitor to common folder.

PR #1377, opened 2022-04-11
update the alert exp to reflect the duration directly

bmeng:


squeed:
@bmeng what's the reasoning behind this? These alerts are not the same.

The alert, as written, fires if the control plane is 2 minutes behind for 10 minutes. It doesn't just catch a hung pipeline, but rather a **slow pipeline** as well. That's why I wrote the alert like that -- make sense?

squeed:
@bmeng you're right -- the description should be updated instead. Something like "the southbound database has been at least 120 seconds stale for too long"

bmeng:
cc @martinkennelly @bpickard22 @squeed PTAL

martinkennelly:
@bmeng  can you rebase?

bmeng:
@martinkennelly Thanks, this is updated.
I also put the runbook url to the same alerts in the self-hosted files.

martinkennelly:
@bmeng Apologies for nit picking but can you put the updated URLs in a new commit?

PR #1378, opened 2022-04-11
Bug 2072766: Reserve port TCP/9104 for cluster-network-operator

andreaskaris:
Cluster network operator is a host networked pod. Reserve port
TCP/9104 to make the kubernetes scheduler aware of that port
requirement.

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
@@ -89,6 +89,11 @@ spec:
               fieldPath: metadata.name
         image: quay.io/openshift/origin-cluster-network-operator:latest
         name: network-operator
+        ports:
+        - containerPort: 9104
+          hostPort: 9104
+          name: cno
+          protocol: TCP
         resources:
           requests:
             cpu: 10m
--- a/manifests/0000_70_cluster-network-operator_03_deployment.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment.yaml
@@ -21,6 +21,11 @@ spec:
     spec:
       containers:
       - name: network-operator
+        ports:
+          - containerPort: 9104
+            hostPort: 9104
+            name: cno
+            protocol: TCP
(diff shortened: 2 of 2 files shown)


jcaamano:
/lgtm

@zshi-redhat how does this work for hypershift?

jcaamano:
/approve cancel
Did not have the intention to approve just yet

zshi-redhat:
> /lgtm
> 
> @zshi-redhat how does this work for hypershift?

CNO is deployed by control-plane-operator in hypershift: https://github.com/openshift/hypershift/pull/1101/files#diff-06207575f3cfe4069b485056885b8807a7ad0f29e07948aa573434531aa9665cR210.
It runs as regular CNI pod instead of hostnetworked pod, so we cannot assign host port to CNO pod.
What will this reserved port 9104 be used for?

jcaamano:
> > > > /lgtm
> > > > @zshi-redhat how does this work for hypershift?
> > > 
> > > 
> > > CNO is deployed by control-plane-operator in hypershift: https://github.com/openshift/hypershift/pull/1101/files#diff-06207575f3cfe4069b485056885b8807a7ad0f29e07948aa573434531aa9665cR210. It runs as regular CNI pod instead of hostnetworked pod, so we cannot assign host port to CNO pod in hypershift deployment(there could be multiple CNO instances running on the same node). What will this reserved port 9104 be used for?
> > 
> > 
> > It's a port that the openshift controller library uses, I don't know what for. Thanks for clarifying. /approve
> 
> To clarify, just wanted to make sure these manifests were not being used for hypershift.

@zshi-redhat passed as parameter to the cno binary but not in the case of hypershift that I could see in the code you shared:
https://github.com/openshift/cluster-network-operator/blob/1f46ed85e9381d428cd445320177f77364540088/manifests/0000_70_cluster-network-operator_03_deployment.yaml#L36

jcaamano:
> What will this reserved port 9104 be used for?

I think that it might be used for health checks coming from the kube apiserver aggregation layer, at least.

andreaskaris:
@jcaamano Can you give another lgtm here? Thanks!

PR #1360, opened 2022-03-30
Bug 1983056: Kuryr: Update CRD from upstream

dulek:
This is based on [1] and required by openshift/kuryr-kubernetes#645.

[1] https://opendev.org/openstack/kuryr-kubernetes/commit/d5f5db7005028f9ba88ed9dcc0b5d09d0bbd8bda

PR #1380, opened 2022-04-12
Bug 1983056: Include info about static Pod to CRD

MaysaMacedo:
Kuryr is adding support to use the K8S_POD_UID passed
from CRI and it needs to skip its usage when the Pod is static
due to a know issue[1]. This commit adds a new field to
the KuryrPort CRD to easily identify if the Pod is static.

[1] https://github.com/k8snetworkplumbingwg/multus-cni/issues/773


MaysaMacedo:
there is already one https://github.com/openshift/cluster-network-operator/pull/1360

PR #1256, opened 2021-12-15
Bug 2022144: sbdb and nbdb containers leave pid around if they restarted or crashed

msherif1234:
This PR remove pidfile on postStart hook to avoid run_n/sb_ovsdb from an early return.

Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>

PR #1383, opened 2022-04-15
Bug 2072011: enable mem triming just once after raft election

msherif1234:
avoid extensive logs in sbdb and nbdb containers with enable mem trim logs

### DIFF
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
@@ -141,6 +141,7 @@ spec:
             /usr/share/ovn/scripts/ovn-ctl stop_nb_ovsdb
             echo "$(date -Iseconds) - nbdb stopped"
             rm -f /var/run/ovn/ovnnb_db.pid
+            rm -f /var/run/ovn/ovnnbdb-mem-trim-enabled
             exit 0
           }
           # end of quit
@@ -382,7 +383,15 @@ spec:
                 echo "NB DB Raft leader is unknown to the cluster node."
                 exit 1
               else
-                /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+                # enable memory triming once
+                if [ ! -f /var/run/ovn/ovnnbdb-mem-trim-enabled ]; then
+                  /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on
+                  if [[ $? -eq 0 ]]; then
+                    touch /var/run/ovn/ovnnbdb-mem-trim-enabled
+                  else
+                    echo "NB DB failed to enable memory trim on compaction"
+                  fi
+                fi
               fi
 
         env:
@@ -491,6 +500,7 @@ spec:
             /usr/share/ovn/scripts/ovn-ctl stop_sb_ovsdb
             echo "$(date -Iseconds) - sbdb stopped"
             rm -f /var/run/ovn/ovnsb_db.pid
+            rm -f /var/run/ovn/ovnsbdb-mem-trim-enabled
             exit 0
           }
           # end of quit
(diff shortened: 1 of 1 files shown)


dcbw:
You know what... I thought there was a way to read back memory trimming, but there's not... so nevermind. Let's just leave the logspam.

msherif1234:
> You know what... I thought there was a way to read back memory trimming, but there's not... so nevermind. Let's just leave the logspam.

this hack should do at least for now till we have a way to read and confirm, its not good to ship with such logs ?

bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
```diff
@@ -382,7 +383,11 @@ spec:
                 echo "NB DB Raft leader is unknown to the cluster node."
                 exit 1
               else
-                /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+                # enable memory triming once
+                if [ ! -f /var/run/ovn/ovnnb_db_mem_trim.txt ]; then
+                  /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+                  echo "done" > /var/run/ovn/ovnnb_db_mem_trim.txt
```
rbbratta, line 389: I think the convention is to use `touch something-executed`

e.g.
https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml#L9

```shell    
    # This file is not needed anymore in 4.7+, but when rolling back to 4.6
    # the ovs pod needs it to know ovs is running on the host.
    touch /var/run/ovs-config-executed
```

bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
```diff
@@ -382,7 +383,11 @@ spec:
                 echo "NB DB Raft leader is unknown to the cluster node."
                 exit 1
               else
-                /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+                # enable memory triming once
+                if [ ! -f /var/run/ovn/ovnnb_db_mem_trim.txt ]; then
+                  /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+                  touch /var/run/ovn/ovnnbdb-mem-trim-enabled
```
rbbratta, line 389: Should also be a ` &&`
```shell
/usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null && touch /var/run/ovn/ovnnbdb-mem-trim-enabled
```
  of `if/then` so that we only touch on success.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
```diff
@@ -366,6 +366,7 @@ spec:
                 /usr/share/ovn/scripts/ovn-ctl stop_nb_ovsdb
                 echo "$(date -Iseconds) - nbdb stopped"
                 rm -f /var/run/ovn/ovnnb_db.pid
+                rm -f /var/run/ovn/ovnnbdb-mem-trim-enabled
```
rbbratta, line 369: This won't cleanup on crash (e.g. `SIGSEGV`), we need to place the enabled file inside to container filesystem so it is removed when the container dies for any reason.

Placing `ovnnb_db.pid` on the host filesystem is a bad special case that we had a different bug for  https://github.com/openshift/cluster-network-operator/pull/1256  

In general we should not be keeping runtime container state on the persistent host file system, or if we really have to, then we have to `rm -f /var/run/ovn/ovnnbdb-mem-trim-enabled` in `postStart`.

msherif1234: better place for cleanup is the trap TERM handler

rbbratta: TERM won't catch the uncatchable signals.

rbbratta: err well I guess only `SIGKILL` and `SIGSTOP` can't be caught.  QE tests with `SIGKILL` now, so after that test memory compaction won't be enabled.

`SIGSEGV` is the main signal we expect and we don't trap it currently, the systemd core dump service catches that.

But again, why are we defeating the container lifecycle by putting a file outside the container?  Stop fighting the container.

rbbratta: even the `ovnnb_db.pid` is invalid outside the container since we have a PID namespace.

msherif1234: SIGKILL can't be trapped so I think if u are testing with SIGKILL we won't be able to even stop any of the dbs u will probably see another class of issues right ?

PR #1387, opened 2022-04-20
Plug proxying through the networking stack for Hypershift

alvaroaleman:
Currently, the deployed networking components never have a proxy
configured because it is never needed as it is always possible to reach
the KAS and OVN SBDB directly.

With Hypershift, the above assumption might be broken. This PR plugs
through proxy settings to the MTU prober, SDN, multus and into ovn-k.
In ovn-k it doesn't currently use these settings, because contrary to
all other components it is not enough to just set HTTP{,S}_PROXY and
NO_PROXY env vars as ovn-k doesn't use http but a custom protocol over
TLS and thus doesn't respect these vars.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"log"
 	"reflect"
-	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"strings"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 
 	"github.com/pkg/errors"
 
--- a/bindata/cloud-network-config-controller/controller.yaml
+++ b/bindata/cloud-network-config-controller/controller.yaml
@@ -51,6 +51,18 @@ spec:
           value: "{{.KubernetesServicePort}}"
         - name: KUBERNETES_SERVICE_HOST
           value: "{{.KubernetesServiceHost}}"
+{{ if .HTTP_PROXY }}
+        - name: "HTTP_PROXY"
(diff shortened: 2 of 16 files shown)


pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -267,8 +268,19 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	// If we use hypershift, the components need proxy settings. In selfhosted, they
+	// are always local to the API so there it is never needed.
+	var proxyConfig configv1.ProxyStatus
+	if hyperShiftConfig := network.NewHyperShiftConfig(); hyperShiftConfig.Enabled {
```
alvaroaleman, line 274: Technically this shouldn't be needed btw, as in selfhosted the KAS internal address is in `NO_PROXY`

squeed: We tend to gather required settings via the Bootstrap mechanism -- can you move this to pkg/platform/platform.go? (And update the InfraStatus type accordingly?)

squeed:
@alvaroaleman one small change I'd like to see (that will mean a lot less plumbing / method signature changes), then this looks good to me.

pkg/controller/operconfig/operconfig_controller.go
alvaroaleman: @squeed updated, PTAL

zshi-redhat:
@alvaroaleman  I included your PR in https://github.com/openshift/cluster-network-operator/pull/1408 which has merged.
Shall I close this one?

alvaroaleman:
yeah, not needed anymore

PR #1390, opened 2022-04-22
Bug 2075475: Add default-route field to egress-router k8s.v1.cni.cncf.io/networks

andreaskaris:
Add default-route field to egress-router additional networks def so
that multus/ovn-k inject the correct routes into the pod towards
the cluster. Currently, the egress-router-cni relies on the gateway
field of the CNI definition to delete the default route and inject
its own, so in the future we will have to fix egress-router-cni, too.
For the moment, both the correct annotation and egress-router-cni's
incorrect behavior can coexist.

### DIFF
--- a/bindata/egress-router/001-deployment.yaml
+++ b/bindata/egress-router/001-deployment.yaml
@@ -19,7 +19,13 @@ spec:
         app: egress-router-cni
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        k8s.v1.cni.cncf.io/networks: egress-router-cni-nad
+        k8s.v1.cni.cncf.io/networks: |
+          [
+            {
+              "name":"egress-router-cni-nad",
+              "default-route": ["{{.Gateway}}"]
+            }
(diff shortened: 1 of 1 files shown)


andreaskaris:
Verification:
https://bugzilla.redhat.com/show_bug.cgi?id=2075475#c19
and
https://bugzilla.redhat.com/show_bug.cgi?id=2075475#c20

andreaskaris:
@danwinship Could you have a look at this given that you're familiar with it?

danwinship:
/lgtm
/hold
As I said in the Slack conversation, I think this is the right plan, though I'm not super multus-knowledgable-y and couldn't actually say if this is the right syntax, etc. If you haven't already can you bring up a test cluster with this build and make sure egress router still works? Then feel free to `/hold cancel`

andreaskaris:
Yeah I tested on both OVN and SDN and it seems to be doing the right thing. So unless I missed something, it should be o.k.; I'll make sure to work with the CI team as soon as this merged to have this tested thoroughly

PR #1393, opened 2022-04-25
ovnkube: export OVS metrics along with OVN metrics

dcbw:
Lets us finally get host OVS metrics with ovnkube.

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -398,6 +398,7 @@ spec:
             --metrics-bind-address "127.0.0.1:29103" \
             --ovn-metrics-bind-address "127.0.0.1:29105" \
             --metrics-enable-pprof \
+            --export-ovs-metrics \
             --disable-snat-multiple-gws \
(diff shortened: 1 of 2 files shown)


dcbw:
/retest
migration-rollback issue is a flake
```
{Operator degraded (InvalidOperatorConfig): Not applying unsafe configuration change: invalid configuration: [cannot change default network type when not doing migration]. Use 'oc edit network.operator.openshift.io cluster' to undo the change.  Operator degraded (InvalidOperatorConfig): Not applying unsafe configuration change: invalid configuration: [cannot change default network type when not doing migration]. Use 'oc edit network.operator.openshift.io cluster' to undo the change.}
```

flavio-fernandes (commented):
/lgtm

ref: https://github.com/ovn-org/ovn-kubernetes/pull/2924

npinaeva:
@dcbw we should be careful about metrics we expose by default - need at least check naming conventions for all of them and verify the values too. So since I think we agreed that this is now mostly needed for perf/scale team I'd prefer to disable this by default, and let scale team enable it when needed. 
And we can do proper review and enabling by default for ovs metrics later, as a part of our debuggability effort. wdyt?

PR #1394, opened 2022-04-25
ovn-kube hypershift: fix pipefailure that prevents HA startup

squeed:
The probe command fails with a non-zero exit code, causing the startup command to prematurely exit. Oops.

/cc @zshi-redhat


zshi-redhat:
lgtm

Wondering why it didn't happen for non-hypershift deployment or do we need to change this for self-hosted ovnk-master

squeed:
/override ci/prow/e2e-network-migration-rollback
this PR doesn't affect this code at all

PR #1395, opened 2022-04-26
Bug 2071859: Switch dnsPolicy to Default for OVN hostNetwork pods

creydr:
**- What this PR does and why is it needed**

This commit changes the spec.dnsPolicy to Default for OVN hostNetwork pods
so they use the hosts DNS. Even this is currently already the case for
pods having hostNetwork:true and spec.dnsPolicy DNSClusterFirst (which
is the default) this makes it more clear (https://github.com/kubernetes/kubernetes/blob/0424c7c74d926b4fe3193059e003e9056b429d28/pkg/kubelet/network/dns/dns.go#L303-L322)

**- How to verify it**
1. Deploy a cluster with this patch
2. Check that all of the ovn pods which use host network have the dnsPolicy `Default`:
    ```
    $ oc -n openshift-ovn-kubernetes get po -ojsonpath='{.items[?(@.spec.hostNetwork)].spec.dnsPolicy}' 
    Default Default Default Default Default Default Default Default Default%
    ```

**- Description for the changelog**
none

### DIFF
--- a/bindata/network/ovn-kubernetes/common/error-cni.yaml
+++ b/bindata/network/ovn-kubernetes/common/error-cni.yaml
@@ -97,6 +97,7 @@ spec:
         kubernetes.io/os: "linux"
     spec:
       hostNetwork: true
+      dnsPolicy: Default
       nodeSelector:
         kubernetes.io/os: linux
         network.operator.openshift.io/dpu: ''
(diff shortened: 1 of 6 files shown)


creydr:
/retest

@tssurya could you PTAL at this PR? It comes from [BZ 2071859](https://bugzilla.redhat.com/show_bug.cgi?id=2071859)
/assign @tssurya

tssurya:
lgtm

tssurya (approved):

PR #1399, opened 2022-04-27
Operconfig: (almost) always request that we resync

squeed:
We have been returning an empty Result{} for most failures, which means that we won't ever try and re-sync the network unless the configuration changes. However, this is wrong: most failures are transient and not actually related to the operconfig. This means that we might not (ever) re-render if we hit a transient failure.


squeed:
This PR is wrong. Hang on.

PR #1400, opened 2022-04-27
Operconfig: (almost) always request that we resync

squeed:
We have been returning an empty Result{} and err for most failures, which means
    that we won't ever try and re-sync the network unless the configuration
    changes. However, this is wrong: most failures are transient and not
    actually related to the operconfig. This means that we might not (ever)
    re-render if we hit a transient failure.
    
    Additionally, returning a non-nil err causes the reconciliation to be
    quickly retried, and evendtually dropped. If there is a transient
    condition for more than 1000 seconds, controller-runtime will eventually
    give up entirely.
    
    So, don't return err, and manually request a retry.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"reflect"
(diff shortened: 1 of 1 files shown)


squeed:
There's a way better way to do this.
/close

PR #1409, opened 2022-04-29
Bug 2080255: SDN: Re-add list/watch/get permissions for nodes needed for EgressIP

andreaskaris:

### DIFF
--- a/bindata/network/openshift-sdn/002-rbac.yaml
+++ b/bindata/network/openshift-sdn/002-rbac.yaml
@@ -20,6 +20,7 @@ rules:
   - endpoints
   - services
   - pods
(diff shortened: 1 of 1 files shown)


bindata/network/openshift-sdn/002-rbac.yaml
```diff
@@ -31,6 +31,13 @@ rules:
   verbs:
   - list
   - watch
+- apiGroups: [""]
+  resources:
+  - nodes
```
danwinship, line 36: you can just add `- nodes` to the existing `apiGroups: [""]` section above; it was separate before because it had different verbs than the other resources

PR #1416, opened 2022-05-02
Re-reconcile network on configmap, stop watching all configmaps in proxy controllers

squeed:
This PR encompasses three separate changes that take the same shape:

1. Trigger a re-reconcile of the network when a Node is added or a ConfigMap changes in the CNO's namespace. This is because we saw HyperShift hang when no Nodes existed, and thus we couldn't run the MTU prober.
2. Switch the configmap-ca-injector to use a label-scoped watch, rather than watching all configmaps in the entire world.
3. Switch the ProxyConfig controller to use a namespace-scoped watch

### DIFF
--- a/pkg/client/fake/fake_client.go
+++ b/pkg/client/fake/fake_client.go
@@ -11,6 +11,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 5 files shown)


squeed:
@JacobTanenbaum can you give the proxy changes a look-over?

squeed:
/hold
want to do a small refactor.

squeed:
This is ready for review. @JacobTanenbaum can you take a look?

PR #1420, opened 2022-05-03
Hypershift: Fix ovnkube-master priority class and set resource requests on token-minter

alvaroaleman:
These two are needed before we can use ovnkube as default in CI, we have verifications for these that currently fail.

Ref  https://github.com/openshift/hypershift/pull/1331

/assign @squeed


squeed:
Test doesn't cover these objects

PR #1421, opened 2022-05-04
Remove ObjectMeta.ClusterName usage

kyrtapz:
Instead of using the deprecated `ObjectMeta.ClusterName` field, use an annotation that serves the same purpose

https://github.com/kubernetes/kubernetes/pull/108717

/cc @squeed @zshi-redhat 

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -24,6 +24,12 @@ type Object interface {
 	runtime.Object
 }
 
+// GetClusterName returns the names.ClusterNameAnnotation annotation for the specified object.
+// If the annotation does not exist it will return an empty string.
+func GetClusterName(obj Object) string {
+	return obj.GetAnnotations()[names.ClusterNameAnnotation]
+}
+
 // ApplyObject submits a server-side apply patch for the given object.
 // This causes fields we own to be updated, and fields we don't own to be preserved.
 // For more information, see https://kubernetes.io/docs/reference/using-api/server-side-apply/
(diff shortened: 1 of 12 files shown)


pkg/apply/apply.go
```diff
@@ -24,16 +24,22 @@ type Object interface {
 	runtime.Object
 }
 
+// GetClusterName returns the names.ClusterNameAnnotation annotation for the specified object.
+// If the annotation does not exist it will return an empty string.
+func GetClusterName(obj Object) string {
+	return obj.GetAnnotations()[names.ClusterNameAnnotation]
```
squeed, line 30: This should probably return the default cluster name if empty, since it's useful in other circumstances.

pkg/apply/apply.go
```diff
@@ -157,7 +163,7 @@ func getCopySource(ctx context.Context, obj Object, client cnoclient.Client) (Ob
 
 	ret.SetNamespace(obj.GetNamespace())
 	ret.SetName(obj.GetName())
-	ret.SetClusterName(obj.GetClusterName())
+	ret.SetClusterName(GetClusterName(obj))
```
squeed, line 166: We should just delete this line.

pkg/apply/apply.go
squeed: Hmm, now I'm not so sure.

kyrtapz: We rely on empty cluster name if we mean `default` I think we can keep this as is

squeed: agreed

PR #1425, opened 2022-05-05
Add gm metric record to use for telemetry exposure

tssurya:
/cc @npinaeva @simonpasquier

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -13,6 +13,8 @@ spec:
   groups:
   - name: cluster-network-operator-master.rules
     rules:
+    - record: cluster:ovnkube_master_egress_routing_via_host:max
+      expr: max(ovnkube_master_egress_routing_via_host)
     - alert: NoRunningOvnMaster
       annotations:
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -13,6 +13,8 @@ spec:
   groups:
   - name: cluster-network-operator-master.rules
     rules:
+    - record: cluster:ovnkube_master_egress_routing_via_host:max
+      expr: max(ovnkube_master_egress_routing_via_host)
```
andreaskaris, line 17: Does avg() make sense here? As in ... if all 3 have the same gm mode, we get 0 or 1 ... but if there's a disagreement, we get a fraction?

tssurya: there cannot be disagreement for this metric. only the leader exposes it. so it will always be 1, 0, 0 so max makes sense.

andreaskaris: ah ok thanks :-)

simonpasquier (commented):
/lgtm

PR #1431, opened 2022-05-09
Bug 2086143: Status controller: use a label, rather than watching all objects

squeed:
Previously, we watched all DaemonSets, Deployments, and StatefulSets across the cluster, and kept a separate list of interesting names. But this is a waste of memory and bandwidth.

Instead, set up a label-selected watch, and label our objects with that. Then just trigger a re-reconcile on that.


pkg/controller/statusmanager/kube.go
```diff
@@ -0,0 +1,124 @@
...
+
+type patchAnnotations struct {
+	Metadata md `json:"metadata"`
+}
+type md struct {
+	Annotations map[string]interface{} `json:"annotations"`
+}
+
+func makeAnnotationPatch(key string, value *string) ([]byte, error) {
+	patch := &patchAnnotations{
```
danwinship, line 37: FWIW you could do

    patch := &metav1.PartialObjectMetadata{

pkg/controller/statusmanager/kube.go
```diff
@@ -0,0 +1,124 @@
...
+		obj.Name,
+		types.MergePatchType,
+		patchData,
+		metav1.PatchOptions{},
+	)
+	return err
+}
+
+// setSSAnnotation sets an annotation on a statefulset; or unsets it if value is nil
+func (status *StatusManager) setSSAnnotation(obj *appsv1.StatefulSet, key string, value *string) error {
```
danwinship, line 84: I know you're just copying the existing names but `SS` in particular is not obvious to me. None of these function names would really be too long if you just fully unabbreviated them...

Although also, given the whole dynamic client thing couldn't you just squash them into a single method? Or were we trying to get away from the dynamic client?

pkg/controller/statusmanager/pod_status.go
```diff
@@ -72,24 +73,36 @@ type statefulsetState struct {
...
-			continue
+	if (len(daemonSets) + len(deployments) + len(statefulSets)) == 0 {
+		progressing = append(progressing, "Deploying")
+	}
+
+	for _, ds := range daemonSets {
+		dsName := ClusteredName{
+			Namespace:   ds.Namespace,
+			Name:        ds.Name,
+			ClusterName: apply.GetClusterName(ds),
```
danwinship, line 105: though not really related to this patch:
1. It is not clear why `GetClusterName` should be in `pkg/apply`
2. `dsName := NewClusteredName(ds)`
```
func NewClusteredName(obj metav1.Object) ClusteredName {
        return ClusteredName{
                Namespace:   obj.GetNamespace(),
                Name:        obj.GetName(),
                ClusterName: obj.GetAnnotations()[name.ClusterNameAnnotation],
        }
}
```

pkg/controller/statusmanager/pod_watcher.go
```diff
@@ -0,0 +1,67 @@
...
+		); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Reconcile triggers a re-update of Status.
+func (p *PodWatcher) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	klog.Infof("Operand updated, re-generating status")
```
danwinship, line 55: that's super vague...

pkg/controller/statusmanager/status_manager.go
```diff
@@ -65,30 +68,94 @@ func (c ClusteredName) String() string {
...
-	statefulSets   []ClusteredName
-	relatedObjects []configv1.ObjectReference
+	dsInformer cache.SharedIndexInformer
+	dsLister   DaemonSetLister
 
-	hyperShiftConfig *network.HyperShiftConfig
+	depInformer cache.SharedIndexInformer
+	depLister   DeploymentLister
+
+	// StatefulSets may exist in more than one cluster (in the case of Hypershift)
```
danwinship, line 90: It seems weird to treat this like there's some inherent difference between StatefulSets and Deployments/DaemonSets, as opposed to this just being a contingent truth about OCP 4.11. (Which is what it is, right?) Maybe there would be some simplifications if you made the three types more consistent in the reorg?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -65,30 +68,94 @@ func (c ClusteredName) String() string {
...
 		hyperShiftConfig: network.NewHyperShiftConfig(),
 	}
+	s.initInformers()
+	return s
+}
+
+func (s *StatusManager) initInformers() {
+	s.dsInformer = v1appsinformers.NewFilteredDaemonSetInformer(
+		s.client.Default().Kubernetes(),
+		"", // all namespaces
```
danwinship, line 110: `metav1.NamespaceAll`
(and again below)

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -294,19 +251,17 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
-			deployments = append(deployments, statusmanager.ClusteredName{ClusterName: apply.GetClusterName(obj), Namespace: obj.GetNamespace(), Name: obj.GetName()})
-		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "StatefulSet" {
-			statefulSets = append(statefulSets, statusmanager.ClusteredName{ClusterName: apply.GetClusterName(obj), Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		// Label all DaemonSets, Deployments, and StatefulSets with the label that generates Status.
+		if obj.GetAPIVersion() == "apps/v1" && (obj.GetKind() == "DaemonSet" || obj.GetKind() == "Deployment" || obj.GetKind() == "StatefulSet") {
+			l := obj.GetLabels()
+			if l == nil {
+				l = map[string]string{}
+			}
+			l[names.GenerateStatusLabel] = "true"
```
danwinship, line 263: Standard convention is that `""` means true and (unset) means false

pkg/names/names.go
```diff
@@ -41,6 +41,12 @@ const CreateWaitAnnotation = "networkoperator.openshift.io/create-wait"
 // that they are not critical to the functioning of the pod network
 const NonCriticalAnnotation = "networkoperator.openshift.io/non-critical"
 
+// GenerateStatusLabel can be set by the various Controllers to tell the
+// StatusController that this object is relevant, and should be included
+// when genetating status from deployed pods.
+// Currently, this is looked for on Deployments, DaemonSets, and StatefulSets
+const GenerateStatusLabel = "networkoperator.openshift.io/generate-status"
```
danwinship, line 48: "generate-status" sounds like it means "someone should generate status for this pod". I'd suggest something more "deployed by CNO" / "tracked by CNO" ish.

pkg/controller/statusmanager/kube.go
squeed: oh nice!

pkg/controller/statusmanager/kube.go
squeed: There are several annoying effects in play here.

The first is that it's not possible to create an arbitrary label-scoped informer using the controller-runtime client (at least yet). One needs to manually create an informer. Assuming you don't want to pass Unstructured around, then you need to use the typed client to create the informers. Okay, fine, we create the informers ourselves and just wire it up. No big deal

The second problem is exclusively around testing. Basically, we offer 3 clients: the controller-runtime client, the Unstructured client, and the Structured client. Annoyingly, the fake clients **all have separate stores**. This means that the status controller's extensive test suite needs to use the same client for the same objects.

But now that I type this up, I think I have a way to make this problem go away, so it's good that you mentioned this.

pkg/controller/statusmanager/status_manager.go
squeed: This is all a good point. As it stands right now, only StatefulSets are deployed in multiple clusters. That could certainly change, however, so, you're right, this should all take the same shape.

pkg/controller/statusmanager/pod_status.go
squeed: 1. Agreed, but changing that is another PR.
2. That is a good idea.

pkg/controller/statusmanager/status_manager.go
squeed: Good point.

pkg/names/names.go
squeed: well, I was trying to make this explicit: this is consumed by the status controller. I could imagine naming this "networkoperator/operand-of=$controller-name", which would be cute, but doesn't specify what it **does**.

squeed:
@danwinship updated, thanks for your feedback, as always.

squeed:
CI is green :)

squeed:
This is looking good. It reduces memory usage on a CI run from 120 MB to 80MB.
cno-memory

squeed:
Update: Got a memory pull from a busier cluster, CNO usage was up to 400 MB. Now it should stay steady around 80.

pkg/names/names.go
danwinship: > well, I was trying to make this explicit: this is consumed by the status controller.

Yeah, but I'm saying that "generate-status" doesn't imply that to me at all. Especially, it's not clear that the "status" there refers to _someone else's_ status, not the object's own status.

squeed: How about (all namespaced with `networkoperator.openshift.io`
1. `controller=$controller`
2. `managed-by=$controller`
3. `operand-of=$controller`

squeed: Leaning towards `networkoperator.openshift.io/generates-operator-status`.

squeed: OK, updated based on this discussion.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -344,19 +302,17 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
 	relatedObjects := []configv1.ObjectReference{}
 	relatedClusterObjects := []network.RelatedObject{}
 	for _, obj := range objs {
-		if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "DaemonSet" {
-			daemonSets = append(daemonSets, statusmanager.ClusteredName{ClusterName: apply.GetClusterName(obj), Namespace: obj.GetNamespace(), Name: obj.GetName()})
-		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "Deployment" {
-			deployments = append(deployments, statusmanager.ClusteredName{ClusterName: apply.GetClusterName(obj), Namespace: obj.GetNamespace(), Name: obj.GetName()})
-		} else if obj.GetAPIVersion() == "apps/v1" && obj.GetKind() == "StatefulSet" {
-			statefulSets = append(statefulSets, statusmanager.ClusteredName{ClusterName: apply.GetClusterName(obj), Namespace: obj.GetNamespace(), Name: obj.GetName()})
+		// Label all DaemonSets, Deployments, and StatefulSets with the label that generates Status.
```
npinaeva, line 308: Do we ever need to remove the label, or all outdated object should be deleted by deleteRelatedObjectsNotRendered?

squeed: I doubt we will ever remove the label, but doing so would be handled correctly: the informer would trigger and we would re-generate status from the remaining objects.

squeed: (The informer treats removing the label as a "delete")

npinaeva: I mean do we need to clean the label for non-already-related objects - in this PR we only add this label, but if some object that was labeled gets outdated (e.g during upgrade old daemonsets will still be labeled) then controller will keep considering that object as a part of a status since it has the label.
So the question is - is it even possible, or all "stale" objects should be just deleted fast enough to not confuse status manager?

squeed: I'm not sure I entirely understand the scenario you're talking about. But, in any case, I don't think the status manager will be "confused". It will always generate a reasonable status based on the objects in the cluster with the label. If a stale object is deleted, then it's deletion will cause status to be re-generated.

We're no longer maintaining a list of desired objects. The source of truth is the informer's store. So now, hopefully, we should be *more consistent*.

kyrtapz:
Overall looks good to me!
It does not work in hypershift though, we do not currently have permissions to watch deployments and daemonsets in the hosted cluster namespace: `ci/prow/e2e-hypershift` https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1431/pull-ci-openshift-cluster-network-operator-master-e2e-hypershift/1524400890348310528/artifacts/e2e-hypershift/dump/artifacts/namespaces/clusters-53aea1ea-d139/core/pods/logs/cluster-network-operator-7668f465df-j924l-cluster-network-operator.log

squeed:
Oh duh, good catch! fixing.

squeed:
@kyrtapz fixed, thanks!

squeed:
/retest
(image hiccup)

squeed:
/retest
now that CI is fixed

squeed:
/test e2e-gcp
/test e2e-gcp-ovn
quota issues

note: hypershift is green again.

squeed:
seems to be failing a lot; this PR is fine

PR #1432, opened 2022-05-09
Bug 2023295: Cleanup CNO relatedObjects

npinaeva:
Remove NetworkPolicy, EgressFirewall, EgressIP and CloudPrivateIPConfig from CNO relatedObjects, since
1. it doesn't match relatedObjects definition ("Common uses are: 1. the detailed resource driving the operator 2. operator namespaces 3. operand namespaces")
2. it makes must-gather collect instances of these resource for all namespaces, and must-gather shouldn't collect resources from custom namespaces

egressips and cloudprivateipconfigs will be collected as cluster-scoped resources https://github.com/openshift/must-gather/pull/300 and egressfirewalls and networkpolicies will be collected as namespaced resources. https://github.com/openshift/oc/pull/1128

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -179,6 +179,12 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				log.Printf("Object Kind is network.operator.openshift.io, skip")
 				continue
 			}
+			// @npinaeva objects without a name shouldn't be listed as relatedObjects in the first place,
+			// and we should never try to delete them
+			if currentObj.Name == "" {
+				log.Printf("Object without a name GVK %+v, skip", gvk)
+				continue
+			}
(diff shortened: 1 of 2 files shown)


squeed:
@npinaeva should we wait for those other PRs to merge before this goes in?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -179,6 +179,15 @@ func (status *StatusManager) deleteRelatedObjectsNotRendered(co *configv1.Cluste
 				log.Printf("Object Kind is network.operator.openshift.io, skip")
 				continue
 			}
+			// @npinaeva listed objects were removed from CNO relatesObjects,
```
squeed, line 182: typo: related.

npinaeva:
It looks like we shouldn't add relatedObjects without a name in the first place, but maybe replacing my newly-added if block with no-name skip will be a good idea? Because I deleted all objects that don't have a name and they will match no-name condition.

squeed:
/lgtm
seems the "dependent" PRs are about to go in themselves

PR #1433, opened 2022-05-09
Bug 2082235: manifests: Add in service, service-cert, and ServiceMonitor

squeed:
That way we actually use a securely generated certificate for the CNO.


squeed:
@kyrtapz can you take a quick look at this?

PR #1367, opened 2022-04-04
Bug 2082611: Limit Kuryr pods permissions

dulek:
This splits the Kuryr RBAC into separate ClusterRoles for controller nad
daemon and makes sure we don't keep unnecessary permissions. I also
dropped the ones related to sriov and dpdk support as we don't configure
them in OpenShift.

PR #1435, opened 2022-05-09
Bug 2083266: Limit Kuryr pods permissions

dulek:
This splits the Kuryr RBAC into separate ClusterRoles for controller nad
daemon and makes sure we don't keep unnecessary permissions. I also
dropped the ones related to sriov and dpdk support as we don't configure
them in OpenShift.

### DIFF
--- a/bindata/network/kuryr/001-rbac.yaml
+++ b/bindata/network/kuryr/001-rbac.yaml
@@ -2,15 +2,16 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: kuryr
+  name: kuryr-controller
 rules:
 - apiGroups: [""]
   resources:
-  - namespaces
   - nodes
-  - endpoints
-  - services
-  - services/status
+  verbs:
+  - get
+  - list
+- apiGroups: [""]
+  resources:
   - pods
   verbs:
   - get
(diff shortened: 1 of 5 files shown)


dulek:
@MaysaMacedo, @gryf, please note that I needed to add delete permissions for pods to support the webhook.

bindata/network/kuryr/001-rbac.yaml
```diff
@@ -19,13 +20,18 @@ rules:
   - update
   - patch
   - delete
```
MaysaMacedo, line 22: Is this needed?
https://github.com/openshift/cluster-network-operator/pull/1367/files#diff-b214c5fc4fe51960ff3611bc799ae5fd2815388ae066350a211c899dcec45068R18-R25

dulek: Admission controller does DNS pod deletion to force them to run with TCP DNS: https://github.com/openshift/kuryr-kubernetes/blob/3385e1dbb0498e75ef2c03221f93228512505cbf/kuryr_kubernetes/cmd/webhook_server.py#L57-L59

I hate it, we don't support OSP13 in 4.10 anyway, but I'd rather not remove a "functionality" as a bugfix.

dulek:
@eurijon, this now only needs your approval.

PR #1439, opened 2022-05-12
WIP: [RFE-2181] To support transparent proxy configuration

TrilokGeer:
Added configType support during installation to signify transparent or explicit proxy configuration to enable user-ca-bundle configured with cluster proxy and determine configuration type for cluster-wide proxy.


TrilokGeer:
Consensus reached to use wildcard "*" set to noproxy field to indicate transparent proxy configuration. Hence closing this PR.

PR #1440, opened 2022-05-12
Bug 2087135: Fixing Hypershift nodeport flow

tsorya:
Bug 2087135: Fixing Hypershift nodeport flow:
    1. In case of NodePort ClusterIp should not be set
    2. Using the right client(management) to get NodePort service from
       management cluster
    3. Using right variable to set ovnsdb nodeport for ovnkube-node

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -157,6 +157,8 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVNDbServiceType"] = corev1.ServiceTypeClusterIP
 	data.Data["OVNSbDbRouteHost"] = nil
 	data.Data["OVN_SB_NODE_PORT"] = nil
+	data.Data["OVN_NB_DB_ENDPOINT"] = fmt.Sprintf("ssl:%s:%s", bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.OVNSbDbRouteHost, OVN_SB_DB_ROUTE_PORT)
+	data.Data["OVN_SB_DB_ENDPOINT"] = fmt.Sprintf("ssl:%s:%s", bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.OVNSbDbRouteHost, OVN_SB_DB_ROUTE_PORT)
 	pubStrategy := bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.ServicePublishingStrategy
 	if pubStrategy != nil && pubStrategy.Type == hyperv1.Route {
 		if pubStrategy.Route != nil && pubStrategy.Route.Hostname != "" {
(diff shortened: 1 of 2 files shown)


zshi-redhat:
LGTM
cc @kyrtapz @squeed

squeed:
/approve
/lgtm
/retest-required
(hypershift ci passes)

pkg/network/ovn_kubernetes.go
```diff
@@ -157,19 +157,19 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 		if pubStrategy.Route != nil && pubStrategy.Route.Hostname != "" {
 			data.Data["OVNSbDbRouteHost"] = pubStrategy.Route.Hostname
 		}
 	} else if pubStrategy != nil && pubStrategy.Type == hyperv1.NodePort {
 		data.Data["OVNDbServiceType"] = corev1.ServiceTypeNodePort
-		if pubStrategy.NodePort != nil && pubStrategy.NodePort.Port > 0 {
-			data.Data["OVN_SB_NODE_PORT"] = pubStrategy.NodePort.Port
-		}
+		data.Data["OVN_SB_NODE_PORT"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.OVNSbDbRouteNodePort
+		data.Data["OVN_NB_DB_ENDPOINT"] = fmt.Sprintf("ssl:%s:%d", bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.OVNSbDbRouteHost, bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.OVNSbDbRouteNodePort)
```
squeed, line 170: @zshi-redhat the NB_DB_ENDPOINT doesn't seem useful, right? Can we get rid of it?

zshi-redhat: ovnk still uses nbdb endpoint for hybrid case if I remember correctly, we would need to remove that in ovnk, then change cno, although  hybrid won't work today in hypershift in 4.11 due to the same reason.

squeed:
@tsorya is there a bug for this? If not, can you please file one (otherwise it can't merge, thanks to feature freeze)

tsorya:
> @tsorya is there a bug for this? If not, can you please file one (otherwise it can't merge, thanks to feature freeze)

will open one

pkg/network/ovn_kubernetes.go
```diff
@@ -537,7 +537,8 @@ func bootstrapOVNHyperShiftConfig(hc *HyperShiftConfig, kubeClient cnoclient.Cli
 	case hyperv1.NodePort:
 		{
 			svc := &corev1.Service{}
-			err = kubeClient.Default().CRClient().Get(context.TODO(), types.NamespacedName{Namespace: hc.Namespace, Name: "ovnkube-master-external"}, svc)
+			clusterClient := kubeClient.ClientFor(client.ManagementClusterName)
```
squeed, line 540: good catch.

PR #1441, opened 2022-05-12
Bug 2086461: AWS: Use hardcoded MTU to speed up cluster creation

alvaroaleman:
Currently with Hypershift on AWS, we need to wait for a node to join the
cluster before we start setting up the controlplane components of the
networking stuck, because we need the MTU which we find out through
probing with a pod. This slows down cluster creation.

This change makes us short-circuit the pod deployment on AWS and use a
hardcoded value instead to avoid that.

### DIFF
--- a/pkg/controller/operconfig/mtu_probe.go
+++ b/pkg/controller/operconfig/mtu_probe.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	"github.com/openshift/cluster-network-operator/pkg/render"
 
+	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


alvaroaleman:
ci/prow/e2e-hypershift succeeded :)

alvaroaleman:
This saves about 70 seconds for the initial rollout (718s with this change vs 792s without)

squeed:
This seems reasonable.

I think, ultimately, the "right" way to do this would be to add a `UnderlayMTU` field to the `configv1.PlatformStatus` type, but that's obviously a lot more work to plumb in.

The question is: is the assumption that **all** AWS installations are MTU 9001 correct? I'm not sure sure. How might we detect and handle any divergent cases?

squeed:
> The question is: is the assumption that **all** AWS installations are MTU 9001 correct? I'm not sure. How might we detect and handle any divergent cases?

First idea: only assume MTU if the platform is AWS and we're operating in Hypershift.

alvaroaleman:
The only test that fails in the hypershift conformnance run is ` etcd cluster has the same number of master nodes and voting members from the endpoints configmap` which is a new test that is incompatible with Hypershift and fails the same way [in our periodic runs](https://testgrid.k8s.io/redhat-hypershift#aws-conformance-4.11), a fix for it is in the pipeline: https://github.com/openshift/origin/pull/27149

Thus I think this is good to go, ptal.

PR #1443, opened 2022-05-13
multi-networkpolicy: Enable on SR-IOV networks

zeeke:
Add `sriov` to the array of network plugins multus-networkpolicy-iptables should manage.


s1061123:
@zeeke thank you for PR. BTW, did you test the diff works fine in cluster (i.e. replace CNO and verify that)?

s1061123 (approved):

PR #1446, opened 2022-05-16
Bug 2086544: Stop passing hosted cluster token as a parameter to ovnkube-master

kyrtapz:
Since `k8s-token-file` parameter is provided there is no need to pass `k8s-token` as well.
Previously if the token was provided it could be logged as the script using it runs with `-x`.
Additionally add a wait loop for the token file to appear.

/cc @squeed @zshi-redhat 


squeed:
/approve
/lgtm
/hold
until hypershift CI passes, feel free to un-hold.

squeed:
hypershift ci passed
/hold cancel

PR #1448, opened 2022-05-18
Bug 2087556: Fix rendering DPU manifests

zshi-redhat:


pliurh:
/lgtm
The patch works in my environment.

PR #1449, opened 2022-05-18
[CFE-465] As a developer I want to support wildcard character for noProxy

TrilokGeer:
Added wildcard '*' validation support to signify transparent proxy when there is no proxy set for http/https config.

### DIFF
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -36,7 +36,9 @@ const (
 // noProxy fields of proxyConfig are valid.
 func (r *ReconcileProxyConfig) ValidateProxyConfig(proxyConfig *configv1.ProxySpec) error {
 	if !isSpecHTTPProxySet(proxyConfig) && !isSpecHTTPSProxySet(proxyConfig) {
-		return fmt.Errorf("httpProxy or httpsProxy must be set when using proxy")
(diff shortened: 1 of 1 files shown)


pkg/controller/proxyconfig/controller.go
```diff
@@ -329,6 +329,11 @@ func (r *ReconcileProxyConfig) Reconcile(ctx context.Context, request reconcile.
 	return reconcile.Result{}, nil
 }
 
+// isOnlyWildCardForNoProxy returns true if Ispec.NoProxy of proxyConfig is set to "*".
+func isOnlyWildCardSetForNoProxy(proxyConfig *configv1.ProxySpec) bool {
+	return proxyConfig.NoProxy == noProxyWildcard
```
wking, line 334: Personally, this `==` comparison is straightforward enough that I'd use it inline, instead of wrapping it in `isOnlyWildCardSetForNoProxy`.  But :shrug:, it will work either way.

danwinship:
/hold

This is completely non-obvious API, and goes against the ordinary usage of `http_proxy`/`https_proxy`/`no_proxy`. If we want to have a mode for "assume there is a transparent proxy", then that should be indicated explicitly in the configuration, not inferred via magic values in the existing config.

TrilokGeer:
closed in favor of https://github.com/openshift/installer/pull/6232

PR #1450, opened 2022-05-18
Bug 2076776: remove patch permissions from ovnkube-node service account

msherif1234:
remove patch permission for ovnkube-node service for the following resources
- namespaces
- endpoints
- services
- configmaps

Kept same resource permissions for :-
- pods because its in use by `SetAnnotationsOnPod` used by ovn `addLogicalPort`
- nodes because its in use by `SetTaintOnNode` and `RemoveTaintFromNode`

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac.yaml
@@ -7,22 +7,65 @@ metadata:
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: openshift-ovn-kubernetes-node
+  namespace: openshift-ovn-kubernetes
+rules:
+- apiGroups: [""]
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups: [certificates.k8s.io]
+  resources: ['certificatesigningrequests']
+  verbs:
+    - create
+    - get
+    - delete
+    - update
+    - list
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: openshift-ovn-kubernetes-node
+  namespace: openshift-ovn-kubernetes
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: openshift-ovn-kubernetes-node
+subjects:
+- kind: ServiceAccount
+  name: ovn-kubernetes-node
+  namespace: openshift-ovn-kubernetes
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: openshift-ovn-kubernetes-node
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/common/002-rbac.yaml
```diff
@@ -13,16 +13,22 @@ metadata:
...
+  - get
+  - list
+  - watch
+  - patch
+- apiGroups: [""]
+  resources:
+  - namespaces
+  - endpoints
   - services
   - configmaps
```
squeed, line 27: We don't need access to all configmaps across all namespaces. Can you add a separate Role (not ClusterRole) for the openshift-ovn-kubernetes namespace?

bindata/network/ovn-kubernetes/common/002-rbac.yaml
```diff
@@ -13,16 +13,22 @@ metadata:
 rules:
 - apiGroups: [""]
   resources:
-  - endpoints
```
squeed, line 16: How does the unidler work? (I forget). Does the node update the Endpoints object?

danwinship: When someone connects to an idled service, ovn-kubernetes emits an Event which is observed by openshift-controller-manager, which removes the annotation and scales up the service.

danwinship:
> * pods because its in use by `SetAnnotationsOnPod` used by ovn `addLogicalPort`

We should create a jira issue about looking into reducing the permissions there.

> * nodes because its in use by `SetTaintOnNode` and `RemoveTaintFromNode`

We should probably remove that functionality like we did in sdn though, and have it just emit an Event and exit with an error in that case. It's a thing that's not supposed to happen anyway; no reason to be giving the pod extra privileges it's not even expected to ever use.

msherif1234:
> > * pods because its in use by `SetAnnotationsOnPod` used by ovn `addLogicalPort`
> 
> We should create a jira issue about looking into reducing the permissions there.
> 
> > * nodes because its in use by `SetTaintOnNode` and `RemoveTaintFromNode`
> 
> We should probably remove that functionality like we did in sdn though, and have it just emit an Event and exit with an error in that case. It's a thing that's not supposed to happen anyway; no reason to be giving the pod extra privileges it's not even expected to ever use.
I was planning on doing the same as u did for sdn but I think I saw ur PR was reverted so wanted to watch this settle b4 using it with ovnk

danwinship:
> I was planning on doing the same as u did for sdn but I think I saw ur PR was reverted so wanted to watch this settle b4 using it with ovnk in separate PR

It wasn't reverted; there was just a followup fixup because I'd removed too many permissions.

Anyway, I'm fine with doing that as a separate PR, I was just pointing out that we should definitely do it.

squeed:
This seems reasonable.
/approve
Can we move CertificateSigningReuests to the Role (and out of the ClusterRole) as well?

PR #1453, opened 2022-05-19
ensures type: Directory for multus host paths

dougbtv:
In case there's a possible timing conflict with the MCO creating these paths, this will wait for the paths.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -185,12 +185,15 @@ spec:
         - name: system-cni-dir
           hostPath:
             path: {{ .SystemCNIConfDir }}
+            type: Directory
         - name: multus-cni-dir
           hostPath:
             path: {{ .MultusCNIConfDir }}
+            type: Directory
         - name: cnibin
           hostPath:
             path: {{ .CNIBinDir }}
+            type: Directory
         - name: os-release
           hostPath:
             path: /etc/os-release


bindata/network/multus/multus.yaml
```diff
@@ -185,12 +185,15 @@ spec:
         - name: system-cni-dir
           hostPath:
             path: {{ .SystemCNIConfDir }}
+            type: Directory
```
dcbw, line 188: @dougbtv is DirectoryOrCreate useful here, or do we need specific permissions on the directory? IIRC DirectoryOrCreate will use 755 and kubelet's user/group which may not be what we want, so maybe it's not useful.

squeed: Agreed, MCO owns these directories, so DirectoryOrCreate is not the right choice here

squeed:
@dougbtv Seems reasonable, but needs a bug for this release.

dougbtv:
/retitle Bug 2089968: ensures type: Directory for multus host paths

squeed:
/lgtm
(test basically succeeded, one uninteresting failure)

bindata/network/multus/multus.yaml
dcbw: Oh hah, yeah, I'll be containernetworking-cni RPMs create the dir, or maybe CRIO RPMs do. Fun times.

PR #1454, opened 2022-05-19
Bug 2086461: Hypershift: Also add default for Azure mtu

alvaroaleman:
This change adds a default for the Azure mtu to speed up cluster
creation, similiar like what already exists for AWS.

Ref:
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-tcpip-performance-tuning

/assign @squeed

### DIFF
--- a/pkg/controller/operconfig/mtu_probe.go
+++ b/pkg/controller/operconfig/mtu_probe.go
@@ -30,6 +30,7 @@ const (
 	cmNamespace = "openshift-network-operator"
 	cmName      = "mtu"
 	awsMTU      = 9001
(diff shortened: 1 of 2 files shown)


danwinship:
do we not support hypershift on GCP? (And if we do, then maybe a map from platform type to default mtu?)

alvaroaleman:
@danwinship currently it only supports AWS, Azure and a bring your own node workflow (in which we won't be able to know ahead of time what MTU to use)

squeed:
In the end, it would be good to add a map, but this seems OK for now
/lgtm
/approve

PR #1462, opened 2022-05-26
Bug 2073452: Copying CNI binaries should be an atomic operation.

dougbtv:
It was previously copying directly to where the binaries are executed, which can cause for half-written binaries to be executed.


dcbw:
https://coreos.slack.com/archives/CM4ERHBJS/p1653496014119409

dcbw:
https://coreos.slack.com/archives/CM4ERHBJS/p1653496014119409

PR #1466, opened 2022-05-31
Revert "Copying CNI binaries should be an atomic operation."

neisw:
This reverts commit c0db22ad50e955cdfa0f41df777b049b6179619c.


deads2k:
This reverts the last merge into the repo, so we we'll match an image that was already in the payload once before.

PR #1475, opened 2022-06-03
Bug 2093399: Drop Node update permission for ovnkube-node

msherif1234:
[PR#1127](https://github.com/openshift/ovn-kubernetes/pull/1127) remove node tainting used when MTU is too small then we don't need to update/patch node resource


danwinship:
/lgtm
assuming the tests pass

msherif1234:
`oc.kube.SetAnnotationsOnNode` needs node patch operation so we can't drop patch access

PR #1414, opened 2022-05-02
ovn: fix northd preStop command handling

dcbw:
Need to exec a shell so we can set the environment variable
we need that tells ovn-ctl not to touch the databases when
stopping northd.

@trozet

PR #1477, opened 2022-06-09
[release-4.10] Bug 2095111: ovn: fix northd preStop command handling

dcbw:
Need to exec a shell so we can set the environment variable
we need that tells ovn-ctl not to touch the databases when
stopping northd.

(cherry picked from commit 599a9d5060e3e5bb79f93e2f321bc4592dfa7a22)
Backport of https://github.com/openshift/cluster-network-operator/pull/1414

### DIFF
--- a/bindata/network/ovn-kubernetes/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/ovnkube-master.yaml
@@ -95,9 +95,9 @@ spec:
           preStop:
             exec:
               command:
-                - OVN_MANAGE_OVSDB=no
-                - /usr/share/ovn/scripts/ovn-ctl
-                - stop_northd
+                - /bin/bash
+                - -c
+                - OVN_MANAGE_OVSDB=no /usr/share/ovn/scripts/ovn-ctl stop_northd
         env:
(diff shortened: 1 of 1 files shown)


dcbw:
@jluhrsen @mccv1r0 could either one of you take a look and /lgtm if good?

mccv1r0 (approved):

maiqueb (approved):

mffiedler:
@rbbratta PTAL

bindata/network/ovn-kubernetes/ovnkube-master.yaml
```diff
@@ -95,9 +95,9 @@ spec:
           preStop:
             exec:
               command:
-                - OVN_MANAGE_OVSDB=no
-                - /usr/share/ovn/scripts/ovn-ctl
-                - stop_northd
+                - /bin/bash
+                - -c
+                - OVN_MANAGE_OVSDB=no /usr/share/ovn/scripts/ovn-ctl stop_northd
```
rbbratta, line 100: If this is all we need, then it would be better to avoid the attack surface that is bash by using coreutils `/usr/bin/env` which was designed for this purpose.

PR #1484, opened 2022-06-13
Bug 2093057: Verify pod's connectivity in postStart hook

kyrtapz:
When network policies match the pod, postStart hook does not have any connectivity in its first run because status.podIP is not yet set(This affects clusters using OpenshiftSDN CNI):
https://github.com/kubernetes/kubernetes/issues/85966

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
@@ -316,6 +316,23 @@ spec:
               - -c
               - |
                 set -x
+
+                # With some plugins like openshift-sdn, NetworkPolicy depends on each pod's Status.PodIP being added to the policy rules.
+                # Kubelet only sets a pod's Status.PodIP when all containers of the pod have started at least once (successfully or unsuccessfully).
+                # Container start is blocked by postStart hooks. See https://github.com/kubernetes/kubernetes/issues/85966 for more details.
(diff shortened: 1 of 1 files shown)


dcbw:
@kyrtapz could you clarify the fix a bit more in the commit message and the code comments? You're describing the message, but not really why restarting the pod makes things work. How about something like this?

```
With some plugins like openshift-sdn, NetworkPolicy depends on each pod's Status.PodIP being added to the policy rules. Kubelet only sets a pod's Status.PodIP when all containers of the pod have started at least once (successfully or unsuccessfully). Container start is blocked by postStart hooks. See https://github.com/kubernetes/kubernetes/issues/85966 for more details.

The NB and SB DB postStart hooks block until the DBs join the RAFT cluster (which requires network connectivity) or until a timeout is reached. Thus the first time the DB containers start their pod's Status.PodIP is not yet set, so NetworkPolicy blocks connectivity, the DB can't join the RAFT cluster, and the postStart hook fails after a long timeout.

To work around this long timeout, exit early if the container has no connectivity, presumably because this is the first time the container has started and Status.PodIP is not yet set. This lets kubelet proceed with setting Status.PodIP, after which it will eventually restart the failed containers, which will now have network connectivity and succeed.
```

kyrtapz:
This change won't be needed once https://github.com/openshift/sdn/pull/439 gets in

PR #1490, opened 2022-06-17
k8s 1.24 bump: add RBAC coordination leases for ovn-k master and replace node-role.kubernetes.io/master

ricky-rav:
Modifications required by changes in kubernetes 1.24


flavio-fernandes (commented):
/lgtm

Very similar to what we did in https://github.com/openshift/cluster-network-operator/pull/1461

PR #1491, opened 2022-06-17
Bug 2060079: Enhance sensitivity of SDN alert NodeProxyApplySlow

martinkennelly:
Before this patch, alert NodeProxyApplySlow calculates the 95th
percentile of the _total_ time series
kubeproxy_sync_proxy_rules_duration_seconds and alerts if it is
above 15 seconds.

Because it looks at the total time series, it lacks the ability
to be sensitive to periodic spikes in proxy rules sync latency.

This should merge after code freeze so I do not blow up alerting for 4.11 and it has sufficient soak test time.

/cc @danwinship

### DIFF
--- a/bindata/network/openshift-sdn/alert-rules.yaml
+++ b/bindata/network/openshift-sdn/alert-rules.yaml
@@ -40,7 +40,7 @@ spec:
         summary: OpenShift SDN pod {{"{{"}} $labels.pod {{"}}"}} on node {{"{{"}} $labels.node {{"}}"}} is taking too long to update proxy rules for services.
         description: Configuration of proxy rules for Kubernetes services in the node is taking too long and stale endpoints may exist.
       expr: |
-        histogram_quantile(.95, kubeproxy_sync_proxy_rules_duration_seconds_bucket)
+        histogram_quantile(.95, sum(rate(kubeproxy_sync_proxy_rules_duration_seconds_bucket[5m])) by (le, namespace, pod))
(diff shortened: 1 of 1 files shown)


JacobTanenbaum:
I think that making the alert more sensitive would not do any harm, `syncProxyRules()`  should occur faster then this alert would fire

martinkennelly:
Thanks Jacob. I am hoping to get insight from our scale team who performed [this](https://docs.openshift.com/container-platform/4.10/scalability_and_performance/planning-your-environment-according-to-object-maximums.html) testing.
During the test of max services + pods, I should see `syncProxyRules()` stressed to its max and this will give me the max upper latency limit we support. From there, I can have a better informed guess.

danwinship:
So yeah, the idea is that this should be triggering when iptables syncs are _currently_ slow, and should stop triggering when they stop being slow. AIUI the current behavior is more like "it triggers when the average across all time is too slow, and stays triggered even when things get better if the average is still too high", which is totally useless.

But I don't know promql enough to know if the PR is correct

martinkennelly:
> So yeah, the idea is that this should be triggering when iptables syncs are _currently_ slow, and should stop triggering when they stop being slow. AIUI the current behavior is more like "it triggers when the average across all time is too slow, and stays triggered even when things get better if the average is still too high", which is totally useless.
> 
> But I don't know promql enough to know if the PR is correct

Yeah, currently it considers all the time series instead of just reacting to a period of time.
I will share a google doc with you with before and after screenshots show this PRs improvement.
If that doesn't satisfy you, we can ask the monitoring team to review this.

PR #1493, opened 2022-06-21
Fix link to document on CVO run level

nishipy:
Currently, the link to CVO run level doc is outdated in `docs/architecture.md`, so this PR fixes it.


nishipy:
Hi @trozet, could you review this?

PR #1508, opened 2022-07-13
Bug 1894268: Allow users to specify ovnkube join subnet

pliurh:
Some users need to be able to specify the v4 and v6 join subnets
to use for ovn-k. In particular, they may be already using the
ones that ovn-k uses as default, and they need a way to be able to
specify a different set of subnets as join subnets. This commit
creates the config option for the same. Eventually this will be
consumed by the CNO using rendered bootstrap data and passed to the
ovn-k daemonsets.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -23,6 +23,7 @@ import (
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/render"
+	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
 	"github.com/openshift/cluster-network-operator/pkg/util/k8s"
 	hyperv1 "github.com/openshift/hypershift/api/v1alpha1"
 	"github.com/pkg/errors"
@@ -108,6 +109,8 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["TokenAudience"] = os.Getenv("TOKEN_AUDIENCE")
 	data.Data["MTU"] = c.MTU
 	data.Data["RoutableMTU"] = nil
+	data.Data["V4JoinSubnet"] = c.V4InternalSubnet
+	data.Data["V6JoinSubnet"] = c.V6InternalSubnet
 
 	if conf.Migration != nil && conf.Migration.MTU != nil {
 		if *conf.Migration.MTU.Network.From > *conf.Migration.MTU.Network.To {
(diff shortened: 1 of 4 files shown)


bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
```diff
@@ -756,6 +756,18 @@ spec:
             fi
           done
 
+          #gateway join subnets
+          gateway_v4_join_subnet=
+          gateway_v6_join_subnet=
+          V4JoinSubnet={{.V4JoinSubnet}}
+          V6JoinSubnet={{.V6JoinSubnet}}
+          if [[ -n $V4JoinSubnet ]]; then
+            gateway_v4_join_subnet="--gateway-v4-join-subnet=${V4JoinSubnet}"
```
danwinship, line 765: I had thought we were trying to do `ovnkube.conf` options now rather than command-line arguments? Did we change strategies at some point?

manifests/0000_70_cluster-network-operator_01_crd.yaml
```diff
@@ -19,111 +19,155 @@ spec:
   - name: v1
     schema:
       openAPIV3Schema:
-        description: "Network describes the cluster's desired network configuration. It is consumed by the cluster-network-operator. \n Compatibility level 1: Stable within a major release for a minimum of 12 months or 3 minor releases (whichever is longer)."
+        description: "Network describes the cluster's desired network configuration.
+          It is consumed by the cluster-network-operator. \n Compatibility level 1:
+          Stable within a major release for a minimum of 12 months or 3 minor releases
+          (whichever is longer)."
```
danwinship, line 25: This file seems to have gotten line-wrapped but none of the actual API changed? What caused that? (More importantly, "is it going to accidentally get flipped back the next time someone changes something"?)

vendor/github.com/openshift/api/operator/v1/0000_70_cluster-network-operator_01.crd.yaml-patch
```diff
@@ -1,3 +0,0 @@
-- op: add
-  path: /spec/versions/name=v1/schema/openAPIV3Schema/properties/spec/properties/defaultNetwork/properties/ovnKubernetesConfig/properties/policyAuditConfig/default
-  value: {}
```
danwinship, line 3: why did this go away?

abhat:
Does this need to be exposed at all? Can't we figure out an appropriate v4 and v6 join subnet to use based on the other CIDRs (service, cluster, node subnets) that they provide in the config?

vendor/github.com/openshift/api/operator/v1/0000_70_cluster-network-operator_01.crd.yaml-patch
pliurh: https://github.com/openshift/api/pull/1102

pliurh:
> Does this need to be exposed at all? Can't we figure out an appropriate v4 and v6 join subnet to use based on the other CIDRs (service, cluster, node subnets) that they provide in the config?

I don't think we can make a good guess with all the information we could get, due to:
1. The join subnet size can limit the size of the cluster. We don't know how big the cluster could be scaled up.
2. We don't know what IP ranges could be occupied by users' current networks that need to communicate with our cluster. 
So I think it would be better to expose the knob to users since they should know it better than we do.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
pliurh: I am not aware of the strategy. I can refactor the patch.

pliurh:
@abhat @danwinship I refactored the PR. PTAL.

pkg/network/ovn_kubernetes.go
```diff
@@ -669,6 +671,18 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 		if oc.GenevePort != nil && (*oc.GenevePort < 1 || *oc.GenevePort > 65535) {
 			out = append(out, errors.Errorf("invalid GenevePort %d", *oc.GenevePort))
 		}
+		if oc.V4InternalSubnet != "" {
```
zshi-redhat, line 674: In the api doc, it says:
```
It must not overlap with any other subnet being used by OpenShift or by the node network. The size of the subnet must be larger than the number of nodes
```
should we validate the above or is it already validated somewhere else?

zshi-redhat:
> > Does this need to be exposed at all? Can't we figure out an appropriate v4 and v6 join subnet to use based on the other CIDRs (service, cluster, node subnets) that they provide in the config?
> 
> I don't think we can make a good guess with all the information we could get, due to:
> 
> 1. The join subnet size can limit the size of the cluster. We don't know how big the cluster could be scaled up.
> 2. We don't know what IP ranges could be occupied by users' current networks that need to communicate with our cluster.
> 
> So I think it would be better to expose the knob to users since they should know it better than we do.

I suppose end user would also want to know how big the subnet should be configured since this is not a change that can be made during runtime. they will have to decide the subnet beforehand (before the cluster is deployed or scaled)?

pliurh:
> I suppose end user would also want to know how big the subnet should be configured based on expected cluster size, number of nodes/pods since this is not a change that can be made during runtime. they will have to decide the subnet beforehand (before the cluster is deployed or scaled)?

Yeah, I think we can put those guides in the document. The join-network shall be big enough to contain one IP address per-node + one IP for ovn cluster router.

pkg/network/ovn_kubernetes.go
pliurh: I don't think we can validate the size of the cluster, as it's extendable after installation. But I can add the overlapping validation.

zshi-redhat:
> > I suppose end user would also want to know how big the subnet should be configured based on expected cluster size, number of nodes/pods since this is not a change that can be made during runtime. they will have to decide the subnet beforehand (before the cluster is deployed or scaled)?
> 
> Yeah, I think we can put those guides in the document. The join-network shall be big enough to contain one IP address per-node + one IP for ovn cluster router.

It sounds like the size of join network can be deduced from cluster-subnets.
Given the size of subnet (default to /24) allocated to each node and the cluster-subnets, we shall be able to tell how many nodes can be added in the cluster, right?

pliurh:
> It sounds like the size of join network can be deduced from cluster-subnets.
Given the size of subnet (default to /24) allocated to each node and the cluster-subnets, we shall be able to tell how many nodes can be added in the cluster, right?

Added this to validation, PTAL.

PR #1509, opened 2022-07-18
Bug 2103680: Revert "controller/clusterconfig: Take advantage of patch."

kyrtapz:
`DisableNetworkDiagnostics` is a bool variable(not a pointer) which means we were always setting it to false.

Revert `b2cf367fd3c8c3e676808e386c826f0abae1dbf6` for now, I think we might need to update the API and deprecate `DisableNetworkDiagnostics` field for something new that can be unset.
 
This reverts commit b2cf367fd3c8c3e676808e386c826f0abae1dbf6.

### DIFF
--- a/pkg/controller/clusterconfig/clusterconfig_controller.go
+++ b/pkg/controller/clusterconfig/clusterconfig_controller.go
@@ -6,7 +6,6 @@ import (
 	"log"
 
 	configv1 "github.com/openshift/api/config/v1"
-	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/apply"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
@@ -14,7 +13,6 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/network"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


andreaskaris:
Have you looked into https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/#server-side-apply-support-in-client-go instead of reverting that? It seems to present a solution to the issue of datatypes with non-pointers.

andreaskaris:
Something like this might be worth a try (obviously with added package names etc.) to be backwards compatible with
our current API while still keeping the Server Side Apply feature:

* create your own ApplyNetworkSpec:

  type ApplyNetwork struct {
      metav1.TypeMeta   `json:",inline"`
      metav1.ObjectMeta `json:"metadata,omitempty"`
  
      Spec   ApplyNetworkSpec   `json:"spec,omitempty"`
      Status NetworkStatus `json:"status,omitempty"`
  }

  type ApplyNetworkSpec struct {
      OperatorSpec `json:",inline"`
      ClusterNetwork []ClusterNetworkEntry `json:"clusterNetwork"`
      ServiceNetwork []string `json:"serviceNetwork"`
      DefaultNetwork DefaultNetworkDefinition `json:"defaultNetwork"`
      AdditionalNetworks []AdditionalNetworkDefinition `json:"additionalNetworks,omitempty"`
      DisableMultiNetwork *bool `json:"disableMultiNetwork,omitempty"`
      UseMultiNetworkPolicy *bool `json:"useMultiNetworkPolicy,omitempty"`
      DeployKubeProxy *bool `json:"deployKubeProxy,omitempty"`
      DisableNetworkDiagnostics *bool `json:"disableNetworkDiagnostics"`  // with this changed to *bool
      KubeProxyConfig *ProxyConfig `json:"kubeProxyConfig,omitempty"`
      ExportNetworkFlows *ExportNetworkFlows `json:"exportNetworkFlows,omitempty"`
      Migration *NetworkMigration `json:"migration,omitempty"`
  }

Create a copy of pkg/network/cluster_config.go 
`MergeClusterConfig(operConf *operv1.NetworkSpec, clusterConf configv1.NetworkSpec)` to merge the
above type and call:

      operConfig := ApplyNetwork{
          TypeMeta:   metav1.TypeMeta{APIVersion: operv1.GroupVersion.String(), Kind: "Network"},
          ObjectMeta: metav1.ObjectMeta{Name: names.OPERATOR_CONFIG},
      } 
      network.MergeApplyClusterConfig(&operConfig.Spec, clusterConfig.Spec)

kyrtapz:
> Have you looked into https://kubernetes.io/blog/2021/08/06/server-side-apply-ga/#server-side-apply-support-in-client-go instead of reverting that? It seems to present a solution to the issue of datatypes with non-pointers.
> 
> So a solution would be to create a new "applyconfiguration" struct which is exactly the same as the NetworkSpec, but with a pointer instead of DisableNetworkDiagnostics. You could then apply that one.
> 
> So, instead of deprecating the field of the NetworkSpec (and thus causing a user facing change), we could add an applyconfiguration for the NetworkSpec. Unfortunately, only for that one single field because everything else is defined as pointers ... But probably still better than a user facing change.

That looks very promising, I will give it a go!

PR #1519, opened 2022-08-02
Add controller to synchronize the API and Ingress VIP fields

creydr:
For the dual stack API & Ingress VIP support, we have to add additional fields to onprem platform statuses (https://github.com/openshift/api/pull/1152). 
To have consistent APIs between versions and keep backwards compatible for components consuming the old VIP fields, we need to keep the new and old VIP fields in sync. This is addressed by this PR by adding a new controller to the CNO which keeps the old and new VIP fields in `Infrastructure.config.openshift.io/cluster` in sync.

**Special notes for reviewers:**
* we added an `OWNERS` file in the folder with adding the bare metal team as reviewers for this controller, as it could make sense to have one of the team members reviewing changes on this component.

### DIFF
--- a/pkg/controller/infrastructureconfig/infrastructureconfig_controller.go
+++ b/pkg/controller/infrastructureconfig/infrastructureconfig_controller.go
@@ -1,0 +1,108 @@
+package infrastructureconfig
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"reflect"
+
+	configv1 "github.com/openshift/api/config/v1"
+	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
+	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"github.com/openshift/cluster-network-operator/pkg/names"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const ControllerName = "infrastructureconfig"
+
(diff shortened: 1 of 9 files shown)


creydr:
/retitle Add controller to synchronize the API and Ingress VIP fields
/hold cancel as https://github.com/openshift/api/pull/1152 merged and we revendored openshift/api in 10f2629f67737dc0a06d89b79986190470ff988c

creydr:
/hold cancel 
as https://github.com/openshift/api/pull/1152 merged and we revendored openshift/api in 10f2629f67737dc0a06d89b79986190470ff988c

pkg/controller/infrastructureconfig/infrastructureconfig_controller.go
```diff
@@ -0,0 +1,105 @@
...
+		err = fmt.Errorf("Error while reading infrastructures.%s/cluster: %w", configv1.GroupName, err)
+		log.Println(err)
+		return reconcile.Result{}, err
+	}
+
+	// Sync API & Ingress VIPs
+	updatedInfraConfig := r.apiAndIngressVIPsSyncer.VipsSynchronize(infraConfig)
+
+	if !reflect.DeepEqual(updatedInfraConfig.Status, infraConfig.Status) {
+		err = r.client.Default().CRClient().Status().Update(ctx, updatedInfraConfig, &client.UpdateOptions{})
```
deads2k, line 93: This will work better as a server-side-apply to avoid clobbering unknown fields as schemas evolve: see https://github.com/openshift/cluster-etcd-operator/pull/833/files#r941800222 as an example.

pkg/controller/infrastructureconfig/sync_vips.go
```diff
@@ -0,0 +1,107 @@
...
+		return updatedInfraConfig
+	}
+
+	if err := syncVIPs(apiVIPs, apiVIP); err != nil {
+		log.Printf("Error on syncing api VIP fields: %v", err) // this is only a warning -> continue
+	}
+
+	if err := syncVIPs(ingressVIPs, ingressVIP); err != nil {
+		log.Printf("Error on syncing ingress VIP fields: %v", err) // this is only a warning -> continue
+	}
```
jcaamano, line 68: nit: if this is a warning, I would probably choose a different wording other than `Error` for this log.

creydr: :+1: updated it

PR #1521, opened 2022-08-04
Render CRDs for both OSDN and OVNK during migration

pliurh:
During SDN migration, we need to support custom resource translation
for features like egressIP and egressFirewall. So we need the CRDs
of both CNIs to be rendered.

### DIFF
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -62,6 +62,16 @@ func Render(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult
 	}
 	objs = append(objs, o...)
 
+	if conf.Migration != nil && conf.Migration.NetworkType != "" {
+		// During SDN Migration, CNO needs to convert the custom resources of
+		// egressIP, egressFirewall, etc. Therefore we need to render the CRDs for
+		// both OpenShiftSDN and OVNKubernetes.
+		o, err = renderCRDForMigration(conf, manifestDir)
+		if err != nil {
+			return nil, progressing, err
+		}
+		objs = append(objs, o...)
+	}
 	// render kube-proxy
 	// DPU_DEV_PREVIEW
 	// There is currently a restriction that renderStandaloneKubeProxy() is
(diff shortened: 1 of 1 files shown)


pkg/network/render.go
```diff
@@ -62,6 +62,13 @@ func Render(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult
 	}
 	objs = append(objs, o...)
 
+	if conf.Migration != nil {
+		o, progressing, err = renderCRDForMigration(conf, manifestDir)
```
kyrtapz, line 66: could you add a comment explaining why we need to render the new CRDs during migration?

pliurh: fixed

kyrtapz: thanks! the comment looks good.
Can you move the comment to the first commit and extract egress firewall migration to a separate PR? I think it would be good to get some more eyes on that.

pliurh: Done.

pkg/network/render.go
```diff
@@ -496,6 +506,29 @@ func renderDefaultNetwork(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.B
 	}
 }
 
+func renderCRDForMigration(conf *operv1.NetworkSpec, manifestDir string) ([]*uns.Unstructured, bool, error) {
+	migration := conf.Migration
+	progressing := false
```
jcaamano, line 511: `progressing` is always false? I see other render methods in this file that don't return any `progressing` so basically it does not seem like we are complying with an interface or harmonizing this render method with the other ones.  So I wonder why we have it.

jcaamano (commented):
Question:
are we planning to leave the old CRDs around after migration? A comment in the commit message on how this has been considered would be welcomed.

pkg/network/render.go
pliurh: Yes, you're right. I removed it.

pliurh:
> Question: are we planning to leave the old CRDs around after migration? A comment in the commit message on how this has been considered would be welcomed.

After migration, the `spec.migration` field shall be removed. Therefore, the old CRDs will no longer be rendered and removed by CNO.

PR #1524, opened 2022-08-09
Bug 2116982: multus-admission-controller SNO number of replicas

omertuc:
Fixes bug introduced by ee502976611d06df811054d93a11936f54c1c18d

The number of replicas for the multus admission controller should be
determined according to the control-plane or infrastructure topology of
the cluster instead of being hard-coded 1.

This is because on SNO clusters, we want core component deployments to
have a smaller foot-print, at the cost of lowered availability. We have
an e2e test that makes sure all `openshift.*` namespace Deployments and
StatefulSets have 1 replicas when the topologies indicate SNO.

See `Test_getMultusAdmissionControllerReplicas` for exactly how many
replicas are chosen depending on the various combinations of
`ControlPlaneTopology`/`InfrastructureTopology` values.


pkg/bootstrap/types.go
```diff
@@ -73,10 +73,11 @@ type BootstrapResult struct {
 }
 
 type InfraStatus struct {
-	PlatformType         configv1.PlatformType
-	PlatformRegion       string
-	PlatformStatus       *configv1.PlatformStatus
-	ExternalControlPlane bool
```
omertuc, line 79: I felt it was best to delete `ExternalControlPlane` (and its test `pkg/platform/platform_test.go`) because it would in a way contain duplicate information that's already carried by the newly added `ControlPlaneTopology` and `InfrastructureTopology` struct fields.

I'm willing to undo this change, if required

pkg/network/render.go
```diff
@@ -600,16 +601,30 @@ func renderAdditionalNetworks(conf *operv1.NetworkSpec, manifestDir string) ([]*
...
 
+func getMultusAdmissionControllerReplicas(bootstrapResult *bootstrap.BootstrapResult) int {
+	replicas := 2
+	if bootstrapResult.Infra.ControlPlaneTopology == configv1.ExternalTopologyMode {
+		if bootstrapResult.Infra.InfrastructureTopology == configv1.SingleReplicaTopologyMode {
+			replicas = 1
+		}
+	} else if bootstrapResult.Infra.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
+		replicas = 1
+	}
```
eranco74, line 612:
```suggestion
	if bootstrapResult.Infra.ControlPlaneTopology == configv1.SingleReplicaTopologyMode {
		replicas = 1
	}
```

omertuc: When control plane is external it gets scheduled on the workers, so we should look at the infra topology

https://github.com/openshift/cluster-network-operator//blob/3d24e5dfe62c67acd8243f16169ae0d19db5e3ff/bindata/network/multus-admission-controller/admission-controller.yaml#L82-L85

eranco74: :man_facepalming: 
I didn't notice the difference, thanks.

pkg/platform/platform.go
```diff
@@ -27,15 +27,16 @@ func InfraStatus(client cnoclient.Client) (*bootstrap.InfraStatus, error) {
...
+		PlatformStatus:         infraConfig.Status.PlatformStatus,
+		ControlPlaneTopology:   infraConfig.Status.ControlPlaneTopology,
+		InfrastructureTopology: infraConfig.Status.InfrastructureTopology,
+		APIServers:             map[string]bootstrap.APIServer{},
 	}
 
 	// If we use hypershift, the components need proxy settings. In selfhosted, they
 	// are always local to the API so there it is never needed.
-	if res.ExternalControlPlane {
+	if res.ControlPlaneTopology == configv1.ExternalTopologyMode {
```
dougbtv, line 39: Is this why we're deleting the platform_test.go?

omertuc: See https://github.com/openshift/cluster-network-operator/pull/1524#discussion_r941638270

dougbtv: Otherwise, I like the change. Thanks for the patch. OK to move forward if it is indeed correct to be removing `platform_test.go`

dougbtv: Thanks, seems OK to me.

PR #1525, opened 2022-08-11
HyperShift: Move CNCC to the controll-plane namespace

kyrtapz:
When running in HyperShift deploy CNCC in the management cluster.


pkg/network/cloud_network.go
```diff
@@ -1,6 +1,8 @@
 package network
 
 import (
+	"github.com/openshift/cluster-network-operator/pkg/client"
```
alvaroaleman, line 4: Tiny nit: Maybe use a different import group?

alvaroaleman (commented):
The change looks good, I didn't actually test it out though

pkg/network/cloud_network.go
```diff
@@ -38,18 +41,27 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, cloudBootstrap
 	data.Data["PlatformTypeAzure"] = v1.AzurePlatformType
 	data.Data["PlatformTypeGCP"] = v1.GCPPlatformType
 	data.Data["CloudNetworkConfigControllerImage"] = os.Getenv("CLOUD_NETWORK_CONFIG_CONTROLLER_IMAGE")
-	data.Data["KubernetesServiceHost"] = cloudBootstrapResult.APIServers[bootstrap.APIServerDefault].Host
-	data.Data["KubernetesServicePort"] = cloudBootstrapResult.APIServers[bootstrap.APIServerDefault].Port
+	data.Data["KubernetesServiceHost"] = cloudBootstrapResult.APIServers[bootstrap.APIServerDefaultLocal].Host
```
zshi-redhat, line 44: @kyrtapz  I forgot the difference between APIServerDefaultLocal and APIServerDefault, do we need to distinguish it for self-hosted and managed?

kyrtapz: this variable should do exactly that: https://github.com/openshift/cluster-network-operator/blob/3b3909bd4f64390fc2b0c80b1be92ff6370209a5/pkg/bootstrap/types.go#L105-L108

kyrtapz: Yes, that is correct

PR #1526, opened 2022-08-15
OVN-K alerts: first tranche

martinkennelly:
See commits for details.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/alert-rules.yaml
+++ b/bindata/network/ovn-kubernetes/common/alert-rules.yaml
@@ -18,8 +18,7 @@ spec:
         runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/NodeWithoutOVNKubeNodePodRunning.md
         description: |
           Networking is degraded on nodes that do not have a functioning ovnkube-node pod. Existing workloads on the
-          node may continue to have connectivity but any additional workloads will not be provisioned on the node. Any
-          network policy changes will not be implemented on existing workloads on the node.
+          node may continue to have connectivity but any changes to the networking control plane will not be implemented.
       expr: |
         (kube_node_info unless on(node) (kube_pod_info{namespace="openshift-ovn-kubernetes",pod=~"ovnkube-node.*"}
         or kube_node_labels{label_kubernetes_io_os="windows"})) > 0
@@ -34,5 +33,31 @@ spec:
           Networking is degraded on nodes when OVN controller is not connected to OVN southbound database connection. No networking control plane updates will be applied to the node.
       expr: |
         max_over_time(ovn_controller_southbound_database_connected[5m]) == 0
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNodePodAddError
+      annotations:
+        summary: OVN Kubernetes is experiencing pod creation errors at an elevated rate.
+        description: OVN Kubernetes experiences pod creation errors at an elevated rate. The
+          pods will be retried.
+      expr: |
+        (sum by(instance, namespace) (rate(ovnkube_node_cni_request_duration_seconds_count{command="ADD",err="true"}[5m]))
+          /
+        sum by(instance, namespace) (rate(ovnkube_node_cni_request_duration_seconds_count{command="ADD"}[5m])))
+        > 0.1
+      for: 15m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNodePodDeleteError
+      annotations:
+        summary: OVN Kubernetes experiencing pod deletion errors at an elevated rate.
+        description: OVN Kubernetes experiences pod deletion errors at an elevated rate. The
+          pods will be retried.
+      expr: |
+        (sum by(instance, namespace) (rate(ovnkube_node_cni_request_duration_seconds_count{command="DEL",err="true"}[5m]))
+          /
+        sum by(instance, namespace) (rate(ovnkube_node_cni_request_duration_seconds_count{command="DEL"}[5m])))
+        > 0.1
+      for: 15m
       labels:
         severity: warning
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -325,7 +325,7 @@ spec:
     # OVN northbound and southbound database performance alerts
     - alert: OVNKubernetesNorthboundDatabaseCPUUsageHigh
       annotations:
-        summary: OVN northbound database {{ $labels.instance }} is greater than {{ $value }} percent CPU usage for a period of time.
+        summary: OVN northbound database {{"{{"}} $labels.instance {{"}}"}} is greater than {{"{{"}} $value | humanizePercentage {{"}}"}} percent CPU usage for a period of time.
         description: High OVN northbound CPU usage indicates high load on the networking control plane.
       expr: (sum(rate(container_cpu_usage_seconds_total{container="nbdb"}[5m])) BY (instance, name, namespace)) > 0.8
       for: 15m
@@ -333,7 +333,7 @@ spec:
         severity: info
     - alert: OVNKubernetesSouthboundDatabaseCPUUsageHigh
       annotations:
-        summary: OVN southbound database {{ $labels.instance }} is greater than {{ $value }} percent CPU usage for a period of time.
+        summary: OVN southbound database {{"{{"}} $labels.instance {{"}}"}} is greater than {{"{{"}} $value | humanizePercentage {{"}}"}} percent CPU usage for a period of time.
         description: High OVN southbound CPU usage indicates high load on the networking control plane.
       expr: (sum(rate(container_cpu_usage_seconds_total{container="sbdb"}[5m])) BY (instance, name, namespace)) > 0.8
       for: 15m
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -210,3 +210,115 @@ spec:
       labels:
         namespace: {{.HostedClusterNamespace}}
         severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseInboundConnectionError
+      annotations:
+        summary: OVN northbound database server(s) is experiencing inbound RAFT connectivity errors.
+        description: OVN northbound database server(s) is experiencing inbound RAFT connectivity errors which may indicate degraded OVN
+          database high availability.
+      expr: |
+        # Without min_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        # ..error_total is set to zero when error resolves itself
+        min_over_time(ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Northbound"}[5m]) > 0
+      for: 5m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesSouthboundDatabaseInboundConnectionError
+      annotations:
+        summary: OVN southbound database server(s) is experiencing inbound RAFT connectivity errors.
+        description: OVN southbound database server(s) is experiencing inbound RAFT connectivity errors which may indicate degraded OVN
+          database high availability.
+      expr: |
+        # Without min_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        # ..error_total is set to zero when error resolves itself
+        min_over_time(ovn_db_cluster_inbound_connections_error_total{db_name="OVN_Southbound"}[5m]) > 0
+      for: 5m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseOutboundConnectionError
+      annotations:
+        summary: OVN northbound database server(s) is experiencing outbound RAFT connectivity errors.
+        description: OVN northbound database server(s) outbound RAFT connectivity errors may indicate degraded OVN
+          database high availability.
+      expr: |
+        # Without min_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        # ..error_total is set to zero when error resolves itself
+        min_over_time(ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Northbound"}[5m]) > 0
+      for: 5m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesSouthboundDatabaseOutboundConnectionError
+      annotations:
+        summary: OVN southbound database server(s) is experiencing outbound RAFT connectivity errors.
+        description: OVN southbound database server(s) outbound RAFT connectivity errors which may indicate degraded OVN
+          database high availability.
+      expr: |
+        # Without min_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        # ..error_total is set to zero when error resolves itself
+        min_over_time(ovn_db_cluster_outbound_connections_error_total{db_name="OVN_Southbound"}[5m]) > 0
+      for: 5m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseInboundConnectionMissing
+      annotations:
+        summary: OVN northbound database server(s) do not have expected number of inbound RAFT connections.
+        description: OVN northbound database server(s) do not have expected number of inbound connections for a RAFT cluster
+          which may indicate degraded OVN database high availability.
+      expr: |
+        # Expected sum of inbound connections is number of control plane nodes * number of control plane nodes minus one
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        sum(max_over_time(ovn_db_cluster_inbound_connections_total{db_name="OVN_Northbound"}[5m]))
+        != count(max_over_time(kube_node_role{role="master"}[5m])) * (count(max_over_time(kube_node_role{role="master"}[5m]))-1)
+      for: 5m
+      labels:
+        namespace: {{.HostedClusterNamespace}}
+        severity: warning
+    - alert: OVNKubernetesSouthboundDatabaseInboundConnectionMissing
+      annotations:
+        summary: OVN southbound database server(s) do not have expected number of inbound RAFT connections.
+        description: OVN southbound database server(s) do not have expected number of inbound connections for a RAFT cluster
+          which may indicate degraded OVN database high availability.
+      expr: |
+        # Expected sum of inbound connections is number of control plane nodes * number of control plane nodes minus one
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        sum(max_over_time(ovn_db_cluster_inbound_connections_total{db_name="OVN_Southbound"}[5m]))
+        != count(max_over_time(kube_node_role{role="master"}[5m])) * (count(max_over_time(kube_node_role{role="master"}[5m]))-1)
+      for: 5m
+      labels:
+        namespace: {{.HostedClusterNamespace}}
+        severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseOutboundConnectionMissing
+      annotations:
+        summary: OVN northbound database server(s) do not have expected number of outbound RAFT connections.
+        description: OVN northbound database server(s) do not have expected number of outbound connections for a RAFT cluster
+          which may indicate degraded OVN database high availability.
+      expr: |
+        # Expected sum of outbound connections is number of control plane nodes * number of control plane nodes minus one
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        sum(max_over_time(ovn_db_cluster_outbound_connections_total{db_name="OVN_Northbound"}[5m]))
+        != count(max_over_time(kube_node_role{role="master"}[5m])) * (count(max_over_time(kube_node_role{role="master"}[5m]))-1)
+      for: 5m
+      labels:
+        namespace: {{.HostedClusterNamespace}}
+        severity: warning
+    - alert: OVNKubernetesSouthboundDatabaseOutboundConnectionMissing
+      annotations:
+        summary: OVN southbound database server(s) do not have expected number of outbound RAFT connections.
+        description: OVN southbound database server(s) do not have expected number of outbound connections for a RAFT cluster
+          which may indicate degraded OVN database high availability.
+      expr: |
+        # Expected sum of outbound connections is number of control plane nodes * number of control plane nodes minus one
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        sum(max_over_time(ovn_db_cluster_outbound_connections_total{db_name="OVN_Southbound"}[5m]))
+        != count(max_over_time(kube_node_role{role="master"}[5m])) * (count(max_over_time(kube_node_role{role="master"}[5m]))-1)
+      for: 5m
+      labels:
+        namespace: {{.HostedClusterNamespace}}
+        severity: warning
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -140,3 +140,45 @@ spec:
       for: 25m
       labels:
         severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseLeaderError
+      annotations:
+        summary: OVN northbound database(s) have no RAFT leader
+        description: OVN northbound database(s) have no RAFT leader. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        count(max_over_time(ovn_db_cluster_server_role{db_name="OVN_Northbound", server_role="leader"}[5m])) by (namespace) == 0
+      labels:
+        severity: critical
+    - alert: OVNKubernetesSouthboundDatabaseLeaderError
+      annotations:
+        summary: OVN southbound database(s) have no RAFT leader
+        description: OVN southbound database(s) have no leader. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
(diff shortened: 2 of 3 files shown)


martinkennelly:
Please do not merge this until monitoring team have review it.

martinkennelly:
@dcbw Are you interested in taking a look at this?

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -183,3 +183,29 @@ spec:
       for: 15m
       labels:
         severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseClusterMemberError
+      annotations:
+        summary: OVN northbound database server(s) has not been a member of the databases high availability for a period of time.
```
npinaeva: high availability cluster?

npinaeva (commented):
I can't verify if prometheus queries are correct, but alerts descriptions look good, and I tried all the queries on a test cluster, the values are as expected.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -209,3 +209,111 @@ spec:
...
+      annotations:
+        summary: OVN southbound database server(s) do not have expected number of outbound connections.
+        description: OVN southbound database server(s) do not have expected number of outbound connections which may
+          indicate degraded OVN database high availability.
+      expr: |
+        # Expected sum of outbound connections is number of control plane nodes * number of control plane nodes minus one
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        sum(max_over_time(ovn_db_cluster_outbound_connections_total{db_name="OVN_Southbound"}[5m]))
+        != count(max_over_time(kube_node_role{role="master"}[5m])) * (count(max_over_time(kube_node_role{role="master"}[5m]))-1)
```
trozet: is this right? What are these outbound connections to? If it's SBDB connections on 9642 it would be to northd, as well as all ovn-controllers, and ovnkube-masters. If its raft it would just be 2 right?

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
```diff
@@ -328,3 +326,44 @@ spec:
...
       labels:
         severity: critical
+    # OVN kubernetes node functional alerts
+    - alert: NodeWithoutOVNKubeNodePodRunning
+      annotations:
+        summary: All Linux nodes should be running an ovnkube-node pod, {{"{{"}} $labels.node {{"}}"}} is not.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/NodeWithoutOVNKubeNodePodRunning.md
+        description: |
+          Networking is degraded on nodes that do not have a functioning ovnkube-node pod. Existing workloads on the
+          node may continue to have connectivity but any additional workloads will not be provisioned on the node. Any
```
trozet, line 362: The network policy stuff seems too specific. You can just say "Networking on this node may function abnormally until the node is restored"

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -329,3 +328,12 @@ spec:
       for: 10m
       labels:
         severity: critical
+    # OVN northd performance alerts
+    - alert: OVNKubernetesNorthdCPUUsageHigh
+      annotations:
+        summary: OVN northd {{ $labels.instance }} is greater than {{ $value }} percent CPU usage for a period of time.
+        description: High OVN northd CPU usage indicates high load on the networking control plane.
+      expr: (sum(rate(container_cpu_usage_seconds_total{container="northd"}[5m])) BY (instance, name) * 100) > 80
```
trozet, line 353: so over 15 minutes the northd CPU should not average over 80% cpu? @dceara seem reasonable?

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -316,6 +316,23 @@ spec:
...
+        description: High OVN northbound CPU usage indicates high load on the networking control plane.
+      expr: (sum(rate(container_cpu_usage_seconds_total{container="nbdb"}[5m])) BY (instance, name) * 100) > 80
+      for: 15m
+      labels:
+        severity: info
+    - alert: OVNKubernetesSouthboundDatabaseCPUUsageHigh
+      annotations:
+        summary: OVN southbound database {{ $labels.instance }} is greater than {{ $value }} percent CPU usage for a period of time.
+        description: High OVN southbound CPU usage indicates high load on the networking control plane.
+      expr: (sum(rate(container_cpu_usage_seconds_total{container="sbdb"}[5m])) BY (instance, name) * 100) > 80
```
trozet, line 332: 80% seems safe, in my experience anything over 70% is a warning sign SBDB is in trouble

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -316,6 +316,23 @@ spec:
       for: 15m
       labels:
         severity: warning
+    # OVN northbound and southbound database performance alerts
+    - alert: OVNKubernetesNorthboundDatabaseCPUUsageHigh
+      annotations:
+        summary: OVN northbound database {{ $labels.instance }} is greater than {{ $value }} percent CPU usage for a period of time.
+        description: High OVN northbound CPU usage indicates high load on the networking control plane.
+      expr: (sum(rate(container_cpu_usage_seconds_total{container="nbdb"}[5m])) BY (instance, name) * 100) > 80
```
trozet, line 324: NBDB cpu should always be pretty low. Definitely less than 50%. @dceara do you agree? @mohit-sheth you may have some data to back this up.

trozet (commented):
I don't really have the experience to evaluate the expressions, but left some general feedback questions about the alerts.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -117,3 +117,25 @@ spec:
...
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        max(max_over_time(ovn_db_cluster_term{db_name="OVN_Northbound"}[5m])) - min(max_over_time(ovn_db_cluster_term{db_name="OVN_Northbound"}[5m])) > 0
+      for: 25m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesSouthboundDatabaseTermLag
+      annotations:
+        summary: OVN southbound databases RAFT term have not been equal for a period of time.
```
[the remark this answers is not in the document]
martinkennelly: Its a RAFT term. A term indicates a period of time. It increments. A new term value is created when there is a new candidate attempting to be a leader of the RAFT cluster. The winning candidate sets the term value and the followers, if they accept it, also adopt this term value. Each of the RAFT cluster members store this term value and it helps them detect stale information.
This alert, will alert if there are any cluster members who are persistently not agreeing on the term number. It can help detect split brain.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -139,3 +139,47 @@ spec:
...
         severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseLeaderError
+      annotations:
+        summary: OVN northbound database(s) have no RAFT leader
+        description: OVN northbound database(s) have no RAFT leader. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        count(max_over_time(ovn_db_cluster_server_role{db_name="OVN_Northbound", server_role="leader"}[5m])) == 0
+      for: 15m
```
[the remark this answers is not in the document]
martinkennelly: It whens if the promql expression is true for 15 minutes it will fire an alert.
Because the promql expression looks back 5 minutes, it means the total time to fire will be 20 minutes of continuous time with no leader seen.
I picked this time to detect permanently broken leader election. 
Do you want to reduce it? It does seem too long on second thought.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
martinkennelly: Its RAFT. 2 is correct. I count the number of leaders and subtract one.

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
martinkennelly: Its confusing alright. I didnt mean k8 network policy, but network control plane policy.

Changed to:

```
          Networking is degraded on nodes that do not have a functioning ovnkube-node pod. Existing workloads on the
          node may continue to have connectivity but any changes to the networking control plane will not be implemented.
```

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Over 20 minutes actually because of the expression.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
igsilya: How frequently these conditions are checked?  I see that interval here is 25 minutes (+5 ?), during that time databases may transfer leadership up to 9 times, that will trigger re-elections and term changes.  Can this cause false-positive alerts?

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
dceara: I agree.  NBDB cpu should be pretty low in most cases.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
igsilya: I don't think that description is correct here. The problem is that northd almost always re-computes the whole network topology every time even small change is made.  On a high-scale cluster it may take several seconds per one run.  Assuming some changes are made in the high-scale cluster every few seconds, northd may stay at 100% CPU usage for a really long time.  And it will take almost the same time for it to process a very small change and a really big one.  So, the alert doesn't really reflect any issues with northd or other components, especially at scale.  And it may produce a lot of false-positives.

dceara: It really depends on the workload/cluster configuration.  I'm curious if @mohit-sheth has some ovn-northd CPU usage data from recent node-density/density-cni runs with larger scale clusters (>= 120 nodes).

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Thats correct. On the scale cluster I got, I rarely saw nb db go over 30%. This NB DB alert is a bit pointless but  I added it anyway in case some issue could cause it to utilise a lot of CPU and this alert would capture it.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Accept the description is incorrect.
I tested it with nodes > 120 after scale testing runs.
I think you raise a reasonable point though that this needs better testing.
I will remove the CPU usage from this PR and introduce it later following more tests.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
martinkennelly: Either every 30s or 1m. I cannot remember which.
I checked it on a large cluster > 100 nodes after multiple scale runs and it was ok.
I artificially induce leadership transfer 20 times over 30 minutes and check the results.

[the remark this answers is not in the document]
trozet: should the summary be "OVN Southbound DB Raft Term not incrementing over time" ?

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
trozet: I would say if this is true for even over a minute it should alert. If there is no leader then it represents some kind of outage or problem between raft nodes that should never happen. I think under normal conditions, raft leadership will always be transferred, or if an election timer goes off it should take only a few seconds max to find a new leader. I'll defer to @igsilya to what he thinks would be a reasonable value here.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
trozet: should the summary then indicate these are raft connections? 
OVN southbound database server(s) do not have expected number of outbound raft connections

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
martinkennelly: I think my summary is better. We want to detect if term values haven't been the same for a period of time and not that they are incrementing over time.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
martinkennelly: I reduced it to five minutes. We have to consider data is scared every 30s and evaluated every 1m. I am uncomfortable having an alert lower than 5 minutes for this without a lot of testing.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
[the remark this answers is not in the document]
martinkennelly: I performed this test by repeatedly deleting the SB DBs RAFT leader container. 
I saw the leader changing and the term becoming consistent between all the new followers within a max time of 30 seconds, usually it was quicker.
The alert promql expression didn't become true during the entire event.
If it did become true, It would need to hold true at least another 30 times in a row.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -0,0 +1,346 @@
...
+    - alert: OVNKubernetesNorthboundDatabaseLeaderError
+      annotations:
+        summary: OVN northbound database(s) have no RAFT leader
+        description: OVN northbound database(s) have no RAFT leader. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        count(max_over_time(ovn_db_cluster_server_role{db_name="OVN_Northbound", server_role="leader"}[5m])) == 0
+      labels:
+        severity: warning
```
bmeng, line 151: It should elect new leader immediately if the existing leader is in trouble. So it sounds like not a warning level.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -0,0 +1,346 @@
...
+        severity: info
+    # OVN northd functional alerts
+    - alert: OVNKubernetesNorthdInactive
+      annotations:
+        summary: No OVN northd is active. Networking is degraded.
+        description: No OVN northd is available within the high availability set. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        count(max_over_time(ovn_northd_status[5m]) == 1) != 1
```
bmeng, line 343: What does the metric `ovn_northd_status` mean here? In my cluster, I can see 3 of them and 1 has value 1, the others are 0.

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
```diff
@@ -0,0 +1,383 @@
...
+        description: No OVN northd is available within the high availability set. Networking control plane is degraded.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        count(max_over_time(ovn_northd_status[5m]) == 1) != 1
+      for: 5m
+      labels:
+        severity: critical
+    # OVN kubernetes node functional alerts
+    - alert: NodeWithoutOVNKubeNodePodRunning
```
bmeng, line 345: The alerts below will be deployed to hosted clusters only?
Seems at least this one `NodeWithoutOVNKubeNodePodRunning` is also useful for management cluster.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Specifies whether this instance of ovn-northd is standby(0) or active(1) or paused(2). Active holds the lock to write to SB DB. Standbys continue to get updates from NB DB but do not write to SB DB.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
```diff
@@ -0,0 +1,346 @@
...
+          taking a long time to apply. This usually means there is a large load on OVN component 'northd' or it is not
+          functioning.
+      expr: |
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        max_over_time(ovnkube_master_nb_e2e_timestamp[5m]) - max_over_time(ovnkube_master_sb_e2e_timestamp[5m]) > 120
+      for: 10m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNorthboundDatabaseClusterIDError
```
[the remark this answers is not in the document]
martinkennelly: It cannot be recovered automatically right now.
Agreed. Ill upgrade the severity and create a runbook.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Agreed. Ill bump the severity and create a runbook.

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
martinkennelly: The alerts here are only deployed on the management clusters and not the hosted clusters.

bindata/network/ovn-kubernetes/managed/alert-rules.yaml
martinkennelly: Also, northd is in HA mode in OCP so thats why youre seeing 3 time series.

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
bmeng: As my understanding the yaml defined in `self-hosted` will be deployed to hosted clusters? Could you help double confirm?

martinkennelly:
/retest

Failures unrelated to this PR.

bindata/network/ovn-kubernetes/self-hosted/alert-rules.yaml
martinkennelly: @zshi-redhat My understanding is the managed alerts are for the hosted clusters and the self hosted alerts will be used in the management cluster. Correct?

zshi-redhat: @martinkennelly  
The `managed` alerts in directory managed are for the hosted clusters - yes
The `self-hosted` alerts in directory self-hosted are for non-hypershift openshift deployment, which can be used in the management cluster in management cluster is using ovnk network type.

martinkennelly:
@wking Any interest in reviewing these alerts?

martinkennelly:
/retest
no ovn alerts seen firing

bindata/network/ovn-kubernetes/common/alert-rules.yaml
```diff
@@ -18,11 +18,36 @@ spec:
...
         severity: warning
+    - alert: OVNKubernetesNodePodAddError
+      annotations:
+        summary: OVN Kubernetes is experiencing pod creation errors at an elevated rate.
+        description: OVN Kubernetes experiences pod creation errors at an elevated rate for the last 5 minutes. The
+          pods will be retried.
+      expr: |
+        (sum(rate(ovnkube_node_cni_request_duration_seconds_bucket{command="ADD",err="true"}[5m]))
+          /
+        sum(rate(ovnkube_node_cni_request_duration_seconds_bucket{command="ADD"}[5m])))
```
simonpasquier, line 36: I believe you want `ovnkube_node_cni_request_duration_seconds_count` rather than `ovnkube_node_cni_request_duration_seconds_bucket`? The latter is for computing percentiles.

```suggestion
        (sum(rate(ovnkube_node_cni_request_duration_seconds_count{command="ADD",err="true"}[5m]))
          /
        sum(rate(ovnkube_node_cni_request_duration_seconds_count{command="ADD"}[5m])))
```

bindata/network/ovn-kubernetes/common/alert-rules.yaml
```diff
@@ -18,11 +18,36 @@ spec:
...
+        > 0.1
+      for: 15m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNodePodDeleteError
+      annotations:
+        summary: OVN Kubernetes experiencing pod deletion errors at an elevated rate.
+        description: OVN Kubernetes is experiences pod creation errors at an elevated rate for the last 5 minutes. The
```
simonpasquier, line 44:
```suggestion
        description: OVN Kubernetes is experiences pod deletion errors at an elevated rate for the last 5 minutes. The
```

bindata/network/ovn-kubernetes/common/alert-rules.yaml
```diff
@@ -18,11 +18,36 @@ spec:
...
+          pods will be retried.
+      expr: |
```
simonpasquier, line 46: same remark about `_count` vs. `_bucket`.

bindata/network/ovn-kubernetes/common/alert-rules.yaml
simonpasquier: it might also be useful to sum by instance/node maybe?

```suggestion
        (sum by(instance) (rate(ovnkube_node_cni_request_duration_seconds_bucket{command="ADD",err="true"}[5m]))
          /
        sum by(instance) (rate(ovnkube_node_cni_request_duration_seconds_bucket{command="ADD"}[5m])))
```

martinkennelly:
Thank you @simonpasquier for your review. We (SDN team) really appreciate your time here.

martinkennelly:
/retest
No failures relating to alerts here.

martinkennelly:
Forced pushed to update northd desc / summary after input from @qiowang721 
```
-        summary: No OVN northd is active. Networking is degraded.
-        description: No OVN northd is available within the high availability set. Networking control plane is degraded.
+        summary: Exactly one OVN northd instance must have an active status.
+        description: Exactly one OVN northd must have an active status within the high availability set.
+          Networking control plane is degraded.

```

martinkennelly:
/unhold
/retest

Lots of infra setup errors in the CI yesterday - hoping it was transient.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -36,10 +38,31 @@ spec:
...
-      for: 10m
+        # Without max_over_time, failed scrapes could create false negatives, see
+        # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
+        max(max_over_time(ovnkube_master_leader[5m])) == 0
+      for: 5m
       labels:
         severity: critical
+    - alert: V4SubnetAllocationThresholdExceeded
+      annotations:
+        summary: More than 80% of v4 subnets available to assign to the nodes are allocated. Current v4 subnet allocation percentage is {{"{{"}} $value {{"}}"}}.
```
simonpasquier, line 49:
```suggestion
        summary: More than 80% of v4 subnets available to assign to the nodes are allocated. Current v4 subnet allocation percentage is {{"{{"}} $value | humanizePercentage {{"}}"}}.
```

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -36,10 +38,31 @@ spec:
...
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/V4SubnetAllocationThresholdExceeded.md
+        description: More than 80% of IPv4 subnets are used. Insufficient IPv4 subnets could degrade provisioning of workloads.
+      expr: ovnkube_master_allocated_v4_host_subnets/ovnkube_master_num_v4_host_subnets * 100 > 80
```
simonpasquier, line 52:
```suggestion
      expr: ovnkube_master_allocated_v4_host_subnets / ovnkube_master_num_v4_host_subnets > 0.8
```

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -36,10 +38,31 @@ spec:
...
+      for: 10m
+      labels:
+        severity: warning
+    - alert: V6SubnetAllocationThresholdExceeded
+      annotations:
+        summary: More than 80% of the v6 subnets available to assign to the nodes are allocated. Current v6 subnet allocation percentage is {{"{{"}} $value {{"}}"}}.
```
simonpasquier, line 58:
```suggestion
        summary: More than 80% of the v6 subnets available to assign to the nodes are allocated. Current v6 subnet allocation percentage is {{"{{"}} $value | humanizePercentage {{"}}"}}.
```

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -36,10 +38,31 @@ spec:
...
+        description: More than 80% of IPv6 subnets are used. Insufficient IPv6 subnets could degrade provisioning of workloads.
+      expr: ovnkube_master_allocated_v6_host_subnets/ovnkube_master_num_v6_host_subnets * 100 > 80
```
simonpasquier, line 60:
```suggestion
      expr: ovnkube_master_allocated_v6_host_subnets / ovnkube_master_num_v6_host_subnets > 0.8
```

simonpasquier (commented):
I think that lots of new alerts won't carry a `namespace` label (for instance, `OVNKubernetesNorthboundDatabaseClusterIDError`). As mentioned in a previous comment, it should be fixed.

martinkennelly:
@simonpasquier resolved the namespace issue. Thank you highlighting this.
For most of the metrics, I depending on a namespace label being present, but if that wasn't possible, I added the "namespace" key to "labels".

martinkennelly:
/retest
No alerts from here firing.
Opened a bz for panic seen in controller manager operator.

martinkennelly:
@trozet could you please re-review this again after the previous comments by Simon?

PR #1527, opened 2022-08-17
Bug 2103680: avoid overrriding disableNetworkDiagnostics on reconciliation

jcaamano:
Unfortunately disableNetworkDiagnostics is not setup as a pointer and is
overriden to the default value on the server-side apply when the cluster
config is reconciliated.


kyrtapz:
/retitle Bug 2103680: avoid overrriding disableNetworkDiagnostics on reconciliation

PR #1528, opened 2022-08-17
Bug OCPBUGS-216: Kuryr: Bump timeoutSeconds for livenessProbe

dulek:
By default livenessProbe times out after 1 second. We just saw that this
is sometimes too little for the Kuryr health server to respond, so this
commit bumps that time to 10s.


MaysaMacedo:
Testing on 4.8 https://github.com/openshift/cluster-network-operator/pull/1529

MaysaMacedo:
/hold cancel
Only one test failed and Kuryr logs are ok. 
`[sig-node] Variable Expansion should fail substituting values in a volume subpath with backticks [Slow] [Conformance] [Suite:k8s] `

PR #1531, opened 2022-08-22
Hypershift: Allow configuring hostname and labels on the route

alvaroaleman:
This change makes it possible to configure the hostname and labels on
the Hypershift SBDB route through environment variables.

Configuring these is needed when the cluster is private, as in that case
a custom router is used to avoid sending the traffic over the internet.

In order to avoid duplicating the logic to determine if a cluster is
private and then needing to keep it in sync with the Hypershift
Controlplane operator, the simple environment variable approach was
chosen.

Ref https://issues.redhat.com/browse/HOSTEDCP-517

/assign @kyrtapz

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/008-route.yaml
+++ b/bindata/network/ovn-kubernetes/managed/008-route.yaml
@@ -8,6 +8,11 @@ metadata:
   namespace: {{.HostedClusterNamespace}}
   annotations:
     network.operator.openshift.io/cluster-name:  {{.ManagementClusterName}}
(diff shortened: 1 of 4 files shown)


alvaroaleman:
> error: unable to retrieve source image registry.build02.ci.openshift.org/ci-op-6rc80j72/stable manifest sha256:359e0d053fe671c09114411997774029f44fdf90598bc443aa1929a44469bfdc: received unexpected HTTP status: 504 Gateway Time-out

/retest-required

PR #1534, opened 2022-08-24
Migrate Egress Firewall Configuration during SDN migration and Rollback

pliurh:
With this patch, CNO will convert EgressNetworkPolicy CRs to
EgressFirewall CRs when migrating from OpenShift SDN to OVNKubernetes.
And convert EgressFirewall CRs to EgressNetworkPolicy CRs when
rolling back.


nicklesimba:
Everything looks good to me. I don't have any comments for these changes.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -400,6 +400,16 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	if operConfig.Spec.Migration != nil {
```
kyrtapz, line 403: should we also check for `Migration.NetworkType`?
There is a chance someone want's to migrate to a different MTU and `NetworkType` can be unset.

Also, do we need to check that `Migration.NetworkType` actually changed to handle a scenario where the value is set to the current CNI or one that is not supported at all?

pkg/controller/operconfig/migration.go
```diff
@@ -0,0 +1,80 @@
...
+
+		egressNetworkPolicy := &uns.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "network.openshift.io/v1",
+				"kind":       "EgressNetworkPolicy",
+				"metadata": map[string]interface{}{
+					"name":      ef.GetName(),
+					"namespace": ef.GetNamespace(),
+				},
+				"spec": ef.Object["spec"],
```
npinaeva, line 72: If I am reading the spec right, egress firewall rule may have `ports` section https://docs.openshift.com/container-platform/4.11/networking/ovn_kubernetes_network_provider/configuring-egress-firewall-ovn.html#egressnetworkpolicy-rules_configuring-egress-firewall-ovn, while egressnetworkpolicy doesn't have this field https://docs.openshift.com/container-platform/4.11/networking/openshift_sdn/configuring-egress-firewall.html#egressnetworkpolicy-rules_openshift-sdn-egress-firewall
it means that we can properly migrate from ovn to sdn only if this field is not set, or e.g. just drop ports section and log big warning

pkg/controller/operconfig/operconfig_controller.go
jcaamano: +1

Or another version of this: should we check that the resources we are migrating to don't already exist?

pkg/controller/operconfig/migration.go
pliurh: That is not our target. We only handle the migration from SDN to OVN and the rollback. So we only need to ensure that all the SDN config can be converted to OVN, not vice versa. Users shall not add unsupported egress firewall configuration before rolling back.

pkg/controller/operconfig/operconfig_controller.go
pliurh: > should we also check for Migration.NetworkType?

I'll add that.

>Or another version of this: should we check that the resources we are migrating to don't already exist?

I use `apply.ApplyObject` to create the resources. So I don't think we need to check if one resource already exists.

jcaamano: Not asking from the perspective of it failing or not. But if it exists, we can potentially overwrite something. Is that ok?

pkg/controller/operconfig/migration.go
jcaamano: So to understand better how this is supposed to work, users should ensure that the configuration is compatible or else can't migrate? What is an `unsupported egress firewall configuration`? Are we missing some context here that can help us review?

pliurh: What users have in Openshift SDN can be migrated to OVNKubernetes. And if they find some issue after the migration, they shall be able to roll back to SDN with the same configuration. So we don't expect, a user to do the opposite migration for a cluster with OVN at install time. So any fields supported only by OVN are out of the scope of the feature migration.

pkg/controller/operconfig/operconfig_controller.go
jcaamano: The answer to this is that the user is not expected to make any changes during migration or within the time window where it decides whether the rollback has been successful or not. After that, rollback is no longer supported.
Also the user is not expected to apply the egress firewall CRD and add egress firewalls manually before migration.
I don't think that a check for existence would do any harm and probably make the migration more robust, but not a strong opinion on it. I am fine with it as it is since any information loss would be responsibility of the customer if the above requirements are clearly layed out in the migration documentation.

pkg/controller/operconfig/migration.go
jcaamano: The answer to this is that the user is not expected to make any changes during migration or within the time window where it decides whether the rollback has been successful or not. So basically they should not make any change in egress firewalls that are not supported in egress network policies until migration is done.
Egress firewall should be spec schema compatible with egress network policy.

jcaamano:
Should we have unit tests for `convertEgressNetworkPolicyToEgressFirewall` to verify schema compatibility?

pliurh:
> Should we have unit tests for convertEgressNetworkPolicyToEgressFirewall to verify schema compatibility?

Can we address this in a separate CR? It requires some effort.

pkg/controller/operconfig/migration.go
```diff
@@ -0,0 +1,96 @@
...
+	operv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/cluster-network-operator/pkg/apply"
+	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
+)
+
+var gvrEgressFirewall = schema.GroupVersionResource{Group: "k8s.ovn.org", Version: "v1", Resource: "egressfirewalls"}
+var gvrEgressNetworkPolicy = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "egressnetworkpolicies"}
+
+func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
+	switch operConfig.Spec.DefaultNetwork.Type {
```
kyrtapz, line 23: Since we only check that `Migration.NetworkType` is set and not it's value we will always render those resources, even if the `Migration.NetworkType` is bogus. Understandably it is an user error but I think it would be easy to switch over `Migration.NetworkType` and not `DefaultNetwork.Type`
What do you think?

pliurh: I agree.

pkg/controller/operconfig/operconfig_controller.go
nicklesimba: >Also, do we need to check that Migration.NetworkType actually changed to handle a scenario where the value is set to the current CNI or one that is not supported at all?

I don't think it matters if the Migration.NetworkType is set to the current CNI. If for example current CNI is OVN and Migration.NetworkType is also OVN, it will attempt to migrate SDN resources to OVN. If this has been done already, that is okay, no change will be applied. If the SDN resources no longer exist, an error will be returned, and will disappear shortly after when migration finishes.

What value for CNI would not be supported? I suppose we can check like this:
```
if operConfig.Spec.Migration != nil && operConfig.Spec.Migration.NetworkType != "" 
&& (operConfig.Spec.Migration.NetworkType == "OpenshiftSDN" ||
operconfig.Spec.Migration.NetworkType == "OVNKubernetes") {
	migration := operConfig.Spec.Migration
	if migration.Features == nil || migration.Features.EgressFirewall {
		err = migrateEgressFirewallCRs(ctx, operConfig, r.client)
		if err != nil {
		}
	}
}

pkg/controller/operconfig/migration.go
nicklesimba: Update: I'll address these changes in my own PR since they aren't critical. For now, let's get this merged.

pkg/controller/operconfig/operconfig_controller.go
nicklesimba: Update: I'll address these changes in my own PR since they aren't critical. For now, let's get this merged.

PR #1536, opened 2022-08-24
Migrate Egress IP configuration during SDN migration and rollback

nicklesimba:
Patch that addresses blocker epic [OVN Migration Further Solution Enhancement](https://issues.redhat.com/browse/NP-20). 

With this patch, CNO will convert HostSubnets and Netnamespaces configured with egressIP to egressIP CR when migrating from Openshift SDN to OVN-Kubernetes. When rolling back, CNO will convert egressIP CR into configuration updates for HostSubnets and Netnamespaces. When said resources get recreated during rollback, the configuration updates will be applied, restoring egressIP config.

Of note: this feature only supports automatically configured egressIP migration for SDN to OVN migration. Manually configured SDN egressIPs are NOT supported and **the documentation should explicitly state this**.

Review comments are welcome and encouraged, as this feature is blocking the release of 4.12.

### DIFF
--- a/pkg/controller/operconfig/migration.go
+++ b/pkg/controller/operconfig/migration.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -18,12 +19,28 @@ import (
 )
 
 const defaultEgressFirewallName = "default"
+const egressCIDRAnnotationName = "networkoperator.openshift.io/sdn-to-ovn-migration"
+const egressIPNodeConfig = "cloud.network.openshift.io/egress-ipconfig"
+const egressAssignable = "k8s.ovn.org/egress-assignable"
 const multicastEnabledSDN = "netnamespace.network.openshift.io/multicast-enabled"
 const multicastEnabledOVN = "k8s.ovn.org/multicast-enabled"
 
 var gvrEgressFirewall = schema.GroupVersionResource{Group: "k8s.ovn.org", Version: "v1", Resource: "egressfirewalls"}
 var gvrEgressNetworkPolicy = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "egressnetworkpolicies"}
+var gvrEgressIp = schema.GroupVersionResource{Group: "k8s.ovn.org", Version: "v1", Resource: "egressips"}
+var gvrHostSubnet = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "hostsubnets"}
 var gvrNetnamespace = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "netnamespaces"}
+var gvrCloudPrivateIPConfig = schema.GroupVersionResource{Group: "cloud.network.openshift.io", Version: "v1", Resource: "cloudprivateipconfigs"}
+
+type NodeEgressIpConfig struct {
+	Interface string
+	IfAddr    map[string]string
+	Capacity  map[string]int
+}
+
+type OVNMigrationNodeAnnotation struct {
+	EgressCIDRs []string
+}
 
 func migrateMulticastEnablement(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
 	switch operConfig.Spec.Migration.NetworkType {
@@ -42,6 +59,21 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
 		return convertEgressNetworkPolicyToEgressFirewall(ctx, client)
 	case string(operv1.NetworkTypeOpenShiftSDN):
 		return convertEgressFirewallToEgressNetworkPolicy(ctx, client)
+	}
+
+	return nil
+}
+
+func migrateEgressIpCRs(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
+	switch operConfig.Spec.Migration.NetworkType {
+	case string(operv1.NetworkTypeOVNKubernetes):
+		egressIpList, netNamespaceList, err := convertSdnEgressIpToOvnEgressIp(ctx, client)
+		if err != nil {
+			return err
+		}
+		return applyEgressIpList(ctx, client, egressIpList, netNamespaceList)
+	case string(operv1.NetworkTypeOpenShiftSDN):
+		return convertOvnEgressIpToSdnEgressIp(ctx, client)
 	}
 
 	return nil
@@ -206,6 +238,209 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
 	return nil
 }
 
+func convertSdnEgressIpToOvnEgressIp(ctx context.Context, client cnoclient.Client) ([]*uns.Unstructured, []*uns.Unstructured, error) {
+
+	// 1. query for hostsubnets and netnamespaces
+	hostSubnetList, err := cnoclient.ListAllOfSpecifiedType(gvrHostSubnet, ctx, client)
+	if err != nil {
+		return nil, nil, err
+	}
+	netNamespaceList, err := cnoclient.ListAllOfSpecifiedType(gvrNetnamespace, ctx, client)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// 2. iterate through hostsubnets
+	//    - any with egressIP configured will cause update to node label "egress-assignable"
+	hostSubnetFound := false
+	for _, hsn := range hostSubnetList {
+		hostSubnetHasEgressIpConfigAutomatic := hostSubnetHasEgressIpConfigAutomatic(*hsn)
+		hostSubnetHasEgressIpConfigManual := hostSubnetHasEgressIpConfigManual(*hsn)
+
+		// mark node from hostsubnet with annotation as follows:
+		// - k8s.ovn.org/egress-assignable: ""
+		if hostSubnetHasEgressIpConfigAutomatic || hostSubnetHasEgressIpConfigManual {
+			hostSubnetFound = true
+			if err := labelNodeAndRemoveHostSubnetConfig(ctx, client, hsn); err != nil {
+				return nil, nil, err
+			}
+		}
+
+		if hostSubnetHasEgressIpConfigManual {
+			log.Printf("Manual configuration of SDN egressIP detected and is unsupported for migration; OVN egressIPs will be generated but will not maintain individual node assignments from SDN hostsubnets")
+		}
+	}
+
+	if !hostSubnetFound {
+		log.Printf("did not find a hostsubnet object with egressIP configured, quitting process early")
+		return nil, nil, nil
+	} else {
+		log.Printf("found hostsubnet object with egressIP configured, continuing...")
+	}
+
+	// 3. iterate through netnamespaces
+	//    - any with egressIP configured will cause an egressIP ovn resource to be created via k8s api
+	//    - a corresponding namespace label will be added to match the egressIP resource's namespace selector field
+	egressIpList := []*uns.Unstructured{}
+	for _, nns := range netNamespaceList {
+		if netNamespaceHasEgressIpConfig(*nns) {
+			// config detected, translating to ovn
+			// - delete cloudprivateipconfig if it exists
+			// - create egressip custom resource for ovn
+			if err := deleteCloudPrivateIpConfigs(ctx, client, nns.Object["egressIPs"].([]interface{})); err != nil {
+				return nil, nil, err
+			}
+
+			egressIpName := fmt.Sprint("egressip-", nns.GetName())
+			egressIP := unstructuredEgressIpObject(egressIpName, nns.Object["egressIPs"].([]interface{}), nns.Object["netname"])
+			egressIpList = append(egressIpList, egressIP)
+		}
+	}
+
+	return egressIpList, netNamespaceList, nil // success
+}
+
+func convertOvnEgressIpToSdnEgressIp(ctx context.Context, client cnoclient.Client) error {
+	egressIpRollbackReady, err := sdnEgressIpResourcesReady(ctx, client)
+	if !egressIpRollbackReady {
+		return nil // wait for all SDN egressIP resources to be created before rollback
+	} else if err != nil {
+		return err
+	}
+
+	// 1. query for egressips
+	egressIpList, err := cnoclient.ListAllOfSpecifiedType(gvrEgressIp, ctx, client)
+	if err != nil {
+		return err
+	}
+
+	// 2. query for nodes
+	nodeList, err := client.Default().Kubernetes().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// 3. iterate through egressIP objects and generate netnamespaces
+	//    - value of `namespaceSelector` will determine netnamespace's name
+	//    - egressIP list will be copied to egressIPs field for netnamespace
+	for _, eip := range egressIpList {
+		// translate egressIPs to SDN netnamespaces
+		// - delete cloudprivateipconfig if it exists
+		// - create egressip to netnamespace resources for sdn
+
+		egressIps := eip.Object["spec"].(map[string]interface{})["egressIPs"].([]interface{})
+
+		if err := deleteCloudPrivateIpConfigs(ctx, client, egressIps); err != nil {
+			return err
+		}
+
+		eipNamespace, found, err := uns.NestedString(eip.Object, "spec", "namespaceSelector", "matchLabels", "kubernetes.io/metadata.name")
+		if !found || err != nil {
+			return fmt.Errorf("kubernetes.io/metadata.name not found in egressIP object %s, probable underlying error: %v", eip.GetName(), err)
+		}
+
+		if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			nns, err := client.Default().Dynamic().Resource(gvrNetnamespace).Get(ctx, eipNamespace, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			nns.Object["egressIPs"] = egressIps
+			_, err = client.Default().Dynamic().Resource(gvrNetnamespace).Update(ctx, nns, metav1.UpdateOptions{})
+			return err
+		}); err != nil {
+			return err
+		}
+	}
+
+	// 4. iterate through nodes and generate hostsubnets for any that are egress-assignable
+	for _, node := range nodeList.Items {
+		if node.Labels == nil {
+			continue
+		}
+		if _, ok := node.Labels[egressAssignable]; ok {
+			// generate egressCIDRs field
+			// if "egressCIDRAnnotationName" node annotation exists, automatic sdn config was used and we shall reuse old config
+			// else, manual sdn config was used and we shall use node's subnet CIDR (cannot restore egressIPs)
+			var egressCIDRsArr []string
+			if _, ok := node.Annotations[egressCIDRAnnotationName]; ok {
+				ovnMigrationNodeAnnotation := OVNMigrationNodeAnnotation{}
+				if err := json.Unmarshal([]byte(node.Annotations[egressCIDRAnnotationName]), &ovnMigrationNodeAnnotation); err != nil {
+					return err
+				}
+				egressCIDRsArr = ovnMigrationNodeAnnotation.EgressCIDRs
+			} else {
+				egressIpConfig := node.Annotations[egressIPNodeConfig]
+				egressIpConfigArr := make([]NodeEgressIpConfig, 0)
+				if err := json.Unmarshal([]byte(egressIpConfig), &egressIpConfigArr); err != nil {
+					return err
+				}
+				if len(egressIpConfigArr) <= 0 {
+					return fmt.Errorf("unexpected error: egress-ipconfig annotation is empty")
+				}
+				nodeSubnet, ok := egressIpConfigArr[0].IfAddr["ipv4"]
+				if !ok {
+					return fmt.Errorf("unexpected error: egress-ipconfig annotation missing ipv4 entry")
+				}
+				egressCIDRsArr = []string{nodeSubnet}
+			}
+
+			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+				hsn, err := client.Default().Dynamic().Resource(gvrHostSubnet).Get(ctx, node.Name, metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+
+				if err := uns.SetNestedStringSlice(hsn.Object, egressCIDRsArr, "egressCIDRs"); err != nil {
+					return err
+				}
+				_, err = client.Default().Dynamic().Resource(gvrHostSubnet).Update(ctx, hsn, metav1.UpdateOptions{})
+				return err
+			}); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil // success
+}
+
+func sdnEgressIpResourcesReady(ctx context.Context, client cnoclient.Client) (bool, error) {
+	// get all nodes
+	// get all hostsubnets
+	// iterate over all node names and return false if any of them don't have an associated hostsubnet
+	nodeList, err := client.Default().Kubernetes().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
+	hostSubnetList, err := cnoclient.ListAllOfSpecifiedType(gvrHostSubnet, ctx, client)
+	if err != nil {
+		return false, err
+	}
+
+	for _, node := range nodeList.Items {
+		found := false
+		for _, hostSubnet := range hostSubnetList {
+			if hostSubnet.Object["host"] == node.Name {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false, nil
+		}
+	}
+
+	result, err := netNamespacesExistForAllNamespaces(ctx, client)
+	if !result {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
 func netNamespacesExistForAllNamespaces(ctx context.Context, client cnoclient.Client) (bool, error) {
 	// get all namespaces
 	// get all netnamespaces
@@ -233,4 +468,151 @@ func netNamespacesExistForAllNamespaces(ctx context.Context, client cnoclient.Cl
 	}
 
 	return true, nil
+}
+
+func netNamespaceHasEgressIpConfig(nns uns.Unstructured) bool {
+	if nns.Object["egressIPs"] != nil {
+		return len(nns.Object["egressIPs"].([]interface{})) > 0
+	}
+	return false
+}
+
+func hostSubnetHasEgressIpConfigAutomatic(hsn uns.Unstructured) bool {
+	if hsn.Object["egressIPs"] != nil && hsn.Object["egressCIDRs"] != nil {
+		return len(hsn.Object["egressIPs"].([]interface{})) > 0 && len(hsn.Object["egressCIDRs"].([]interface{})) > 0
+	}
+	return false
+}
+
+func hostSubnetHasEgressIpConfigManual(hsn uns.Unstructured) bool {
+	if hsn.Object["egressIPs"] != nil && hsn.Object["egressCIDRs"] == nil {
+		return len(hsn.Object["egressIPs"].([]interface{})) > 0
+	}
+	return false
+}
+
+func labelNodeAndRemoveHostSubnetConfig(ctx context.Context, client cnoclient.Client, hsn *uns.Unstructured) error {
+	// label node as egressassignable
+	hostStr := fmt.Sprintf("%v", hsn.Object["host"])
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		nodeObj, err := client.Default().Kubernetes().CoreV1().Nodes().Get(ctx, hostStr, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		if nodeObj.Labels == nil {
+			nodeObj.Labels = make(map[string]string)
+		}
+		if _, ok := nodeObj.Labels[egressAssignable]; !ok {
+			nodeObj.Labels[egressAssignable] = ""
+		}
+
+		// copy egressCIDRs to node label if possible
+		// - if egressCIDRs contains values (automatic config), on rollback this annotation will provide the respective hostsubnet with its original egressCIDRs values
+		// - if egressCIDRs is empty (manual config), on rollback this label will not exist, and so we default to node's subnet for egressCIDR field.
+		egressCIDRs, found, err := uns.NestedStringSlice(hsn.Object, "egressCIDRs")
+		if err != nil {
+			return fmt.Errorf("egressCIDRs not found in HostSubnet object %s, probable underlying error: %v", hsn.GetName(), err)
+		} else if found {
+			// automatic configuration
+			ovnMigrationAnnotation := OVNMigrationNodeAnnotation{
+				EgressCIDRs: egressCIDRs,
+			}
+
+			if nodeObj.Annotations == nil {
+				nodeObj.Annotations = make(map[string]string)
+			}
+			if _, ok := nodeObj.Annotations[egressCIDRAnnotationName]; !ok {
+				egressCIDRsText, err := json.Marshal(ovnMigrationAnnotation)
+				if err != nil {
+					return err
+				}
+				nodeObj.Annotations[egressCIDRAnnotationName] = string(egressCIDRsText)
+			}
+		}
+		_, err = client.Default().Kubernetes().CoreV1().Nodes().Update(ctx, nodeObj, metav1.UpdateOptions{})
+		return err
+	}); err != nil {
+		return err
+	}
+
+	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		hsn, err := client.Default().Dynamic().Resource(gvrHostSubnet).Get(ctx, hsn.GetName(), metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		// now remove egressIP config from hostsubnet
+		hsn.Object["egressCIDRs"] = nil
+		hsn.Object["egressIPs"] = nil
+		_, err = client.Default().Dynamic().Resource(gvrHostSubnet).Update(ctx, hsn, metav1.UpdateOptions{})
+		return err
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func deleteCloudPrivateIpConfigs(ctx context.Context, client cnoclient.Client, egressIps []interface{}) error {
+	for _, egressIp := range egressIps {
+		egressIpStr := fmt.Sprintf("%v", egressIp)
+		// delete cloudprivateipconfig
+		err := client.Default().Dynamic().Resource(gvrCloudPrivateIPConfig).Delete(ctx, egressIpStr, metav1.DeleteOptions{})
+		if err != nil {
+			if errors.IsNotFound(err) {
+				continue
+			} else {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func applyEgressIpList(ctx context.Context, client cnoclient.Client, egressIpList []*uns.Unstructured, netNamespaceList []*uns.Unstructured) error {
+	for _, egressIp := range egressIpList {
+		if err := apply.ApplyObject(ctx, client, egressIp, ""); err != nil {
+			return err
+		}
+	}
+
+	for _, nns := range netNamespaceList {
+		if netNamespaceHasEgressIpConfig(*nns) {
+			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+				nns, err := client.Default().Dynamic().Resource(gvrNetnamespace).Get(ctx, nns.GetName(), metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				// now remove egressIP config from netnamespace
+				nns.Object["egressIPs"] = nil
+				_, err = client.Default().Dynamic().Resource(gvrNetnamespace).Update(ctx, nns, metav1.UpdateOptions{})
+				return err
+			}); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func unstructuredEgressIpObject(egressIpName string, egressIps []interface{}, netname interface{}) *uns.Unstructured {
+	return &uns.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "k8s.ovn.org/v1",
+			"kind":       "EgressIP",
+			"metadata": map[string]interface{}{
+				"name": egressIpName,
+			},
+			"spec": map[string]interface{}{
+				"egressIPs": egressIps,
+				"namespaceSelector": map[string]interface{}{
+					"matchLabels": map[string]interface{}{
+						"kubernetes.io/metadata.name": netname,
+					},
+				},
+			},
+		},
+	}
 }
--- a/pkg/controller/operconfig/migration_test.go
+++ b/pkg/controller/operconfig/migration_test.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"testing"
 
-	// . "github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
 	v1 "github.com/openshift/api/network/v1"
 	"github.com/openshift/cluster-network-operator/pkg/client/fake"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+				"metadata": map[string]interface{}{
+					"name":            netName,
+					"resourceVersion": nnsResourceVersion,
+				},
+				"netid":     nnsNetId,
+				"netname":   netName,
+				"egressIPs": egressIps,
+			},
+		}
+		_, err = client.Default().Dynamic().Resource(gvrNetnamespace).Update(ctx, netnamespace, metav1.UpdateOptions{})
```
pliurh, line 292: Shall we use `Patch` instead of `Update` here? Also, we should check if the netnamespace object is already in the desired state before updating it.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+					"host":        node.Name,
+					"hostIP":      hostIp,
+					"metadata": map[string]interface{}{
+						"name":            node.Name,
+						"resourceVersion": hsnResourceVersion,
+					},
+					"subnet": hsnSubnet,
+				},
+			}
+			_, err = client.Default().Dynamic().Resource(gvrHostSubnet).Update(ctx, hostSubnet, metav1.UpdateOptions{})
```
pliurh, line 332: Ditto. Shall we use Patch instead of Update here?

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		_, err = client.Default().Dynamic().Resource(gvrNetnamespace).Update(ctx, netnamespace, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	// 4. iterate through nodes and generate hostsubnets for any that are egress-assignable
+	for _, node := range nodeList.Items {
+		if _, ok := node.Labels["k8s.ovn.org/egress-assignable"]; ok {
+			// parse out node's subnet from egressip annotation
```
pliurh, line 301: I think we need to document this. It means users cannot expect the hostsubnet object will be exactly the same as the one before migration.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	nnsList, err := client.Default().Dynamic().Resource(gvrNetnamespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	// 2. iterate through hostsubnets
+	//    - any with egressIP configured will cause update to node label "egress-assignable"
+	hostSubnetFound := false
+	for _, hsn := range hsnList.Items {
+		if hostSubnetHasEgressIpConfigManual(hsn) {
```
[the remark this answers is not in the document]
nicklesimba: I don't think this is a bad idea, we just need to make it explicit in the docs that manual config is not supported and this is the best we can do if the user uses manual config and also wants to keep the feature migration enabled.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+						},
+					},
+				},
+			}
+			_, err = client.Default().Dynamic().Resource(gvrEgressIp).Create(ctx, egressIP, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+
+			// also need to label namespace with the same label.
```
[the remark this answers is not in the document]
nicklesimba: Yes you are correct, I can remove this section of code because it is redundant.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	//    - a corresponding namespace label will be added to match the egressIP resource's namespace selector field
+	for i, nns := range nnsList.Items {
+		if netNamespaceHasEgressIpConfig(nns) {
+			// config detected, translating to ovn
+			// - delete cloudprivateipconfig if it exists
+			// - create egressip custom resource for ovn
+
+			for _, egressIp := range nns.Object["egressIPs"].([]interface{}) {
+				egressIpStr := fmt.Sprintf("%v", egressIp)
+				err = client.Default().Dynamic().Resource(gvrCloudPrivateIPConfig).Delete(ctx, egressIpStr, metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
nicklesimba: We can do this check in the rollback code. However, here this loop is redundant and I plan on removing it entirely. 

If there is any cloudprivateipconfig, it shall be removed by SDN automatically. This is because code previously deletes any hostsubnet egressIP config. As a result SDN will clean up all CloudPrivateIpConfig attached to the hostsubnet egressIP configs.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,313 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	for _, eip := range eipList.Items {
+		// translate egressIPs to SDN netnamespaces
+		// - delete cloudprivateipconfig if it exists
+		// - create egressip to netnamespace resources for sdn
+
+		egressIps := eip.Object["spec"].(map[string]interface{})["egressIPs"].([]interface{})
+
+		for _, egressIp := range egressIps {
+			egressIpStr := fmt.Sprintf("%v", egressIp)
+			err = client.Default().Dynamic().Resource(gvrCloudPrivateIPConfig).Delete(ctx, egressIpStr, metav1.DeleteOptions{})
```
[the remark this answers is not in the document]
nicklesimba: Yes, in rollback case this makes sense.

pkg/controller/operconfig/migration.go
nicklesimba: I don't see a benefit of switching to `Patch`, `Update` works fine in my tests. Also when performing a rollback, the netnamespace object will not be in the desired state (namely, that egressIPs are configured) because when rolling back, SDN creates the netnamespaces/hostsubnets with no additional configuration. Therefore the `egressCIDRs/egressIPs` fields are left empty and updating is necessary.

pkg/controller/operconfig/migration.go
nicklesimba: See my [previous reply](https://github.com/openshift/cluster-network-operator/pull/1536/files/8630b5f0d5bdfdfcaa5e4a5af1dfb9490b283554#r959866164).

pkg/controller/operconfig/migration.go
nicklesimba: If the SDN use case is normal and they have hostsubnet object using node subnet CIDR for `EgressCIDRs`, the object will be the exact same as before migration. However if they use some other value for `EgressCIDRs` the configuration rollback migration will generate will not be exactly the same, right, I agree. Definitely worth documenting these details!
There isn't a way to rollback perfectly for every possible use case the SDN API allows.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	return nil
+}
+
+func deleteCloudPrivateIpConfigs(ctx context.Context, client cnoclient.Client, egressIps []interface{}) error {
+	for _, egressIp := range egressIps {
+		egressIpStr := fmt.Sprintf("%v", egressIp)
+		// check if cloudprivateipconfig exists before attempting to delete
+		if _, err := client.Default().Dynamic().Resource(gvrCloudPrivateIPConfig).Get(ctx, egressIpStr, metav1.GetOptions{}); err != nil {
+			errMessage := fmt.Sprintf("cloudprivateipconfigs.cloud.network.openshift.io \"%s\" not found", egressIpStr)
+			if err.Error() == errMessage {
```
jcaamano, line 424: Use `IsNotFound` from `k8s.io/apimachinery/pkg/api/errors`

It would make sense to directly `Delete` the object nad just ignore the error if `IsNotFound`, no need to lookup.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		// - delete cloudprivateipconfig if it exists
+		// - create egressip to netnamespace resources for sdn
+
+		egressIps := eip.Object["spec"].(map[string]interface{})["egressIPs"].([]interface{})
+
+		if err := deleteCloudPrivateIpConfigs(ctx, client, egressIps); err != nil {
+			return err
+		}
+
+		eipNamespace := eip.Object["spec"].(map[string]interface{})["namespaceSelector"].(map[string]interface{})["matchLabels"].(map[string]interface{})["kubernetes.io/metadata.name"]
```
jcaamano, line 236: There are utility methods in `"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"` that might help having cleaner & more robust code when accessing or manipulating unstructured objects and is what is typically used for this.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		eipNamespaceStr := fmt.Sprintf("%v", eipNamespace)
+		nns, err := client.Default().Dynamic().Resource(gvrNetnamespace).Get(ctx, eipNamespaceStr, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		nnsResourceVersion := nns.Object["metadata"].(map[string]interface{})["resourceVersion"]
+		nnsNetId := nns.Object["netid"]
+		netName := eip.Object["spec"].(map[string]interface{})["namespaceSelector"].(map[string]interface{})["matchLabels"].(map[string]interface{})["kubernetes.io/metadata.name"]
```
jcaamano, line 244: This is the same as `eipNamespace`?

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+				"kind":       "NetNamespace",
+				"metadata": map[string]interface{}{
+					"name":            netName,
+					"resourceVersion": nnsResourceVersion,
+				},
+				"netid":     nnsNetId,
+				"netname":   netName,
+				"egressIPs": egressIps,
+			},
+		}
```
jcaamano, line 258: If I am reading this right we are just updating the egressIPs. Why no set them directly in `nns` instead of building a new Unstructured?

pkg/controller/operconfig/migration.go
jcaamano: Can we store this information in annotations on the resources created on migration and use it for the rollback?

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		if _, ok := node.Labels["k8s.ovn.org/egress-assignable"]; ok {
+			// parse out node's subnet from egressip annotation
+			egressIpConfigParsed := strings.SplitAfter(node.Annotations["cloud.network.openshift.io/egress-ipconfig"], "\"ipv4\":\"")
+			nodeSubnet := egressIpConfigParsed[1]
+			nodeSubnet = nodeSubnet[:strings.IndexByte(nodeSubnet, '"')]
+			nodeSubnetArr := []string{nodeSubnet}
+
+			hostIpParsed := strings.SplitAfter(node.Annotations["k8s.ovn.org/host-addresses"], "[\"")
+			hostIp := hostIpParsed[1]
+			hostIp = hostIp[:strings.IndexByte(hostIp, '"')]
```
jcaamano, line 275: I think unmarshaling would be more robust than parsing

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+			Object: map[string]interface{}{
+				"apiVersion": "network.openshift.io/v1",
+				"kind":       "NetNamespace",
+				"metadata": map[string]interface{}{
+					"name":            netName,
+					"resourceVersion": nnsResourceVersion,
+				},
+				"netid":     nnsNetId,
+				"netname":   netName,
+				"egressIPs": egressIps,
```
jcaamano, line 256: Should this be inside a spec?

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+					"apiVersion":  "network.openshift.io/v1",
+					"kind":        "HostSubnet",
+					"egressCIDRs": nodeSubnetArr,
+					"host":        node.Name,
+					"hostIP":      hostIp,
+					"metadata": map[string]interface{}{
+						"name":            node.Name,
+						"resourceVersion": hsnResourceVersion,
+					},
+					"subnet": hsnSubnet,
```
jcaamano, line 295: Should some of this stuff be inside a "spec"?
Also, if we are just updating egressCIDRs, why not do so directly in `hsnSubnet`?

jcaamano (commented):
I think we should write some tests for this.

pkg/controller/operconfig/migration.go
nicklesimba: For the manual case, users should not expect to get this information back on rollback, because support for manual migration is very limited. 

For automatic case, we can store this information somewhere and reuse it for rollback, sure. I am thinking it is acceptable to preserve the EgressCIDRs field by labeling the node with it. On rollback we can read this value if it exists. 

To avoid error in a manual config scenario, if the value does not exist, we default to node subnet.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +112,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
 	}
 	return nil
 }
+
+func convertSdnEgressIpToOvnEgressIp(ctx context.Context, client cnoclient.Client) error {
+
+	// 1. query for hostsubnets and netnamespaces
+	hsnList, err := client.Default().Dynamic().Resource(gvrHostSubnet).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
nicklesimba: > I think we should write some tests for this.
I agree. This is not a trivial effort however, so I will do this in a separate PR.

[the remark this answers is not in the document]
nicklesimba: I do not know how to make `List` calls paginated. Perhaps we can discuss this on Slack privately?

pkg/controller/operconfig/migration.go
nicklesimba: I'm not familiar with these libraries but that sounds way less painful. Leaving this comment unresolved for now since it's a lower priority compared to some other TODOs.

pkg/controller/operconfig/migration.go
nicklesimba: Yes, this was a mistake. Removing this and using eipNamespace to avoid redundancy.

pkg/controller/operconfig/migration.go
nicklesimba: No, see `bindata/network/openshift-sdn/001-crd.yaml`. It is a property, like apiVersion. I find it...strange as well. But it is what it is.

pkg/controller/operconfig/migration.go
nicklesimba: This was the product of old code sticking around. You are correct, I can just set them directly in nns.

pkg/controller/operconfig/migration.go
nicklesimba: This was the product of old code sticking around. You are correct, I can just set them directly in `hsn`. (I assume you mean `hsn` instead of `hsnSubnet`).

Regarding the "spec" comment:
> No, see bindata/network/openshift-sdn/001-crd.yaml. It is a property, like apiVersion. I find it...strange as well. But it is what it is.

maiqueb (commented):
Very superficial review ...

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: > I do not know how to make `List` calls paginated or what that does. Perhaps we can discuss this on Slack privately? For what it's worth, I do not have problems with excessive loads at the moment so I don't know how important this is. EgressFirewall migration code has similar `List` calls and has worked without issue.

See for example
https://github.com/openshift/sdn/blob/master/pkg/network/common/list_pager.go

This is important at scale.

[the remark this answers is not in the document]
nicklesimba: > I disagree with this - if there are no tests, how do we know this feature does what we want it to ? And do we ensure it does not break when future PRs are merged ?

I have a few comments in response.
- To be clear, in an ideal world I **100 percent** agree with your reasoning. However, this feature needs to be in by end of this sprint, and one week of development time is lost due to the face-to-face meetup in Raleigh. This makes it even more urgent to get this code in.
- I will not support the merging of this code until the ready-to-merge code has been tested manually by me and nothing is broken on my end.
- To your point about breaking when future code is introduced, I don't think this code (which is only executed during a migration) will even be hit at ALL by the average user. Hypothetically, for any user trying to perform a migration, why are they provisioning an OpenshiftSDN cluster using 4.12 code in the first place? 4.12 has not released yet and so risk is to be expected.
- See review comments on [the egressFirewall PR that has been merged](https://github.com/openshift/cluster-network-operator/pull/1534#issuecomment-1228585457). My PR is significantly larger than the egressFirewall PR and so I believe this code should be held to the same standard, where e2e CI can be provided at a later date (still, I believe CI is *high priority* and is what I shall work on next.
- Unit tests are reasonable to add, but again, requires some effort. I value an e2e test much higher than unit tests so I would rather focus development time on that.

pkg/controller/operconfig/migration.go
```diff
@@ -94,3 +111,305 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	}
+
+	// 3. iterate through netnamespaces
+	//    - any with egressIP configured will cause an egressIP ovn resource to be created via k8s api
+	//    - a corresponding namespace label will be added to match the egressIP resource's namespace selector field
+	for i, nns := range nnsList.Items {
+		if netNamespaceHasEgressIpConfig(nns) {
+			// config detected, translating to ovn
+			// - delete cloudprivateipconfig if it exists
+			// - create egressip custom resource for ovn
```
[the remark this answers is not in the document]
nicklesimba: This is a necessary step. CloudPrivateIpConfig from SDN may not be reused, and will share the same name as the new OVN resource. If it exists while trying to add the new resource, adding the new resource will fail.

pkg/controller/operconfig/migration.go
nicklesimba: I looked into this a little bit, and it seems like the code will only benefit stylistically from using said utility methods. That's definitely valuable, but I still think this is low priority for now given the 4.12 timeline - the code is overall still pretty readable (more imporantly, it is _functional_) and I want to focus more on e2e testing as the deadline approaches.

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
nicklesimba: > See for example
> https://github.com/openshift/sdn/blob/master/pkg/network/common/list_pager.go
>
> This is important at scale.
I added paginated list calls in my latest commit. I made these changes to ALL list calls in `migration.go` including egressFirewall migration code.

nicklesimba:
@pliurh @jcaamano is there anything that remains to be done before this code is merged? I believe I have addressed review comments - please feel free to add discussion as needed.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		// - delete cloudprivateipconfig if it exists
+		// - create egressip to netnamespace resources for sdn
+
+		egressIps := eip.Object["spec"].(map[string]interface{})["egressIPs"].([]interface{})
+
+		if err := deleteCloudPrivateIpConfigs(ctx, client, egressIps); err != nil {
+			return err
+		}
+
+		eipNamespace := eip.Object["spec"].(map[string]interface{})["namespaceSelector"].(map[string]interface{})["matchLabels"].(map[string]interface{})["kubernetes.io/metadata.name"]
```
jcaamano, line 227: Addressing `matchLabels` is an unsafe access as it is not required.

I guess we are assuming that these are only the egress ips that we previously migrated, but the code could be more robust anyway.

Overall I would have liked to see more use of [unstructure](https://pkg.go.dev/k8s.io/apimachinery@v0.25.0/pkg/apis/meta/v1/unstructured) to access unstructured objects, then we would not have to review or guess what and when is something required and must be there and which might not, and we get conversions handled for us for the most common data types.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		eipNamespaceStr := fmt.Sprintf("%v", eipNamespace)
+		nns, err := client.Default().Dynamic().Resource(gvrNetnamespace).Get(ctx, eipNamespaceStr, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		nns.Object["egressIPs"] = egressIps
+		if err := apply.ApplyObject(ctx, client, nns, ""); err != nil {
```
jcaamano, line 235: Same concern with ApplyObject as I had elsewhere

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+
+				r := bytes.NewReader([]byte(egressIpConfig))
+				decoder := json.NewDecoder(r)
+
+				egressIpConfigArr := make([]NodeEgressIpConfig, 0)
+				if err := decoder.Decode(&egressIpConfigArr); err != nil {
+					return err
+				}
+				nodeSubnet := egressIpConfigArr[0].IfAddr["ipv4"]
+				egressCIDRsArr = []string{nodeSubnet}
```
jcaamano, line 262: I wonder way a standard un-marshaling could not be done here?

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+			}
+
+			hsn, err := client.Default().Dynamic().Resource(gvrHostSubnet).Get(ctx, node.Name, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+
+			hsn.Object["egressCIDRs"] = egressCIDRsArr
+
+			if err := apply.ApplyObject(ctx, client, hsn, ""); err != nil {
```
jcaamano, line 272: Same concerns I had with ApplyObject elsewhere

pkg/controller/operconfig/migration.go
jcaamano: Yeah, I was just stating the idea, did not actually expect to be done in this PR.

pkg/controller/operconfig/migration.go
nicklesimba: > I guess we are assuming that these are only the egress ips that we previously migrated

Yes this is correct, so it is not an unsafe access if the user is rolling back according to what we allow in documentation.

> Overall I would have liked to see more use of [unstructure](https://pkg.go.dev/k8s.io/apimachinery@v0.25.0/pkg/apis/meta/v1/unstructured) to access unstructured objects

Perhaps this can be addressed later in a separate PR.

pkg/controller/operconfig/migration.go
nicklesimba: I found this approach to be elegant and it does not require any additional dependencies (all are from standard library).

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: > * To be clear, in an ideal world I **100 percent** agree with your reasoning. However, this feature needs to be in by end of this sprint (it is the ONLY release blocker for 4.12), and one week of development time is lost due to the face-to-face meetup in Raleigh. This makes it even more critical to get this code in.

I don't think this is a reason to not do tests. It would be a reason to scope smaller amount of work in a sprint/release. No tests just translates in lower quality code.

> * To your point about breaking when future code is introduced, I don't think this code (_which is only executed during a migration_) will even be hit at ALL by the average user. Hypothetically, for any user trying to perform a migration, why are they provisioning an OpenshiftSDN cluster using 4.12 code in the first place? 4.12 has not released yet and so risk is to be expected.

I don't think this is related to end user usage or circumstances. It's about code maintainability, which needs to happen as long as the code is there.

> * See review comments on [the egressFirewall PR that has been merged](https://github.com/openshift/cluster-network-operator/pull/1534#issuecomment-1228585457). My PR is significantly larger than the egressFirewall PR and so I believe this code should be held to the same standard, where e2e CI can be provided at a later date (still, I believe CI _is_ high priority and it is what I shall work on next.

That was an exception because it was the end of the sprint (while we are not at the end of our sprint) and it was stated tests would be added in a follow-up PR. It might change from being an exception to being a mistake if at the end the tests are never added or if it is used as a precedent to request the same kind of exception multiple other times.

> * Unit tests are reasonable to add, but again, requires some effort. I value an e2e test much higher than unit tests so I would rather focus development time on that. (also e2e tests for this shall live in `openshift/release`)

Writing unit tests is not a separate effort from writing the code itself.

> 
> Ultimately I don't control when code gets merged so if the consensus is that tests are absolutely necessary prior to merge...then hopefully I will be able to provide CI soon.

I am unaware if we have any kind of policy regarding unit tests. I will ask. I just made the suggestion because it seemed to me it would add quality to the PR.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		}
+	}
+
+	// copy egressCIDRs to node label if possible
+	// - if egressCIDRs contains values (automatic config), on rollback this label will provide the respective hostsubnet with its original egressCIDRs values
+	// - if egressCIDRs is empty (manual config), on rollback this label will not exist, and so we default to node's subnet for egressCIDR field.
+	egressCIDRs, ok := hsn.Object["egressCIDRs"]
+	if ok {
+		// automatic configuration
+		if _, ok := nodeObj.Annotations["sdn.to.ovn.migration/sdnEgressCIDRs"]; !ok {
```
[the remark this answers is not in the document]
nicklesimba: I will make the annotation key names constant.

I propose instead this key name `networkoperator.openshift.io/sdn-to-ovn-migration-egressCIDRs`. I don't really see a reason to generalize the key name fully encoded in the annotation, since egressCIDRs are the only thing we need to preserve/restore anyway. 

We do not save egressIPs to restore them because it is for manual SDN config, which we do not support.

If we are rolling back to an automatic SDN config, egressIPs will be restored automatically by SDN once netnamespace and hostsubnet get configured with their old egressIPs and egressCIDRs respectively (this is currently implemented).

pkg/controller/operconfig/list_pager.go
```diff
@@ -0,0 +1,48 @@
...
+func ListAllNamespaces(ctx context.Context, client cnoclient.Client) ([]*v1.Namespace, error) {
+	list := []*v1.Namespace{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.Default().Kubernetes().CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		list = append(list, obj.(*v1.Namespace))
+		return nil
+	})
+	return list, err
+}
```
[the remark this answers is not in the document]
nicklesimba: I'm putting it in pkg/client - it makes the most sense since it uses many client calls.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +120,325 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+			egressCIDRsText, err := json.Marshal(egressCIDRs)
+			if err != nil {
+				return err
+			}
+			nodeObj.Annotations["sdn.to.ovn.migration/sdnEgressCIDRs"] = string(egressCIDRsText)
+			if err := apply.ApplyObject(ctx, client, nodeObj, ""); err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
nicklesimba: > I am not sure about the use of ApplyObject here (or in general for the migration process).

I would agree with most of your logic. However when creating egressIP for migration to OVN I think it is okay.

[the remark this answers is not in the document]
nicklesimba: > Consider [RetryOnConflict](https://github.com/kubernetes/client-go/blob/v0.25.0/util/retry/util.go#L103) if you think conflicts are likely to happen.

There shouldn't be any conflicts here, I think I was just using apply.ApplyObject without fully understanding the impact of it.

pkg/controller/operconfig/migration.go
nicklesimba: > For example, if multicast migration creates an annotation during rollback, it will be forcefully overwritten by an Update call. Thank you for this suggestion.

This is already resolved, but just for book-keeping, I think the highlighted comment I had made earlier was incorrect. The reason things were being "forcefully overwritten" by `Update` is because I was defining an entirely new object and sending that into the `Update`. The proper way is to perform a `Get`, modifying any necessary fields (in this case the egressIPs field) and then run an `Update` on the object, that way the original config is preserved, alongside any updates you provide. Due to some comments by @jcaamano it appears using `Update` will be safer than `ApplyObject` (see discussion in later review comments)

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: Shouldn't this be as easy as defining a struct
```
type OVNMigrationNodeAnnotation {
    egressCIDRs []string
}
```
and then just un/marshaling this struct into/from the annotation?
And then we don't have to worry about assumptions on what might come in the future that will basically not play nice with the annotation key `sdn-to-ovn-migration-egressCIDRs` because it is too specific?

pkg/controller/operconfig/migration.go
jcaamano: So this is not only a stilistic type of suggestion, [unstructure](https://pkg.go.dev/k8s.io/apimachinery@v0.25.0/pkg/apis/meta/v1/unstructured) does the correct safety access check on the maps and uses type casting for conversions instead of Sprintf.

An example could be:
```
eipNamespace, found, err := unstructured.NestedString(eip.Object, "spec", "namespaceSelector", "matchLabels", "kubernetes.io/metadata.name")
if !found || err != nil {
    return fmt.Errorf(err, "kubernetes.io/metadata.name not found in egress ip object %s, probable underlying error: %v", eipName, err)
} 
```

in this case though `matchLabels` is defined as a string and not as an object so I am not sure it can be accessed as a `map[string]interface{}` but if your code works then this example should work as well.

pkg/controller/operconfig/migration.go
jcaamano: But I was wondering why this wouldn't work (I don't see additional dependencies here)
```
if err := json.Unmarshal([]byte(egressIpConfig), &egressIpConfigArr); err != nil {
    return err
}
```
instead of
```
r := bytes.NewReader([]byte(egressIpConfig))
decoder := json.NewDecoder(r)

egressIpConfigArr := make([]NodeEgressIpConfig, 0)
if err := decoder.Decode(&egressIpConfigArr); err != nil {
    return err
}
```

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +46,27 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
 }
 
+func migrateEgressIpCRs(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
+	switch operConfig.Spec.Migration.NetworkType {
+	case string(operv1.NetworkTypeOVNKubernetes):
+		egressIpList, err := convertSdnEgressIpToOvnEgressIp(ctx, client)
+		if err != nil {
+			return err
+		}
+		return applyEgressIpList(ctx, client, egressIpList)
```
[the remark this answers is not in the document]
nicklesimba: I'll move that part of the code into the applyEgressIpList, after the egressIPs have been applied without error.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +128,382 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+		_, err := client.Default().Dynamic().Resource(gvrHostSubnet).Update(ctx, hsn, metav1.UpdateOptions{})
+		if errors.IsConflict(err) {
+			// refresh hsn with Get() in the event of conflict
+			var getErr error
+			hsn, getErr = client.Default().Dynamic().Resource(gvrHostSubnet).Get(ctx, hsn.GetName(), metav1.GetOptions{})
+			if getErr != nil {
+				return getErr
+			}
+		}
+		return err
```
jcaamano, line 431: Check the `RetryOnConflic` docs:
```
// RetryOnConflict is used to make an update to a resource when you have to worry about
// conflicts caused by other code making unrelated updates to the resource at the same
// time. fn should fetch the resource to be modified, make appropriate changes to it, try
// to update it, and return (unmodified) the error from the update function
```
You need to return the error from `Update` unmodified.
So the use pattern is how you used it earlier: first Get and then Update and return the error from Update.

jcaamano: Please check other comments I had, I am not sure if they are being hidden from you when you marked the conversation as resolved.

pkg/controller/operconfig/migration.go
nicklesimba: Yes standard un-marshaling works fine. I guess since my code is to some degree confusing I'll switch to standard un-marshaling instead. 

Like:
```
egressIpConfig := node.Annotations[egressIPNodeConfig]
egressIpConfigArr := make([]NodeEgressIpConfig, 0)
if err := json.Unmarshal([]byte(egressIpConfig), &egressIpConfigArr); err != nil {
	return err
}
nodeSubnet := egressIpConfigArr[0].IfAddr["ipv4"]
egressCIDRsArr = []string{nodeSubnet}

File: pkg/controller/operconfig/migration.go (hunk 21 above)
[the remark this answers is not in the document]
Comment (nicklesimba) [reply pkg/controller/operconfig/migration.go]: I think because it _is_ so specific, it will actually play nice with code that comes in the future. I'm not sure what assumptions we are supposed to worry about here. However, I don't feel strongly one way or the other on this issue, so I went ahead and followed your suggestion.

File: pkg/controller/operconfig/migration.go (hunk 17 above)
Comment (nicklesimba) [reply pkg/controller/operconfig/migration.go:228]: From the egressIP yaml definition:
```
matchLabels:
  additionalProperties:
    type: string
  description: matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels map is equivalent to an element of matchExpressions, whose key field is "key", the operator is "In", and the values array contains only "value". The requirements are ANDed.
  type: object
```  

matchLabels is defined as a map object, not as a string.

Like you said matchLabels is optional, so I went ahead and replaced this instance with your example since it's a fair point. Other instances either have checks on optional fields or only check required fields, and they are significantly less verbose, so I left those unchanged.

File: pkg/controller/operconfig/migration.go (hunk 25 above)
Comment (nicklesimba) [reply pkg/controller/operconfig/migration.go:431]: > You need to return the error from Update unmodified.
So the use pattern is how you used it earlier: first Get and then Update and return the error from Update.

I wanted to avoid potentially unecessary Get() calls, but I see the issue. I suppose an extra Get() call is harmless while the current code could lead to unforeseen issues. I'll make this change (here and also at another spot in the code)

File: pkg/controller/operconfig/migration.go (hunk 15 above)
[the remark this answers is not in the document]
Comment (nicklesimba) [reply pkg/controller/operconfig/migration.go]: I have added unit tests. Please take a look. I would like to aim to have this PR merged by Friday, so any feedback is welcome. @jcaamano @pliurh

File: pkg/controller/operconfig/migration.go (hunk 21 above)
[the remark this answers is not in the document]
Comment (jcaamano) [reply pkg/controller/operconfig/migration.go]: My assumption was that if we were to need additional information saved in the future, I would it is better to have 1 annotation rather than n annotations.

File: pkg/controller/operconfig/migration_test.go [hunk 26]
```diff
@@ -0,0 +1,494 @@
...
+				egressIpsStringSlice := make([]string, len(egressIpsSlice))
+				for i := range egressIpsSlice {
+					egressIpsStringSlice[i] = egressIpsSlice[i].(string)
+				}
+				egressIpValueList = append(egressIpValueList, egressIpsStringSlice...)
+			}
+
+			// Check if values match exactly
+			less := func(a, b string) bool { return a < b }
+			if !(cmp.Diff(egressIpValueList, tc.expectedEgressIpList, cmpopts.SortSlices(less)) == "") {
```
jcaamano, line 198: gomega can do this, and is a dependency we already have.
Initialize a gomega instance with t at the beginning of the test case run:
```
import (
    . "github.com/onsi/gomega"
)

...

    t.Run(tc.name, func(t *testing.T) {
        g := gomega.WithT(t)

    ...
```

and then you can use

```
g.Expect(egressIpValueList).To(ConsistOf(tc.expectedEgressIpList...), "optional error explanation")
```

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	for i, nns := range netNamespaceList {
+		if netNamespaceHasEgressIpConfig(*nns) {
+			// config detected, translating to ovn
+			// - delete cloudprivateipconfig if it exists
+			// - create egressip custom resource for ovn
+			if err := deleteCloudPrivateIpConfigs(ctx, client, nns.Object["egressIPs"].([]interface{})); err != nil {
+				return nil, err
+			}
+
+			egressIpName := fmt.Sprint("egressip-", i)
```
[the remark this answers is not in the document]
nicklesimba: Sure, I see what you're saying and I can't think netnamespaces are a more descriptive way of naming anyway.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func applyEgressIpList(ctx context.Context, client cnoclient.Client, egressIpList []*uns.Unstructured) error {
+	for _, egressIp := range egressIpList {
+		if err := apply.ApplyObject(ctx, client, egressIp, ""); err != nil {
```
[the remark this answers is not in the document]
nicklesimba: It's not any different here than creating an egressFirewall with `ApplyObject` for that feature migration. I think this is okay the way it is, let me know if you can think of any issues.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+			return err
+		}
+	}
+
+	netNamespaceList, err := cnoclient.ListAllOfSpecifiedType(gvrNetnamespace, ctx, client)
+	if err != nil {
+		return err
+	}
+
+	for _, nns := range netNamespaceList {
```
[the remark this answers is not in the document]
nicklesimba: To answer this, we should look at the overall function being split as a whole 
```
case string(operv1.NetworkTypeOVNKubernetes):
	egressIpList, err := convertSdnEgressIpToOvnEgressIp(ctx, client)
	if err != nil {
		return err
	}
	return applyEgressIpList(ctx, client, egressIpList)
```

The function is split here because we can return the egressIpList prior to it being applied. This is useful for my unit test, and doesn't affect the code since the egressIpList gets applied in the function afterward anyway. 

This is needed for the SDNtoOVN function's unit test because fakeClient [has issues with ApplyObject function](https://github.com/kubernetes/kubernetes/issues/104052) Meanwhile, my OVNtoSDN function does not use ApplyObject and so fakeClient has no issues.

So yes we can move the CloudPrivateIpConfig deletion into the second loop, but we cannot consolidate everything into one loop/one overall function. I think this can remain the same unless there is some issue with functionality.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+					return err
+				}
+				egressCIDRsArr = ovnMigrationNodeAnnotation.EgressCIDRs
+			} else {
+				egressIpConfig := node.Annotations[egressIPNodeConfig]
+				egressIpConfigArr := make([]NodeEgressIpConfig, 0)
+				if err := json.Unmarshal([]byte(egressIpConfig), &egressIpConfigArr); err != nil {
+					return err
+				}
+				nodeSubnet := egressIpConfigArr[0].IfAddr["ipv4"]
```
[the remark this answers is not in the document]
nicklesimba: Assuming the user is rolling back to an SDN cluster and has not messed with their cluster settings, they should have this node annotation by default: 
```
cloud.network.openshift.io/egress-ipconfig: [{"interface":"nic0","ifaddr":{"ipv4":"10.0.128.0/17"},"capacity":{"ip":10}}]
```

When unmarshaling this, the whole thing is treated as a giant array with 1 element. Parsing further you can see how I got the access I did. So yes, this access should be safe as long as the user is coming from an SDN cluster and isn't trying to rollback from some arbitrary starting point.

abhat:
linter error from the utest changes: ```pkg/controller/operconfig/migration_test.go:23:16: Error return value is not checked (errcheck)```

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+
+	// 4. iterate through nodes and generate hostsubnets for any that are egress-assignable
+	for _, node := range nodeList.Items {
+		if node.Labels == nil {
+			continue
+		}
+		if _, ok := node.Labels[egressAssignable]; ok {
+			// generate egressCIDRs field
+			// if "egressCIDRAnnotationName" node annotation exists, automatic sdn config was used and we shall reuse old config
+			// else, manual sdn config was used and we shall use node's subnet CIDR (cannot restore egressIPs)
```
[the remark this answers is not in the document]
nicklesimba: Coming from an SDN cluster, the user will define the actual egressIP value (via hostsubnet and netnamespace). The user shouldn't be using _any node's specific IP address per se_, but should be using the node's subnet CIDR. Here's an example:

- So if some worker nodes A,B, and C have IPs `10.0.128.1`, `10.0.128.2`, and `10.0.128.3`, it is fine to define on a Netnamespace object an egressIP of `10.0.128.4`. 
- On the hostsubnet side, providing hostsubnet corresponding to worker node C with egressCIDR of `10.0.128.0/17` will allow the egressIP `10.0.128.4` to exist on worker node C. 

Providing a hostsubnet (and therefore its corresponding node) with egressCIDR equal to the node subnet CIDR effectively is the same as marking a node as "egress-assignable" in OVN. Let me know if this is clear, it can be pretty confusing.

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +132,388 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+
+	return nil // success
+}
+
+func sdnEgressIpResourcesReady(ctx context.Context, client cnoclient.Client) (bool, error) {
+	// get all nodes
+	// get all hostsubnets
+	// iterate over all node names and return false if any of them don't have an associated hostsubnet
+	// here!!
+	nodeList, err := client.Default().Kubernetes().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
nicklesimba: From Dan Winship:
"(FWIW defaultPageSize is 500 and we don't even support clusters with that many nodes so this will basically always be equivalent to doing a non-paged List.)"

pkg/controller/operconfig/test_helpers.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type fakeClusterClient struct {
+	cnoclient.ClusterClient
+	crclient crclient.Client
+}
+
+func (f *fakeClusterClient) CRClient() crclient.Client {
+	return f.crclient
+}
```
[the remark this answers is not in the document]
nicklesimba: These functions are originally from `mtu_probe_test.go` - I simply moved them to a more reasonable spot given that my test code uses these functions a couple of times. It makes more sense keeping this stuff in its own file than locking it to a single arbitrary test when multiple tests use it. I think this change can remain.

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: mkay, I was thinking on a 50 default page size for some reason

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: Ok. I might be paranoid here as well but it was me doing this I would do the safe check.

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: Thanks for explaining.

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: Perhaps we could pass the list of netns here as well, to make sure we are acting on the same ones that we originally saw.

jcaamano:
@nicklesimba thanks for doing the rework on additional PRs, would squash them to the reasonable amount at the end?

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
nicklesimba: I don't think this is necessary. The netnamespaces are created automatically by SDN, for every namespace that exists. Due to the readiness check in my code (see below) 
```egressIpRollbackReady, err := sdnEgressIpResourcesReady(ctx, client)```
...the cluster will have all the netnamespaces matching the namespaces it needs for rollback migration.

Once this condition is met, the CNO has stabilized and we don't expect a lot of restarts/etc. that could cause a netnamespace or namespace to disappear, and the CNO code itself does not delete namespaces or netnamespaces at any point. The only edge case is if a user deletes these resources, which...they shouldn't be configuring their cluster during a rollback at all, so I don't think we need to concern ourselves with this.

File: pkg/controller/operconfig/migration.go (hunk 30 above)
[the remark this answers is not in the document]
Comment (nicklesimba) [reply pkg/controller/operconfig/migration.go:273]: I could add some kind of check for peace of mind. This annotation will always be in this format when it is created by the cluster. I guess if a user changes it, that's bad news -- they shouldn't be messing with this annotation -- and we could report an error.
Like so:
```
if len(egressIpConfigArr) <= 0 {
	return fmt.Errorf("unexpected error: egress-ipconfig annotation is empty")
}
nodeSubnet, ok := egressIpConfigArr[0].IfAddr["ipv4"]
if !ok {
	return fmt.Errorf("unexpected error: egress-ipconfig annotation missing ipv4 entry")
} 
egressCIDRsArr = []string{nodeSubnet}
```

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
nicklesimba: I think I'll leave this as is for now just to keep it consistent between my different PRs...I'm definitely open to changing this later if the need arises; it's not a difficult change to make.

nicklesimba:
I've squashed commits down - hopefully we are nearing a ready-to-merge state.

pkg/controller/operconfig/migration.go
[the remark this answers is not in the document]
jcaamano: I am not talking about the rollback.

In convertSDNtoOVN we read the NetNS and build a list of egress IPs. 
The we pass the list of egress ips to applyEgressIPs where we read again the NetNS.

What I mean is returning back from convertSDNtoOVN the list of NetNS as well and passing it to applyEgressIPs along with the egressIPs.

I understand how improbable it is that the list of NetNS read in convertSDNtoOVN is different from the one read in applyEgressIPs, but the way I see it there is no reason to risk it if the solution is so trivial, avoids an extra read and it just feels more consistent all around.

I will leave it up to you though.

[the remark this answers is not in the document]
nicklesimba: Sure, I can make this change, it does avoid an extra read.

(Separate reminder for myself - unstructured is imported twice - fix this.)

pkg/controller/operconfig/migration.go
nicklesimba: (Leaving this unresolved so we don't miss the note on documentation)

nicklesimba:
> @nicklesimba thanks for doing the rework on additional PRs, would squash them to the reasonable amount at the end?

@jcaamano how do we feel about this PR? Are we ready to label for merge? Any additional comments? Thanks.

pkg/controller/operconfig/migration.go
nicklesimba: Actually, I'm going to resolve this. Because I have since added the feature Jaime suggested for automatic config, there is not much nuance here. 

The config will be restored in the automatic case, so there is nothing worth documenting because behavior is as expected. 

For manual case we don't support officially in the first place, so we don't need to provide more details for a "use at your own risk" feature.

PR #1537, opened 2022-08-25
microshift: compact ovn databases periodically

zshi-redhat:
Closes #[NP-542](https://issues.redhat.com//browse/NP-542)

### DIFF
--- a/bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
+++ b/bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
@@ -1,4 +1,4 @@
-a--
+---
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
--- a/bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
+++ b/bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
@@ -199,7 +199,8 @@ spec:
                 echo "$(date -Iseconds) - nbdb stopped"
                 rm -f /var/run/ovn/ovnnb_db.pid
         readinessProbe:
-          timeoutSeconds: 5
+          timeoutSeconds: 10
+          periodSeconds: 600
           exec:
             command:
             - /bin/bash
@@ -207,6 +208,7 @@ spec:
             - |
               set -xeo pipefail
               /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+              /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/compact 2>/dev/null
 
         env:
         - name: OVN_LOG_LEVEL
@@ -303,7 +305,8 @@ spec:
                 echo "$(date -Iseconds) - sbdb stopped"
                 rm -f /var/run/ovn/ovnsb_db.pid
         readinessProbe:
-          timeoutSeconds: 5
+          timeoutSeconds: 10
+          periodSeconds: 600
           exec:
             command:
             - /bin/bash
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
```diff
@@ -199,14 +199,15 @@ spec:
...
-          timeoutSeconds: 5
+          timeoutSeconds: 10
           exec:
             command:
             - /bin/bash
             - -c
             - |
               set -xeo pipefail
               /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/memory-trim-on-compaction on 2>/dev/null
+              /usr/bin/ovn-appctl -t /var/run/ovn/ovnnb_db.ctl --timeout=5 ovsdb-server/compact 2>/dev/null
```
maiqueb, line 210: Do we want to also bump this timeout ?

zshi-redhat: @maiqueb thanks for the review!
why this needs to be bumped? it may take more time when compacting database?

maiqueb: I honestly do not know, I just saw you bumped the readiness probe timeout, and assumed we'd probably need to also extend this timeout. 

Disregard my comment if that is not the case.

zshi-redhat: oh, I see your point. 
I bumped the overall timeout to 10s,  because by adding this commit it now has two cmds with each being 5s timeout.

maiqueb (approved):
Thanks.

bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
```diff
@@ -303,14 +305,17 @@ spec:
                 echo "$(date -Iseconds) - sbdb stopped"
                 rm -f /var/run/ovn/ovnsb_db.pid
         readinessProbe:
-          timeoutSeconds: 5
+          timeoutSeconds: 10
+          periodSeconds: 600
```
maiqueb, line 309: We go from compacting every 10 second to once per hour, right ?

zshi-redhat: From 10s to 10m, yes.
In microshift, it doesn't need to check the leader status as is done in openshift, thus reduce the frequence.
database won't grow that fast, so 10m sounds reasonable.

maiqueb: :S my math is horrible. Too hungry to multiply stuff by 10.

bindata/network/ovn-kubernetes/microshift/master/daemonset.yaml
```diff
@@ -1,4 +1,4 @@
-a--
+---
```
jcaamano, line 1: doh

jcaamano:
@zshi-redhat have you checked the effect of compacting the database? I don't know right now if in a standard deployment this now happens on a secondary thread or if it causes a leadership transfer. But on microshift, is there any kind of unavailability during compaction?

zshi-redhat:
> @zshi-redhat have you checked the effect of compacting the database? I don't know right now if in a standard deployment this now happens on a secondary thread or if it causes a leadership transfer. But on microshift, is there any kind of unavailability during compaction?

@jcaamano microshift uses standalone ovn database, no leader election.
the compact, if not manually done, will be triggered automatically with some conditions. I didn't see issues of unavailability, perhaps I didn't run intensive testing ? the scale of microshift workload is about 10-20 pods, 10 svc etc.

jcaamano:
I guess we will see what happens.

/lgtm
/approve

PR #1541, opened 2022-08-30
Bug SDN-3458: HyperShift: Differentiate resources deployed by different CNO instances in status manager

kyrtapz:
CNO should only reconcile its status based on the resources it rendered.
In the management cluster there is the default CNO and one CNO per hosted clusters control namespace.
By using a label value set to the infrastructure name we can differentiate the resources rendered by different CNO instances.

In stand-alone environments the value is a non-empty constant so it doesn't overlap with older CNO versions.

Additionally start watching deployments in the hosted cluster control plane namespace sine we added CNCC and multus admission controller to it.

### DIFF
--- a/pkg/controller/statusmanager/pod_watcher.go
+++ b/pkg/controller/statusmanager/pod_watcher.go
@@ -32,10 +32,10 @@ type PodWatcher struct {
 
 // initInformersFor sets up the DaemonSet, Deployment, and StatefulSet informers
 // for the given cluster and namespace
-// minor hack: Hypershift doesn't have permission to create or watch Deployments and Daemonsets
-// in the management cluster. So statefulSetOnly works around that :-/
-func (s *StatusManager) initInformersFor(clusterName, namespace string, statefulSetOnly bool) {
-	if !statefulSetOnly {
+// minor hack: Hypershift doesn't have permission to create or watch Daemonsets
+// in the management cluster. So includeDaemonsets works around that :-/
+func (s *StatusManager) initInformersFor(clusterName, namespace string, includeDaemonsets bool) {
+	if includeDaemonsets {
 		inf := v1appsinformers.NewFilteredDaemonSetInformer(
 			s.client.ClientFor(clusterName).Kubernetes(),
 			namespace,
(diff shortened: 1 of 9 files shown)


pkg/controller/statusmanager/status_manager.go
```diff
@@ -113,6 +114,16 @@ func New(client cnoclient.Client, name string) *StatusManager {
 		ssInformers:  map[string]cache.SharedIndexInformer{},
 		ssListers:    map[string]StatefulSetLister{},
 	}
+	cluster := ""
+	if status.hyperShiftConfig.Enabled {
+		cluster = status.hyperShiftConfig.Name
```
zshi-redhat, line 119: Should this be `bootstrapResult.Infra.InfraName`?
I noticed infraName is assigned to `names.GenerateStatusLabel` label.

kyrtapz: yes, absolutely. 
It is just a leftover from my previous approach

kyrtapz: Done, the PR is ready for a review @zshi-redhat

pkg/controller/statusmanager/pod_watcher.go
```diff
@@ -75,11 +76,11 @@ func (s *StatusManager) initInformersFor(clusterName, namespace string, stateful
 
 // AddPodWatcher wires up the PodWatcher to the controller-manager.
 func (s *StatusManager) AddPodWatcher(mgr manager.Manager) error {
-	s.initInformersFor("", metav1.NamespaceAll, false)
```
patjlm, line 78: in the hypershift hosted control plane case, why would we watch all namespaces ?

kyrtapz: We want to watch all of the resources created by the CNO in the hosted cluster.

patjlm: but that would not be on the same API endpoint, would it? This question is probably out of scope of this PR, sorry. Happy to take it elsewhere.

kyrtapz: Correct, CNO talks to both the management and hosted clusters API servers. 
`initInformersFor` uses the `clusterName` to determine which client to use (empty string means hosted cluster API server)

kyrtapz:
test e2e-hypershift-ovn

PR #1543, opened 2022-08-30
Migrate Multicast configuration during SDN migration and rollback

nicklesimba:
Patch that addresses blocker epic [OVN Migration Further Solution Enhancement](https://issues.redhat.com/browse/NP-20).

Multicast migration is simple, only annotations need to be updated.

### DIFF
--- a/pkg/client/list_pager.go
+++ b/pkg/client/list_pager.go
@@ -16,20 +16,9 @@ import (
 func ListAllOfSpecifiedType(resourceType schema.GroupVersionResource, ctx context.Context, client Client) ([]*uns.Unstructured, error) {
 	list := []*uns.Unstructured{}
 	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
-		return client.Default().Dynamic().Resource(resourceType).List(ctx, metav1.ListOptions{})
+		return client.Default().Dynamic().Resource(resourceType).List(ctx, opts)
 	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
 		list = append(list, obj.(*uns.Unstructured))
-		return nil
-	})
-	return list, err
-}
-
-func ListAllNodes(ctx context.Context, client Client) ([]*v1.Node, error) {
-	list := []*v1.Node{}
-	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
-		return client.Default().Kubernetes().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
-	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
-		list = append(list, obj.(*v1.Node))
 		return nil
 	})
 	return list, err
@@ -38,7 +27,7 @@ func ListAllNodes(ctx context.Context, client Client) ([]*v1.Node, error) {
 func ListAllNamespaces(ctx context.Context, client Client) ([]*v1.Namespace, error) {
 	list := []*v1.Namespace{}
 	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
-		return client.Default().Kubernetes().CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+		return client.Default().Kubernetes().CoreV1().Namespaces().List(ctx, opts)
 	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
 		list = append(list, obj.(*v1.Namespace))
 		return nil
--- a/pkg/client/list_pager.go
+++ b/pkg/client/list_pager.go
@@ -1,0 +1,47 @@
+package client
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/pager"
+	// cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
+)
+
+// for use with dynamic client
+func ListAllOfSpecifiedType(resourceType schema.GroupVersionResource, ctx context.Context, client Client) ([]*uns.Unstructured, error) {
+	list := []*uns.Unstructured{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.Default().Dynamic().Resource(resourceType).List(ctx, metav1.ListOptions{})
+	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		list = append(list, obj.(*uns.Unstructured))
+		return nil
+	})
+	return list, err
+}
+
+func ListAllNodes(ctx context.Context, client Client) ([]*v1.Node, error) {
+	list := []*v1.Node{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.Default().Kubernetes().CoreV1().Nodes().List(ctx, metav1.ListOptions{})
+	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		list = append(list, obj.(*v1.Node))
+		return nil
+	})
+	return list, err
+}
+
+func ListAllNamespaces(ctx context.Context, client Client) ([]*v1.Namespace, error) {
+	list := []*v1.Namespace{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.Default().Kubernetes().CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		list = append(list, obj.(*v1.Namespace))
+		return nil
+	})
+	return list, err
+}
--- a/pkg/controller/operconfig/migration.go
+++ b/pkg/controller/operconfig/migration.go
@@ -21,6 +21,17 @@ const defaultEgressFirewallName = "default"
 var gvrEgressFirewall = schema.GroupVersionResource{Group: "k8s.ovn.org", Version: "v1", Resource: "egressfirewalls"}
 var gvrEgressNetworkPolicy = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "egressnetworkpolicies"}
 
+func migrateMulticastEnablement(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
+	switch operConfig.Spec.Migration.NetworkType {
+	case string(operv1.NetworkTypeOVNKubernetes):
+		return enableMulticastOVN(ctx, client)
+	case string(operv1.NetworkTypeOpenShiftSDN):
+		return enableMulticastSDN(ctx, client)
+	}
+
+	return nil
+}
+
 func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, client cnoclient.Client) error {
 	switch operConfig.Spec.Migration.NetworkType {
 	case string(operv1.NetworkTypeOVNKubernetes):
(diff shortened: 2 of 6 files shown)


pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+	}
+
+	// 2. iterate through netnamespaces
+	//    - any with multicast-enabled="true" annotation will cause an update to the corresponding
+	//      namespace to add the necessary OVN annotation
+	for _, nns := range nnsList.Items {
+		if nns.Object["metadata"].(map[string]interface{})["annotations"] == nil {
+			continue
+		}
+		multicastAnnotation := nns.Object["metadata"].(map[string]interface{})["annotations"].(map[string]interface{})["netnamespace.network.openshift.io/multicast-enabled"]
```
danwinship, line 61:
```
multicastAnnotation := nns.GetAnnotations()["netnamespace.network.openshift.io/multicast-enabled"]
```

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+		if multicastAnnotation == "true" {
+			// first update namespace to have the same annotation
+			nspStr := fmt.Sprintf("%v", nns.Object["netname"])
+			namespaceObj, err := client.Default().Kubernetes().CoreV1().Namespaces().Get(ctx, nspStr, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			namespaceObj.Annotations["k8s.ovn.org/multicast-enabled"] = "true"
```
danwinship, line 69: to be extra safe you need to check if `namespaceObj.Annotations == nil` and create it if so.

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+
+			if err := apply.ApplyObject(ctx, client, namespaceObj, ""); err != nil {
```
danwinship, line 71: You should not use `apply.ApplyObject` for this; that is for creating/updating objects _that CNO owns_, and may do things you don't want, like updating `ManagedFields`. You should just update the object normally. (Except that you have to use `RetryOnConflict`.)

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+	// 2. iterate through netnamespaces
+	//    - any with multicast-enabled="true" annotation will cause an update to the corresponding
+	//      namespace to add the necessary OVN annotation
+	for _, nns := range nnsList.Items {
+		if nns.Object["metadata"].(map[string]interface{})["annotations"] == nil {
+			continue
+		}
+		multicastAnnotation := nns.Object["metadata"].(map[string]interface{})["annotations"].(map[string]interface{})["netnamespace.network.openshift.io/multicast-enabled"]
+		if multicastAnnotation == "true" {
+			// first update namespace to have the same annotation
```
danwinship, line 63: (there's no "second"...)

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+			}
+		}
+	}
+	return nil
+}
+
+func enableMulticastSDN(ctx context.Context, client cnoclient.Client) error {
+	multicastRollbackReady, err := netNamespacesExistForAllNamespaces(ctx, client)
+	if !multicastRollbackReady {
+		return nil // wait for all SDN netnamespace resources to be created before rollback
```
danwinship, line 82: It seems to me that returning `nil` will just cause it to think the migration completed successfully?

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+	} else if err != nil {
+		return err
+	}
+
+	// 2. query for namespaces
```
danwinship, line 87: "1"

pkg/controller/operconfig/migration.go
```diff
@@ -32,6 +43,84 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+					"kind":       "NetNamespace",
+					"metadata": map[string]interface{}{
+						"annotations": map[string]interface{}{
+							"netnamespace.network.openshift.io/multicast-enabled": "true",
+						},
+						"name": ns.Name,
+					},
+				},
+			}
+			if err := apply.ApplyObject(ctx, client, netNamespaceObj, ""); err != nil {
```
danwinship, line 111: again, using `ApplyObject` seems wrong here

pkg/controller/operconfig/migration.go
```diff
@@ -97,3 +186,32 @@ func convertEgressFirewallToEgressNetworkPolicy(ctx context.Context, client cnoc
...
+	}
+	netnamespaceList, err := client.Default().Dynamic().Resource(gvrNetnamespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	for _, namespace := range namespaceList.Items {
+		found := false
+		for _, netnamespace := range netnamespaceList.Items {
+			if netnamespace.Object["netname"] == namespace.Name {
```
danwinship, line 207: `if netnamespace.GetName() == namespace.Name` ?

pkg/controller/operconfig/migration.go
```diff
@@ -20,6 +20,7 @@ const defaultEgressFirewallName = "default"
 
 var gvrEgressFirewall = schema.GroupVersionResource{Group: "k8s.ovn.org", Version: "v1", Resource: "egressfirewalls"}
 var gvrEgressNetworkPolicy = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "egressnetworkpolicies"}
+var gvrNetnamespace = schema.GroupVersionResource{Group: "network.openshift.io", Version: "v1", Resource: "netnamespaces"}
```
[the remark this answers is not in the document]
nicklesimba: I will squash a bunch of commits down and then force push prior to merging.

pkg/controller/operconfig/migration.go
nicklesimba: This is acceptable behavior AFAIU, and is actually necessary for this migration rollback to work. Here is my understanding:

- No error has occurred, so we don't want to return an error. 

- See where the [function is called](https://github.com/openshift/cluster-network-operator/pull/1543/files#diff-7c8b030b905dedc903155ca81c4f4089318866fece8d863fa881ece4f46f9333L403-R412), the migration process will execute until the operConfig.Spec.Migration field becomes nil. We actually do want the migration to "complete successfully" by doing nothing and returning nil -- until the multicastRollback is ready (NetNamespaces should exist for all namespaces), at which point the migration truly will execute and complete successfully. 
- Note that this code, the egressFirewall migration code, and the egressIP migration code all will continue to execute as part of the CNO until the migration field is set to null.
- According to openshift docs the user must manually set the migration field to null by patching the CNO once certain conditions for migration rollback are met. My code is in line with this, because the NetNamespace objects will be recreated before said conditions are met, and the code beyond what you have highlighted will execute.

pliurh: Will there be any issue if we return an error here? An error message can help us to debug. Also normally, we shall return an error if the desired change cannot be done in a reconciliation cycle.

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +45,99 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+	namespaceList, err := cnoclient.ListAllNamespaces(ctx, client)
+	if err != nil {
+		return err
+	}
+
+	// 2. iterate through namespaces
+	//    - any with multicast-enabled=true annotation will cause an update to the corresponding
+	//      netnamespace to add the necessary SDN annotation
+	for _, ns := range namespaceList {
+		if ns.Annotations["k8s.ovn.org/multicast-enabled"] == "true" {
```
pliurh, line 104: Shall we check if `ns.Annotations` is nil before using it?

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +45,99 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+			// update netnamespace to have multicast annotation
+			netNamespaceObj := &uns.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "network.openshift.io/v1",
+					"kind":       "NetNamespace",
+					"metadata": map[string]interface{}{
+						"annotations": map[string]interface{}{
+							"netnamespace.network.openshift.io/multicast-enabled": "true",
```
pliurh, line 112: It would be better to use constants to store the annotation keys we use in this file.

pkg/controller/operconfig/migration.go
danwinship: You don't need to; for purposes of reading, a `nil` map is equivalent to an empty map. (But you need to check when writing, because writing to a `nil` map will cause a panic.)

pkg/controller/operconfig/migration.go
danwinship: Yes... I was going to say it should get the ovn annotation names by vendoring the ovn-kubernetes sources and getting them from there, but it's not even exported in ovn-kubernetes...

But having them as constants at least here (rather than hardcoding them in multiple places) is definitely good.

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +45,99 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+				return err
+			}
+			if namespaceObj.Annotations == nil {
+				namespaceObj.Annotations = make(map[string]string)
+			}
+			namespaceObj.Annotations["k8s.ovn.org/multicast-enabled"] = "true"
+
+			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+				_, err := client.Default().Kubernetes().CoreV1().Namespaces().Update(ctx, namespaceObj, metav1.UpdateOptions{})
+				return err
```
danwinship, line 77: You need to wrap everything back to the `Namespaces().Get()` inside the `RetryOnConflict`; if you get back a conflict error, that means someone else updated the object on the apiserver since you called `Get()`, so retrying the same `Update` is just going to fail again.

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +45,99 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+		return err
+	}
+
+	// 2. iterate through namespaces
+	//    - any with multicast-enabled=true annotation will cause an update to the corresponding
+	//      netnamespace to add the necessary SDN annotation
+	for _, ns := range namespaceList {
+		if ns.Annotations["k8s.ovn.org/multicast-enabled"] == "true" {
+			// update netnamespace to have multicast annotation
+			netNamespaceObj := &uns.Unstructured{
```
danwinship, line 106: This should work like the sdn-to-ovn direction does; fetch the existing NetNamespace and modify it.
(eg, as currently written, this would delete any egress IPs on the NetNamespace)

pkg/controller/operconfig/migration.go
```diff
@@ -32,12 +45,99 @@ func migrateEgressFirewallCRs(ctx context.Context, operConfig *operv1.Network, c
...
+					"metadata": map[string]interface{}{
+						"annotations": map[string]interface{}{
+							"netnamespace.network.openshift.io/multicast-enabled": "true",
+						},
+						"name": ns.Name,
+					},
+				},
+			}
+			if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+				_, err := client.Default().Dynamic().Resource(gvrNetnamespace).Update(ctx, netNamespaceObj, metav1.UpdateOptions{})
```
danwinship, line 119: (as above, the scope of the RetryOnConflict needs to be larger)

pkg/controller/operconfig/migration.go
nicklesimba: If the err in the highlighted code is non nil, we do return that. So yes it is as Dan said.

nicklesimba: > Is there a Jira issue for updating the docs to mention the new criteria?

I'm not sure what you mean. The criteria has not changed.

pkg/client/list_pager.go
```diff
@@ -0,0 +1,47 @@
...
+
+// for use with dynamic client
+func ListAllOfSpecifiedType(resourceType schema.GroupVersionResource, ctx context.Context, client Client) ([]*uns.Unstructured, error) {
+	list := []*uns.Unstructured{}
+	err := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+		return client.Default().Dynamic().Resource(resourceType).List(ctx, metav1.ListOptions{})
+	}).EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+		list = append(list, obj.(*uns.Unstructured))
+		return nil
+	})
```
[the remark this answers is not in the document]
nicklesimba: I think doing it the current way has worked fine, and it makes accesses to items in `migration.go` more readable since it's just a slice of the items, which is all the data my code needs to be interacting with.

pkg/client/list_pager.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/pager"
+	// cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
```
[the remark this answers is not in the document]
nicklesimba: I do use this package quite a few times, not sure why it was commented out in this commit.

PR #1544, opened 2022-08-31
OKD-49: Adds support for scos to multus

aleskandro:
This PR adds support for scos to the cnibincopy.sh script for multus.

Closes [OKD-49](https://issues.redhat.com//browse/OKD-49)

cc @LorbusChris

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -38,7 +38,7 @@ data:
     rhelmajor=
     # detect which version we're using in order to copy the proper binaries
     case "${ID}" in
-      rhcos) rhelmajor=8
+      rhcos|scos) rhelmajor=8
       ;;
       rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .)
       ;;


LorbusChris:
Thanks @aleskandro!
It's a bit confusing since SCOS has major version 9, but like on FCOS, we can simply copy the EL8 builds to the host here.

/lgtm
/assign @abhat

LorbusChris:
@aleskandro that'd need a RHEL9 or UBI9 base image, which we currently don't have (nor yet need). I think that can be added here when that base image is created eventually.

For now it should be OK to use the ubi8 based builds for EL9 (RHCOS and SCOS) nodes (it's what we do on FCOS too).

aleskandro:
> @aleskandro that'd need a RHEL9 or UBI9 base image, which we currently don't have (nor yet need). I think that can be added here when that base image is created eventually.
> For now it should be OK to use the ubi8 based builds for EL9 (RHCOS and SCOS) nodes (it's what we do on FCOS too).

Hi @LorbusChris, I added a comment in the commit message for future reference when we'll get the on-el9-built binary in the multus-cni image.

PR #1545, opened 2022-08-31
Bug: OCPBUGS-736: Kuryr: Use machine net MTU to create service net

dulek:
We configure Kuryr to create Neutron networks with MTU that CNO gets
from the machines network. We however still create network for services
with the default MTU. This commit fixes that and also uses machines
network MTU when creating the services network.


dulek:
/test e2e-openstack-kuryr

Seems flaky to me, some FIP association failure.

gryf (approved):
/lgtm

PR #1546, opened 2022-08-31
Multus admission controller: Wait for token in Hypershift

alvaroaleman:
The token is created and rotated by a sidecar to the admission
controller, which means most of the time it won't be there yet when the
process tries to read it during startup. Wait until it is there before
starting.

/assign @dougbtv @kyrtapz 

We have a check in our CI that pods don't restart and this makes it fail regularly, e.G. here: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hypershift/1689/pull-ci-openshift-hypershift-main-e2e-aws/1565002270200827904#1:build-log.txt%3A1927

### DIFF
--- a/bindata/network/multus-admission-controller/admission-controller.yaml
+++ b/bindata/network/multus-admission-controller/admission-controller.yaml
@@ -82,15 +82,29 @@ spec:
       - name: multus-admission-controller
         image: {{.MultusAdmissionControllerImage}}
         command:
-        - /usr/bin/webhook
-        args:
-        - -bind-address=0.0.0.0
-        - -port=6443
-        - -tls-private-key-file=/etc/webhook/tls.key
-        - -tls-cert-file=/etc/webhook/tls.crt
-        - -alsologtostderr=true
-        - -metrics-listen-address=127.0.0.1:9091
-        - -ignore-namespaces=openshift-etcd,openshift-console,openshift-ingress-canary,{{.IgnoredNamespace}}
+        - /bin/bash
+        - -c
+        - |-
+          set -euo pipefail
+{{- if .HyperShiftEnabled}}
+          retries=0
+          while [ ! -f /var/run/secrets/hosted_cluster/token ]; do
+            (( retries += 1 ))
+            sleep 1
+            if [[ "${retries}" -gt 30 ]]; then
+              echo "$(date -Iseconds) - Hosted cluster token not found"
+                exit 1
+            fi
+          done
+{{- end }}
+          exec /usr/bin/webhook \
+            -bind-address=0.0.0.0 \
+            -port=6443 \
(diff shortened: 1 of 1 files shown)


kyrtapz:
We are checking this for [cncc](https://github.com/openshift/cluster-network-operator/blob/master/bindata/cloud-network-config-controller/managed/controller.yaml#L109) and [ovn-k master](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml#L750) already but we missed it here.
Thanks @alvaroaleman!

bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -82,15 +82,24 @@ spec:
...
+        - /bin/bash
+        - -c
+        - |-
+          set -euo pipefail
+{{- if .HyperShiftEnabled}}
+          while ! [[ -f /var/run/secrets/hosted_cluster/token ]]; do
+             echo "Waiting for client token"
+             sleep 2
+          done
+{{- end }}
```
jcaamano, line 94: Should we eventually, after some time, exit with an error so that a possible problem is more evident to the outside world?
And this time probably needs to account for an offset delay from the time the file is created to when it appears in the mounted path.
Meaning I would not like this to always have restarts because the file is never going to be there at first but also not to be hung up forever hiding a possible problem.

alvaroaleman: @jcaamano good point, updated it to eventually fail. This should at most take a couple of seconds

bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -82,15 +82,29 @@ spec:
...
+          retries=0
+          while [ ! -f /var/run/secrets/hosted_cluster/token ]; do
+            (( retries += 1 ))
+            sleep 1
+            if [[ "${retries}" -gt 30 ]]; then
+              echo "$(date -Iseconds) - Hosted cluster token not found"
+                exit 1
+            fi
+          done
+{{- end }}
```
jcaamano, line 99: Wait.

/var/run/secrets/hosted_cluster/token is already used form the init container, why not wait there?

Also the wait fits better there I think since the init container is already expected to run to completion.

kyrtapz: The file is not read in the init container, the path is used to create a kubeconfig.
We cannot wait for it in the init container since it is actually created and managed by the sidecar container here: https://github.com/openshift/cluster-network-operator/blob/53b1e88d3e89886e94261a253748bf2898d2da6c/bindata/network/multus-admission-controller/admission-controller.yaml#L70

PR #1549, opened 2022-09-03
SDN-3442:Add alert for OVNKubernetesOvnKubeMasterDatabasesMonitorLoss

msherif1234:
ovnNBDB_query
OVNSBDB_query
SBDB_monitor_alert
NBDB_monitor_alert

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -352,3 +352,23 @@ spec:
       for: 5m
       labels:
         severity: critical
+    - alert: OVNKubernetesMasterNorthboundDatabaseMonitorLoss
+      annotations:
+        summary: OVN kubernetes master can not monitor northbound database(s).
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/OVNKubernetesMasterNorthboundDatabaseMonitorLoss.md
+        description: |
+          OVN Kubernetes master cannot monitor northbound database(s). This maybe the monitoring of the OVN leader itself, or the database which indicates the leaders status.
+      expr: |
+        max_over_time(ovnkube_master_libovsdb_monitors{primary_model="OVN_Northbound"}[5m]) != 2
+      labels:
+        severity: warning
+    - alert: OVNKubernetesMasterSouthboundDatabaseMonitorLoss
+      annotations:
+        summary: OVN kubernetes master can not monitor the southbound database(s).
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/OVNKubernetesMasterSouthboundDatabaseMonitorLoss.md
+        description: |
+          OVN Kubernetes master cannot monitor southbound database(s). This maybe the monitoring of the OVN leader itself, or the database which indicates the leaders status.
+      expr: |
+        max_over_time(ovnkube_master_libovsdb_monitors{primary_model="OVN_Southbound"}[5m]) != 2
+      labels:
+        severity: warning
--- a/bindata/network/ovn-kubernetes/self-hosted/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/alert-rules-control-plane.yaml
@@ -351,3 +351,23 @@ spec:
       for: 5m
       labels:
         severity: critical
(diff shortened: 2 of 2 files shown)


bindata/network/ovn-kubernetes/common/alert-rules.yaml
```diff
@@ -26,3 +26,13 @@ spec:
       for: 20m
       labels:
         severity: warning
+   - alert: OVNKubernetesOvnKubeMasterLostConnectionToDatabases
+      annotations:
+        summary: Networking control plane is degraded on node {{"{{"}} $labels.node {{"}}"}} because OVNKubeMaster lost connections to OVN SB/NB databases.
```
tssurya, line 31: If its master loosing connection to db's then its control plane alerts. This file is for data plane alerts (ovnkube-node alerts). You should probably add this to control-plane alerts file and test it.

tssurya (changes requested):

msherif1234:
runbook added by https://github.com/openshift/runbooks/pull/67

bindata/network/ovn-kubernetes/common/alert-rules.yaml
msherif1234: will attempt testing this when all pieces in place and content make sense

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -85,3 +85,13 @@ spec:
...
       labels:
         severity: warning
+    - alert: OVNKubernetesOvnKubeMasterLostConnectionToDatabases
+      annotations:
+        summary: ovnkube-master can not connect to the southbound/northbound DB for too long.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/OVNKubernetesOvnKubeMasterLostConnectionToDatabases.md
+        description: |
+          Networking is degraded on nodes when OVNKubeMaster lost connections to SB/NB databases . No networking control plane updates will be applied to the node.
+      expr: |
+        max_over_time(ovnkube_master_libovsdb_monitors[5m]) == 0
```
npinaeva, line 95: this metric is labelled, so i guess we need to create 2 different alerts for nbdb and sbdb like here https://github.com/openshift/cluster-network-operator/pull/1526/files#diff-3e4b0f06a694a6622c6a7cde162da4c82454271e9c709de8292aa1267dd3f12cR117
And the expected value of this metric is 2, so we probably want to alert when it !=2 or < 2 @dcbw can you help us inderstand what is the best way to create alert based on this monitors metric?

msherif1234: does this look any better I am not sure where I can find more on what this metric contains so I can know what to look for

msherif1234: also can I use one runbook for both alerts or we need one for each ?

martinkennelly: runbook for each unfortunately. They have to start with the alert name.

bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -352,3 +352,23 @@ spec:
...
       labels:
         severity: critical
+    - alert: OVNKubernetesOvnKubeMasterLostConnectionToNBDatabase
+      annotations:
+        summary: ovnkube-master can not connect to the northbound DB for too long.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-network-operator/OVNKubernetesOvnKubeMasterLostConnectionToDatabases.md
+        description: |
+          Networking is degraded on nodes when OVNKubeMaster lost connections to NB databases . No networking control plane updates will be applied to the node.
+      expr: |
+        min_over_time(ovnkube_master_libovsdb_monitors{db_name="OVN_Northbound"}[5m]) > 2
```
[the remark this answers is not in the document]
msherif1234: this will need to revisted with SNO and ushift right ? I assume they don't support alert atm

martinkennelly:
@msherif1234 This should be closed. I misunderstood this metric. The original intent was that this metric would update if a monitor lost connectivity. This metric doesn't represent this.
It will only decrement if we unregistered a monitor.

PR #1557, opened 2022-09-14
Use local MTU when node replica is zero in hypershift

zshi-redhat:
Closes #[HOSTEDCP-562](https://issues.redhat.com//browse/HOSTEDCP-562)

### DIFF
--- a/pkg/controller/operconfig/mtu_probe.go
+++ b/pkg/controller/operconfig/mtu_probe.go
@@ -48,6 +48,10 @@ func (r *ReconcileOperConfig) probeMTU(ctx context.Context, oc *operv1.Network, 
 			klog.Infof("AWS cluster, omitting MTU probing and using default of %d", azureMTU)
 			return azureMTU, nil
 		}
+		if infra.PlatformType == configv1.BareMetalPlatformType {
+			klog.Infof("Baremetal cluster, omitting MTU probing and using host local MTU")
+			return 0, nil
+		}
 	}
 	mtu, err := r.readMTUConfigMap(ctx)
 	if err == nil {


zshi-redhat:
cc @kyrtapz

zshi-redhat:
> I think with this change there is a high chance of CNO always picking the local MTU for `none` platform in hypershift as during the first CNO reconciliation there is probably no nodes available yet. If we really want to use the local mtu for `none` platform in hypershift we should just add it as another case here: https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/operconfig/mtu_probe.go#L42-L51

I don't think we want to always skip the mtu dection for `none` platform.
There was an issue with zero node pool in `none` platform that ovnkube-master and other CNO managed components are not created due to zero node. I wonder if we shall skip the mtu when node count is zero in hypershift case.

zshi-redhat:
@kyrtapz  PTAL, I don't see a better way to skip the mtu probe for none platform in hypershift, this commit uses host local mtu in such case. I think we probably need a API field for MTU longer term in hypershift, wdyt?

kyrtapz:
/lgtm
I do not see a better way forward right now.

zshi-redhat:
@jcaamano PTAL

jcaamano:
@zshi-redhat I am missing a proper explanation of what we are doing here.

What I understand is that we are skipping the MTU probe for baremetal hosted cluster. Doing so just because of a test case spawning 0 worker hosted clusters (reading from https://issues.redhat.com//browse/HOSTEDCP-562) which would be pretty unrealistic, does not make a lot of sense to me. So I might be misunderstanding something.

zshi-redhat:
> @zshi-redhat I am missing a proper explanation of what we are doing here.
> 
> What I understand is that we are skipping the MTU probe for baremetal hosted cluster. Doing so just because of a test case spawning 0 worker hosted clusters (reading from https://issues.redhat.com//browse/HOSTEDCP-562) which would be pretty unrealistic, does not make a lot of sense to me. So I might be misunderstanding something.

Yes, you got it right that the issue happens when 0 worker in the nodepool. 
I think it may happen, as a deployment model, that user requests a control plane first and later add a nodepool according to real workload need. And we don't say this is unsupported in hypershift.

/cc @csrwng for more input on the 0 worker use case.

jcaamano:
@zshi-redhat 

I guess there is no information available on the number of initial nodes that are being deployed?
Would it make sense to at least try to probe for some time instead of not trying at all?

zshi-redhat:
> @zshi-redhat
> 
> I guess there is no information available on the number of initial nodes that are being deployed? 

Yes, guest CNO doesn't have the necessary permissions to access the nodepool information which is exposed in the management cluster. 

> Would it make sense to at least try to probe for some time instead of not trying at all?

This was how the first version of this PR does, the problem is (quoted from Patryk): there is a high chance of CNO always picking the local MTU for none platform in hypershift as during the first CNO reconciliation there is probably no nodes available yet.

zshi-redhat:
> > @zshi-redhat
> > I guess there is no information available on the number of initial nodes that are being deployed?
> 
> Yes, guest CNO doesn't have the necessary permissions to access the nodepool information which is exposed in the management cluster.
> 
> > Would it make sense to at least try to probe for some time instead of not trying at all?
> 
> This was how the first version of this PR does, the problem is (quoted from Patryk): there is a high chance of CNO always picking the local MTU for none platform in hypershift as during the first CNO reconciliation there is probably no nodes available yet.

@jcaamano  do it make sense?

jcaamano:
> > there is a high chance of CNO always picking the local MTU for none platform in hypershift as during the first CNO reconciliation there is probably no nodes available yet.
> 
> @jcaamano do it make sense?

Hmm not really. 

Wouldn't that depend on the amount of extra time we would be willing to wait?

And isn't that (pick up the local MTU) what is happening with the latest changes anyway?

So what is Patryk saying?:
That we should not bother to wait because we would pick up the local MTU anyway so let's do that directly?
If there is no nodes during the first CNO reconciliation, doesn't that mean that we stay degraded until there's nodes because we can't probe the MTU?

BTW, isn't it an option to configure the MTU directly so that probing is not needed?

zshi-redhat:
> > > there is a high chance of CNO always picking the local MTU for none platform in hypershift as during the first CNO reconciliation there is probably no nodes available yet.
> > 
> > 
> > @jcaamano do it make sense?
> 
> Hmm not really.
> 
> Wouldn't that depend on the amount of extra time we would be willing to wait?
> 
Yes, we can extend the mtu probe timeout, but the time out should not be based on the prediction of user behavior (e.g. how long it would take for user to add a node). 

> And isn't that (pick up the local MTU) what is happening with the latest changes anyway?

Yes, latest change allows to skip mtu probe for baremetal and proceed on creating ovnkube master for the control plane. Previously the CNO will stuck at the mtu probe.
> 
> So what is Patryk saying?: That we should not bother to wait because we would pick up the local MTU anyway so let's do that directly? If there is no nodes during the first CNO reconciliation, doesn't that mean that we stay degraded until there's nodes because we can't probe the MTU?

Yes, CNO stays degraded until mtu probe succeeds. This is arguable that why we cannot have a ready control plane w/o a node.
> 
> BTW, isn't it an option to configure the MTU directly so that probing is not needed?

Yes, this will require API change in hypershift node pool (haven't talked with hypershift team).

jcaamano:
Thanks @zshi-redhat, sorry for so many questions. Getting to understand the extent of the problem.

> > BTW, isn't it an option to configure the MTU directly so that probing is not needed?
> 
> Yes, this will require API change in hypershift node pool (haven't talked with hypershift team).

I am referring to the cluster network MTU option that is already there. If a MTU value is provided, probing is not needed:

https://github.com/openshift/cluster-network-operator/blob/830daae9472c1e3f525c0af66bc7ea4054de9989/pkg/controller/operconfig/operconfig_controller.go#L229-L230

So isn't a better overall limitation to say for now that
```
If you don't plan to deploy any nodes, please specify an MTU for your cluster network
```
than
```
we will use you management cluster MTU for your hosted cluster unconditionally 
```

> Yes, we can extend the mtu probe timeout, but the time out should not be based on the prediction of user behavior (e.g. how long it would take for user to add a node).

But it could be based on the problem we get from the probe deployment. Like, if it is pending because no nodes are available. Just an idea, not sure if reasonable.


BTW, why is this specific to baremetal?

zshi-redhat:
> Thanks @zshi-redhat, sorry for so many questions. Getting to understand the extent of the problem.
> 
> > > BTW, isn't it an option to configure the MTU directly so that probing is not needed?
> > 
> > 
> > Yes, this will require API change in hypershift node pool (haven't talked with hypershift team).
> 
> I am referring to the cluster network MTU option that is already there. If a MTU value is provided, probing is not needed:
> 
> https://github.com/openshift/cluster-network-operator/blob/830daae9472c1e3f525c0af66bc7ea4054de9989/pkg/controller/operconfig/operconfig_controller.go#L229-L230
> 
> So isn't a better overall limitation to say for now that
> 
> ```
> If you don't plan to deploy any nodes, please specify an MTU for your cluster network
> ```
> 
> than
> 
> ```
> we will use you management cluster MTU for your hosted cluster unconditionally 
> ```
> 
> > Yes, we can extend the mtu probe timeout, but the time out should not be based on the prediction of user behavior (e.g. how long it would take for user to add a node).
> 
> But it could be based on the problem we get from the probe deployment. Like, if it is pending because no nodes are available. Just an idea, not sure if reasonable.

With hypershift, CNO is deployed by [control plane operator](https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L264) (a controller in hypershift), which doesn't expose user-facing option for mtu config like we do in OpenShift. Hypershift user doesn't interact with CNO network spec directly. 
> 
> BTW, why is this specific to baremetal?

The issue was found in baremetal platform and the mtu is [hardcoded](https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/operconfig/mtu_probe.go#L42-L50) for some other platforms such as aws, azure:

jcaamano:
@zshi-redhat 
 
> With hypershift, CNO is deployed by [control plane operator](https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L264) (a controller in hypershift), which doesn't expose user-facing option for mtu config like we do in OpenShift. Hypershift user doesn't interact with CNO network spec directly.

Can the control plane operator potentially have more information than us on the number of nodes that are going to be deployed such that it would make sense for them to configure a MTU (that would prevent probing) in case the number of worker nodes is 0?

zshi-redhat:
> @zshi-redhat
> 
> > With hypershift, CNO is deployed by [control plane operator](https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L264) (a controller in hypershift), which doesn't expose user-facing option for mtu config like we do in OpenShift. Hypershift user doesn't interact with CNO network spec directly.
> 
> Can the control plane operator potentially have more information than us on the number of nodes that are going to be deployed such that it would make sense for them to configure a MTU (that would prevent probing) in case the number of worker nodes is 0?

@jcaamano  Hosted control plane operator doesn't have visibility on the nodepools. The hypershift team suggested to avoid coupling the CNO and nodepools and instead to query the actual number of nodes, but it is hard to tell from CNO perspective when to stop querying the number of nodes.
Paste the link to previous discussion with @enxebre  and @csrwng : https://redhat-internal.slack.com/archives/C02M2K4B1FX/p1670242886544159?thread_ts=1669971178.515479&cid=C02M2K4B1FX

PR #1558, opened 2022-09-14
Allow empty vSphere status field in VIP sync

creydr:
VSphere UPI doesn't populate VSphere field. So with vSphere we should only handle the case, that `Status.Platform == configv1.VSpherePlatformType && Status.PlatformStatus.VSphere != nil`

### DIFF
--- a/pkg/controller/infrastructureconfig/sync_vips.go
+++ b/pkg/controller/infrastructureconfig/sync_vips.go
@@ -17,38 +17,39 @@ type apiAndIngressVipsSynchronizer struct{}
 
 // VipsSynchronize synchronizes new API & Ingress VIPs with old fields.
 // It returns if the status was updated and the updated infrastructure object.
+//
 //nolint:staticcheck
 func (*apiAndIngressVipsSynchronizer) VipsSynchronize(infraConfig *configv1.Infrastructure) *configv1.Infrastructure {
 	var apiVIPs, ingressVIPs *[]string // new fields
 	var apiVIP, ingressVIP *string     // old/deprecated fields
 
 	updatedInfraConfig := infraConfig.DeepCopy()
-	switch updatedInfraConfig.Status.Platform {
-	case configv1.BareMetalPlatformType:
+	switch {
+	case updatedInfraConfig.Status.Platform == configv1.BareMetalPlatformType:
 		apiVIPs = &updatedInfraConfig.Status.PlatformStatus.BareMetal.APIServerInternalIPs
 		apiVIP = &updatedInfraConfig.Status.PlatformStatus.BareMetal.APIServerInternalIP
 		ingressVIPs = &updatedInfraConfig.Status.PlatformStatus.BareMetal.IngressIPs
 		ingressVIP = &updatedInfraConfig.Status.PlatformStatus.BareMetal.IngressIP
 
-	case configv1.VSpherePlatformType:
+	case updatedInfraConfig.Status.Platform == configv1.VSpherePlatformType && updatedInfraConfig.Status.PlatformStatus.VSphere != nil: // VSphere UPI doesn't populate VSphere field
 		apiVIPs = &updatedInfraConfig.Status.PlatformStatus.VSphere.APIServerInternalIPs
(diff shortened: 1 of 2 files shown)


pkg/controller/infrastructureconfig/sync_vips.go
```diff
@@ -17,38 +17,39 @@ type apiAndIngressVipsSynchronizer struct{}
...
-	case configv1.BareMetalPlatformType:
+	switch {
+	case updatedInfraConfig.Status.Platform == configv1.BareMetalPlatformType:
 		apiVIPs = &updatedInfraConfig.Status.PlatformStatus.BareMetal.APIServerInternalIPs
 		apiVIP = &updatedInfraConfig.Status.PlatformStatus.BareMetal.APIServerInternalIP
 		ingressVIPs = &updatedInfraConfig.Status.PlatformStatus.BareMetal.IngressIPs
 		ingressVIP = &updatedInfraConfig.Status.PlatformStatus.BareMetal.IngressIP
 
-	case configv1.VSpherePlatformType:
+	case updatedInfraConfig.Status.Platform == configv1.VSpherePlatformType && updatedInfraConfig.Status.PlatformStatus.VSphere != nil: //VSphere UPI doesn't populate VSphere field
```
tssurya, line 34: nit: space after the double comment line. `// VSphere`

tssurya (approved):
/lgtm
no functionality change, PR just restructures the case to be more accommodating and explicit.

So if platform type is vsphere its not necessary the status is always set? Shouldn't that be fixed by VSphere UPI? Seems like same problem doesn't happen for the other platforms UPI install? Could be good to get a background of why this special case for vsphere.

pkg/controller/infrastructureconfig/sync_vips_test.go
```diff
@@ -267,6 +267,19 @@ func Test_apiAndIngressVipsSynchronizer_VipsSynchronize(t *testing.T) {
...
+			name: "should not panic on empty PlatformStatus.VSphere field (VSphere UPI doesn't populate VSphere field)",
+			givenStatus: configv1.InfrastructureStatus{
+				Platform: configv1.VSpherePlatformType,
+				PlatformStatus: &configv1.PlatformStatus{
+					VSphere: nil,
+				},
+			},
+			wantStatus: configv1.InfrastructureStatus{
+				Platform:       configv1.VSpherePlatformType,
+				PlatformStatus: &configv1.PlatformStatus{},
```
tssurya, line 280: @creydr : here the given status is nil, so we shouldn't take any action right? your switch case matches explicitly on the `!= nil` part, so should want status after reconcile also be nil?

creydr: Yes exactly. This test is only, to make sure we don't panic if `vsphere = nil` (what can happen in vsphere UPI). 
For the `!= nil` part, we have another test: https://github.com/openshift/cluster-network-operator/blob/1a56d16d64512522fbcbc95e0570f4a57ab14f11/pkg/controller/infrastructureconfig/sync_vips_test.go#L187-L208

pkg/controller/infrastructureconfig/sync_vips.go
creydr: :+1 updated it

creydr:
> So if platform type is vsphere its not necessary the status is always set? Shouldn't that be fixed by VSphere UPI? Seems like same problem doesn't happen for the other platforms UPI install? Could be good to get a background of why this special case for vsphere.

@cybertron @sinnykumari do you know, why vSphere UPI behaves that way and has an empty platformStatus.VSphere field?

jcpowermac:
> > So if platform type is vsphere its not necessary the status is always set? Shouldn't that be fixed by VSphere UPI? Seems like same problem doesn't happen for the other platforms UPI install? Could be good to get a background of why this special case for vsphere.
> 
> @cybertron @sinnykumari do you know, why vSphere UPI behaves that way and has an empty platformStatus.VSphere field?

@creydr vSphere UPI was created prior to IPI. We have users that use their own external LB. Its not defined because it doesn't use the api or ingress vips.

PR #1561, opened 2022-09-20
OCPBUGS-1515: Use custom uint128 type when validating v6InternalSubnet

pliurh:
For IPv6, default int value is not large enough to store the number of IPs. In some case, it causes int overflow.

In golang, there is no built-in type for uint128. So create a custom uint128 type for calculating v6 IP number.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"math/big"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


abhat:
Should we use https://cs.opensource.google/go/go/+/master:src/net/netip/uint128.go instead?

pliurh:
> Should we use https://cs.opensource.google/go/go/+/master:src/net/netip/uint128.go instead?

I don't think we can use it, because 1) it's a private type, and 2) it doesn't have the method we need here.

pkg/network/ovn_kubernetes.go
```diff
@@ -1526,5 +1522,64 @@ func isInternalSubnetLargeEnough(conf *operv1.NetworkSpec, v4 bool) bool {
...
+		i.hi = 0
+		i.lo = 0
+	} else if bits >= 64 {
+		i.hi = i.lo << (bits - 64)
+		i.lo = 0
+	} else {
+		i.hi <<= bits
+		i.hi |= i.lo >> (64 - bits)
+		i.lo <<= bits
+	}
```
jcaamano, line 1584: Could we place this in a reusable place with some tests?
Any reason not to use `Int` from `math/big`?

pliurh: I was not aware it exists. Let me refactor the patch with `math/big`. Thanks.

PR #1563, opened 2022-09-20
[release-4.10] OCPBUGS-1538: Make northd probe interval default to 10 seconds

tssurya:
/cc @trozet 
brings in https://github.com/openshift/cluster-network-operator/pull/1386/commits/eebce7172b9fc236a454809dc1768fcf7b3703fd and https://github.com/openshift/cluster-network-operator/pull/1494/files (which is the right fix)


trozet (approved):
/lgtm

thanks @tssurya !

tssurya:
@trozet : needs the risk assessed label

tssurya:
@anuragthehatter : PTAL

PR #1566, opened 2022-09-22
OCPBUGS-1341: Set owner reference for pod network connectivity check

pperiyasamy:
This would set node object which is hosting network-check-source pod as owner reference for PodNetworkConnectivityCheck objects so that these are automatically garbage collected when node is actually removed from cluster.

### DIFF
--- a/pkg/controller/connectivitycheck/connectivity_check_controller.go
+++ b/pkg/controller/connectivitycheck/connectivity_check_controller.go
@@ -3,12 +3,13 @@ package connectivitycheck
 import (
 	"context"
 	"fmt"
-	"github.com/openshift/cluster-network-operator/pkg/network"
 	"net"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/openshift/cluster-network-operator/pkg/network"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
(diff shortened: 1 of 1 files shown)


flavio-fernandes:
cool beans! TY Peri.
/assign @trozet

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -137,15 +142,33 @@ func (c *connectivityCheckTemplateProvider) generate(ctx context.Context, syncCo
...
 	for _, pod := range pods {
 		if pod.Spec.NodeName == "" {
 			// network-checker pod hasn't been assigned a node yet, skip
 			continue
 		}
+		var node *v1.Node
+		var ok bool
+		if node, ok = nodes[pod.Spec.NodeName]; !ok {
+			var err error
+			node, err = c.kubeClient.CoreV1().Nodes().Get(ctx, pod.Spec.NodeName, metav1.GetOptions{})
```
trozet, line 155: should we be using a node lister here instead of kube client? All of the other resources look to be using a lister.

pperiyasamy: yes, changed it to use node lister now.

trozet (approved):
/lgtm

PR #1567, opened 2022-09-22
OCPBUGS-1083: Improve alert expressions

npinaeva:
always use `for:` to make sure alert doesn't fire immediately 
for aggregation expressions make sure to use sum/count before max_over_time, since sum/count is the value we want to track

### DIFF
--- a/bindata/network/ovn-kubernetes/common/alert-rules.yaml
+++ b/bindata/network/ovn-kubernetes/common/alert-rules.yaml
@@ -33,6 +33,7 @@ spec:
           Networking is degraded on nodes when OVN controller is not connected to OVN southbound database connection. No networking control plane updates will be applied to the node.
       expr: |
         max_over_time(ovn_controller_southbound_database_connected[5m]) == 0
+      for: 5m
       labels:
         severity: warning
     - alert: OVNKubernetesNodePodAddError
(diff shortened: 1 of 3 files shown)


tssurya:
> always use `for:` to make sure alert doesn't fire immediately for aggregation expressions make sure to use sum/count before max_over_time, since sum/count is the value we want to track
> 
> Signed-off-by: Nadia Pinaeva [npinaeva@redhat.com](mailto:npinaeva@redhat.com)

on the original PR martin asked me to remove the `for` that was added,

npinaeva:
> on the original PR martin asked me to remove the `for:` that was initially present in the PR,

Thanks for pointing that out, but I think that was a mistake: when pod is killed, this alert will fire without `for` every time, because at the end of pod's life this metric goes to 0, and with max_over_time this value just shifts to the right.

npinaeva:
maybe merge https://github.com/openshift/cluster-network-operator/pull/1569 first to allow us some time to experiment with alerting expressions

bindata/network/ovn-kubernetes/common/alert-rules.yaml
```diff
@@ -33,6 +33,7 @@ spec:
           Networking is degraded on nodes when OVN controller is not connected to OVN southbound database connection. No networking control plane updates will be applied to the node.
       expr: |
         max_over_time(ovn_controller_southbound_database_connected[5m]) == 0
+      for: 5m
```
candita, line 36: Please consider making this 15 or 20 minutes, same as your other alerts.  We are experiencing severe CI issues due to this new alert.

npinaeva:
Quick fix didn't work, we need more time to properly update and test new expressions
Close for now in favor of https://github.com/openshift/cluster-network-operator/pull/1578

PR #1571, opened 2022-09-23
rebase to k8s v1.25.0

jluhrsen:


jluhrsen:
https://issues.redhat.com/browse/SDN-3201

ricky-rav:
Thanks Jamo!

The `k8s.io` part looks good! 

Now:
- github.com/openshift/* libraries too should be updated
- not sure what to do with sigs.k8s.io libraries, since they don't follow the k8s version. Let's update them too?
- I see you've updated gomega as well. There might be some incompatibilities from one version to another, but maybe not this time. Notice how in ovn-k we're still using v1.15.0 :-)

jluhrsen:
> Thanks Jamo!
> 
> The `k8s.io` part looks good!
> 
> Now:

can you check now please?

> * github.com/openshift/* libraries too should be updated

done.

> * not sure what to do with sigs.k8s.io libraries, since they don't follow the k8s version. Let's update them too?

got em. this had one hiccup that I think I resolved by updating a Get() function to match the new sig in the
updated lib.

> * I see you've updated gomega as well. There might be some incompatibilities from one version to another, but maybe not this time. Notice how in ovn-k we're still using v1.15.0 :-)

I don't know about this, so looking for your advice. I didn't purposely update gomega so it must have been
auto-updated with one of the other 'go get' commands I ran.

pkg/apply/apply.go
```diff
@@ -131,13 +131,15 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
...
-//  Finalizers
+//
+//	Name
+//	Namespace
+//	ClusterName
+//	Labels
+//	OwnerReferences
+//	ManagedFields
+//	Finalizers
+//
```
ricky-rav, line 142: wow, so now when it sees more than one space in a comment, it replaces it with a tab? :-o

ricky-rav:
/lgtm
... but let's wait for CI to pass :)

Thanks, Jamo!

pkg/apply/apply.go
jluhrsen: hopefully it's four spaces and not a tab :)

PR #1576, opened 2022-09-29
Jira OCPBUGS-1736: Always set PROXY variables for CNCC

andreaskaris:
Most components such as OVNK, SDN, etc. require PROXY env variables only for hypershift. The CNCC on the other hand must talk to the underlying cloud platform and must respect PROXY settings.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-1736

### DIFF
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -35,15 +35,11 @@ func InfraStatus(client cnoclient.Client) (*bootstrap.InfraStatus, error) {
 		APIServers:             map[string]bootstrap.APIServer{},
 	}
 
-	// If we use hypershift, the components need proxy settings. In selfhosted, they
-	// are always local to the API so there it is never needed.
-	if res.ControlPlaneTopology == configv1.ExternalTopologyMode {
-		proxy := &configv1.Proxy{}
-		if err := client.Default().CRClient().Get(context.TODO(), types.NamespacedName{Name: "cluster"}, proxy); err != nil {
-			return nil, fmt.Errorf("failed to get proxy 'cluster': %w", err)
-		}
-		res.Proxy = proxy.Status
+	proxy := &configv1.Proxy{}
+	if err := client.Default().CRClient().Get(context.TODO(), types.NamespacedName{Name: "cluster"}, proxy); err != nil {
+		return nil, fmt.Errorf("failed to get proxy 'cluster': %w", err)
 	}
+	res.Proxy = proxy.Status
 
 	// Extract apiserver URLs from the kubeconfig(s) passed to the CNO
 	for name, c := range client.Clients() {
--- a/pkg/controller/operconfig/mtu_probe.go
+++ b/pkg/controller/operconfig/mtu_probe.go
@@ -155,9 +155,14 @@ func renderMTUProber(infra *bootstrap.InfraStatus) ([]*uns.Unstructured, error) 
 	data.Data["KUBERNETES_SERVICE_PORT"] = infra.APIServers[bootstrap.APIServerDefault].Port
 	data.Data["DestNS"] = cmNamespace
 	data.Data["DestName"] = cmName
-	data.Data["HTTP_PROXY"] = infra.Proxy.HTTPProxy
-	data.Data["HTTPS_PROXY"] = infra.Proxy.HTTPSProxy
-	data.Data["NO_PROXY"] = infra.Proxy.NoProxy
+	data.Data["HTTP_PROXY"] = ""
+	data.Data["HTTPS_PROXY"] = ""
+	data.Data["NO_PROXY"] = ""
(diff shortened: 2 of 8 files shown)


andreaskaris:
It's atm not clear to me whether I have to readd the proxy vars for the managed hypershift side of things:
https://github.com/openshift/cluster-network-operator/commit/370f8f3b186ffd9f3f78ab9f79839f70a25d84fb#diff-455422beb3a7c7d9264deb6f3b37c645c371083e9e0bb4365d85b13197750281R98

pkg/platform/platform.go
```diff
@@ -35,15 +35,15 @@ func InfraStatus(client cnoclient.Client) (*bootstrap.InfraStatus, error) {
...
-			return nil, fmt.Errorf("failed to get proxy 'cluster': %w", err)
-		}
-		res.Proxy = proxy.Status
+	// If we use hypershift, the components need proxy settings. In selfhosted,
+	// only the CNCC needs proxy setttings.
+	proxy := &configv1.Proxy{}
+	if err := client.Default().CRClient().Get(
+		context.TODO(), types.NamespacedName{Name: "cluster"}, proxy,
+	); err != nil && res.ControlPlaneTopology == configv1.ExternalTopologyMode {
+		return nil, fmt.Errorf("failed to get proxy 'cluster': %v", err)
```
andreaskaris, line 44: So I wonder if this is even correct for hypershift - shall we really fail if the proxy/cluster object cannot be found? Shouldn't we merely log a warning?

andreaskaris:
@kyrtapz

kyrtapz:
> It's atm not clear to me whether I have to readd the proxy vars for the managed hypershift side of things: [370f8f3#diff-455422beb3a7c7d9264deb6f3b37c645c371083e9e0bb4365d85b13197750281R98](https://github.com/openshift/cluster-network-operator/commit/370f8f3b186ffd9f3f78ab9f79839f70a25d84fb#diff-455422beb3a7c7d9264deb6f3b37c645c371083e9e0bb4365d85b13197750281R98)

Starting 4.12 CNCC runs in the management cluster in HyperShift which means it should not use the proxy settings of the hosted cluster. Instead it should use the Proxy settings used by the management cluster: https://github.com/openshift/cluster-network-operator/pull/1577

pkg/platform/platform.go
kyrtapz: We deny cluster configs deletion: https://github.com/openshift/kubernetes/blob/master/openshift-kube-apiserver/admission/customresourcevalidation/config/deny_delete_cluster_config_resource.go#L31

If there is no custom proxy configuration `Proxy` has an empty spec:
```
apiVersion: config.openshift.io/v1
kind: Proxy
metadata:
  creationTimestamp: "2022-10-06T07:26:01Z"
  generation: 1
  name: cluster
  resourceVersion: "450"
  uid: 07de649c-391d-4937-9af7-e5582f8324b4
spec:
  trustedCA:
    name: ""
status: {}
```

I think we should treat `Proxy` the same as `Infrastructure` and always return an error if get failed.

pkg/platform/platform.go
```diff
@@ -35,15 +35,15 @@ func InfraStatus(client cnoclient.Client) (*bootstrap.InfraStatus, error) {
...
 
-	// If we use hypershift, the components need proxy settings. In selfhosted, they
-	// are always local to the API so there it is never needed.
-	if res.ControlPlaneTopology == configv1.ExternalTopologyMode {
-		proxy := &configv1.Proxy{}
-		if err := client.Default().CRClient().Get(context.TODO(), types.NamespacedName{Name: "cluster"}, proxy); err != nil {
-			return nil, fmt.Errorf("failed to get proxy 'cluster': %w", err)
-		}
-		res.Proxy = proxy.Status
+	// If we use hypershift, the components need proxy settings. In selfhosted,
```
kyrtapz, line 38: I don't think that comment makes sense here.
The code below just retrieves the proxy config.

MaysaMacedo:
It seems you might need to create a Proxy config [before validating the cluster config](https://github.com/openshift/cluster-network-operator/blob/68de0f6086bf2393a1d4c83262f16ad907907b61/pkg/network/cluster_config.go#L120), otherwise tests will fail with Not Found.

kyrtapz:
lint is failing:
```
 pkg/network/cluster_config_test.go:50:13: Error return value is not checked (errcheck)
	createProxy(client)
	           ^ 
```

andreaskaris:
@kyrtapz Can you review this again? Thanks!

MaysaMacedo:
I had a chance to give a try to this PR combined with the CNCC fix, the installation with OpenShift SDN succeeded. However, with OVNKubernetes there was one issue that prevented it from finishing, I opened a bug for it. While must-gather is available, I have an env if someone would like to take a peak. https://issues.redhat.com/browse/OCPBUGS-2536

PR #1581, opened 2022-10-04
Kuryr: Add missing keystoneauth options

dulek:
Seems like we've omitted passing a few OpenStack authentication options from clouds.yaml into kuryr.conf. This got noticed as application credentials doesn't get passed to Kuryr. This commit adds app creds as well as other missing options.


dulek:
@pierreprinetti - this should enable app credentials with Kuryr.

dulek:
Kuryr jobs seem cool here, we're having issue with this thing: https://github.com/openshift/origin/pull/27435

PR #1582, opened 2022-10-05
ovn-k, managed: pass join-subnet to control-plane

qinqon:
At hypershift the control plane of tenants nodes use different ovnkube.conf than the workers that are running at VMs. This change copy the missing argument to configure gateway router CIDR to the control-plane config.


qinqon:
@dcbw is CI failing on unrelated ? looks like...

jcaamano:
/override ci/prow/e2e-network-mtu-migration-ovn-ipv4

These failures seem to be related to some tests being renamed, attempting fix here:
https://github.com/openshift/release/pull/32975

PR #1583, opened 2022-10-06
SDN-3515: HyperShift: multus admission controller: expose metrics over HTTPs

kyrtapz:
The usage of kube-rbac-proxy is not supported in hosted clusters control namespace as it required a clusterrole binding.
Follow the example of other components and expose encrypted metrics directly.
Additionally, replace metrics id for multus admission controller.


kyrtapz:
/retest
@zshi-redhat ptal

PR #1585, opened 2022-10-07
IPsec dynamic mtu migration

mohammadheib:


pkg/network/ovn_kubernetes.go
```diff
@@ -910,7 +1126,18 @@ func fillOVNKubernetesDefaults(conf, previous *operv1.NetworkSpec, hostMTU int)
 		var syslogfacility string = "local0"
 		sc.PolicyAuditConfig.SyslogFacility = syslogfacility
 	}
-
+	if previous != nil && previous.DefaultNetwork.OVNKubernetesConfig != nil &&
```
dcbw, line 1129: @mohammadheib per our conversation I think we should move this to the render phase, rather than fill-defaults. That way we can report Progressing status too.

PR #1586, opened 2022-10-12
Bug OCPBUGS-736: Kuryr: If set use MTU from Config for svc net

dulek:
This commit fixes #1545 to make sure that Kuryr bootstrap honors the MTU set in KuryrConfig when creating the network for services. The MTU logic had to be moved from rendering to the bootstrap phase.

### DIFF
--- a/pkg/network/kuryr.go
+++ b/pkg/network/kuryr.go
@@ -110,16 +110,7 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
 	data.Data["ConfigMapHash"] = hash
 
 	// Pods Network MTU
-	mtu := b.PodsNetworkMTU
-	if c.MTU != nil {
-		if mtu >= *c.MTU {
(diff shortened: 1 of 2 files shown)


dulek:
@MaysaMacedo: I tested that this works, it's ready for a review.

MaysaMacedo:
Looks good to me. Just one question, should the mtu field definition change?
https://github.com/openshift/cluster-network-operator/blob/master/manifests/0000_70_cluster-network-operator_01_crd.yaml#L210-L215

dulek:
> Looks good to me. Just one question, should the definition of mtu field change? https://github.com/openshift/cluster-network-operator/blob/master/manifests/0000_70_cluster-network-operator_01_crd.yaml#L210-L215

Pretty good point, but I'll rather update API definition first. Here's a PR: https://github.com/openshift/api/pull/1332

dulek:
@MaysaMacedo: Can you consider an /lgtm on this?

PR #1587, opened 2022-10-13
OCPBUGS-6736: Remove nbctl logfile

bpickard22:
Problem: The ovn-nbctl logs were never rotated and are stored in a logfile that continually grows.
The nbctl daemon is run in detatched mode, which disables the file-descriptors, so we cannot just log to stdout

Solution: Simplest solution is to just remove the logs, especially since nbctl is not used in newer versions 


bpickard22:
4.7 is in maintenance mode, and this is not a critical fix. If it is deemed important enough, I will re-open this

palonsoro:
This is for 4.10, not 4.7

trozet (approved):
/lgtm

PR #1590, opened 2022-10-17
Remove the allow_ra sysctl for ipv4 from default systl whitelist

mmirecki:
The allow_ra sysctl controlls router advertisment settings which are an ipv6 thing only. As such it should only be present for ipv6.


yuvalk:
/lgtm
looks good to me
note, that there's only `/proc/sys/net/ipv6/conf/interface/accept_ra` so the ipv4 sysctl simply does not exists

mmirecki:
> /lgtm looks good to me note, that there's only `/proc/sys/net/ipv6/conf/interface/accept_ra` so the ipv4 sysctl simply does not exists

which will result in an error when the tuning cni will write to it

PR #1600, opened 2022-10-24
Bug 1896533: moved SetDegraded call out of object loop to process all items first

nicklesimba:
This is a follow up to #1128, for this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1896533

Signed-off-by: nicklesimba <simha.nikhil@gmail.com>

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -370,6 +370,7 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	r.status.SetRelatedClusterObjects(relatedClusterObjects)
 
 	// Apply the objects to the cluster
+	setDegraded := false
 	for _, obj := range objs {
 		// TODO: OwnerRef for non default clusters. For HyperShift this should probably be HostedControlPlane CR
 		if apply.GetClusterName(obj) == "" {
@@ -402,10 +403,14 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 					continue
 				}
 			}
-			r.status.SetDegraded(statusmanager.OperatorConfig, "ApplyOperatorConfig",
-				fmt.Sprintf("Error while updating operator configuration: %v", err))
-			return reconcile.Result{}, err
+			setDegraded = true
 		}
+	}
+
+	if setDegraded {
+		r.status.SetDegraded(statusmanager.OperatorConfig, "ApplyOperatorConfig",
+			fmt.Sprintf("Error while updating operator configuration: %v", err))
+		return reconcile.Result{}, err
 	}
 
 	if operConfig.Spec.Migration != nil && operConfig.Spec.Migration.NetworkType != "" {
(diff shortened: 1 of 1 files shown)


pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -402,12 +403,16 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
-			r.status.SetDegraded(statusmanager.OperatorConfig, "ApplyOperatorConfig",
-				fmt.Sprintf("Error while updating operator configuration: %v", err))
-			return reconcile.Result{}, err
+			setDegraded = true
 		}
 	}
 
+	if setDegraded {
+		r.status.SetDegraded(statusmanager.OperatorConfig, "ApplyOperatorConfig",
+			fmt.Sprintf("Error while updating operator configuration: %v", err))
```
dougbtv, line 412: aren't we going to need a way to store the error above, and then use it here? e.g. if the error is cleared when given a set of net-attach-defs `[i, j, k]` where `j` fails, but `k` completes, the error is nil when it gets past the loop and then processes this block for `setDegraded` is true?

nicklesimba: You're entirely right, great catch!

nicklesimba: For what it's worth, there's also the scenario where both `i` and `j` fail and cause degraded error. In this case I think we should be okay with returning the last error that occurs (`j`) and once it's resolved the error for `i` will appear.

dougbtv: I think last error is probably fine! Thanks!

nicklesimba:
/retitle Bug 1896533: moved SetDegraded call out of object loop to process all items first

djuran:
For someone who doesn't really understand what SetDegraded does

Will this change ensure that the CNO will continue to process all additionalNetworks even after encountering one for which the Namespace doesn't (yet) exist?

dougbtv:
@djuran that is indeed the intention! the degraded status will let you know that one (or more) additionalNetworks are invalid, but it will process the whole set instead of bailing out on the first invalid additionalNetwork

dougbtv:
Works as expected with these additional networks:

```
  additionalNetworks:
  - name: one
    namespace: openshift-multus
    type: Raw
    rawCNIConfig: |-
      {
        "cniVersion": "0.3.1",
        "name": "one",
        "type": "loopback"
      }
  - name: two
    namespace: noexist
    type: Raw
    rawCNIConfig: |-
      {
        "cniVersion": "0.3.1",
        "name": "two",
        "type": "loopback"
      }
  - name: three
    namespace: openshift-multus
    type: Raw
    rawCNIConfig: |-
      {
        "cniVersion": "0.3.1",
        "name": "three",
        "type": "loopback"
      }
```

with the degraded message being:

```
network                                    4.12.0-0.nightly-2022-10-25-210451   True        False         True       139m    Error while updating operator configuration: could not apply (k8s.cni.cncf.io/v1, Kind=NetworkAttachmentDefinition) noexist/two; Namespace error for networkattachment definition, consider possible solutions: (1) Edit config files to include existing namespace (2) Create non-existent namespace (3) Delete erroneous network-attachment-definition: could not apply (k8s.cni.cncf.io/v1, Kind=NetworkAttachmentDefinition) noexist/two: failed to apply / update (k8s.cni.cncf.io/v1, Kind=NetworkAttachmentDefinition) noexist/two: namespaces "noexist" not found
```

awesome. You can provide this step to QE and let them know that they should expect `one` & `three` to be created, but that the operator should go degraded and tell you something is wrong with `two`

PR #1605, opened 2022-10-26
Remove northbound DB on schema downgrade

kyrtapz:
To avoid dealing with stale data on downgrades remove the northbound database so it gets rebuilt from k8s API


danwinship:
/uncc
I'm not the right person to review this...

PR #1606, opened 2022-10-26
OCPBUGS-2598: ipsec: Run ovs-monitor-ipsec in the foreground and change probes

andreaskaris:
ipsec: Run ovs-monitor-ipsec in the foreground and change probes

Up until now, we ran ovs-monitor-ipsec in the background and ran a sleep
infinity after that. We then used the following in our liveness and
readiness probes:
  ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
We never parsed the output of either of these commands thus the probes
failed only when ovs-monitor-ipsec was down or when pluto failed. Probes
also caused issues in environments with large numbers of IPsec endpoints
as they caused the ovn-ipsec pods to be killed on pod startup.
Optimize this by running ovs-monitor-ipsec in the foreground and by only
keeping `ipsec status` in the probes. Also shorten the
initialDelaySeconds.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-2598

---------------------

As discussed with Tim, the current probes are of very little value:
~~~
ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
~~~
First of all,  `ovs-appctl -t ovs-monitor-ipsec ipsec/status` actually calls `ipsec status` so this was redundant
Secondly, we never parsed the output of either command to make sure that all tunnels were up. So we ran the status commands for the sake of running a status check but never new if everything was o.k. 

ovs-appctl -t ovs-monitor-ipsec ipsec/status actually blocks when a refresh() happens until all tunnels are added. One might argue that this is a valid check to make sure that everything is o.k.; however, this never verified if a tunnel was successfully brought up. All this did is cause liveness probes to fail until the operation is done, thus causing ovn-ipsec pods to be killed in large environments: https://issues.redhat.com/browse/OCPBUGS-2598 

Our new plan consists of 3 parts:
a) run in foreground and drop the probes
b) look into adding probes that check if all of the tunnels are actually up; while in the process of adding tunnels, add a taint to the node to make it unschedulable; if something goes wrong, the probe can still kill the pod if that's needed; either way, the probe must become way smarter
c) fix ovs-monitor-ipsec to come up faster -----> @mheib ttps://bugzilla.redhat.com/show_bug.cgi?id=2138135

This PR addresses a) so that we can close https://issues.redhat.com/browse/OCPBUGS-2598

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec.yaml
@@ -207,13 +207,13 @@ spec:
           # Start the pluto IKE daemon
           /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
 
-          # Environment variables are for workaround for https://mail.openvswitch.org/pipermail/ovs-dev/2020-October/375734.html
-          # We now start ovs-monitor-ipsec which will monitor for changes in the ovs
+          # Start ovs-monitor-ipsec which will monitor for changes in the ovs
           # tunnelling configuration (for example addition of a node) and configures
           # libreswan appropriately.
-          OVS_LOGDIR=/var/log/openvswitch OVS_RUNDIR=/var/run/openvswitch OVS_PKGDATADIR=/usr/share/openvswitch /usr/share/openvswitch/scripts/ovs-ctl --ike-daemon=libreswan --no-restart-ike-daemon start-ovs-ipsec
-
-          sleep infinity
+          # We are running this in the foreground so that the container will be restarted when ovs-monitor-ipsec fails.
+          /usr/libexec/platform-python /usr/share/openvswitch/scripts/ovs-monitor-ipsec \
+            --pidfile=/var/run/openvswitch/ovs-monitor-ipsec.pid --ike-daemon=libreswan --no-restart-ike-daemon \
+            --log-file --monitor unix:/var/run/openvswitch/db.sock
         env:
         - name: K8S_NODE
           valueFrom:
@@ -236,17 +236,6 @@ spec:
             cpu: 10m
             memory: 100Mi
         terminationMessagePolicy: FallbackToLogsOnError
-        readinessProbe:
-          exec:
-            command:
-            - /bin/bash
-            - -c
-            - |
-              #!/bin/bash
-              ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
-          initialDelaySeconds: 120
-          periodSeconds: 30
-          timeoutSeconds: 40
         livenessProbe:
           exec:
             command:
@@ -254,8 +243,8 @@ spec:
             - -c
             - |
               #!/bin/bash
-              ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
-          initialDelaySeconds: 60
+              ipsec status
+          initialDelaySeconds: 15
           periodSeconds: 60
       nodeSelector:
         beta.kubernetes.io/os: "linux"


bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
           exec:
             command:
             - /bin/bash
             - -c
             - |
               #!/bin/bash
               ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
-          initialDelaySeconds: 60
-          periodSeconds: 60
+          initialDelaySeconds: 15
```
andreaskaris, line 257: Wait for 15 seconds before starting the probes (to avoid false positives)

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
+          periodSeconds: 30
```
andreaskaris, line 258: Probe every 30 seconds

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
+          timeoutSeconds: 20
```
andreaskaris, line 259: Report failure if we do not get a response after 20 seconds

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
+          failureThreshold: 10
```
andreaskaris, line 260: After 30 seconds * 10 failures, kill and restart the pod  (that gives us 5 minutes)

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
             - /bin/bash
             - -c
             - |
               #!/bin/bash
               ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
-          initialDelaySeconds: 120
-          periodSeconds: 30
-          timeoutSeconds: 40
-        livenessProbe:
+          periodSeconds: 60
```
andreaskaris, line 247: Run livenessProbe every 60 seconds

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -236,27 +236,28 @@ spec:
...
+          timeoutSeconds: 20
```
andreaskaris, line 248: A single probe timed out after 20 seconds (before, this was at the default value = 1 seconds, so this should be more resilient now)

dcbw:
@mohammadheib

andreaskaris:
As discussed with Tim, the current probes are of very little value:
~~~
ovs-appctl -t ovs-monitor-ipsec ipsec/status && ipsec status
~~~
First of all,  `ovs-appctl -t ovs-monitor-ipsec ipsec/status` actually calls `ipsec status` so this was redundant
Secondly, we never parsed the output of either command to make sure that all tunnels were up. So we ran the status commands for the sake of running a status check but never new if everything was o.k. 

ovs-appctl -t ovs-monitor-ipsec ipsec/status actually blocks when a refresh() happens until all tunnels are added. One might argue that this is a valid check to make sure that everything is o.k.; however, this never verified if a tunnel was successfully brought up. All this did is cause liveness probes to fail until the operation is done, thus causing ovn-ipsec pods to be killed in large environments: https://issues.redhat.com/browse/OCPBUGS-2598 

Our new plan consists of 3 parts:
a) run in foreground and drop the probes
b) look into adding probes that check if all of the tunnels are actually up; while in the process of adding tunnels, add a taint to the node to make it unschedulable; if something goes wrong, the probe can still kill the pod if that's needed; either way, the probe must become way smarter
c) fix ovs-monitor-ipsec to come up faster -----> @mheib ttps://bugzilla.redhat.com/show_bug.cgi?id=2138135

This PR addresses a) so that we can close https://issues.redhat.com/browse/OCPBUGS-2598

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -207,13 +207,13 @@ spec:
           # Start the pluto IKE daemon
           /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
 
-          # Environment variables are for workaround for https://mail.openvswitch.org/pipermail/ovs-dev/2020-October/375734.html
-          # We now start ovs-monitor-ipsec which will monitor for changes in the ovs
+          # Start ovs-monitor-ipsec which will monitor for changes in the ovs
```
mohammadheib, line 210: Hi @andreaskaris,
thank you for submitting this fix.
i see that you are trying to save time by just executing the script in the foreground, i agree with you that we don't really care about the output of readinessProbe, livenessProbe executed commands but sometimes these commands return code can be  useful and can help to indicate an issue in IPsec,  for example:
if you have an issue with Pluto or it's not running for some reason this change will keep reporting that the pod is running fine but nothing really works, in the other hand the ipsec status command will returns non-zero code which will then make the pod failed to create.

please see this output example:
with your change enabling ipsec when pluto is not running will run fine:
`python3 /usr/share/openvswitch/scripts/ovs-monitor-ipsec --pidfile=/var/run/openvswitch/ovs-monitor-ipsec.pid --ike-daemon=libreswan --log-file --monitor unix:/var/run/openvswitch/db.sock 
2022-10-31T12:03:16Z |  0  | ovs-monitor-ipsec | INFO | Restarting IKE daemon
2022-10-31T12:03:16Z |  1  | ovs-monitor-ipsec | INFO | Restarting LibreSwan
Redirecting to: systemctl restart ipsec.service
Job for ipsec.service failed because the control process exited with error code.
See "systemctl status ipsec.service" and "journalctl -xe" for details.
2022-10-31T12:03:16Z |  3  | reconnect | INFO | unix:/var/run/openvswitch/db.sock: connecting...
2022-10-31T12:03:16Z |  6  | reconnect | INFO | unix:/var/run/openvswitch/db.sock: connected


keeps running .....
`

with the current implantation the  ipsec status command will return with code 33:
`ipsec status
whack: Pluto is not running (no "/run/pluto/pluto.ctl")
[root@localhost ~]# echo $?
33
`
what about maybe we keep the current implantation and remove the ovs-appctl -t ovs-monitor-ipsec ipsec/status from the probs and keep the ipsec status command, that will save us more time since we do the same command twice now.

andreaskaris: Also, our plan is to add a more complex probe / another container that actually runs a check that's a bit smarter and that will taint the node if not all tunnels are up. The idea is to avoid scheduling workloads to nodes where the ipsec pod is causing issues. That said, that would be in a follow-up commit

mohammadheib: > 

yes, you are right, we don't need the readinessProbe since we don't get traffic.

andreaskaris: @mohammadheib  Does this look better to you, now? (I'd still like to run the command itself in the foreground instead of the `sleep`, so I left it that way)

trozet (approved):
/lgtm

PR #1612, opened 2022-11-03
Remove references to the hosts kubeconfig

kyrtapz:
Since https://github.com/openshift/cluster-network-operator/pull/976 we are no longer using the hosts kubeconfig.

Remove the references to the old approach.

This addresses https://bugzilla.redhat.com/show_bug.cgi?id=2083334, security bugs do not work with the bot:

> When the change merges, the Bugzilla associated with the CVE must be set to MODIFIED. Since the DPTP bugzilla bot is not permitted to scan bugs with the SECURITY group in Bugzilla, The REP will not be able to use the bot's public functionality of moving their bug to MODIFIED.
> https://docs.google.com/document/d/1KuenDafC3Ukw19jY55tkVeH8nNVVAi8TEAfqynoVfzY/edit#heading=h.ikdk6suc575k

### DIFF
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -4,28 +4,21 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"log"
 	"math/rand"
-	"net/url"
 	"os"
 	"time"
 
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/operator"
+	"github.com/openshift/cluster-network-operator/pkg/version"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-	"k8s.io/client-go/tools/clientcmd"
-
-	_ "github.com/openshift/cluster-network-operator/pkg/client"
-	"github.com/openshift/cluster-network-operator/pkg/version"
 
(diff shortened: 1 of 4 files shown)


cmd/cluster-network-operator/main.go
```diff
@@ -4,9 +4,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"log"
+	"github.com/openshift/cluster-network-operator/pkg/version"
```
jcaamano, line 7: oops bad import group

jcaamano (changes requested):

cmd/cluster-network-operator/main.go
kyrtapz: fixed

jcaamano:
@kyrtapz wanted to check with you a couple of things:

1. no potential downgrade issue here, since probably the environment file has been available since a few versions already.
2. could you check with @zshi-redhat if microshift depends on any of this?

kyrtapz:
>     1. no potential downgrade issue here, since probably the environment file has been available since a few versions already.
There should be no downgrade issues, `/etc/kubernetes/apiserver-url.env` is available since 4.7 and we stopped reading the kubeconfig here: https://github.com/openshift/cluster-network-operator/pull/976
https://bugzilla.redhat.com/show_bug.cgi?id=2083334#c1
> 
>     2. could you check with @zshi-redhat if microshift depends on any of this?
From what I understand CNO is not run in microshift so this should not affect it at all(Thanks @pmtk!)
@zshi-redhat please help us confirm that there is no indirect impact on microshift.

jcaamano:
After talking with @zshi-redhat, this is a non issue for microshift.

PR #1614, opened 2022-11-09
OCPBUGS-3460: CNI binary copy should account for the possibility of symlinks

dougbtv:
I still am wholly unsure how the symlink gets created in the reported bug, but, this should make sure that the destination is unlinked before the copy.


s1061123 (approved):
/lgtm
/retest-required

PR #1617, opened 2022-11-09
OCPBUGS-64: CNI binary copy should account for the possibility of symlinks [backport 4.10]

dougbtv:
This also includes the fix for the atomic move during CNI binary copy.


dougbtv:
/retitle OCPBUGS-64: CNI binary copy should account for the possibility of symlinks [backport 4.10]

s1061123 (approved):
/lgtm

PR #1618, opened 2022-11-11
Delete pod network connectivity check when target node is removed

pperiyasamy:
Currently pod network connectivity check object contains source node set in owner reference field, this cleans up object only when source node is removed from cluster. But this won't delete the object when node hosting target endpoint is removed. Hence this commit sets target node as owner reference as well.


pperiyasamy:
Closing this PR in favor of https://github.com/openshift/library-go/pull/1430.
Thanks @kyrtapz for suggesting a nice fix.

PR #1629, opened 2022-11-18
OCPBUGS-2532: Disable UDP aggregation on s390x

danwinship:
The golang ethtool library doesn't seem to work right on s390x, and this is blocking testing. Might be reverted later...

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	goruntime "runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -624,9 +625,17 @@ func getDisableUDPAggregation(cl crclient.Reader) bool {
...
 	if disableUDPAggregation == "true" {
 		return true
+	} else if disableUDPAggregation == "false" {
+		return false
 	} else {
 		klog.Warningf("Ignoring unexpected udp-aggregation-config override value disable-udp-aggregation=%q", disableUDPAggregation)
 	}
+
+	// Disable on s390x because it sometimes doesn't work there; see OCPBUGS-2532
+	if goruntime.GOARCH == "s390x" {
```
martinkennelly, line 635: For other reviewers, I _think_ this `GOARCH` value is from here: https://github.com/golang/go/blob/master/src/go/build/syslist.go

martinkennelly:
/retest-required
For jobs ovn-windows and sdn-multi, it looks like image registry pull issues. Unrelated to this PR.

PR #1630, opened 2022-11-19
SDN-3597: OVN-K alerts: add OVS overflow alerts

martinkennelly:
OVS vswitchd and kernel module talk via a full duplex link comms using the netlink socket API. Reciever requires a buffer. Buffer overflow will cause packet loss.

Overflows can occur if node is under extreme load and numerous userspace flow lookups. It may also occur if OVS vswitchd is malfunctioning.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/alert-rules.yaml
+++ b/bindata/network/ovn-kubernetes/common/alert-rules.yaml
@@ -71,3 +71,21 @@ spec:
       expr: increase(ovnkube_resource_retry_failures_total[10m]) > 0
       labels:
         severity: warning
+    - alert: OVNKubernetesNodeOVSOverflowUserspace
+      annotations:
+        summary: OVS vSwitch daemon drops packets due to buffer overflow.
+        description: Netlink messages dropped by OVS vSwitch daemon due to netlink socket buffer overflow.
+          This will result in packet loss.
+      expr: increase(ovs_vswitchd_netlink_overflow[5m]) > 0
+      for: 15m
+      labels:
+        severity: warning
+    - alert: OVNKubernetesNodeOVSOverflowKernel
+      annotations:
+        summary: OVS kernel module drops packets due to buffer overflow.
+        description: Netlink messages dropped by OVS kernel module due to netlink socket buffer overflow.
+          This will result in packet loss.
+      expr: increase(ovs_vswitchd_dp_flows_lookup_lost[5m]) > 0
+      for: 15m
+      labels:
+        severity: warning


martinkennelly:
/hold 
Pending review of Adrian Ms comments via slack.

amorenoz:
@martinkennelly, do you want me to review the PR?

martinkennelly:
Talked to OVS expert Adrian Moreno regarding this since he has worked on customer issue where a workload can overwhelm a netlink socket and messages are dropped. He has graciously educated me.

Identifying that this issue is occurring is the goal of these alerts.

Some context provided by Adrian thats worth sharing. Prior to ovs < 2.16, ovs exposed a netlink socket per vport (ovs bridge port), however with following versions, it exposes a netlink socket per cpu.
When a workloads networking is being processed by cpuX, it *may* emit netlink messages to its associated netlink socket - remember each cpu has an associated netlink socket for ovs >= 2.16.
Associating the problematic workloads networking with a cpu can be done with tools. Adrian will document this either in KCS or runbook. I vote runbook and link it directly with these alerts.

So, the plan is these alerts will alert the sre/cluster admin to an issue. The investigation following this alert will be documented later.

martinkennelly:
@amorenoz Ready for review

For anyone unfamiliar with promql expressions, here is my attempt at a plain English description of both alerts.
Every 30 seconds when the alert expr is evaluated, it will look back at the last 5 mins, and if it has seen an increase in that period of the last 5 mins, the alert goes into pending stage. If this alert is consistently pending for 15mins, then a warning alert fires. This means the metrics must be consistently increasing over a period of time and not just a one off increase.

martinkennelly:
For QE, et all, I was able to induce an increase for both metrics by spawning a cluster.
Then setting flow limit to zero on _every_ node, this causes a lot of upcalls:
```
ovs-vsctl --no-wait set Open_vSwitch . other_config:flow-limit=0
```


I then executed a workload:
```
kube-burner ocp node-density-heavy
```

martinkennelly:
Interestingly, from my testing, this alert will also fire if vswitchd isnt reading the netlink messages fast enough (cpu starved).

npinaeva:
/lgtm
thanks @martinkennelly!

martinkennelly:
Note for reviewers - this PR cannot be merged with Adrian Ms lgtm.

martinkennelly:
Rebased - no changes

amorenoz:
In most cases both metrics should trigger simultaneously as counters should change at the same pace if ovs-vswitchd is running. I don't know if that's an issue in this case.

Otherwise the change looks good to me. It'll surely be of value for customers and GSS teams.

martinkennelly:
> In most cases both metrics should trigger simultaneously as counters should change at the same pace if ovs-vswitchd is running. I don't know if that's an issue in this case.
> 
> Otherwise the change looks good to me. It'll surely be of value for customers and GSS teams.

I have seen one metric increment and the other did not, that is why I split them into separate alerts and didnt OR them in one alert. I think its fine. Thanks for review Adrian.

trozet (approved):

PR #1639, opened 2022-11-24
Allow SDN migration from Kuryr to OVNKubernetes

dulek:
This PR aims to unblock checks that currently block Kuryr->OVNKubernetes migrations. Interestingly that worked just fine in 4.11, but feature migrations introduced in 4.12 started to block that. Also a validateMigration function is introduced that will
block Kuryr->OpenShiftSDN migrations that are considered not supported.

I also added error logging to feature migration code as these are not printed and I spent quite a while trying to figure out why the migration isn't progressing as it did in 4.11.


dulek:
/retest

I think mtu-migration failures are a flake.

gryf (approved):
/lgtm

PR #1648, opened 2022-12-02
OCPBUGS-4350: Fix handling of deployment and statefulset updates

kyrtapz:
When a deployment or a statefulset is being updated `UpdatedReplicas` indicates how many replicas are on the latest version. When the field was not taken into account all statefulsets and deployments were considered ready prematurely during upgrades.

cc: @zshi-redhat 
/assign @jcaamano 

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -148,6 +148,9 @@ func (status *StatusManager) SetFromPods() {
 		if isNonCritical(ss) && ss.Status.ReadyReplicas == 0 && !status.installComplete {
 			progressing = append(progressing, fmt.Sprintf("StatefulSet %q is waiting for other operators to become ready", ssName.String()))
 			ssProgressing = true
+		} else if ss.Status.UpdatedReplicas < ss.Status.Replicas {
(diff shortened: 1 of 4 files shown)


pkg/controller/statusmanager/pod_status.go
```diff
@@ -200,6 +203,9 @@ func (status *StatusManager) SetFromPods() {
 		if isNonCritical(dep) && dep.Status.UnavailableReplicas > 0 && !status.installComplete {
 			progressing = append(progressing, fmt.Sprintf("Deployment %q is waiting for other operators to become ready", depName.String()))
 			depProgressing = true
+		} else if dep.Status.UpdatedReplicas < dep.Status.Replicas {
```
zshi-redhat, line 206: @kyrtapz  Do we have the same issue with deployment? Is it becacuse the deployments are not running in the management cluster in 4.12 so we didn't see the upgrade issue?

kyrtapz: I am not sure that we saw the bug with deployments but ovn-k master statefulset would likely be the last thing to update since it has to wait for ovn-k node update. Regardless of that I think deployment update handling should also be fixed in this PR.

zshi-redhat: Yes, makes sense.
I was just to understand if there will be same issue for deployment.

PR #1649, opened 2022-12-05
OCPBUGS-1341: Enhance check controller to remove old check objects

pperiyasamy:
This would make connectivitycheck controller to use new library-go api so that stale checks objects are removed automatically as part of controller sync.

### DIFF
--- a/pkg/controller/connectivitycheck/connectivity_check_controller.go
+++ b/pkg/controller/connectivitycheck/connectivity_check_controller.go
@@ -12,10 +12,10 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/network"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
-	"github.com/openshift/api/operatorcontrolplane/v1alpha1"
 	configv1client "github.com/openshift/client-go/config/clientset/versioned"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	applyconfigv1alpha1 "github.com/openshift/client-go/operatorcontrolplane/applyconfigurations/operatorcontrolplane/v1alpha1"
 	operatorcontrolplaneclient "github.com/openshift/client-go/operatorcontrolplane/clientset/versioned"
 	"github.com/openshift/cluster-network-operator/pkg/controller/eventrecorder"
 	"github.com/openshift/library-go/pkg/controller/factory"
@@ -26,8 +26,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apiextensionsinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	applyconfigmetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
@@ -45,7 +45,6 @@ type NetworkConnectivityCheckController interface {
 // Checks between network-check-source pod and every LB
 // Checks between network-check-source pod and network-check-target service and endpoints this being managed by a Daemonset
 func NewNetworkConnectivityCheckController(
-	kubeClient kubernetes.Interface,
 	operatorClient v1helpers.OperatorClient,
 	operatorcontrolplaneClient *operatorcontrolplaneclient.Clientset,
 	apiextensionsClient *apiextensionsclient.Clientset,
@@ -91,7 +90,7 @@ func NewNetworkConnectivityCheckController(
 		nodeLister:                        kubeInformersForNamespaces.InformersFor("").Core().V1().Nodes().Lister(),
 		infrastructureLister:              configInformers.Config().V1().Infrastructures().Lister(),
 	}
-	return c.WithPodNetworkConnectivityCheckFn(generator.generate)
+	return c.WithPodNetworkConnectivityCheckApplyFn(generator.generate)
 }
 
 type networkConnectivityCheckController struct {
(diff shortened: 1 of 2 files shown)


pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -158,25 +161,25 @@ func (c *connectivityCheckTemplateProvider) generate(ctx context.Context, syncCo
...
-			WithSource("network-check-source-" + strings.Split(pod.Spec.NodeName, ".")[0])(check)
-			check.Spec.SourcePod = pod.Name
-			nodeRef := metav1.OwnerReference{
-				APIVersion: "v1",
-				Kind:       "Node",
-				UID:        node.GetUID(),
-				Name:       node.GetName(),
+		for i := range checks {
+			WithSource("network-check-source-" + strings.Split(pod.Spec.NodeName, ".")[0])(checks[i])
+			checks[i].Spec.SourcePod = &pod.Name
```
kyrtapz, line 166: I don't think that this is correct, `pod` is an iteration variable so all of the checks will point to the last pod name returned by `c.diagnosticsPodLister.List(labels.Set{"app": "network-check-source"}.AsSelector())`

kyrtapz:
Can you extract non functional changes to a separate commit? It would make reviewing significantly easier

pkg/controller/connectivitycheck/connectivity_check_controller.go
pperiyasamy: @kyrtapz This is still inside pods iteration loop and not even on another goroutine, so i don't think there would be a variable jump. do you see a problem ?

pperiyasamy:
> Can you extract non functional changes to a separate commit? It would make reviewing significantly easier

sure, done with 2 commits now.

pkg/controller/connectivitycheck/connectivity_check_controller.go
pperiyasamy: Now I got the problem what you are referring to. will update with a fix.

kyrtapz: this code is a little confusing to me, it looks like for every pod all of the checks will be modified to point to that specific pod?

pperiyasamy: Yes @kyrtapz , i guess this is fixed correctly now. please check.

flavio-fernandes:
/retest-required
TY Peri! This looks good to me, but I am not very experimented with this code. @kyrtapz @trozet PTAL

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -26,13 +26,18 @@ import (
...
 	apiextensionsinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	applyconfigmetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
 )
 
+var (
```
kyrtapz, line 37: NIT: these are used only in one function, do they need to be defined at package level?

pkg/controller/connectivitycheck/connectivity_check_template.go
```diff
@@ -28,24 +20,47 @@ func NewPodNetworkConnectivityCheckTemplate(address, namespace string, options .
...
+func WithTarget(target string) func(*applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+	return func(check *applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+		name := strings.Replace(*check.Name, "$(TARGET)", target, -1)
+		check.Name = &name
+	}
+}
+
+// DeepCopy returns clone for given check object. Now this method considers copying only
+// relevant fields used by check controller.
+func DeepCopy(check *applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) *applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration {
```
kyrtapz, line 49: This function is not really doing a deep copy, can we rename it to something more fitting?
Also, does it need to be exported?

pkg/controller/connectivitycheck/connectivity_check_template.go
```diff
@@ -28,24 +20,47 @@ func NewPodNetworkConnectivityCheckTemplate(address, namespace string, options .
...
 	}
 }
 
 // WithSource option replaces the $(SOURCE) token in the name.
-func WithSource(source string) func(*v1alpha1.PodNetworkConnectivityCheck) {
-	return func(check *v1alpha1.PodNetworkConnectivityCheck) {
-		check.Name = strings.Replace(check.Name, "$(SOURCE)", source, -1)
+func WithSource(source string) func(*applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+	return func(check *applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+		name := strings.Replace(*check.Name, "$(SOURCE)", source, -1)
```
kyrtapz, line 34: Do we need to check if `Name` is not nil?

pkg/controller/connectivitycheck/connectivity_check_template.go
```diff
@@ -28,24 +20,47 @@ func NewPodNetworkConnectivityCheckTemplate(address, namespace string, options .
...
 	}
 }
 
 // WithTarget option replaces the $(TARGET) token in the name.
-func WithTarget(target string) func(*v1alpha1.PodNetworkConnectivityCheck) {
-	return func(check *v1alpha1.PodNetworkConnectivityCheck) {
-		check.Name = strings.Replace(check.Name, "$(TARGET)", target, -1)
+func WithTarget(target string) func(*applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+	return func(check *applyconfigv1alpha1.PodNetworkConnectivityCheckApplyConfiguration) {
+		name := strings.Replace(*check.Name, "$(TARGET)", target, -1)
```
kyrtapz, line 42: Do we need to check if `Name` is not nil?

pkg/controller/connectivitycheck/connectivity_check_template.go
pperiyasamy: done

pkg/controller/connectivitycheck/connectivity_check_template.go
pperiyasamy: done

pkg/controller/connectivitycheck/connectivity_check_template.go
pperiyasamy: changed the function name and made it private. is this ok now ?

PR #1654, opened 2022-12-08
OCPBUGS-2532: fix for #1629

Prashanth684:
PR #1629 disabled udp aggregation for s390x. this PR modifies that fix to move the check to the begining of the function as the udp-aggregation-config configmap is not present by default


danwinship:
This isn't quite what I'd intended; see #1655 (and please check my logic there!)
/close

PR #1655, opened 2022-12-08
OCPBUGS-2532: Fix default disable-udp-aggregation value on s390x

danwinship:
We were bailing out early and returning false ("don't disable it") if the config file didn't exist, well before the check of "return true by default on s390x"...

(As with the previous PR, this is untested, because I don't have a mainframe lying around, so please double-check my logic...)


madeelrh:
cc @andreaskaris

PR #1660, opened 2022-12-09
[release-4.12] OCPBUGS-4686: Revert "Remove references to the hosts kubeconfig"

kyrtapz:
There are OpenShift environments that do not use MCO and rely on kubecfg being available on the host.
Revert the removal of the kubeconfig fallback to address that.

/assign @trozet @jcaamano 

This reverts commit a737ab6d1e523bcb9413f72dd0935bb078ff9019.

### DIFF
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -4,21 +4,28 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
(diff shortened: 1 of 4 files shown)


relyt0925:
Thank you so much for your help here!!!! IBM Cloud ROKS team is working on adding this to the 4.12 candidate so we can ensure we are prepped for this to go into 4.13

PR #1662, opened 2022-12-12
OCPBUGS-4746: Revert "Remove references to the hosts kubeconfig"

kyrtapz:
There are OpenShift environments that do not use MCO and rely on kubecfg being available on the host.
Revert the removal of the kubeconfig fallback to address that.

/assign @trozet @jcaamano 
This reverts commit 14ac289f2b9f52430a6eb7f868ebc8a564df75a5.

### DIFF
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -4,21 +4,28 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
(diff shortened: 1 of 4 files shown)


kyrtapz:
/close
after talking to Jaime we decided we should remove the fallback in 4.13 as not to keep the dead code(ROKS will handle the creation of apiserver-url.env in 4.13)

PR #1665, opened 2022-12-13
OCPBUGS-3272: Unhealthy Readiness Probe failing ci

bpickard22:
When a cluster is created, it typically takes ~2 minutes for  the databases to start up and form the Raft cluster. Before this occurs, there will be no DB endpointsm so the node containers that are waiting for them and failing readiness tests will cause alerts, which cause ci failures / flakes

We can avoid this problem by waiting 2 minutes to start the liveness probe, and reducing its polling rate

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -570,7 +570,7 @@ spec:
           exec:
             command: ["test", "-f", "/etc/cni/net.d/10-ovn-kubernetes.conf"]
           initialDelaySeconds: 5
-          periodSeconds: 5
+          periodSeconds: 30
       {{- if .OVNPlatformAzure}}
       - name: drop-icmp
         image: "{{.OvnImage}}"
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
@@ -465,7 +465,7 @@ spec:
           exec:
             command: ["test", "-f", "/etc/cni/net.d/10-ovn-kubernetes.conf"]
           initialDelaySeconds: 5
-          periodSeconds: 5
+          periodSeconds: 30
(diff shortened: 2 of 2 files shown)


bpickard22:
@dcbw PTAL

jluhrsen:
IIRC, CI will fail if we get 20+ of these alerts. So, I think as long as you make sure the polling delay and intervals would ensure we don't get 20+ alerts while a cluster is coming up you should be able to avoid the CI failures.

bpickard22:
Another possible solution here would be leaving the initial delay at 5 seconds but reducing the polling rate to around 30 seconds to help with the case when the container restarts but the dbs are still healthy

flavio-fernandes (commented):
/lgtm

trozet (changes requested):
don't you also need this change in the managed path as well?

jcaamano:
@bpickard22 we had discussed at one point in out meeting to increase `periodSeconds` but leave `initialDelaySeconds` as is; as to not unnecessarily delay start (even though it might take 2 minutes for things to be ready, maybe one day in the future it takes less time). We are probably talking about 1 probe failure difference, between 4 or 5 during those 2 minutes.
Do you know what is the alert threshold? How many probes have to fail to get the alert?
Also you mention in the PR message `waiting 2 minutes to start the liveness probe` but is really 30 seconds, right? Just crosschecking.

bpickard22:
@jcaamano i forgot to correct the commit, message, I will take care of that, and I have changed the intialDelaySeconds back to 5 seconds.  The alert threshold is 20 failures (from here) https://github.com/openshift/origin/blob/2154d73eed665b1a0951f59ee638e52242d64a6c/pkg/synthetictests/duplicated_events.go#L25) so as long as we stay under that we are good.

bpickard22:
@trozet made the change in the managed path

flavio-fernandes:
@bpickard22 the job e2e-network-migration has not passed since Jan/12/23.
Maybe time ask for an override and approve?

dcbw:
unrelated to this change

PR #1669, opened 2022-12-16
Fix CNO crashing when Kuryr without MTU is set

dulek:
2cb56d8176756e57dd95d4940c620e88cae02342 moved around code setting default value for MTU when Kuryr is configured. The problem is that it also stopped to set it, which made CNO unable to update the .status on Network CR and the kube-apiserver-operator from correctly populating API IP in the certs.

This commit makes sure we always populate KuryrConfig.MTU with a value.


dulek:
/test e2e-openstack-kuryr

I tried locally again and it worked, so let's see.

dulek:
/assign @dcbw

Please take a look, this unblocks 4.13 with Kuryr and is a fairly obvious mistake.

dulek:
/retest-required
Kuryr job passed!

PR #1676, opened 2023-01-11
ovn-kubernetes: Allow node_mgmt_port_netdev_flags for non-DPU modes

wizhaoredhat:
The node_mgmt_port_netdev_flags is a feature flag of ovn-kubernetes and shouldn't be restricted to any particular modes of operation.

Signed-off-by: William Zhao <wizhao@redhat.com>

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -446,9 +446,9 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            ${node_mgmt_port_netdev_flags} \
             {{- if eq .OVN_NODE_MODE "dpu-host" }}
             --ovnkube-node-mode dpu-host \
-            ${node_mgmt_port_netdev_flags} \
             {{- end }}
             --metrics-bind-address "127.0.0.1:29103" \
             --ovn-metrics-bind-address "127.0.0.1:29105" \
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
@@ -353,9 +353,9 @@ spec:
             --loglevel "${OVN_KUBE_LOG_LEVEL}" \
             --inactivity-probe="${OVN_CONTROLLER_INACTIVITY_PROBE}" \
             ${gateway_mode_flags} \
+            ${node_mgmt_port_netdev_flags} \
             {{- if eq .OVN_NODE_MODE "dpu-host" }}
             --ovnkube-node-mode dpu-host \
-            ${node_mgmt_port_netdev_flags} \
             {{- end }}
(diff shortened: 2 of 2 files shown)


zshi-redhat:
For reviewers:  the `node_mgmt_port_netdev_flags` is currently enabled by adding `OVNKUBE_NODE_MGMT_PORT_NETDEV` env var in the `env-overrides` configmap. It is a dev preview feature for DPU and tech preview feature for NIC (e.g. NVIDIA ConnectX-6).

trozet:
@jluhrsen do we have a bug for the migration job perma fail?

trozet (commented):
This seems to contradict what is in ovnk?
https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L1259

```
	&cli.StringFlag{
		Name: "ovnkube-node-mgmt-port-netdev",
		Usage: "valid only when ovnkube-node-mode is either dpu or dpu-host. " +
			"when provided, use this netdev as management port. it will be renamed to ovn-k8s-mp0 " +
			"and used to allow host network services and pods to access k8s pod and service networks. ",
		Value:       OvnKubeNode.MgmtPortNetdev,
		Destination: &cliConfig.OvnKubeNode.MgmtPortNetdev,
```

I guess this PR is to enable:
https://github.com/ovn-org/ovn-kubernetes/pull/3160
https://github.com/ovn-org/ovn-kubernetes/issues/3159

I'm a little confused about the real use case here with using full mode and netdev mgmt port.

jluhrsen:
> @jluhrsen do we have a bug for the migration job perma fail?

just created it @trozet 

https://issues.redhat.com/browse/SDN-3681

wizhaoredhat:
> This seems to contradict what is in ovnk? https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L1259
> 
> ```
> 	&cli.StringFlag{
> 		Name: "ovnkube-node-mgmt-port-netdev",
> 		Usage: "valid only when ovnkube-node-mode is either dpu or dpu-host. " +
> 			"when provided, use this netdev as management port. it will be renamed to ovn-k8s-mp0 " +
> 			"and used to allow host network services and pods to access k8s pod and service networks. ",
> 		Value:       OvnKubeNode.MgmtPortNetdev,
> 		Destination: &cliConfig.OvnKubeNode.MgmtPortNetdev,
> ```
> 
> I guess this PR is to enable: [ovn-org/ovn-kubernetes#3160](https://github.com/ovn-org/ovn-kubernetes/pull/3160) [ovn-org/ovn-kubernetes#3159](https://github.com/ovn-org/ovn-kubernetes/issues/3159)
> 
> I'm a little confused about the real use case here with using full mode and netdev mgmt port.

@trozet That string flag is definitely an oversight. It should include a description when using it in Full mode.

Yes it is to enable those list of PRs you mentioned. The real use case for Full mode is to offload "some" traffic that goes through the mgmt port in the case when host pods tries to communicated with pods on the same node. It is a very limited use case for Full mode. But we are using Full mode as an easy way to prototype changes for DPU mode. This PR just makes it easier for us to debug stuff in Full mode. We actually have to design and implement a way to easily reserve VFs for system host pods (e.g. ovnkube-node).

trozet (approved):
/lgtm

PR #1677, opened 2023-01-12
OCPBUGS-4417: Added missing API field podref to OverlappingRangeIPReservation CRD

nicklesimba:
The OverlappingRangeIPReservation CRD was missing the podref field. Because of this, an issue occurs when the ip-reconciler code from whereabouts get executed, where the ip-reconciler will incorrectly view a pod as not alive, and consequently incorrectly delete the OverlappingRangeIPReservation object from kubernetes. 

I updated the CRD to include the podref field.

Signed-off-by: nicklesimba <simha.nikhil@gmail.com>

### DIFF
--- a/bindata/network/additional-networks/crd/001-crd.yaml
+++ b/bindata/network/additional-networks/crd/001-crd.yaml
@@ -143,6 +143,8 @@ spec:
             properties:
               containerid:
                 type: string
+              podref:
+                type: string
             required:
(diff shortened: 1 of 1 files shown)


nicklesimba:
I made a dummy PR with no changes to see if network-migration job will fail there - to isolate if the failure is due to my PR or not. Waiting on results...

PR #1678, opened 2023-01-12
OSASINFRA-3075 - sync_vips: allow external LB on OpenStack platform

EmilienM:
If the VIPs aren't defined in the PlatformStatus, don't try to sync the
vips, they don't exist.


EmilienM:
I'm closing it because for now we don't need this one anymore, see https://github.com/openshift/enhancements/pull/1322

PR #1679, opened 2023-01-12
OCPBUGS-5802: Update github.com/Masterminds/sprig to v3

bn222:
Sprig is a dependency of cno which is in turn a dependency of multiple projects while the old sprig has a vulnerability.


jcaamano:

not related to this PR

PR #1688, opened 2023-01-18
sdn: stop mounting host's / in sdn container

dcbw:
With https://github.com/openshift/sdn/pull/496 merged we no longer need host's / in the container.


danwinship:
/retest

`F0118 22:29:20.984819   58103 cmd.go:118] Failed to start sdn: failed to set up iptables: failed to ensure chain OPENSHIFT-NOTRACK exists: error creating chain \"OPENSHIFT-NOTRACK\": exit status 125: chroot: cannot change root directory to '/host': No such file or directory`

looks like it didn't pick up the new sdn image

PR #1695, opened 2023-01-26
Fix/Update ovn-kind-cno.sh script

jluhrsen:
4.12 uses k8s 1.25 and although 1.25.6 is the most recent k8s version of 1.25, the most recent kindest/node tag is based on 1.25.3

also cleans up a few small typo/grammar things


jluhrsen:
/hold

this doesn't work yet. not sure how many more things will need to change.

jluhrsen:
this script is probably just dead. might make sense to just remove so it doesn't send someone else down a dead end road

PR #1696, opened 2023-01-31
Make the cno react to hybrid overlay toggling

JacobTanenbaum:
the ability to toggle the hybrid overlay has been added to ovn-kubernetes, this will make cno roll out the changes automatically

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
@@ -849,6 +849,10 @@ spec:
             fi
           done
 
+          if [ "{{.OVNHybridOverlayEnable}}" == "true" ]; then
+            echo "HybridOverlay enabled"
+          fi
+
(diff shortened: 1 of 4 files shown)


kyrtapz:
Unfortunately I do not know the historic reasons for not restarting ovn-k on configmap changes but I cannot think of a scenario where we want to change the config without applying it straight away.
Please note that we use one configmap for both ovnk-master and ovnk-node, this means that we would always restart both even if the config change only applied to one.

For now, I agree that it would be good to use the same approach(annotation) so it makes it a little easier if we want to change it in the future.

jcaamano:
> Please note that we use one configmap for both ovnk-master and ovnk-node, this means that we would always restart both even if the config change only applied to one.

True but I have the feeling that the correct procedure to avoid that is having (and documenting) different configurations for each instead of white-boxing from CNO which specific settings affect which component. 

> For now, I agree that it would be good to use the same approach(annotation) so it makes it a little easier if we want to change it in the future.

Let's see what other people think in the meeting.

JacobTanenbaum:
closing in favor of https://github.com/openshift/cluster-network-operator/pull/1709

PR #1699, opened 2023-02-01
ovn-k, multi-homing: enable the feature

maiqueb:
The OVN-Kubernetes feature will be enabled whenever the `conf.DisableMultiNetwork` knob is disabled.

Manually testing deploying this PR w/ the following YAML gets us two pods interconnected via this cluster-wide logical switch on upcoming openshift-4.13:
```yaml
---
apiVersion: k8s.cni.cncf.io/v1
metadata:
  name: flatl2
spec:
    config: '{                                                                           
        "cniVersion": "0.4.0",
        "name": "flatl2",
        "netAttachDefName": "default/flatl2",
        "topology": "layer2",
        "type": "ovn-k8s-cni-overlay",                                          
        "subnets": "192.168.100.0/24",                                                          
        "logFile": "/var/log/ovn-kubernetes/ovn-k8s-cni-overlay.log",           
        "logLevel": "5",                                                        
        "logfile-maxsize": 100,                                                 
        "logfile-maxbackups": 5,                                                
        "logfile-maxage": 5                                                     
    }'
---
apiVersion: v1
kind: Pod
metadata:
  name: pod1
  annotations:
    k8s.v1.cni.cncf.io/networks: 'flatl2'
spec:
  containers:
  - name: pod1
    image: k8s.gcr.io/e2e-test-images/agnhost:2.26
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    securityContext:
      runAsUser: 1000
      privileged: false
      seccompProfile:
        type: RuntimeDefault
      capabilities:
        drop: ["ALL"]
      runAsNonRoot: true
      allowPrivilegeEscalation: false
---
apiVersion: v1
kind: Pod
metadata:
  name: pod2
  annotations:
    k8s.v1.cni.cncf.io/networks: 'flatl2'
spec:
  containers:
  - name: pod2
    image: k8s.gcr.io/e2e-test-images/agnhost:2.26
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    securityContext:
      runAsUser: 1000
      privileged: false
      seccompProfile:
        type: RuntimeDefault
      capabilities:
        drop: ["ALL"]
      runAsNonRoot: true
      allowPrivilegeEscalation: false
```

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -431,6 +431,11 @@ spec:
             node_mgmt_port_netdev_flags="--ovnkube-node-mgmt-port-netdev ${OVNKUBE_NODE_MGMT_PORT_NETDEV}"
           fi
 
+          multi_network_enabled_flag=
+          if [[ "{{.OVN_MULTI_NETWORK_ENABLE}}" == "true" ]]; then
+            multi_network_enabled_flag="--enable-multi-network"
+          fi
+
           exec /usr/bin/ovnkube --init-node "${K8S_NODE}" \
             --nb-address "{{.OVN_NB_DB_ENDPOINT}}" \
             --sb-address "{{.OVN_SB_DB_ENDPOINT}}" \
@@ -456,6 +461,7 @@ spec:
             --export-ovs-metrics \
             --disable-snat-multiple-gws \
             ${export_network_flows_flags} \
+            ${multi_network_enabled_flag} \
             ${gw_interface_flag}
         env:
         # for kubectl
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
@@ -828,6 +828,11 @@ spec:
             exit 1
           fi
 
+          multi_network_enabled_flag=
+          if [[ "{{.OVN_MULTI_NETWORK_ENABLE}}" == "true" ]]; then
+            multi_network_enabled_flag="--enable-multi-network"
+          fi
+
           echo "I$(date "+%m%d %H:%M:%S.%N") - ovnkube-master - start ovnkube --init-master ${K8S_NODE}"
           exec /usr/bin/ovnkube \
             --init-master "${K8S_NODE}" \
@@ -850,6 +855,7 @@ spec:
             --nb-cert-common-name "{{.OVN_CERT_CN}}" \
             --enable-multicast \
             --disable-snat-multiple-gws \
+            ${multi_network_enabled_flag} \
             --acl-logging-rate-limit "{{.OVNPolicyAuditRateLimit}}"
         volumeMounts:
         # for checking ovs-configuration service
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -351,6 +351,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["NorthdThreads"] = 4
 	}
 
+	data.Data["OVN_MULTI_NETWORK_ENABLE"] = true
+	if conf.DisableMultiNetwork != nil && *conf.DisableMultiNetwork {
+		data.Data["OVN_MULTI_NETWORK_ENABLE"] = false
+	}
+
 	var manifestSubDir string
 	manifestDirs := make([]string, 0, 2)
 	manifestDirs = append(manifestDirs, filepath.Join(manifestDir, "network/ovn-kubernetes/common"))
--- a/bindata/network/ovn-kubernetes/common/003-rbac-controller.yaml
+++ b/bindata/network/ovn-kubernetes/common/003-rbac-controller.yaml
@@ -117,6 +117,11 @@ rules:
 - apiGroups: ['authorization.k8s.io']
   resources: ['subjectaccessreviews']
   verbs: ['create']
+- apiGroups:
+  - k8s.cni.cncf.io
+  resources:
+  - network-attachment-definitions
+  verbs: ["list", "get", "watch"]
 
 
 ---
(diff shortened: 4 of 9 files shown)


kyrtapz:
Does this only change the configmap? it would mean the changing the value of DisableMultiNetwork wouldn't affect anything straight away since ovn-k doesn't react to configmap changes in runtime.

A possible solution might be to do something similar to what Jacob did here: https://github.com/openshift/cluster-network-operator/pull/1696/files

bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
```diff
@@ -456,6 +462,7 @@ spec:
             --export-ovs-metrics \
             --disable-snat-multiple-gws \
             ${export_network_flows_flags} \
+            ${multi_network_enabled_flag} \
```
kyrtapz, line 465: This flag is redundant right? You enable multi-network in the configmap anyway

maiqueb: I honestly don't know ... 

I thought I had to update the command. Are you saying that just by `echo`ing the string above it'll cause a rollout ? 

That explains the PR you showed me !!!

kyrtapz: yup, if the spec changes then a rollout will happen :)

maiqueb:
@trozet / @zshi-redhat do we want to enable this also for microshift ?

bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
maiqueb: Thank you for leading me to the correct solution; should have understood it the first time.

zshi-redhat:
> @trozet / @zshi-redhat do we want to enable this also for microshift ?

@maiqueb  Not this time, multus is not yet supported in microshift :-)

jcaamano:
Worth considering: 
https://github.com/openshift/cluster-network-operator/pull/1696#issuecomment-1424114088

maiqueb:
> Worth considering: [#1696 (comment)](https://github.com/openshift/cluster-network-operator/pull/1696#issuecomment-1424114088)

I'm OK with following your suggestion if we align on it across the board.

pkg/network/ovn_kubernetes.go
```diff
@@ -349,6 +349,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["NorthdThreads"] = 4
 	}
 
+	if conf.DisableMultiNetwork != nil && !*conf.DisableMultiNetwork {
+		data.Data["OVN_MULTI_NETWORK_ENABLE"] = true
+	} else {
+		data.Data["OVN_MULTI_NETWORK_ENABLE"] = false
+	}
+
```
jcaamano, line 357: nit: double negation

```
	if conf.DisableMultiNetwork != nil && *conf.DisableMultiNetwork {
		data.Data["OVN_MULTI_NETWORK_ENABLE"] = false
	} else {
		data.Data["OVN_MULTI_NETWORK_ENABLE"] = true
	}
```

maiqueb: fair enough.

Would it make more sense to have @jcaamano ? 
```golang
data.Data["OVN_MULTI_NETWORK_ENABLE"] = true
if conf.DisableMultiNetwork != nil && *conf.DisableMultiNetwork {
    data.Data["OVN_MULTI_NETWORK_ENABLE"] = false
}
```

maiqueb: Done.

trozet:
lgtm but will defer to @jcaamano to give final approval

/approve

bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
```diff
@@ -828,6 +828,12 @@ spec:
             exit 1
           fi
 
+          multi_network_enabled_flag=
+          if [[ "{{.OVN_MULTI_NETWORK_ENABLE}}" == "true" ]]; then
+            multi_network_enabled_flag="--enable-multi-network"
+          fi
+          echo "multi_network_enabled_flag=${multi_network_enabled_flag}"
+
```
jcaamano, line 836: I think for now we should pass the flag to the binary (instead of the echo) like we do with the gateway_mode_flags.

maiqueb: that was my original implementation, in https://github.com/openshift/cluster-network-operator/pull/1699/files#diff-a5b1a97252b229ace4443a7d954a6976d02cc4a6a09b78e3a1c820aaec48fe12

I was told that was redundant, and to align with what it is currently being done for hybrid overlay. 

I really don't care, it can go either way as far as I'm concerned, but seems aligning across all features is not possible.

jcaamano: It is true that passing the flag to the binary is redundant as we already have it in the config map.

Similar things could be said about having a meaningless echo on the script though.

We are already redundantly passing config flags for gateway mode, maybe others. I think that from the script standpoint itself is what makes more sense until we have a better way of doing this.

@kyrtapz WDYT?

kyrtapz: Currently there are several ways we use to trigger the rollout, this needs a refactor across CNO even if we do not want to rollout on every config change. This should be done separately and we should not hold this PR for it.
@maiqueb When I suggested that this flag is redundant I assumed you will get rid of it(together with it's definition) for something else. 

> We are already redundantly passing config flags for gateway mode, maybe others. I think that from the script standpoint itself is what makes more sense until we have a better way of doing this.

Let's go with that for now and we will just include `multi_network_enabled_flag` in the wider refactor in the future.

jcaamano:
/lgtm
/hold

Will do a final check with @trozet today to make sure we are on the same page

PR #1707, opened 2023-02-07
Allow cidr expansion

jluhrsen:
- only the mask can change and must become smaller (allowing for more
  ip addresses across the entire cluster)
- only supported for OVNKuberentes

### DIFF
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -71,6 +71,10 @@ const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migrati
 // to indicate the current IP Family mode of the cluster: "single-stack" or "dual-stack"
 const NetworkIPFamilyModeAnnotation = "networkoperator.openshift.io/ip-family-mode"
 
+// ClusterNetworkCIDRsAnnotation is an annotation on the OVN networks.operator.openshift.io daemonsets
+// to indicate the current list of clusterNetwork CIDRs available to the cluster.
+const ClusterNetworkCIDRsAnnotation = "networkoperator.openshift.io/cluster-network-cidr"
+
 // OVNRaftClusterInitiator is an annotation on the networks.operator.openshift.io CR to indicate
 // which node IP was the raft cluster initiator. The NB and SB DB will be initialized by the same member.
 const OVNRaftClusterInitiator = "networkoperator.openshift.io/ovn-cluster-initiator"
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -420,6 +420,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		return nil, progressing, errors.Wrapf(err, "failed to set IP family %s annotation on daemonsets or statefulsets", ipFamilyMode)
 	}
 
+	// logic to pretty print the clusterNetwork CIDR (possibly only one) in its annotation
+	var clusterNetworkCIDRs []string
+	for _, c := range conf.ClusterNetwork {
+		clusterNetworkCIDRs = append(clusterNetworkCIDRs, c.CIDR)
+	}
+
+	err = setOVNObjectAnnotation(objs, names.ClusterNetworkCIDRsAnnotation, strings.Join(clusterNetworkCIDRs, ","))
+	if err != nil {
+		return nil, progressing, errors.Wrapf(err, "failed to set %s annotation on daemonsets or statefulsets", clusterNetworkCIDRs)
+	}
+
 	// don't process upgrades if we are handling a dual-stack conversion.
 	if updateMaster && updateNode {
 		updateNode, updateMaster = shouldUpdateOVNKonUpgrade(bootstrapResult.OVN, os.Getenv("RELEASE_VERSION"))
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -325,7 +326,9 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 		return nil
 	}
 
-	// Currently the only change we allow is switching to/from dual-stack.
+	// Currently the only changes we allow are:
+	//   -  switching to/from dual-stack.
+	//   -  ClusterNetwork modification (check isClusterNetworkChangeSafe()) for what is supported
 
 	// validateIPPools() will have ensured that each config is independently either
 	// a valid single-stack config or a valid dual-stack config. Make sure we have
@@ -338,13 +341,13 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	case len(prev.ServiceNetwork) > len(next.ServiceNetwork):
 		// Going from dual to single
 		dualStack, singleStack = prev, next
+	case !reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork):
+		// If the ServiceNetwork has changed, but it's not part of a single<->dual stack migration
+		// then we do not support
+		return errors.Errorf("unsupported change to ServiceNetwork")
 	default:
-		// They didn't change single-vs-dual
-		if reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork) {
-			return errors.Errorf("unsupported change to ClusterNetwork")
-		} else {
-			return errors.Errorf("unsupported change to ServiceNetwork")
-		}
+		// this is not a single/dual stack migration; check if the clusterNetwork change is ok
+		return isClusterNetworkChangeSafe(prev, next)
 	}
 
 	// Validate that this is either a BareMetal or None PlatformType. For all other
@@ -382,6 +385,61 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 		}
 	}
 
+	return nil
+}
+
+func isClusterNetworkChangeSafe(prev, next *operv1.NetworkSpec) error {
+
+	// quick check to make sure clusterNetwork slices are of same size as we do not
+	// support adding/removing additional clusterNetwork entries unless it's for a
+	// single/dual stack migration. in those cases validation is done in isNetworkChangeSafe()
+	if len(prev.ClusterNetwork) != len(next.ClusterNetwork) {
+		return errors.Errorf("adding/removing clusterNetwork entries of the same type is not supported")
+	}
+
+	// Only support changing ClusterNetwork CIDR if it's OVNK
+	if !reflect.DeepEqual(next.DefaultNetwork.Type, operv1.NetworkTypeOVNKubernetes) {
+		return errors.Errorf("network type is %v. changing clusterNetwork entries is only supported for OVNKubernetes", next.DefaultNetwork.Type)
+	}
+
+	// sort prev and next just in case there was some re-ordering of the slice, since we
+	// want to know for sure we are comparing the same elements in each below
+	sort.SliceStable(prev.ClusterNetwork, func(i, j int) bool {
+		return prev.ClusterNetwork[i].CIDR < prev.ClusterNetwork[j].CIDR
+	})
+	sort.SliceStable(next.ClusterNetwork, func(i, j int) bool {
+		return next.ClusterNetwork[i].CIDR < next.ClusterNetwork[j].CIDR
+	})
+
+	// since we do not allow the clusterNetwork[] size to change, it should be safe to compare
+	// prev[i] to next[i] in this validation
+	for i, e := range prev.ClusterNetwork {
+		prevIp, prevMask, err := net.ParseCIDR(e.CIDR)
+		if err != nil {
+			return errors.Errorf("error parsing CIDR from ClusterNetwork entry %s: %v", e.CIDR, err)
+		}
+		nextIp, nextMask, err := net.ParseCIDR(next.ClusterNetwork[i].CIDR)
+		if err != nil {
+			return errors.Errorf("error parsing CIDR from ClusterNetwork entry %s: %v", next.ClusterNetwork[i].CIDR, err)
+		}
+		prevHostPrefix := e.HostPrefix
+		nextHostPrefix := next.ClusterNetwork[i].HostPrefix
+
+		// changing hostPrefix is not allowed
+		if prevHostPrefix != nextHostPrefix {
+			return errors.Errorf("modifying a clusterNetwork's hostPrefix value is unsupported")
+		}
+
+		if !prevIp.Equal(nextIp) {
+			return errors.Errorf("modifying IP network value for clusterNetwork CIDR is unsupported")
+		}
+
+		prevMaskSize, _ := prevMask.Mask.Size()
+		nextMaskSize, _ := nextMask.Mask.Size()
+		if prevMaskSize < nextMaskSize {
+			return errors.Errorf("reducing IP range with a larger CIDR mask for clusterNetwork CIDR is unsupported")
+		}
+	}
 	return nil
 }
 
--- a/pkg/network/render_test.go
+++ b/pkg/network/render_test.go
@@ -2,16 +2,16 @@ package network
 
 import (
 	"fmt"
-	"strings"
-	"testing"
-
 	. "github.com/onsi/gomega"
+	"github.com/openshift/cluster-network-operator/pkg/client/fake"
+	"k8s.io/client-go/kubernetes/scheme"
+	"strings"
+
 	configv1 "github.com/openshift/api/config/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
-	"github.com/openshift/cluster-network-operator/pkg/client/fake"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/kubernetes/scheme"
+	"testing"
 )
 
 // NOTE: IsChangeSafe() requires you to have called Validate() beforehand, so we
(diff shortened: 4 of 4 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -415,6 +415,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	var cidrs string
+	for _, c := range conf.ClusterNetwork {
+		clusterNetworkCidrs = append(clusterNetworkCidrs, c.CIDR)
+	}
+	if len(conf.ClusterNetwork) == 1 {
+		cidrs = clusterNetworkCidrs[0]
+	} else {
+		cidrs = strings.Join(clusterNetworkCidrs, ", ")
+	}
+
```
kyrtapz, line 429: there is no need to check the length, it should be enough to call:
`err = setOVNObjectAnnotation(objs, names.ClusterNetworkCidr, strings.Join(clusterNetworkCidrs, ",")`
Additionally, I think we should ditch the space in `, `

pkg/network/ovn_kubernetes.go
```diff
@@ -415,6 +415,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	err = setOVNObjectAnnotation(objs, names.ClusterNetworkCidr, cidrs)
+	if err != nil {
+		return nil, progressing, errors.Wrapf(err, "failed to set ClusterNetworkCidr annotation on daemonsets or statefulsets")
```
kyrtapz, line 432: NIT: instead of `ClusterNetworkCidr` use `%s` with `ClusterNetworkCidr` arg

pkg/names/names.go
```diff
@@ -71,6 +71,10 @@ const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migrati
 // to indicate the current IP Family mode of the cluster: "single-stack" or "dual-stack"
 const NetworkIPFamilyModeAnnotation = "networkoperator.openshift.io/ip-family-mode"
 
+// NetworkIPFamilyAnnotation is an annotation on the OVN networks.operator.openshift.io daemonsets
```
kyrtapz, line 74: Please fix the comment

pkg/network/render.go
```diff
@@ -338,13 +338,16 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	case len(prev.ServiceNetwork) > len(next.ServiceNetwork):
 		// Going from dual to single
 		dualStack, singleStack = prev, next
+	case !reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork):
+		return errors.Errorf("unsupported change to ServiceNetwork")
+	case validateCidrExpansion(prev.ClusterNetwork, next.ClusterNetwork) &&
```
kyrtapz, line 343: Can this be moved to `IsChangeSafe` instead? it looks like a better place to verify changes

pkg/network/render.go
```diff
@@ -385,6 +388,46 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+	// quick check to make sure clusterNetwork slices are of same size as we do not
+	// support adding/removing additional clusterNetwork entries
+	if len(prev) != len(next) {
+		log.Printf("Adding/Removing clusterNetwork entries of the same type is not supported")
+		return false
+	}
+
+	// CIDR Mask expansion is only supported for OVNK.
+
+	// since we do not allow the clusterNetwork[] size to change, it should be safe to compare
```
kyrtapz, line 402: I am not sure if that is a safe assumption, users could potentially reorder `clusterNetwork`?

pkg/network/render.go
```diff
@@ -385,6 +388,46 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	return nil
 }
 
+func validateCidrExpansion(prev, next []operv1.ClusterNetworkEntry) bool {
```
kyrtapz, line 391: Can we rename that to `isClusterNetworkChangeSafe`?

pkg/network/render.go
```diff
@@ -338,13 +338,16 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	case len(prev.ServiceNetwork) > len(next.ServiceNetwork):
 		// Going from dual to single
 		dualStack, singleStack = prev, next
+	case !reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork):
+		return errors.Errorf("unsupported change to ServiceNetwork")
+	case validateCidrExpansion(prev.ClusterNetwork, next.ClusterNetwork) &&
+		reflect.DeepEqual(next.DefaultNetwork.Type, operv1.NetworkTypeOVNKubernetes):
```
kyrtapz, line 344: `reflect.DeepEqual(next.DefaultNetwork.Type, operv1.NetworkTypeOVNKubernetes)` should be a part of the function we call to verify whether the clusternetwork change is safe

kyrtapz:
verify-gofmt failed: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1707/pull-ci-openshift-cluster-network-operator-master-verify/1625601442792345600

pkg/network/ovn_kubernetes.go
jluhrsen: seems like overkill, but in the case with just one CIDR, it prints with an ugly trailing `,`

I did get rid of the space after the comma

pkg/network/ovn_kubernetes.go
jluhrsen: done

pkg/names/names.go
jluhrsen: done

pkg/network/render.go
jluhrsen: good catch. next update will sort prev and next each time to make it deterministic

pkg/network/render.go
jluhrsen: yup. done

pkg/network/render.go
jluhrsen: sounds good. done.

pkg/network/ovn_kubernetes.go
kyrtapz: `strings.Join` would not print a sperator if the array is one element:
https://go.dev/play/p/waewUDXbb5O?v=goprev

pkg/names/names.go
```diff
@@ -71,6 +71,10 @@ const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migrati
 // to indicate the current IP Family mode of the cluster: "single-stack" or "dual-stack"
 const NetworkIPFamilyModeAnnotation = "networkoperator.openshift.io/ip-family-mode"
 
+// ClusterNetworkCidr is an annotation on the OVN networks.operator.openshift.io daemonsets
+// to indicate the current list of clusterNetwork CIDRs available to the cluster.
+const ClusterNetworkCidr = "networkoperator.openshift.io/cluster-network-cidr"
```
kyrtapz, line 76:
```suggestion
const ClusterNetworkCIDRAnnotation = "networkoperator.openshift.io/cluster-network-cidr"
```

pkg/network/ovn_kubernetes.go
```diff
@@ -420,6 +420,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		return nil, progressing, errors.Wrapf(err, "failed to set IP family %s annotation on daemonsets or statefulsets", ipFamilyMode)
 	}
 
+	// logic to pretty print the clusterNetwork CIDR (possibly only one) in its annotation
+	var clusterNetworkCidrs []string
```
kyrtapz, line 424: NIT
```suggestion
	var clusterNetworkCIDRs []string
```

pkg/network/render.go
```diff
@@ -338,13 +339,15 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	case len(prev.ServiceNetwork) > len(next.ServiceNetwork):
 		// Going from dual to single
 		dualStack, singleStack = prev, next
+	case !reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork):
+		return errors.Errorf("unsupported change to ServiceNetwork")
+	case isClusterNetworkChangeSafe(prev, next):
+		// this was a CIDR Expansion change not a single/dual stack change. can just get out of here now.
```
kyrtapz, line 345: If `isClusterNetworkChangeSafe` changes to include more/other verification this comment would no longer be valid

pkg/network/render.go
```diff
@@ -338,13 +339,15 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	case len(prev.ServiceNetwork) > len(next.ServiceNetwork):
 		// Going from dual to single
 		dualStack, singleStack = prev, next
+	case !reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork):
+		return errors.Errorf("unsupported change to ServiceNetwork")
+	case isClusterNetworkChangeSafe(prev, next):
```
kyrtapz, line 344: I don't think `isClusterNetworkChangeSafe` should be a part of this switch statement, can we move that before checking `ServiceNetwork` changes and only return if it fails?
Also, please update the comments:
```
// Currently the only change we allow is switching to/from dual-stack.
// validateIPPools() will have ensured that each config is independently either
// a valid single-stack config or a valid dual-stack config. Make sure we have
// one of each.
```

pkg/network/render.go
```diff
@@ -385,6 +388,59 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	return nil
 }
 
+func isClusterNetworkChangeSafe(prev, next *operv1.NetworkSpec) bool {
```
kyrtapz, line 391: can we stick to `isXXXChangeSafe(...) error` signature? We can then just return a appropriate error instead of just logging.

pkg/network/render.go
```diff
@@ -385,6 +388,59 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+
```
kyrtapz, line 392: I think it would be good for clarity to remove the first part of the following statement from `isNetworkChangeSafe`:
```
if reflect.DeepEqual(prev.ClusterNetwork, next.ClusterNetwork) && reflect.DeepEqual(prev.ServiceNetwork, next.ServiceNetwork) {
		return nil
}
```
and move `if reflect.DeepEqual(prev.ClusterNetwork, next.ClusterNetwork) {return nil}`  to `isClusterNetworkChangeSafe`

pkg/network/render.go
```diff
@@ -385,6 +388,59 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+	// sort prev and next just in case there was some re-ordering of the slice, since we
+	// want to know for sure we are comparing the same elements in each below
+	sort.SliceStable(prev.ClusterNetwork, func(i, j int) bool {
+		return prev.ClusterNetwork[i].CIDR < prev.ClusterNetwork[j].CIDR
+	})
+	sort.SliceStable(next.ClusterNetwork, func(i, j int) bool {
+		return next.ClusterNetwork[i].CIDR < next.ClusterNetwork[j].CIDR
+	})
+
+	// since we do not allow the clusterNetwork[] size to change, it should be safe to compare
```
kyrtapz, line 415: `clusterNetwork` length can change in conversion from/to dual-stack https://docs.openshift.com/container-platform/4.12/networking/ovn_kubernetes_network_provider/converting-to-dual-stack.html

pkg/network/render.go
```diff
@@ -385,6 +388,59 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+	// prev[i] to next[i] in this validation
+	for i, e := range prev.ClusterNetwork {
+		prevIp, prevMask, _ := net.ParseCIDR(e.CIDR)
```
kyrtapz, line 418: lets not ignore the error returned from `net.ParseCIDR`

pkg/network/render.go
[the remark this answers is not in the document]
kyrtapz: answering my own question, when ServiceNetwork and the ClusterNetwork change we assume it is a dualstack conversion and run the following checks on ClusterNetwork:
```
	// Validate that the shared ClusterNetwork entries are unchanged, and that ALL of
	// the new ones in dualStack are of the opposite IP family from the shared ones.
	// (ie, you cannot go from [ipv4] to [ipv4, ipv6, ipv4], even though the latter
	// would have been valid as a new install.)
	EntryZeroIsIPv6 := utilnet.IsIPv6CIDRString(singleStack.ClusterNetwork[0].CIDR)
	for i := range dualStack.ClusterNetwork {
		if i < len(singleStack.ClusterNetwork) {
			if !reflect.DeepEqual(singleStack.ClusterNetwork[i], dualStack.ClusterNetwork[i]) {
				// Changed or re-ordered an existing ClusterNetwork element
				return errors.Errorf("cannot change ClusterNetwork")
			}
		} else if utilnet.IsIPv6CIDRString(dualStack.ClusterNetwork[i].CIDR) == EntryZeroIsIPv6 {
			// Added a new element of the existing IP family
			return errors.Errorf("cannot change ClusterNetwork")
		}
	}
```

kyrtapz:
@jluhrsen Does ovn-k properly clear all the setup it did for previous clusterNetworks on restart? I am concerned that we are going to have stale rules in OVN after CIDR expansion.

jluhrsen:
> @jluhrsen Does ovn-k properly clear all the setup it did for previous clusterNetworks on restart? I am concerned that we are going to have stale rules in OVN after CIDR expansion.

This is something that needs to be tested. Planning to do this manually. There are a few things I want to validate manually. [Here](https://docs.google.com/document/d/1v0C362-k3252XbrkKm50NgzqeMYsLkmLtm23DzbHunc/edit#bookmark=id.am7yp1gh21fm) are some low level checks @trozet helped me figure out to check. [here](https://docs.google.com/document/d/1v0C362-k3252XbrkKm50NgzqeMYsLkmLtm23DzbHunc/edit#bookmark=id.ajz46z6acvgr) is where I'm trying to document my manual testing efforts.

pkg/names/names.go
```diff
@@ -71,6 +71,10 @@ const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migrati
 // to indicate the current IP Family mode of the cluster: "single-stack" or "dual-stack"
 const NetworkIPFamilyModeAnnotation = "networkoperator.openshift.io/ip-family-mode"
 
+// ClusterNetworkCidr is an annotation on the OVN networks.operator.openshift.io daemonsets
```
kyrtapz, line 74:
```suggestion
// ClusterNetworkCIDRAnnotation is an annotation on the OVN networks.operator.openshift.io daemonsets
```

pkg/names/names.go
```diff
@@ -71,6 +71,10 @@ const NetworkMigrationAnnotation = "networkoperator.openshift.io/network-migrati
...
+// to indicate the current list of clusterNetwork CIDRs available to the cluster.
+const ClusterNetworkCIDRAnnotation = "networkoperator.openshift.io/cluster-network-cidr"
```
kyrtapz, line 76: shouldn't that be 
```suggestion
const ClusterNetworkCIDRsAnnotation = "networkoperator.openshift.io/cluster-network-cidrs"
```

pkg/network/render.go
```diff
@@ -385,6 +388,61 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	return nil
 }
 
+func isClusterNetworkChangeSafe(prev, next *operv1.NetworkSpec) error {
+
+	// quick check to make sure clusterNetwork slices are of same size as we do not
+	// support adding/removing additional clusterNetwork entries unless it's for a
+	// single/dual stack migration. in those cases validation is done in isNetworkChangeSafe()
+	if len(prev.ClusterNetwork) != len(next.ClusterNetwork) {
+		return errors.Errorf("Adding/Removing clusterNetwork entries of the same type is not supported")
```
kyrtapz, line 397: I think in general errors should not be capitalized: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/network/render.go
```diff
@@ -385,6 +388,61 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+	sort.SliceStable(next.ClusterNetwork, func(i, j int) bool {
+		return next.ClusterNetwork[i].CIDR < next.ClusterNetwork[j].CIDR
+	})
+
+	// since we do not allow the clusterNetwork[] size to change, it should be safe to compare
+	// prev[i] to next[i] in this validation
+	for i, e := range prev.ClusterNetwork {
+		prevIp, prevMask, err := net.ParseCIDR(e.CIDR)
+		if err != nil {
+			return errors.Errorf("Error parsing CIDR from ClusterNetwork entry %s", e.CIDR)
```
kyrtapz, line 419: it would be good to wrap the error returned from `net.ParseCIDR` into the returned one.

pkg/network/render.go
```diff
@@ -385,6 +388,61 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
+	// quick check to make sure clusterNetwork slices are of same size as we do not
+	// support adding/removing additional clusterNetwork entries unless it's for a
+	// single/dual stack migration. in those cases validation is done in isNetworkChangeSafe()
+	if len(prev.ClusterNetwork) != len(next.ClusterNetwork) {
+		return errors.Errorf("Adding/Removing clusterNetwork entries of the same type is not supported")
+	}
+
+	// Only support changing ClusterNetwork CIDR if it's OVNK
+	if !reflect.DeepEqual(next.DefaultNetwork.Type, operv1.NetworkTypeOVNKubernetes) {
+		return errors.Errorf("Adding/Removing clusterNetwork entries of the same type is not supported")
```
kyrtapz, line 402: Please fix the error message

pkg/network/render_test.go
```diff
@@ -196,29 +215,117 @@ func TestIsChangeSafe(t *testing.T) {
...
+	// original is 10.128.0.0/15, but shrinking the ip range with a /16 mask should not be allowed.
+	next.ClusterNetwork[0].CIDR = "10.128.0.0/16"
+	err := IsChangeSafe(prev, next, infra)
+	g.Expect(err).To(MatchError(ContainSubstring("Reducing IP range with a larger CIDR mask for clusterNetwork CIDR is unsupported")))
+}
+
+func TestDisallowShrinkingClusterNetworkCIDRMaskForSDN(t *testing.T) {
+	g, infra, prev, next := setupTestInfraAndBasicRenderConfigs(t, "sdn", "sdn")
+
+	// original is 10.128.0.0/15, but shrinking the ip range with a /16 mask should not be allowed.
```
kyrtapz, line 274:
```suggestion
	// Changes to the cluster network's CIDR mask is not allowed for OpenShiftSDN.
```

pkg/network/render_test.go
```diff
@@ -380,3 +483,33 @@ func Test_getMultusAdmissionControllerReplicas(t *testing.T) {
 		})
 	}
 }
+
+func fillDefaults(conf, previous *operv1.NetworkSpec) {
+	FillDefaults(conf, previous, 1400)
+}
+
+func setupTestInfraAndBasicRenderConfigs(t *testing.T, prevType, nextType string) (
```
kyrtapz, line 491: NIT: Can we simplify the function to take `operv1.Network` as prev and next?
Then we could just makes copies of the args and skip the if logic.
example of new call:
`setupTestInfraAndBasicRenderConfigs(t, OpenShiftSDNConfig, OpenShiftSDNConfig)`

kyrtapz (approved):
/lgtm
some small changes requested

jluhrsen:
> /lgtm some small changes requested

thanks @kyrtapz , everything should be addressed now

jluhrsen:
https://issues.redhat.com/browse/SDN-3674
/retest
/hold cancel

jluhrsen:
https://issues.redhat.com/browse/SDN-3671

PR #1711, opened 2023-02-10
always create env.sh when run_vs_existing_cluster

jluhrsen:
when working with run-locally.sh on new a new cluster, if that /tmp/env.sh file has not been removed, it will try to use it again and the script hangs waiting for something that will never show up. The 'oc' command to create it from scratch is not overly time consuming, so just run it every time.


trozet (approved):
/approve

PR #1717, opened 2023-02-20
Allow updates to pods

bn222:
https://issues.redhat.com/browse/NHE-443

https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/cni/cni_dpu.go#L70

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac.yaml
@@ -49,6 +49,7 @@ rules:
   - list
   - watch
   - patch
+  - update
 - apiGroups: [""]
   resources:
   - namespaces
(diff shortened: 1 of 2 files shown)


zshi-redhat:
/lgtm
cc @jcaamano

zshi-redhat:
In dpu-host mode, ovnkube-node on the host needs to update pod annotation to inform the ovnkube-node on the dpu about which interface to plug to OVS.

bn222:
> In dpu-host mode, ovnkube-node on the host needs to update pod annotation to inform the ovnkube-node on the dpu about which interface to plug to OVS.

This is correct. The important code for this can be found here: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/cni/cni_dpu.go#L70

PR #1720, opened 2023-02-27
Enables nodeSelector to be used in egress firewall rule

trozet:
Follows the upstream implementation here:
https://github.com/ovn-org/ovn-kubernetes/pull/3002

### DIFF
--- a/bindata/network/ovn-kubernetes/common/001-crd.yaml
+++ b/bindata/network/ovn-kubernetes/common/001-crd.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.0
+    controller-gen.kubebuilder.io/version: v0.11.3
   creationTimestamp: null
   name: egressfirewalls.k8s.ovn.org
 spec:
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -74,6 +74,53 @@ spec:
                           description: dnsName is the domain name to allow/deny traffic to. If this is set, cidrSelector must be unset.
                           pattern: ^([A-Za-z0-9-]+\.)*[A-Za-z0-9-]+\.?$
                           type: string
+                        nodeSelector:
+                          description: nodeSelector will allow/deny traffic to the
```
tssurya, line 78: I compared the changes with upstream ones that landed, the `cidrSelector` and `dnsName` descriptions are not updated in this one.

bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -74,6 +74,53 @@ spec:
...
+                            Kubernetes node IP of selected nodes. If this is, set
```
tssurya, line 79:
```suggestion
                            Kubernetes node IP of selected nodes. If this is set,
```
I could care less about a comma positioning usually but in CRD yaml descriptions which often customers/users read with the oc describe better to be accurate. (I know this is missing upstream too, but this will be the customer facing change I guess..)

bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -74,6 +74,53 @@ spec:
...
+                                type: string
+                              description: matchLabels is a map of {key,value} pairs.
+                                A single {key,value} in the matchLabels map is equivalent
+                                to an element of matchExpressions, whose key field
+                                is "key", the operator is "In", and the values array
+                                contains only "value". The requirements are ANDed.
+                              type: object
+                          type: object
                       type: object
                       minProperties: 1
```
tssurya, line 125: in the upstream PR the min and max properties were removed since before it was only cidrSelector versus dnsName and now we here we haven't removed them although I feel this is the correct behaviour, not what went in upstream, left a comment on that PR asking for more clarification:  https://github.com/ovn-org/ovn-kubernetes/pull/3002#discussion_r1121251873

tssurya (changes requested):

trozet:
updated, use hide whitespace when you review

PR #1724, opened 2023-03-06
OCPBUGS-8364: manifests: create namespace earlier

vrutkovs:
During initial install several manifests (i.e. metrics service) are being created on runlevel 50, but namespace is being created on runlevel 70.

This moves namespace to runlevel 50, so that during installation it would be created earlier


vrutkovs:
/hold

This is a very hacky way to speed up cluster install. It would be best to improve scc controller / installation proceedure

PR #1728, opened 2023-03-09
OCPBUGS-8692: HyperShift: Set affinity, tolerations and co-location for all hcp resources created by CNO

kyrtapz:
This duplicates the following funcionality of HyperShift control-plane operator:
https://github.com/openshift/hypershift/blob/main/support/config/deployment.go#L264-L265
https://hypershift-docs.netlify.app/how-to/distribute-hosted-cluster-workloads/

I will work on [nodeSelector](https://github.com/openshift/hypershift/blob/main/support/config/deployment.go#L255) implementation in CNO as a follow up PR.

/cc @jcaamano @zshi-redhat @enxebre @csrwng

### DIFF
--- a/bindata/cloud-network-config-controller/managed/controller.yaml
+++ b/bindata/cloud-network-config-controller/managed/controller.yaml
@@ -54,6 +54,7 @@ spec:
                   matchLabels:
                     hypershift.openshift.io/hosted-control-plane: {{.HostedClusterNamespace}}
                 topologyKey: kubernetes.io/hostname
+      priorityClassName: hypershift-control-plane
       securityContext:
         runAsNonRoot: true
         seccompProfile:
--- a/bindata/cloud-network-config-controller/managed/controller.yaml
+++ b/bindata/cloud-network-config-controller/managed/controller.yaml
@@ -9,6 +9,9 @@ metadata:
     release.openshift.io/version: "{{.ReleaseVersion}}"
     networkoperator.openshift.io/non-critical: ""
     network.operator.openshift.io/cluster-name:  {{.ManagementClusterName}}
+  labels:
+    # used by PodAffinity to prefer co-locating pods that belong to the same hosted cluster.
+    hypershift.openshift.io/hosted-control-plane: {{.HostedClusterNamespace}}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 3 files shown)


kyrtapz:
images are currently broken: https://redhat-internal.slack.com/archives/C02CZNQHGN8/p1678362561543709

bindata/cloud-network-config-controller/managed/controller.yaml
```diff
@@ -26,6 +29,32 @@ spec:
         type: infra
         openshift.io/component: network
     spec:
+      affinity:
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 50
```
jcaamano, line 35: I am curious about this weight. Can there be nodes labeled with `hypershift.openshift.io/cluster={{.HostedClusterNamespace}}`  but not labeled with `hypershift.openshift.io/control-plane`?

bindata/cloud-network-config-controller/managed/controller.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
     release.openshift.io/version: "{{.ReleaseVersion}}"
     networkoperator.openshift.io/non-critical: ""
     network.operator.openshift.io/cluster-name:  {{.ManagementClusterName}}
+  labels:
+    # used by PodAffinity to prefer co-locating pods that belong to the same hosted cluster.
+    hypershift.openshift.io/hosted-control-plane: {{.HostedClusterNamespace}}
```
jcaamano, line 14: Is this a standard hypershift label? Did not see it in https://hypershift-docs.netlify.app/how-to/distribute-hosted-cluster-workloads/.

jcaamano (approved):
Just a couple of questions

bindata/cloud-network-config-controller/managed/controller.yaml
kyrtapz: it is a label that the control-plane-operator uses: https://github.com/openshift/hypershift/blob/main/support/config/deployment.go#L27

bindata/cloud-network-config-controller/managed/controller.yaml
kyrtapz: IIURC `hypershift.openshift.io/control-plane` is a general label for all HCP pods that should have a lower priority than the more specific, per cluster `hypershift.openshift.io/cluster`.
In theory, I think it might be possible to only see `hypershift.openshift.io/cluster` but @enxebre might have more info on real world scenarios.

PR #1735, opened 2023-03-14
Fix tier label, privileged, HOSTNAME/NODENAME in whereabouts reconciler 

andreaskaris:
Fix tier label - it's not needed to run the whereabouts-reconciler downstream
Remove privileged security Context - elevated privileges are not needed to run the whereabouts-reconciler downstream
Retrieve node name from the downward API - this is needed to accomodate a recent bugfix in the whereabouts-reconciler
Grant multus clusterrole the right to get nodes - this is needed to accomodate the same bugfix in the whereabouts-reconciler

### DIFF
--- a/bindata/network/multus/002-rbac.yaml
+++ b/bindata/network/multus/002-rbac.yaml
@@ -41,6 +41,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups: [""]
+  resources:
+  - nodes
+  verbs:
+  - get
 - apiGroups: ["", "events.k8s.io"]
   resources:
   - events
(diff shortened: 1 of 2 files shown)


andreaskaris:
This would have to go in simultaneously with https://github.com/openshift/whereabouts-cni/pull/119
E.g.: launch openshift/cluster-network-operator#1735,openshift/whereabouts-cni#119 azure,ovn

andreaskaris:
@dougbtv I think that this one here is ready for review

dougbtv (approved):

dougbtv:
as you mentioned, I think you're right we just want the CNO PR to merge first :+1: thanks for identifying the dependency, and for the PR too of course

andreaskaris:
@dcbw Can you give me an override here, too? Thanks1

dcbw:
errors during gather-extra, not the test

PR #1736, opened 2023-03-14
OCPBUGS-10009: HyperShift: Support HostedControlPlane node selector

kyrtapz:
With https://github.com/openshift/hypershift/pull/1592 the hosted control plane can have a `nodeSelector` field that should be applied to all resources deployed in the hosted clusters control namespace.

Bumped the hypershift dep to v1betav1 since v1alpha1 is deprecated, thanks @martinkennelly


pkg/controller/statusmanager/status_manager.go
```diff
@@ -3,21 +3,22 @@ package statusmanager
...
 	"fmt"
-	"k8s.io/apimachinery/pkg/labels"
 	"log"
 	"os"
 	"reflect"
 	"strings"
 	"sync"
 
+	"github.com/openshift/cluster-network-operator/pkg/platform"
+	"k8s.io/apimachinery/pkg/labels"
```
jcaamano, line 13: nit: the imports in this file are already all over the place but the platform immport could go where the network import was and the apimachinery import could go in the apimachinery import group.

jcaamano (commented):
@kyrtapz Two questions:

* Should we expect the hypershift CI job to pass
* Is HostedControlPlane changes something that we are or should we be reconciling on?

kyrtapz:
>     * Should we expect the hypershift CI job to pass
CI is now passing :)
>     * Is HostedControlPlane changes something that we are or should we be reconciling on?
It is not required with this change, when `nodeSelector` changes CNO will be the first to restart with the new fields in the deployment, once it does it will pick up the changes and redeploy ovn-k master, cncc and multus adm controller.

jcaamano:
@kyrtapz did you see the imports comment? We can ignore if you would like as long as I get the ACK ;)

pkg/controller/statusmanager/status_manager.go
kyrtapz: good point, sorry for missing the comment sooner

kyrtapz:
Bumped the hypershift dep to `v1betav1` since `v1alpha1` is deprecated, thanks @martinkennelly

enxebre:
@kyrtapz @muraee we'd want this into 4.12 as well.

muraee:
@kyrtapz we are moving to drop `v1alpha1` in 4.12 as well, is there any plans to backport this PR to 4.12 or should I open a new dedicated PR to move to `v1beta1`?

kyrtapz:
> @kyrtapz we are moving to drop `v1alpha1` in 4.12 as well, is there any plans to backport this PR to 4.12 or should I open a new dedicated PR to move to `v1beta1`?

Yeah, here is the pr: https://github.com/openshift/cluster-network-operator/pull/1816
Sorry for missing the previous comment.

PR #1739, opened 2023-03-15
OVN-K alerts: add runbook for SB DB disconnection

martinkennelly:


martinkennelly:
/close
Patryk already has PR up

PR #1744, opened 2023-03-21
OCPBUGS-10031: operConfig reconcile can return nil error on failure

jcaamano:
When an object reconciliation fails ~other than the last~.


jcaamano:
/cc @nicklesimba @dougbtv 

PTAL

kyrtapz:
/lgtm
That was not even the last apply error right? since apply error was scoped to the if block:
```
if err := apply.ApplyObject(ctx, r.client, obj, ControllerName); err != nil {
```

PR #1745, opened 2023-03-21
OCPBUGS-10647: multus-admission-controller should not run as root under Hypershift

michael-topchiev:
This PR changes pod security context of `multus-admission-controller` to run with a specific user ID when under Hypershift management.

Instead of duplicating code from Hypershift CPO to detect management cluster capabilities https://github.com/openshift/hypershift/blob/9d04882e2e6896d5f9e04551331ecd2129355ecd/support/capabilities/management_cluster_capabilities.go#L102-L109 (which possibly requires additional permissions assigned to CNO service account), user ID to run with is passed from Hypershift as an env variable.

### DIFF
--- a/bindata/network/multus-admission-controller/admission-controller.yaml
+++ b/bindata/network/multus-admission-controller/admission-controller.yaml
@@ -212,6 +212,10 @@ spec:
       - name: admin-kubeconfig
         secret:
           secretName: service-network-admin-kubeconfig
+{{- if not (eq .RunAsUser "")}}
+      securityContext:
+        runAsUser: {{.RunAsUser}}
+{{- end }}
       tolerations:
         - key: "hypershift.openshift.io/control-plane"
           operator: "Equal"
--- a/pkg/network/hypershift.go
+++ b/pkg/network/hypershift.go
@@ -31,6 +31,7 @@ var (
 	routeHost      = os.Getenv("OVN_SBDB_ROUTE_HOST")
 	routeLabels    = map[string]string{}
 	routeLabelsRaw = os.Getenv("OVN_SBDB_ROUTE_LABELS")
+	runAsUser      = os.Getenv("RUN_AS_USER")
 )
 
 const (
@@ -53,6 +54,7 @@ type HyperShiftConfig struct {
 	Name               string
 	Namespace          string
 	OVNSbDbRouteHost   string
+	RunAsUser          string
 	OVNSbDbRouteLabels map[string]string
 	RelatedObjects     []RelatedObject
 }
@@ -62,6 +64,7 @@ func NewHyperShiftConfig() *HyperShiftConfig {
 		Enabled:            hyperShiftEnabled(),
 		Name:               name,
 		Namespace:          namespace,
+		RunAsUser:          runAsUser,
 		OVNSbDbRouteHost:   routeHost,
 		OVNSbDbRouteLabels: routeLabels,
 	}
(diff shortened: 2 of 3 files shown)


dougbtv:
I'm good with the "user ID to run with is passed from Hypershift as an env variable" approach.

michael-topchiev:
Corresponding Hypershift PR submitted https://github.com/openshift/hypershift/pull/2319

danwinship:
/hold
@dougbtv can't we just make it _always_ run as non-root?

zshi-redhat:
> > /hold @dougbtv can't we just make it _always_ run as non-root?
> 
> @dougbtv We test and use CNO with Hypershift, so the scope was specifically narrowed to Hypershift-managed case to avoid potential impact on other use-cases.
> 
> With this PR the variable `RUN_AS_USER` works only if CNO is Hypershift-managed. If it makes sense, this PR can be changed to make parameter `RUN_AS_USER` to always work, regardless whether under Hypershift or not, expanding its scope. Or you want to always run `multus-admission-controller` with a hard-coded runAsUser ID?

@michael-topchiev Looks like the multus-admission-controller already run with hardcoded `runAsUser`: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus-admission-controller/admission-controller.yaml#L188,  probably we don't need this change?

michael-topchiev:
I see that without Hypershift `multus-admission-controller` runs as user 65534. The problem this PR addresses is that `multus-admission-controller` runs with `root` when CNO is Hypershift-managed, so some change IS needed.

michael-topchiev:
> > I see that without Hypershift `multus-admission-controller` runs as user 65534. The problem this PR addresses is that `multus-admission-controller` runs with `root` when CNO is Hypershift-managed and runs on Kubernetes management cluster (not Openshift), so some change IS needed.
> 
> I see, I didn't realize the runAsUser is in the `if` clause for the normal deployment. Maybe we just always run it as non-root user as @danwinship suggested.

Normally Hypershift uses the Openshift default security context for control plane services when it runs on Openshift, and uses runAsUser `1001` (https://github.com/openshift/hypershift/blob/a0377ae793b20fb565ac8852ee4db1e5bb4780da/support/config/constants.go#L22) when running on Kubernetes. I am not sure about impact of hard-coded rinAsUser uid `65534` on Hypershift-managed control plane. The advantage though is that no Hypershift code changes will be needed to pass `RUN_AS_USER` env variable.

Maybe @enxebre or @csrwng can tell if Hypershift will be fine with `multus-admission-controller` alway running with hard-coded runAsUser `65534` both on Openshift and Kubernetes.

michael-topchiev:
@danwinship @dougbtv this PR fixes a high-priority Hypershift security issue and is in line with the Hypershift's pod security context assignment rules. I believe this code fix should be good as is. The corresponding Hypershift PR is approved and merged (https://github.com/openshift/hypershift/pull/2319). Would it be possible to review this PR expeditiously?

zshi-redhat:
@nicklesimba  I noticed you used to change the runAsUser for regular openshift deployment, could you double check if this change makes sense for multus-admission-controller?

zshi-redhat:
@jcaamano PTAL.

zshi-redhat:
@michael-topchiev  could you take a look at why e2e-hypershift-ovn job is failing?

michael-topchiev:
> @michael-topchiev could you take a look at why e2e-hypershift-ovn job is failing?

Not sure, asked Hypershift team https://redhat-internal.slack.com/archives/C01C8502FMM/p1680803943113969

dougbtv:
If you can make the modification to always have the runAsUser set, I definitely give that a +1, agreed, this never needs root privileges.

PR #1752, opened 2023-03-28
OCPBUGS-10937: multus-admission-controller mounts secret with mode 0640

zshi-redhat:
CNO managed component (multus-admission-controller) to conform to hypershift control plane expectations that All secrets should be mounted to not have global read. change from 420(0644) to 416(0640)

### DIFF
--- a/bindata/network/multus-admission-controller/admission-controller.yaml
+++ b/bindata/network/multus-admission-controller/admission-controller.yaml
@@ -213,18 +213,23 @@ spec:
       volumes:
       - name: webhook-certs
         secret:
+{{- if .HyperShiftEnabled}}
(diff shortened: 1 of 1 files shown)


zshi-redhat:
cc @nicklesimba @dougbtv

dcbw:
@zshi-redhat seems like the many of the jobs fail networking bring up now, can you investigate?

zshi-redhat:
This fix is specific to hypershift and the e2e-hypershift-ovn job passed.

zshi-redhat:
@dcbw could you take another look?  Previous failure was due to the mode 640 cannot be applied to self-managed deployment, I made it specific to hypershift (managed).

PR #1754, opened 2023-03-28
Bug OCPBUGS-8070: Merge legacy field manager to current field manager

martinkennelly:
Previous to 4.11, we created Update operations with a manager called 'cluster-network-operator', however following 4.11, we changed the field manager name

field because the legacy field manager owned and
preserved the field.

Need API server teams review. There is no clear documentation upstream on how to do this, so I have experimented.

I need to write test cases but I need an API server with kube-controller (?), because the fake API server will not merge the managed fields.

/hold
/cc


martinkennelly:
Requested API server team review for this PR in forum-apiserver.

PR #1756, opened 2023-03-30
OCPBUGS-11136: Add liveness and readiness probes

michael-topchiev:
This PR adds simple Liveness and Readiness probes so they can be used from any client application, like Hypershift.


michael-topchiev:
Dropping the PR because there is no hard evidence that the service really need the probes.

PR #1770, opened 2023-04-05
OCPBUGS-11448: add Hypershift release-image annotation to multus

michael-topchiev:
This PR fixes the problem described in [OCPBUGS-11448](https://issues.redhat.com/browse/OCPBUGS-11448) bug.

The proposed solution is for CNO, when managed by Hypershift, to read annotations from its own deployment manifest and propagate "hypershift.openshift.io/release-image" annotation down to multus-admission-controller deployment.

Currently Hypershift does not assign CNO necessary permission to be able to `get` its own deployment manifest. To address the need, issue [OCPBUGS-11450](https://issues.redhat.com/browse/OCPBUGS-11450) is created.

Corresponding [Hypershift PR](https://github.com/openshift/hypershift/pull/2384) is submitted

Update:
After discussion, the solution changed to read env variable OPENSHIFT_RELEASE_IMAGE that is passed from Hypershift to CNO.


kyrtapz:
CNO sets `release.openshift.io/version` on all of the resources it deploys: https://github.com/openshift/cluster-network-operator/blob/b3539db3b76066b51dd8162212441c6eeacc3aba/bindata/cloud-network-config-controller/managed/controller.yaml#L9 The value for that is passed to it as an environment variable form HyperShift:  https://github.com/openshift/hypershift/blob/161d250fb2bcb3785583763b306b5127b1bf5a7d/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L448

If the `hypershift.openshift.io/release-image` is really needed here I think it might be better to also pass it as an env to CNO.
Additionally, shouldn't this also apply to all resources deployed by CNO? (ovnkube-master, cloud-network-config-controller, multus-addmission-controller)

michael-topchiev:
Yes, we can pass an additional env variable `OPENSHIFT_RELEASE_IMAGE` from Hypershift to CNO instead. The disadvantage is that the number of such variables continues to grow and every time new variable is introduced both Hypershift and CNO code have to change in lockstep. On the other hand, retrieving the annotation value from CNO deployment in this case, and in possible future cases like this, do not require Hypershift changes.

If all agree that passing the annotation value as an Env variable is preferred, I am ok with that and will make changes to this PR and corresponding changes to Hypershift code as well.

kyrtapz:
> Yes, we can pass an additional env variable `OPENSHIFT_RELEASE_IMAGE` from Hypershift to CNO instead. The disadvantage is that the number of such variables continues to grow and every time new variable is introduced both Hypershift and CNO code have to change in lockstep. On the other hand, retrieving the annotation value from CNO deployment in this case, and in possible future cases like this, do not require Hypershift changes.

Agreed, it is not very sustainable for us to keep adding options via env vars. Considering that CNO is [already retrieving the HostedControlPlane](https://github.com/openshift/cluster-network-operator/blob/04b74dbd6eb504dc0004f96f87acd57210c567db/pkg/network/multus_admission_controller.go#L100) CR why don't we just use that to get the `OPENSHIFT_RELEASE_IMAGE` (`hcp.spec.releaseImage`)?


In https://github.com/openshift/cluster-network-operator/pull/1736/commits/4c80be9b4d9ec7a7b98c51fa103dca9b971407c8 I am moving getting the HostedControlPlane so it is accessible from InfraStatus.

kyrtapz:
> I do not think that will work because the purpose of the annotation is for managed service providers to track CP upgrades and the annotation value should change only after CNO is upgraded, i.e when `hcp.spec.releaseImage` changes `multus-admission-controller' deployment must not change until CNO upgrades.

Oh right, that is a valid point. I still don't think CNO should retrieve this value from it's deployment but it should be configured as an env, this is very similar to how the `ReleaseVersion` is used. I want to avoid CNO retrieving yet another resource from the API to get something that is not going to change during its runtime.

cc: @jcaamano wdyt?

michael-topchiev:
As discussed above, the code changed to read env variable `OPENSHIFT_RELEASE_IMAGE` that is passed from Hypershift to CNO.

kyrtapz:
@michael-topchiev does this also need to be applied to [ovn-kube-master statefulset](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml) and [cloud-network-config-controller deployment](https://github.com/openshift/cluster-network-operator/blob/master/bindata/cloud-network-config-controller/managed/controller.yaml)?

Other than that the change looks good to me.

michael-topchiev:
> @michael-topchiev does this also need to be applied to [ovn-kube-master statefulset](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml) and [cloud-network-config-controller deployment](https://github.com/openshift/cluster-network-operator/blob/master/bindata/cloud-network-config-controller/managed/controller.yaml)?
> 
> Other than that the change looks good to me.

Not sure if those two services are used by any managed service providers. 

@enxebre @csrwng please let us know if `ovn-kube-master statefulset` and `cloud-network-config-controller deployment` also need to have "hypershift.openshift.io/release-image" annotation.

csrwng:
> please let us know if ovn-kube-master statefulset and cloud-network-config-controller deployment also need to have "hypershift.openshift.io/release-image" annotation.

The only reason would be consistency. However, I do not know of any current use case where that annotation is used outside of IBM's use case.

michael-topchiev:
@kyrtapz the annotation is added to `ovn-kube-master` statefulset and `cloud-network-config-controller` deployment for consistency.

kyrtapz:
/lgtm
NIT: @michael-topchiev could you rebase on master instead of merging it in in the second commit?
@csrwng  It looks like the hypershift job is not happy, are there any known issues?

michael-topchiev:
Rebased

michael-topchiev:
Re-based the PR branch.

kyrtapz:
/lgtm
@dougbtv ptal for approve

PR #1773, opened 2023-04-06
OCPBUGS-11046: Fix allowlist ds template

mlguerrero12:
This issue does not allow the allowlist controller to update the sysctl allowlist file of the nodes.


mlguerrero12:
@dougbtv, could you please have a look at this?

PR #1786, opened 2023-04-19
OCPBUGS-5027: Make the operator degraded on panic

kyrtapz:
This PR introduces two changes to:
- Make the Operator/ClusterOperator degraded on panic if any of the controllers crashes
- Do not set the Operator as available if the network config status is not set
   I am not sure if this could be done better but I am open to suggestions, CNO sets the `Available` condition once it is done deploying all of the pods, I added an additional condition that it also has to wait for the config to get updated.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -14,14 +14,17 @@ import (
 	"github.com/openshift/library-go/pkg/operator/managementstatecontroller"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
-
 	"github.com/openshift/library-go/pkg/operator/management"
+
+	"sigs.k8s.io/controller-runtime/pkg/config/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	ctmanager "sigs.k8s.io/controller-runtime/pkg/manager"
 )
@@ -51,6 +54,9 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
 			return o.client.Default().RESTMapper(), nil
 		},
 		MetricsBindAddress: "0",
+		Controller: v1alpha1.ControllerConfigurationSpec{
+			RecoverPanic: pointer.Bool(true),
+		},
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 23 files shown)


pkg/operator/operator.go
```diff
@@ -44,13 +50,50 @@ func RunOperator(ctx context.Context, controllerConfig *controllercmd.Controller
...
+				return err
+			}
+			return nil
+		})
+		if err != nil {
+			klog.Errorf("Failed to update cluster operator degraded condition on panic %v", err)
+			return
+		}
+
+		panic(panicValue)
```
dulek, line 85: We still want to panic instead of suppressing it?

jcaamano: I think so. Consistency of the process itself cannot be ensured after a panic. The panic handling should do relatively simple things that if gone wrong don't risk the system so it's even bold that we attempt to update the status to degraded.
@kyrtapz can you point to a reference of other operators doing this for my peace of mind?

kyrtapz: I think we want to forward the panic as we are not really recovering from it here, we are just setting the status.
There is no assurance about the state of CNO after the crash occurs so  we should not allow it to continue, potentially causing more issues.

library-go controller wrapper(which CNO doesn't use) handles panics using `runtime.HandleCrash`: https://github.com/openshift/library-go/blob/a467dafcef6988afbcce43400c339d7ca68a3298/pkg/controller/factory/base_controller.go#L80
The panic will be forwarded if `runtime.ReallyCrash` is true: https://github.com/openshift/library-go/blob/db40400cfd41917bd76b1177023e3e5c69e9aa62/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go#L54
It's value can be [controlled](https://github.com/openshift/library-go/blob/9575de9b95b2542f69a00ff81ac4518b50cfdafe/pkg/serviceability/panic.go#L35) by an env variable: https://github.com/openshift/library-go/blob/4a2b231285b3a45bdbd3d86e2374965c626c3fae/pkg/controller/controllercmd/cmd.go#L115 (`false` by default)

CNO uses upstream controller runtime right now and none of its controllers implement panic recovery.
An alternative to the proposed approach would be to add panic handling to every controller manually.

jcaamano: Wondering if we would lose info if we don't go through status manager `setDegraded`?
Also, there we do create or update and I am not sure if we need that here.

What would be the reasons not to go through status manager? Aside from redudinc the surface of further panics?

dulek: Alright, thanks for extensive explanations! It makes sense to me.

kyrtapz: > What would be the reasons not to go through status manager? Aside from reducing the surface of further panics?

That was my original intent but after analyzing it further and talking to Jaime I think it is better to rely on status manager. In the previous approach there was a chance that while handling the panic (running deferred funcs) some other controller would clean the degraded status, using statusmanager with a new statusLevel addresses that issue.

dulek:
Still looks good to me in terms of bug being fixed.

PR #1788, opened 2023-04-21
OCPBUGS-11565: High API requests due to allowlist and operconfig reconcilers running too often

donpenney:
The allowlist and operconfig reconcilers are running too often, which significantly drives up the cluster-network-operator kube-apiserver requests.

The allowlist controller creates a watcher that triggers the reconciler for any configmap change in any namespace. The first thing the reconciler does is a GET request for the 1 configmap it manages to check existence, creating it if it doesn't already exist. The second thing is to exit if the object being reconciled is not the configmap it manages. This results in the reconciler being run almost constantly, up to a few thousand times per hour, with a configmap GET request every time. By changing the controller to use a cmInformer to limit the watcher to a specific namespace, as is done with other controllers watching configmaps, the reconciler is run only when needed.

The operconfig reconciler accesses and manages a larger number of resources, and requeues itself every 3 minutes in order to update node status if needed. The requeue functionality is reliant on the uniqueness of the reconcile request object, ie. namespace and name. Originally, the controller launched this recurring reconciler with just a watcher on the network type. A later update added configmap and node watchers as additional triggers, with a request object transformer to set the request name to match the default network name. However, this transformer also set the namespace, where the network reconcile request was unnamespaced. This created a second unique recurring reconciler, as can be seen in the logs by pairing the timestamps of "Operconfig Controller complete" logs with the subsequent requeued reconciler start log "Reconciling Network.operator.openshift.io cluster" three minutes later. As a result, the reconciler is now run twice every three minutes, rather than once. By updating the request object transformer to leave the namespace unset, the reconcile requests triggered by configmap and node changes now match the network request object, and the requeue sees these as the same trigger, leaving a single recurring reconciler requeued.

### DIFF
--- a/pkg/controller/allowlist/allowlist_controller.go
+++ b/pkg/controller/allowlist/allowlist_controller.go
@@ -2,7 +2,9 @@ package allowlist
 
 import (
 	"context"
+	"log"
 	"os"
+	"strings"
 	"time"
 
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
@@ -17,11 +19,15 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	v1coreinformers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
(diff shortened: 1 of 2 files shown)


donpenney:
The e2e-openstack-ovn test is failing, but the failures are not consistent. Is this a flaky test, or could this somehow be related to the operconfig_controller change? Given that the reconciler handles the unnamespaced network object and deals with objects in various namespaces, it doesn't seem like this change should impact anything other than recurrence.

If this isn't a flaky test, however, and this change is somehow triggering the failures, I can either split out the change from this PR, or alter the change to undo the object transformation change and instead do a namespace check when returning ReconcileAfter at the end of the Reconcile function (only triggering the recurring reconciler for the namespaced reconcile request, and not the unnamespaced network object. This is how I originally coded the change, and have already verified it resolves the recurrence issue, without changing how the reconcile is called, but the current implementation of updating the transformation seemed like the cleaner approach.

@flavio-fernandes @browsell

flavio-fernandes:
/assign @pperiyasamy 
Hi Peri. Can you please track this as part of https://issues.redhat.com/browse/OCPBUGS-11565 ?

dougbtv:
I'm quite willing to approve, but I want to get a review from @mlguerrero12 first please, and thanks!

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -37,19 +45,38 @@ func Add(mgr manager.Manager, status *statusmanager.StatusManager, c cnoclient.C
...
+	// watch for changes in all configmaps in our namespace
+	cmInformer := v1coreinformers.NewConfigMapInformer(
+		r.client.Default().Kubernetes(),
+		names.MULTUS_NAMESPACE,
+		0, // don't resync
+		cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+
+	r.client.Default().AddCustomInformer(cmInformer) // Tell the ClusterClient about this informer
+
+	if err := c.Watch(&source.Informer{Informer: cmInformer},
```
mlguerrero12, line 67: you can simply return here

donpenney: Ok. I need to rebase the PR, so I'll update the code and retest

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -76,6 +97,7 @@ func (r *ReconcileAllowlist) Reconcile(ctx context.Context, request reconcile.Re
 	if request.Namespace != names.MULTUS_NAMESPACE || request.Name != names.ALLOWLIST_CONFIG_NAME {
```
mlguerrero12, line 97: I suppose we can delete the namespace check

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -76,6 +97,7 @@ func (r *ReconcileAllowlist) Reconcile(ctx context.Context, request reconcile.Re
...
 		return reconcile.Result{}, nil
 	}
+	log.Printf("Reconcile allowlist for %s/%s", request.Namespace, request.Name)
```
mlguerrero12, line 100: please use klog to keep consistency on this file

pkg/controller/allowlist/allowlist_controller.go
donpenney: Ok, will do

pkg/controller/allowlist/allowlist_controller.go
donpenney: Ok, will do. I had been following code of the other controllers and didn't notice this controller was using klog.

pkg/controller/allowlist/allowlist_controller.go
donpenney: Done

pkg/controller/allowlist/allowlist_controller.go
donpenney: Done

pkg/controller/allowlist/allowlist_controller.go
donpenney: Done

donpenney:
@dougbtv Job history shows this e2e-hypershift-ovn has been failing repeatedly for almost a week, with the last successful run April 27. Seems to be AWS issues?

PR #1795, opened 2023-04-27
OCPBUGS-15220: Add zone node preference to multus-admission-controller

brucejcong:


bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -68,6 +71,13 @@ spec:
                   matchLabels:
                     hypershift.openshift.io/hosted-control-plane: {{.AdmissionControllerNamespace}}
                 topologyKey: kubernetes.io/hostname
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchLabels:
+                app: multus-admission-controller
+                hypershift.openshift.io/hosted-control-plane: {{.AdmissionControllerNamespace}}
```
pliurh, line 79: Do we need this line? With line 45, I think `app: multus-admission-controller` is sufficient here.

pliurh:
It looks good to me. @s1061123 Could you take a look?

PR #1800, opened 2023-05-02
[WIP] Use encapsulation=true for IBM Cloud

mccv1r0:
Use encapsulation=true for IBM Cloud

e.g. ovn-nbctl set nb_global . options:ipsec_encapsulation=true


bindata/network/ovn-kubernetes/self-hosted/ovnkube-master.yaml
```diff
@@ -381,6 +381,15 @@ spec:
                 {{ else }}
                 ipsec=false
                 {{ end }}
+                if [ "${ipsec}" == "true" ]; then
```
dcbw, line 384: @mccv1r0 you could combine this with the `set nb_global` below and do it all in one?

mccv1r0: It has to be 

```sudo ovn-nbctl set nb_global . ipsec=${ipsec} options:ipsec_encapsulation=${encapsulation}```

where ${encapsulation} defaults to false and is set true only for IBMCloud

dcbw:
/retest
registry issues; random flakes

libesz:
Hey @mccv1r0 . As I can see this changes the self-hosted cluster behavior, like IPI. I wonder if hypershift based installs could be fixed as well, which use manifests from the [managed](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml) directory.

mccv1r0:
>  I wonder if hypershift based installs could be fixed as well

Given all we (the industry) learned over the years re IPsec/IKE, IMHO we should use encapsulation all the time.

In an effort to have this ready before an escalation, I have a [PR ](https://github.com/openshift/api/pull/1472) that is laying the groundwork for letting encapsulation be used anytime one wants to turn it on, regardless of platform.

libesz:
Thanks @mccv1r0, great news! May I ask if you plan to have this API working in Openshift 4.14?

rbbratta:
ping

PR #1803, opened 2023-05-08
OCPBUGS-13219: Use `IfNotPresent` instead of `Always` in OVNK upgrades pre-puller

jhernand:
Currently the `ovnkube-upgrades-prepuller` daemon set uses the `Always` image pull policy. As a result this pod requires the presence of working image registry server. That is not really needed, and complicates upgrades in environments where the registry is not available at the time this pod runs. To avoid that this patch chages the image pull policy to `IfNotPresent`.

Related: https://issues.redhat.com/browse/OCPBUGS-13219
Related: https://issues.redhat.com/browse/MGMT-13733


oourfali:
lgtm

pkg/network/ovn_kubernetes_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
```
martinkennelly, line 19: dupe import

martinkennelly: Besides that - looks good.

trozet (approved):
/lgtm

PR #1807, opened 2023-05-08
SDN-3708: Support configurable masquerade subnet in ovn-k 

bpickard22:
Adds the necessary knobs in the cno to consume the api change (https://github.com/openshift/api/pull/1410) exposing the internal masquerade subnet in ovn-k where we reserve addresses for ip masquerading

### DIFF
--- a/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
@@ -66,6 +66,13 @@ data:
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
 
+{{- if (index . "V4InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
+{{- end }}
+{{- if (index . "V6InternalMasqueradeSubnet")}}
+    v6-internal-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
+{{- end }}
+
     [logging]
     libovsdblogfile=/var/log/ovnkube/libovsdb.log
     logfile-maxsize=100
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -119,8 +119,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["TokenAudience"] = os.Getenv("TOKEN_AUDIENCE")
 	data.Data["MTU"] = c.MTU
 	data.Data["RoutableMTU"] = nil
+	// v4 and v6 join subnet are used when the user wants to use the addresses that we reserve for the join subnet in ovn-k
+	// TODO: this field is being deprecated and will turn into c.GatewayConfig.IPv4/6.InternalJoinSubnet when we introduce the transit switch config into the api
 	data.Data["V4JoinSubnet"] = c.V4InternalSubnet
 	data.Data["V6JoinSubnet"] = c.V6InternalSubnet
+	// v4 and v6InternalMasqueradeSubnet are used when the user wants to use the addresses that we reserve in ovn-k for ip masquerading
+	if c.GatewayConfig != nil && c.GatewayConfig.IPv4.InternalMasqueradeSubnet != "" {
+		data.Data["V4InternalMasqueradeSubnet"] = c.GatewayConfig.IPv4.InternalMasqueradeSubnet
+	}
+	if c.GatewayConfig != nil && c.GatewayConfig.IPv6.InternalMasqueradeSubnet != "" {
+		data.Data["V6InternalMasqueradeSubnet"] = c.GatewayConfig.IPv6.InternalMasqueradeSubnet
+	}
 	data.Data["EnableUDPAggregation"] = !bootstrapResult.OVN.OVNKubernetesConfig.DisableUDPAggregation
 	data.Data["NETWORK_NODE_IDENTITY_ENABLE"] = bootstrapResult.Infra.NetworkNodeIdentityEnabled
 	data.Data["NodeIdentityCertDuration"] = OVN_NODE_IDENTITY_CERT_DURATION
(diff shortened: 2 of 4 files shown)


pkg/network/ovn_kubernetes_test.go
```diff
@@ -179,6 +179,7 @@ func TestRenderedOVNKubernetesConfig(t *testing.T) {
 		egressIPConfig         *operv1.EgressIPConfig
 		masterIPs              []string
 		v4InternalSubnet       string
+		v4InternalMasqueradeSubnet string
```
flavio-fernandes, line 182: NIT align all lines above and below this to accommodate the longer string

ricky-rav:
@bpickard22 let's make sure CNO builds with the changes in ovn_kubernetes.go :) 
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1807/pull-ci-openshift-cluster-network-operator-master-verify/1687179374593314816

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 				}
 			}
 		}
+		if oc.V4InternalMasqueradeSubnet != "" {
+			if !cnHasIPv4 {
+				out = append(out, errors.Errorf("v4InternalMasqueradeSubnet and ClusterNetwork must have matching IP families"))
```
ricky-rav, line 905: let's include `oc.V4InternalMasqueradeSubnet` in the error message

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+			}
+			_, v4Net, err = net.ParseCIDR(oc.V4InternalMasqueradeSubnet)
+			if err != nil {
+				out = append(out, errors.Errorf("v4InternalMasqueradeSubnet is invalid: %s", err))
+			}
+		}
+		if oc.V6InternalMasqueradeSubnet != "" {
+			if !cnHasIPv4 {
```
ricky-rav, line 913: so here it's `!cnHasIPv6`

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+				out = append(out, errors.Errorf("v6InternalMasqueradeSubnet and ClusterNetwork must have matching IP families"))
```
ricky-rav, line 914: let's print `oc.V6InternalMasqueradeSubnet` in the error message.

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+			}
+			_, v4Net, err = net.ParseCIDR(oc.V6InternalMasqueradeSubnet)
```
ricky-rav, line 916: `v6net`

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+			if err != nil {
+				out = append(out, errors.Errorf("v6InternalMasqueradeSubnet is invalid: %s", err))
+			}
+		}
+		for _, cn := range conf.ClusterNetwork {
+			if utilnet.IsIPv6CIDRString(cn.CIDR) {
+				if oc.V6InternalMasqueradeSubnet != "" {
+					_, v6ClusterNet, _ := net.ParseCIDR(cn.CIDR)
+					if iputil.NetsOverlap(*v6Net, *v6ClusterNet) {
+						out = append(out, errors.Errorf("v6InternalMasqueradeSubnet overlaps with ClusterNetwork %s", cn.CIDR))
```
ricky-rav, line 926: let's include `v6InternalMasqueradeSubnet` in the error message

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+					}
+				}
+			} else {
+				if oc.V4InternalMasqueradeSubnet != "" {
+					_, v4ClusterNet, _ := net.ParseCIDR(cn.CIDR)
+					if iputil.NetsOverlap(*v4Net, *v4ClusterNet) {
+						out = append(out, errors.Errorf("v4InternalMasqueradeSubnet overlaps with ClusterNetwork %s", cn.CIDR))
```
ricky-rav, line 933: let's include `v4InternalMasqueradeSubnet` in the error message

pkg/network/ovn_kubernetes.go
```diff
@@ -897,6 +900,58 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+					}
+				}
+			}
+		}
+		for _, sn := range conf.ServiceNetwork {
+			if utilnet.IsIPv6CIDRString(sn) {
+				if oc.V6InternalMasqueradeSubnet != "" {
+					_, v6ServiceNet, _ := net.ParseCIDR(sn)
+					if iputil.NetsOverlap(*v6Net, *v6ServiceNet) {
+						out = append(out, errors.Errorf("v6InternalMasqueradeSubnet overlaps with ServiceNetwork %s", sn))
```
ricky-rav, line 943: same here for `oc.v6InternalMasqueradeSubnet`

pkg/network/ovn_kubernetes_test.go
ricky-rav: The IDE should do it automagically :)

ricky-rav (changes requested):
I gave the PR a first pass. We need the [OVNK PR ](https://github.com/ovn-org/ovn-kubernetes/pull/3594#discussion_r1281023454) to merge first, the [API PR ](https://github.com/openshift/api/pull/1410)second, and then once we update `vendor/github.com/openshift/api/` with the new API, we can finally test this CNO PR.

pkg/network/ovn_kubernetes.go
```diff
@@ -901,14 +904,14 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 		var err error
 		if oc.V4InternalSubnet != "" {
 			if !cnHasIPv4 {
-				out = append(out, errors.Errorf("v4InternalSubnet and ClusterNetwork must have matching IP families"))
+				out = append(out, errors.Errorf("v4InternalSubnet %v and ClusterNetwork must have matching IP families"))
```
ricky-rav, line 907: here there's nothing to format for %v :)

pkg/network/ovn_kubernetes.go
```diff
@@ -901,14 +904,14 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
 			}
 			_, v4Net, err = net.ParseCIDR(oc.V4InternalSubnet)
 			if err != nil {
 				out = append(out, errors.Errorf("v4InternalSubnet is invalid: %s", err))
 			}
 			if !isV4InternalSubnetLargeEnough(conf) {
-				out = append(out, errors.Errorf("v4InternalSubnet is no large enough for the maximum number of nodes which can be supported by ClusterNetwork"))
+				out = append(out, errors.Errorf("v4InternalSubnet is not large enough for the maximum number of nodes which can be supported by ClusterNetwork"))
```
ricky-rav, line 914: let's print what v4InternalSubnet is

pkg/network/ovn_kubernetes.go
```diff
@@ -920,7 +923,7 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
 				out = append(out, errors.Errorf("v6InternalSubnet is invalid: %s", err))
 			}
 			if !isV6InternalSubnetLargeEnough(conf) {
-				out = append(out, errors.Errorf("v6InternalSubnet is no large enough for the maximum number of nodes which can be supported by ClusterNetwork"))
+				out = append(out, errors.Errorf("v6InternalSubnet is not large enough for the maximum number of nodes which can be supported by ClusterNetwork"))
```
ricky-rav, line 926: let's print what v6InternalSubnet is

ricky-rav:
@bpickard22 so in order to test CNO against the API changes that are still to be merged, you can modify the `go.mod` file to point to your own fork of openshift/api with the commit hash from your API PR. I tried it here: https://github.com/ricky-rav/cluster-network-operator/commit/94ac16e1c1d624c25da0731e0e1abbee4c5fc797

Basically in the "replace" part of the go.mod file, you type:
`github.com/openshift/api => github.com/bpickard22/api 680fd20165a4385e0b92a50ecd91d6c3763bd5d1
`

Then you run `go mod vendor` and it will rewrite the commit ID into the "v0.0.0..." form:
`github.com/openshift/api => github.com/bpickard22/api v0.0.0-20230804184907-680fd20165a4
`

Of course you'll undo this once the API PR merges, but at least we can move forward with this CNO in the meantime :)

pkg/network/ovn_kubernetes.go
```diff
@@ -141,6 +141,9 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["RoutableMTU"] = nil
 	data.Data["V4JoinSubnet"] = c.V4InternalSubnet
 	data.Data["V6JoinSubnet"] = c.V6InternalSubnet
+	data.Data["V4InternalMasqueradeSubnet"] = c.V4InternalMasqueradeSubnet
```
ricky-rav, line 144: this and the following ones will be `c.GatewayConfig.V4InternalMasqueradeSubnet`

pkg/network/ovn_kubernetes.go
bpickard22: this is the join subnet which is not part of this feature, but we can go ahead and update that log

pkg/network/ovn_kubernetes.go
bpickard22: why would these be `c.GatewayConfig.V4InternalMasqueradeSubnet` when the join subnet above in the same config is not?

zshi-redhat:
@bpickard22 Is this PR ready to merge?

bpickard22:
> @bpickard22 Is this PR ready to merge?

@zshi-redhat ci is failing due to hypershift deps still. thought rebasing against master without my go.mod commit then only updating the api dep in go.mod would fix this but it did not so i am still debugging how to fix that, but once that is taken care of this will be good to merge

bindata/network/ovn-kubernetes/managed/common/004-config.yaml
```diff
@@ -136,13 +136,19 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+{{- if (index . "V4InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
+{{- end }}
+{{- if (index . "V6InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
```
rbbratta, line 143: Typo?  should this be `v6-internal-masquerade-subnet` here?

bindata/network/ovn-kubernetes/self-hosted/common/004-config.yaml
```diff
@@ -62,12 +62,19 @@ data:
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
 
+{{- if (index . "V4InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
+{{- end }}
+{{- if (index . "V6InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
```
rbbratta, line 69: Typo?  `v6`?

bindata/network/ovn-kubernetes/managed/common/004-config.yaml
rbbratta: Upstream is `v6-masquerade-subnet`?   Did we add `internal-` somewhere?

bpickard22: good catch on the typo corrected,

bpickard22: yes we added the internal name https://github.com/openshift/api/pull/1410/files#diff-2e98c189fdf696589c6d28124557b9cf96a329e82f790b4e61209351cc68648d

ricky-rav (changes requested):
the changes look good to me, but I would merge the two commits into one and remove the commented debug code that you left in ovn_kubernetes.go. Other than that, all good. Thanks, Ben! :)

ricky-rav:
/lgtm
Thanks, Ben!

bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
```diff
@@ -66,6 +66,13 @@ data:
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
 
+{{- if (index . "V4InternalMasqueradeSubnet")}}
+    v4-internal-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
+{{- end }}
+{{- if (index . "V6InternalMasqueradeSubnet")}}
+    v6-internal-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
+{{- end }}
+
```
jcaamano, line 75: I don't think that this will cause a rollout/restart of ovn-k, making it a day-1 only configuration parameter. Is that ok?

pkg/network/ovn_kubernetes.go
```diff
@@ -861,14 +870,66 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+		}
+		if oc.GatewayConfig != nil {
+			if oc.GatewayConfig.IPv4.InternalMasqueradeSubnet != "" {
+				if !cnHasIPv4 {
+					out = append(out, errors.Errorf("v4InternalMasqueradeSubnet %s and ClusterNetwork must have matching IP families", oc.GatewayConfig.IPv4.InternalMasqueradeSubnet))
+				}
+				_, v4Net, err = net.ParseCIDR(oc.GatewayConfig.IPv4.InternalMasqueradeSubnet)
+				if err != nil {
+					out = append(out, errors.Errorf("v4InternalMasqueradeSubnet is invalid: %s", err))
+				}
```
jcaamano, line 893: I see in other places that the error returned from `net.ParseCIDR` is ignored, I suspect because it is assumed to be properly validated by the API server. But here you don't. If there is an error here, are you sure you can derenference safely `*v4Net` later on?

pkg/network/ovn_kubernetes.go
```diff
@@ -861,14 +870,66 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
+						_, v4ClusterNet, _ := net.ParseCIDR(cn.CIDR)
+						if iputil.NetsOverlap(*v4Net, *v4ClusterNet) {
+							out = append(out, errors.Errorf("v4InternalMasqueradeSubnet %s overlaps with ClusterNetwork %s", oc.GatewayConfig.IPv4.InternalMasqueradeSubnet, cn.CIDR))
+						}
+					}
+				}
+				for _, sn := range conf.ServiceNetwork {
+					if utilnet.IsIPv4CIDRString(sn) {
+						_, v4ServiceNet, _ := net.ParseCIDR(sn)
+						if iputil.NetsOverlap(*v4Net, *v4ServiceNet) {
```
jcaamano, line 905: Should we check that there is no overlap with the join subnet as well? Perhaps it would be useful to have an utility method to pass a map of network names to networks, that just checks that there are no overlaps among any of them.

jcaamano (changes requested):
I think that the first commit needs to be dropped as it should be merged already.

bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
bpickard22: yes that is the current intended behavior

pkg/network/ovn_kubernetes.go
bpickard22: correct the api would catch any invalid subnet configuration (or at least it should), just being cautious here to catch anything that i missed with my api validations. I am unsure if a value that could cause a panic would occur here because we already check for nil

pkg/network/ovn_kubernetes.go
bpickard22: I think I would like to get this in before code freeze and then add some utility that allows us to check for overlap when we add the transit switch config this coming release, I can add a todo here if thats alright.

pkg/network/ovn_kubernetes.go
jcaamano: Where do you check for nil?

pkg/network/ovn_kubernetes.go
jcaamano: Regardless of the utility, don't you think we need to check for overlaps with the join subnet in this PR?

bpickard22: yes will add that now

pkg/network/ovn_kubernetes.go
bpickard22: we check that the gatewayConfig is not nil, so if no value is passed into the internalMasqueradeSubnet it should just be an empty string im pretty sure

pkg/network/ovn_kubernetes.go
bpickard22: @jcaamano added a check for just the join and masq subnets here, will add a utility when the transit swtich config is introduced here

pkg/network/ovn_kubernetes.go
jcaamano: So you have
```
_, v4MasqNet, err = net.ParseCIDR(oc.GatewayConfig.IPv4.InternalMasqueradeSubnet)
```
if that returns non-nil `err` and potentially nil `v4MasqNet` then you just append the error but continue on
```
out = append(out, errors.Errorf("v4InternalMasqueradeSubnet is invalid: %s", err)
```
and then a few lines below you dereference v4MasqNet which will panic if it is nil
```
if iputil.NetsOverlap(*v4MasqNet, *v4ClusterNet)
```

bpickard22: @jcaamano added a check to make sure that the v4/v6 masqNet return isnt nil

pkg/network/ovn_kubernetes.go
```diff
@@ -813,45 +822,53 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
 			}
-			_, v4Net, err = net.ParseCIDR(oc.V4InternalSubnet)
+			_, v4JoinNet, err = net.ParseCIDR(oc.V4InternalSubnet)
 			if err != nil {
 				out = append(out, errors.Errorf("v4InternalSubnet is invalid: %s", err))
 			}
+			if v4JoinNet == nil {
+				out = append(out, errors.Errorf("Unable to parse cidr for v4InternalSubnet %s", oc.V4InternalSubnet))
+				return out
+			}
```
jcaamano, line 838: If `v4JoinNet == nil` then `err != nil` so it doesn't make sense to check and append an error for both things.
Also this method is trying to validate all without breaking out early, I think we can keep doing that.

I would do this
```
_, v4JoinNet, err = net.ParseCIDR(oc.V4InternalSubnet)
if err != nil {
    out = append(out, errors.Errorf("v4InternalSubnet is invalid: %s", err))
} else if !isV4InternalSubnetLargeEnough(conf) {
    out = append(out, errors.Errorf("v4InternalSubnet %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", oc.V4InternalSubnet))
}

...
...

    if v4JoinNet != nil {
        _, v4ClusterNet, _ := net.ParseCIDR(cn.CIDR)
	if iputil.NetsOverlap(*v4JoinNet, *v4ClusterNet) {
	    out = append(out, errors.Errorf("v4InternalSubnet %s overlaps with ClusterNetwork %s", oc.V4InternalSubnet, cn.CIDR))
	}
    }
```

Similar approach on the other instances.

bpickard22: latest change implements this

jcaamano: You missed the second part, I added a comment for it

pkg/network/ovn_kubernetes.go
```diff
@@ -813,62 +822,138 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
 				out = append(out, errors.Errorf("v4InternalSubnet is invalid: %s", err))
+			} else if !isV4InternalSubnetLargeEnough(conf) {
+				out = append(out, errors.Errorf("v4InternalSubnet %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", oc.V4InternalSubnet))
 			}
-			if !isV4InternalSubnetLargeEnough(conf) {
-				out = append(out, errors.Errorf("v4InternalSubnet is no large enough for the maximum number of nodes which can be supported by ClusterNetwork"))
+			if v4JoinNet != nil {
+				for _, cn := range conf.ClusterNetwork {
+					if utilnet.IsIPv4CIDRString(cn.CIDR) {
+						if oc.V4InternalSubnet != "" {
```
jcaamano, line 840: You should replace `oc.V4InternalSubnet != ""` with `v4JoinNet != nil`

Similar in other instances

bpickard22: I actually think that is a redundant if check that can just be removed

PR #1810, opened 2023-05-11
Remove unused Dockerfile.rhel7 file

martinkennelly:
This can be accomplished because of PR:
https://github.com/openshift/release/pull/38079

cc @jluhrsen


Dockerfile.rhel7
```diff
@@ -1,14 +0,0 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.14 AS builder
```
martinkennelly, line 1: @jluhrsen There is a different here between Dockerfile and this - in Dockerfile its: 
```
FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.14 AS builder
```

dcbw:
Ugh, we actually *do* use Dockerfile.rhel7:

[ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-network-operator/openshift-cluster-network-operator-master.yaml#L30)

So we have to first change that to `Dockerfile` and then we can kill the .rhel7 version.

Dockerfile.rhel7
jluhrsen: so I think we need this to fix it back up:
https://github.com/openshift/release/pull/39281

dcbw:
See https://issues.redhat.com/browse/DPTP-3458 to figure out why our Dockerfile rename changes keep getting overwritten.

martinkennelly:
depends on https://github.com/openshift/release/pull/39364

dcbw:
openshift/release PR merged so

/retest

PR #1459, opened 2022-05-25
Bug 2090437: Bump CNO to k8s 1.24

msherif1234:
update CNO to kube1.24
Signed-off-by: Mohamed Mahmoud <mmahmoud@redhat.com>

PR #1816, opened 2023-05-22
OCPBUGS-13891: [release-4.12] HyperShift: Support HostedControlPlane node selector

kyrtapz:
This is a clean backport of https://github.com/openshift/cluster-network-operator/pull/1802.
The change is the first commit bumping the hypershift dependency, this was requried since we use different dependencies in 4.12 and 4.13.

Additionally this PR cherry-picks: https://github.com/openshift/cluster-network-operator/pull/1837/commits/ff5d378a5e8a9af3d20692d11c1b9270549d075e

/cc @jcaamano @martinkennelly


go.mod
```diff
@@ -80,12 +79,12 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
-	github.com/openshift/hypershift v0.0.0-20220525174911-c7c2b57c98ca
+	github.com/openshift/hypershift v0.1.2
```
jcaamano, line 82: I am a bit confused here, would it make sense to put this in the first require section or in the replace section?

kyrtapz: Good spot!
I don't know why this moved to the second require, shouldn't only the first one have direct dependencies?
It looks like it happened here: https://github.com/openshift/cluster-network-operator/commit/f58b1a997381731e45801917c4e6d7fe1e5c48ff
If we decide to change that I would prefer to do it in master first, wdyt?

kyrtapz: https://github.com/openshift/cluster-network-operator/pull/1459#discussion_r882638796 maybe it was just a mistake

csrwng:
/hold
until this is fixed: https://issues.redhat.com/browse/OCPBUGS-14476

jeffnowicki:
> /hold until this is fixed: ~https://issues.redhat.com/browse/OCPBUGS-14476~

@csrwng this is the issue/bug you intended to reference: https://issues.redhat.com/browse/OCPBUGS-14714

kyrtapz:
> /hold until this is fixed: https://issues.redhat.com/browse/OCPBUGS-14714

Added the fix in the last commit

celebdor:
https://issues.redhat.com/browse/OCPBUGS-14714 is already verified. I think the hold can be lifted.
/hold cancel

PR #1822, opened 2023-05-25
OCPBUGS-11882: Added safe-to-evict annotation to ovnkube-master pods

jparrill:
This PR fixes part of this bug: https://issues.redhat.com/browse/OCPBUGS-11882.

Basically adding this annotation to the pods which contains volumes or emptyDirs will allow the ClusterAutoscaler to drain the pods properly.


bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
```diff
@@ -53,6 +53,7 @@ spec:
       annotations:
         hypershift.openshift.io/release-image: {{.ReleaseImage}}
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
```
enxebre, line 56: does this annotation somehow skip PDBs? If so that's highly undesired and will need to use `safe-to-evict-local-volumes` https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

enxebre: yeh we don't want this behaviour https://github.com/kubernetes/autoscaler/blob/d1740df93b76d99bf9302ad0c62978deb0ec1d5b/cluster-autoscaler/utils/drain/drain.go#L39-L40

jparrill: I've changed that just pointing to the `EmptyDir` concrete volumes. Also added to the `multus-admission-controller` deployment.

PR #1826, opened 2023-06-01
Rebase to kube 1.27

trozet:
Past most errors, remaining this seem to be brokenin cloud specific cluster-api providers


flavio-fernandes:
/cc @trozet  : 

```
INFO [runner] linters took 1m5.104670358s with stages: goanalysis_metalinter: 1m5.039557737s, structcheck: 21.726µs
pkg/controller/allowlist/allowlist_controller.go:179:9: SA1019: wait.Poll is deprecated: This method does not return errors from context, use PollWithContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release. (staticcheck)
        return wait.Poll(time.Second, time.Minute, func() (done bool, err error) {
               ^
cmd/cluster-network-operator/main.go:24:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
        rand.Seed(time.Now().UTC().UnixNano())
        ^
pkg/controller/operconfig/mtu_probe.go:69:8: SA1019: wait.PollWithContext is deprecated: This method does not return errors from context, use PollWithContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release. (staticcheck)
        err = wait.PollWithContext(ctx, 5*time.Second, 100*time.Second, func(ctx context.Context) (bool, error) {
              ^
pkg/network/ovn_kubernetes.go:1103:9: SA1019: wait.PollImmediate is deprecated: This method does not return errors from context, use PollWithContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release. (staticcheck)
        err := wait.PollImmediate(OVN_MASTER_DISCOVERY_POLL*time.Second, time.Duration(timeout)*time.Second, func() (bool, error) {
               ^
pkg/network/ovn_kubernetes.go:1122:5: SA1019: wait.ErrWaitTimeout is deprecated: This type will be made private in favor of Interrupted() for checking errors or ErrorInterrupted(err) for returning a wrapped error. (staticcheck)
        if wait.ErrWaitTimeout == err {
           ^
pkg/controller/proxyconfig/controller.go:236:3: assigned to trustBundle, but reassigned without using the value (wastedassign)
                trustBundle, err = r.mergeTrustBundlesToConfigMap(proxyData, systemData)
                ^
pkg/controller/signer/signer-controller.go:138:3: assigned to csr, but never used afterwards (wastedassign)
                csr, err = r.clientset.CertificatesV1().CertificateSigningRequests().UpdateApproval(ctx, request.Name, csr, metav1.UpdateOptions{})
                ^
```

vendor/sigs.k8s.io/controller-runtime/pkg/client/fake/client.go
```diff
@@ -48,13 +61,15 @@ import (
...
 }
 
 type fakeClient struct {
-	tracker    versionedTracker
-	scheme     *runtime.Scheme
-	restMapper meta.RESTMapper
+	tracker               versionedTracker
+	scheme                *runtime.Scheme
+	restMapper            meta.RESTMapper
+	withStatusSubresource sets.Set[schema.GroupVersionKind]
```
flavio-fernandes, line 72: .

jcaamano (approved):
Just a nit

go.mod
```diff
@@ -152,6 +151,7 @@ require (
 	k8s.io/kms v0.27.2 // indirect
 	k8s.io/kube-aggregator v0.26.1 // indirect
 	sigs.k8s.io/cluster-api-provider-aws/v2 v2.0.2 // indirect
+	sigs.k8s.io/cluster-api-provider-ibmcloud v0.2.4 // indirect
```
kyrtapz, line 154: Just curious, was it needed to move that here?

trozet: wasnt needed, just a side effect of me hacking at this so many times

pkg/operator/operator.go
```diff
@@ -3,6 +3,8 @@ package operator
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"sigs.k8s.io/controller-runtime/pkg/config"
```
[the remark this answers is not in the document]
trozet: yeah...since that is the only issue I'm inclined to leave it for now and fix it later.

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -69,6 +70,14 @@ func set(t *testing.T, client cnoclient.Client, obj crclient.Object) {
 	}
 }
 
+func setStatus(t *testing.T, client cnoclient.Client, obj crclient.Object) {
```
flavio-fernandes, line 73: TYSM, @trozet !!! you rock.

PR #1844, opened 2023-06-21
Enable EgressService controller

oribon:
* Add the CRD
* Add RBAC permissions


oribon:
/hold
https://github.com/ovn-org/ovn-kubernetes/pull/3567 should be merged downstream first

PR #1845, opened 2023-06-21
OCPBUGS-18114: Update node selector in various YAMLs

ricky-rav:
- Replace deprecated `beta.kubernetes.io/os` node selector with `kubernetes.io/os`
https://kubernetes.io/docs/reference/labels-annotations-taints/#beta-kubernetes-io-os-deprecated
- Add linux node selector to openshift-sdn/sdn-controller daemonset

### DIFF
--- a/bindata/network/openshift-sdn/controller.yaml
+++ b/bindata/network/openshift-sdn/controller.yaml
@@ -116,8 +116,9 @@ spec:
             mountPath: /etc/pki/tls/metrics-certs
             readOnly: True
       hostNetwork: true
-{{- if not .ExternalControlPlane }}
       nodeSelector:
+        kubernetes.io/os: "linux"
+{{- if not .ExternalControlPlane }}
         node-role.kubernetes.io/master: ""
 {{- end }}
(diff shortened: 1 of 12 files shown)


bindata/network/openshift-sdn/controller.yaml
```diff
@@ -119,6 +119,7 @@ spec:
 {{- if not .ExternalControlPlane }}
       nodeSelector:
         node-role.kubernetes.io/master: ""
+        kubernetes.io/os: "linux"
```
kyrtapz, line 122: Should `kubernetes.io/os: "linux"` selector be limited only to deployments without `ExternalControlPlane`? I don't think it should

rbbratta:
https://issues.redhat.com/browse/OCPBUGS-18114

bindata/network/openshift-sdn/controller.yaml
ricky-rav: true

flavio-fernandes (commented):
/lgtm

PR #1849, opened 2023-06-23
move IPsec to host

yuvalk:
copying the commit msg:

    ipsec: move pluto to host
    
    this is done to allow co-existance with host configured ipsec tunnels
    
    To avoid mounting the whole host fs, only the configuration files are
    shared. including NSS DB and the whole ipsec configuration
    
    hostPID is used, so that the pid file make sense from the host as well.
    that required moving the termination logic to a preStop lifecycle step.
    
    the rest stays pretty much the same :-)

upgrade now works.
the interim solution for 4.14 is to keep 2 sets of daemonsets
where only one is active, according to whether libreswan is found on the host or not.


bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -251,24 +268,50 @@ spec:
...
           type: DirectoryOrCreate
-      - name: signer-ca
-        configMap:
+        name: host-var-run-ovs
+      - configMap:
+          defaultMode: 420
           name: signer-ca
-      - name: etc-openvswitch
-        hostPath:
+        name: signer-ca
```
rbbratta, line 282: It doesn't look like there is an impact on cert-rotation, but asking anyway for [SDN-4039](https://issues.redhat.com/browse/SDN-4039)

yuvalk: the impact might be related to the fact that the NSS DB is now persistent, so better remove old cert&keys otherwise eventually it'll explode.

yuvalk:
> ovn-ipsec pods stuck in Init.

yeah, because it's missing https://github.com/openshift/release/pull/40898

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+          # Each node has a key pair which is used by OVS to encrypt/decrypt/authenticate traffic
+          # between each node. The CA cert is used as the root of trust for all certs so we need
+          # the CA to sign our certificate signing requests with the CA private key. In this way,
+          # we can validate that any signed certificates that we receive from other nodes are
+          # authentic.
+          echo "Configuring IPsec keys"
+
+          # If the certificate does not exist or it will expire in the next 6 months
+          # (15770000 seconds), we will generate a new one.
+          if [ ! -e /etc/openvswitch/keys/ipsec-cert.pem ] || [ ! openssl x509 -noout -dates -checkend 15770000 ];
```
rbbratta, line 70: shellcheck says this is invalid.  `don't wrap commands in []/[[]]`

`[ ! openssl x509 -noout -dates -checkend 15770000 ]`  doesn't seem to make sense, maybe it has never worked?


```shell
bash: [: too many arguments
```

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+        command:
+        - /bin/bash
+        - -c
+        - |
+          #!/bin/bash
+          set -exuo pipefail
+
+          # check if need to run on host
+          rpm -qa libreswan
+          if [ "$?" -eq 0 ]; then
```
rbbratta, line 53:
```shell
if rpm -q libreswan; then
    echo "host has libreswan, ipsec will run on host"
    exit 0
fi
```
We can use `rpm -q` instead of `rpm -qa` for specific packages?

We can use the if statement with just the `rpm -q` command.

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+            spec:
+              request: ${csr_64}
+              signerName: network.openshift.io/signer
+              usages:
+              - ipsec tunnel
+          EOF
+
+            # Wait until the certificate signing request has been signed.
+            counter=0
+            until [ ! -z $(kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' 2>/dev/null) ]
```
rbbratta, line 113: `[! -z ` should be `[ -n `

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+              fi
+            done
+
+            # Decode the signed certificate.
+            kubectl get csr/$(hostname) -o jsonpath='{.status.certificate}' | base64 -d | openssl x509 -outform pem -text -out /etc/openvswitch/keys/ipsec-cert.pem
+
+            kubectl delete csr/$(hostname)
+
+            # Get the CA certificate so we can authenticate peer nodes.
+            cat /signer-ca/ca-bundle.crt | openssl x509 -outform pem -text > /etc/openvswitch/keys/ipsec-cacert.pem
```
rbbratta, line 130:
```shell
openssl x509 -outform pem -text < /signer-ca/ca-bundle.crt > /etc/openvswitch/keys/ipsec-cacert.pem
```

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+            # Generate an SSL private key and use the key to create a certitificate signing request
+            umask 077 && openssl genrsa -out /etc/openvswitch/keys/ipsec-privkey.pem 2048
+            openssl req -new -text \
+                        -extensions v3_req \
+                        -addext "subjectAltName = DNS:${cn}" \
+                        -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
+                        -key /etc/openvswitch/keys/ipsec-privkey.pem \
+                        -out /etc/openvswitch/keys/ipsec-req.pem
+
+            csr_64=$(cat /etc/openvswitch/keys/ipsec-req.pem | base64 | tr -d "\n")
```
rbbratta, line 87:
```shell
csr_64=$(base64 -w0 /etc/openvswitch/keys/ipsec-req.pem)
```

`-w0` to avoid line-wrap

rbbratta (commented):
our bash is poor quality.

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+            kill -9 $(cat /var/run/pluto/pluto.pid 2>/devnull) 2>/dev/null || true
+
+            /usr/sbin/ipsec --stopnflog
+            exit 0
+          }
+          trap cleanup SIGTERM
+
+          # check if need to run on host
+          rpm -qa libreswan
+          if [ "$?" -eq 0 ]; then
```
rbbratta, line 191: same.  `if rpm -q libreswan; then`

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+
+            # Don't allow ovs monitor to cleanup persistent state
+            kill $(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null) 2>/dev/null || true
+            # Don't allow pluto to clear xfrm state and policies on exit
+            kill -9 $(cat /var/run/pluto/pluto.pid 2>/devnull) 2>/dev/null || true
+
+            /usr/sbin/ipsec --stopnflog
+            exit 0
+          }
+          trap cleanup SIGTERM
```
rbbratta, line 187: the trap will still run on `delete`?  If we have host `libreswan` we want to disable `cleanup`?

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+          echo "ovnkube-node has configured node."
+
+          # After a restart of this container (or on initial startup), we flush xfrm state and policy
+          # before we start pluto and ovs-monitor-ipsec in order to start in a known good state. This
+          # will result in a small interruption in traffic until pluto and ovs-monitor-ipsec start again.
+          ip x s flush
+          ip x p flush
+
+          # Workaround for https://github.com/libreswan/libreswan/issues/373
+          ulimit -n 1024
```
rbbratta, line 217: Hopefully host `libreswan` doesn't have this `ulimit` issue.  TODO: check the host RPM.

https://github.com/libreswan/libreswan/issues/373

rbbratta: Fixed in https://github.com/libreswan/libreswan/commit/5d9bd100e85181a68a8ee3e5ab3a7e8e9e2e403a

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -0,0 +1,298 @@
...
+
+          # Every time we restart this container, we will create a new key pair if
+          # we are close to key expiration or if we do not already have a signed key pair.
+          #
+          # Each node has a key pair which is used by OVS to encrypt/decrypt/authenticate traffic
+          # between each node. The CA cert is used as the root of trust for all certs so we need
+          # the CA to sign our certificate signing requests with the CA private key. In this way,
+          # we can validate that any signed certificates that we receive from other nodes are
+          # authentic.
+          echo "Configuring IPsec keys"
```
rbbratta, line 66: If we skip this code when using host-based E-W we move cert-rotation into a manual process which would require documentation for FIPS?

docs/enabling_ns_ipsec.md
```diff
@@ -0,0 +1,100 @@
...
+        echo "importing cert to NSS"
+        certutil -A -a -i /etc/pki/certscacert.p12 -d sql:/var/lib/ipsec/nss -n "CAname" -t 'CT,,'
+        certutil -A -a -i /etc/pki/certscert.p12 -d sql:/var/lib/ipsec/nss -n "Certname" -t 'CT,,'
+  systemd:
+     units:
+     - name: ipsec-import-certs
+       enabled: true
+       contents: |
+         [Unit]
+         Description=Imort external certs into ipsec NSS
```
rbbratta, line 85: s/Imort/Import/

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: no, cert rotation for E-W is still handled the exact same way (in the init container)
the diff is that the key is later loaded into the host NSS 

BTW: ovs_monitor, will finally have what to cleanup :-)
 https://github.com/openvswitch/ovs/blob/master/ipsec/ovs-monitor-ipsec.in#L731
 
 
 BBTW: for N-S we are not handling certs at all, so there rotation is manual/ handled by customer. (can use cert-manager, etc)

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: note that the ipsec-pre414.yaml
is exactly the script that existed there before this PR (was called ipsec.yaml)

I'll rearrange the PR so that it's a `git mv`

historically, I initially thought we wont need it anymore, but We've agreed (with @trozet) that for now we'll hold both to avoid upgrade issues

yuvalk:
> our bash is poor quality.

yeah, but I suggest dealing with it separately 

after I see this change works properly, 
I can add an additional commit for that (or even lets have it in it's own PR)

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: I agree `-n` would be more readable in this case

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: `openssl x509 -in /etc/openvswitch/keys/ipsec-cert.pem -outform pem -text -out /etc/openvswitch/keys/ipsec-cacert.pem`

yuvalk:
> poor

it have some issues
but some of it is just "old school" :-)

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: the pre414 is for case where there is not host libreswan

anyhow, in both cases we want to guard against xfrm state flushes,
so you'll see I had to move it, in the new code, to pod lifecycle, because the host part is now using hostPID, for which kubelet wont do SIGTERM at all..

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
yuvalk: well, seems you are right - it never worked:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1928/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1687090636676665344/artifacts/e2e-ovn-ipsec-step-registry/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovn-ipsec-9kpnd_ovn-keys.log

zshi-redhat:
> *** Still WIP - till I implement the upgrade part

@yuvalk Should this be removed in the commit message?

yuvalk:
/retest
now that https://github.com/openshift/release/pull/40898 is merged

bindata/network/ovn-kubernetes/common/ipsec-pre414.yaml
```diff
@@ -174,6 +182,11 @@ spec:
           }
           trap cleanup SIGTERM
 
+          if rpm --dbpath=/usr/share/rpm -q libreswan; then
+            echo "host have libreswan, ovn-ipsec will sleep infinity"
```
trozet, line 186: "host has libreswan and therefore ipsec will be configured by <other host IPSEC DS name?>, this ovn ipsec container will sleep to infinity"

docs/enabling_ns_ipsec.md
```diff
@@ -0,0 +1,100 @@
...
+         After=network.target
```
trozet, line 86: does this need to happen before networking or kubelet?

trozet (commented):
does upgrade now work? If so update the description? Also include some more detail of how that works.

yuvalk:
> does upgrade now work? If so update the description? Also include some more detail of how that works.

yes, upgrade works, as we agreed
I've modified the description

yuvalk:
> should either remove the code, or add a comment above it referencing why it is commented out

good catch, I'll delete the line

but wanted to elaborate here
or see the commit msg for 50c0b669bc58b942647a1130bbc10f5c2024f0d3:
```
    ipsec: remove csr delete
    
    it seems that after commit 10dd8c57a98558c03f6aa72a6e8d90292db890ba, we dont need to delete the csr anymore. in fact, we no longer have the needed permissions.
    the only reason we didnt hit this in CI is an issue  after my fix of the if statememnt,
    this is actually failing ovn-keys, so I had to fix it
    
    see that PR#1928 logs - https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/1928/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1687090636676665344/artifacts/e2e-ovn-ipsec-step-registry/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovn-ipsec-9kpnd_ovn-keys.log
    
    this fixes #1960
```
the reason we didnt hit an issue there before is #1961 (which I've also fixed)

anyhow, I'll modify the commit so line will be removed

docs/enabling_ns_ipsec.md
yuvalk: hmm, good catch!
before network..

note: this doc is for n-s which is a host thing.

yuvalk: not sure what's the right way to run something before network though

yuvalk: the after network is not that bad, because it's a very short script that will probably still run before network-online is reached (where ipsec is started)

bindata/network/ovn-kubernetes/common/ipsec.yaml
```diff
@@ -58,10 +58,11 @@ spec:
           # authentic.
           echo "Configuring IPsec keys"
 
+          cert_pem=/etc/openvswitch/keys/ipsec-cert.pem
```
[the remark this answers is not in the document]
yuvalk: so I verified, the openssl command would fail if the cert does not exist (as well as many other conditions)
I think we're good here

trozet (approved):
/lgtm

PR #1851, opened 2023-06-27
OCPBUGS-15282: Add release version annotation to whereabouts-reconciler

jcaamano:
CNO uses it to track whether updates to the daemonset have been done or
not. If not set, CNO won't set the overall operator version to the
target upgrade and the upgrade won't complete.


jcaamano:
@dougbtv @weliang1 PTAL

jcaamano:
Not related, opened bug https://issues.redhat.com/browse/OCPBUGS-15475

/override ci/prow/e2e-vsphere-ovn-windows

PR #1866, opened 2023-07-05
OCPBUGS-15794: fix: add missing annotation for workload partitioning

eggfoobar:
To support the workload partitioning we need to add this required annotations
(https://github.com/openshift/enhancements/pull/703).


eggfoobar:
/unhold

The original failure was due to unrelated hypershift provisioning

PR #1869, opened 2023-07-06
Change rhel7/8 to rhel8/9

s1061123:
Now that the binary is being dynamically compiled downstream, the binary must match the RHEL version of RHCOS because it is copied to the host.


dcbw:
/retest
capacity issues

jupierce:
No failures which seem connected to this change. Urgent need to merge this for FIPS remediation, so 
/unhold

PR #1871, opened 2023-07-07
NETOBSERV-1047: Add a cluster monitoring dashboard for ovn

jotak:
Some screenshots:

**Networking / Linux Subsystem Stats**

Capture d’écran du 2023-10-24 10-39-33

Capture d’écran du 2023-10-24 10-39-39

**Networking / Infrastructure**

Capture d’écran du 2023-10-24 10-47-16

Capture d’écran du 2023-10-24 10-47-25

Capture d’écran du 2023-10-24 10-47-32


jotak:
cc @martinkennelly @npinaeva 
Just opening as draft, not yet asking for a detailed review, but in order to validate if I'm starting with the right approach - and to have a basis for discussion.
One thing that isn't clear yet to me is what's the idiomatic way to create resources at the operator startup, but programmatically (not as a static manifest) - I've found for watching / reconciling, but not for the initial creation

jotak:
PS: don't pay attention to the dashboard content itself: it's just a temporary copy-pasted thing

jotak:
Marking as ready for review. The dashboards content is not finalized, but the controller mechanics can be reviewed

jotak:
There's a question I have here, given the fact the CNO runs in-payload / pre-configured: the dashboards deployed here are not really meant to be reconciled, they could simply be installed once for all at startup. Does it worth it to have a controller/reconciler then? Any idea if/how I could do that differently?

jotak:
Current OVN dashboard:

image

image

jotak:
OVN dashboard after last commit:

image

pkg/controller/dashboards/dashboard_controller.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (r *ReconcileDashboard) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	klog.Infof("Reconcile dashboards")
+
+	// Fetch the Network.operator.openshift.io instance to get Network Type
+	operConfig := &operv1.Network{TypeMeta: metav1.TypeMeta{APIVersion: operv1.GroupVersion.String(), Kind: "Network"}}
+	err := r.client.Default().CRClient().Get(ctx, types.NamespacedName{Name: names.CLUSTER_CONFIG}, operConfig)
+	if err != nil {
+		err = fmt.Errorf("unable to retrieve Network.operator.openshift.io object: %w", err)
```
martinkennelly, line 117: nit: Can we start the error with a capital letter as outputting to klog usually starts with a capital letter? I assume its fine for the degraded message too. Same below.

pkg/controller/dashboards/dashboard_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		r.status.SetDegraded(statusmanager.DashboardConfig, "DashboardError", err.Error())
+		return reconcile.Result{}, err
+	}
+
+	err = r.applyManifests(ctx, operConfig)
+	if err != nil {
+		err = fmt.Errorf("failed to apply dashboard manifests: %w", err)
+		klog.Error(err)
+		r.status.SetDegraded(statusmanager.DashboardConfig, "DashboardError", err.Error())
+		// Do not throw an error
```
martinkennelly, line 128: why? If youve a reason, document it here.

pkg/controller/dashboards/dashboard_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	r.status.SetNotDegraded(statusmanager.DashboardConfig)
+
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileDashboard) applyManifests(ctx context.Context, cfg *operv1.Network) error {
+	klog.Infof("Applying dashboards manifests")
```
martinkennelly, line 138: nit: `klog.Info()`

pkg/controller/dashboards/dashboard_controller.go
```diff
@@ -0,0 +1,160 @@
...
+var (
+	//go:embed netstats.json
+	netstatsContent string
+	//go:embed ovn-health.json
+	ovnHealthContent string
+	dashboardRefs    []dashboardRef = []dashboardRef{
+		{
+			name:      "grafana-dashboard-network-stats",
+			json:      netstatsContent,
+			tplSuffix: "NetStats",
```
martinkennelly, line 43: super nit: feel free to ignore - put a comment here that if they alter `tplsuffixs`, they must update the config maps?

pkg/controller/dashboards/netstats.json
```diff
@@ -0,0 +1,1495 @@
...
+                            "intervalFactor":2,
+                            "legendFormat":"{{dev}}",
+                            "refId":"A"
+                        }
+                    ],
+                    "thresholds":[
+                    ],
+                    "timeFrom":null,
+                    "timeShift":null,
+                    "title":"Bytes Sent, by Device",
```
martinkennelly, line 690: It can be any linux network interface right? I _think_ `interface` is better than device.

pkg/controller/dashboards/netstats.json
```diff
@@ -0,0 +1,1495 @@
+{
+    "__inputs":[
```
martinkennelly, line 2: Is there any way to put a summary/description at the top of these dashboards?

pkg/controller/dashboards/ovn-health.json
```diff
@@ -0,0 +1,1249 @@
...
+                            "show":false
+                        }
+                    ]
+                }
+            ],
+            "repeat":null,
+            "repeatIteration":null,
+            "repeatRowId":null,
+            "showTitle":true,
+            "title":"Network Setup",
```
martinkennelly, line 657: what about `Network Configuration`?

martinkennelly:
@CFields651 Can you take a look at this?
/cc @venkataanil

pkg/controller/dashboards/netstats.json
martinkennelly: Id like to inform the user that OVS/OVN drops arent included here somehow and this is just linux networking stack drops. I dont know how to accomplish this.

martinkennelly:
Do you think you need help reviewing the promql expressions from the monitoring team or are you confident enough?

CFields651:
@jotak @martinkennelly it's great to see that progress is being made here!  I like the OVN dashboard with latency on pod creation and deletion.  I think it would be great to include some of the pod_network_connectivity_check_tcp_connect_latency_gauge counters.  The challenge is that there are a lot of them and the more nodes in the cluster the more guages you get.  Here's a suggestion; There is a check from where the network-check-source pod is running to the openshift api.  You can find the endpoint for the check like this:
oc get svc/api -n openshift-apiserver
NAME   TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)   AGE
api    ClusterIP   172.30.117.229   <none>        443/TCP   154d

Then the query looks like this:  
oc exec prometheus-k8s-0 -n openshift-monitoring --  promtool query instant  http://localhost:9090 'pod_network_connectivity_check_tcp_connect_latency_gauge{targetEndpoint="172.30.117.229:443"}'

I think that mapping that over time just like pod latency creation and deletion would be valuable.

pkg/controller/dashboards/dashboard_controller.go
jotak: Hmm perhaps I'm wrong here ... My rationale would be that dashboards aren't a critical piece of the CNO, so an error there shouldn't bring the whole thing down. But I'm not used to dealing with multiple controllers , I guess if the error is propagated here, it won't affect the other controllers and the CNO globally ?

pkg/controller/dashboards/netstats.json
jotak: yes correct. I was reusing the node-exporter terminology (they use "device" or "dev") but I imagine "interface" is more correct - that's also what we use in netobserv

pkg/controller/dashboards/netstats.json
jotak: I don't think we can have a dashboard description text (I can ask the monitoring team) - but how about adding this precision in the drop panels, something like "at interface":
- Current Drops at Interface
- Packets Received Dropped at Interface, by Node
etc.
 ?

bindata/dashboards/configmaps.yaml
```diff
@@ -0,0 +1,31 @@
+{{ if .IsOVN }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{.DashboardNameOVNHealth}}
+  namespace: {{.DashboardNamespace}}
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
```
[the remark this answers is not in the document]
jotak: I assumed they are used for contextual installation - they are also set in many of the resources that the CNO creates - but let me ask on slack...
I guess there's no specific types of installations where we specifically don't want dashboards?

jotak:
@CFields651 interesting,
There's other options when we have too many metrics:
- We can show a topK (e.g. just the top 5 or such)
- We can show stats - for instance here I'm showing just the min, max and avg:

image

Side question: this metric is said as being ALPHA - do we expect any change soon (renaming, relabelling, etc.) ?

jotak:
Also, I assume the unit of this metric is nanos? It gives very high values such as 3_000_000
A good practice, when designing metrics, is to include the unit in the name, and also if possible to use the base unit (seconds here) given that everything is floating point in prometheus
cf https://prometheus.io/docs/practices/naming/#metric-names

jotak:
I can add a new section called "TCP latency probes" with text panels with avg/min/max and timeseries chart with these same three stats - let me try that

martinkennelly:
>Side question: this metric is said as being ALPHA - do we expect any change soon (renaming, relabelling, etc.) ?

Its a valid question. There is talk about this being possible disabled (it can be disabled via CNO option..) in some scenarios so i think bear this in mind.

>Small remark :-) - it's generally a good practice, when designing metrics, is to include the unit in the name, and also if possible to use the base unit (seconds here) given that everything is floating point in prometheus

Yeah, we didnt develop this component - the API developed it a long time ago for us. File bugs if you see something wrong with it. Its been badly neglected for years and few have paid attention to it.

jotak:
@CFields651  @martinkennelly here's a try with added TCP latency metric:

image

wdyt? keep it, change it or throw it away?

BTW: if this metric is optional, when turned off, it will show empty charts. We have several options then. I could build the dashboard more conditionally , but TBH, this isn't in the scope defined for this enhancement, I would rather do it in a follow-up - it would require a very different implementation. Or we can keep it as is, or remove these charts.
My opinion is that it's not so wrong if there's an empty chart there: users can collapse the corresponding row to not see it; and it's also a reminder to them that they can have more metrics if they enable that feature.

martinkennelly:
>I would rather do it in a follow-up 

Thats fine by me until we discover what else we want to add.
Network connectivity metrics is alpha and hasn't been maintained in awhile unfortunately. I am slow to build on it unless we have ownership in regards to maintaining it.

Its really hard for me to review your promql expressions and dashboard jsons in general as I am not an expert at that stuff.
Do you have anyone in mind who could have a look at them? monitoring team? You can raise request for a review with them via their ask for help func in slack on their team channel.

Setup of the dashboards lgtm.
Can you print out the dashboards here again?

martinkennelly:
I submitted a request to the monitoring team to review the promql expressions but hopefully the dashboard too. Not sure if they can or will. 

https://redhat-internal.slack.com/archives/C0VMT03S5/p1697801205443769

jotak:
btw, I don't know if I should worry or not about the failing tests, looking at details I don't see something related to my changes (but it's hard to me to navigate and find relevant info in all these logs), is it usual to have failing/flaky tests here?

jotak:
@martinkennelly I updated the top description to add up-to-date screenshots

jan--f:
> I submitted a request to the monitoring team to review the promql expressions but hopefully the dashboard too. Not sure if they can or will review the dashboard config part.
> 
> https://redhat-internal.slack.com/archives/C0VMT03S5/p1697801205443769

I came here via :arrow_up: . 
I skimmed the 3k lines of json and I have some notes:

- To me the dashboards look fine. I don't however know in which context these will be displayed. It might be worth checking with the console team if its feasible to align these dashboards with what the console does.
- I see a lot of range_vectors with short time ranges a la `rate(some_metric[1m])`. I don't know your prometheus config and specifically the scrape interval. But it bear mentioning that `rate` requires at least 2 samples to work. You should be confident that can always be maintained even if scrapes fail. For the in-cluster stack we have a scrape interval of 30s and use `4m` or longer for all range vectors.

And last but not least I wonder how you plan to maintain these dashboards and what the development workflow looks like? Changing them in Grafana and exporting the result can become very cumbersome and error prone (from experience). I'd recommend you look into generating dashboards from a templating language. For OCP monitoring we use the kubernetes-mixin project, that uses jsonnet to generate dashboards, e.g. https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/network-usage/cluster-total.libsonnet. While there is a learning curve and initial investment, the maintenance overhead is a lot more manageable.

martinkennelly:
@jotak I am on PTO for a week - when i come back, ill review.

jotak:
hey @jan--f 
thanks for your feedback!
- Can you elaborate on what you said with the console team is doing? Are you referring to the existing dashboards (such as in `kubernetes-mixin`) which, indeed, may have some overlap ? I think the main difference with these dashboards is the desire here to break down network metrics per interface ; also the fact that those existing dashboards are directly pulled from upstream projects make them less easy to amend than ones provided here, which we 100% control.
- On intervals: the issue with longer intervals is that it tends to smooth out the peaks and loose accuracy, especially when looking at 99p kinds of metric. I _think_ the scrape interval defined for these metrics are 15s, but I will double-check. In any case, during my testing, I didn't see missed datapoints.
- On jsonnet: I had a discussion about it with @simonpasquier and I told myself this wouldn't worth the investment. In fact, even if those json blobs are awful to read and navigate into, there are however more straightforward to edit compared to having several layers of abstraction. This works as long as we're fine with static dashboard, and that may change if at some point we need to display things conditionally. In netobserv, we've come up with an alternative way of [golang-based dashboard generation](https://github.com/netobserv/network-observability-operator/blob/1c0449c8d4c1d169e893aea343546e00ffd7d124/pkg/helper/dashboard.go). But I also hear that the monitoring team is working on a new dashboard framework based on "Perses" (not sure of the name) , so I wouldn't like to spend too much time investing in jsonnet if this had to be thrown away soon?

martinkennelly:
Looking for review from maintainer for this.
It /lgtm from me.

martinkennelly:
/assign @trozet 
Tim will review this week.

trozet (approved):
/lgtm

PR #1877, opened 2023-07-07
OCPBUGS-15962: Add logic to pick the ovn-k8s-cni-overlay binary from the right dir

abhat:
Add logic to pick the ovn-k8s-cni-overlay binary from the right source directory.

### DIFF
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
@@ -290,7 +290,49 @@ spec:
             source "/env/${K8S_NODE}"
             set +o allexport
           fi
-          cp -f /usr/libexec/cni/ovn-k8s-cni-overlay /cni-bin-dir/
+
+          function log()
+          {
+              echo "$(date --iso-8601=seconds) [cnibincopy] ${1}"
+          }
+
+          # collect host os information
+          . /host/etc/os-release
+          rhelmajor=
+          # detect which version we're using in order to copy the proper binaries
+          case "${ID}" in
+            rhcos|scos)
+              rhelmajor=$(echo $RHEL_VERSION | sed -E 's/([0-9]+)\.{1}[0-9]+(\.[0-9]+)?/\1/')
+            ;;
+            rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .)
+            ;;
+            fedora)
+              if [ "${VARIANT_ID}" == "coreos" ]; then
+                rhelmajor=8
+              else
+                log "FATAL ERROR: Unsupported Fedora variant=${VARIANT_ID}"
+                exit 1
+              fi
+            ;;
+            *) log "FATAL ERROR: Unsupported OS ID=${ID}"; exit 1
+            ;;
+          esac
+
+          # Set which directory we'll copy from, detect if it exists
+          sourcedir=/usr/libexec/cni/
+          case "${rhelmajor}" in
+          8)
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -290,7 +290,49 @@ spec:
...
+          sourcedir=/usr/libexec/cni/
+          case "${rhelmajor}" in
+          8)
+            sourcedir=/usr/libexec/cni/rhel8/
+          ;;
+          9)
+            sourcedir=/usr/libexec/cni/rhel9/
+          ;;
+          *)
+            log "ERROR: RHEL Major Version Unsupported, rhelmajor=${rhelmajor}"
```
wking, line 331: on future-proofing option would be `sourcedir="/usr/libexec/cni/rhel${rhelmajor}"`, which would fail in the `cp` below if that directory didn't exist or lacked the overlay shim.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -290,7 +290,49 @@ spec:
...
+          esac
+
+          # Set which directory we'll copy from, detect if it exists
+          sourcedir=/usr/libexec/cni/
+          case "${rhelmajor}" in
+          8)
+            sourcedir=/usr/libexec/cni/rhel8/
+          ;;
+          9)
+            sourcedir=/usr/libexec/cni/rhel9/
```
wking, line 328: nit: trailing `/` here will double up with the one you add explicitly in `$sourcedir/ovn-k8s-cni-overlay` below.  Maybe drop the trailing slashes from the `sourcedir` value, and rely on the explicit slash from the `$sourcedir/ovn-k8s-cni-overlay` expansion?

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -290,7 +290,49 @@ spec:
...
+          ;;
+          *)
+            log "ERROR: RHEL Major Version Unsupported, rhelmajor=${rhelmajor}"
+          ;;
+          esac
+          
+          cp -f $sourcedir/ovn-k8s-cni-overlay /cni-bin-dir/
```
wking, line 335: nit: current `sourcedir` population leaves no room for whitespace or other shell-sensitive characters, but quoting here `"${sourcedir}/ovn-k8s-cni-overlay"` gives you an extra layer of defense in case some future iteration of this script might include shell-sensitive characters in the `sourcedir` value.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
abhat: i would rather we explicitly fail it and know as early as possible. Of course in either case we will barf whether it's a few lines later or more explicitly earlier.

wking: in the "RHEL 10 gets a new `/rhel10` directory, using `sourcedir="/usr/libexec/cni/rhel${rhelmajor}"` would be one less place that needed touching.  Maybe:

```sh
sourcedir="/usr/libexec/cni/rhel${rhelmajor}"
test -d "${sourcedir}" || log "ERROR: RHEL Major Version Unsupported, rhelmajor=${rhelmajor}"
```

To get your explicit logs while still automatically picking up potential future `/rhel10`?

PR #1878, opened 2023-07-07
[WIP][DNM]Ic alert refactor

bpickard22:
PR to allow for review only, this commit will be given to @ricky-rav as part of his work in CNO for ovn-ic

@ricky-rav i need some eyes on the labels for the service monitor to make sure I am using the right syntax

Also need someone from the monitoring team to take a look at the alert expressions to make sure I am going about that correctly


tssurya:
@bpickard22 : please rebase this on top of #1838 : I'd like to see a green CI as far as alerts are concerned.

PR #1879, opened 2023-07-10
HOSTEDCP-1063: allow webhooks in hosted clusters to reach multus-admission-controller service

muraee:
We are preventing/deleting any webhook that runs in a hosted cluster and targets a service running in the management cluster here: https://github.com/openshift/hypershift/pull/2775

This PR adds a label to the `multus-admission-controller` service to ensure `multus.openshift.io` webhook doesn't get deleted.

Fixes #[HOSTEDCP-1063](https://issues.redhat.com/browse/HOSTEDCP-1063)

### DIFF
--- a/bindata/network/multus-admission-controller/001-service.yaml
+++ b/bindata/network/multus-admission-controller/001-service.yaml
@@ -6,6 +6,9 @@ metadata:
   namespace: {{.AdmissionControllerNamespace}}
   labels:
     app: multus-admission-controller
+{{- if .HyperShiftEnabled}}
+    hypershift.openshift.io/allow-guest-webhooks: "true"
+{{- end }}
   annotations:
 {{- if .HyperShiftEnabled}}
     network.operator.openshift.io/cluster-name: {{.ManagementClusterName}}


danwinship:
/approve
but I'll leave for someone else to lgtm. I'm not sure that this is the right solution, as opposed to moving the admission controller outside of the management cluster?
/cc @dougbtv

muraee:
@dougbtv could you review this please

muraee:
@danwinship right, it might be better to move it to the guest cluster, but for now we would like this merged ASAP so we can merge the new changes and not break existing clusters.

dougbtv:
yeah, Dan's right. This should probably move outside the management cluster, it makes sense to me. I filed a backlog for it, and in the name of making sure this webhook isn't deleted, let's move forward with this for now.

muraee:
@dougbtv @danwinship the `e2e-vsphere-ovn-windows` is failing consistently, I can't see how adding a label would do that. can we override the failing tests?

danwinship:
Yup, job history shows it's failing for everyone
/override ci/prow/e2e-vsphere-ovn-windows

PR #1880, opened 2023-07-10
OCPBUGS-15977: Add logic to pick the openshift-sdn shims from right directories

abhat:
Add logic to pick the sdn shim binaries from the correct source directories based on host os release. This change is required so that we pick the right binary based on the base OS version to not fail glibc linkage for FIPS.


abhat:
/override ci/prow/e2e-gcp-ovn. 

this change is unrelated to OVN. overriding the job.

PR #1881, opened 2023-07-10
Enable APB on IC

tssurya:
quay.io/itssurya/dev-images:ic-cno-hack-v4 
/cc @npinaeva


zshi-redhat:
/close

https://redhat-internal.slack.com/archives/GQ0CU2623/p1692090217665869

PR #1882, opened 2023-07-10
[WIP] Ricky pr1838 plus UT fix

flavio-fernandes:
This is a WIP PR that fixes the UT issues we currently have in Ricky PR.

Ref:  https://github.com/openshift/cluster-network-operator/pull/1838


flavio-fernandes:
/close

PR #1886, opened 2023-07-12
[wip] Ricky pr 1838 metrics for controller

flavio-fernandes:
to be tested
self-hosted: adding service port for the controller container in ovnkube


flavio-fernandes:
/close

PR #1890, opened 2023-07-13
Bug 16136: change whereabouts ip reconciler exec

s1061123:
It changes to invoke entrypoint.sh instead of ip-control-loop directly to run corresponding container base OS version binary.


s1061123:
This requires https://github.com/openshift/whereabouts-cni/pull/147 hence blocked for now.

zshi-redhat:
/retitle OCPBUGS-16136: change whereabouts ip reconciler exec

PR #327, opened 2019-09-27
Enable Windows/Linux hybrid clusters 

JacobTanenbaum:
Add env variables to ovn-kubernetes images to enable windows/linux hybrid clusters

PR #1896, opened 2023-07-13
OVN-Kubernetes: Use the status subresource permissions for setting labels and annotations

kyrtapz:
Changes to metadata via status subresources are not restricted for the basic kubernetes types.
This means ovn-kubernetes can set the status/annotations/labels with only the status subresource permissions.

Limit ovn-k permissions.
/cc


kyrtapz:
@JacobTanenbaum is pod delete permission somehow required for hybrid-overlay?
I saw you have added it here: https://github.com/openshift/cluster-network-operator/pull/327/files#diff-f479b1895359064ded1dbab4d640eb9b5b66d9d78e155d90f4658fd47a665552R3

jcaamano:
@kyrtapz is there k8s documentation for this? I guess not but wanted to check if we are tapping on some unintended use case or side effect.

kyrtapz:
> @kyrtapz is there k8s documentation for this? I guess not but wanted to check if we are tapping on some unintended use case or side effect.

I haven't found more than: https://github.com/kubernetes/kubernetes/issues/92022#issuecomment-642680124

npinaeva:
do you think we can add this comment that status permissions also apply to labels and annotations right in the yamls? I bet I will forget about that :)

JacobTanenbaum:
@kyrtapz I don't recall why I added the delete verb. It might have been a misunderstanding on my part about what that meant

kyrtapz:
> do you think we can add this comment that status permissions also apply to labels and annotations right in the yamls? I bet I will forget about that :)

I added some notes in ovn-k, where we will use it:
https://github.com/ovn-org/ovn-kubernetes/pull/3773/files#diff-8844fbe76462e98c9522cc2065c762a3381bf17865f7e720879877adc0a4a464R318
do you think that is sufficient?

kyrtapz:
> @kyrtapz I don't recall why I added the delete verb. It might have been a misunderstanding on my part about what that meant

Gotcha, I think we should remove it then. If we find out it was required we will re-add it with an explanation.

kyrtapz:
/retest-required
https://redhat-internal.slack.com/archives/CDCP2LA9L/p1690471015267919?thread_ts=1688058648.145139&cid=CDCP2LA9L

kyrtapz:
@jcaamano ptal

PR #1903, opened 2023-07-17
OCPBUGS-15918: Skip rendering 0.0.0.0/0 for cluster proxy status

pperiyasamy:
The CNO uses machineCIDR and machineNetwork CIDR parameters in install config of the cluster config map object to render proxy/cluster status. In some cases, machineCIDR contains 0.0.0.0/0 which makes CNO to render noProxy with 0.0.0.0/0 in proxy/cluster status. This is invalid config and leads to installation fails at times. This PR ensures 0.0.0.0/0 is not rendered into proxy/cluster status.

### DIFF
--- a/pkg/util/proxyconfig/no_proxy.go
+++ b/pkg/util/proxyconfig/no_proxy.go
@@ -14,6 +14,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+const defaultCIDR = "0.0.0.0/0"
+
 // MergeUserSystemNoProxy merges user supplied noProxy settings from proxy
 // with cluster-wide noProxy settings. It returns a merged, comma-separated
 // string of noProxy settings. If no user supplied noProxy settings are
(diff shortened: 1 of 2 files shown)


flavio-fernandes (commented):
/lgtm

kyrtapz:
@pperiyasamy could you help me understand what is wrong with `0.0.0.0/0` being set in the `noProxy`?
I understand that it doesn't make much sense to do that but if that is what was specified I don't see how it is invalid.

pperiyasamy:
> @pperiyasamy could you help me understand what is wrong with `0.0.0.0/0` being set in the `noProxy`? I understand that it doesn't make much sense to do that but if that is what was specified I don't see how it is invalid.

AFAIK MCO reacts to `0.0.0.0/0` and sets this to `NO_PROXY` in `/etc/mco/proxy.env`  which would trigger all traffic to bypass the Proxy which is not sensible or feasible. This also makes OCP installation being stuck and will eventually fail as the MachineConfigPool is failing to rollout. @cdoern I think you can help us to understand the issue better.

PR #1907, opened 2023-07-18
[release-4.9] OCPBUGS-16312: pick the ovn-k8s-cni-overlay and openshift-sdn shims from right directories 

jcaamano:


dcbw:
Have to wait for this one until https://github.com/openshift/sdn/pull/560 and https://github.com/openshift/ovn-kubernetes/pull/1759 merge

PR #1916, opened 2023-07-24
[release-4.8] OCPBUGS-16685: pick the ovn-k8s-cni-overlay and openshift-sdn shims from right directories

jcaamano:


jcaamano:
e2e-vsphere jobs seem to be stuck on permanent platform issues. The change here should be unrelated and not sign of issues in other jobs.

/override ci/prow/e2e-vsphere-ovn 
/override ci/prow/e2e-vsphere-windows

PR #1918, opened 2023-07-24
[release-4.6] OCPBUGS-16687: pick the ovn-k8s-cni-overlay and openshift-sdn shims from right directories

jcaamano:


jcaamano:
gather-extra pull fails but tests pass, there should be no relationship with the changes on this PR with that problem specific to the ipv6 job

/override  ci/prow/e2e-metal-ipi-ovn-ipv6

PR #1931, opened 2023-08-02
make MAXLOGFILES a real variable

jluhrsen:
missed the $ symbol to make it a variable instead of just the dumb string "MAXLOGFILES" when trying to do -gt comparison


kyrtapz:
/lgtm
@jluhrsen were you able to verify the rotation works for both self-hosted and managed?

jluhrsen:
> /lgtm @jluhrsen were you able to verify the rotation works for both self-hosted and managed?

I only personally tested in self-hosted. the scripts are the same. I was looking for a way to de-dup that kind of
stuff, but didn't figure it out yet.

PR #1934, opened 2023-08-07
Remove certificatesigningrequests/update permission from ovnkubenode

kyrtapz:
ovnkube-node creates and reads `certificatesigningrequests`, there is no need for it to have the update permission.
Upstream docs: https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#authorization

/cc @jcaamano

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -152,7 +152,6 @@ rules:
     - create
     - get
     - list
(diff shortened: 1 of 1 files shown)


flavio-fernandes (commented):
/lgtm

kyrtapz:
rebased
`bindata/network/ovn-kubernetes/common/002-rbac.yaml` changed to `bindata/network/ovn-kubernetes/common/002-rbac-node.yaml`

PR #1904, opened 2023-07-17
OCPBUGS-16019: prevent creation of multiple cni-sysctl-allowlist-ds pods

mlguerrero12:
The daemonset has a BestEffort QoS which  means that in some cases, the pods won't ever be scheduled.
This also prevents unwanted retries when one or more pods are not ready due to issues with the cluster.
https://issues.redhat.com/browse/OCPBUGS-15818 (BestEffort Daemonset cni-sysctl-allowlist-ds)

When there are issues with the cluster such as in https://issues.redhat.com/browse/OCPBUGS-15284, some pods are stuck in the terminating state which affects subsequent reconciliations of the allowlist controller.

PR #1936, opened 2023-08-08
OCPBUGS-17457: prevent creation of multiple cni-sysctl-allowlist-ds pods

mlguerrero12:
Manual cherry-pick of https://github.com/openshift/cluster-network-operator/pull/1904


mlguerrero12:
@aravindhp, can we also override the windows job here in this backport?

PR #1938, opened 2023-08-08
separate libovsdblogs from main ovnkube-master logs

flavio-fernandes:
The intent of this PR is to split out the logs for libovsdb client out of the ovnkube-node logs,
so must-gather collection has a lot less data to deal with.

Modifications to must-gather gather-netwotk-logs, as well as sos-report must be in place in
order to ensure these new log files are not left behind.

**TODO**:
~~- make sure we are happy with the name of the logs chosen~~
~~- consider using non defaults (i.e. make explicit) values for aging, number of backups and max-size. Their defaults are kept here:   https://github.com/openshift/ovn-kubernetes/blob/f05d79a6080efe659ad6445ccd2a5636cfabf172/go-controller/pkg/config/config.go#L78-L80~~
~~- what if pod/container restart? Maybe logs need to be in a volume-mount? Maybe co-locate it with the dbs?~~

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/common/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/managed/common/004-config.yaml
@@ -136,6 +136,12 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+
+    [logging]
+    libovsdblogfile=/var/log/ovnkube/libovsdb.log
+    logfile-maxsize=100
+    logfile-maxbackups=5
+    logfile-maxage=0
 {{- if .OVNHybridOverlayEnable }}
 
     [hybridoverlay]
--- a/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
@@ -937,6 +937,8 @@ spec:
           name: host-var-lib-cni-networks-ovn-kubernetes
         - mountPath: /run/openvswitch
           name: run-openvswitch
+        - mountPath: /var/log/ovnkube/
+          name: etc-openvswitch
         - mountPath: /run/ovn/
           name: run-ovn
         - mountPath: /etc/openvswitch
--- a/bindata/network/ovn-kubernetes/self-hosted/common/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/common/004-config.yaml
@@ -61,6 +61,12 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+
+    [logging]
+    libovsdblogfile=/var/log/ovnkube/libovsdb.log
+    logfile-maxsize=100
+    logfile-maxbackups=5
+    logfile-maxage=0
 {{- if .OVNHybridOverlayEnable }}
 
     [hybridoverlay]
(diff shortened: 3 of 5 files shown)


bindata/network/ovn-kubernetes/managed/common/004-config.yaml
```diff
@@ -127,6 +127,9 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+
+    [logging]
+    libovsdblogfile=/var/log/ovnkube/libovsdb.log
```
flavio-fernandes, line 132: @kyrtapz do you have an opinion on how to mimic the size and rotation of logs that kubectl uses? We can specify them here using the following flags (https://github.com/openshift/ovn-kubernetes/blob/f05d79a6080efe659ad6445ccd2a5636cfabf172/go-controller/pkg/config/config.go#L78-L80):
- logfile-maxsize
- logfile-maxbackups
- logfile-maxage

flavio-fernandes, line 132: what of pod/container restart? Maybe logs need to be in a volume-mount? Maybe co-locate it with the dbs?

flavio-fernandes: As we discussed in meeting, let's disable max-age and use the other default values.

https://github.com/natefinch/lumberjack/blob/4cb27fcfbb0f35cb48c542c5ea80b7c1d18933d0/lumberjack.go#L89-L94
```
	// MaxAge is the maximum number of days to retain old log files based on the
	// timestamp encoded in their filename.  Note that a day is defined as 24
	// hours and may not exactly correspond to calendar days due to daylight
	// savings, leap seconds, etc. The default is not to remove old log files
	// based on age.
	MaxAge int `json:"maxage" yaml:"maxage"`
```

flavio-fernandes:
/assign @kyrtapz    
Finally fixed UTs !  PTAL, Pat

bindata/network/ovn-kubernetes/managed/common/004-config.yaml
```diff
@@ -130,6 +130,10 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+
+    [logging]
```
kyrtapz, line 134: NIT: IIUC the default MaxSize is set to 100 megabytes, can we set that explicitly here so it is clear for us when reading the config itself?

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -819,6 +819,8 @@ spec:
           name: var-lib-openvswitch
         - mountPath: /run/openvswitch/
           name: run-openvswitch
+        - mountPath: /var/log/ovnkube/
```
kyrtapz, line 822: Note: you are using etc-openvswitch which is mounted to a different host-path in IC and non-IC, I think that should be fine though.

flavio-fernandes: right. This is only used in ovn-ic cases, so I hope that will not be too confusing.

bindata/network/ovn-kubernetes/self-hosted/common/004-config.yaml
```diff
@@ -61,6 +61,12 @@ data:
 {{- if (index . "V6JoinSubnet") }}
     v6-join-subnet="{{.V6JoinSubnet}}"
 {{- end }}
+
+    [logging]
+    libovsdblogfile=/var/log/ovnkube/libovsdb.log
+    logfile-maxsize=100
+    logfile-maxbackups=5
+    logfile-maxage=0
```
jcaamano, line 69: @flavio-fernandes  It looks like this configuration will also affect the CNI logging. Just making sure we are aware and that's ok

https://github.com/ovn-org/ovn-kubernetes/blob/8055a47068e0689d339e6b8a66922fe809e2afe9/go-controller/pkg/config/cni.go#L32-L34

flavio-fernandes: yup, it is ok, iirc.

PR #1941, opened 2023-08-09
Skip rendering 0.0.0.0/0 for cluster proxy status

jluhrsen:
The CNO uses machineCIDR and machineNetwork CIDR parameters in install config of the cluster config map object to render proxy/cluster status. In some cases, machineCIDR contains 0.0.0.0/0 which makes CNO to render noProxy with 0.0.0.0/0 in proxy/cluster status. This is invalid config and leads to installation fails at times.
This commit ensures 0.0.0.0/0 is not rendered into proxy/cluster status.


jluhrsen:
no longer needed

PR #1946, opened 2023-08-10
[Do not merge] Whereabouts, set cron expression to every 15 minutes

dougbtv:
For a patched build, do not merge.

Sets cron expression for the "audit-style scheduled process" for the whereabouts reconciliation to run every 15 minutes.


wwcd:
Can the scheduling time be configurable? 15 minutes is still too long for CT applications, because the addresses provided by operators are limited.

PR #1952, opened 2023-08-17
OCPBUGS-16051, OCPBUGS-3176: Enables IP Forwarding config in CNO

trozet:
CNO will detect an upgrade scenario from 4.13->4.14. In this case, IP Forwarding is set in the API to Global mode. This is desired to not break users during upgrade who have features enabled that rely on this behavior (like Metal LB). In 4.13 IP Forwarding is set via MCO. During upgrade, CNO upgrades first and will detect the upgrade, and enable global IP forwarding. Upon node reboot during upgrade, MCO will remove the forwarding files, but the ovnkube-node container will re-enable forwarding at start up.

For non 4.13->4.14 upgrade scenarios, whatever the setting in the API is will be used (by default this is restricted forwarding).

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -344,7 +344,7 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	// Generate the objects
-	objs, progressing, err := network.Render(&operConfig.Spec, bootstrapResult, ManifestPath, r.client, r.featureGates)
+	objs, progressing, err := network.Render(&newOperConfig.Spec, bootstrapResult, ManifestPath, r.client, r.featureGates)
 	if err != nil {
 		log.Printf("Failed to render: %v", err)
 		r.status.SetDegraded(statusmanager.OperatorConfig, "RenderError",
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -337,6 +337,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVN_GATEWAY_MODE"] = OVN_SHARED_GW_MODE
 	}
 
+	data.Data["IP_FORWARDING_MODE"] = operv1.IPForwardingRestricted
+	if c.GatewayConfig != nil {
+		data.Data["IP_FORWARDING_MODE"] = c.GatewayConfig.IPForwarding
+	}
+
 	// leverage feature gates
 	data.Data["OVN_ADMIN_NETWORK_POLICY_ENABLE"] = featureGates.Enabled(configv1.FeatureGateAdminNetworkPolicy)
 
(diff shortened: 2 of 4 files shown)


trozet:
/hold

until I can verify its working as expected

trozet:
/hold cancel

looks like its working @tssurya PTAL

/assign @tssurya

pkg/network/ovn_kubernetes.go
```diff
@@ -1544,6 +1549,16 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
 		ipsecStatus.Version = ipsecDaemonSet.GetAnnotations()["release.openshift.io/version"]
 	}
 
+	// If we are upgrading from 4.13 -> 4.14 set new API for IP Forwarding mode to Global.
+	// This is to ensure backwards compatibility.
+	if masterStatus != nil {
```
tssurya, line 1554: is this like the IC hack? can be removed in 4.15?

pkg/network/ovn_kubernetes.go
```diff
@@ -1544,6 +1549,16 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
...
+		klog.Infof("4.13 -> 4.14 upgrade detected. Will set IP Forwarding API to Global mode for backwards compatibility")
+		if conf.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig == nil {
+			conf.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig = &operv1.GatewayConfig{}
+		}
+		conf.Spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig.IPForwarding = operv1.IPForwardingGlobal
```
tssurya, line 1559: IIUC, we are basically keeping the IPForwarding to be global for all older clusters?
perhaps I am missing something, but are we really patching the api object to reflect this?
conf here is meant to be read only right? or is CNO actually updating this after you set this?

tssurya: maybe https://github.com/openshift/cluster-network-operator/blob/e125ef95ef43805571c8f2b275a1a1c982932d74/pkg/network/ovn_kubernetes.go#L793 is a better place for this? feels like this render will be called always versus the bootstrap logic is a one time thing?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -344,7 +344,7 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	}
 
 	// Generate the objects
-	objs, progressing, err := network.Render(&operConfig.Spec, bootstrapResult, ManifestPath, r.client, r.featureGates)
+	objs, progressing, err := network.Render(&newOperConfig.Spec, bootstrapResult, ManifestPath, r.client, r.featureGates)
```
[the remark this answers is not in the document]
trozet: yeah I looked into it, it never really mattered before because Alex updated the bootstrapResult and relied on that to sync the new change rather than the operConfig. But I need to use it now, and I dont think it should hurt anything.

pkg/network/ovn_kubernetes.go
trozet: yeah, once we past 4.14 the api will exist, so we will remove this. In that case a user will be upgrading from 4.14 -> 4.15 and already have their API set.

pkg/network/ovn_kubernetes.go
trozet: Yeah the agreement we came up with @knobunc and @cgoncalves was that for clusters upgrading we would enable global forwarding. This would preserve their existing functionality in case they rely on metal lb or some other feature that requires forwarding. For fresh 4.14 installs we will keep the default as disabled.

re: "is a better place for this? feels like this render will be called always versus the bootstrap logic is a one time thing?" and "or is CNO actually updating this after you set this?"

It works because after the bootstrap we do:

```
	// Bootstrap any resources
	bootstrapResult, err := network.Bootstrap(newOperConfig, r.client)
	if err != nil {
		log.Printf("Failed to reconcile platform networking resources: %v", err)
		r.status.SetDegraded(statusmanager.OperatorConfig, "BootstrapError",
			fmt.Sprintf("Internal error while reconciling platform networking resources: %v", err))
	}

	if !reflect.DeepEqual(operConfig, newOperConfig) {
		if err := r.UpdateOperConfig(ctx, newOperConfig); err != nil {
			log.Printf("Failed to update the operator configuration: %v", err)
			r.status.SetDegraded(statusmanager.OperatorConfig, "UpdateOperatorConfig",
				fmt.Sprintf("Internal error while updating operator configuration: %v", err))
		}
	}
```

Where we update the API config if it changed. The modification to the newOperConfig is done in Bootstrap function, should I be doing it somewhere else? Either way this should only happen one time if the code is executed more than once.

tssurya: ack no i won't push for a different place since its just a one time thing...lgtm

pkg/controller/operconfig/operconfig_controller.go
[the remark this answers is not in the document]
dulek: This change caused https://issues.redhat.com/browse/OCPBUGS-18517. And yeah, it might not be clever to inject values in Render phase. Please take a look at #1988 and the approach tried there.

PR #1953, opened 2023-08-17
OCPBUGS-18175: Fix bond-cni's default directory in multus manifest

s1061123:


zshi-redhat:
@s1061123  looks like the PR has the jira/valid-reference label, but not the jira/valid-bug label, maybe we need to move the NP-796 bug under OCPBUG jira project?

zshi-redhat:
/retitle OCPBUGS-18175: Fix bond-cni's default directory in multus manifest

PR #1954, opened 2023-08-17
OCPBUGS-17916: Fix IC configmap lookup in pod_status.go

ricky-rav:
During an upgrade from non-IC to IC, the CNO status logic looks up a well-known configmap that indicates whether the an upgrade to IC is ongoing in order not to report the new operator version (4.14) until the second and final phase of the IC upgrade is done.

The following corrections are needed:

 - CNO shouldn't report a new version if IC configmap can't be retrieved for whatever reason, as suggested in a review to the CNO PR that enabled IC support: https://github.com/openshift/cluster-network-operator/pull/1874#pullrequestreview-1560616992
- looking up the key "ongoing-upgrade" inside the IC configmap is enough; checking for its value to be "true" is not correct, since after code reviews it was decided not to set the value to "true", but to set it to an empty string;
- (optimization) CNO shouldn't look up the IC configmap if the cluster is not running ovn kubernetes

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -253,18 +253,20 @@ func (status *StatusManager) SetFromPods() {
 		}
 	}
 
-	// hack for 2-phase upgrade from non-IC to IC ovnk:
-	// don't update the version field until phase 2 is over
-	if icConfigMap, err := util.GetInterConnectConfigMap(status.client.ClientFor("").Kubernetes()); err == nil {
-		// When an upgrade from <= 4.13 is ongoing, the IC configmap exists and exhibits ongoing-upgrade=''.
-		// When multizone control-plane and node have been rolled out (end of phase 2), the configmap is deleted.
-		if _, ok := icConfigMap.Data["ongoing-upgrade"]; ok {
+	if status.isOVNKubernetes != nil && *status.isOVNKubernetes {
+		// hack for 2-phase upgrade from non-IC to IC ovnk:
+		// don't update the version field until phase 2 is over
+		if icConfigMap, err := util.GetInterConnectConfigMap(status.client.ClientFor("").Kubernetes()); err == nil {
+			// When an upgrade from <= 4.13 is ongoing, the IC configmap exists and exhibits ongoing-upgrade=''.
+			// When multizone control-plane and node have been rolled out (end of phase 2), the configmap is deleted.
+			if _, ok := icConfigMap.Data["ongoing-upgrade"]; ok {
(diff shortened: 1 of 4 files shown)


pkg/controller/statusmanager/pod_status.go
```diff
@@ -258,7 +258,7 @@ func (status *StatusManager) SetFromPods() {
 	if icConfigMap, err := util.GetInterConnectConfigMap(status.client.ClientFor("").Kubernetes()); err == nil {
 		// When an upgrade from <= 4.13 is ongoing, the IC configmap exists and exhibits ongoing-upgrade=true.
 		// When multizone control-plane and node have been rolled out (end of phase 2), the configmap is deleted.
-		if ongoingUpgrade, ok := icConfigMap.Data["ongoing-upgrade"]; ok && ongoingUpgrade == "true" {
+		if _, ok := icConfigMap.Data["ongoing-upgrade"]; ok {
```
dcbw, line 261: @ricky-rav the code no longer agrees with the comment...

ricky-rav: Ops, thanks!
While I'm at it, I'll add a couple of more fixes. I'll add the WIP tag to the title and will remove it once I've run all the necessary upgrades :)

ricky-rav:

This fix allows CNO to report the new version it has upgraded to only at the end of the 2-phase upgrade.

jcaamano (approved):
@kyrtapz PTAL for lgtm

jcaamano:
/payload periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade

pkg/controller/statusmanager/pod_status.go
```diff
@@ -419,3 +423,14 @@ func (status *StatusManager) listAllStatusObjects() (dss []*appsv1.DaemonSet, de
 	}
 	return
 }
+
+func (status *StatusManager) isClusterRunningOVNKubernetes() bool {
+	daemonSets, _, _ := status.listAllStatusObjects()
+	for _, ds := range daemonSets {
+		dsName := NewClusteredName(ds)
```
kyrtapz, line 430: NIT: Is there a point creating the `dsName`? why don't you get the name straight from `ds`

jcaamano:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade 10

jcaamano:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade 10

PR #1965, opened 2023-08-23
OCPBUGS-17391: do not run ovn-ctl stop... in preStop hook

jluhrsen:
JIRA: https://issues.redhat.com/browse/OCPBUGS-17391

it's killing the container and due to recent changes somewhere in openshift (possibly these cri-o stop handling changes [0]), deleting the ovnkube pods will see them stuck in Terminating state with the containers all Exited but kubelet showing them as Running. Eventually (~3m) the pod will finally restart, but if we let cri-o sigterm kill the containers instead the pod will restart normally right away.

[0] https://github.com/cri-o/cri-o/pull/7168


jluhrsen:
[PR](https://github.com/openshift/cluster-network-operator/pull/1978) to do this in master is posted now.

PR #1970, opened 2023-08-24
use sleep instead of wait

jluhrsen:
wait will end up killing the parent PID after the current process ends and that will be PID 1 which exits the container. This happens when the prestop hook runs the ovn-ctl stop command. Because of this and new changes in cri-o the container is in exit state, but kubelet believes it's still running.

this allows the prestop hook to execute all it's steps and it's intention is not to stop the container anyway.

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
@@ -160,7 +160,7 @@ spec:
           trap quit TERM INT
 
           echo "$(date -Iseconds) - starting ovn-northd"
-          exec ovn-northd \
+          nohup ovn-northd \
             --no-chdir "-vconsole:${OVN_LOG_LEVEL}" -vfile:off "-vPATTERN:console:{{.OVN_LOG_PATTERN_CONSOLE}}" \
             --ovnnb-db "{{.OVN_NB_DB_LIST}}" \
             --ovnsb-db "{{.OVN_SB_DB_LIST}}" \
@@ -170,7 +170,7 @@ spec:
             -c /ovn-cert/tls.crt \
             -C /ovn-ca/ca-bundle.crt &
 
-          wait $!
+          while true; do sleep 10; done
         lifecycle:
           preStop:
             exec:
@@ -321,14 +321,14 @@ spec:
             if ${cluster_found}; then
               echo "Cluster already exists for DB: ${db}"
               # join existing cluster
-              exec /usr/share/ovn/scripts/ovn-ctl ${OVN_ARGS} \
+              nohup /usr/share/ovn/scripts/ovn-ctl ${OVN_ARGS} \
               --db-nb-cluster-remote-port={{.OVN_NB_RAFT_PORT}} \
               --db-nb-cluster-remote-addr=${init_ip} \
               --db-nb-cluster-remote-proto=ssl \
               --ovn-nb-log="-vconsole:${OVN_LOG_LEVEL} -vfile:off -vPATTERN:console:{{.OVN_LOG_PATTERN_CONSOLE}}" \
               run_nb_ovsdb &
 
-              wait $!
+              while true; do sleep 10; done
             else
               # either we need to initialize a new cluster or wait for master to create it
               if [[ "${pod_dns_name}" == "${CLUSTER_INITIATOR_IP}" ]]; then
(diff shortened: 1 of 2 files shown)


jluhrsen:
> > Why is this against `release-4.13` instead of `master`? How does this relate to #1966?
> 
> Because master is more complicated now with the IC work. I just wanted a fast way to test different approaches. Once we find something that works (this does), then I'll do the work on master and get that merged first.
> 
> > Why don't we just run the job without `&` and then not need anything clever afterward?
> 
> will that work? if we run `ovn-ctl stop...` wont it kill the process which is a child of the main bash PID 1 process? that's the problem i'm trying to avoid.
> 
> I'll try it though and see. will push a new PR.

no, it does not work:

```
sh-5.1# crictl ps -a | egrep sbdb
f60e49b46c3d9       b5bac72450ab9c1c2b9c509ea8f5ab5c007af7607fe96f3e926a0672eb1abca1                                                                  16 minutes ago       Running             sbdb                                          0                   996311fd71643       ovnkube-master-2ctcx
sh-5.1# crictl exec -it f60e49b46c3d9 sh
sh-5.1# ps -elf
F S UID          PID    PPID  C PRI  NI ADDR SZ WCHAN  STIME TTY          TIME CMD
4 S root           1       0  0  80   0 -  1313 do_wai 20:34 ?        00:00:00 /usr/bin/sh /usr/share/ovn/scripts/ovn-ctl --db-sb-cluster-local-port=9644 --db-sb-cluster-local-addr=10.0.186.196 --no-monitor --db-sb-cluster-local-proto=ssl --ovn-sb-db-ssl-key=/ovn-cert/tls.key --ovn-sb-db-ssl-cert=/ovn-cert/tls.crt --
4 S root         208       1  0  80   0 - 44764 do_pol 20:34 ?        00:00:02 ovsdb-server -vconsole:info -vfile:off -vPATTERN:console:%D{%Y-%m-%dT%H:%M:%S.###Z}|%05N|%c%T|%p|%m --log-file=/var/log/ovn/ovsdb-server-sb.log --remote=punix:/var/run/ovn/ovnsb_db.sock --pidfile=/var/run/ovn/ovnsb_db.pid --unixctl=/var/ru
4 S root        1316       0  0  80   0 -  1112 do_wai 20:51 pts/0    00:00:00 sh
0 R root        1335    1316  0  80   0 -  1787 -      20:51 pts/0    00:00:00 ps -elf
sh-5.1# /usr/share/ovn/scripts/ovn-ctl stop_sb_ovsdb
Exiting ovnsb_db (208).
FATA[0046] execing command in container: command terminated with exit code 137 
```

running ovn-ctl stop_sb_ovsdb in the sbdb container makes the entire container exit. This is from a cluster built
with [this PR](https://github.com/openshift/cluster-network-operator/pull/1973) which just removes the & and wait.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
```diff
@@ -170,7 +170,7 @@ spec:
             -c /ovn-cert/tls.crt \
             -C /ovn-ca/ca-bundle.crt &
 
-          wait $!
```
flavio-fernandes, line 173: `sleep infinty`   may be easier than doing a loop

flavio-fernandes (commented):
Very interesting find, Jamo!
Please make mentioning of  https://github.com/cri-o/cri-o/pull/7168
as well as https://issues.redhat.com/browse/OCPBUGS-17391 in this PR.
I'm okay with this simplification, as long as we can assure that will be running with the newer crio.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
jluhrsen: I can do that.

dcbw: So what happens if the ovn thing (northd, controller, DBs, whatever) exits for some reason? Won't the sleep loop cover that up and leave the container running, while the process we care about is dead?

The `wait $!` would at least exit the container and let kubelet restart it, right? how does that work here?

jluhrsen:
> Very interesting find, Jamo! Please make mentioning of [cri-o/cri-o#7168](https://github.com/cri-o/cri-o/pull/7168) as well as https://issues.redhat.com/browse/OCPBUGS-17391 in this PR. I'm okay with this simplification, as long as we can assure that will be running with the newer crio.

this is not the right PR at the moment. once we find a solution I will run it here with a 4.13 PR and then if it works I'll do the work on master. It's easier to iterate on 4.13 is why.

bindata/network/ovn-kubernetes/managed/ovnkube-master.yaml
jluhrsen: yes, that's correct. I can see that could be a major problem then.

hmmm....

Do we have to stop the process in the preStop hook? we have this code:

```
          quit() {
            echo "$(date -Iseconds) - stopping sbdb"
            /usr/share/ovn/scripts/ovn-ctl stop_sb_ovsdb
            echo "$(date -Iseconds) - sbdb stopped"
            rm -f /var/run/ovn/ovnsb_db.pid
            exit 0
          }
          # end of quit
          trap quit TERM INT
```
          
that is called on a SIGTERM which is what we get in normal cases with cri-o stopping the containers. Is that
good enough and we can just remove the prestop hooks?

btw, the 'echo' "rm -f" and "exit 0" lines will not exectute since the "ovn-ctl stop..." command kills things all
by itself.

jluhrsen: btw, I have [this PR](https://github.com/openshift/cluster-network-operator/pull/1965/commits/08951fde319539e8d40d540d6cd7097df7ee355b) removing the ovn-ctl stop command from the prestop hook. I know it gets us around the
issue I'm concerned with. what would be the downside of this?

jcaamano:
So it looks to me like we could just remove the preStop hooks. I tried to blame the change to see if I could find out a specific circumstance we need them for and I looped back to @flavio-fernandes  and @dcbw  :stuck_out_tongue: 

https://github.com/openshift/cluster-network-operator/commit/4eb76c54ee3ffd62d8244d3f5bf2c412bb149629
https://bugzilla.redhat.com/show_bug.cgi?id=1943363

Not a lot of specific info there so it looks like it might have been added just in case. They do mention upgrades. So perhaps we need to test upgrading **from** this change to something else and check that goes smoothly

jluhrsen:
this was a bad idea because of Dan's comments:

> So what happens if the ovn thing (northd, controller, DBs, whatever) exits for some reason? Won't the sleep loop cover that up and leave the container running, while the process we care about is dead?
> 
> The wait $! would at least exit the container and let kubelet restart it, right? how does that work here?

PR #1930, opened 2023-08-01
Join ovnkube-controller and ovnkube-node container for multizone setup 

npinaeva:
TOOD: Remove ovn-k image tweak commit when https://github.com/openshift/ovn-kubernetes/pull/1833 is merged

PR #1971, opened 2023-08-25
OCPBUGS-17782, SDN-3664: Join ovnkube-controller and ovnkube-node container for multizone setup

trozet:
Copied from https://github.com/openshift/cluster-network-operator/pull/1930 and removed hack commit for quicker CI signal

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
@@ -698,155 +698,8 @@ spec:
             memory: 300Mi
         terminationMessagePolicy: FallbackToLogsOnError
 
-      - name: kube-rbac-proxy-controller
-        image: {{.KubeRBACProxyImage}}
-        command:
-        - /bin/bash
-        - -c
-        - |
-          #!/bin/bash
-          set -euo pipefail
-          TLS_PK=/etc/pki/tls/metrics-cert/tls.key
-          TLS_CERT=/etc/pki/tls/metrics-cert/tls.crt
-          # As the secret mount is optional we must wait for the files to be present.
-          # The service is created in monitor.yaml and this is created in sdn.yaml.
-          TS=$(date +%s)
-          WARN_TS=$(( ${TS} + $(( 20 * 60)) ))
-          HAS_LOGGED_INFO=0
-
-          log_missing_certs(){
-              CUR_TS=$(date +%s)
-              if [[ "${CUR_TS}" -gt "WARN_TS"  ]]; then
-                echo $(date -Iseconds) WARN: ovn-node-metrics-cert not mounted after 20 minutes.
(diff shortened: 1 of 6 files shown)


npinaeva:
egress ip fix https://github.com/openshift/origin/pull/28219

trozet:
/hold

until https://github.com/openshift/origin/pull/28219 is merged and serial passes

tssurya:
/retitle SDN-3664: Join ovnkube-controller and ovnkube-node container for multizone setup

trozet:
/retitle OCPBUGS-17782, SDN-3664: Join ovnkube-controller and ovnkube-node container for multizone setup

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -1083,7 +946,7 @@ spec:
         resources:
           requests:
             cpu: 10m
-            memory: 300Mi
+            memory: 600Mi
```
tssurya, line 949: this is measured and justified? or its just combination of 300+300 of both containers?

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -973,13 +832,17 @@ spec:
             --metrics-bind-address "127.0.0.1:29103" \
```
tssurya, line 832: so the same container port is going to be used to scrape metrics from both node and controller threads right ? since its same process?

pkg/network/ovn_kubernetes.go
```diff
@@ -2240,7 +2240,7 @@ func isInterConnectEnabledOnDaemonset(ds *appsv1.DaemonSet, containerName string
 }
 
 func isInterConnectEnabledOnNodeDaemonset(ds *appsv1.DaemonSet) bool {
-	return isInterConnectEnabledOnDaemonset(ds, util.OVN_NODE)
+	return isInterConnectEnabledOnDaemonset(ds, util.OVN_NODE) || isInterConnectEnabledOnDaemonset(ds, util.OVN_CONTROLLER)
```
tssurya, line 2243: hmm is this for backwards compatibility we check this for upgrades of sorts? since we don't really have the ovn_node container anymore...

tssurya (approved):
/lgtm
this change was not as big as i thought it would be :D
shall we trigger payloads for this? -> at the very lease the upgrades need to be tested so that we are 100% sure disruptions pass.

pkg/network/ovn_kubernetes.go
ricky-rav: I asked Nadia to add this line. I need to know during upgrades whether interconnect is enabled (`--enable-interconnect`) on the ovnk node pod as well as on the ovnk master / control-plane pods. Checking ovnkube-node container is now valid only for 4.13 and 4.14 single zone, so we need to check the ovnkube-controller container in 4.14 multizone (and future versions).

PR #1975, opened 2023-08-29
OVN-Kubernetes ipsec: Do not try to remove CSRs

kyrtapz:
Follow up to https://github.com/openshift/cluster-network-operator/pull/1928/ in which the ipsec daemonset starting using a randomized name for CSRs. 
It should not try to remove the CSR at the end since it doesn't have the permissions to do so and CSRs are automatically garbage collected anyway.

/cc @jcaamano


trozet:
same fix is in the host ipsec PR as https://github.com/openshift/cluster-network-operator/pull/1849

kyrtapz:
> same fix is in the host ipsec PR as #1849

Thanks, since it merged closing this one.

PR #1978, opened 2023-08-30
OCPBUGS-17391: remove prestop hooks for northd, sbdbd and nbdb

jluhrsen:
JIRA: https://issues.redhat.com/browse/OCPBUGS-17391

when the prestop hook executes the ovn-ctl stop... command this causes the containers main script to exit which also propagates to the main container (PID 1) process which exits causing the container to exit. This is not neccessarily a problem except that with recent changes in CRI-O [0] there is some problem with how kubelet views this case. the containers are Exited for crio but seen as Ready from kubelet's perspective. Because of this, the pod gets stuck in Terminating state until it times out (~3m) and then containers will restart.

removing the prestop hook let's the SIGTERM reach the configured trap (example [1]) run which effectively does the same thing as what the prestop hook would have done. This prevents the pod from being stuck in Terminating state.

[0] https://github.com/cri-o/cri-o/pull/7168
[1] https://github.com/openshift/cluster-network-operator/blob/c55f19132864a9d8fe347ad6a8280eaa2f60e839/bindata/network/ovn-kubernetes/self-hosted/single-zone-interconnect/ovnkube-master.yaml#L74-L82

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
@@ -338,13 +338,6 @@ spec:
             --n-threads={{.NorthdThreads}} &
 
           wait $!
-        lifecycle:
-          preStop:
-            exec:
-              command:
-                - /bin/bash
-                - -c
-                - OVN_MANAGE_OVSDB=no /usr/share/ovn/scripts/ovn-ctl stop_northd
         env:
         - name: OVN_LOG_LEVEL
           value: info
@@ -511,16 +504,6 @@ spec:
                 if ! retry 20 "northd-backoff" "${OVN_NB_CTL} set nb_global . options:northd-backoff-interval-ms={{.OVN_NORTHD_BACKOFF_MS}}"; then
                   exit 1
                 fi
-          preStop:
-            exec:
-              command:
-                - /bin/bash
-                - -c
-                - |
-                  echo "$(date -Iseconds) - stopping nbdb"
-                  /usr/share/ovn/scripts/ovn-ctl stop_nb_ovsdb
-                  echo "$(date -Iseconds) - nbdb stopped"
-                  rm -f /var/run/ovn/ovnnb_db.pid
         readinessProbe:
           initialDelaySeconds: 90
           timeoutSeconds: 5
@@ -654,16 +637,6 @@ spec:
                 if ! retry 20 "ipsec" "${OVN_SB_CTL} get sb_global . ipsec"; then
                   exit 1
                 fi
-          preStop:
-            exec:
-              command:
-                - /bin/bash
-                - -c
-                - |
-                  echo "$(date -Iseconds) - stopping sbdb"
-                  /usr/share/ovn/scripts/ovn-ctl stop_sb_ovsdb
-                  echo "$(date -Iseconds) - sbdb stopped"
-                  rm -f /var/run/ovn/ovnsb_db.pid
         readinessProbe:
           initialDelaySeconds: 90
           timeoutSeconds: 5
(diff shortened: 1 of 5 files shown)


flavio-fernandes (commented):
nice. less is more!
/lgtm
^^
pending upgrade jobs pass w/out problems

jcaamano:
@jluhrsen 

The first reference I find of the prestop hooks being added is
https://github.com/openshift/cluster-network-operator/commit/f0d409748f66f91adb9b029509dfe94e8c1ced74

Those were just for NBDB and SBDB and at that time those were not running in the background with a signal handler on top, and there is no specific mention on why the latter wouldn't work. So I would say it could have been a viable alternative at the time.

The second is 
https://github.com/openshift/cluster-network-operator/commit/4eb76c54ee3ffd62d8244d3f5bf2c412bb149629
from bug 
https://bugzilla.redhat.com/show_bug.cgi?id=1943363
There is no specific mention there also on why would both things be needed. But there is a reproducer, which is core dumps generated on upgrades
https://bugzilla.redhat.com/show_bug.cgi?id=1943363#c13

So perhaps we could try an upgrade between two versions with this fix and see if we get those coredumps. But not only on 4.14 branch as raft might have something to do with it. I would try 4.13 as well.

In regards to the cri-o, is difficult to track how it changed behavior. Things I would check for is how much time does it take for the northd/nbdb/sbdb processes to exit (might go over the timeout and might be killed instead), and for crio error messages in the journal that could give us a clue what is actually going on. I could spot differences on how some unexpected scenarios would be handled but they look unlikely to happen.

I don't think we have many choices though. So we might as well remove the prestop hooks which shouldn't really be used for what we are using them and see what happens; than try any other hack that we don't equally know how well is going to work.

@jluhrsen let me know if you want to check any additional thing, specially upgrading on top of this change on 4.13, otherwise I agree to LGTM this.

jluhrsen:
> @jluhrsen let me know if you want to check any additional thing, specially upgrading on top of this change on 4.13, otherwise I agree to LGTM this.

@jcaamano , I manually brought up a 4.14 cluster without the prestop hooks and upgraded it to a release based off of this PR. no coredumps and cluster upgrade completed successfully. I had meant to start with a 4.13 cluster though, so I will work on that next. but we have this data point now.

jcaamano:
> BTW, @kyrtapz mentioned to me that on node reboot, its systemd crio service termination which sends a SIGTERM to all processes running on all containers, so presumably in that scenario the bash script would get a TERM signal, and the DBs/north would get another TERM signal of its own. This means that potentially we would not have a graceful shutdown on reboot but I don't know how the prestop hooks would help there anyway. I think you tried this out as well and did not see anything weird?

In regards to this found this other bug
https://bugzilla.redhat.com/show_bug.cgi?id=1944239

That bug seems to imply that appctl exit is preferred over sigterm but unsure about the extent of the concern. I have started a slack thread and tagged interested people.
However, as I mentioned, don't see how the prestop hooks or this PR have anything to do with it, so just for pure knowledge.

jcaamano:
@jluhrsen @kyrtapz I just learned about a k8s feature called graceful node shutdown that should allow a proper lifecycle shutdown of pods on node reboot, not sure if openshift makes use of it
https://kubernetes.io/blog/2021/04/21/graceful-node-shutdown-beta/

jluhrsen:
> @jluhrsen I think we need to test the upgrade between two releases both of them containing this fix, so no pre-stop hooks on either, which is what I think you did, just making sure. Yeah, 4.13 is a different valuable data point as well as that has RAFT.+

I was finally able to test this with a 4.13 release based on [this PR](https://github.com/openshift/cluster-network-operator/pull/2001) which removes the prestop hooks and a release based on this PR which also has the hooks removed. I did not find any core dumps and the upgrade was successful.

jcaamano:
I reproduced this looking out for crio errors that would hint how its new code is triggering the issue but did not find any.

I think that enough time was spent ensuring this does not bring back any of the issues for which the prestop hooks were introduced. 

As mentioned before, they should not cover node reboots.

PR #1979, opened 2023-08-31
OCPBUGS-18363: Add '/etc/cni/multus/net.d' into volumemount in multus pod

s1061123:


s1061123:
/title OCPBUGS-18363: Add '/etc/cni/multus/net.d' into volumemount in multus pod

s1061123:
/retitle OCPBUGS-18363: Add '/etc/cni/multus/net.d' into volumemount in multus pod

PR #1981, opened 2023-08-31
OCPBUGS-15201: Disable weak SSH cipher suites

flavio-fernandes:
During an internal audit conducted by one of our customers in OCP, it was discovered that the openshift-sdn component, specifically on port 9101, has exposed certain cipher suites that are considered weak. We have identified the following weak ciphers that are currently being accepted:

[TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256](https://ciphersuite.info/cs/TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256/)
[TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256](https://ciphersuite.info/cs/TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256/)

Reported-at: https://issues.redhat.com/browse/OCPBUGS-15201

### DIFF
--- a/bindata/kube-proxy/kube-proxy.yaml
+++ b/bindata/kube-proxy/kube-proxy.yaml
@@ -142,7 +142,7 @@ spec:
           exec /usr/bin/kube-rbac-proxy \
             --logtostderr \
             --secure-listen-address=:{{.MetricsPort}} \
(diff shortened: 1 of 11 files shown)


flavio-fernandes:
/assign @danwinship 

Hi Dan, can you PTAL and help me in finding out if there are some ciphers we could be adding as well as removing the 2 weak ones?

danwinship:
port 9101 is the kube-rbac-proxy for metrics.
while it looks like we are explicitly requesting certain cipher suites in `bindata/network/openshift-sdn/sdn.yaml`, this seems kind of weird. Lots of OCP components are doing the same kube-rbac-proxy-for-metrics hack. We ought to make it so it does the right thing automatically.

There is no reason to add additional ciphers.

flavio-fernandes:
Another related issue, fixed recently by Jamo:  https://issues.redhat.com/browse/OCPBUGS-3267

PR #1985, opened 2023-09-05
OCPBUGS-14819: HyperShift: Use the local konnectivity proxy when checking proxy readiness

kyrtapz:
When CNO checks proxy readiness it should send the requests through the hosted cluster.

Depends on HyperShift changes here: https://github.com/openshift/hypershift/pull/2974, which merged already.

### DIFF
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -11,13 +11,14 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/types"
-
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/cluster-network-operator/pkg/platform"
 	"github.com/openshift/cluster-network-operator/pkg/util/validation"
+	netproxy "golang.org/x/net/proxy"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


vendor/modules.txt
```diff
@@ -588,7 +588,9 @@ golang.org/x/net/http/httpproxy
 golang.org/x/net/http2
 golang.org/x/net/http2/hpack
 golang.org/x/net/idna
+golang.org/x/net/internal/socks
```
flavio-fernandes, line 591: silly ask: no trying to reflect this chance in go.mod+go.sum files as well?

pkg/controller/proxyconfig/validation.go
```diff
@@ -231,9 +233,26 @@ func validateReadinessEndpointWithRetries(caBundle []*x509.Certificate, proxy, e
...
+func runReadinessProbe(caBundle []*x509.Certificate, proxyURL, endpoint *url.URL) error {
 	transport := &http.Transport{
-		Proxy: http.ProxyURL(proxy),
+		Proxy: http.ProxyURL(proxyURL),
+	}
+
+	if platform.NewHyperShiftConfig().Enabled {
+		hostedProxyURL, err := url.Parse(platform.HostedClusterLocalProxy)
+		if err != nil {
+			return fmt.Errorf("failed to parse hosted cluster proxy: %v", err)
```
flavio-fernandes, line 244: NIT: wrap error using %w instead of %v

pkg/controller/proxyconfig/validation.go
```diff
@@ -231,9 +233,26 @@ func validateReadinessEndpointWithRetries(caBundle []*x509.Certificate, proxy, e
...
+		}
+		hcpProxy, err := proxy.FromURL(hostedProxyURL, nil)
+		if err != nil {
+			return fmt.Errorf("failed to create hosted cluster proxy: %v", err)
```
flavio-fernandes, line 248: NIT: wrap error using %w instead of %v

pkg/controller/proxyconfig/validation.go
```diff
@@ -231,9 +233,26 @@ func validateReadinessEndpointWithRetries(caBundle []*x509.Certificate, proxy, e
...
 // caBundle is used to authenticate endpoint if endpoint contains an
 // https scheme.
-func runReadinessProbe(caBundle []*x509.Certificate, proxy, endpoint *url.URL) error {
+func runReadinessProbe(caBundle []*x509.Certificate, proxyURL, endpoint *url.URL) error {
 	transport := &http.Transport{
-		Proxy: http.ProxyURL(proxy),
+		Proxy: http.ProxyURL(proxyURL),
+	}
+
+	if platform.NewHyperShiftConfig().Enabled {
```
flavio-fernandes, line 241: it is a shame that `func hyperShiftEnabled() bool` is not public. But I think this work around (of getting the config and looking at Enabled) is fine for now.

pkg/controller/proxyconfig/validation.go
```diff
@@ -14,7 +13,10 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/cluster-network-operator/pkg/platform"
 	"github.com/openshift/cluster-network-operator/pkg/util/validation"
+	"golang.org/x/net/proxy"
```
flavio-fernandes, line 18: Q: would you consider calling this import `netProxy` ? I say that because it is a bit 'scary' when looking at functions that remain using `proxy` as parameters, like: `func validateReadinessEndpoint(caBundle []*x509.Certificate, proxy, endpoint string) error {`

flavio-fernandes (commented):
some minor nits and questions... otherwise
/lgtm

vendor/modules.txt
kyrtapz: We are already pulling in `golang.org/x/net` which I think is why my changes are not introducing changes to go modules.

pkg/controller/proxyconfig/validation.go
kyrtapz: we use the same approach across CNO so I would leave it as is for now.

PR #1988, opened 2023-09-06
OCPBUGS-18517: Kuryr: Set MTU on Bootstrap, not Render phase

dulek:
Since #1952 `network.Render` is no longer supplied with operConfig used later, but a copy of it. This means that MTU set by Kuryr's render phase won't be used for setting the Network.Status and will lead to panic.

We have this issue because for Kuryr default MTU is detected on bootstrap phase and there's no fixed value.

This commit solves this by making "0" a default value for Kuryr's MTU. Having 0 means that Kuryr will attempt to autodetect the MTU. Moreover the MTU on KuryrConfig will now be set on bootstrap phase, so update of the OperConfig done after the bootstrap phase will make sure to update it too.

### DIFF
--- a/pkg/network/kuryr.go
+++ b/pkg/network/kuryr.go
@@ -112,7 +112,6 @@ func renderKuryr(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapR
 
 	// Pods Network MTU
 	data.Data["PodsNetworkMTU"] = b.PodsNetworkMTU
-	c.MTU = &b.PodsNetworkMTU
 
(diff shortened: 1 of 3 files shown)


gryf (approved):
LGTM

dulek:
/hold cancel

The Kuryr jobs still fails the tests and we [link this with upgrade of our testing cloud to OSP 17.1](https://issues.redhat.com/browse/OCPBUGS-17066). This PR however unblocks the job to pass the installation, so it's a significant progress. I'd like to get this merged even with the job failing.

dulek:
/retest-required

Just to be sure if UI isn't hiding some label to me.

PR #1990, opened 2023-09-07
OCPBUGS-18676: ovnkube: set northd backoff-interval and use a single thread to save CPU

dcbw:
northd has an option to sleep for a short amount of time after processing changes from NB/SB that allows it to trade off a bit of latency for a lot of CPU savings. Since events from NB come frequently during scale tests northd doesn't have a lot of time to sleep. Until we have more incremental processing, most of that CPU time is burned just recalculating things that haven't changed, so it's mostly wasted.

Letting northd sleep has been shown in density-light and density-cni 120 node scale tests to have almost no adverse effect on P99 PodReady times, but a huge improvement in CPU utilization.

Upstream equivalent is https://github.com/ovn-org/ovn-kubernetes/pull/3877

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
@@ -506,6 +506,11 @@ spec:
                 if ! retry 20 "ipsec" "${OVN_NB_CTL} set nb_global . ipsec=${ipsec}"; then
                   exit 1
                 fi
+
+                # Tell northd to sleep a bit so it takes less CPU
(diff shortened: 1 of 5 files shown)


dcbw:
From an NBDB container on the cluster:
```
sh-5.1# ovn-nbctl list NB_Global | grep backoff   
options             : {e2e_timestamp="1694097130", ipsec_encapsulation="false", mac_prefix="12:ff:3d", max_tunid="16711680", name=ci-op-qirq0bxi-b3a20-jmtrb-master-0, northd-backoff-interval-ms="300", northd_internal_version="23.06.1-20.27.2-70.6", northd_probe_interval="10000", svc_monitor_mac="b2:37:8d:4b:ce:00"}
```

dcbw:
/retest

AWS `InsufficientInstanceCapacity`

trozet:
Critical fix for perf/scale. From our testing it this cuts northd CPU in half, while having no impact to pod latency.

/lgtm

PR #1999, opened 2023-09-13
OCPBUGS-18892: make ipsec.service required

yuvalk:
so pluto will run on the host by systemd only without the duality of
possibly running it via systemd or from the e-w container

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -190,6 +190,11 @@ spec:
           done
           echo "ovnkube-node has configured node."
 
+          if ! pgrep pluto; then
+            echo "pluto is not running, enable the service and/or check system logs"
+            exit 2
+          fi
+
           # After a restart of this container (or on initial startup), we flush xfrm state and policy
           # before we start pluto and ovs-monitor-ipsec in order to start in a known good state. This
           # will result in a small interruption in traffic until pluto and ovs-monitor-ipsec start again.
@@ -204,8 +209,6 @@ spec:
           /usr/libexec/ipsec/_stackmanager start
           # Check nss database status
           /usr/sbin/ipsec --checknss
-          # Start the pluto IKE daemon
-          /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
 
           # Start ovs-monitor-ipsec which will monitor for changes in the ovs
           # tunnelling configuration (for example addition of a node) and configures
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -132,7 +132,7 @@ spec:
           # Updating the certificates does not need to be an atomic operation as
           # the will get read and loaded into NSS by the ovs-monitor-ipsec process
           # which has not started yet.
-          ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=$cert_pem/g \
+          ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=$cert_pem \
                                                      other_config:private_key=/etc/openvswitch/keys/ipsec-privkey.pem \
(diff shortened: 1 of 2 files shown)


yuvalk:
this PR requires https://github.com/openshift/release/pull/43328 first

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -205,7 +210,8 @@ spec:
           # Check nss database status
           /usr/sbin/ipsec --checknss
           # Start the pluto IKE daemon
-          /usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
+          ## we'll no longer start it ourseleves but rely on prerequisit to enable the systemd service
+          #/usr/libexec/ipsec/pluto --leak-detective --config /etc/ipsec.conf --logfile /var/log/openvswitch/libreswan.log
```
dcbw, line 214: @yuvalk Just remove the command itself, don't comment it out. Kill the "# start the pluto ike daemon" comment too since it'll get started elsewhere.

But really, how does the IKE daemon get started now? How does the systemd unit get enabled now?

yuvalk: will do

dcbw:
> > But really, how does the IKE daemon get started now? How does the systemd unit get enabled now?
> > I thought it might be confusing :-/
> > so for now, we have 2 sets of ipsec DSs
> > ipsec-containerized - used if libreswan is not found on host
> > ipsec-host - used if libreswan is found on host (and with this pr, will also fail if pluto is not running)
> 
> and just as the ipsec os extension is a prerequisite for n-s ipsec, we'll make pluto.service a pre-requisit too. so for now, it will be enabled by the user.
> 
> if it's just e-w, nothing changes. and pluto will still run in the container https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml#L217

@yuvalk OK; but it looks like we'll run *both* ipsec-containerized.yaml and ipsec-host.yaml since both are enabled when `{{if .OVNIPsecDaemonsetEnable}}` is true. Why do we need both of them?

yuvalk:
> @yuvalk OK; but it looks like we'll run _both_ ipsec-containerized.yaml and ipsec-host.yaml since both are enabled when `{{if .OVNIPsecDaemonsetEnable}}` is true. Why do we need both of them?

this been discussed in the original enablement PR.
basically for now (4.14) we dont have a good upgrade solution that will enable us to move from containerized version to host only implementation.
so we decided to keep both.
each of them is guarded by a set of `if`s that verify only one of them is active
e.g from the containerized ds:
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml#L51
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml#L183
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml#L259

cant have one failing because of the statusmanager guarding code that verify all DSs are running properly
https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/statusmanager/pod_status.go#L89


plan for 4.15 is to properly solve the upgrade issue and from then on, keep only the host version.
for that it'll also automatically apply the needed MC (to enable the os extension and the service)

dcbw:
Ok; thanks for the explanation; so back to this PR...  we *need* to make upgrades work. We can't not have them work for people that have already deployed ipsec.

Why can't we check whether pluto is already running using a socket or something, before launching pluto?

yuvalk:
> Ok; thanks for the explanation; so back to this PR... we _need_ to make upgrades work. We can't not have them work for people that have already deployed ipsec.
wait, upgrade absolutely work!

when I said "upgrade issue" I only meant that we are currently running with 2 DSs. in the future we plan to improve the upgrade process so that we can always run with the host based solution for everyone (including all the upgrade scenarios)

trozet (approved):
/lgtm

PR #2000, opened 2023-09-13
ovnkube: simplify northd threading and SNO templating

dcbw:
It's all single thread now, plus we don't need an if/else block for SNO anymore since it's just a single boolean now.


trozet (approved):
/lgtm

trozet:
closing in favor of https://github.com/openshift/cluster-network-operator/pull/2234

/close

PR #2003, opened 2023-09-14
OCPBUGS-19018: use $CPE_NAME to find the OS major version

derekhiggins:
RHEL_VERSION isn't present in scos

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -39,6 +39,7 @@ data:
     # detect which version we're using in order to copy the proper binaries
     case "${ID}" in
       rhcos|scos)
+        RHEL_VERSION=$(echo "${CPE_NAME}" | cut -f 5 -d :)
         rhelmajor=$(echo $RHEL_VERSION | sed -E 's/([0-9]+)\.{1}[0-9]+(\.[0-9]+)?/\1/')
       ;;
       rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .)
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -113,6 +113,7 @@ spec:
           # detect which version we're using in order to copy the proper binaries
           case "${ID}" in
             rhcos|scos)
+              RHEL_VERSION=$(echo "${CPE_NAME}" | cut -f 5 -d :)
               rhelmajor=$(echo $RHEL_VERSION | sed -E 's/([0-9]+)\.{1}[0-9]+(\.[0-9]+)?/\1/')
             ;;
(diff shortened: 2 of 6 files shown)


LorbusChris:
Maybe it'd be a bit clearer to separate the `rhcos` and `scos` cases here, OTOH it probably doesn't matter.

/lgtm

trozet:
@derekhiggins what does scos mean? What is the CPE_NAME variable?

LorbusChris:
SCOS == CentOS **S**tream **C**ore**OS**
CPE == Common Platform Enumeration (i.e. a stable tag)

The CPE_NAME is from `/usr/lib/os-release`, i.e. on SCOS it's `CPE_NAME="cpe:/o:centos:centos:9::coreos"`

I wouldn't mind using the CPE_NAME only on SCOS, and leaving the RHCOS case as-is.

derekhiggins:
> I wouldn't mind using the CPE_NAME only on SCOS, and leaving the RHCOS case as-is.

I can do this, but it would require that I duplicate the sed line to remove the minor version
>        rhelmajor=$(echo $RHEL_VERSION | sed -E 's/([0-9]+)\.{1}[0-9]+(\.[0-9]+)?/\1/')

I know the minor version isn't in this part of the CPE but it was in the past https://access.redhat.com/solutions/6564691

trozet (approved):
/approve

bindata/network/multus/multus.yaml
```diff
@@ -39,6 +39,8 @@ data:
     # detect which version we're using in order to copy the proper binaries
     case "${ID}" in
       rhcos|scos)
+        IFS=':' read -ra parts <<< "$CPE_NAME"
```
LorbusChris, line 42: probably a non-issue but should this be
```suggestion
        local IFS=':' read -ra parts <<< "$CPE_NAME"
```
so the changed IFS doesn't leak?

LorbusChris: Quoting @trozet:

> probably, and should probably be in its own function

Let's address this
/hold

derekhiggins: hmm, I didn't think it would leak because its being set inline, but may IFS is special, either way I could just use
> echo $CPE_NAME | cut -f 4 -d :
like the lines below

LorbusChris: I think you may be right about the inline setting of IFS here.

Nonetheless, I personally find using `cut` easier to understand. Mind changing it?

derekhiggins: Done, I'd have used cut in the first place if I noticed the pattern already being used in the file...

LorbusChris:
/lgtm

I'll leave it to @trozet to lift the hold

LorbusChris:
I've got the green light from Tim
/unhold

PR #2007, opened 2023-09-19
OCPBUGS-19377: Kuryr: Fix deriving MTU from previous config

dulek:
PR https://github.com/openshift/cluster-network-operator/pull/1988 introduced a bug causing upgrades to fail, because even when Kuryr MTU was detected from the previous version of the KuryrConfig, CNO would override it to 0. This commit fixes that.


MaysaMacedo:
@dulek would you mind adding a commit message? Thanks

dulek:
> @dulek would you mind adding a commit message? Thanks

Done.

dulek:
/retest-required

Installation with Kuryr passed, tests failed. This is the current situation due to https://issues.redhat.com/browse/OCPBUGS-18818 which we investigate, but this patch should be good to go.

PR #2009, opened 2023-09-19
OCPBUGS-19625: Multus per-node certificate request

dougbtv:


s1061123:
Hi

- multus serviceaccunt no longer requried in per node cert because multus will query k8s api using new cert that is retrieved by per-node cert
- Current `ClusterRole` should be applied to `system:multus` as follows:

```
---
kind: ClusterRoleBinding
metadata:
  name: multus-group
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: multus
subjects:
  - kind: Group
    name: system:multus
    apiGroup: rbac.authorization.k8s.io
```

bindata/network/multus/multus.yaml
```diff
@@ -243,6 +249,14 @@ spec:
         - name: "NO_PROXY"
           value: "{{ .NO_PROXY}}"
 {{ end }}
+        - name: K8S_NODE
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: MULTUS_NODE_NAME
```
kyrtapz, line 256: CI failure:
```
level=error msg=Cluster operator network Degraded is True with ApplyOperatorConfig: Error while updating operator configuration: could not apply (apps/v1, Kind=DaemonSet) openshift-multus/multus: failed to apply / update (apps/v1, Kind=DaemonSet) openshift-multus/multus: failed to create typed patch object (openshift-multus/multus; apps/v1, Kind=DaemonSet): .spec.template.spec.containers[name="kube-multus"].env: duplicate entries for key [name="MULTUS_NODE_NAME"]
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/2009/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn/1705053800420610048



It might be related to the fact that Dan already added this variable here: https://github.com/openshift/cluster-network-operator/pull/2020

bindata/network/multus/002-rbac.yaml
```diff
@@ -53,26 +54,95 @@ rules:
...
+
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: multus-group
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: multus-proper
```
kyrtapz, line 122: This seems to be enabled unconditionally but we should allow for the fallback option that would disable the per-node cert/webhook.
See here: https://github.com/openshift/cluster-network-operator/pull/1983/files#diff-cd68b4caec2e8a6afb4c120d980469834c62524b90a300f2adf90b6d6bec0c55R49-R58
And: https://github.com/openshift/cluster-network-operator/pull/1983/files#diff-13e60e30d72f7b5c263fabc48ae9533c32531df3983d7e70d8a4e1a0a1c1a97dR25-R28

bindata/network/multus/multus.yaml
```diff
@@ -299,6 +313,9 @@ spec:
             items:
             - key: daemon-config.json
               path: daemon-config.json
+        - name: host-run-multus-certs
+          hostPath:
+            path: /run/multus_certs
```
kyrtapz, line 318: This will be removed if a node reboots right?
I don't think we want to wipe the certs if they are still valid.

bindata/network/node-identity/common/003-node-identity-configmap.yml
```diff
@@ -0,0 +1,22 @@
...
+      "commonNamePrefix":"system:multus",
+      "organizations": ["system:multus"],
+      "groups": ["system:nodes", "system:multus", "system:authenticated"],
+      "userPrefixes": ["system:node", "system:multus"]
+    }]
+  additional-pod-admission-cond.json: |
+    [{
+      "commonNamePrefix":"system:multus",
+      "allowedPodAnnotations": ["k8s.v1.cni.cncf.io/network-status"]
+    }]
```
kyrtapz, line 22: NIT: missing new line

bindata/network/multus/multus.yaml
```diff
@@ -121,6 +121,11 @@ data:
         "logToStderr": true,
         "logLevel": "verbose",
         "binDir": "{{ .CNIBinDir }}",
+        "perNodeCertificate": {
```
kyrtapz, line 124: Looks like currently multus hard-codes cert-duration to 10min: https://github.com/k8snetworkplumbingwg/multus-cni/pull/1159/files#diff-562bab31c287044ff5fd9a5ba21041103225e129ca4373634b389356bdce0070R101

 I think that the cert rotation shouldn't happen that often in a production environment.
In comparison, kubelet and ovn-kube ruse 24h.

kyrtapz (changes requested):

bindata/network/multus/multus.yaml
dougbtv: ok, yep, I think it needs a rebase likely. good  eye

bindata/network/multus/multus.yaml
dougbtv: Great observation. Turns out that we'll need a modification to ensure that we re-use it on the Multus end instead of re-request. This will certainly be an improvement, but we'll address it as a follow-on item.

bindata/network/multus/002-rbac.yaml
dougbtv: Good call. Added conditions for RBAC + serviceaccount + serviceaccountname, as well as multus configuration to disable cert per node mode.

bindata/network/multus/002-rbac.yaml
```diff
@@ -53,37 +54,153 @@ rules:
...
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
 
+# We keep this service account for upgrades.
+# It will be unreferenced and unused after the upgrade.
+# We assign it a new ClusterRoleBinding to a less permissive ClusterRole
```
jcaamano, line 116: The comment no longer applies

bindata/network/multus/002-rbac.yaml
```diff
@@ -53,37 +54,153 @@ rules:
...
+    name: system:multus
+    apiGroup: rbac.authorization.k8s.io
+{{ else }}
+# Network node identity is disabled, therefore, we use a service account for Multus itself
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: multus-proper
+  namespace: openshift-multus
```
jcaamano, line 159: We don't need this here anymore, we will always have the SA

dougbtv: got it. I can change the following clusterrolebinding to bind to the extant `multus` serviceaccount.

PR #2010, opened 2023-09-19
multinetworkpolicy: allow Neighbor Discovery Protocol traffic

zeeke:
MultusNetworkPolicy iptables implementation denies all traffic that is not specified by policy rules. This can bring to some unexpected scenario, especially when working with IPv6.

Unlike ARP in IPv4, Neighbor Discovery Protocol (NDP) uses IP packets to exchange information between peers, so they are affected by iptables rules. In a scenario where a server is the subject of a MultiNetworkPolicy and a client is the only one who can send traffic to the server, the latter should be able to send the neighbor advertisements, otherwise the client will never know the link-layer address of the server.

multus-networkpolicy doc: https://github.com/openshift/multus-networkpolicy/blob/master/docs/configurations.md#add-custom-iptablesip6tables-rules

### DIFF
--- a/bindata/network/multus-networkpolicy/custom-iptables-rules.yaml
+++ b/bindata/network/multus-networkpolicy/custom-iptables-rules.yaml
@@ -1,0 +1,17 @@
+---
+# Following rules are applied to every pod with an MultiNetworkPolicy working on it, allowing 
+# a base networking that couldn't be expressed by the policy syntax.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: multi-networkpolicy-custom-rules
+  namespace: openshift-multus
+data:
+  
+  custom-v6-rules.txt: |
+    # accept NDP
+    -p icmpv6 --icmpv6-type neighbor-solicitation -j ACCEPT
+    -p icmpv6 --icmpv6-type neighbor-advertisement -j ACCEPT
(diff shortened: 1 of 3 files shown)


zeeke:
/retest

@s1061123 , @dougbtv please take a look when you have time

zeeke:
This PR mainly addresses the following scenario:

Given a namespace nsX with 3 pods, podA and podB, podC attached to a secondary network.

A multinetworkpolicy policy1, defined as
```
apiVersion: k8s.cni.cncf.io/v1beta1
kind: MultiNetworkPolicy
metadata:
  name: policy1
  namespace: nsX
  annotations:
    k8s.v1.cni.cncf.io/policy-for: net1
spec:
  podSelector:
    matchLabels:
      app: podA
  ingress:
    - from:
      - podSelector:
          matchLabels:
            app: podB
```

The policy allows incoming connections to podA from podB. It denies any other incoming connections to podA. Anyway, podA should reach podC, and to achieve it, should receive at least Neighbor Advertisement messages from podC.
Without this PR, those messages are dropped by podA iptables, resulting in podA not being able to reach podC.

@s1061123 I can rework this PR to deal with Neighbor Solicitation and Neighbor Advertisement packets only, and we can think about other message types later on.

PR #2019, opened 2023-09-21
[release-4.14] Relax conditions to get IC upgrade started

ricky-rav:
We've seen cases in which an upgrade to 4.14 was being performed on a 4.13 cluster where one node was in a Ready,SchedulingDisabled state. The ovnkube-node instance on that node had been removed and CNO considered ovnkube-node DaemonSet as progressing (5 out 6 instances were available). This completely bypassed the IC upgrade path and no IC configmap was created.

Let's relax the conditions on when CNO adds the IC configmap, so we can get IC upgrade started in this case too.

Signed-off-by: Riccardo Ravaioli <rravaiol@redhat.com>
(cherry picked from commit 2fe8d1c7c260769a17c60791b53ea2cb577269c5)

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -2265,10 +2265,12 @@ func isInterConnectEnabledOnMasterDaemonset(ds *appsv1.DaemonSet) bool {
 }
 
 // migration from single zone to multizone is about to start if:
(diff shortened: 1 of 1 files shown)


ricky-rav:
We're still missing upgrade jobs in the release repo for CNO...
https://github.com/openshift/release/tree/master/ci-operator/config/openshift/cluster-network-operator

PR #2020, opened 2023-09-21
OCPBUGS-19550: multus: set MULTUS_NODE_NAME to filter pods to local node

dcbw:
This allows multus to set an apiserver filter to only watch pods on its own node. Should reduce memory usage as multus doesn't care about pods on other nodes.

Clusters with ~60,000 pods show multus using around 1.8G of RSS which seems fairly unnecessary.

QE: can be verified by looking at the top of `kube-multus` container logs in multus pods:

```
$ oc logs -n openshift-multus -c kube-multus multus-llgcj | grep "Filtering pod"
2023-09-21T12:44:11Z [verbose] Filtering pod watch for node "ci-op-fvc0qy12-b3a20-dkw5b-worker-b-m7qmq"
```


dcbw:
/override ci/prow/e2e-aws-sdn-network-migration-rollback

registry timeouts

PR #2021, opened 2023-09-21
OCPBUGS-18396: Fix config status MTU migration not being updated

jcaamano:
Render has side effects in the passed in operConfig that are reflected later on in the status when it is being updated. This stopped working when the the passed in operConfig was changed to a copy that was not then used when applying the status.

It makes sense to use the updated operConfig for everything that comes after, not just Render, so change that to fix the issue.

Fixes: https://github.com/openshift/cluster-network-operator/pull/1952

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -343,8 +343,13 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
-	// Generate the objects
-	objs, progressing, err := network.Render(&newOperConfig.Spec, bootstrapResult, ManifestPath, r.client, r.featureGates)
+	// once updated, use the new config
+	operConfig = newOperConfig
+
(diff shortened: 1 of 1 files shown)


jcaamano:
cc @kyrtapz

jcaamano:
/retitle OCPBUGS-18396: Fix config status MTU migration not being updated

jcaamano:
@dulek PTAL

Would https://github.com/openshift/cluster-network-operator/pull/1988 and https://github.com/openshift/cluster-network-operator/pull/2007 be reverted or stay with this change?

jcaamano:
ci/prow/e2e-network-mtu-migration-ovn-ipv6 seems to be fine, it failed on a final test data collection that has also been oberseved to fail in the normal ipv6 job:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-network-operator/2030/pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-ipv6/1706259564141219840

PR #2037, opened 2023-09-27
OCPBUGS-18089,OCPBUGS-18088: ovnkube: container scripts cleanup

dcbw:
Clean up a bunch of things that are done for us by other things, or just aren't necessary.

Then move all the common functionality to a bash script library shared between all the containers so we can use the same code from multiple daemonsets. Tip of the hat to Tomo for the ConfigMap trick.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -367,3 +367,182 @@ data:
         exit 1
       fi
     }
+
+    function log()
+    {
+        echo "$(date --iso-8601=seconds) [{$1}] ${2}"
+    }
+
+    # cni-bin-copy() detects the host OS and copies the correct shim binary to
+    # the CNI binary directory.
+    #
+    # Requires the following volume mounts:
+    #   /host
+    #   /cni-bin-dir
+    cni-bin-copy()
+    {
+      # collect host os information
+      . /host/etc/os-release
+      rhelmajor=
+      # detect which version we're using in order to copy the proper binaries
+      case "${ID}" in
+        rhcos|scos)
+          RHEL_VERSION=$(echo "${CPE_NAME}" | cut -f 5 -d :)
+          rhelmajor=$(echo $RHEL_VERSION | sed -E 's/([0-9]+)\.{1}[0-9]+(\.[0-9]+)?/\1/')
+        ;;
+        rhel) rhelmajor=$(echo "${VERSION_ID}" | cut -f 1 -d .)
+        ;;
+        fedora)
+          if [ "${VARIANT_ID}" == "coreos" ]; then
+            rhelmajor=8
+          else
+            log "cnibincopy" "FATAL ERROR: Unsupported Fedora variant=${VARIANT_ID}"
+            exit 1
+          fi
+        ;;
+        *) log "cnibincopy" "FATAL ERROR: Unsupported OS ID=${ID}"; exit 1
+        ;;
+      esac
+
+      # Set which directory we'll copy from, detect if it exists
+      sourcedir=/usr/libexec/cni/
+      case "${rhelmajor}" in
+        8)
+          sourcedir=/usr/libexec/cni/rhel8
+        ;;
+        9)
+          sourcedir=/usr/libexec/cni/rhel9
+        ;;
+        *)
+          log "cnibincopy" "ERROR: RHEL Major Version Unsupported, rhelmajor=${rhelmajor}"
+        ;;
+      esac
+
+      cp -f "$sourcedir/ovn-k8s-cni-overlay" /cni-bin-dir/
+    }
+
+    # start-ovnkube-node starts the ovnkube-node process. This function does not
+    # return.
+    start-ovnkube-node()
+    {
+      local log_level=$1
+      local metrics_port=$2
+      local ovn_metrics_port=$3
+
+      # copy the right CNI shim for the host OS
+      cni-bin-copy
+
+      ovn_config_namespace=openshift-ovn-kubernetes
+      echo "I$(date "+%m%d %H:%M:%S.%N") - disable conntrack on geneve port"
+      iptables -t raw -A PREROUTING -p udp --dport {{.GenevePort}} -j NOTRACK
+      iptables -t raw -A OUTPUT -p udp --dport {{.GenevePort}} -j NOTRACK
+      ip6tables -t raw -A PREROUTING -p udp --dport {{.GenevePort}} -j NOTRACK
+      ip6tables -t raw -A OUTPUT -p udp --dport {{.GenevePort}} -j NOTRACK
+      echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node"
+
+      if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
+        gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+      elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
+        gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
+      else
+        echo "Invalid OVN_GATEWAY_MODE: \"{{.OVN_GATEWAY_MODE}}\". Must be \"local\" or \"shared\"."
+        exit 1
+      fi
+
+      export_network_flows_flags=
+      if [[ -n "${NETFLOW_COLLECTORS}" ]] ; then
+        export_network_flows_flags="--netflow-targets ${NETFLOW_COLLECTORS}"
+      fi
+      if [[ -n "${SFLOW_COLLECTORS}" ]] ; then
+        export_network_flows_flags="$export_network_flows_flags --sflow-targets ${SFLOW_COLLECTORS}"
+      fi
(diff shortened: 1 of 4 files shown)


bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -689,161 +370,8 @@ spec:
...
-            ${admin_network_policy_enabled_flag} \
-            --enable-multicast \
-            --zone ${K8S_NODE} \
-            --enable-interconnect \
-            --acl-logging-rate-limit "{{.OVNPolicyAuditRateLimit}}" \
-            ${gw_interface_flag} \
-            --enable-multi-external-gateway=true  \
-            ${ip_forwarding_flag} \
-            ${NETWORK_NODE_IDENTITY_ENABLE}
+          . /ovnkube-lib/ovnkube-lib.sh || exit 1
```
jcaamano, line 373: K8S_NODE sourced above is aslo sourced within ovnkube-lib.sh

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -0,0 +1,547 @@
...
+    ovndb-readiness-probe()
+    {
+      # dbname should be 'sb' or 'nb'
+      local dbname=$1
+
+      local ctlfile
+      if [[ "${dbname}" = "sb" ]]; then
+        ctlfile=${nbdb_ctl}
+      elif [[ "${dbname}" = "nb" ]]; then
+        ctlfile=${sbdb_ctl}
```
jcaamano, line 310: Just a heads up that you started using this variable in a commit but defined it in a later commit.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    # run-ovn-northd() starts ovn-northd and does not return until
+    # northd exits.
+    #
+    # Requires the following volume mounts:
+    #   /etc/openvswitch/
+    #   /var/lib/openvswitch/
+    #   /run/openvswitch/
+    #   /run/ovn/
```
jcaamano, line 85: `/var/log/ovn/` as well?

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -8,7 +8,6 @@ metadata:
   {{ else }}
   name: ovnkube-node
   {{ end }}
-  name: ovnkube-node
```
jcaamano, line 11: lol, k8s did not complain?

bindata/network/ovn-kubernetes/self-hosted/multi-zone-interconnect/ovnkube-node.yaml
```diff
@@ -746,162 +404,8 @@ spec:
...
-            ${admin_network_policy_enabled_flag} \
-            --enable-multicast \
-            --zone ${K8S_NODE} \
-            --enable-interconnect \
-            --acl-logging-rate-limit "{{.OVNPolicyAuditRateLimit}}" \
-            ${gw_interface_flag} \
-            --enable-multi-external-gateway=true \
-            ${ip_forwarding_flag} \
-            ${NETWORK_NODE_IDENTITY_ENABLE}
+          . /ovnkube-lib/ovnkube-lib.sh || exit 1
```
jcaamano, line 407: K8S_NODE sourced above is aslo sourced within ovnkube-lib.sh

jcaamano (commented):
Great stuff!

I lost all my comments mid-review so hopefully I recalled all of them. Mostly nits though so not too concerned.

Should we review how we use env overrides? Wouldn't we want to provide that only for ovn-k specific containers and not ovn ones? I think I saw a place where env overrides are mounted but K8S_NOdE env is not defined, ovn-controller I think. I don't know if that's intentional but I would do a final pass for this.

dcbw:
> I think I saw a place where env overrides are mounted but K8S_NOdE env is not defined, ovn-controller I think. I don't know if that's intentional but I would do a final pass for this.

@jcaamano yeah this was intentional and that's why I added the `if [[ -n "${K8S_NODE}" &&` bit; it seemed like enough containers could use per-node overrides of things that it was useful to make generic.

jcaamano:
> > I think I saw a place where env overrides are mounted but K8S_NOdE env is not defined, ovn-controller I think. I don't know if that's intentional but I would do a final pass for this.
> 
> @jcaamano yeah this was intentional and that's why I added the `if [[ -n "${K8S_NODE}" &&` bit; it seemed like enough containers could use per-node overrides of things that it was useful to make generic.

But does it make sense to mount env overrides in ovn-controller container without setting the K8S_NODE env variable there? So we don't want per-node override there?

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
jluhrsen, line 1: Thanks for doing this @dcbw. There is a bug I just started to look at which I think this might help with.
I'm testing with 10s now with this PR. I'll close it once I know it helps and you can get this change done
here.

How come you didn't modify this in the self-hosted/ folder yamls?

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
dcbw: Fixed

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
dcbw: @jcaamano hmm, can you elaborate? do you mean `dbname` or `ctlfile`?  AFAICS I define and use both of those in `ovnkube: move nbdb container script to common library` but I'm sure I'm missing something

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
dcbw: Nope, apparently not.

bindata/network/ovn-kubernetes/self-hosted/multi-zone-interconnect/ovnkube-node.yaml
dcbw: Fixed.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
dcbw: Fixed.

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
dcbw: @jluhrsen I did self-hosted & managed multi-zone-interconnect only. I haven't looked, but I figured the single-zone and migration ones wouldn't be quite the same, though we could try to do those in a follow-up.

dcbw: @jluhrsen let me know if you'd like 10s and I'll change to that. I did 30 because it seemed like 90 was no longer necessary, but I wasn't sure that removing it entirely was safe.

But actually, looking at the definition of readiness in https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes we're clearly not using it for its intended purpose, and I'll bet we can drop it down to 10.

dcbw:
> But does it make sense to mount env overrides in ovn-controller container without setting the K8S_NODE env variable there? So we don't want per-node override there?

@jcaamano I see where you're going with this; Looks like I mistakenly removed the K8S_NODE env from the controller containers. I'll add it back.

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-node.yaml
jluhrsen: I'm guessing we could get rid of the delay altogether and still be fine, but just know that I tried with 10s and 
did a small local test and most presubmits passed in case you want to change that. it would reduce rollout times
even more.

As for self-hosted, I was only referring to this timer. just noticed you didn't change those. we can do it in a follow
up if needed.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
jcaamano: Sorry, wasn't clear enough here.

I meant `sbdb_ctl` which started to be used in `ovnkube: move nbdb container script to common library` but was defined in `ovnkube: move sbdb container script to common library``

jcaamano:
/hold

for https://github.com/openshift/cluster-network-operator/pull/2037#issuecomment-1749281088
and https://github.com/openshift/cluster-network-operator/pull/2037#discussion_r1349240645

but either way that goes this LGTM

Taking the opportunity to retest but the IPv6 jobs looks like they might need an override as that infra setup looks to be very flakey these days.

PR #2039, opened 2023-09-27
OCPBUGS-19861: Multus per-node certificates should have 24h duration

dougbtv:
Also uses `/etc/cni/multus/certs` for certificate storage on host

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -126,7 +126,8 @@ data:
         "perNodeCertificate": {
           "enabled": true,
           "bootstrapKubeconfig": "/hostroot/var/lib/kubelet/kubeconfig",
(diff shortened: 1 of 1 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -126,7 +126,8 @@ data:
         "perNodeCertificate": {
           "enabled": true,
           "bootstrapKubeconfig": "/hostroot/var/lib/kubelet/kubeconfig",
-          "certDir": "/run/multus/certs"
+          "certDir": "/etc/cni/multus/certs",
+          "certDuration": "24h"
```
kyrtapz, line 130: NIT: We could probably make this a variable if in the future we want to enable customizing that value

kyrtapz:
/lgtm
I had one comment which can be addressed separately.

PR #2042, opened 2023-09-28
OCPBUGS-19918: get ipsecStatus from host daemonset

yuvalk:
ds name changed and since both ipsec-host and ipsec-containerized are
paired, we can get by getting either


trozet (approved):
/lgtm

PR #2044, opened 2023-09-28
Network metrics daemon: change priority class to openshift-user-critical

martinkennelly:
I noticed during CI tests that sometimes this component will attempt to start at the same time the CNI is attemting to start.

This can cause multus to return an error during CNI ADD and cause test failures.

Metrics are user facing.
See openshift docs on priority class definitions: https://docs.openshift.com/container-platform/4.13/nodes/pods/nodes-pods-priority.html

### DIFF
--- a/bindata/network/network-metrics/001-daemonset.yaml
+++ b/bindata/network/network-metrics/001-daemonset.yaml
@@ -38,7 +38,7 @@ spec:
                 operator: DoesNotExist
               - key: network.operator.openshift.io/dpu
                 operator: DoesNotExist
-      priorityClassName: "system-node-critical"
+      priorityClassName: "openshift-user-critical"
       tolerations:
         - operator: Exists
       containers:


martinkennelly:
I was in two minds about whether to use `system-cluster-critical` or `openshift-user-critical`. I chose the latter because I dont think the metrics are important for the cluster to function. My goal is to not want this component to be the same scheduling priority as the CNI.

bindata/network/network-metrics/001-daemonset.yaml
```diff
@@ -38,7 +38,7 @@ spec:
                 operator: DoesNotExist
               - key: network.operator.openshift.io/dpu
                 operator: DoesNotExist
-      priorityClassName: "system-node-critical"
+      priorityClassName: "openshift-user-critical"
```
dougbtv, line 41: This is definitely the right one for metrics, seeing it's user facing and doesn't have an impact on core node functionality. Good one!

dougbtv (approved):

PR #2048, opened 2023-10-04
OCPBUGS-19370: Added HCP label to CNO pods

jparrill:
Added 2 missing labels in the CNO pods on Hypershift context:
    
- hypershift.openshift.io/hosted-control-plane
- hypershift.openshift.io/control-plane
    
The `hypershift.openshift.io/hosted-control-plane` label needs to be present to signal the Management Cluster router to fulfil the route.

The `hypershift.openshift.io/control-plane` is needed to mark those pods as part of the ControlPlane avoiding to fall into the serving-component-nodes.

Signed-off-by: Juan Manuel Parrilla Madrid <jparrill@redhat.com>

### DIFF
--- a/bindata/network/multus-admission-controller/admission-controller.yaml
+++ b/bindata/network/multus-admission-controller/admission-controller.yaml
@@ -9,6 +9,7 @@ metadata:
 {{- if .HyperShiftEnabled}}
     # used by PodAffinity to prefer co-locating pods that belong to the same hosted cluster.
     hypershift.openshift.io/hosted-control-plane: {{.AdmissionControllerNamespace}}
+    hypershift.openshift.io/control-plane: "true"
 {{- end }}
   annotations:
     kubernetes.io/description: |
@@ -42,6 +43,10 @@ spec:
       labels:
         app: multus-admission-controller
         namespace: {{.AdmissionControllerNamespace}}
+{{- if .HyperShiftEnabled}}
+        hypershift.openshift.io/hosted-control-plane: {{.AdmissionControllerNamespace}}
+        hypershift.openshift.io/control-plane: "true"
+{{- end }}
         component: network
         type: infra
         openshift.io/component: network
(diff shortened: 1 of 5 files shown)


jparrill:
waiting on @mjlshen to confirm if the label is relevant in more pod templates than the included in this PR

bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-control-plane.yaml
```diff
@@ -38,6 +38,7 @@ spec:
         type: infra
         openshift.io/component: network
         hypershift.openshift.io/control-plane-component: ovnkube-control-plane
+        hypershift.openshift.io/hosted-control-plane: {{.HostedClusterNamespace}}
```
enxebre, line 41: they should also have "hypershift.openshift.io/control-plane":true, both are used for our affinity opinions

kyrtapz:
@jparrill Can you add a commit message with an explanation around what's the purpose of the new labels?

bindata/network/multus-admission-controller/admission-controller.yaml
```diff
@@ -42,6 +43,8 @@ spec:
       labels:
         app: multus-admission-controller
         namespace: {{.AdmissionControllerNamespace}}
+        hypershift.openshift.io/hosted-control-plane: {{.AdmissionControllerNamespace}}
```
kyrtapz, line 46: These are hypershift specific so they should be under `{{- if .HyperShiftEnabled}}` condition.

enxebre:
>The hypershift.openshift.io/control-plane is needed to mark those pods as part of the ControlPlane avoding to fall into the serving-component-nodes.

Thanks! The statement above is not exactly accurate. All the dedicated serving content pods have this label as well. All control plane workloads should have it so they respond to tolerations and affinity for managed cluster Nodes taints and labels. See more details here https://hypershift-docs.netlify.app/how-to/distribute-hosted-cluster-workloads/

/lgtm

kyrtapz:
/lgtm
@jcaamano ptal for approve

PR #2051, opened 2023-10-05
OCPBUGS-20104: Don't run network node identity as root

jcaamano:


jcaamano:
/retitle OCPBUGS-20104: Don't run network node identity as root

abhat:
Do you need an additional CAP_NET to be able to bind to localhost in the nodes default network namespace?

jcaamano:
> Do you need an additional CAP_NET to be able to bind to localhost in the nodes default network namespace?

Only for privileged ports < 1024 I believe

PR #2052, opened 2023-10-05
[SDN-4071] two replicas for CM instead of 3 

JacobTanenbaum:
https://issues.redhat.com/browse/SDN-4071

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -407,6 +407,13 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		data.Data["OVN_MULTI_NETWORK_POLICY_ENABLE"] = true
 	}
 
+	//there only needs to be two cluster managers
+	clusterManagerReplicas := 2
+	if len(bootstrapResult.OVN.MasterAddresses) < 2 {
+		clusterManagerReplicas = len(bootstrapResult.OVN.MasterAddresses)
+	}
+	data.Data["ClusterManagerReplicas"] = clusterManagerReplicas
+
 	var manifestSubDir string
 	manifestDirs := make([]string, 0, 2)
 	manifestDirs = append(manifestDirs, filepath.Join(manifestDir, "network/ovn-kubernetes/common"))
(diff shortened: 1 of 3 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -125,7 +125,13 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 		data.Data["OvnControlPlaneImage"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.ControlPlaneImage
 	}
-	data.Data["OvnkubeMasterReplicas"] = len(bootstrapResult.OVN.MasterAddresses)
+	//there only needs to be at most two cluster managers
+	clusterManagerReplicas := 2
+	if len(bootstrapResult.OVN.MasterAddresses) < 2 {
+		clusterManagerReplicas = len(bootstrapResult.OVN.MasterAddresses)
+	}
+	data.Data["OvnkubeMasterReplicas"] = clusterManagerReplicas
+
```
ricky-rav, line 134: Here we're forcing `OvnkubeMasterReplicas` to take the number of replicas we want for cluster manager, that is two at most or one if there's only one master.
In the single-zone-interconnect folder, we still need the correct value for `OvnkubeMasterReplicas` for setting up the number of RAFT replicas and for the alerts that we're bringing over from 4.13:
```
$ grep -rnil OvnkubeMasterReplicas
bindata/network/ovn-kubernetes/self-hosted/single-zone-interconnect/alert-rules-control-plane.yaml
bindata/network/ovn-kubernetes/self-hosted/multi-zone-interconnect/ovnkube-control-plane.yaml
bindata/network/ovn-kubernetes/managed/single-zone-interconnect/ovnkube-master.yaml
bindata/network/ovn-kubernetes/managed/single-zone-interconnect/alert-rules-control-plane.yaml
bindata/network/ovn-kubernetes/managed/multi-zone-interconnect/ovnkube-control-plane.yaml
pkg/network/ovn_kubernetes.go
```
As for `clusterManagerReplicas`, I would either create a new key inside `data.Data` or set it directly in the YAML for ovnkube-control-plane. What do you think?

ricky-rav (changes requested):

ricky-rav:
/lgtm
Thanks, Jacob!

PR #2053, opened 2023-10-05
reduce nbdb & sbdb initialDelaySeconds: 90s -> 10s

jluhrsen:
the initial need for this delay was to allow RAFT clusters to settle. Without it now, we don't need to wait as long.

JIRA: https://issues.redhat.com/browse/OCPBUGS-18349


dcbw:
Partly superceded by https://github.com/openshift/cluster-network-operator/pull/2037

PR #2063, opened 2023-10-13
[release-4.14] OCPBUGS-20472: hosted cluster upgrade failure from 4.13 stable to 4.14

flavio-fernandes:
Add missing volume mount to ovnkube-node pods while in the single zone phase

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/single-zone-interconnect/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/single-zone-interconnect/ovnkube-node.yaml
@@ -658,6 +658,10 @@ spec:
           name: host-var-lib-cni-networks-ovn-kubernetes
         - mountPath: /run/openvswitch
           name: run-openvswitch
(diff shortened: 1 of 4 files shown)


ricky-rav:
I prefer this solution to having the ovnkube-config ConfigMap copied in two places :)
Let's try it out.
Thanks, Flavio!

bindata/network/ovn-kubernetes/managed/single-zone-interconnect/ovnkube-node.yaml
```diff
@@ -658,6 +658,10 @@ spec:
           name: host-var-lib-cni-networks-ovn-kubernetes
         - mountPath: /run/openvswitch
           name: run-openvswitch
+        # gives write access to libovsdb logs until moving to multi-zone
```
flavio-fernandes, line 661: @tssurya is this comment in line with what you had in mind?

trozet (approved):
/approve

PR #2065, opened 2023-10-13
OCPBUGS-20519: hosted cluster upgrade failure from 4.13 stable to 4.14

flavio-fernandes:
Add missing volume mount to ovnkube-node pods while in the single zone phase


abhat:
This failure isn't related to any changes the PR is introducing. The job is failing since the last couple of days.

PR #2076, opened 2023-10-20
[release-4.14] OCPBUGS-26573: Improve troubleshooting IC upgrades

ricky-rav:
If ever a cluster is not making progress during phase 1 or phase 2 of the upgrade to OVN IC, a cluster admin can now edit the interconnect configmap and add a new key (`fast-forward-to-multizone`) to bypass the two-phase upgrade and let CNO apply directly the multizone YAMLs.

The current code only allows the cluster to move forward when it's in phase 1 by setting `zone-mode=multizone` and `temporary=false`:
CurrentICupgrades drawio

Let's improve that by allowing to jump to multizone in either phase of the upgrade. At the same time, remove the `temporary` field from the configmap, since the `ongoing-upgrade` field already tracks that CNO is going through an upgrade:
ICupgrades drawio

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -1256,20 +1256,18 @@ const (
 )
 
 type targetZoneModeType struct {
-	// zoneMode indicates the target zone mode that CNO is supposed to converge to.
+	// zoneMode indicates the target zone mode that CNO is supposed to converge to. It defaults
+	// multizone.
 	zoneMode InterConnectZoneMode
-	// "configMapFound" indicates whether the interconnect configmap was found; when not found,
-	// the zone mode defaults to multizone.
+	// "configMapFound" indicates whether the interconnect configmap was found. The configmap
+	// is created by CNO itself during a 4.13->4.14 upgrade and deleted once the upgrade is over.
 	configMapFound bool
-	// ongoingUpgrade is true when the configmap was pushed by CNO itself; it is used by status manager
-	// to know it has to keep reporting the old version when this parameter is set.
-	ongoingUpgrade bool
 	// fastForwardToMultiZone can be manually set by the cluster admin through the interconnect configmap
 	// in order to force CNO to deploy multizone OVNK regardless of the current status of OVNK components
-	// throughout an upgrade to interconnect. This would effectively get the cluster out of phase 1 and phase 2
+	// during an upgrade to interconnect. This would effectively get the cluster out of phase 1 and phase 2
 	// and make OVNK jump to the YAMLs in the multi-zone-interconnect folder.
 	// This can be useful in case of problems during  upgrades, if ever the logic in prepareUpgradeToInterconnect
-	// is not moving forward (e.g. one or more nodes are down and ovnk node is considered as "progressing")
+	// is not moving forward (e.g. one or more nodes are down and ovnkube-node is considered as "progressing")
 	fastForwardToMultiZone bool
 }
 
@@ -1303,10 +1301,6 @@ func getTargetInterConnectZoneMode(kubeClient cnoclient.Client) (targetZoneModeT
 	} else {
 		klog.Infof("no target zone mode in interconnect configmap, defaulting to multizone")
 		targetZoneMode.zoneMode = zoneModeMultiZone
-	}
-
-	if _, ok := interConnectConfigMap.Data["ongoing-upgrade"]; ok {
-		targetZoneMode.ongoingUpgrade = true
 	}
 
 	if _, ok := interConnectConfigMap.Data["fast-forward-to-multizone"]; ok {
@@ -2396,7 +2390,7 @@ func doesVersionEnableInterConnect(string) bool {
 // If we're upgrading from a 4.13 cluster, which has no OVN InterConnect support, three phases are necessary.
 // Phase 1:
 //
-//	a) prepareUpgradeToInterConnect pushes a configMap with zone-mode=singlezone, ongoing-upgrade='';
+//	a) prepareUpgradeToInterConnect pushes a configMap with zone-mode=singlezone;
 //	b) renderOVNKubernetes selects the YAMLs from the single-zone folder
 //	   (node DaemonSet, master DaemonSet [StatefulSet for hypershift], ovnkube-sbdb route [hypershift]);
 //	c) shouldUpdateOVNKonUpgrade rolls out first node DaemonSet then master DaemonSet/StatefulSet in single-zone mode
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1258,16 +1258,19 @@ const (
 type targetZoneModeType struct {
 	// zoneMode indicates the target zone mode that CNO is supposed to converge to.
 	zoneMode InterConnectZoneMode
-	// "temporary", if true, indicates that the target zone mode is only temporary;
-	// it is used along with zoneMode=singlezone in order to temporarily switch to single-zone mode when upgrading
-	// from versions with no interconnect support (<=4.13). It has no effect if zoneMode=multizone.
-	temporary bool
```
tssurya, line 1264: so this variable is now removed in 4.14.7 or whatever z stream this lands in right?
how will the new upgrades from nonIC to 4.14.z where this change lands look like? I see
`remove the temporary field from the configmap, since the ongoing-upgrade field already tracks that CNO is going through an upgrade` in the PR message so that means we no longer need to know where the phase is temporary or not, just need to know if upgrade is ongoing or not? why did we need the `temporary` flag originally? seems like it was redundant with ongoingUpgrade?

I guess my TL;DR question is is there a difference between going from 4.13.0 to 4.14.0 and 4.13.0 to 4.14.z where this lands functionally for the end user if this field is going away? Do we need to make updates to the original enhancement?

pkg/network/ovn_kubernetes.go
```diff
@@ -1302,15 +1305,19 @@ func getTargetInterConnectZoneMode(kubeClient cnoclient.Client) (targetZoneModeT
...
 	if _, ok := interConnectConfigMap.Data["ongoing-upgrade"]; ok {
 		targetZoneMode.ongoingUpgrade = true
 	}
 
-	klog.Infof("target zone mode: %+v", targetZoneMode)
+	if _, ok := interConnectConfigMap.Data["fast-forward-to-multizone"]; ok {
+		targetZoneMode.fastForwardToMultiZone = true
+		if targetZoneMode.zoneMode != zoneModeMultiZone {
+			klog.Warningf("Forcing interconnect zone mode to multizone due to 'fast-forward-to-multizone' being set")
+			targetZoneMode.zoneMode = zoneModeMultiZone
```
tssurya, line 1316: so this is sort of a disruptive change in case ongoingUpgrade=true and someone sets this right? example, single zone is rolling out and stuck and now we set `fastForwardToMultiZone` to true to unblock and move forward with multizone.. will the single zone pods be left behind or will cleanup happen? 

Also I guess we can add something to the PR description or enhancement update around this option to say "at your own risk - to be used only with guidance from eng or something"

pkg/network/ovn_kubernetes.go
```diff
@@ -1258,16 +1258,19 @@ const (
...
 	// zoneMode indicates the target zone mode that CNO is supposed to converge to.
 	zoneMode InterConnectZoneMode
-	// "temporary", if true, indicates that the target zone mode is only temporary;
-	// it is used along with zoneMode=singlezone in order to temporarily switch to single-zone mode when upgrading
-	// from versions with no interconnect support (<=4.13). It has no effect if zoneMode=multizone.
-	temporary bool
 	// "configMapFound" indicates whether the interconnect configmap was found; when not found,
 	// the zone mode defaults to multizone.
 	configMapFound bool
 	// ongoingUpgrade is true when the configmap was pushed by CNO itself; it is used by status manager
```
tssurya, line 1264: do we need to make updates to what ongoingUpgrade means? It seems like its also taking over the role that "temporary" was doing...

pkg/network/ovn_kubernetes.go
```diff
@@ -2463,11 +2468,6 @@ func prepareUpgradeToInterConnect(ovn bootstrap.OVNBootstrapResult, client cnocl
 				"path":  "/data/zone-mode",
 				"value": fmt.Sprint(zoneModeMultiZone),
 			},
-			{
-				"op":    "replace",
-				"path":  "/data/temporary",
-				"value": "false",
-			},
 		}
```
tssurya, line 2471: hmm maybe stupid question, but why isn't ongoing-upgrade also patched?

tssurya (commented):
Overall logic is clear where you are trying to add a fast-forward knob. Left some inline questions...
Do we need some unit tests?
Did you do live testing and results were ok? (asking because we are kinda going in blind here...)

pkg/network/ovn_kubernetes.go
ricky-rav: I could have simplified the logic earlier on, but I wanted to leave a window open in case we were determined to allow single zone... 

Here's how it went.

The `temporary` field was how I envisaged to distinguish between being permanently in single zone (temporary=false) or just momentarily during the upgrade to 4.14 (temporary=true). The idea was that a configmap would track an upgrade to IC or a voluntary switch to single zone.

Then I needed another flag in the configmap, `ongoing-upgrade`, so that the CNO status manager could know whether an upgrade to IC was ongoing: it would only report version=4.14 when the upgrade was done, which was signaled by `ongoingUpgrade` getting removed from the configmap.

Well after 4.14 GA, we decided not to enable switching back to single zone, so now we can finally get rid of `temporary`.

pkg/network/ovn_kubernetes.go
ricky-rav: Yes, using the new flag is disruptive: CNO will push the final multizone YAMLs regardless of where the cluster is during the 4.13->4.14 upgrade. After this PR merges, my idea was to write a document about this for support. Either that or I update the enhancement...

As for the single-zone pods, they will be removed as a consequence of the new ones being pushed.

pkg/network/ovn_kubernetes.go
ricky-rav: Actually, since we've decided that we're not allowing a cluster to be permanently in single zone, I might as well remove "ongoingUpgrade" and just refer to `configMapFound`, since only CNO is supposed to push the configmap now... let me try it out!

ricky-rav:
I'll test the new changes tomorrow morning...

pkg/network/ovn_kubernetes.go
tssurya: ack did you try it? all ok on this PR?

tssurya (approved):
/lgtm

pkg/network/ovn_kubernetes.go
ricky-rav: yes, I retested everything after pushing the current changes: https://github.com/openshift/cluster-network-operator/pull/2076#issuecomment-1851982323

trozet (approved):
/approve
/label backport-risk-assessed

ricky-rav:
@anuragthehatter for the final diagrams & configmap please refer to https://github.com/openshift/enhancements/pull/1567

PR #2082, opened 2023-10-25
OCPBUGS-21924: Bump library-go

kyrtapz:


go.mod
```diff
@@ -20,11 +18,11 @@ require (
 	github.com/vishvananda/netlink v1.1.0
 	github.com/vishvananda/netns v0.0.0-20200728191858-db3c7e526aae // indirect
 	golang.org/x/net v0.14.0
-	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.27.3
-	k8s.io/apimachinery v0.27.3
-	k8s.io/code-generator v0.27.2
-	k8s.io/component-base v0.27.2
+	gopkg.in/yaml.v2 v2.4.0 // indirect
```
jcaamano, line 21: don't indirect deps have its own block?

jcaamano:
@kyrtapz can you please add the commands that lead you to these changes?

kyrtapz:
> @kyrtapz can you please add the commands that lead you to these changes?

Done, I also added an additional commit running `go mod tidy` on master (first commit)

go.mod
kyrtapz: Apparently when a dependency change from direct to indirect it doesn't change sections.
I removed that line and re-ran `go mod tidy` to clean this up.

jcaamano:
/approve
/lgtm
/hold

@kyrtapz remove the hold when you are satisfied with the CI. I would rather get all the greens we can get as long as we have the time, but up to you.

kyrtapz:
> /approve /lgtm /hold
> 
> @kyrtapz remove the hold when you are satisfied with the CI. I would rather get all the greens we can get as long as we have the time, but up to you.

Since https://github.com/openshift/cluster-network-operator/pull/2128 went in first I reworked the PR so the bump is extremely straight forward now. 
@jcaamano please re-add the lgtm :)

kyrtapz:
/hold
need to update the code to match the 1.28 library-go

PR #2083, opened 2023-10-25
OCPBUGS-22244: Remove IPsec daemonset after disabling IPsec in OVN

pperiyasamy:
The ipsec daemonset pods are not removed currently because it keeps ovs monitor ipsec process running so that ovn disable ipsec option can be processed in ovs monitor script and then ipsec can be disabled in kernel.
Hence this PR makes operator to wait for ovn ipsec disable is rolled out into all nodes and then it goes out and remove ipsec daemonset.


pperiyasamy:
Closing it as this case is addressed with PR https://github.com/openshift/cluster-network-operator/pull/2087 already.

PR #2084, opened 2023-10-26
OCPBUGS-11179: change CNO to use custom ServiceAccount

jcaamano:
To be compliant with CIS benchmark rule, network operator should run with separate service account instead of using default


jcaamano:
/hold

want to make sure I evaluate CI results before merging

PR #2086, opened 2023-10-30
use $CPE_NAME to find the OS major version

luenta:
RHEL_VERSION isn't present in scos, but $CPE_NAME
is present in /etc/os-release on both distros.

As I could tell it has been fixed above 4.13, so I'd like to fix in this branch too.


LorbusChris:
Thanks for stepping up @luenta! PRs are always welcome, but unfortunately backports require additional QE scrutiny. Since we're not planning on cutting another OKD 4.13 release, this change wouldn't land anyway. OKD 4.14 will be out soon, so please stay tuned.

/close

PR #2088, opened 2023-10-31
Narrow down CNO RBAC

jcaamano:
Recently, a CNO service account was introduced bound to the cluster admin role. Those permissions are more than we need, so try to narrow them down with specific roles.


jcaamano:
/hold 

WIP

PR #2089, opened 2023-10-31
[OCPBUGS-22729] Add IP validation in schema for static hop IP

jordigilh:
Adds IP validation in the static hop IP field for the Admin Policy Based External Route CR to ensure the field is matches an IPv4 or IPv6 format.

```

@tshefi FYI


jordigilh:
`format:ip` doesn't solve the issue. Looking for a different approach

PR #2092, opened 2023-11-02
WIP-Enable NS IPSec in CRD

JoshSalomon:
Do not merge yet


pkg/network/ovn_kubernetes.go
```diff
@@ -336,6 +336,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 
+	if c.IPSecExternConfig != nil {
+		data.Data["OVNIPsecDaemonsetEnable"] = true
+		//TODO: Remove
+		klog.Infof("DBG-Josh: IPsec NS enabled: %t", data.Data["OVNIPsecDaemonsetEnable"])
+	} else {
+		// Intentionally kept empty in case it was enebled before
+	}
+
+	klog.Infof("IPSec: IPSecEnabe [%t], IPSecDaemonsetEnable [%t]", data.Data["OVNIPsecEnable"], data.Data["OVNIPsecDaemonsetEnable"])
```
JoshSalomon, line 347: @yuvalk - the existing IPSec functionality (EW) holds 2 flags, one for OVNIPsecEnable and one for the OVNIPsecDaemonsetEnable (which once is on, it is never turned off) - should we have the same for OVNIPsecExternEnable? - otherwise the OVNIPsecDaemonsetEnable flag does not say whether the user asked to turn NS IPsec on or off

PR #2094, opened 2023-11-03
[release-4.13] OCPBUGS-22293: remove all managed fields used by old manager

jluhrsen:
Previous to 4.11, we used client side update operations with a manager called 'cluster-network-operator', however following 4.11, we changed the field manager name

field because the depreciated field manager owned and preserved the field.

This patch removes the depreciated field manager
regardless of the operation it used previously.

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
 	utilpointer "k8s.io/utils/pointer"
 )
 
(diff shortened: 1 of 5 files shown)


jcaamano:
/lgtm
/approve
/label backport-risk-assessed

@jluhrsen I missed the cherry-pick comment (`git cherry-pick -x`) so I could easily refer to the cherry-picked commits.
Also I am not sure if the vendored code comes from the cherrypick or from actual vendoring. But I did crosscheck that the vendored code is the same as the dependency it should come from (client-go v0.26.1)

wking:
[`Cluster resource quota should control resource limits across namespaces` and 1s `disruption/ci-cluster-network-liveness` hiccup seem unrelated][1]:

/retest-required

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2094/pull-ci-openshift-cluster-network-operator-release-4.13-e2e-aws-ovn-network-migration/1745482656763613184

PR #2096, opened 2023-11-07
OCPBUGS-23011: Block upgrades to 4.15 with Kuryr

dulek:
Kuryr is not supported on 4.15. This commit makes sure that for clusters with Kuryr configured in `.spec.defaultNetwork.type`, the upgrade to 4.15 will be blocked.

### DIFF
--- a/pkg/controller/statusmanager/status_manager_test.go
+++ b/pkg/controller/statusmanager/status_manager_test.go
@@ -135,7 +135,7 @@ func TestStatusManager_set(t *testing.T) {
 	set(t, client, no)
 
 	condUpdate := operv1.OperatorCondition{
-		Type:   operv1.OperatorStatusTypeUpgradeable,
+		Type:   string(configv1.OperatorUpgradeable),
 		Status: operv1.ConditionTrue,
 	}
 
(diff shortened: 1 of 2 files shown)


dulek:
I updated the PR to add a bit of unit testing to this.

mandre (commented):
/lgtm

dulek:
/retest

Let's try these jobs once again to be sure.

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -253,6 +253,59 @@ func TestStatusManager_set(t *testing.T) {
 	}
 }
 
+func TestStatusManager_set_kuryr(t *testing.T) {
+	client := fake.NewFakeClient()
+	status := New(client, "testing", "")
+
+	// make the network.operator object
+	log.Print("Creating Network Operator Config")
```
MaysaMacedo, line 261: leftover?

MaysaMacedo (commented):
We will need it in 4.14 as well, right?

pkg/controller/statusmanager/status_manager_test.go
dulek: I've copied this from TestStatusManager_set() above. It proves to be useful when debugging the tests.

dulek:
> We will need it in 4.14 as well, right?

This is being merged to 4.14. Then we'll bump the upgrade graph to make sure that you need to go through this version before upgrading to 4.15.

dcbw:
@dulek it's not immediately clear to me from the code, but will setting Upgradable=false block Z-stream upgrades, or just Y stream ones?

dcbw:
@dulek Hmm... The code const that's getting used says:


```
        // Upgradeable indicates that the operator configuration itself (not prereqs) can be auto-upgraded by the CVO
        OperatorStatusTypeUpgradeable = "Upgradeable"
```

which doesn't mention minor versions at all. But the CVO does check the const you mention `configv1.OperatorUpgradeable` which is kinda confusing. I'm trying to get clarification of why we have both of these constants.

Given that the CVO, which is the thing that actually uses this constant, is using `configv1.OperatorUpgradeable` internally, would you mind changing both of the occurrances to that instead?

dulek:
> @dulek Hmm... The code const that's getting used says:
> 
> ```
>         // Upgradeable indicates that the operator configuration itself (not prereqs) can be auto-upgraded by the CVO
>         OperatorStatusTypeUpgradeable = "Upgradeable"
> ```
> 
> which doesn't mention minor versions at all. But the CVO does check the const you mention `configv1.OperatorUpgradeable` which is kinda confusing. I'm trying to get clarification of why we have both of these constants.
> 
> Given that the CVO, which is the thing that actually uses this constant, is using `configv1.OperatorUpgradeable` internally, would you mind changing both of the occurrances to that instead?

@dcbw: I'm guessing `OperatorUpgradeable` is for `ClusterVersion`, but sure, I can change it.

mandre (commented):
/lgtm

dulek:
```
Error: unknown flag: --cert-rotation-scale
unknown flag: --cert-rotation-scale
```

I'm fairly sure HyperShift failures are not fault of this patch. @dcbw, could you consider overriding it?

PR #2099, opened 2023-11-08
[release-4.12] OCPBUGS-24039: remove all managed fields used by old manager

jluhrsen:
Previous to 4.11, we used client side update operations with a manager called 'cluster-network-operator', however following 4.11, we changed the field manager name

field because the depreciated field manager owned and preserved the field.

This patch removes the depreciated field manager
regardless of the operation it used previously.


jluhrsen:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.12-upgrade-from-stable-4.11-e2e-gcp-ovn-upgrade 10

pkg/apply/fieldmanager.go
```diff
@@ -29,7 +33,7 @@ func mergeManager(ctx context.Context, clusterClient client.ClusterClient, us *u
 			continue
 		}
 		if m.Operation == metav1.ManagedFieldsOperationUpdate {
-			patch, err = csaupgrade.UpgradeManagedFieldsPatch(us, map[string]sets.Empty{fromManager: {}}, intoManager)
+			patch, err = upgradeManagedFieldsPatchLocal(us, map[string]struct{}{fromManager: {}}, intoManager)
```
jcaamano, line 36: nit: 
I think that if you just set up a csaupgrade sub-package here that is the exact same thing that you are copying with a comment where you copied from would work better. I thiunk the comment at least is nice to have.

jcaamano (commented):
@jluhrsen don't you need to remove the vendored code?

jluhrsen:
> @jluhrsen don't you need to remove the vendored code?

yeah. I mean, it wasn't being used so it was just garbage. I think my newest update addresses this but the way I did it makes the github review look ugly.

I did a straight cherry-pick of the original "Deprecate legacy field manager" which brings in that vendored code. the next commit removes it and makes the the local copy.

that ok or some better way to do it?

pkg/apply/fieldmanager.go
jluhrsen: ok, I moved it all to a new local sub-module (csaupgrade.go) in ./pkg/apply/ I added some comments to explain at the top of the file. And the commit message is there too for anyone in the future trying to understand what's going on.

jcaamano:
> > @jluhrsen don't you need to remove the vendored code?
> 
> yeah. I mean, it wasn't being used so it was just garbage. I think my newest update addresses this but the way I did it makes the github review look ugly.
> 
> I did a straight cherry-pick of the original "Deprecate legacy field manager" which brings in that vendored code. the next commit removes it and makes the the local copy.
> 
> that ok or some better way to do it?

It's fine I think, only that github interprets is a moved file, which is probably better because it lets me compare it and it becomes apparent that there are more changes to the imports than needed. Except the `sets` import, was any other import change needed?

pkg/apply/csaupgrade.go
```diff
@@ -0,0 +1,250 @@
...
+	"fmt"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"reflect"
+	"sigs.k8s.io/structured-merge-diff/v4/fieldpath"
+)
+
+/*
+this sub-package is taken from k8s.io/client-go/util/csaupgrade
```
jcaamano, line 14: I wouldn't say sub-package here to avoid confusion, the way you did it is just this single file.

jluhrsen: removed

jluhrsen:
> > > @jluhrsen don't you need to remove the vendored code?
> > 
> > 
> > yeah. I mean, it wasn't being used so it was just garbage. I think my newest update addresses this but the way I did it makes the github review look ugly.
> > I did a straight cherry-pick of the original "Deprecate legacy field manager" which brings in that vendored code. the next commit removes it and makes the the local copy.
> > that ok or some better way to do it?
> 
> It's fine I think, only that github interprets is a moved file, which is probably better because it lets me compare it and it becomes apparent that there are more changes to the imports than needed. Except the `sets` import, was any other import change needed?

my bad. my IDE put them as a separate import section and I didn't noticed. it's cleaned up now

sdodson:
/test e2e-aws-ovn-network-migration
Maybe we get lucky, proposing override in channel though.

PR #2112, opened 2023-11-16
[release-4.14] OCPBUGS-24037: remove all managed fields used by old manager

jluhrsen:
the old manager "cluster-network-operator" was changed to "cluster-network-operator/operconfig" in 4.11 when server-side-apply was migrated to from client-side-apply. However, the old operator still had it's own managed fields and some interaction between those and the change to remove preStop hooks in ovnkube-master's daemonset containers was causing upgrades to stick in the network operator.

this will just remove the old manager (and it's managed fields) entirely. it's done before the update to apply the object instead of after.

JIRA: https://issues.redhat.com/browse/OCPBUGS-22293

deal w/ deprecated field manager before apply

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -125,29 +125,37 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
 		log.Printf("could not encode %s for apply", objDesc)
 		return fmt.Errorf("could not encode for patching: %w", err)
 	}
-	us, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions, subresources...)
+	// consider removing in OCP 4.18 when we know field manager 'cluster-network-operator' no longer possibly
+	// exists in any object from all upgrade paths
+	// Retrieve the current state of the resource
+	if isDepFieldManagerCleanupNeeded(subcontroller) {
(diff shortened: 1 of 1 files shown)


pkg/apply/apply.go
```diff
@@ -125,29 +125,37 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
...
+
+			if doesManagerOpExist(us.GetManagedFields(), depreciatedFieldManager, metav1.ManagedFieldsOperationUpdate,
+				metav1.ManagedFieldsOperationApply) {
+
+				us.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+				if err = mergeManager(ctx, clusterClient, us, depreciatedFieldManager, fieldManager, rm.Resource); err != nil {
+					klog.Errorf("Failed to merge field managers %q for object %q %s %s: %v", depreciatedFieldManager,
+						gvk.String(), obj.GetNamespace(), obj.GetName(), err)
+				} else {
+					klog.Infof("Depreciated field manager %s for object %q %s %s", depreciatedFieldManager,
```
rbbratta, line 147: typo `s/Depreciated/Deprecated/`

rbbratta: err, maybe it's an old typo.

sdodson:
Based on https://github.com/openshift/cluster-network-operator/pull/2112#issuecomment-1881969953 i'm going to add all required to merge labels.

jluhrsen:
the e2e-vsphere-ovn-windows job has a pass rate under 50% so it may take a few retests for it to pass

PR #2113, opened 2023-11-16
OCPBUGS-22869: hypershift, hosted clusters: enable multi-homing and multi-net features

maiqueb:
On HyperShift hosted cluster we were not enabling the multi-net (or multi-net policy) features; as such, the network attachment definition informers were not started, which caused multi-homed pods from being started in the hosted cluster.

Also enabled the multi-net policy feature depending on the operator configuration, since it was also missing from the configuration.


dougbtv:
just a couple conditional configs, I think I'm ok to issue a

PR #2114, opened 2023-11-16
OCPBUGS-24036: remove all managed fields used by old manager

jluhrsen:
the old manager "cluster-network-operator" was changed to "cluster-network-operator/operconfig" in 4.11 when server-side-apply was migrated to from client-side-apply. However, the old operator still had it's own managed fields and some interaction between those and the change to remove preStop hooks in ovnkube-master's daemonset containers was causing upgrades to stick in the network operator.

this will just remove the old manager (and it's managed fields) entirely. it's done before the update to apply the object instead of after.

JIRA: https://issues.redhat.com/browse/OCPBUGS-22293

deal w/ deprecated field manager before apply

### DIFF
--- a/pkg/apply/apply.go
+++ b/pkg/apply/apply.go
@@ -125,29 +125,37 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
 		log.Printf("could not encode %s for apply", objDesc)
 		return fmt.Errorf("could not encode for patching: %w", err)
 	}
-	us, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions, subresources...)
+	// consider removing in OCP 4.18 when we know field manager 'cluster-network-operator' no longer possibly
+	// exists in any object from all upgrade paths
+	// Retrieve the current state of the resource
+	if isDepFieldManagerCleanupNeeded(subcontroller) {
+		us, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get current state of %s: %w", objDesc, err)
+		}
+		if us != nil {
+			us.SetGroupVersionKind(gvk)
+
+			if doesManagerOpExist(us.GetManagedFields(), depreciatedFieldManager, metav1.ManagedFieldsOperationUpdate,
+				metav1.ManagedFieldsOperationApply) {
+
+				us.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+				if err = mergeManager(ctx, clusterClient, us, depreciatedFieldManager, fieldManager, rm.Resource); err != nil {
+					klog.Errorf("Failed to merge field managers %q for object %q %s %s: %v", depreciatedFieldManager,
+						gvk.String(), obj.GetNamespace(), obj.GetName(), err)
+				} else {
+					klog.Infof("Depreciated field manager %s for object %q %s %s", depreciatedFieldManager,
+						gvk.String(), obj.GetNamespace(), obj.GetName())
+				}
+			}
+		}
+	}
+
+	_, err = clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions, subresources...)
 	if err != nil {
 		return fmt.Errorf("failed to apply / update %s: %w", objDesc, err)
 	}
 
-	// consider removing in OCP 4.18 when we know field manager 'cluster-network-operator' no longer possibly
-	// exists in any object from all upgrade paths
-	if isDepFieldManagerCleanupNeeded(subcontroller) {
-		us.SetGroupVersionKind(gvk)
-
-		if doesManagerOpExist(us.GetManagedFields(), depreciatedFieldManager, metav1.ManagedFieldsOperationUpdate,
-			metav1.ManagedFieldsOperationApply) {
-
-			us.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
-			if err = mergeManager(ctx, clusterClient, us, depreciatedFieldManager, fieldManager, rm.Resource); err != nil {
-				klog.Errorf("Failed to merge field managers %q for object %q %s %s: %v", depreciatedFieldManager,
-					gvk.String(), obj.GetNamespace(), obj.GetName(), err)
-			} else {
-				klog.Infof("Depreciated field manager %s for object %q %s %s", depreciatedFieldManager,
-					gvk.String(), obj.GetNamespace(), obj.GetName())
-			}
-		}
-	}
 	log.Printf("Apply / Create of %s was successful", objDesc)
 	return nil
 }


martinkennelly:
### For other reviewers
The bug attached to this PR failed when apply was performed by CNO and resulted in error `must specify a handler type`.
Located in kapi validation here: kubernetes/pkg/apis/core/validation/validation.go

Code block:
```
func validateHandler(handler commonHandler, fldPath *field.Path) field.ErrorList {
	numHandlers := 0
	allErrors := field.ErrorList{}
	if handler.Exec != nil {
		if numHandlers > 0 {
			allErrors = append(allErrors, field.Forbidden(fldPath.Child("exec"), "may not specify more than 1 handler type"))
		} else {
			numHandlers++
			allErrors = append(allErrors, validateExecAction(handler.Exec, fldPath.Child("exec"))...)
		}
	}
	if handler.HTTPGet != nil {
		if numHandlers > 0 {
			allErrors = append(allErrors, field.Forbidden(fldPath.Child("httpGet"), "may not specify more than 1 handler type"))
		} else {
			numHandlers++
			allErrors = append(allErrors, validateHTTPGetAction(handler.HTTPGet, fldPath.Child("httpGet"))...)
		}
	}
	if handler.TCPSocket != nil {
		if numHandlers > 0 {
			allErrors = append(allErrors, field.Forbidden(fldPath.Child("tcpSocket"), "may not specify more than 1 handler type"))
		} else {
			numHandlers++
			allErrors = append(allErrors, validateTCPSocketAction(handler.TCPSocket, fldPath.Child("tcpSocket"))...)
		}
	}
	if handler.GRPC != nil {
		if numHandlers > 0 {
			allErrors = append(allErrors, field.Forbidden(fldPath.Child("grpc"), "may not specify more than 1 handler type"))
		} else {
			numHandlers++
			allErrors = append(allErrors, validateGRPCAction(handler.GRPC, fldPath.Child("grpc"))...)
		}
	}
	if numHandlers == 0 {
		allErrors = append(allErrors, field.Required(fldPath, "must specify a handler type"))
	}
	return allErrors
}
```

That func above is ONLY called for poststart/prestop hooks when theyre not nil:
```
func validateLifecycle(lifecycle *core.Lifecycle, fldPath *field.Path) field.ErrorList {
	allErrs := field.ErrorList{}
	if lifecycle.PostStart != nil {
		allErrs = append(allErrs, validateHandler(handlerFromLifecycle(lifecycle.PostStart), fldPath.Child("postStart"))...)
	}
	if lifecycle.PreStop != nil {
		allErrs = append(allErrs, validateHandler(handlerFromLifecycle(lifecycle.PreStop), fldPath.Child("preStop"))...)
	}
	return allErrs
}
```
`PreStop` is defined but doesnt have a handler defined (i.e either exec or httpget or tcpsocket). This suggests the hander was removed but not the fields above it (`Prestop`).
I checked the MG in the case where upgrade stalled (must-gather.local.3228488807846621353).
I can see theres two managers and they own the same prestop fields.
Theres differences though that jamo pointed out yesterday in the managed field section for the preStop fields.
The "new" manager managed fields for prestop look like:
```
                  f:preStop:
                    f:exec:
                      f:command: {}
```
and old field manager:

```
                   f:preStop:
                    .: {}
                    f:exec: {}
```

Looks like api server bug "merging" the fields when using managed fields and multiple owners. Guessing, when the new manager drops field `command`, then api server removes it, but doesnt remove `prestop` & `prestop.exec`  entirely because `exec` is owned by the old manager but this is an invalid configuration and that fails validation.

So how do we move forward? 
We can investigate fixing the kapi validation bug but we need to work around it now for CUs because apply will fail and upgrade will be blocked and thats why Jamo has posted this patch.

Reviewing now.

pkg/apply/fieldmanager.go
```diff
@@ -47,11 +48,45 @@ func mergeManager(ctx context.Context, clusterClient client.ClusterClient, us *u
 			if err != nil {
 				return fmt.Errorf("failed to patch (type %s) for object %s %s: %v", patchType, objGVR.String(), us.GetName(), err)
 			}
+			// remove the old manager
```
martinkennelly, line 51: I dont understand why this line and the reset below is needed. 
The previous lines *should* remove the old manager.  
Can you paste here what the managed fields look like at this point following the patch?

jluhrsen: you are right. the changes you did to deprecate the old fieldmanager were enough and the extra I added
was overkill. However, your changes' code path wasn't executing because it was being called after the APPLY
and the APPLY is where the upgrade failed and got stuck. [Here](https://github.com/openshift/cluster-network-operator/pull/2136) is a PR that adds a GET before the APPLY and
if it sees an old manager, it will then execute your code path. I tested with it, and it works like we want.

martinkennelly:
I can replicate this with simple pods but also on OCP. I can confirm that the old field manager has invalid field manager configuration and that is why the apply cannot be applied. Its clear we need to remove the old field managers configuration before attempting to apply.

The root of the problem is the old field manager owns the `Prestop`/`PostStart` field and the `exec` field. The new manager drops ownership of all these fields which leaves just `Prestop`/`PostStart` + `exec` fields owned by the old manager. However, this is an invalid configuration. There must also own a handler field (i.e. `command` / `httpGet`, etc). This fails the api server validation phase and that why is why the apply is failing.

### To replicate on any OCP ver without having to go through many upgrade jumps
1. Unmanage CNO 
```
oc patch clusterversion version --type json -p '[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps","name":"network-operator","namespace":"openshift-network-operator","unmanaged":true}]}]'
```
2. Scale down CNO replicates to 0
```
oc edit deployment network-operator -n openshift-network-operator
```
Confirm no CNO pods are running.

3. We need to hack the managed fields to add in a `Prestop`/ `PostStop` hook + "bad" managed field config for the ovnkube-node daemonset
Save the artifact to file
```
oc -n openshift-ovn-kubernetes get daemonset ovnkube-node -o yaml > /tmp/ok-node
```

Add in a `postStart` hook. I added it to the ovn-acl-logging container. Doesnt matter which container you add it to.
```
    lifecycle:
        postStart:
           exec:
             command:
               - sh
```

```
oc apply -f $file --field-manager=another-field-manager
```

This will perform a clientside update to the artifact.
Save the artifact to file again and output the managed fields. Confirm the `postStop` hook is there.

```
oc -n openshift-ovn-kubernetes get daemonset ovnkube-node --show-managed-fields -o yaml > /tmp/ok-node
```

Do a server side apply using the existing field manager. We do this so the existing field manager also owns the `postStop` and its subfields:

```
oc apply -f /tmp/b --field-manager=cluster-network-operator/operconfig --server-side
```

Edit the file and change the managed field section the field manager "another-field-manager" for the `postStart`` from:
```
                f:postStart:
                    f:exec:
                      f:command: {}

```

to

```
                f:postStart:
                    .: {}
                    f:exec: {}
```

This mimics what we see in the MG for the old field manager. It doesnt own the `command` field.

Clientside apply this update using current field manager.

```
oc apply -f $file --field-manager=cluster-network-operator/oper
```

Confirm that manager `cluster-network-operator/operconfig` owns the `postStart`, `exec` and `command` fields.
Confirm that manager `another-field-manager` owns just the `postStart` and `exec` fields only.

4. Now scale the CNO back up and it should fail to apply and remove the `postStart` field.

pkg/apply/apply.go
```diff
@@ -125,14 +125,14 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
...
 	}
-	us, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Patch(ctx, name, types.ApplyPatchType, data, patchOptions, subresources...)
-	if err != nil {
-		return fmt.Errorf("failed to apply / update %s: %w", objDesc, err)
-	}
-
 	// consider removing in OCP 4.18 when we know field manager 'cluster-network-operator' no longer possibly
 	// exists in any object from all upgrade paths
-	if isDepFieldManagerCleanupNeeded(subcontroller) {
+	// Retrieve the current state of the resource
```
martinkennelly, line 130: nit: ideally id put in a comment why we needed to do this and what OCP version it can be removed.

jluhrsen: isn't that what's up above in lines 128-130? although correct me if I'm wrong, but I think this can be removed in 4.16 right? 4.15 will have these changes from GA onward and any cluster that eventually upgrades to 4.15 would have this resolved, so 4.16 and above wouldn't need it.

But, does it hurt? I suppose in the future another field manager name change could occur?

jluhrsen:
> /retest-required

@sdodson , the migration jobs are failing for eerily similar reasons and could possibly be related to this
PR. I am digging in to what's going on.

sdodson:
Is migration supported in 4.15 right now, I thought that was an aspirational goal for maybe 4.15.z? If not maybe these jobs should be made non blocking.

trozet:
@sdodson offline migration is already supported, live migration is what is being targeted for 4.15. These jobs are for offline migration.

trozet (approved):
/lgtm

trozet:
/hold

looks like something with the new server side apply is messing with the migration fields. @pliurh PTAL

I see in the CNO code its says:
```

func (r *ReconcileOperConfig) UpdateOperConfig(ctx context.Context, operConfig *operv1.Network) error {
	config := operConfig.DeepCopy()
	// Since ApplyObject uses server side apply operconfig controller
	// takes ownership of all fields set in operConfig.
	// It shouldn't own .Spec.Migration as it is not modifying it anywhere.
	// Setting the value to nil will ensure that the value of that field will stay unchanged (including the fieldManager).
	config.Spec.Migration = nil
```

but the rollback script is timing out at:
```
# Reset the spec.migration before we can set it to other value
oc patch Network.operator.openshift.io cluster --type='merge' --patch '{"spec":{"migration":null}}'
# Wait until CNO update the applied-cluster cm to the latest
timeout 60s bash <<EOT
until 
  ! oc get network -o yaml | grep migration  > /dev/null
do
  echo "migration field is not cleaned by CNO"
  sleep 3
done
EOT
```

I'm not sure what the expectation is here. @jluhrsen reports some jobs it makes it past this stage. I'm not sure if removing the old field manager stops us from updating the applied value, or if something is just taking longer now.

jluhrsen:
> I'm not sure what the expectation is here. @jluhrsen reports some jobs it makes it past this stage. I'm not sure if removing the old field manager stops us from updating the applied value, or if something is just taking longer now.

yeah, some jobs make it past this point. I get the sense that something about this PR is causing some slowness
that is not there otherwise. Increasing the timeout from 60s to 120s in some steps helped a little. I think the cluster
may not be fully stable (clusteroperators ready) when the scripts are moving on to the next steps.

also, this shouldn't even be executing the code to remove the old field manager, because there is no old field
manager in the clusters these jobs run. Not sure what's going on...

sdodson:
Perhaps patch instead of SSA should be considered? Similar to this change? https://github.com/kubernetes/kubernetes/pull/121103

pliurh:
@trozet In live migration, we use the `clusterconfig-controller` to set and remove the `spec.migration` field of network.operator. We don't want the `operconfig-controller` to set this field in `UpdateOperConfig`. Otherwise there will be more than one controller owns this field. Then it cannot be removed by either of the owners with the server-side apply. However, in the existing offline migration job, we use the client-side apply (kubectl) to remove the `spec.migration` field of network.operator. So it will not be subject to the `managedFields`.

jluhrsen:
I've updated this PR with one small change to move the `Get()` to only happen if `isDepFieldManagerCleanupNeeded(subcontroller)` is true. I run multiple tests with a TEST PR over the weekend
and the network-migration jobs are much more stable. I'm not sure why this works, but it is a little more efficient
in the end. Otherwise, any time `ApplyObject()` was getting called this `Get()` would also be called whether it was needed
or not.

pkg/apply/apply.go
```diff
@@ -125,29 +125,37 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
...
-			metav1.ManagedFieldsOperationApply) {
-
-			us.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
-			if err = mergeManager(ctx, clusterClient, us, depreciatedFieldManager, fieldManager, rm.Resource); err != nil {
-				klog.Errorf("Failed to merge field managers %q for object %q %s %s: %v", depreciatedFieldManager,
-					gvk.String(), obj.GetNamespace(), obj.GetName(), err)
-			} else {
-				klog.Infof("Depreciated field manager %s for object %q %s %s", depreciatedFieldManager,
-					gvk.String(), obj.GetNamespace(), obj.GetName())
+		us, err := clusterClient.Dynamic().Resource(rm.Resource).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
```
martinkennelly, line 132: i was think you could do all this once and when its successful, never do it again for the lifetime of the operator, this would reduce the api calls. Up to you though. Not hard req for me.

trozet (approved):
I have no idea how the Get was causing that much delay...but it seems to have fixed the problem somehow...

/hold cancel
/lgtm

sdodson:
/retitle OCPBUGS-24691: remove all managed fields used by old manager
Move to the 4.16.0 bug after we branched and now master is 4.16

jluhrsen:
> /cherry-pick release-4.15

@sdodson , thanks for doing this. just FYI that I already have [4.14](https://github.com/openshift/cluster-network-operator/pull/2112), [4.13](https://github.com/openshift/cluster-network-operator/pull/2094) and [4.12](https://github.com/openshift/cluster-network-operator/pull/2099) PRs up and hoping to get those all green with CI so maybe things will move faster.

PR #2119, opened 2023-11-20
Added a configmap and controller for cron schedule

nicklesimba:
Added a configmap with "reconciler_cron_file" key value pair. Also added a controller so that if the user changes the value for reconciler_cron_file, it updates instead of getting clobbered/overwritten by the default value. The value of reconciler_cron_file (i.e. "30 4 * * *") will be put into a file of the same name on whereabouts ip-reconciler pods, such that when the reconciler is enabled on openshift clusters, the file should become available.

This change is meant to be used in conjunction with later whereabouts changes. In the future, whereabouts will search for the reconciler_cron_file file, and if present, it will use the cron schedule from the file. If the file is not present (for example, on an upstream, non-openshift cluster), a default value from the flatfile will be used instead.


zshi-redhat:
@nicklesimba could you fix the ci/prow/unit test error please?

bindata/cronjob-scheduler/config/configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
dougbtv, line 1: I think we want all these files in this dir in the bindata/network/multus folder. Put them all on the same folder depth there and just number them.

bindata/cronjob-scheduler/daemonset/daemonset.yaml
```diff
@@ -0,0 +1,56 @@
+---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: cron-scheduler-configmap-ds
```
dougbtv, line 5: Do we need this file at all? Maybe I don't get it, but... Don't we just modify the deployment for the reconciler?

bindata/network/multus/005-cron-schedule-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{.CronSchedulerConfigmap}}
+  namespace: openshift-multus
+  annotations:
+    kubernetes.io/description: |
+      Configmap containing user customizable cronjob schedule
```
dougbtv, line 8: Nit, update to: 

> Configmap containing user customizable schedule for the Whereabouts ip reconciler using a cron expression

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -37,8 +37,8 @@ import (
 const (
 	allowlistDsName      = "cni-sysctl-allowlist-ds"
 	allowlistAnnotation  = "app=cni-sysctl-allowlist-ds"
-	manifestDir          = "../../bindata/allowlist/daemonset"
-	allowlistManifestDir = "../../bindata/network/multus/004-sysctl-configmap.yaml"
+	manifestDir          = "bindata/allowlist/daemonset"
```
dougbtv, line 40: why are we referencing the allowlist here?

pkg/controller/cronjob_scheduler/cronjob_scheduler_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	manifestDir              = "bindata/cronjob-scheduler/daemonset"
+	cronConfigmapDsName      = "cron-scheduler-configmap-ds"
+	cronConfigmapManifestDir = "bindata/cronjob-scheduler/config/configmap.yaml"
```
dougbtv, line 40: Hrmmm, ok I see now that you're referencing this here. Now I'm not sure it the files from that dir should be in the bindata/network/multus because those get rendered, and you want to use it here. Now I think I get it.

I'm now not sure where files like this should go. Can you find another examples of something else in the CNO that references files like this and see where they're stored?

bindata/cronjob-scheduler/config/configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: cron-scheduler-configmap
```
dougbtv, line 4: Maybe we should make sure this is associated with whereabouts, maybe...

> whereabouts-reconciler-cron-scheduler-configmap

pkg/names/names.go
```diff
@@ -35,6 +35,11 @@ const MULTUS_NAMESPACE = "openshift-multus"
 // ALLOWLIST_CONFIG_NAME is the name of the allowlist ConfigMap
 const ALLOWLIST_CONFIG_NAME = "cni-sysctl-allowlist"
 
+// CRON_SCHEDULER_CONFIG_NAME is the name of the cronjob schedule configmap
+const CRON_SCHEDULER_CONFIG_NAME = "cron-scheduler-configmap"
+
+const USER_EDITABLE_CRON_SCHEDULER_CONFIGMAP = "cron-scheduler-configmap"
```
dougbtv, line 41: If these values are the same, do we need both?

dougbtv (changes requested):

pkg/controller/cronjob_scheduler/cronjob_scheduler_controller.go
nicklesimba: I based it off the allowlist, so my other example would be `bindata/allowlist/config/configmap.yaml`.

bindata/cronjob-scheduler/config/configmap.yaml
nicklesimba: I like your suggestion. I'm going with `whereabouts-reconciler-scheduler-configmap`. Not exactly a name that rolls off the tongue but it'll do.

bindata/cronjob-scheduler/daemonset/daemonset.yaml
nicklesimba: I think I was just copying it because the allowlist had something similar, but I think you're right - the deployment in multus.yaml should be good enough.

pkg/names/names.go
nicklesimba: We don't, good catch.

bindata/network/multus/005-cron-schedule-configmap.yaml
dougbtv: Can you update this to include that it's for a whereabouts reconciler cron? thanks!

bindata/cronjob-scheduler/daemonset/daemonset.yaml
dougbtv: Let's remove this

bindata/cronjob-scheduler/config/configmap.yaml
dougbtv: Yep, sounds good, let's update this one!

bindata/allowlist/daemonset/daemonset.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
     spec:
       containers:
-        - name: kube-multus-additional-cni-plugins
+        - name: cni-sysctl-allowlist-pod
```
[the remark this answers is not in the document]
dougbtv: Yep, revert this file, thanks

.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
```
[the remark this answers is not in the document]
dougbtv: again, remove this file

dougbtv:
closed in favor of #2209

PR #2131, opened 2023-11-27
CFE-888: Enable DNSNameResolver feature-gate 

chiragkyal:
Enable `DNSNameResolver` feature-gate for ovn-kubernetes 

Related to EP : https://github.com/openshift/enhancements/pull/1335

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -152,6 +152,15 @@ rules:
   - egressqoses/status
   verbs:
   - patch
+{{- if .DNS_NAME_RESOLVER_ENABLE }}
+- apiGroups: ["network.openshift.io"]
+  resources:
+  - dnsnameresolvers
+  verbs:
+  - get
+  - list
+  - watch
+{{- end }}
 {{- if .OVN_ADMIN_NETWORK_POLICY_ENABLE }}
 - apiGroups: ["policy.networking.k8s.io"]
   resources:
--- a/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
@@ -101,6 +101,19 @@ rules:
     - egressservices/status
   verbs:
     - update
+{{- if .DNS_NAME_RESOLVER_ENABLE }}
+- apiGroups: ["network.openshift.io"]
+  resources:
+  - dnsnameresolvers
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+{{- end }}
 - apiGroups: ["cloud.network.openshift.io"]
   resources:
   - cloudprivateipconfigs
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -557,6 +557,11 @@ data:
         admin_network_policy_enabled_flag="--enable-admin-network-policy"
       fi
 
+      dns_name_resolver_enabled_flag=
+      if [[ "{{.DNS_NAME_RESOLVER_ENABLE}}" == "true" ]]; then
+        dns_name_resolver_enabled_flag="--enable-dns-name-resolver"
+      fi
+
       # If IP Forwarding mode is global set it in the host here.
       ip_forwarding_flag=
       if [ "{{.IP_FORWARDING_MODE}}" == "Global" ]; then
@@ -606,6 +611,7 @@ data:
         ${multi_network_enabled_flag} \
         ${multi_network_policy_enabled_flag} \
         ${admin_network_policy_enabled_flag} \
+        ${dns_name_resolver_enabled_flag} \
         --enable-multicast \
         --zone ${K8S_NODE} \
         --enable-interconnect \
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ghodss/yaml"
 	. "github.com/onsi/gomega"
 	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -91,7 +92,10 @@ func TestRenderOVNKubernetes(t *testing.T) {
 			},
 		},
 	}
-	featureGatesCNO := featuregates.NewFeatureGate([]configv1.FeatureGateName{configv1.FeatureGateAdminNetworkPolicy}, []configv1.FeatureGateName{})
+	featureGatesCNO := featuregates.NewFeatureGate(
+		[]configv1.FeatureGateName{configv1.FeatureGateAdminNetworkPolicy, configv1.FeatureGateDNSNameResolver},
+		[]configv1.FeatureGateName{},
+	)
 	fakeClient := cnofake.NewFakeClient()
 
 	objs, _, err := renderOVNKubernetes(config, bootstrapResult, manifestDirOvn, fakeClient, featureGatesCNO)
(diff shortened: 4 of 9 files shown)


bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
```diff
@@ -150,6 +150,19 @@ rules:
...
+  resources:
+  - dnsnameresolvers
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
npinaeva, line 164: we'll need to split these rights between node and control pane once we finalize the ovn-k implementation. It is likely to be node can only read, and control-plane can create/update/delete

bindata/network/ovn-kubernetes/common/003-rbac-controller.yaml
```diff
@@ -95,6 +95,19 @@ rules:
   - patch
   - update
   - watch
+{{- if .DNS_NAME_RESOLVER_ENABLE }}
```
npinaeva, line 98: we removed this file, so on rebase you can just skip these changes

pkg/network/ovn_kubernetes_test.go
```diff
@@ -189,7 +199,7 @@ func TestRenderedOVNKubernetesConfig(t *testing.T) {
 		disableGRO               bool
 		disableMultiNet          bool
 		enableMultiNetPolicies   bool
-		enableAdminNetPolicies   bool
+		featureGates             configv1.CustomFeatureGates
```
npinaeva, line 202: I am thinking, maybe to make less changes in the future, we could replace this parameter with `enableFeatureGates []FeatureGateName` and then build a real config by iterating over all known feature gates and adding the ones that are listed to `Enabled` and others to `Disabled`. In that case you don't need to change tests that don't enable any feature gates, and only make changes when a new feature gate should be enabled. What do you think?

npinaeva (commented):
generally looks good, we just need to wait for ovn-k PR to be merged to finalize it

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
chiragkyal: Sure, for now, I've dropped create, delete, patch and update permissions from node. Let me know if it needs further refinement.

pkg/network/ovn_kubernetes_test.go
chiragkyal: Agreed, and that's a good suggestion. I've updated the structure, hope it looks cleaner now.

bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
```diff
@@ -71,6 +71,19 @@ rules:
   - patch
   - update
   - watch
+{{- if .DNS_NAME_RESOLVER_ENABLE }}
+- apiGroups: ["network.openshift.io"]
+  resources:
+  - dnsnameresolvers
+  verbs:
+  - create
```
kyrtapz, line 79: Does the control-plane really require all of these verbs?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -565,6 +565,11 @@ data:
         admin_network_policy_enabled_flag="--enable-admin-network-policy"
       fi
 
+      dns_name_resolver_enabled_flag=
```
kyrtapz, line 568: `start-ovnkube-node` is only used by ovnkube-node, this means that the `ovnkube-cluster-manager` won't rollout if `DNS_NAME_RESOLVER_ENABLE` changes.

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
```diff
@@ -156,6 +156,15 @@ rules:
     - get
     - list
     - watch
+{{- if .DNS_NAME_RESOLVER_ENABLE }}
+- apiGroups: ["network.openshift.io"]
+  resources:
+  - dnsnameresolvers
```
kyrtapz, line 162: Where is the `dnsnameresolvers` CRD applied?

pkg/network/ovn_kubernetes_test.go
```diff
@@ -819,7 +826,50 @@ logfile-maxage=0`,
 			controlPlaneReplicaCount: 2,
 			disableMultiNet:          true,
 			enableMultiNetPolicies:   true,
-			enableAdminNetPolicies:   true,
+			enabledFeatureGates:      []configv1.FeatureGateName{configv1.FeatureGateAdminNetworkPolicy},
```
kyrtapz, line 829: `configv1.FeatureGateDNSNameResolver` is not set in the featuregates here

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
npinaeva: I think it will, as it uses `004-config.yaml` file for features setup

bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
npinaeva: I think so, because cluster manager creates and deletes these objects (similar to cloudprivateipconfigs I suppose)

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
arkadeepsen: The CRD will be created when along with the cluster-dns-operator: https://github.com/openshift/cluster-dns-operator/pull/394/files#diff-d5348dedaad45595f0bd40f1b4730932489203aa511b1432951f2bfad8ab8a79

pkg/network/ovn_kubernetes_test.go
arkadeepsen: I think this test is specific to ANP. The `enableAdminNetPolicies` field is changed to `enabledFeatureGates` for using the same field for different feature gates in different test cases. The test case for `DNSNameResolver` is added below.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
```
npinaeva, line 8: I think linter is unhappy because of this package, which I think requires a linter version update to make it work. Since we don't have much time to merge extra PRs, maybe we could replace this package with `"k8s.io/utils/strings/slices"` which is already used in the repo and provides a similar `Contains` function?

chiragkyal: I checked that the `Contains` function from `"k8s.io/utils/strings/slices"` expects `[]string` but `enabledFeatureGates` is of type  `[]configv1.FeatureGateName`, so it's not directly compatible here. Should we use a `map` instead of a slice? 


```go
enabledFeatureGates      map[configv1.FeatureGateName]bool
```

npinaeva: ah sorry for the wrong pointer, I would suggest `"k8s.io/apimachinery/pkg/util/sets"` in that case

chiragkyal: Okay, I've updated to use `"k8s.io/apimachinery/pkg/util/sets"`

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
kyrtapz: Lest consider a scenario in which the featuregate is enabled during cluster runtime(e.g cluster is moved to techpreview):
1. FeatureGate gets enabled
2. CNO restarts
3. CNO renders the new 008-script-lib.yaml and 004-config.yaml confimaps
4. ovnkube-node daemonset is rolled out because it is annotated with the has of the script: https://github.com/openshift/cluster-network-operator/blob/6aa2f0fbf2b06ce52db1a2a039bbdbcaf588c409/pkg/network/ovn_kubernetes.go#L356-L378
5. ovnkube-controller is not rolled out and it doesn't reload the configmap dynamically on its own. 

does this make sense?

npinaeva: we agreed it is a bigger problem and created a card for it https://issues.redhat.com/browse/SDN-4832

arkadeepsen:
All other dev PRs are merged

/hold cancel

npinaeva:
manual CI-bot report
```
passed: (37.3s) 2024-05-16T08:39:04 "[sig-network][OCPFeatureGate:DNSNameResolver][Feature:EgressFirewall] when using openshift ovn-kubernetes should ensure egressfirewall with wildcard dns rules is created [Suite:openshift/conformance/parallel]"
```
/lgtm

npinaeva:
build failed :(
/retest-required

PR #2132, opened 2023-11-28
SDN-4308: Update status merge for APBRoute and EgressFirewall.

npinaeva:
Update APBRoute IP validation.
Depends on https://github.com/ovn-org/ovn-kubernetes/pull/3750

### DIFF
--- a/bindata/network/ovn-kubernetes/common/001-crd.yaml
+++ b/bindata/network/ovn-kubernetes/common/001-crd.yaml
@@ -163,8 +163,6 @@ spec:
                   x-kubernetes-list-type: set
                 status:
                   type: string
-              required:
-                - messages
               type: object
           required:
             - spec
@@ -725,9 +723,6 @@ spec:
                 description: A concise indication of whether the AdminPolicyBasedRoute
                   resource is applied with success
                 type: string
-            required:
-            - lastTransitionTime
-            - messages
             type: object
         required:
         - spec
--- a/bindata/network/ovn-kubernetes/common/001-crd.yaml
+++ b/bindata/network/ovn-kubernetes/common/001-crd.yaml
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.3
-  creationTimestamp: null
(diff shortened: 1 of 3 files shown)


bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -713,14 +720,14 @@ spec:
...
                 type: array
+                x-kubernetes-list-type: set
               status:
                 description: A concise indication of whether the AdminPolicyBasedRoute
                   resource is applied with success
                 type: string
             required:
             - lastTransitionTime
             - messages
-            - status
```
jcaamano, line 723: I talked with @npinaeva about this being a backward incompatible change. Need to confirm if we are ok with this.

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
```diff
@@ -138,6 +138,7 @@ rules:
   resources:
   - adminpolicybasedexternalroutes
   - egressfirewalls
+  - egressfirewalls/status
```
jcaamano, line 141: Why did you need this and not  `adminpolicybasedexternalroutes/status` as well?

bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
```diff
@@ -61,6 +61,10 @@ rules:
   - egressips
   - egressservices
   - egressservices/status
+  - adminpolicybasedexternalroutes
+  - adminpolicybasedexternalroutes/status
+  - egressfirewalls
+  - egressfirewalls/status
```
jcaamano, line 67: I guess we are following the status quo here but we should really only need update perms on `adminpolicybasedexternalroutes/status` and `egressfirewalls/status`

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
npinaeva: I actually need, but it should've already been here, because node updates status already :confused:

npinaeva: created a separate PR for that, since it is a different problem and should be backported to 4.14 https://github.com/openshift/cluster-network-operator/pull/2139

trozet (approved):

PR #2139, opened 2023-11-30
OCPBUGS-22710: Add apbroute/status patch rights for ovnkube-node to update status

npinaeva:
Looks like we have 0 tests for APBRoute in CNO, but if you deploy a 4.14 cluster and create APBRoute, you will find its status is empty and logs like
`
W1130 14:58:55.081710 5185 master_controller.go:227] Failed to update AdminPolicyBasedExternalRoutes default status: failed to update AdminPolicyBasedExternalRoutes default status: adminpolicybasedexternalroutes.k8s.ovn.org "default" is forbidden: User "system:ovn-node:ci-ln-f2400f2-72292-sgnts-master-2" cannot update resource "adminpolicybasedexternalroutes/status" in API group "k8s.ovn.org" at the cluster scope
`
Randomly found here https://github.com/openshift/cluster-network-operator/pull/2132/files#r1410561628
Must be related to https://issues.redhat.com/browse/OCPBUGS-22710

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -137,6 +137,7 @@ rules:
 - apiGroups: ["k8s.ovn.org"]
   resources:
   - adminpolicybasedexternalroutes
+  - adminpolicybasedexternalroutes/status
   - egressfirewalls
   - egressips
   - egressqoses


jordigilh:
Not here: not related to https://issues.redhat.com/browse/OCPBUGS-22710 as the referred bug was caused by a scenario where it failed to update the status struct when the`lastTransitionTime` field was nil, which is not allowed.
This case is due to the controller not having the correct RBAC permissions to write the status field.

PR #2141, opened 2023-11-30
OCPBUGS-24176: Ignore invalid PEM blocks

kyrtapz:
Previously, the CertificateData function would error out when unable to decode the next PEM block, causing issues with files containing empty lines at the end. 

This causes problems for HyperShift clusters with a management cluster that uses the `legacyVulnerableCABundle`:
https://github.com/openshift/service-ca-operator/blob/fe887613d031435fb0c76eb397764ba68fe09014/pkg/controller/cabundleinjector/starter.go#L30C5-L35
https://github.com/openshift/service-ca-operator/blob/fe887613d031435fb0c76eb397764ba68fe09014/pkg/controller/cabundleinjector/starter.go#L72

This change addresses the problem by using crypto.CertsFromPEM from library-go, which returns early when it cannot decode the next PEM block without throwing an error.

### DIFF
--- a/pkg/util/validation/trustbundle.go
+++ b/pkg/util/validation/trustbundle.go
@@ -2,17 +2,12 @@ package validation
 
 import (
 	"crypto/x509"
-	"encoding/pem"
 	"fmt"
 
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/library-go/pkg/crypto"
 
 	corev1 "k8s.io/api/core/v1"
-)
-
-const (
-	// certPEMBlock is the type taken from the preamble of a PEM-encoded structure.
-	certPEMBlock = "CERTIFICATE"
 )
 
 // TrustBundleConfigMap validates that ConfigMap contains a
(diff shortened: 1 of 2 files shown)


pkg/util/validation/trustbundle.go
```diff
@@ -27,53 +22,27 @@ func TrustBundleConfigMap(cfgMap *corev1.ConfigMap, caDataKey string) ([]*x509.C
 	if len(trustBundleData) == 0 {
 		return nil, nil, fmt.Errorf("data key %q is empty from ConfigMap %q", names.TRUSTED_CA_BUNDLE_CONFIGMAP_KEY, cfgMap.Name)
 	}
-	certBundle, _, err := CertificateData(trustBundleData)
+	certBundle, err := crypto.CertsFromPEM(trustBundleData)
```
trozet, line 25: "This change addresses the problem by using crypto.CertsFromPEM from library-go, which returns early when it cannot decode the next PEM block without throwing an error."

So it returns early and has no error? What if the first PEM block is the invalid one?

trozet (approved):
/lgtm

trozet (approved):
/approve

kyrtapz:
trying to fix tide..

/test verify

PR #2142, opened 2023-11-30
OCPBUGS-24299: network-node-identity mounts secrets with mode 0640

jonesbr17:
CNO managed component (network-node-identity) to conform to hypershift control plane expectations that all secrets should be mounted to not have global read. change from 420(0644) to 416(0640)


csrwng:
/retitle OCPBUGS-23397: network-node-identity mounts secrets with mode 0640

csrwng:
/retitle OCPBUGS-24299: network-node-identity mounts secrets with mode 0640

kyrtapz:
@jonesbr17 can you re-push the commit again(without any functional changes)? CI is confused with upgrade jobs and that should help with it.

PR #2147, opened 2023-12-05
OCPBUGS-19817: ipsec: add pluto restart

yuvalk:
we need to restart pluto after flushing xfrm state&policy
or else it loses all the previous state.

and we do want to flush, to get a "clean slate"

this fixes OCPBUGS-19817


yuvalk:
> @dcbw can we maybe override the failing required lanes IDK why they are failing, but they cant be affected by something inside the ipsec DS, as it's not even enabled in those lanes..
> 
> the important ipsec test ci/prow/e2e-ovn-ipsec-step-registry passed

nm, all required passed
please approve

PR #2101, opened 2023-11-09
SDN-4061: Remove interconnect upgrade logic, single-zone YAMLs and everything related to OVNK centralized architecture

ricky-rav:

PR #2149, opened 2023-12-06
SDN-4308,OCPBUGS-24650: Set enable-multi-external-gateway flag in ovnkube.conf

npinaeva:
to pass to both ovnkube-node and control-plane pods. There is no feature gate, so it is set unconditionally.


trozet (approved):
/lgtm

npinaeva:
/hold
till https://github.com/openshift/cluster-network-operator/pull/2101 is merged to change less files

trozet (approved):
/lgtm

PR #2148, opened 2023-12-05
OCPBUGS-24436: Add probes to node-network-identity

jonesbr17:
The node-network-identity deployment should contain liveness and readiness probes where applicable.

PR #2151, opened 2023-12-07
OCPBUGS-24601: Add minReadySeconds to network-node-identity

Joseph-Goergen:
The node-network-identity deployment should be set to assist in a controlled rollout of the microservice pods.

### DIFF
--- a/bindata/network/node-identity/managed/node-identity.yaml
+++ b/bindata/network/node-identity/managed/node-identity.yaml
@@ -13,6 +13,7 @@ metadata:
     hypershift.openshift.io/hosted-control-plane: {{.HostedClusterNamespace}}
     hypershift.openshift.io/control-plane: "true"
 spec:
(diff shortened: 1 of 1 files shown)


kyrtapz:
@Joseph-Goergen thanks for the PR.
Do you think this change is still required with https://github.com/openshift/cluster-network-operator/pull/2148/files#diff-9043a12081a351cc300d0683cbfb7f044c1629e6f6276e658eb19588f692702fR148?

Joseph-Goergen:
@kyrtapz I think it does. We typically set `minReadySeconds` to 15 with a probe(which includes an additional 15 seconds  from`initialDelaySeconds`) and 30 if there is no probe.

tssurya:
/retest-required
@Joseph-Goergen : Please don't backport this unless this is really critical. We have 0 backport policy unless there are good justifications to do so.

PR #2152, opened 2023-12-08
OCPBUGS-24633: ipsec add pluto restart

yuvalk:
we need to restart pluto after flushing xfrm state&policy or else it loses all the previous state.

and we do want to flush, to get a "clean slate"

(cherry picked from commit 2f5879349699b8477a2fcef9bc6d725e741c59c8)


trozet (approved):
/lgtm
/label backport-risk-assessed

PR #2153, opened 2023-12-08
OCPBUGS-24635: network node identity: tolarate all taints

kyrtapz:
The network node identity in self-hosted clusters should tolerate all taints, not only the common ones. 
This mimics the behavior of the API server pods: https://github.com/openshift/cluster-kube-apiserver-operator/blob/03c5a03d1bd0a952655aa180bd2708802e8cc1a4/bindata/assets/kube-apiserver/pod.yaml#L268-L269


kyrtapz:
The migration job is perma-failing and this PR doesn't affect any of the migration logic.
https://issues.redhat.com/browse/OCPBUGS-25863

@jcaamano can we override the failing job?

PR #2155, opened 2023-12-08
Ocpbugs 22293 4.15 optimized

jluhrsen:


jluhrsen:
I think this proved useful. I've updated the [main PR](https://github.com/openshift/cluster-network-operator/pull/2114) to be like this now and will go from there.

PR #2159, opened 2023-12-12
OCPBUGS-23729: Set replicas of Multus admission controller according to Hypershift ControllerAvailabilityPolicy

pliurh:


pliurh:
@enxebre PTAL

pliurh:
@zshi-redhat PTAL

kyrtapz:
/lgtm

This is the same approach we use for ovn-k:
https://github.com/openshift/cluster-network-operator/blob/f5757f9eabd0a109ed331a4663dc29ea38742e80/pkg/network/node_identity.go#L87-L90

PR #2160, opened 2023-12-12
OCPBUGS-23199: add env var in whereabouts-reconciler daemonset

zshi-redhat:
WHEREABOUTS_NAMESPACE env var is used by whereabouts-reconciler to garbage collect the Pod IPs. When unset, it defaults to kube-system namespace, while ippool.whereabouts.cni.cncf.io custom resource only exist in openshift-multus namespace on openshift. Setting WHEREABOUTS_NAMESPACE env var will tell whereabouts-reconciler where to get the ippool customer resource

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -726,11 +726,6 @@ spec:
           - >
             /usr/src/whereabouts/bin/entrypoint.sh -log-level debug
         image: {{.WhereaboutsImage}}
-        env:
-        - name: WHEREABOUTS_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
         resources:
           requests:
             cpu: "50m"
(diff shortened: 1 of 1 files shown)


maiqueb (approved):
/lgtm

Thank you !!! Good catch.

In upstream, we derive this variable from the daemonset namespace, but I guess this is simpler / more straight-forward.

maiqueb:
> /approve

:man_facepalming: I don't have approve powers on CNO. 

@dougbtv could you approve ?

bindata/network/multus/multus.yaml
```diff
@@ -747,6 +747,8 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: spec.nodeName
+        - name: WHEREABOUTS_NAMESPACE
+          value: "openshift-multus"
```
dougbtv, line 751: Thank you! Can we also remove the above duplicated `env` section @ https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus/multus.yaml#L729-L733 thanks

dougbtv (changes requested):

dougbtv (approved):

bindata/network/multus/multus.yaml
maiqueb: wow, really good catch.

I guess this was overridding the `env` stanza ...

PR #2161, opened 2023-12-13
OCPBUGS-25337: Avoid removal of ipsec-host daemonset when node is NotReady

pperiyasamy:
It's not reliable to check on ready machine count as it won't work in case of node reboot. Moreover it's enough to check for ipsec machine config extension in machine config pool status to confirm whether ipsec extension is completely rolled out or not. Hence this commit removes additional checks on ready machine count.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -1432,13 +1432,10 @@ func shouldUpdateOVNKonPrepull(ovn bootstrap.OVNBootstrapResult, releaseVersion 
 }
 
 // isUserIPsecMachineConfigPresent returns true if user owned MachineConfigs for IPsec plugin
-// are already present in master and worker nodes, otherwise returns false.
+// are already present either in master or worker nodes, otherwise returns false.
 func isUserIPsecMachineConfigPresent(infra bootstrap.InfraStatus) bool {
-	if infra.MasterIPsecMachineConfig == nil && infra.WorkerIPsecMachineConfig == nil {
-		return false
-	}
-	return !containsNetworkOwnerRef(infra.MasterIPsecMachineConfig.OwnerReferences) ||
-		!containsNetworkOwnerRef(infra.WorkerIPsecMachineConfig.OwnerReferences)
+	return (infra.MasterIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.MasterIPsecMachineConfig.OwnerReferences)) ||
+		(infra.WorkerIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.WorkerIPsecMachineConfig.OwnerReferences))
 }
 
 func containsNetworkOwnerRef(ownerRefs []metav1.OwnerReference) bool {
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1434,7 +1434,7 @@ func shouldUpdateOVNKonPrepull(ovn bootstrap.OVNBootstrapResult, releaseVersion
 // isUserIPsecMachineConfigPresent returns true if user owned MachineConfigs for IPsec plugin
 // are already present in master and worker nodes, otherwise returns false.
 func isUserIPsecMachineConfigPresent(infra bootstrap.InfraStatus) bool {
-	if infra.MasterIPsecMachineConfig == nil && infra.WorkerIPsecMachineConfig == nil {
+	if infra.MasterIPsecMachineConfig == nil || infra.WorkerIPsecMachineConfig == nil {
```
jcaamano, line 1437: Why this change?

pkg/network/ovn_kubernetes.go
```diff
@@ -1460,9 +1460,7 @@ func isIPsecMachineConfigActive(infra bootstrap.InfraStatus) bool {
 	}
 	ipSecPluginOnMasterNodes := hasSourceInMachineConfigStatus(infra.MasterMCPStatus, infra.MasterIPsecMachineConfig.Name)
 	ipSecPluginOnWorkerNodes := hasSourceInMachineConfigStatus(infra.WorkerMCPStatus, infra.WorkerIPsecMachineConfig.Name)
-	return infra.MasterMCPStatus.MachineCount == infra.MasterMCPStatus.ReadyMachineCount &&
-		infra.WorkerMCPStatus.MachineCount == infra.WorkerMCPStatus.ReadyMachineCount &&
-		ipSecPluginOnMasterNodes && ipSecPluginOnWorkerNodes
+	return ipSecPluginOnMasterNodes && ipSecPluginOnWorkerNodes
```
jcaamano, line 1463: But what does this mean for the normal N/S ipsec deployment? 
Is it ok if we rollout the N/S ipsec daemonset before the nodes have actually rebooted after applying the machine configs?
Shouldn't we, instead, track `isNSIPsecActive` and then where we consider `isIpsecMachineConfigActive` use `isIpsecMachineConfigActive || isNSIPsecActive` instead?

pkg/network/ovn_kubernetes.go
pperiyasamy: This is just to avoid nil pointer reference at line.no. 1440 when user rolled out ipsec MachineConfig only in master or worker roles and not on both simultaneously.

pkg/network/ovn_kubernetes.go
pperiyasamy: In the case of N/S IPsec deployment, only ipsec MachineConfigs are deployment with ipsec DaemonSets. so no problem here.
The MachineConfigStatus is updated with IPsec MachineConfig Name once rollout is complete on all nodes. So `isIpsecMachineConfigActive`  becomes true after MC installed on both master and workers nodes. Hence timing of ipsec DS rollout doesn't change.

pkg/network/ovn_kubernetes.go
jcaamano: It also changes the logic. It will return false if either master or worked MCs are not found while line 1440 will return true if any MC does not have the owner ref. Why not
```
return (infra.MasterIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.MasterIPsecMachineConfig.OwnerReferences)) ||
       (infra.WorkerIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.WorkerIPsecMachineConfig.OwnerReferences))
```

pperiyasamy: yes @jcaamano , done.

PR #2169, opened 2023-12-19
OCPBUGS-25669: Update ovn-k managed control-plane RBAC

npinaeva:
to use the same service account as self-hosted control-plane. Based on 20ec0344851a766e5fc6e3c394f93af2acd9ac6a ovn-kubernetes-controller should only be used for microshift (since single-zone doesn't exist anymore).
Microshift doesn't use this serviceAccount either, so delete it


npinaeva:
cc @ricky-rav @kyrtapz does it make sense? If we only use `ovn-kubernetes-controller` SA for microshift, should we rename/update it? e.g. I didn't add apbroute rights to it, because microshift doesn't enable that feature

kyrtapz:
/lgtm

ovnkube-controllplane in hypershift should use the same service account as we do for self hosted:
https://github.com/openshift/cluster-network-operator/blob/da0e780f23ad99fd5f9824a7421d9d7994d686cd/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L32

@ricky-rav @npinaeva I think we should consider renaming/refactoring/removing the `003-rbac-controller.yaml` to avoid this confusion in the future. 
AFAIR microshift is not even using the bindata files we have for it in CNO, is that correct @zshi-redhat?

zshi-redhat:
> /lgtm
> 
> ovnkube-controllplane in hypershift should use the same service account as we do for self hosted:
> 
> https://github.com/openshift/cluster-network-operator/blob/da0e780f23ad99fd5f9824a7421d9d7994d686cd/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L32
> 
> @ricky-rav @npinaeva I think we should consider renaming/refactoring/removing the `003-rbac-controller.yaml` to avoid this confusion in the future. AFAIR microshift is not even using the bindata files we have for it in CNO, is that correct @zshi-redhat?

@kyrtapz  you're right.  The 003-rbac-controller.yaml was never used by microshift.
The idea was that we update the CNO [bindata for microshift](https://github.com/openshift/cluster-network-operator/tree/master/bindata/network/ovn-kubernetes/microshift), then copy them to the [microshift repo](https://github.com/openshift/microshift/tree/main/assets/components/ovn), but we rarely followed it, most of the time, we just update the assets in microshift repo directly.

kyrtapz:
> > /lgtm
> > ovnkube-controllplane in hypershift should use the same service account as we do for self hosted:
> > https://github.com/openshift/cluster-network-operator/blob/da0e780f23ad99fd5f9824a7421d9d7994d686cd/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L32
> > 
> > @ricky-rav @npinaeva I think we should consider renaming/refactoring/removing the `003-rbac-controller.yaml` to avoid this confusion in the future. AFAIR microshift is not even using the bindata files we have for it in CNO, is that correct @zshi-redhat?
> 
> @kyrtapz you're right. The 003-rbac-controller.yaml was never used by microshift. The idea was that we update the CNO [bindata for microshift](https://github.com/openshift/cluster-network-operator/tree/master/bindata/network/ovn-kubernetes/microshift), then copy them to the [microshift repo](https://github.com/openshift/microshift/tree/main/assets/components/ovn), but we rarely followed it, most of the time, we just update the assets in microshift repo directly.

So it sounds like we diverged already and I don't see the value of keeping the microshift yamls in CNO. Can we remove it to avoid us having to try to update them every time we change something for other platforms? I doubt they work as is anyway, we don't run any tests for microshift on our side.

zshi-redhat:
> > > /lgtm
> > > ovnkube-controllplane in hypershift should use the same service account as we do for self hosted:
> > > https://github.com/openshift/cluster-network-operator/blob/da0e780f23ad99fd5f9824a7421d9d7994d686cd/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L32
> > > 
> > > @ricky-rav @npinaeva I think we should consider renaming/refactoring/removing the `003-rbac-controller.yaml` to avoid this confusion in the future. AFAIR microshift is not even using the bindata files we have for it in CNO, is that correct @zshi-redhat?
> > 
> > 
> > @kyrtapz you're right. The 003-rbac-controller.yaml was never used by microshift. The idea was that we update the CNO [bindata for microshift](https://github.com/openshift/cluster-network-operator/tree/master/bindata/network/ovn-kubernetes/microshift), then copy them to the [microshift repo](https://github.com/openshift/microshift/tree/main/assets/components/ovn), but we rarely followed it, most of the time, we just update the assets in microshift repo directly.
> 
> So it sounds like we diverged already and I don't see the value of keeping the microshift yamls in CNO. Can we remove it to avoid us having to try to update them every time we change something for other platforms? I doubt they work as is anyway, we don't run any tests for microshift on our side.

I'm fine with removing it if there is no other usage. @pliurh  wdyt?

ricky-rav:
/lgtm
Thanks guys for figuring out what else we can trim :)

pliurh:
I don't think we need to keep this manifest for Microshift either.

npinaeva:
(hopefully) fixed a unit test that checked stale ClusterRole

jcaamano:
What about this reference here?
https://github.com/openshift/cluster-network-operator/blob/975816eb56c09b51faab78f875815eb0d3ee4cb8/bindata/network/ovn-kubernetes/self-hosted/007-flowschema.yaml#L30-L33

npinaeva:
> What about this reference here?
> 
> https://github.com/openshift/cluster-network-operator/blob/975816eb56c09b51faab78f875815eb0d3ee4cb8/bindata/network/ovn-kubernetes/self-hosted/007-flowschema.yaml#L30-L33

oh wow how did I miss it, thank you Jaime!

PR #2170, opened 2023-12-19
OCPBUGS-24379: Remove egressip write permissions from ovn-kubernetes-node

kyrtapz:
ovn-kubernetes-node does not write anything to egressip and doesn't interact with cloudprivateipconfigs at all. This commit removes the unnecessary permissions.
This PR specifically handles the egressip resource, other features will be handled separately.


kyrtapz:
Verified manually on AWS and I was able to create and use egressIP without issues.
CI is not doing great recently.

ricky-rav:
/lgtm
I verified that (1) cloudprivateipconfigs are only used in cluster manager, so the ovnkube-node doesn't need any permissions for it; (2) egress ip is only retrieved, listed and watched in ovnkube-node, which doesn't need any update or patch permissions for this resource type.

PR #2171, opened 2023-12-19
OCPBUGS-24383: Limit write permissions for egressservices

kyrtapz:
Based on https://github.com/openshift/cluster-network-operator/pull/2170

ovn-kubernetes-node should only be allowed to read the egressservices resources. Additionally ovn-kubernetes-control-plane should only be allowed to update egressservices/status.
This PR specifically handles the egressservices resource, other features will be handled separately.

The only place we modify the egresservices resource is a status update here: https://github.com/ovn-org/ovn-kubernetes/blob/7447607048cf6fca57b0a51dd3361e694fea851d/go-controller/pkg/kube/kube.go#L456
Other than that both the ovn-kubernetes-node and ovn-kubernetes-control-plane read the egresservices from the API.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -140,7 +140,6 @@ rules:
   - adminpolicybasedexternalroutes/status
   - egressfirewalls
   - egressfirewalls/status
-  - egressips
   - egressqoses
   - egressservices
   - egressservices/status
@@ -150,6 +149,13 @@ rules:
   - patch
   - update
   - watch
+- apiGroups: ["k8s.ovn.org"]
+  resources:
+    - egressips
+  verbs:
+    - get
+    - list
+    - watch
 {{- if .OVN_ADMIN_NETWORK_POLICY_ENABLE }}
 - apiGroups: ["policy.networking.k8s.io"]
   resources:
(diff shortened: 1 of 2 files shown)


kyrtapz:
Verified manually, the status was properly set and the rules were defined in OVN:
```
apiVersion: v1
items:
- apiVersion: k8s.ovn.org/v1
  kind: EgressService
  metadata:
...
    name: agnhost
    namespace: default
...
  spec:
    sourceIPBy: LoadBalancerIP
  status:
    host: ip-10-0-59-74.ec2.internal


[root@ip-10-0-59-74 ~]# ovn-nbctl lr-policy-list ovn_cluster_router
Routing Policies
      1004 inport == "rtos-ip-10-0-59-74.ec2.internal" && ip4.dst == 10.0.59.74 /* ip-10-0-59-74.ec2.internal */         reroute                10.128.0.2
       102 (ip4.src == $a4548040316634674295 || ip4.src == $a13607449821398607916) && ip4.dst == $a14918748166599097711           allow               pkt_mark=1008
       102 ip4.src == 10.128.0.0/14 && ip4.dst == 10.128.0.0/14           allow
       102 ip4.src == 10.128.0.0/14 && ip4.dst == 100.64.0.0/16           allow
       101                             ip4.src == 10.128.2.13         reroute                10.128.0.2
       101                             ip4.src == 10.129.2.19         reroute                10.128.0.2
       101                             ip4.src == 10.131.0.20         reroute                10.128.0.2

```

ricky-rav:
/lgtm
I verified that only ovnk cluster-manager needs to call `Update` through `setEgressServiceHost` and that the logic in ovnkube-node only needs to list, watch and get egress services.

PR #2178, opened 2023-12-21
OCPBUGS-24176: Ignore invalid PEM blocks

kyrtapz:
Copy of https://github.com/openshift/cluster-network-operator/pull/2141 as I was unable to make the the tide happy...
@trozet please approve again.

Previously, the CertificateData function would error out when unable to decode the next PEM block, causing issues with files containing empty lines at the end.

This causes problems for HyperShift clusters with a management cluster that uses the legacyVulnerableCABundle:
https://github.com/openshift/service-ca-operator/blob/fe887613d031435fb0c76eb397764ba68fe09014/pkg/controller/cabundleinjector/starter.go#L30C5-L35
https://github.com/openshift/service-ca-operator/blob/fe887613d031435fb0c76eb397764ba68fe09014/pkg/controller/cabundleinjector/starter.go#L72

This change addresses the problem by using crypto.CertsFromPEM from library-go, which returns early when it cannot decode the next PEM block without throwing an error.

### DIFF
--- a/pkg/controller/proxyconfig/validation.go
+++ b/pkg/controller/proxyconfig/validation.go
@@ -15,6 +15,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/util/validation"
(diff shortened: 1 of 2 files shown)


trozet (approved):
/lgtm

PR #2179, opened 2023-12-22
OCPBUGS-25760: Keep the live migration annotation consistant with the enhancement doc

pliurh:
https://github.com/openshift/enhancements/blob/master/enhancements/network/sdn-live-migration.md#workflow-description


martinkennelly:
/test e2e-hypershift-ovn

Job maybe passing again. See some passes recently.

PR #2183, opened 2024-01-04
HOSTEDCP-1308: Add KAS access label to ovnkube-control-plane pod

Patryk-Stefanski:
Adding KAS access label to ovnkube-control-plane pod
Jira: [HOSTEDCP-1308](https://issues.redhat.com/browse/HOSTEDCP-1308)


kyrtapz:
@Patryk-Stefanski iirc ovnkube-control-plane doesn't access the mgmt cluster, why is that label required?

Patryk-Stefanski:
@kyrtapz when I removed the SA token mount from the pod the container started crashing witht the following error : ```F1212 12:18:13.705048 1 ovnkube.go:136] unable to create kubernetes rest config, err: TLS-secured apiservers require token/cert and CA certificate```

PR #2187, opened 2024-01-05
OCPBUGS-25207: Render IPsec MachineConfig after cluster settles with MachineConfigs

pperiyasamy:
The Machine Config Operator doesn't understand how to process IPsec MachineConfig as it's rolled out by network operator and that wasn't present in install manifest. This makes cluster operator "machine-config" to degraded state. Hence this commit waits for both master and worker machine config pool to be in ready state and then renders IPsec MachineConfig.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -529,7 +529,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 // When pre-4.14 cluster is Hypershift cluster running with ipsecConfig set, then just render ovn-ipsec-containerized daemonset as
 // MachineConfig kind is not supported there.
 // All Other cases are not supported in pre-4.14 deployments.
-func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootstrap.BootstrapResult) (renderIPsecMachineConfig, renderIPsecDaemonSet,
+func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootstrap.BootstrapResult) (renderCNOIPsecMachineConfig, renderIPsecDaemonSet,
 	renderIPsecOVN, renderIPsecHostDaemonSet, renderIPsecContainerizedDaemonSet, renderIPsecDaemonSetAsCreateWaitOnly bool) {
 	isHypershiftHostedCluster := bootstrapResult.Infra.HostedControlPlane != nil
 	isIPsecEnabled := conf.IPsecConfig != nil
@@ -542,6 +542,8 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	isUserIPsecMachineConfigPresent := isUserIPsecMachineConfigPresent(bootstrapResult.Infra)
 	isIpsecMachineConfigActive := isIPsecMachineConfigActive(bootstrapResult.Infra)
 	isIPsecMachineConfigNotActiveOnUpgrade := isIpsecUpgrade && !isIpsecMachineConfigActive && !isHypershiftHostedCluster
+	isMachineConfigClusterOperatorReady := bootstrapResult.Infra.MachineConfigClusterOperatorReady
+	isCNOIPsecMachineConfigPresent := isIPsecMachineConfigPresent(bootstrapResult.Infra) && !isUserIPsecMachineConfigPresent
 
 	// We render the ipsec deployment if IPsec is already active in OVN
 	// or if IPsec config is enabled.
@@ -561,7 +563,9 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// MachineConfig IPsec extensions are needed for the ipsec deployment except
 	// when the containerized deployment is used in hypershift hosted clusters.
 	// We will rollout unless the user has rolled out its own.
-	renderIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	renderCNOIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	// Wait for MCO to be ready unless we had already rendered the IPsec MachineConfig
+	renderCNOIPsecMachineConfig = renderCNOIPsecMachineConfig && (isCNOIPsecMachineConfigPresent || isMachineConfigClusterOperatorReady)
 
 	// We render OVN IPsec if IPsec is enabled or it's upgrade is in progress.
 	// If NS IPsec is enabled as well, we need to wait to IPsec MachineConfig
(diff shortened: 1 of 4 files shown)


yuvalk:
> The Machine Config Operator doesn't understand how to process IPsec MachineConfig as it's rolled out by network operator and that wasn't present in install manifest.
why?

pperiyasamy:
> > The Machine Config Operator doesn't understand how to process IPsec MachineConfig as it's rolled out by network operator and that wasn't present in install manifest.
> > why?
> 
> this sounds like a hack over MCO bug

@yuvalk This is done based on [comment](https://issues.redhat.com/browse/OCPBUGS-25207?focusedId=23727104&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23727104) from @yuqi-zhang. He may provide an explanation about the issue with MCO.

anuragthehatter:
This PR still shows following error

```
 - lastTransitionTime: "2024-01-05T21:10:50Z"
      message: 'Node ipsec-nstest87-kcpf9-master-0.c.openshift-qe.internal is reporting:
        "machineconfig.machineconfiguration.openshift.io \"rendered-master-0f5a6cc2aa59b61de071c13608aeaecf\"
        not found", Node ipsec-nstest87-kcpf9-master-1.c.openshift-qe.internal is
        reporting: "machineconfig.machineconfiguration.openshift.io \"rendered-master-0f5a6cc2aa59b61de071c13608aeaecf\"
        not found", Node ipsec-nstest87-kcpf9-master-2.c.openshift-qe.internal is
        reporting: "machineconfig.machineconfiguration.openshift.io \"rendered-master-0f5a6cc2aa59b61de071c13608aeaecf\"
        not found"'
      reason: 3 nodes are reporting degraded status on sync
      status: "True"

```

yuqi-zhang:
To elaborate on how the MCO determines this, what happens is that the bootstrap MCO renders the initial configuration to be served by the bootstrap MCS. Then when the in-cluster MCO takes over. It expects the same configuration at that time, otherwise it will error.

Now we could have the MCO just "force apply" whatever the latest is, but that would cause 2 reboots instead of 1 during install time, which is very undesireable for a lot of customers. You can think of this as an early warning system that things changed in the middle of your install process.

The best thing to do for situtations like this is to have the configuration rendered during bootstrap time, if it is meant to be part of the install configuration

anuragthehatter:
cluster install looks good with latest changes

/label qe-approved

yuqi-zhang:
> question is, why does it expects the same configuration?

A design decision from a long time ago, I think it's also that some configuration need to be install-time and not a quick re-apply after cluster comes up. We kept to allow us to be aware of drifts that can be happening and not ship with it

> well, if that's what the end-user wanted, why not do it?

I think we can consider making that an explicit flag that the user has to provide during install time, that they are ok with multiple reboots to apply configuration + some of the configs aren't actually present during install, only day-2 

> specifically for IPsec, we cant do that. because we rely on MCD to first apply the os-extension and so CNO is applying that MC

Ah, I think the current workaround of waiting for the cluster to finish install before attemping this as a "day-2" generation makes sense then

yuqi-zhang:
> sorry, I dont understand what are you trying to protect here?
> in any case, wouldn't it be better to reject those MCs and show an error?

The current error (rendered-master not found) is the rejection + error in a sense. I do think the error message can be more clear though

> the current solution (from this PR) will still incur a second reboot wave, triggered by CNO..

Right, this is explicitly determined as a day-2 configuration

> not sure why does that make more sense rather than fixing the issue in MCO

The current MCO design is an explicit design decision and not a bug. If we believe this behaviour should change, we can open an RFE and discuss there

pkg/network/ovn_kubernetes.go
```diff
@@ -562,6 +562,12 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// when the containerized deployment is used in hypershift hosted clusters.
 	// We will rollout unless the user has rolled out its own.
 	renderIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	// When IPsec MachineConfig is rendered for the first time, check machine config cluster operator is in ready state. This
+	// ensures cluster is settled with Machine Configs present during install time. If the operator is not ready, then do not
+	// render IPsec MachineConfig.
+	if renderIPsecMachineConfig && !isIPsecMachineConfigPresent(bootstrapResult.Infra) && !bootstrapResult.Infra.MachineConfigClusterOperatorReady {
```
jcaamano, line 568: Why could keep the flow that we already had going on here
```
...
isMachineConfigClusterOperatorReady := bootstrapResult.Infra.MachineConfigClusterOperatorReady
isUserIPsecMachineConfigPresent := isUserIPsecMachineConfigPresent(bootstrapResult.Infra)
isCNOIPsecMachineConfigPresent := isIPsecMachineConfigPresent(bootstrapResult.Infra) && !isUserIPsecMachineConfigPresent
isIpsecMachineConfigActive := isIPsecMachineConfigActive(bootstrapResult.Infra)
isIPsecMachineConfigNotActiveOnUpgrade := isIpsecUpgrade && !isIpsecMachineConfigActive && !isHypershiftHostedCluster
...
// Wait for MCO to be ready unless we had already rendered the IPsec MachineConfig
renderIPsecMachineConfig = renderIPsecMachineConfig && (isCNOIPsecMachineConfigPresent || isMachineConfigClusterOperatorReady)
```

Probably renaming `renderIPsecMachineConfig` to `renderCNOIPsecMachineConfig` helps with the understanding

pkg/network/ovn_kubernetes_test.go
```diff
@@ -2118,6 +2118,8 @@ func TestRenderOVNKubernetesEnableIPsec(t *testing.T) {
 	// bootstrap also represents current status
 	// the current cluster is single-stack and has version 1.9.9
 	bootstrapResult := fakeBootstrapResult()
+	bootstrapResult.Infra = bootstrap.InfraStatus{}
+	bootstrapResult.Infra.MachineConfigClusterOperatorReady = true
```
jcaamano, line 2122: Don't you want to test the transition of `MachineConfigClusterOperatorReady` from  `false` to `true` and then back from `true` to `false` as the machine config is applied?

pkg/network/ovn_kubernetes.go
pperiyasamy: yes, nice comment. done.

pkg/network/ovn_kubernetes_test.go
pperiyasamy: sure, added logic to cover mco state transitions.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -2204,7 +2231,7 @@ func TestRenderOVNKubernetesEnableIPsec(t *testing.T) {
 		t.Errorf("The MachineConfig %s must exist, but it's not available", workerMachineConfigIPsecExtName)
 	}
 
-	// At the 2nd pass, test render logic while MC extension rollout is in progress.
+	// At the 3rd pass, test render logic while MC extension rollout is in progress.
```
jcaamano, line 2234: I am not sure we need the 5th pass. Wouldn't MCO be progressing here on 3rd pass? So `MachineConfigClusterOperatorReady` would be false?

pperiyasamy: yes, i think so. removed 5th pass and checking it in 3rd pass itself.

PR #2191, opened 2024-01-09
SDN-4166: Enable N-S IPsec

yuvalk:
- Bump openshift/api module
- Remove check not allowing to change ipsecConfig
- Add the implementation to the new IPsec enablement API


pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -311,6 +311,16 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	// Fill all defaults explicitly
 	network.FillDefaults(&newOperConfig.Spec, prev, mtu)
 
+	//TO-REMOVE - 5 lines
```
JoshSalomon, line 314: Look for //To-Remove comments, these are lines I planned to remove, decided what you want to keep and in any case remove the comments.

yuvalk: done

pkg/network/ovn_kubernetes.go
```diff
@@ -561,12 +579,12 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// MachineConfig IPsec extensions are needed for the ipsec deployment except
```
JoshSalomon, line 579: I believe the comment here is misleading and should mention EW ipsec, this is what the code does.

pkg/network/ovn_kubernetes.go
```diff
@@ -561,12 +579,12 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
...
 	// when the containerized deployment is used in hypershift hosted clusters.
 	// We will rollout unless the user has rolled out its own.
-	renderIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	renderIPsecMachineConfig = (isIPsecEnabled || renderIPsecDaemonSet) && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
 
 	// We render OVN IPsec if IPsec is enabled or it's upgrade is in progress.
```
JoshSalomon, line 584: I believe the comment here is misleading and we should change it to 

// We render OVN IPsec if **EW** IPsec is enabled or it's upgrade is in progress.

Alternatively the code should be changed - (in line 587 it should be **isIPsecFull**

pkg/network/ovn_kubernetes.go
```diff
@@ -561,12 +579,12 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// MachineConfig IPsec extensions are needed for the ipsec deployment except
 	// when the containerized deployment is used in hypershift hosted clusters.
 	// We will rollout unless the user has rolled out its own.
-	renderIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	renderIPsecMachineConfig = (isIPsecEnabled || renderIPsecDaemonSet) && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
 
-	// We render OVN IPsec if IPsec is enabled or it's upgrade is in progress.
+	// We render OVN IPsec if E-W IPsec is enabled or it's upgrade is in progress.
```
kyrtapz, line 584: NIT: is it `EW` or `E-W`? do we need the acronym or could we just spell it out?

pkg/network/ovn_kubernetes.go
```diff
@@ -532,10 +534,26 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	if conf.IPsecConfig != nil {
+		if conf.IPsecConfig.Mode != "" {
+			mode = conf.IPsecConfig.Mode
+		} else {
+			mode = operv1.IPsecModeFull // BW compatibility with existing configs
+		}
+	}
+	isIPsecEnabled := mode != operv1.IPsecModeDisabled
+	isIPsecFull := mode == operv1.IPsecModeFull // Full mode is for NS+EW
+	klog.Infof("IPsec: after looking at %+v, ipsec mode=%s, isIPsecEnabled=%v", conf.IPsecConfig, mode, isIPsecEnabled)
```
kyrtapz, line 555: NIT: is there any value in logging the `isIPsecEnabled`? In most deployments `IPsecConfig` will be nil and I am not sure if we need that log message(looks more like a v5?)

kyrtapz (approved):
/lgtm
minor nits

kyrtapz:
We need to also adapt: https://github.com/openshift/cluster-network-operator/blob/c4a19c2e8691cc8e3dd81ac0de4f3e27ca5d3c55/pkg/network/ovn_kubernetes.go#L886

anuragthehatter:
Just to mention here based on testing this PR-
ipsecConfig `Disabled` mode doesn't disable NS IPsec 
Ref: https://issues.redhat.com/browse/OCPBUGS-26549

yuvalk:
> We need to also adapt:
> 
> https://github.com/openshift/cluster-network-operator/blob/c4a19c2e8691cc8e3dd81ac0de4f3e27ca5d3c55/pkg/network/ovn_kubernetes.go#L885-L887

hopefully I'm doing the right thing now

yuvalk:
> Just to mention here based on testing this PR- ipsecConfig `Disabled` mode doesn't disable NS IPsec Ref: https://issues.redhat.com/browse/OCPBUGS-26549

this one works for me :shrug: 
let's discuss offline

pkg/network/ovn_kubernetes.go
```diff
@@ -532,10 +548,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootstrap.BootstrapResult) (renderIPsecMachineConfig, renderIPsecDaemonSet,
 	renderIPsecOVN, renderIPsecHostDaemonSet, renderIPsecContainerizedDaemonSet, renderIPsecDaemonSetAsCreateWaitOnly bool) {
 	isHypershiftHostedCluster := bootstrapResult.Infra.HostedControlPlane != nil
-	isIPsecEnabled := conf.IPsecConfig != nil
 	isIpsecUpgrade := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.LegacyIPsecUpgrade
 	isOVNIPsecActive := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.OVNIPsecActive
 
+	// Find the IPsec mode from Ipsec.config
+	// Ipsec.config == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
```
jcaamano, line 555: I am curious on this however. I wouldn't expect the possibility of ipsecConfig==nil. Has that really been observed to happen with thew new API in place? Expecting it does not do any harm, I am just curious about it.

pkg/network/ovn_kubernetes.go
```diff
@@ -877,12 +906,12 @@ func validateOVNKubernetes(conf *operv1.NetworkSpec) []error {
...
 	return out
 }
-
 func getOVNEncapOverhead(conf *operv1.NetworkSpec) uint32 {
 	const geneveOverhead = 100
 	const ipsecOverhead = 46 // Transport mode, AES-GCM
 	var encapOverhead uint32 = geneveOverhead
-	if conf.DefaultNetwork.OVNKubernetesConfig.IPsecConfig != nil {
+	mode := getIPsecMode(conf.DefaultNetwork.OVNKubernetesConfig)
+	if mode == operv1.IPsecModeFull || mode == "" {
```
jcaamano, line 914: `mode == ""` is not possible?

pkg/network/ovn_kubernetes.go
```diff
@@ -532,10 +548,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
 	isOVNIPsecActive := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.OVNIPsecActive
 
+	// Find the IPsec mode from Ipsec.config
+	// Ipsec.config == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
+	// ipsecConfig.mode == "" (bw compatibility) || ipsec.Config == Full ==> ipsec is enabled for NS and EW
+	// ipsecConfig.mode == External ==> ipsec is enabled for NS only
+
+	mode := getIPsecMode(conf)
+	isIPsecEnabled := mode != operv1.IPsecModeDisabled
+	isIPsecFull := mode == operv1.IPsecModeFull // Full mode is for NS+EW
```
jcaamano, line 561: Would it work more intuitively to have these boolean instead?
```
isNSIPsecEnabled := mode != operv1.IPsecModeDisabled
isEWIPsecEnabled := mode == operv1.IPsecModeFull
```

pkg/network/ovn_kubernetes.go
```diff
@@ -532,10 +548,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	if conf.IPsecConfig != nil {
+		klog.Infof("IPsec: after looking at %+v, ipsec mode=%s, isIPsecEnabled=%v", conf.IPsecConfig, mode, isIPsecEnabled)
+	} else {
+		klog.V(5).Infof("IPsec: after looking at %+v, ipsec mode=%s, isIPsecEnabled=%v", conf.IPsecConfig, mode, isIPsecEnabled)
+	}
```
jcaamano, line 566: I guess that this has been useful for debugging but I would refine this comment a bit
```
if conf.IPsecConfig == nil || conf.IPsecConfig.Mode == "" { 
    klog.V(5).Infof("Assuming IPsec mode %s after IPsec config %+v", mode, conf.IPsecConfig)
}
```
It wouldn't harm if this lived in `getIPsecMode` at that log level.

pkg/network/ovn_kubernetes.go
JoshSalomon: The name isEWIPsecEnabled is misledding because it is NS+EW hence why we use Full and not EW

pkg/network/ovn_kubernetes.go
JoshSalomon: `mode == ""` is possible during upgrade (for bw compatibility, when ipsec was enabled with `ipsecConfig: {}` , but the user can't set `mode` to "" anymore

pkg/network/ovn_kubernetes.go
jcaamano: I don't think it is more misleading than using expressions like "...if EW IPsec config is enabled..." or "...We render OVN IPsec if East-West IPsec is enabled..." in the comments below. It is clearly assumed in the local logic of this method that the full IPsec mode is decomposed in NS and EW and defining the booleans in line with the comments just make things easier to understand I think.

We can try to tweak the variable names though. What about
```
enablesNSIPsec := mode != operv1.IPsecModeDisabled 
enablesEWIPsec := mode == operv1.IPsecModeFull
```

pkg/network/ovn_kubernetes.go
```diff
@@ -561,12 +585,12 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// MachineConfig IPsec extensions are needed for the ipsec deployment except
 	// when the containerized deployment is used in hypershift hosted clusters.
 	// We will rollout unless the user has rolled out its own.
-	renderIPsecMachineConfig = renderIPsecDaemonSet && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
+	renderIPsecMachineConfig = (isIPsecEnabled || renderIPsecDaemonSet) && !isUserIPsecMachineConfigPresent && !isHypershiftHostedCluster
```
jcaamano, line 588: I think the comment above (or some comment) needs to be updated to reflect the fact that we now now somehow manage NS IPsec because before we only managed in in combination with EW IPsec.

jcaamano: Like the comment above reads `MachineConfig IPsec extensions are needed for the ipsec deployment` but now is not only that since now we can enable NS IPsec with no ipsec deployment at all ....

pkg/network/ovn_kubernetes.go
JoshSalomon: This is for upgrade scenarios for a cluster which did not have ipsec enabled. It can not happen with the new API but it happens in upgrade situations.

yuvalk:
if we want to see the ipsec ci succeeding at least once, we should merge first https://github.com/openshift/release/pull/47486

pkg/network/ovn_kubernetes.go
JoshSalomon: indeed IIUC `getIPsecMode` can't return ""

pkg/network/ovn_kubernetes.go
pperiyasamy: +1, This would make `shouldRenderIPsec` method to be consistent with other state variables.

pperiyasamy:
Would you also update IPsec unit tests according to new API changes ?

yuvalk:
> Would you also update IPsec unit tests according to new API changes ?

fixed

yuvalk:
@jcaamano can you please take another look?
I think I've fixed all the comments you had (or at least I hope so)

pkg/network/ovn_kubernetes.go
yuvalk: fixed

PR #2192, opened 2024-01-11
OCPBUGS-26492: Not set CNO to degraded if API server returns conflict error

pliurh:
In CNO, the network.operator CR can be updated by both operconfig and clusterconfig controllers. There is a chance that these two controllers may encounter a racing condition, resulting in conflict error.

Normally, this type of error can be automatically recovered by retrying the operation. Therefore, we don't set CNO to degrade when encountering this error.


kyrtapz:
/lgtm
This issue is transient and I agree that we should not make CNO degraded on it.
Although the issue is not specific to live-migration it is most likely to happen in that scenario.
Using `RetryOnConflict` doesn't make a lot of sense here as we would be getting the latest object.

trozet (approved):
/approve

PR #2193, opened 2024-01-11
Remove unused code for IPsec upgrade

pperiyasamy:
The IPsec upgrade logic contains code to handle 4.13 upgrade path. But OCP only allow upgrades from one minor version to the next e.g. 4.y to 4.(y+1), As we're already in the 4.15 version, this code is no longer valid. Hence removing unused code from IPsec upgrade path.


ricky-rav:
@pperiyasamy Given that we're now in 4.17, do we need any special handling of ipsec in 4.16->4.17 upgrades?

pperiyasamy:
closing it in favor of https://github.com/openshift/cluster-network-operator/pull/2454.

PR #2197, opened 2024-01-12
OCPBUGS-24690: Disable network-node-identity on ROKS

kyrtapz:
In environments with external control plane topology, the API server is deployed out of cluster. 
This means that CNO cannot easily predict how to deploy and enforce the node identity webhook. 
IBMCloud uses an external control plane topology with Calico as the CNI for both HyperShift based ROKS deployments and IBM ROKS Toolkit based ROKS deployments.
/cc @dougbtv @s1061123

### DIFF
--- a/pkg/network/multus_admission_controller.go
+++ b/pkg/network/multus_admission_controller.go
@@ -50,7 +50,8 @@ func renderMultusAdmissonControllerConfig(manifestDir string, externalControlPla
 	objs := []*uns.Unstructured{}
 	var err error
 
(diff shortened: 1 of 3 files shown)


s1061123 (approved):

rtheis (changes requested):
/hold we need to talk about ROKS since the component is deploy and I'm not sure why we wouldn't want it to be deployed.

rtheis (approved):
/remove-hold
/lgtm

Thanks for being patient and working with us on these changes.

PR #2199, opened 2024-01-12
NO-JIRA: Fix jira component info

danwinship:
At some point someone added a `cluster-network-operator` subcomponent in jira, but never updated the OWNERS file here to reference it

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -18,6 +18,5 @@ approvers:
   - knobunc
   - trozet
 
-# Bugzilla info; "openshift-sdn" is wrong but there is no "generic networking" subcomponent
 component: Networking
-subcomponent: openshift-sdn
+subcomponent: cluster-network-operator


OWNERS
```diff
@@ -18,6 +18,5 @@ approvers:
   - knobunc
   - trozet
 
-# Bugzilla info; "openshift-sdn" is wrong but there is no "generic networking" subcomponent
 component: Networking
-subcomponent: openshift-sdn
+subcomponent: cluster-network-operator
```
tssurya, line 22: does this take effect in JIRA same way it used to in bugzilla?
If so don't we need to do Networking/cluster-network-operator or something? or are we sure this works..

tssurya:
@danwinship : seems like this need a jira bug to satisfy the bot? that is surprising considering this is against master branch, is that a new rule?

PR #2201, opened 2024-01-15
OCPBUGS-26952: Deploy CNO IPsec MC even if user already have one

pperiyasamy:
The CNO does not roll out its own IPsec Machine Config when user already installed Machine Config for IPsec. This leads to IPsec gets disabled temporarily in the data plane when user deletes their own Machine Config. IPsec would be in disabled state until operator rolls out its own IPsec machine config in the subsequent reconcile loop followed by nodes reboot. To avoid this situation, this commit always rolls out its own IPsec Machine Config at the time of IPsec fresh install or upgrade irrespective of user's Machine Config exists or not.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -39,6 +39,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -537,7 +538,7 @@ func getIPsecMode(conf *operv1.OVNKubernetesConfig) operv1.IPsecMode {
 
 // shouldRenderIPsec method ensures the have following IPsec states for upgrade path from 4.14 to 4.15 or later versions:
 // When 4.14 cluster is already installed with MachineConfig for IPsec extension and ipsecConfig is set in network operator
-// config (i.e. IPsec for NS+EW), then reuse the installed MC extension and render ipsec-host daemonset.
+// config (i.e. IPsec for NS+EW), then render CNO's IPsec MC extension and ipsec-host daemonset.
 // When 4.14 cluster is just running with ipsecConfig set in network operator config (i.e. IPsec for EW only), then activate
 // IPsec MachineConfig and render ipsec-host daemonset.
 // When 4.14 cluster is just installed with MachineConfig for IPsec extension (i.e. IPsec for NS only), then just keep MachineConfig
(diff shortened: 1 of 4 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1455,11 +1451,25 @@ func shouldUpdateOVNKonPrepull(ovn bootstrap.OVNBootstrapResult, releaseVersion
 	return true, false
 }
 
-// isUserIPsecMachineConfigPresent returns true if user owned MachineConfigs for IPsec plugin
-// are already present either in master or worker nodes, otherwise returns false.
-func isUserIPsecMachineConfigPresent(infra bootstrap.InfraStatus) bool {
-	return (infra.MasterIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.MasterIPsecMachineConfig.OwnerReferences)) ||
-		(infra.WorkerIPsecMachineConfig != nil && !containsNetworkOwnerRef(infra.WorkerIPsecMachineConfig.OwnerReferences))
+func isCNOIPsecMachineConfigExists(infra bootstrap.InfraStatus) bool {
```
jcaamano, line 1454: nit: `isCNOIPsecMachineConfigPresent`?

The doc text is always useful, why not keep it around since it was already there?

pkg/network/ovn_kubernetes.go
```diff
@@ -1455,11 +1451,25 @@ func shouldUpdateOVNKonPrepull(ovn bootstrap.OVNBootstrapResult, releaseVersion
...
+		return false
+	}
+	for _, machineConfig := range infra.WorkerIPsecMachineConfigs {
+		if containsNetworkOwnerRef(machineConfig.OwnerReferences) {
+			exists = true
+			break
+		}
+		exists = false
+	}
+	return exists
```
jcaamano, line 1472: you can just `return  true` here and just `return false` at the end

Or

```
func isCNOIPsecMachineConfigPresent(infra bootstrap.InfraStatus) bool {
        isCNOIPsecMachineConfigPresentIn := func(mcs [] infra.MasterIPsecMachineConfigs) bool {
            for _, mc := range mcs {
		    if containsNetworkOwnerRef(mc.OwnerReferences) {
			    return true
		    }
	    }
	    return false
        }
        return isCNOIPsecMachineConfigPresentIn(infra.MasterIPsecMachineConfigs) && isCNOIPsecMachineConfigPresentIn(infra.WorkerIPsecMachineConfigs)
}
```

jcaamano (commented):
@pperiyasamy 

can you elaborate why it is not a problem that the user and CNO machine configs are present at the same time?

yuvalk:
> @pperiyasamy
> 
> can you elaborate why it is not a problem that the user and CNO machine configs are present at the same time?

because MCO knows how to handle it properly.
if 2 MCs are trying to set the same systemd unit - you get only one
and for the extensions, I think they have a small problem there, because we get `ipsec` twice, but nevertheless it works fine.

the benefit of always having our own (CNO) managed MC
is that ipsec stays intact even if customer would want to delete their MC
which is what we expect to happen for customers who are using this feature in prior releases (where they had to have their own MC) and no upgrading to this release (and desire to let CNO manage it for them)

jcaamano:
> > @pperiyasamy
> > can you elaborate why it is not a problem that the user and CNO machine configs are present at the same time?
> 
> because MCO knows how to handle it properly. if 2 MCs are trying to set the same systemd unit - you get only one and for the extensions, I think they have a small problem there, because we get `ipsec` twice, but nevertheless it works fine.
> 
> the benefit of always having our own (CNO) managed MC is that ipsec stays intact even if customer would want to delete their MC which is what we expect to happen for customers who are using this feature in prior releases (where they had to have their own MC) and no upgrading to this release (and desire to let CNO manage it for them)

I discussed briefly with @pperiyasamy what if the user is using the same MCs names. I wouldn't discard that happening as a user might copy what is available in newer releases and apply it to old releases.

It might be worth to test & document somehow.

yuvalk:
> > > @pperiyasamy
> > > can you elaborate why it is not a problem that the user and CNO machine configs are present at the same time?
> > 
> > 
> > because MCO knows how to handle it properly. if 2 MCs are trying to set the same systemd unit - you get only one and for the extensions, I think they have a small problem there, because we get `ipsec` twice, but nevertheless it works fine.
> > the benefit of always having our own (CNO) managed MC is that ipsec stays intact even if customer would want to delete their MC which is what we expect to happen for customers who are using this feature in prior releases (where they had to have their own MC) and no upgrading to this release (and desire to let CNO manage it for them)
> 
> I discussed briefly with @pperiyasamy what if the user is using the same MCs names. I wouldn't discard that happening as a user might copy what is available in newer releases and apply it to old releases.
> 
> It might be worth to test & document somehow.

absolutely right
I think we should document that

pkg/network/ovn_kubernetes.go
pperiyasamy: done.

pkg/network/ovn_kubernetes.go
pperiyasamy: usage of closure `isCNOIPsecMachineConfigPresentIn` looks good. changed it.

pperiyasamy:
> > > > @pperiyasamy
> > > > can you elaborate why it is not a problem that the user and CNO machine configs are present at the same time?
> > > 
> > > 
> > > because MCO knows how to handle it properly. if 2 MCs are trying to set the same systemd unit - you get only one and for the extensions, I think they have a small problem there, because we get `ipsec` twice, but nevertheless it works fine.
> > > the benefit of always having our own (CNO) managed MC is that ipsec stays intact even if customer would want to delete their MC which is what we expect to happen for customers who are using this feature in prior releases (where they had to have their own MC) and no upgrading to this release (and desire to let CNO manage it for them)
> > 
> > 
> > I discussed briefly with @pperiyasamy what if the user is using the same MCs names. I wouldn't discard that happening as a user might copy what is available in newer releases and apply it to old releases.
> > It might be worth to test & document somehow.
> 
> absolutely right I think we should document that (because user MC might include extra content - that CNO will run over!)

Yes, The CNO takes the ownership of those IPsec MCs if user already have MCs with same name. So Let's document it so that user doesn't accidentally delete it.

PR #2207, opened 2024-01-17
OCPBUGS-27925,OCPBUGS-30579:[release-4.14] tighten conditions for the state transitions in IC upgrade

ricky-rav:
A few misc fixes for the IC upgrade logic:

- When evaluating whether phase 2 is ongoing, don't rely on ovnkube-node being "progressing", since that might happen also due to transient node conditions _after_ a complete roll out of ovnkube-node (found in https://issues.redhat.com//browse/OCPBUGS-30579)
- Tighten the progression to phase 2 to the presence of the configmap: since CNO removes the configmap as soon as the upgrade is completed, checking for the configmap prevents us from mistakenly going back to the multi-zone-interconnect-tmp folder when some transient conditions on ovnkube-xyz pods could make us flag them as "progressing". (found in https://issues.redhat.com//browse/OCPBUGS-30579)
- Remove all code related to the migration from multizone to single zone, since we've decided not to support that.
- Tighten `shouldUpdateOVNKonInterConnectZoneModeChange` to the presence of the IC configmap, so that it only takes action when the upgrade to IC is happening and doesn't interfere with cluster CIDR expansion (https://issues.redhat.com//browse/OCPBUGS-27925)

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -2121,7 +2121,7 @@ func daemonSetProgressing(ds *appsv1.DaemonSet, allowHung bool) bool {
 	}
 	klog.V(2).Infof("daemonset %s/%s rollout %s; %d/%d scheduled; %d unavailable; %d available; generation %d -> %d",
 		ds.Namespace, ds.Name, s, status.UpdatedNumberScheduled, status.DesiredNumberScheduled,
-		status.NumberUnavailable, status.NumberAvailable, ds.Generation, status.ObservedGeneration)
+		status.NumberUnavailable, status.NumberAvailable, status.ObservedGeneration, ds.Generation)
 
 	if !progressing {
 		klog.V(2).Infof("daemonset %s/%s rollout complete", ds.Namespace, ds.Name)
@@ -2180,7 +2180,7 @@ func deploymentProgressing(d *appsv1.Deployment) bool {
 	}
 	klog.V(2).Infof("deployment %s/%s rollout %s; %d/%d scheduled; %d available; generation %d -> %d",
 		d.Namespace, d.Name, s, status.ReadyReplicas, status.Replicas,
-		status.AvailableReplicas, d.Generation, status.ObservedGeneration)
+		status.AvailableReplicas, status.ObservedGeneration, d.Generation)
 
 	if !progressing {
 		klog.V(2).Infof("deployment %s/%s rollout complete", d.Namespace, d.Name)
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -528,7 +528,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 
 	// Process phase 2 of the upgrade to IC (4.14 single zone  -> 4.14 multizone)
-	updateNode, updateMaster, updateControlPlane := shouldUpdateOVNKonInterConnectZoneModeChange(bootstrapResult.OVN, targetZoneMode.zoneMode)
+	updateNode, updateMaster, updateControlPlane := shouldUpdateOVNKonInterConnectZoneModeChange(bootstrapResult.OVN, targetZoneMode)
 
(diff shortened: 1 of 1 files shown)


jluhrsen:
@ricky-rav , can we get back to this now please?

ricky-rav:
@tssurya can you please take a look when you have time? Thanks!

pkg/network/ovn_kubernetes.go
```diff
@@ -456,18 +459,21 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 
 	manifestDirs = append(manifestDirs, filepath.Join(manifestSubDirBasePath, "common"))
 
-	// choose the YAMLs based on the target zone mode (4.14 only) TODO: starting from 4.15, support multizone only
-	manifestSubDir := filepath.Join(manifestSubDirBasePath, "multi-zone-interconnect") // default is multizone
+	ongoingMigrationFromSingleToMultiZone := (!targetZoneMode.fastForwardToMultiZone &&
```
tssurya, line 462:
```suggestion
	ongoingUpgradePhaseFromSingleToMultiZone := (!targetZoneMode.fastForwardToMultiZone &&
```
I think "migration" is too overloaded, I'd prefer to keep all the upgrade related logic as some sort of upgrade related naming for better readability. (mtu migration, cni live/offline migration, gateway mode migration etc etc)

pkg/network/ovn_kubernetes.go
```diff
@@ -521,25 +527,23 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		fillKubeProxyDefaults(conf, nil)
 	}
 
-	zoneModeMigrationIsOngoing := isZoneModeMigrationAboutToStartOrOngoing(bootstrapResult.OVN, &targetZoneMode)
```
tssurya, line 524: what is the difference between `zoneModeMigrationIsOngoing` and `ongoingMigrationFromSingleToMultiZone`

pkg/network/ovn_kubernetes.go
```diff
@@ -2074,33 +2078,8 @@ func shouldUpdateOVNKonInterConnectZoneModeChange(ovn bootstrap.OVNBootstrapResu
 			return true, true, true
 		}
 
-	} else if targetZoneMode == zoneModeSingleZone {
-		// Multizone -> single zone
```
tssurya, line 2078: ++ no need good riddance

pkg/network/ovn_kubernetes.go
tssurya: AH I see `zoneModeMigrationIsOngoing` had checks to go from multizone to single zone so we don't need this anymore but still end to end what's the current diff here ?

ricky-rav: yes, the code included checks for the migration in the two directions

pkg/network/ovn_kubernetes.go
ricky-rav: It was a handy word ;)

pkg/network/ovn_kubernetes.go
ricky-rav: As for the current diff, CNO considers it's moving from single to multizone according to the existing conditions on OVNK status + the presence of the configmap. Since the configmap is removed by CNO as soon as the upgrade to IC completes, CNO won't mistakenly consider that it has to redo the single->multizone "migration" when transient conditions on any ovnkube-node instance make it look like an upgrade phase happening again.

So we're picking YAMLs from the single zone folder if the configmap says the target is single zone; we're picking YAMLs from the multizone-tmp folder if the conditions for the single->multizone phase hinted at in the  previous paragraph apply (ovnk status + config map); if none of that applies, we're picking YAMLs from the multizone folder, which is our final state. 

When we're not picking yamls from the multizone folder, `ongoingUpgradeToInterconnect` is true.

tssurya: ack thanks for explaining this to me.

tssurya (approved):
/lgtm

thanks Ric!

PR #2225, opened 2024-01-26
NP-903: Only allows live migration on SD managed clusters

pliurh:
In all the SD managed clusters, the namespace "dedicated-admin" shall exist. We use this to figure out if CNO is running in an SD managed cluster.

### DIFF
--- a/pkg/controller/clusterconfig/clusterconfig_controller.go
+++ b/pkg/controller/clusterconfig/clusterconfig_controller.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/network"
+	"github.com/openshift/cluster-network-operator/pkg/util"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -108,6 +109,18 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 	network.MergeClusterConfig(&operConfig.Spec, clusterConfig.Spec)
 
 	if _, ok := clusterConfig.Annotations[names.NetworkTypeMigrationAnnotation]; ok {
+		// Only allows live migration on SD managed clusters
+		inManagedCluster, err := util.IsRunningInManagedCluster(ctx, r.client)
+		if err != nil {
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		if !inManagedCluster {
+			err := fmt.Errorf("Network type live migration is not supported on self-managed clusters")
+			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
+				fmt.Sprintf("Failed to process network type live migration (%v).Use 'oc edit network.config.openshift.io cluster' to fix.", err))
+			return reconcile.Result{}, err
+		}
 		if hcp := hypershift.NewHyperShiftConfig(); hcp.Enabled {
 			err := fmt.Errorf("network type live migration is not supported on HyperShift clusters")
 			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/network"
 	"github.com/openshift/cluster-network-operator/pkg/platform"
+	"github.com/openshift/cluster-network-operator/pkg/util"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/operator/events"
 
@@ -255,6 +256,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	if operConfig.Spec.ManagementState == operv1.Unmanaged {
 		log.Printf("Operator configuration state is %s - skipping operconfig reconciliation", operConfig.Spec.ManagementState)
 		return reconcile.Result{}, nil
+	}
+
+	// Only allows live migration on SD managed clusters
+	inManagedCluster, err := util.IsRunningInManagedCluster(ctx, r.client)
+	if err != nil {
+		log.Println(err)
+		return reconcile.Result{}, err
+	}
+	if !inManagedCluster && operConfig.Spec.Migration != nil && operConfig.Spec.Migration.Mode == operv1.LiveNetworkMigrationMode {
+		r.status.SetDegraded(statusmanager.OperatorConfig, "MergeClusterConfig",
+			"invalid configuration: network type live migration is not supported on self-managed clusters. Use 'oc edit network.operator.openshift.io cluster' to fix.")
+		return reconcile.Result{}, err
 	}
 
 	// Merge in the cluster configuration, in case the administrator has updated some "downstream" fields
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -23,6 +24,7 @@ const SDN_NAMESPACE = "openshift-sdn"
 const MTU_CM_NAMESPACE = "openshift-network-operator"
 const MTU_CM_NAME = "mtu"
 const OVN_NBDB = "nbdb"
+const MANAGED_CLUSTER_NAMESPACE = "dedicated-admin"
 
 func GetInterConnectConfigMap(kubeClient kubernetes.Interface) (*corev1.ConfigMap, error) {
 	return kubeClient.CoreV1().ConfigMaps(OVN_NAMESPACE).Get(context.TODO(), OVN_INTERCONNECT_CONFIGMAP_NAME, metav1.GetOptions{})
@@ -42,4 +44,17 @@ func ReadMTUConfigMap(ctx context.Context, client cnoclient.Client) (int, error)
 
 	klog.V(2).Infof("Found mtu %d", mtu)
 	return mtu, nil
+}
+
+// IsRunningInManagedCluster returns true if the operator is running in a managed cluster.
+// It checks for the existence of the dedicated-admin namespace
+func IsRunningInManagedCluster(ctx context.Context, client cnoclient.Client) (bool, error) {
+	err := client.Default().CRClient().Get(ctx, types.NamespacedName{Name: MANAGED_CLUSTER_NAMESPACE}, &corev1.Namespace{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
 }


martinkennelly:
Depends on https://github.com/openshift/release/pull/48165

martinkennelly:
We have sought advice from SD and the best way we can currently detect if the cluster is managed or not is when a specific namespace is created. In the future, we will move to use metrics and remove this.

pkg/util/util.go
```diff
@@ -43,3 +45,16 @@ func ReadMTUConfigMap(ctx context.Context, client cnoclient.Client) (int, error)
...
+// It checks for the existence of the dedicated-admin namespace
+func IsRunningInManagedCluster(ctx context.Context, client cnoclient.Client) (bool, error) {
+	err := client.Default().CRClient().Get(ctx, types.NamespacedName{Name: MANAGED_CLUSTER_NAMESPACE}, &corev1.Namespace{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
```
tssurya, line 59: we are sure we don't want to double the stored results Into the obj right? trusting this has been tested

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// Only allows live migration on SD managed clusters
```
tssurya, line 261: nit:
```suggestion
	// Only allows live migration on managed clusters
```

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+	inManagedCluster, err := util.IsRunningInManagedCluster(ctx, r.client)
+	if err != nil {
+		log.Println(err)
+		return reconcile.Result{}, err
+	}
+	if !inManagedCluster && operConfig.Spec.Migration != nil && operConfig.Spec.Migration.Mode == operv1.LiveNetworkMigrationMode {
+		r.status.SetDegraded(statusmanager.OperatorConfig, "MergeClusterConfig",
```
tssurya, line 268: so ultimately anyone can set the API value and we are going to make the operator force itself to become degraded? I hope we are doing release notes for this, seems like we support the API officially but one can't use it. Also is this API still in TechPreviewNoUpgrade and this is the requirement from API team for them to allow this in normal clusters (sorry I am missing context on this one)?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+			"invalid configuration: network type live migration is not supported on self-managed clusters. Use 'oc edit network.operator.openshift.io cluster' to fix.")
```
tssurya, line 269: does that mean the `Mode` field is actually mutable at runtime? there is some corner case here for mtu migration? (has(self.mode) && self.mode == 'Live'",message="networkType migration in mode other than 'Live' may not be configured at the same time as mtu migration")

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -108,6 +109,18 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 	network.MergeClusterConfig(&operConfig.Spec, clusterConfig.Spec)
 
 	if _, ok := clusterConfig.Annotations[names.NetworkTypeMigrationAnnotation]; ok {
+		// Only allows live migration on SD managed clusters
```
tssurya, line 112: nit:
```suggestion
		// Only allows live migration on managed clusters
```
not a fan of using "SD" but I'll leave it to you

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -108,6 +109,18 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
...
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		if !inManagedCluster {
+			err := fmt.Errorf("Network type live migration is not supported on self-managed clusters")
+			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
+				fmt.Sprintf("Failed to process network type live migration (%v).Use 'oc edit network.config.openshift.io cluster' to fix.", err))
+			return reconcile.Result{}, err
+		}
 		if hcp := hypershift.NewHyperShiftConfig(); hcp.Enabled {
```
tssurya, line 124: ~wait what is Hypershift cluster not supported mean? are we differentiating "SD HCP" from "normal HCP"? if so that seems super hacky to me, isn't `hcp.Enabled` set in the managed cluster? again sorry for the questions due to lack of knowledge here.~
NVM, the comment was misleading, this is meant to work only on hosted clusters and not on self-hosted and management clusters...

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -108,6 +109,18 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 	network.MergeClusterConfig(&operConfig.Spec, clusterConfig.Spec)
 
 	if _, ok := clusterConfig.Annotations[names.NetworkTypeMigrationAnnotation]; ok {
```
tssurya, line 111: interesting that we use an annotation for the clusterconfig controller side of things to signal a live migration? but the name of the annotation is generic that I don't know if its live or something else.. but unrelated to this PR and this comment is out of context and self homework for myself.

tssurya (commented):
definitely missing some higher level context, so left some inline questions but overall my main concern is with exposing something in the api but not letting people use it if I understand the crux correctly.
however I am only a reviewer @jcaamano / @kyrtapz have more knowledge here.

pkg/util/util.go
martinkennelly: yep - should be fine.

pkg/controller/operconfig/operconfig_controller.go
martinkennelly: > does that mean the Mode field is actually mutable at runtime? 

yes

> there is some corner case here for mtu migration?

```
+openshift:validation:FeatureSetAwareXValidation:featureSet=CustomNoUpgrade;TechPreviewNoUpgrade,rule="!has(self.mtu) || !has(self.networkType) || self.networkType == '' || has(self.mode) && self.mode == 'Live'",message="networkType migration in mode other than 'Live' may not be configured at the same time as mtu migration"
type NetworkMigration struct {
```

martinkennelly: Anyway, this is outside the scope of this PR. @pliurh  can you confirm?

pkg/controller/clusterconfig/clusterconfig_controller.go
martinkennelly: > interesting that we use an annotation for the clusterconfig controller side of things to signal a live migration?

Yes, it gates the migration.

pkg/controller/clusterconfig/clusterconfig_controller.go
martinkennelly: Ill ignore this comment unless others want me to make changes to this PR, then Ill create my own and change this. I agree that id rather drop the SD too.

pkg/util/util.go
```diff
@@ -23,6 +24,7 @@ const SDN_NAMESPACE = "openshift-sdn"
 const MTU_CM_NAMESPACE = "openshift-network-operator"
 const MTU_CM_NAME = "mtu"
 const OVN_NBDB = "nbdb"
+const MANAGED_CLUSTER_NAMESPACE = "dedicated-admin"
```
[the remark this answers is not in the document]
martinkennelly: I think Id rather not push a new PR just for this. Peng is on PTO for 3 weeks. LMK if you feel strongly about this and Ill do it.

martinkennelly:
@tssurya appreciate the review and throughness - I dont think any changes are essential from your comments. LMK otherwise. If there are changes needed, ill push my own PR and add in your suggested changes.

pkg/controller/operconfig/operconfig_controller.go
martinkennelly: Ignoring this unless theres necessary changes and i must create my own PR. I will add it in then.

pkg/util/util.go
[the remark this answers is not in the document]
tssurya: It's ok to not repush for this, unless Jaime has more code wise suggestions. But please let's come back to this in the future and track it somehow

pkg/util/util.go
tssurya: resolve.

pkg/controller/operconfig/operconfig_controller.go
tssurya: sgtm.

pkg/controller/operconfig/operconfig_controller.go
tssurya: ack thanks @martinkennelly that deff sets the context for me. Again not a fan of this change but if this is what we have to do then so be it.

pkg/controller/operconfig/operconfig_controller.go
tssurya: might not be too much outside the scope of this PR because if you are giving a warning for a user to mutate it but you can't really mutate in all circumstances we must notify users of that limitation, but ok again to me if this is super urgent to get in I will not block for this.

pkg/controller/clusterconfig/clusterconfig_controller.go
tssurya: sgtm

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -108,6 +109,18 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
...
+		if err != nil {
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		if !inManagedCluster {
+			err := fmt.Errorf("Network type live migration is not supported on self-managed clusters")
+			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
+				fmt.Sprintf("Failed to process network type live migration (%v).Use 'oc edit network.config.openshift.io cluster' to fix.", err))
+			return reconcile.Result{}, err
+		}
```
jcaamano, line 123: So in theory, we could just do the check in the oper config controller and not here. I am not sure if I am missing why this additional check is being done here, unless it is just to give a more meaningful error? But then (next comment...)

pkg/controller/operconfig/operconfig_controller.go
jcaamano, line 269: I would be more vague in this error given that in this case the user must have modified the operator config itself and not the cluster config so the resolution would not be `oc edit network.operator.openshift.io cluster' to fix.`

jcaamano (commented):
Just a couple of nits

pkg/controller/clusterconfig/clusterconfig_controller.go
tssurya: haha good question, I had the same thought but then on my side I just don't know exactly what the cluster-config oper does which operconfig doesn't.

pkg/controller/operconfig/operconfig_controller.go
tssurya: is the operator.config itself modifiable by the user? I always forget which CRD is for users to modify and which isn't

jcaamano:
known flake : https://github.com/ovn-org/ovn-kubernetes/issues/4060

pkg/controller/operconfig/operconfig_controller.go
martinkennelly: user should only modify `network.config.openshift.io` and not `network.operator.openshift.io`.

Jaime, youre correct, I will make a new PR.

pkg/controller/clusterconfig/clusterconfig_controller.go
martinkennelly: I dont know why its there as well - removing it as its not needed.

martinkennelly:
Pushed a new PR to address:
* Removed "SD" from comments
* Added in `todo` to change from detection of managed cluster from a NS to a metric - no bug for this. Asking SD team. Will add PR later to add the jira if i get one.
* Removed redundant check for a managed cluster in config controller. I could see no reason for it. If theres a reason, we can add it later.
* Updated degraded message in operconfig controller.

https://github.com/openshift/cluster-network-operator/pull/2236

martinkennelly:
/close

In favour of #2236

PR #2234, opened 2024-01-29
OCPBUGS-19551: ovnkube: simplify northd threading and SNO templating

trozet:
It's all single thread now, plus we don't need an if/else block for SNO anymore since it's just a single boolean now.


tssurya (approved):
/lgtm

PR #2236, opened 2024-01-30
NP-903: Only allows live migration on standalone managed clusters

martinkennelly:
In all the managed clusters excl hypershift aka standalone managed clusters, the namespace "dedicated-admin" shall exist. We use this to figure out if CNO is running in a managed cluster.

### DIFF
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -75,6 +75,8 @@ type InfraStatus struct {
 	ControlPlaneTopology   configv1.TopologyMode
 	InfrastructureTopology configv1.TopologyMode
 	InfraName              string
+	// StandaloneManagedCluster is set to true when the cluster is a standalone managed cluster (excl HyperShift)
+	StandaloneManagedCluster bool
 
 	// KubeCloudConfig is the contents of the openshift-config-managed/kube-cloud-config ConfigMap
 	KubeCloudConfig map[string]string
--- a/pkg/controller/clusterconfig/clusterconfig_controller.go
+++ b/pkg/controller/clusterconfig/clusterconfig_controller.go
@@ -10,10 +10,8 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/apply"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
-	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/network"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -92,8 +90,8 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 	}
 
 	// Validate the cluster config
-	if err := network.ValidateClusterConfig(clusterConfig.Spec, r.client); err != nil {
-		log.Printf("Failed to validate Network.Spec: %v", err)
+	if err := network.ValidateClusterConfig(clusterConfig, r.client); err != nil {
+		log.Printf("Failed to validate Network CR: %v", err)
 		r.status.SetDegraded(statusmanager.ClusterConfig, "InvalidClusterConfig",
 			fmt.Sprintf("The cluster configuration is invalid (%v). Use 'oc edit network.config.openshift.io cluster' to fix.", err))
 		return reconcile.Result{}, err
@@ -108,12 +106,6 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 	network.MergeClusterConfig(&operConfig.Spec, clusterConfig.Spec)
 
 	if _, ok := clusterConfig.Annotations[names.NetworkTypeMigrationAnnotation]; ok {
-		if hcp := hypershift.NewHyperShiftConfig(); hcp.Enabled {
-			err := fmt.Errorf("network type live migration is not supported on HyperShift clusters")
-			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
-				fmt.Sprintf("Failed to process network type live migration (%v). Use 'oc edit network.config.openshift.io cluster' to fix.", err))
-			return reconcile.Result{}, err
-		}
 		// https://github.com/openshift/enhancements/blob/master/enhancements/network/sdn-live-migration.md#api
 		if err := r.processNetworkTypeLiveMigration(ctx, request, clusterConfig, operConfig); err != nil {
 			log.Printf("Failed to process SDN live migration: %v", err)
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -23,6 +24,7 @@ const SDN_NAMESPACE = "openshift-sdn"
 const MTU_CM_NAMESPACE = "openshift-network-operator"
 const MTU_CM_NAME = "mtu"
 const OVN_NBDB = "nbdb"
+const STANDALONE_MANAGED_CLUSTER_NAMESPACE = "dedicated-admin" // namespace required for standalone managed clusters (excluding hypershift)
 
 func GetInterConnectConfigMap(kubeClient kubernetes.Interface) (*corev1.ConfigMap, error) {
 	return kubeClient.CoreV1().ConfigMaps(OVN_NAMESPACE).Get(context.TODO(), OVN_INTERCONNECT_CONFIGMAP_NAME, metav1.GetOptions{})
@@ -42,4 +44,19 @@ func ReadMTUConfigMap(ctx context.Context, client cnoclient.Client) (int, error)
 
 	klog.V(2).Infof("Found mtu %d", mtu)
 	return mtu, nil
+}
+
+// IsStandaloneManagedCluster returns true if the operator is running in a managed cluster that isn't managed by HyperShift.
+// It checks for the existence of the dedicated-admin namespace.
+func IsStandaloneManagedCluster(ctx context.Context, client cnoclient.Client) (bool, error) {
+	// TODO(martinkennelly): replace detection of a standalone managed cluster with a metric instead of a namespace when that metric
+	// becomes available.
+	err := client.Default().CRClient().Get(ctx, types.NamespacedName{Name: STANDALONE_MANAGED_CLUSTER_NAMESPACE}, &corev1.Namespace{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
 }
--- a/pkg/controller/operconfig/cluster.go
+++ b/pkg/controller/operconfig/cluster.go
@@ -38,7 +38,7 @@ func (r *ReconcileOperConfig) MergeClusterConfig(ctx context.Context, operConfig
 	}
 	// Validate cluster config
 	// If invalid just warn and proceed.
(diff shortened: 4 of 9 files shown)


jcaamano (commented):
Two questions here @martinkennelly 

When we talk about SD managed clusters, is that the same thing as hypershift managed clusters? If it is, we could do the same check as we do in other places: `if hc := hypershift.NewHyperShiftConfig(); hc.Enabled`. If it is not, then I don't think the `SD managed" qualification is out of line because otherwise the managed term is just overloaded. WDYT @tssurya ? We can look for other terminology if we would like.

Also, are you testing this in any way? Is it going to be tested before back-porting to 4.15?

martinkennelly:
> When we talk about SD managed clusters, is that the same thing as hypershift managed clusters? 

@jcaamano I am not sure about the terminology but I know Hypershift doesnt support SDN, therefore we dont need this migration for Hypershift and its not in the set of supported managed clusters.

> Also, are you testing this in any way? Is it going to be tested before back-porting to 4.15?

Only with CI, and even then CI is mimicking the managed cluster env by artificially creating that namespace before executing the live migration test. I wanted to test with cluster bot and create a rosa cluster but its only possible with a nightly. I will get QE to test it before backporting.

tssurya:
> Two questions here @martinkennelly
> 
> When we talk about SD managed clusters, is that the same thing as hypershift managed clusters? If it is, we could do the same check as we do in other places: `if hc := hypershift.NewHyperShiftConfig(); hc.Enabled`. 

yes IMO they are the same thing as HCP managed clusters.

> If it is not, then I don't think the "SD managed" qualification is out of line because otherwise the managed term is just overloaded. WDYT @tssurya ? We can look for other terminology if we would like.

So really this PR applies to any "managed cluster that has the dedicated-admin namespace created inside it". Today sure it is done probably only by SD, but given HCP as an offering anyone could potentially just have a namespace with that name and this will work just fine which is why I opted for "all managed clusters" and prevented the overload of "SD"

jcaamano:
> yes IMO they are the same thing as HCP managed clusters. but this feature is being limited to the SD managed clusters is my understanding, but yeah probably we are overthinking this given the timeline

So this contradicts what @martinkennelly says above?

> but given HCP as an offering anyone could potentially just have a namespace with that name and this will work just fine which is why I opted for "all managed clusters" and prevented the overload of "SD".

While that is a defect of how we detect an SD managed cluster, we don't really want that, do we? So really the intended functionality is for this to work in SD managed clusters, for lack of a better term.

martinkennelly:
@jcaamano SD managed clusters can include hypershift clusters but we dont support live migration in hypershift. you can see we have a check already that if the CNO is within the hosted control plane, we dont allow live migration.
I apologise for not using watching my terminology. Lets add some comments to ensure the reader can see we mean managed clusters exclusing hypershift.

jcaamano:
> @jcaamano SD managed clusters can include hypershift clusters but we dont support live migration in hypershift. you can see we have a check already that if the CNO is within the hosted control plane, we dont allow live migration. I apologise for not using watching my terminology. Lets add some comments to ensure the reader can see we mean managed clusters excluding hypershift.

Thanks, got it now.

martinkennelly:
/test e2e-aws-live-migration-sdn-ovn

Cluster setup failure

martinkennelly:
/retest

unrelated failures

abhat:
@kyrtapz can you review this when you are back? also [PR #2235](https://github.com/openshift/cluster-network-operator/pull/2235) has more context and discussion.

zshi-redhat:
> @kyrtapz can you review this when you are back? also [PR #2235](https://github.com/openshift/cluster-network-operator/pull/2235) has more context and discussion.

I think what @abhat means is this PR #2225

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// Only allow live migration in managed clusters excluding hypershift clusters
+	isNonHypershiftManagedCluster, err := util.IsNonHypershiftManagedCluster(ctx, r.client)
```
martinkennelly, line 262: I wasnt a fan of this exclusive naming but I couldnt think of a better one. I think the comment above it helps explain that its the set of all managed cluster excluding hypershift.

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -114,6 +115,16 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
...
 			return reconcile.Result{}, err
 		}
+		isStandaloneManagedCluster, err := util.IsStandaloneManagedCluster(ctx, r.client)
+		if err != nil {
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		if !isStandaloneManagedCluster {
+			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
+				"Failed to process network type live migration (%v).Use 'oc edit network.config.openshift.io cluster' to fix.")
```
kyrtapz:
```suggestion
				"Failed to process network type live migration (%v). Use 'oc edit network.config.openshift.io cluster' to fix.")
```

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// Only allow live migration in managed clusters excluding hypershift clusters
+	isStandaloneManagedCluster, err := util.IsStandaloneManagedCluster(ctx, r.client)
+	if err != nil {
+		log.Println(err)
```
kyrtapz, line 264: Please add a log message

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -114,6 +115,16 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
...
 			return reconcile.Result{}, err
 		}
+		isStandaloneManagedCluster, err := util.IsStandaloneManagedCluster(ctx, r.client)
+		if err != nil {
+			log.Println(err)
+			return reconcile.Result{}, err
+		}
+		if !isStandaloneManagedCluster {
+			r.status.SetDegraded(statusmanager.ClusterConfig, "NetworkTypeMigrationFailed",
+				fmt.Sprintf("Failed to process network type live migration (%v).Use 'oc edit network.config.openshift.io cluster' to fix.", err))
```
kyrtapz, line 126:
```suggestion
				fmt.Sprintf("Failed to process network type live migration (%v). Use 'oc edit network.config.openshift.io cluster' to fix.", err))
```

pkg/controller/operconfig/operconfig_controller.go
kyrtapz, line 264: Please add a log message

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -257,6 +258,18 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, nil
 	}
 
+	// Only allow live migration in managed clusters excluding hypershift clusters
+	isStandaloneManagedCluster, err := util.IsStandaloneManagedCluster(ctx, r.client)
```
kyrtapz, line 262: Wouldn't that whole check better fit into `isMigrationChangeSafe`?
https://github.com/openshift/cluster-network-operator/blob/a5edddf194e70a8ab01a44dbc4e3d8214a773f10/pkg/network/render.go#L673

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -114,6 +115,17 @@ func (r *ReconcileClusterConfig) Reconcile(ctx context.Context, request reconcil
 				fmt.Sprintf("Failed to process network type live migration (%v). Use 'oc edit network.config.openshift.io cluster' to fix.", err))
 			return reconcile.Result{}, err
 		}
+		// Only allow live migration in managed clusters excluding HyperShift clusters
```
martinkennelly, line 118: Adding this check back because there are checks in both cluster config and oper config controllers. There is no separation  between the controllers currently - i.e. one does the checks. Continue that pattern.

Also, if we dont fail here, we're translating an unsupported config to oper config. Even though, its harmless, lets stop an unsupported config in both controllers.

kyrtapz (commented):
Overall looks good, some minor comments.
Can you add a test for this configuration to `migration_tests.go`?

pkg/controller/clusterconfig/clusterconfig_controller.go
jcaamano: Makes sense. The operator config is also exposed to the user so agreed.

pkg/controller/operconfig/operconfig_controller.go
martinkennelly: I think validate is better. Move it there.

martinkennelly:
Update resulting from Pats review - buried the support "validation" of the feature inside the validation step in each controller. Its a bit cleaner even though that validation func wasnt meant to detect support per say but better than creating a new func. Added a very small unit test.

Theres some asymmetry in the checking in each controller. Network.Oper validation doesnt look for annotation but this is natural since we do not apply the Network.Config annotations to Network.Oper.

martinkennelly:
/test e2e-azure-ovn 

Bootstrap issue

kyrtapz:
/lgtm
/hold
for CI and @jcaamano lgtm

kyrtapz:
> Update resulting from Pats review - buried the support "validation" of the feature inside the validation step in each controller. Its a bit cleaner even though that validation func wasnt meant to detect support per say but better than creating a new func. Added a very small unit test.

I think we are still missing the unit test :)

pkg/util/util.go
```diff
@@ -43,3 +45,18 @@ func ReadMTUConfigMap(ctx context.Context, client cnoclient.Client) (int, error)
...
+	// becomes available.
+	err := client.Default().CRClient().Get(ctx, types.NamespacedName{Name: STANDALONE_MANAGED_CLUSTER_NAMESPACE}, &corev1.Namespace{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
```
jcaamano, line 62: Shouldn't we just set this information somewhere in InfraStatus() which already handles a context (however appropriate is or isn't the actual context that is being used there)?

infraStatus is already being used in the relevant places and will drive where you can do the validation.
For cluster config, it fits in `ValidateClusterConfig`.
For oper config, it looks more like it would go into `IsChangeSafe`. It looks like `Validate` here is more about performing a static validation of the oper config itself, naming conventions notwithstanding.

martinkennelly: Thanks - It makes sense. Pat suggested the same earlier.
I agree the current intent for `Validate` is ensuring the values are within bounds and not checking if a value is ok wrt external (to the field) constraints (cluster env / supported).

martinkennelly:
Looked at job `e2e-network-mtu-migration-ovn-ipv4` - failure is unrelated. Migration job was successful.

martinkennelly:
@jcaamano @kyrtapz PTAL

martinkennelly:
Pat said via slack he's happy with the changes.

pkg/bootstrap/types.go
```diff
@@ -75,6 +75,8 @@ type InfraStatus struct {
 	ControlPlaneTopology   configv1.TopologyMode
 	InfrastructureTopology configv1.TopologyMode
 	InfraName              string
+	// StandaloneManagedCluster is set to true when the cluster is a standalone managed cluster (excl HyperShift)
```
jcaamano, line 78: nit:
```
// StandaloneManagedCluster is set to true when the cluster is a non-hypershift managed cluster
```
or you could just have
```
// a managed cluster is either a hosted hypershift cluster or standalone
ManagedCluster := infraRes.HostedControlPlane != nil || isStandaloneManagedCluster
```
and use that to check.

jcaamano (commented):
Just that nit which is very opinionated.

Overall this LGTM. @kyrtapz WDYT ?

pkg/bootstrap/types.go
kyrtapz: I don't think that would work, for live migration we support standalone managed clusters but not hypershift.

jcaamano: My bad. So ignore the second part

martinkennelly:
Verified by QE

PR #2238, opened 2024-01-31
OCPBUGS-28676: Remove libreswan rpm existence checks

pperiyasamy:
There is no need for libreswan rpm existence checks anymore in ipsec daemonset pod because libreswan is always expected to be present, otherwise it should be an error. Hence this commit is removing those checks. This also inherently fixes rpm db directory mount issue because rhel and rhcos use different directories for it.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
@@ -90,11 +90,6 @@ spec:
           export KUBECONFIG=/var/run/ovnkube-kubeconfig
 {{ end }}
 
-          if rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host has libreswan and therefore ipsec will be configured by ipsec daemonset, this ovn ipsec container doesnt need to init anything"
-            exit 0
-          fi
-
           # Every time we restart this container, we will create a new key pair if
           # we are close to key expiration or if we do not already have a signed key pair.
           #
@@ -192,9 +187,6 @@ spec:
           name: signer-ca
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -225,11 +217,6 @@ spec:
             exit 0
           }
           trap cleanup SIGTERM
-
-          if rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host has libreswan and therefore ipsec will be configured by ipsec daemonset, this ovn ipsec container will sleep to infinity"
-            sleep infinity
-          fi
 
           # Don't start IPsec until ovnkube-node has finished setting up the node
           counter=0
@@ -287,9 +274,6 @@ spec:
           name: host-var-log-ovs
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -302,11 +286,6 @@ spec:
             - -c
             - |
               #!/bin/bash
-              if rpm --dbpath=/usr/share/rpm -q libreswan; then
-                echo "host has libreswan and therefore ipsec will be configured by ipsec daemonset, this ovn ipsec container is always \"alive\""
-                exit 0
-              fi
-
               if [[ $(ipsec whack --trafficstatus | wc -l) -eq 0 ]]; then
                 echo "no ipsec traffic configured"
                 exit 10
@@ -340,10 +319,6 @@ spec:
       - name: host-cni-netd
         hostPath:
           path: "{{.CNIConfDir}}"
-      - name: host-usr-share-rpm
-        hostPath:
-          path: /usr/share/rpm
-          type: Directory
       tolerations:
       - operator: "Exists"
 {{end}}
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -91,11 +91,6 @@ spec:
           export KUBECONFIG=/var/run/ovnkube-kubeconfig
 {{ end }}
 
-          if ! rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host doesnt have libreswan, therefore ipsec will be configured by ipsec-pre414 daemonset, this ovn ipsec container has nothing to init"
-            exit 0
-          fi
-
           # Every time we restart this container, we will create a new key pair if
           # we are close to key expiration or if we do not already have a signed key pair.
           #
@@ -197,9 +192,6 @@ spec:
           name: etc-openvswitch
         - mountPath: /etc
           name: host-etc
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -216,11 +208,6 @@ spec:
           #!/bin/bash
           set -exuo pipefail
 
-
-          if ! rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host doesnt have libreswan, therefore ipsec will be configured by ipsec-pre414 daemonset, this ovn ipsec container will sleep to infinity"
-            sleep infinity
-          fi
 
           # Don't start IPsec until ovnkube-node has finished setting up the node
           counter=0
@@ -279,11 +266,6 @@ spec:
                    # In order to maintain traffic flows during container restart, we
                    # need to ensure that xfrm state and policies are not flushed.
 
-                   if ! rpm --dbpath=/usr/share/rpm -q libreswan; then
-                     echo "host doesnt have libreswan, therefore ipsec will be configured by ipsec-pre414 daemonset, preStop wont do anything"
-                     exit 0
-                   fi
-
                    # Don't allow ovs monitor to cleanup persistent state
                    kill $(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null) 2>/dev/null || true
         env:
@@ -307,9 +289,6 @@ spec:
           name: host-var-lib
         - mountPath: /etc
           name: host-etc
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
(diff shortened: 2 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -245,6 +245,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["IPsecMachineConfigEnable"] = IPsecMachineConfigEnable
 	data.Data["OVNIPsecDaemonsetEnable"] = OVNIPsecDaemonsetEnable
 	data.Data["OVNIPsecEnable"] = OVNIPsecEnable
+	data.Data["RHEL_OS_ENABLE"] = bootstrapResult.Infra.HostedControlPlane == nil
```
yuvalk, line 248: that's for hypershift.
you need to check whether there are any nodes with RHEL.

bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
```diff
@@ -34,6 +34,12 @@ spec:
             - matchExpressions:
               - key: network.operator.openshift.io/dpu-host
                 operator: DoesNotExist
+{{ if .RHEL_OS_ENABLE }}
+              - key: node.openshift.io/os_id
+                operator: In
+                values:
+                - rhel
+{{ end }}
```
yuvalk, line 42: also need to modify host-usr-share-rpm
on a rhel node rpm db is in `/var/lib/rpm`

and need to create the nss db directory in /var/lib/ipsec/nss 
(need a `mkdir -p` because ipsec directory is also missing)

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -218,7 +224,7 @@ spec:
 
 
           if ! rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host doesnt have libreswan, therefore ipsec will be configured by ipsec-pre414 daemonset, this ovn ipsec container will sleep to infinity"
+            echo "host doesnt have libreswan, therefore ipsec will be configured by ipsec-containerized daemonset, this ovn ipsec container will sleep to infinity"
```
yuvalk, line 227: we should not sleep infinity we should error instead

bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
pperiyasamy: ok @yuvalk , now completely removing rpm check from DS manifests so that both DS can be rendered unconditionally and its pod can be scheduled accordingly based on the underlying OS.

anuragthehatter:
we see both ds's on RHEL nodes?

```
$ oc get ds
NAME                      DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ovn-ipsec-containerized   2         2         2       2            2           kubernetes.io/os=linux   128m
ovn-ipsec-host            3         3         3       3            3           kubernetes.io/os=linux   94m
ovnkube-node              5         5         5       5            5           kubernetes.io/os=linux   128m

```
Both `ovn-ipsec-host` and `ovn-ipsec-containerized` pods contains same `ovn-ipsec` container

pkg/network/ovn_kubernetes.go
```diff
@@ -245,6 +245,8 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["IPsecMachineConfigEnable"] = IPsecMachineConfigEnable
 	data.Data["OVNIPsecDaemonsetEnable"] = OVNIPsecDaemonsetEnable
 	data.Data["OVNIPsecEnable"] = OVNIPsecEnable
+	// Set RHEL_OS_CHECK to true only for non Hybrid hosted cluster.
+	data.Data["RHEL_OS_CHECK"] = bootstrapResult.Infra.HostedControlPlane == nil
```
jcaamano, line 249: `RHEL_OS_CHECK` is not the most intuitive variable name.

Could we reuse something that we already have, like `{{- if not (eq .HostedClusterNamespace "")}}`?

Otherwise, if you think it adds value, I would be in favor of introducing a more explicit variable, like `HYPERSHIFT_HOSTED_CLUSTER"

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -34,6 +34,12 @@ spec:
             - matchExpressions:
               - key: network.operator.openshift.io/dpu-host
                 operator: DoesNotExist
+{{ if .RHEL_OS_CHECK }}
```
jcaamano, line 37: We would not need this if template condition right, as this is never deployed for a hosted cluster.

jcaamano (commented):
I have the feeling that we keep going in circles with ipsec and is difficult for me to follow up.

I have some questions:

1. Isn't it just easier to require that package on RHEL? Or is it not available at all?
2. I guess that N/S ipsec still works on the RHEL nodes?
3. What is the ipsec-containerized status when deployed on a cluster with no RHEL nodes? 

Also, can you take a final look @yuvalk ?

pkg/network/ovn_kubernetes.go
pperiyasamy: you are right, just reusing HostedClusterNamespace now.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: correct, removed it.

pperiyasamy:
> I have the feeling that we keep going in circles with ipsec and is difficult for me to follow up.
> 
> I have some questions:
> 
> 1. Isn't it just easier to require that package on RHEL? Or is it not available at all?

I think @yuvalk can answer to this question, From the discussion [here](https://redhat-internal.slack.com/archives/C06665AQN0J/p1706660724343449), this may not be possible.

> 2. I guess that N/S ipsec still works on the RHEL nodes?

Looks like N/S ipsec never supported in 4.14 for RHEL nodes. @yuvalk ?

> 3. What is the ipsec-containerized status when deployed on a cluster with no RHEL nodes?

Here is the ipsec-containerized DS status when cluster have no RHEL nodes.

```
$ oc get daemonset -n openshift-ovn-kubernetes
NAME                      DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
ovn-ipsec-containerized   0         0         0       0            0           kubernetes.io/os=linux   24h
ovn-ipsec-host            5         5         5       5            5           kubernetes.io/os=linux   24h
ovnkube-node              5         5         5       5            5           kubernetes.io/os=linux   46h
```

> 
> Also, can you take a final look @yuvalk ?

yuvalk:
> I have the feeling that we keep going in circles with ipsec and is difficult for me to follow up.
> 
> I have some questions:
> 
>     1. Isn't it just easier to require that package on RHEL? Or is it not available at all?
No,
the problem is with existing customers of E-W IPsec
currently they are not required to install it
and if they upgrade, they'll get nodes without ipsec
which is a regression.
see the bug.

> 
>     2. I guess that N/S ipsec still works on the RHEL nodes?
We plan on no-support for N-S on RHEL Nodes.
if we do like to support it, we'll need to do the same trick we did in 4.14,
where we deploy 2 ds on each rhel node, and according to it's rpm db, we decide which one is active.
then we can instruct users to install libreswan and also manually manage the connections
do we want to do that now? (my opinion is no)
> 
>     3. What is the ipsec-containerized status when deployed on a cluster with no RHEL nodes?
> 
> 
> Also, can you take a final look @yuvalk ?
I'll do that now

bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
```diff
@@ -90,11 +96,6 @@ spec:
           export KUBECONFIG=/var/run/ovnkube-kubeconfig
 {{ end }}
 
-          if rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host has libreswan and therefore ipsec will be configured by ipsec daemonset, this ovn ipsec container doesnt need to init anything"
-            exit 0
-          fi
-
```
yuvalk, line 97: I think you need this with similar condition to the node selector
on hcp, you dont want the containerized version to deploy on RHCOS workers.. (and you are not using the node selector there..)

yuvalk: you'll have to also solve the rpm db location issue
might not be able to escape it by using 2 different daemon sets

jcaamano:
> No, the problem is with existing customers of E-W IPsec currently they are not required to install it and if they upgrade, they'll get nodes without ipsec which is a regression. see the bug.

So this does look a bit weird. They are using a tool scale up the cluster with RHEL nodes, openshift-ansible I guess it is or based on that. That tool needs to be aligned to whatever openshift release the scale up is being done on. So it would make sense that tool would need to run on openshift upgrades to as well to bring those nodes up to the standards of what that openshift upgrade requires. I thing they might actually do: https://github.com/openshift/openshift-ansible#run-the-rhel-node-upgrade-playbook
So maybe as part of that the package could be installed.

yuvalk:
> > No, the problem is with existing customers of E-W IPsec currently they are not required to install it and if they upgrade, they'll get nodes without ipsec which is a regression. see the bug.
> 
> So this does look a bit weird. They are using a tool scale up the cluster with RHEL nodes, openshift-ansible I guess it is or based on that. That tool needs to be aligned to whatever openshift release the scale up is being done on. So it would make sense that tool would need to run on openshift upgrades to as well to bring those nodes up to the standards of what that openshift upgrade requires. I thing they might actually do: https://github.com/openshift/openshift-ansible#run-the-rhel-node-upgrade-playbook So maybe as part of that the package could be installed.

I'm no expert on that whole RHEL Nodes thing..
I thought it's more of a 'do it yourself' mode
but if there's a MANDATORY playbook, we should certainly and simply add libreswan (and nm-libreswan) there.

then we'll get
a. consistency with rhcos and ovn-ipsec-host (that will be able to run on rhel too)
b. easier ability to do n-s.

the only thing we need to worry about is documenting the need of that package on rhel nodes

bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
pperiyasamy: The  rpm db location is used for checking libreswan package is present or not. But irrespective of we render both ipsec daemon sets, finally there would  be either ipsec-containerized or ipsec-host pod being scheduled on a node and libreswan packages are expected to be present for pod to run.  so these checks are no longer needed, hence removed it.
for hcp clusters, we support only containerized version even on rhcos because machine configs are not supported there.

pperiyasamy:
> I thing they might actually do: https://github.com/openshift/openshift-ansible#run-the-rhel-node-upgrade-playbook
> So maybe as part of that the package could be installed.

@jcaamano @yuvalk now we have changes in openshift-ansible to include ipsec rpms with [PR](https://github.com/openshift/openshift-ansible/pull/12477), updated this PR to fix only rpm db directory mount issue. Please take a look.

yuvalk:
/lgtm
lgtm
though I am a bit worried what will happen in error cases where both will be running
(obviously that would be a bug, but maybe we can somehow at least detect it and issue an error?)

PR #2259, opened 2024-02-07
OCPBUGS-28920: Update ingressconfig_controller to use field Manager

npinaeva:
Update ingressconfig_controller to use subcontroller "ingress_controller"
for field management on Patch calls. It used to apply changes with
deprecated cluster-network-operator field manager, that was overridden
and removed by ApplyObject. Thus, labels applied to
openshift-host-network namespace were periodically removed.

On upgrade, ingress-owned labels will be merged with "cluster-network-operator/operconfig" Apply call here https://github.com/openshift/cluster-network-operator/blob/master/pkg/apply/apply.go#L139-L150 and removed as the new owner didn't specify them. On the next reconcile loop, ingress controller will re-apply the labels with the new fieldManager and everything should work fine ever after.

For some reason, ingress controller was last this that doesn't use apply logic, but Patches instead.

Should end up with something like this
```
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:openshift.io/description: {}
          f:workload.openshift.io/allowed: {}
        f:labels:
          f:policy-group.network.openshift.io/host-network: {}
        f:ownerReferences:
          k:{"uid":"eae5d3b2-cf31-4bfa-940a-fa631adce87d"}: {}
    manager: cluster-network-operator/operconfig
    operation: Apply
    time: "2024-02-07T11:38:45Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:labels:
          f:network.openshift.io/policy-group: {}
          f:policy-group.network.openshift.io/ingress: {}
      f:spec: {}
    manager: cluster-network-operator/ingress_controller
    operation: Apply
    time: "2024-02-07T11:41:11Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:openshift.io/sa.scc.mcs: {}
          f:openshift.io/sa.scc.supplemental-groups: {}
          f:openshift.io/sa.scc.uid-range: {}
    manager: cluster-policy-controller
    operation: Update
    time: "2024-02-07T08:49:53Z"

```

### DIFF
--- a/pkg/controller/ingressconfig/ingressconfig_controller.go
+++ b/pkg/controller/ingressconfig/ingressconfig_controller.go
@@ -142,5 +142,7 @@ func (r *ReconcileIngressConfigs) updatePolicyGroupLabelOnNamespace(ctx context.
 
 	newNamespace.SetLabels(existingLabels)
 
-	return r.client.Patch(context.TODO(), newNamespace, crclient.MergeFrom(namespace))
+	return r.client.Patch(context.TODO(), newNamespace, crclient.MergeFrom(namespace), &crclient.PatchOptions{
+		FieldManager: "cluster-network-operator/ingress_controller",
+	})
 }


pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -64,7 +63,8 @@ var _ reconcile.Reconciler = &ReconcileIngressConfigs{}
 // ReconcileIngressConfigs watches for updates to ingress controller configuration
 // and sets the network policy related labels on the openshift-host-network namespace
 type ReconcileIngressConfigs struct {
-	client crclient.Client
+	client cnoclient.Client
+	mgr    manager.Manager
```
kyrtapz, line 67: Instead of adding the manager to the controller you can extract the client from `cnoclient.Client` with `client.Default().CRCClient()`

kyrtapz (commented):
Overall looks good.
We should probably also address the same issue here https://github.com/openshift/cluster-network-operator/blob/422456d26118e22d5fb0c1d8be372b59a34e588f/pkg/controller/statusmanager/kube.go#L61

kyrtapz:
> Overall looks good. We should probably also address the same issue here
> 
We've decided to do a follow up tracked here: https://issues.redhat.com/browse/OCPBUGS-29288

pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -142,5 +142,7 @@ func (r *ReconcileIngressConfigs) updatePolicyGroupLabelOnNamespace(ctx context.
 
 	newNamespace.SetLabels(existingLabels)
 
-	return r.client.Patch(context.TODO(), newNamespace, crclient.MergeFrom(namespace))
+	return r.client.Patch(context.TODO(), newNamespace, crclient.MergeFrom(namespace), &crclient.PatchOptions{
```
pperiyasamy, line 145: @npinaeva @kyrtapz This seems to be using MergePatchType, shouldn't this be changed to use ApplyPatchType for SSA to kick in ?

PR #2246, opened 2024-02-01
OCPBUGS-27222,NP-874: Bump openshift API

kyrtapz:
Includes moving the live migration to default.
Additionally it pulls the CNCC change, cc: @dulek 

Commands ran:
```
go get github.com/openshift/api@master
go mod tidy && go mod vendor
./hack/update-codegen.sh
```

PR #2263, opened 2024-02-08
[release-4.15] OCPBUGS-28902: Fix CloudPrivateIPConfig CRD in common

dulek:
`update-codegen.sh` only updated the top-level CRD as 001-crd.yaml in `/bindata/cloud-network-config-controller` and skipped the one in `common` subdirectory. This made #2246 to not actually update the CRD.

This commit makes sure the script is updated to put the file into the correct subdirectory.


dulek:
@kyrtapz: Hopefully this is the last one!

PR #2270, opened 2024-02-12
SDN-4163: Add ipsec state to telemetry

JoshSalomon:
This PR adds a metric for the ipsec mode. The metric created is a Gauge named `cluster_network_ipsec_state` with a constant value of 1 and 2 labels:
1. "mode" which lists the ipsec mode, one of `Disabled` `External` or `Full`
2. "is_legacy_api" which states the flavor of the API used - either `true`
   for legacy API (up to OCP V4.14) or `false` for the new API (OCP 4.15+)
    -- If the cluster uses non ovn-k network (a network that does not
        support ipsec) the value of this label is "N/A - ipsec not supported (non-OVN network)"

Then a Prometheus rule  is created to remove other labels and this rule `openshift:cluster_network_ipsec_state:sum` will be collected in the telemetry

Jira ticket https://issues.redhat.com/browse/MON-3707

### DIFF
--- a/bindata/cluster-network-operator/prometheus.yaml
+++ b/bindata/cluster-network-operator/prometheus.yaml
@@ -1,0 +1,18 @@
+##
+# This file is not applied correctly during cluster installation when located in
+# the manifests directory, this is the reason why it is located in the bindata directory.
+##
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+    name: openshift-network-operator-ipsec-rules
+    namespace: openshift-network-operator
+spec:
+    groups:
+    - name: openshift-network.rules
+      rules:
+      - expr: |-
+          sum by (mode,is_legacy_api) (
+            openshift_network_operator_ipsec_state{namespace=~"openshift-network-operator"}
+          )
+        record: openshift:openshift_network_operator_ipsec_state:sum
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -22,6 +22,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/network"
 	"github.com/openshift/cluster-network-operator/pkg/platform"
+	ipsecMetrics "github.com/openshift/cluster-network-operator/pkg/util/ipsec"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 	"github.com/openshift/library-go/pkg/operator/events"
 
@@ -348,6 +349,7 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	updateIPsecMetric(&newOperConfig.Spec)
 	// once updated, use the new config
 	operConfig = newOperConfig
 
@@ -548,6 +550,23 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	// so we can reconcile state again.
 	log.Printf("Operconfig Controller complete")
 	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
+}
+
+func updateIPsecMetric(newOperConfigSpec *operv1.NetworkSpec) {
+	if newOperConfigSpec == nil {
+		// spec is not initilized yet
+		klog.V(5).Infof("IPsec: << updateIPsecTelemetry, new spec is nil, skipping")
+	} else if newOperConfigSpec.DefaultNetwork.OVNKubernetesConfig == nil {
+		// non ovn-k network, ipsec is not supported
+		ipsecMetrics.UpdateIPsecMetricNA()
+	} else {
+		// ovn-k network, ipsec is supported, update the ipsec state metric
+		newOVNKubeConfig := newOperConfigSpec.DefaultNetwork.OVNKubernetesConfig
+		mode := string(network.GetIPsecMode(newOVNKubeConfig))
+		legacyAPI := network.IsIPsecLegacyAPI(newOVNKubeConfig)
+
+		ipsecMetrics.UpdateIPsecMetric(mode, legacyAPI)
+	}
 }
 
 func reconcileOperConfig(ctx context.Context, obj crclient.Object) []reconcile.Request {
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -518,12 +518,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	return objs, progressing, nil
 }
 
-// getIPsecMode return the ipsec mode accounting for upgrade scenarios
+// GetIPsecMode return the ipsec mode accounting for upgrade scenarios
 // Find the IPsec mode from Ipsec.config
 // Ipsec.config == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
 // ipsecConfig.mode == "" (bw compatibility) || ipsec.Config == Full ==> ipsec is enabled for NS and EW
 // ipsecConfig.mode == External ==> ipsec is enabled for NS only
-func getIPsecMode(conf *operv1.OVNKubernetesConfig) operv1.IPsecMode {
+func GetIPsecMode(conf *operv1.OVNKubernetesConfig) operv1.IPsecMode {
 	mode := operv1.IPsecModeDisabled // Should stay so if conf.IPsecConfig == nil
 	if conf.IPsecConfig != nil {
 		if conf.IPsecConfig.Mode != "" {
@@ -535,6 +535,11 @@ func getIPsecMode(conf *operv1.OVNKubernetesConfig) operv1.IPsecMode {
 
 	klog.V(5).Infof("IPsec: after looking at %+v, ipsec mode=%s", conf.IPsecConfig, mode)
 	return mode
+}
+
+// IsIPsecLegacyAPI returns true if the old (pre 4.15) IPsec API is used, and false otherwise.
+func IsIPsecLegacyAPI(conf *operv1.OVNKubernetesConfig) bool {
+	return conf.IPsecConfig == nil || conf.IPsecConfig.Mode == ""
 }
 
 // shouldRenderIPsec method ensures the have following IPsec states for upgrade path from 4.14 to 4.15 or later versions:
@@ -558,7 +563,7 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	isIpsecUpgrade := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.LegacyIPsecUpgrade
 	isOVNIPsecActive := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.OVNIPsecActive
 
-	mode := getIPsecMode(conf)
+	mode := GetIPsecMode(conf)
 
 	// On upgrade, we will just remove any existing ipsec deployment without making any
 	// change to them. So during upgrade, we must keep track if IPsec MachineConfigs are
@@ -975,7 +980,7 @@ func getOVNEncapOverhead(conf *operv1.NetworkSpec) uint32 {
 	const geneveOverhead = 100
 	const ipsecOverhead = 46 // Transport mode, AES-GCM
 	var encapOverhead uint32 = geneveOverhead
-	mode := getIPsecMode(conf.DefaultNetwork.OVNKubernetesConfig)
+	mode := GetIPsecMode(conf.DefaultNetwork.OVNKubernetesConfig)
 	if mode == operv1.IPsecModeFull {
 		encapOverhead += ipsecOverhead
 	}
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -124,6 +124,12 @@ func Render(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.BootstrapResult
 	}
 	objs = append(objs, o...)
 
+	o, err = renderCNO(manifestDir)
+	if err != nil {
+		return nil, progressing, err
+	}
+	objs = append(objs, o...)
+
 	log.Printf("Render phase done, rendered %d objects", len(objs))
 	return objs, progressing, nil
 }
@@ -822,6 +828,17 @@ func renderNetworkPublic(manifestDir string) ([]*uns.Unstructured, error) {
 	manifests, err := render.RenderDir(filepath.Join(manifestDir, "network", "public"), &data)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to render network/public manifests")
+	}
+	return manifests, nil
+}
+
+// renderCNO renders the common objects in the cluster-network-operator directory
+func renderCNO(manifestDir string) ([]*uns.Unstructured, error) {
+	data := render.MakeRenderData()
+
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "cluster-network-operator"), &data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render cluster-network-operator manifests")
 	}
 	return manifests, nil
 }
--- a/pkg/util/ipsec/metrics.go
+++ b/pkg/util/ipsec/metrics.go
@@ -1,0 +1,44 @@
+package ipsec
+
+import (
+	"strconv"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog/v2"
+)
+
+const (
+	cnoNamespace  = "openshift_network_operator"
+	ipsecStateNA  = "N/A - ipsec not supported (non-OVN network)"
+	ipsecDisabled = "Disabled"
+)
+
+var (
+	ipsecStateGauge *metrics.GaugeVec
+)
+
+func init() {
+	ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
+		Namespace: cnoNamespace,
+		Name:      "ipsec_state",
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster, " +
+			"and the API that invoked it, legacy (pre OCP 4.14) or new. " +
+			"In case the network doesn't support ipsec (non-OVN network), " +
+			"the 'is_legacy_api' value is set to '" + ipsecStateNA + "'.",
+	}, []string{"mode", "is_legacy_api"})
+	legacyregistry.MustRegister(ipsecStateGauge)
+}
+
+func UpdateIPsecMetric(state string, isLegacyAPI bool) {
+	klog.V(5).Infof("IPsec mode: %s, isLegacyAPI: %v",
+		state, isLegacyAPI)
+	ipsecStateGauge.Reset()
+	ipsecStateGauge.WithLabelValues(state, strconv.FormatBool(isLegacyAPI)).Set(1)
+}
+
+func UpdateIPsecMetricNA() {
+	klog.V(5).Infof("IPsec is not supported by non-OVN network (disabled)")
+	ipsecStateGauge.Reset()
+	ipsecStateGauge.WithLabelValues(ipsecDisabled, ipsecStateNA).Set(1)
+}


pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,50 @@
...
+	} else if ipsecConfig.Mode == "" {
+		mode = string(operv1.IPsecModeFull)
+		flavor = "4.14"
+	} else {
+		mode = string(ipsecConfig.Mode)
+		flavor = "4.15"
+	}
+	if mode != recordedMode || flavor != recordedFlavor {
+		klog.V(5).Infof("IPsec Telemetry: (%s, %s), writing to %v", mode, flavor, ipsecStateGauge)
+		ipsecStateGauge.WithLabelValues(mode, flavor).Set(1)
```
jan--f, line 46: I'm neither an expert on instrumentation nor sure whether the following is intentional.
`WithLabelValues` will create a new time series if you call it with a never seen before set of label values. So if the expectation here is to always have a single timeseries, you'll need to reset the vector before set. Otherwise you'll end up with multiple series if those label values change.

jan--f:
It's not required but good practice to create a recording rule for telemetry metrics. Something like `sum by(mode, apiFlavor) cluster_network_ipsec_state` for example. This guards telemetry against later additions to the metrics label set, since those additional label dimensions would be aggregated away.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,50 @@
...
+	recordedMode, recordedFlavor string = "", ""
+)
+
+// RegisterIpsecMetrics in the global registry
+func RegisterIpsecMetrics() {
+	registerMetrics.Do(func() {
+		ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
+			Name: "cluster_network_ipsec_state",
+			Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster",
+		}, []string{"mode", "apiFlavor"})
```
jan--f, line 25:
```suggestion
		}, []string{"mode", "api_flavor"})
```
Prometheus has a strong preference for snake_case oder camelCase.

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: @jan--f Thanks - I am not a Prometheus expert and I wondered how to keep a single value (It was not critical,  since in most cases we will have 1 or 2 values, but what I want is always a single value).

JoshSalomon:
@jan--f - This is my first experience with telemetry, so this can be a beginner question. I think I know how to create the prometheus rules, but I couldn't find in the documentation how it is connected to the telemetry - I see that in the file 0000_50_cluster-monitoring-operator_04-config.yaml I add the counter name, so should I do anything else after I create the rule?
(I thought of adding a directory (bindata/network/ipsec) and add a file rules.yaml in this directory - but how should I connect this file and the rule inside into the telemetry?)

pkg/util/ipsec/ipsec_telemetry.go
jan--f: Right, this is unfortunately a discouraged but widespread practice. Prometheus best-practice would recommend to always emit all possible label dimensions and set 0 for the ones not observed.
That can however be expensive. For larger metric sets like this a custom collector is usually used, like in https://github.com/kubernetes/kubernetes/blob/20d0ab7ae808aaddb1556c3c38ca0607663c50ac/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L299

I suppose this metric is expected to rarely change in a running cluster? So the reset way is probably just fine here.

simonpasquier: As @jan--f  said, the preferred option would be to expose all label values with a value of 1 for the mode being enabled:

```
cluster_network_ipsec_state{mode="disabled"} 0
cluster_network_ipsec_state{mode="external"} 1
cluster_network_ipsec_state{mode="full"} 0
```

Using Reset() should work fine though it's more error-prone. The other option if you want to keep only 1 series would be to use the custom collector pattern (https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#hdr-Custom_Collectors_and_constant_Metrics).

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+}
+
+func UpdateIpsecTelemetry(ipsecConfig *operv1.IPsecConfig) {
+	var flavor, mode string
+	klog.V(5).Infof("IPsec Telemetry: %v", ipsecConfig)
+	RegisterIpsecMetrics()
+	if ipsecConfig == nil {
+		mode = string(operv1.IPsecModeDisabled)
+		flavor = "4.14"
```
simonpasquier, line 36: What is the `flavor` label about? Is it the OCP version?

JoshSalomon:
@dougbtv @kyrtapz - Can I get a review for this PR - it blocks my next PR of adding the counter to telemetry.

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: Yes - the API has changed between OCP 4.14 and OCP 4.15

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: @jan--f - this metric is indeed expected to change very rarely in production clusters, once you enable ipsec you rarely disable it (assuming you had a business reason to enable it, it doesn't go away).

JoshSalomon: @simonpasquier - for a metric as simple as this which rarely changes, can I leave the current implementation? I believe the option you suggest will just add more lines of code for something really simple and will make the code cumbersome.

pkg/util/ipsec/ipsec_telemetry.go
simonpasquier: I see no reason to have it as a label since the cluster version is already available in Telemeter.

pkg/util/ipsec/ipsec_telemetry.go
simonpasquier: I'm not sure that the code would be more complex with a custom collector. It's your call since you maintain that piece of code ;)

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: I probably misled you in the previous response. There are 2 flavors to the API according to the version in which the API was invoked, but for backward compatibility we keep the API version during an upgrade until the API is invoked again, so you can find an instance of the old API (4.14) in newer clusters. The reason to keep the API flavor is in order to understand when we can drop the support for the old API flavor, which is non-trivial.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -348,6 +349,7 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		}
 	}
 
+	UpdateIpsecTelemetry(&newOperConfig.Spec)
```
pperiyasamy, line 352: I think it would be good to introduce new field in `ReconcileOperConfig` struct to store IPsec mode and update ipsec metric when it changes to a new value.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,51 @@
...
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog/v2"
+)
+
+var (
+	registerMetrics sync.Once
+	ipsecStateGauge *metrics.GaugeVec
+
+	recordedMode, recordedFlavor string = "", ""
```
pperiyasamy, line 16: As I said, this must be moved under `ReconcileOperConfig`

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,51 @@
...
+
+func UpdateIpsecTelemetry(ipsecConfig *operv1.IPsecConfig) {
+	var flavor, mode string
+	klog.V(5).Infof("IPsec Telemetry: %v", ipsecConfig)
+	RegisterIpsecMetrics()
+	if ipsecConfig == nil {
+		mode = string(operv1.IPsecModeDisabled)
+		flavor = "4.14"
+	} else if ipsecConfig.Mode == "" {
+		mode = string(operv1.IPsecModeFull)
```
pperiyasamy, line 38: I'd reused existing method [getIPsecMode](https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/ovn_kubernetes.go#L535) to retrieve the ipsec mode.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,51 @@
...
+		flavor = "4.14"
+	} else {
+		mode = string(ipsecConfig.Mode)
+		flavor = "4.15"
```
pperiyasamy, line 42: what's the significance of exposing flavor in the metric ? I'm not sure if it's good idea to hardcode these values with 4.15 or 4.14. Did you consider to use https://github.com/openshift/cluster-network-operator/blob/master/vendor/github.com/openshift/api/operator/v1/types.go#L123 ?

JoshSalomon: I am not sure I understand this - we have 2 versions of the API, and they changed between 4.14 and 4.15. I believe it is good to know if someone is still using the older version. I can call it "old" and "new", or any other agreed constant names, but if we collect telemetry, it is an important piece of information. I don't understand how the `Version` field can help me - since I need constants and not something that can change according to the version in which the code is executed.

pkg/util/ipsec/ipsec_telemetry.go
pperiyasamy: yes, you could just make `getIPsecMode` method as public and use it from here. I prefer to have new method for returning flavor.

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: This is an option - I am not sure if booleans are supported by Prometheus labels, but it can be "true" or "false" if not. I will do it.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,47 @@
...
+		}, []string{"mode", "is_legacy_api"})
+		legacyregistry.MustRegister(ipsecStateGauge)
+	})
+}
+
+func UpdateIpsecTelemetry(state string, isLegacyAPI bool) {
+	klog.V(5).Infof("IPsec mode(prev): %s(%s), isLegacyAPI(prev): %v(%s)",
+		state, recordedMode.state, isLegacyAPI, recordedMode.legacyAPI)
+	isLegacyAPIStr := strconv.FormatBool(isLegacyAPI)
+	RegisterIpsecMetrics()
```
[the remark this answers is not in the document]
JoshSalomon: registerMetrics is an instance of sync.Once - it runs only once.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,47 @@
...
+	if state != recordedMode.state || isLegacyAPIStr != recordedMode.legacyAPI {
```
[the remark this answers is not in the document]
JoshSalomon: It is not after upgrade - the API could be invoked from `ipsec: {}` to `ipsec.mode: "Full"`
from a functional perspective, it may not matter but I believe it is good for metrics purposes. Especially since there is no way back from the new API to the legacy one.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -550,6 +552,25 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+
+	newOVNKubeConfig := newOperConfigSpec.DefaultNetwork.OVNKubernetesConfig
+	klog.V(5).Infof("IPsec: calling UpdateIpsecTelemetry, %v", newOVNKubeConfig.IPsecConfig)
+	mode := string(network.GetIPsecMode(newOVNKubeConfig))
+	legacyAPI := network.IsIPsecLegacyAPI(newOVNKubeConfig)
+
+	ipsecTelemetry.UpdateIpsecTelemetry(mode, legacyAPI)
+
+	klog.V(5).Infof("IPsec: << UpdateIpsecTelemetry")
+	// Don't record telemetry when the IPsecConfig hasn't changed
```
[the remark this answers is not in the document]
JoshSalomon: will fix this (tomorrow)

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -550,6 +552,25 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
 }
 
+func UpdateIpsecTelemetry(newOperConfigSpec *operv1.NetworkSpec) {
+	klog.V(5).Infof("IPsec: >> UpdateIpsecTelemetry new: %v", newOperConfigSpec)
```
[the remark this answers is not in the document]
JoshSalomon: will fix this (tomorrow)

pkg/util/ipsec/ipsec_telemetry.go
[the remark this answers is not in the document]
pperiyasamy: ok, but this is still bit confusing calling register from update metrics method. 
Can you at least do it via module's init method if it can not be registered from operconfig controller ?

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,47 @@
...
+
+var (
+	registerMetrics sync.Once
+	ipsecStateGauge *metrics.GaugeVec
+
+	recordedMode ipsecAPIMode // initial values are empty strings
+)
+
+// RegisterIpsecMetrics in the global registry
+func RegisterIpsecMetrics() {
```
pperiyasamy, line 25: Nit, replace Ipsec with IPsec, this is the pattern being used all over the places.

pperiyasamy (commented):
Please discuss with @martinkennelly, @jcaamano about where to maintain current IPsec mode.

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: Regarding the registration - this is not something I invented - some modules (e.g. pkg/cmd/checkendpoints/controller/metrics.go) use this method, but some (e.g. vendor/github.com/grpc-ecosystem/go-grpc-prometheus/client.go) use the init module, I can move it. Good catch on the name (should be IPsec)

JoshSalomon:
> Please discuss with @martinkennelly, @jcaamano about where to maintain current IPsec mode.

@martinkennelly @jcaamano - What I hold is not the ipsec mode is the ipsec *last recorded* mode which has more legal values than ipsec mode (it can be empty on cluster init), moreover it catches the situation where there is nothing to do (disabled mode) which is not caught in the operator code, since it requires no-op. So while in normal work I could have placed this code where the ipsec change state code is, and call the metric change on ipsec change, on startup this is not the case, and I see no reason to clutter the operator code more than it should be, for some metric optimization code which can be encapsulated in a small isolated piece of code.

bindata/network/network-metrics/ipsec-rules.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: monitoring.coreos.com/v1
```
martinkennelly, line 1: is this for sdn + ovn-k? if its just for ovn-k and you want to expose this for every deployment model we have, id put this in `bindata/network/ovn-kubernetes/common/alert-rules.yaml`. If its not meant for managed services and just non-managed standalone clusters, you can adjust where to put it.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+	recordedState     string
+	recordedLegacyAPI string
+)
+
+func init() {
+	ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
+		Name: "cluster_network_ipsec_state",
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster",
+	}, []string{"mode", "is_legacy_api"})
+	legacyregistry.MustRegister(ipsecStateGauge)
```
martinkennelly, line 23: I am confused - how will the metrics be consumed from CNO? 
We dont expose any endpoint for Prometheus to scrape this metric AFAIK.

JoshSalomon: It will be consumed from the telemetry module - the Prometheus query "openshift:cluster_network_ipsec_state:sum" will be gathered by the telemetry agent

martinkennelly: I mean, how will prometheus actually get the metrics info - i dont think we expose a server for prometheus agent to scrape the metrics. 
Did you see the metrics being exposed in OCP console with this PR?

JoshSalomon: Yep - 
image

martinkennelly: I was wrong - its serving on 9104. Thanks!

bindata/network/network-metrics/ipsec-rules.yaml
martinkennelly, line 1: i think move to manifests file and name `07-prometheus-rules.yaml`. WDYT?

This dir is for network metrics daemon.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -550,6 +552,19 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 	return reconcile.Result{RequeueAfter: ResyncPeriod}, nil
 }
 
+func updateIPsecTelemetry(newOperConfigSpec *operv1.NetworkSpec) {
```
martinkennelly, line 555: `updateIPsecMetric(`

Theres no telemetry update here - just updating a metric. Gathering of telemetry based on this metric is outside the scope of this PR.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+
+var (
+	ipsecStateGauge *metrics.GaugeVec
+
+	recordedState     string
+	recordedLegacyAPI string
+)
+
+func init() {
+	ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
```
martinkennelly, line 19: please fill in fields `namespace`, `subsystem` and `name`. Its how the full name is usually constructed in OCP.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+		Name: "cluster_network_ipsec_state",
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster",
```
martinkennelly, line 21: what about a description of the legacy api label?

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog/v2"
+)
+
+var (
+	ipsecStateGauge *metrics.GaugeVec
+
+	recordedState     string
```
martinkennelly, line 14: i dont see the need for this. Just update the metric with the current state. Using these vars just made the code harder to read.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+	ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
+		Name: "cluster_network_ipsec_state",
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster",
+	}, []string{"mode", "is_legacy_api"})
+	legacyregistry.MustRegister(ipsecStateGauge)
+	recordedState = ""
+	recordedLegacyAPI = ""
+}
+
+func UpdateIPsecTelemetry(state string, isLegacyAPI bool) {
```
martinkennelly, line 28: ` UpdateIPsecMetric`

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
...
+	klog.V(5).Infof("IPsec mode(prev): %s(%s), isLegacyAPI(prev): %v(%s)",
+		state, recordedState, isLegacyAPI, recordedLegacyAPI)
+	isLegacyAPIStr := strconv.FormatBool(isLegacyAPI)
```
martinkennelly, line 31: i think you need another state besides true and false - if ipsec is disabled, then the third state should be `unknown` otherwise it looks strange defining is it legacy api or not when its not even configured. WDYT?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -550,6 +552,19 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
...
+	if newOperConfigSpec == nil || newOperConfigSpec.DefaultNetwork.OVNKubernetesConfig == nil {
```
martinkennelly, line 556: what if the user switches to a new CNI from ovn-k? the metrics will still report old values indicating ipsec is enabled. You should reset the metric for this scenario.

pkg/util/ipsec/ipsec_telemetry.go
```diff
@@ -0,0 +1,39 @@
+package ipsec
```
martinkennelly, line 1: rename file `metrics.go`

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: IIUC - this causes having more load on Prometheus - as Prometheus holds a time series it does not collapse identical values - this is just another value in the time series, and this is not what we want to keep even if we can collapse this in the telemetry collection. Actually, this part of the code was reviewed (with comments) and approved by the @jan--f from the telemetry team.
@jan--f does it make sense just to write the same value to Prometheus over and over again when all we need is the latest state and not a time series?

pkg/util/ipsec/ipsec_telemetry.go
JoshSalomon: I see that namespace is hardly used and this metric is cluster level and not really namespaced (in meaning that it is limited to a specific namespace and can differ between namespaces) so I just used only subsystem and name (which is the common way to create metrics as far as I could find)

bindata/network/network-metrics/ipsec-rules.yaml
JoshSalomon: The file was not applied during the installation in the manifests directory, I am not sure why. On the other hand, this is kind of network metric...

pkg/util/ipsec/ipsec_telemetry.go
martinkennelly: Its cluster level but in order to easily inform the consumer where it came from, we populate the `namespace` field.

pkg/util/ipsec/ipsec_telemetry.go
martinkennelly: Its just updating what the prom server (presented by CNO here) presents to the prometheus scraper - so if the value doesnt change, its a no-op and the server will present the same data. It doesnt increase the load on the prometheus backend which store this data.

martinkennelly: I think it just makes it easier to read this code without affecting performance much. It wont increase the load or create additional artifacts in the prometheus backend db (in the prometheus namespace)

bindata/cluster-network-operator/prometheus.yaml
```diff
@@ -0,0 +1,18 @@
+##
+# This file is not applied correctly during cluster installation when located in
+# the monifests directory, this is the reason why it is located in the bindata directory.
```
martinkennelly, line 3: monifests -> manifests

martinkennelly (approved):

kyrtapz:
@JoshSalomon the `e2e-gcp-sdn` job reports that CNO is crashing:
`level=error msg=Cluster operator network Degraded is True with ReconcileError: Panic detected: runtime error: invalid memory address or nil pointer dereference`
https://storage.googleapis.com/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2270/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-sdn/1757339435860496384/build-log.txt

/test e2e-gcp-sdn

PR #2274, opened 2024-02-13
OCPBUGS-29288: Ensure proper deprecation for the default field manager

pperiyasamy:
This PR does the following to remove deprecated field manager clean up code in `apply#ApplyObject` method and also ensures default field manager is set.

- Set default field manager value (via rest.Config's UserAgent) for all of the object's managed fields. Here is sample PR https://github.com/openshift/cluster-network-operator/pull/2281.
- Remove the entire `isDepFieldManagerCleanupNeeded` code from `apply#ApplyObject `for 4.16.
 (because p.1 and p.2 takes care setting up field manager already).


pkg/names/names.go
```diff
@@ -224,3 +224,6 @@ const DefaultClusterName = "default"
 
 // DashboardNamespace is the namespace where dashboards are created
 const DashboardNamespace = "openshift-config-managed"
+
+// DefaultUserAgent is the default UserAgent set on the clients
+const DefaultUserAgent = "cluster-network-operator"
```
kyrtapz, line 229: This is the same value as the fieldManager we've determined to be deprecated,  I don't think this will work in earlier releases where we don't plane to remove the deprecation cleanup logic.

pperiyasamy: ok, would change it to `network-operator` then, so this makes stale fields which are managed by deprecated field manager (`cluster-network-operator`) automatically get removed when this version of code runs, am i right ?

pperiyasamy:
With the change, I can notice both default field manager and overridden field manager are set appropriately on the objects.

```
# kubectl get networks.operator.openshift.io cluster -o yaml --show-managed-fields
apiVersion: operator.openshift.io/v1
kind: Network
metadata:
  creationTimestamp: "2024-02-21T13:40:45Z"
  generation: 92
  managedFields:
  - apiVersion: operator.openshift.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:clusterNetwork: {}
        f:defaultNetwork:
          f:ovnKubernetesConfig:
            f:egressIPConfig: {}
            f:gatewayConfig:
              f:ipv4: {}
              f:ipv6: {}
            f:genevePort: {}
            f:mtu: {}
            f:policyAuditConfig:
              f:destination: {}
              f:maxFileSize: {}
              f:rateLimit: {}
              f:syslogFacility: {}
          f:type: {}
        f:deployKubeProxy: {}
        f:disableMultiNetwork: {}
        f:disableNetworkDiagnostics: {}
        f:logLevel: {}
        f:managementState: {}
        f:observedConfig: {}
        f:operatorLogLevel: {}
        f:serviceNetwork: {}
        f:unsupportedConfigOverrides: {}
        f:useMultiNetworkPolicy: {}
      f:status:
        f:readyReplicas: {}
        f:version: {}
    manager: network-operator/operconfig
    operation: Apply
    time: "2024-02-21T13:41:05Z"
  - apiVersion: operator.openshift.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:spec:
        f:clusterNetwork: {}
        f:defaultNetwork:
          f:type: {}
        f:disableNetworkDiagnostics: {}
        f:managementState: {}
        f:observedConfig: {}
        f:serviceNetwork: {}
        f:unsupportedConfigOverrides: {}
      f:status:
        f:readyReplicas: {}
    manager: network-operator/clusterconfig
    operation: Apply
    time: "2024-02-21T13:44:39Z"
  - apiVersion: operator.openshift.io/v1
    fieldsType: FieldsV1
    fieldsV1:
      f:status:
        f:conditions: {}
        f:version: {}
    manager: network-operator
    operation: Update
    time: "2024-02-21T13:58:32Z"
  name: cluster
  resourceVersion: "29837"
  uid: 83f1b9c1-bc3f-4057-8683-c1cd4f5988d5
spec:
  clusterNetwork:
  - cidr: 10.128.0.0/14
    hostPrefix: 23
  defaultNetwork:
    ovnKubernetesConfig:
      egressIPConfig: {}
      gatewayConfig:
        ipv4: {}
        ipv6: {}
        routingViaHost: false
      genevePort: 6081
      ipsecConfig:
        mode: Disabled
      mtu: 8901
      policyAuditConfig:
        destination: "null"
        maxFileSize: 50
        maxLogFiles: 5
        rateLimit: 20
        syslogFacility: local0
    type: OVNKubernetes
  deployKubeProxy: false
  disableMultiNetwork: false
  disableNetworkDiagnostics: false
  logLevel: Normal
  managementState: Managed
  observedConfig: null
  operatorLogLevel: Normal
  serviceNetwork:
  - 172.30.0.0/16
  unsupportedConfigOverrides: null
  useMultiNetworkPolicy: false
status:
  conditions:
  - lastTransitionTime: "2024-02-21T13:40:45Z"
    status: "False"
    type: ManagementStateDegraded
  - lastTransitionTime: "2024-02-21T13:48:36Z"
    status: "False"
    type: Degraded
  - lastTransitionTime: "2024-02-21T13:40:45Z"
    status: "True"
    type: Upgradeable
  - lastTransitionTime: "2024-02-21T13:58:32Z"
    status: "False"
    type: Progressing
  - lastTransitionTime: "2024-02-21T13:41:05Z"
    status: "True"
    type: Available
  readyReplicas: 0
  version: 4.15.0-0.ci.test-2024-02-21-132851-ci-ln-q9kdwzb-latest


# kubectl get namespace openshift-host-network -o yaml --show-managed-fields
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    openshift.io/description: Namespace for enabling network policy specification
      for host network traffic. Can be used to allow access to or from host network
      components
    openshift.io/sa.scc.mcs: s0:c23,c7
    openshift.io/sa.scc.supplemental-groups: 1000520000/10000
    openshift.io/sa.scc.uid-range: 1000520000/10000
    workload.openshift.io/allowed: management
  creationTimestamp: "2024-02-21T13:41:20Z"
  labels:
    kubernetes.io/metadata.name: openshift-host-network
    policy-group.network.openshift.io/host-network: ""
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:openshift.io/description: {}
          f:workload.openshift.io/allowed: {}
        f:labels:
          f:policy-group.network.openshift.io/host-network: {}
        f:ownerReferences:
          k:{"uid":"83f1b9c1-bc3f-4057-8683-c1cd4f5988d5"}: {}
    manager: network-operator/operconfig
    operation: Apply
    time: "2024-02-21T13:41:20Z"
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          f:openshift.io/sa.scc.mcs: {}
          f:openshift.io/sa.scc.supplemental-groups: {}
          f:openshift.io/sa.scc.uid-range: {}
    manager: cluster-policy-controller
    operation: Update
    time: "2024-02-21T13:41:20Z"
  name: openshift-host-network
  ownerReferences:
  - apiVersion: operator.openshift.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Network
    name: cluster
    uid: 83f1b9c1-bc3f-4057-8683-c1cd4f5988d5
  resourceVersion: "3645"
  uid: 46db7baa-6722-44af-84af-7156973a5f27
spec:
  finalizers:
  - kubernetes
status:
  phase: Active

```

pkg/apply/apply.go
```diff
@@ -106,11 +106,10 @@ func ApplyObject(ctx context.Context, client cnoclient.Client, obj Object, subco
 		}
 	}
 
-	fieldManager := "cluster-network-operator"
+	fieldManager := GetFieldManager(subcontroller)
```
npinaeva: wait a second, I don't think we want to change fieldManager here, because it will cause the same problems when migrating from one fieldManager to another.
I think we wanted to leave Apply field manager as it was (as it is explicitly specificed), and only update field manager used by the client?

pperiyasamy: @npinaeva this change is made to have uniform prefix on the field manger value , it also does force apply on the object. so where do you see a problem ?

npinaeva: fields that are owned now by "cluster-network-operator/subcontroller" will be renamed to "network-operator/subcontroller" which requires one more field manager deprecation and we didn't want that?
cc @kyrtapz

kyrtapz: Yeah, I don't think we need to unify the prefix in this change. We can do that separately in the future if required.

pperiyasamy: Ok @npinaeva / @kyrtapz , as discussed, retaining code to use "cluster-network-operator/subcontroller".

npinaeva:
looks safe now :) lgtm from my side (not a label since tests haven't passed yet, but code looks fine)

PR #2280, opened 2024-02-20
OCPBUGS-29341: Run dhcp-daemon pods as system-node-critical priority

pliurh:
With system-node-critical priority, dhcp-daemon pods will not get evicted before workloads.


s1061123 (approved):
/lgtm

PR #2282, opened 2024-02-20
AUTH-482: set required-scc for openshift workloads

liouk:


bindata/network/node-identity/managed/node-identity.yaml
```diff
@@ -29,6 +29,7 @@ spec:
       annotations:
         hypershift.openshift.io/release-image: {{.ReleaseImage}}
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        openshift.io/required-scc: hostnetwork-v2
```
kyrtapz, line 32: node-identity in managed is not a hostNetworked pod, this should work with restricted-v2?

liouk: Thanks for this -- our monitor tests probably run only on self-hosted therefore I have missed this case. Fixed :+1:

PR #2285, opened 2024-02-28
OCPBUGS-27925: Remove interconnect upgrade logic, single-zone YAMLs and everything related to OVNK centralized architecture

jluhrsen:
This is a manual cherry-pick of [SDN-4061](https://github.com/openshift/cluster-network-operator/pull/2101/commits


jluhrsen:
/hold
this may not be what we want to do, but I want to use a presubmit here to see if these changes are what resolves
the CIDR

PR #2288, opened 2024-03-01
OCPBUGS-30103: ensure local networking deployments within hypershift use the client side load balancer to be resilient to control plane node failures

relyt0925:
In environments with no cloud load balancer health checking and removing bad control plane endpoints: requests to the control plane from data plane components like kube-proxy and mutlus using the control plane dns name directly will see intermittent failures. In hypershift: a node local haproxy load balancer runs that health checks and removes failed control plane ingress points dynamically. When the components move to using this: control plane failures do not lead to further failures in the data plane starting new workload. fixes: https://issues.redhat.com/browse/OCPBUGS-30103


relyt0925:
Currently seeing failure in e2e-hypershift-ovn test in multus admission controller on
```
I0306 15:21:55.445552       1 webhook.go:155] AdmissionReview request allowed: Network Attachment Definition '{"bridge":"watestbr0","cniVersion":"0.3.0","forceAddress":false,"hairpinMode":true,"ipMasq":true,"ipam":{"exclude":["192.168.2.225/30"],"log_file":"/tmp/whereabouts.log","log_level":"debug","range":"192.168.2.225/29","type":"whereabouts"},"isDefaultGateway":true,"name":"whereaboutstestbridge","type":"bridge"}' is valid

I0306 15:21:55.446155       1 webhook.go:118] validating network config spec: { "cniVersion": "0.3.0", "name": "whereaboutsexample", "type": "bridge", "bridge": "watestbr1", "isDefaultGateway": true, "forceAddress": false, "ipMasq": true, "hairpinMode": true, "ipam": { "type": "whereabouts", "sleep_for_race": 5, "range": "192.168.2.240/28", "log_file": "/tmp/whereabouts.log", "log_level": "debug" } }


I0306 15:21:55.446217       1 webhook.go:142] spec is not a valid network config list: error parsing configuration list: no 'plugins' key - trying to parse into standalone config

```

relyt0925:
For now we can proceed forward with IBM Cloud (need this for reliability improvements as well in 4.14) and then ultimately can remove the platform check when we get confirmation all of Red Hat's hypershift cluster certs are updated with the following ip address in this pr:
https://github.com/openshift/hypershift/pull/3699

pkg/network/kube_proxy.go
```diff
@@ -206,8 +209,16 @@ func renderStandaloneKubeProxy(conf *operv1.NetworkSpec, bootstrapResult *bootst
 	data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
 	data.Data["KubeProxyImage"] = os.Getenv("KUBE_PROXY_IMAGE")
 	data.Data["KubeRBACProxyImage"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
-	data.Data["KUBERNETES_SERVICE_HOST"] = bootstrapResult.Infra.APIServers[bootstrap.APIServerDefault].Host
-	data.Data["KUBERNETES_SERVICE_PORT"] = bootstrapResult.Infra.APIServers[bootstrap.APIServerDefault].Port
+	hsc := hypershift.NewHyperShiftConfig()
+	//TODO (relyt0925): when hypershift appropriately signs kube-apiserver certificate with node local
+	//address it is my recommendation to move to using the node local loadbalancer for all workload
+	if hsc.Enabled && bootstrapResult.Infra.PlatformType == v1.IBMCloudPlatformType {
```
kyrtapz, line 215: I assumed that this should be applicable to all hypershift deployments, not only `IBMCloudPlatformType`?

pkg/network/multus.go
```diff
@@ -69,8 +70,14 @@ func renderMultusConfig(manifestDir, defaultNetworkType string, useDHCP bool, us
 	data.Data["WhereaboutsImage"] = os.Getenv("WHEREABOUTS_CNI_IMAGE")
 	data.Data["EgressRouterImage"] = os.Getenv("EGRESS_ROUTER_CNI_IMAGE")
 	data.Data["RouteOverrideImage"] = os.Getenv("ROUTE_OVERRRIDE_CNI_IMAGE")
-	data.Data["KUBERNETES_SERVICE_HOST"] = apihost
-	data.Data["KUBERNETES_SERVICE_PORT"] = apiport
+	hsc := hypershift.NewHyperShiftConfig()
+	if hsc.Enabled && bootstrapResult.Infra.PlatformType == configv1.IBMCloudPlatformType {
```
kyrtapz, line 74: I assumed that this should be applicable to all hypershift deployments, not only IBMCloudPlatformType?

kyrtapz (commented):
Thanks @relyt0925!
Overall change make sense to me.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -252,6 +252,23 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		return reconcile.Result{}, err
 	}
 
+	//Fetch Kubernetes address endpoint
```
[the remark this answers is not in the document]
relyt0925: We can: I can change it to using that way.

pkg/network/multus.go
kyrtapz: Ah, I now understand what you meant here: https://github.com/openshift/cluster-network-operator/pull/2288/files#diff-2c4ca959f28adcdad43be768016a18842630345bec246de204c7c9d8a541c0f5R213-R214
Thanks!
The only thing that seems to be missing in the hypershift PR is only missing a valid jira label.

relyt0925:
/hold

need to test new logic on IBM hypershift location

relyt0925:
/unhold

ready for review

relyt0925:
@kyrtapz just FYI the refactor is complete and tested: I guess now we wait for the ipi fixes to go in so the required tests can pass?

pkg/network/multus.go
kyrtapz: Considering that the hypershift PR is merged can we remove the `IBMCloudPlatformType` in this PR?

relyt0925: I'm worried about your existing production fleet: I would like to get confirmation that all operators are updated before we do so......

relyt0925:
@kyrtapz why I am worried to go past IBM Cloud for now

1) I am not sure when your entire existing fleet of production hypershift operators/control-plane-operators are going to update. The two key pre reqs we need to ensure happen before we move further

When we merge this: it will be expected that any deployment using the cno at this version is in an environment with a kube-apiserver that has certs signed with that node local IP. What I am not sure on at the moment is this: for existing clusters that upgrade to 4.16 (let's say that's the release this goes in at): does every certificate guarenteed to get updated? or is that at a interval based on certificate expiration (which could mean it takes a year to expire). If that guarentee could not be provided: your cno at that release would be incompatible with existing environments (it would work on all new environments created with a CPO that signs that kube-apiserver cert with that logic.

I would prefer to follow up with the IBMCloud toggle removal on a separate PR if possible but happy to drive that. This is impacting some of our customers (I need to start to begin the backport process to 4.14) so would like not to make that all go together if possible (plus I bet the backports anyways you will not want to take "globally" and want to have it scoped to IBM Cloud

relyt0925:
I can proceed to open the pr to simply remove that (a separate PR) that we could get the test signals for and hold till we know we are ready

kyrtapz:
> When we merge this: it will be expected that any deployment using the cno at this version is in an environment with a kube-apiserver that has certs signed with that node local IP. What I am not sure on at the moment is this: for existing clusters that upgrade to 4.16 (let's say that's the release this goes in at): does every certificate guarenteed to get updated? or is that at a interval based on certificate expiration (which could mean it takes a year to expire). If that guarentee could not be provided: your cno at that release would be incompatible with existing environments (it would work on all new environments created with a CPO that signs that kube-apiserver cert with that logic.

I think the certificate would be reconciled to the new one once control-plane-operator upgrades to 4.16. I am good with following up on this in a separate PR.

> I can proceed to open the pr to simply remove that (a separate PR) that we could get the test signals for and hold till we know we are ready

Thank you!

/lgtm

PR #2299, opened 2024-03-06
OCPBUGS-30299: Fix managed cluster detection on ARO

kyrtapz:
In ARO the dedicated-admin namespace does not exit. Detect `openshift-azure-operator` instead.
/cc @martinkennelly @pliurh


martinkennelly (commented):
LGTM (waiting on ci)

PR #2309, opened 2024-03-12
SDN-4607: Allow HyperShift ability to set ConfigMap through env var

bryan-cox:
This PR allows HyperShift the ability to set the CA ConfigMap and key through an environment variable. The PR also removes the service-sa annotation from services used by HyperShift since we will pass through a CA ConfigMap for the services we are using.

Fixes [SDN-4607](https://issues.redhat.com//browse/SDN-4607)

### DIFF
--- a/bindata/network/multus-admission-controller/001-service.yaml
+++ b/bindata/network/multus-admission-controller/001-service.yaml
@@ -13,7 +13,10 @@ metadata:
 {{- if .HyperShiftEnabled}}
     network.operator.openshift.io/cluster-name: {{.ManagementClusterName}}
 {{- end }}
+# Any new HCP deployment will automatically reconcile this secret in the HCP without the service-ca operator
+{{- if or (not .HyperShiftEnabled) (eq .CAConfigMap "openshift-service-ca.crt")}}
     service.alpha.openshift.io/serving-cert-secret-name: multus-admission-controller-secret
+{{- end }}
 spec:
   ports:
   - name: webhook
--- a/bindata/network/multus-admission-controller/monitor.yaml
+++ b/bindata/network/multus-admission-controller/monitor.yaml
@@ -26,8 +26,8 @@ spec:
     tlsConfig:
       ca:
         configMap:
-          key: service-ca.crt
-          name: openshift-service-ca.crt
+          key: {{.CAConfigMapKey}}
+          name: {{.CAConfigMap}}
       cert:
         secret:
           key: tls.crt
--- a/bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
@@ -56,7 +56,11 @@ metadata:
   namespace: {{.HostedClusterNamespace}}
   annotations:
     network.operator.openshift.io/cluster-name:  {{.ManagementClusterName}}
+# If the default CAConfigMap 'openshift-service-ca.crt' was used then the HostedCluster isn't reconciling the secret, so
+# we need to add the annotation so service-ca creates the secret
+{{- if eq .CAConfigMap "openshift-service-ca.crt"}}
     service.beta.openshift.io/serving-cert-secret-name: ovn-control-plane-metrics-cert
+{{- end}}
   labels:
     app: ovnkube-control-plane
 spec:
--- a/bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
@@ -23,8 +23,8 @@ spec:
     tlsConfig:
       ca:
         configMap:
-          key: service-ca.crt
-          name: openshift-service-ca.crt
+          key: {{.CAConfigMapKey}}
+          name: {{.CAConfigMap}}
       cert:
         secret:
           key: tls.crt
--- a/pkg/hypershift/hypershift.go
+++ b/pkg/hypershift/hypershift.go
@@ -29,6 +29,8 @@ var (
 	runAsUser         = os.Getenv("RUN_AS_USER")
 	releaseImage      = os.Getenv("OPENSHIFT_RELEASE_IMAGE")
 	controlPlaneImage = os.Getenv("OVN_CONTROL_PLANE_IMAGE")
+	caConfigMap       = os.Getenv("CA_CONFIG_MAP")
+	caConfigMapKey    = os.Getenv("CA_CONFIG_MAP_KEY")
 )
 
 const (
@@ -92,9 +94,19 @@ type HyperShiftConfig struct {
 	RelatedObjects    []RelatedObject
 	ReleaseImage      string
 	ControlPlaneImage string
+	CAConfigMap       string
+	CAConfigMapKey    string
 }
 
 func NewHyperShiftConfig() *HyperShiftConfig {
+	if caConfigMap == "" {
+		caConfigMap = "openshift-service-ca.crt"
+	}
+
+	if caConfigMapKey == "" {
+		caConfigMapKey = "service-ca.crt"
+	}
+
 	return &HyperShiftConfig{
 		Enabled:           hyperShiftEnabled(),
 		Name:              name,
@@ -102,6 +114,8 @@ func NewHyperShiftConfig() *HyperShiftConfig {
 		RunAsUser:         runAsUser,
 		ReleaseImage:      releaseImage,
 		ControlPlaneImage: controlPlaneImage,
+		CAConfigMap:       caConfigMap,
+		CAConfigMapKey:    caConfigMapKey,
 	}
 }
 
--- a/pkg/network/multus_admission_controller.go
+++ b/pkg/network/multus_admission_controller.go
@@ -81,17 +81,18 @@ func renderMultusAdmissonControllerConfig(manifestDir string, externalControlPla
 		data.Data["TokenMinterImage"] = os.Getenv("TOKEN_MINTER_IMAGE")
 		data.Data["TokenAudience"] = os.Getenv("TOKEN_AUDIENCE")
 		data.Data["RunAsUser"] = hsc.RunAsUser
+		data.Data["CAConfigMap"] = hsc.CAConfigMap
+		data.Data["CAConfigMapKey"] = hsc.CAConfigMapKey
 
-		// Get serving CA from the management cluster since the service resides there
 		serviceCA := &corev1.ConfigMap{}
 		err := client.ClientFor(names.ManagementClusterName).CRClient().Get(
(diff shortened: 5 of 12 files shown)


bindata/network/multus-admission-controller/001-service.yaml
```diff
@@ -13,7 +13,9 @@ metadata:
 {{- if .HyperShiftEnabled}}
     network.operator.openshift.io/cluster-name: {{.ManagementClusterName}}
 {{- end }}
+{{- if not .HyperShiftEnabled or eq .CAConfigMap ""}}
```
kyrtapz, line 16: This fails in CI:
```
I0312 20:47:44.272052       1 log.go:245] Failed to render: failed to render multus admission controller manifests: failed to render file bindata/network/multus-admission-controller/001-service.yaml: failed to render manifest bindata/network/multus-admission-controller/001-service.yaml: template: bindata/network/multus-admission-controller/001-service.yaml:16:7: executing "bindata/network/multus-admission-controller/001-service.yaml" at <not>: wrong number of args for not: want 1 got 5
```

Is it possible for `CAConfigMap` to be empty in HyperShift? it doesn't look like it from the code logic and there is no check in [multus-admission-controller/monitor.yaml](https://github.com/openshift/cluster-network-operator/pull/2309/files#diff-8c63f39010003c98df7342fc123f4fd07c270a9b0a7369f341064350207b1aaf)

Please add a render unit test somewhere in `pkg/network/multus_admission_controller_test.go`

pkg/network/multus_admission_controller.go
```diff
@@ -82,16 +82,31 @@ func renderMultusAdmissonControllerConfig(manifestDir string, externalControlPla
 		data.Data["TokenAudience"] = os.Getenv("TOKEN_AUDIENCE")
 		data.Data["RunAsUser"] = hsc.RunAsUser
 
-		// Get serving CA from the management cluster since the service resides there
+		CAConfigMap, ok := os.LookupEnv("CA_CONFIG_MAP")
```
kyrtapz, line 85: To avoid duplicating this logic it should be moved to `pkg/hypershift/hypershift.go` similar to other HyperShift env vars.

kyrtapz (commented):
> ... The PR also removes the service-sa annotation from services used by HyperShift since we will pass through a CA ConfigMap for the services we are using.

Do you mean `service.alpha.openshift.io/serving-cert-secret-name`? This PR doesn't remove that, but if it did it would break the affected deployments as they are still trying to mount the secret with the certificate created through that annotation, for example:
https://github.com/openshift/cluster-network-operator/blob/ab26470c566d83cb4f07d3464b935dd22d59ad4e/bindata/network/multus-admission-controller/admission-controller.yaml#L242

bindata/network/multus-admission-controller/001-service.yaml
bryan-cox: For older deployments, it's possible it would be empty. I will add a check in [multus-admission-controller/monitor.yaml](https://github.com/openshift/cluster-network-operator/pull/2309/files#diff-8c63f39010003c98df7342fc123f4fd07c270a9b0a7369f341064350207b1aaf).

pkg/network/multus_admission_controller.go
bryan-cox: Ok moved this to hypershift.go

bindata/network/multus-admission-controller/001-service.yaml
kyrtapz: `CAConfigMap` is set to the `CA_CONFIG_MAP` env if it exists or `openshift-service-ca.crt` otherwise. Are you saying that there is a scenario in which `CA_CONFIG_MAP` env is set to an empty value? If so please add a comment.

bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
```diff
@@ -23,8 +23,13 @@ spec:
     tlsConfig:
       ca:
         configMap:
+{{- if .HyperShiftEnabled}}
```
kyrtapz, line 26:
```suggestion
{{- if not .HyperShiftEnabled or eq .CAConfigMap ""}}
```

pkg/hypershift/hypershift.go
```diff
@@ -186,3 +186,19 @@ func SetHostedControlPlaneConditions(hcp *unstructured.Unstructured, operStatus
 	hcp.Object["status"].(map[string]interface{})["conditions"] = conditions
 	return conditions, nil
 }
+
+// GetCAConfigMapAndKey gets the CAConfigMap and key if set through an environment variable
+// otherwise, default to OpenShift service CA
+func GetCAConfigMapAndKey() (string, string) {
```
kyrtapz, line 192: NIT: There is no need to always read the env variable, we can just read it once and add new fields to `HyperShiftConfig` with the defaulting being done in `NewHyperShiftConfig`. See something similar here: 
https://github.com/openshift/cluster-network-operator/blob/23b758650fd6a91bbea398c3a3b429f8d7114b9e/pkg/hypershift/hypershift.go#L21-L2

pkg/network/node_identity.go
```diff
@@ -89,8 +89,24 @@ func renderNetworkNodeIdentity(conf *operv1.NetworkSpec, bootstrapResult *bootst
 	// HyperShift specific
 	if hcpCfg := hypershift.NewHyperShiftConfig(); hcpCfg.Enabled {
 		webhookCAClient = client.ClientFor(names.ManagementClusterName)
-		webhookCALookup = types.NamespacedName{Name: "openshift-service-ca.crt", Namespace: hcpCfg.Namespace}
-		caKey = "service-ca.crt"
+		CAConfigMap, ok := os.LookupEnv("CA_CONFIG_MAP")
```
kyrtapz, line 92: This should rely on the `hypershift.go` implementation instead of duplicating the logic.

bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml
bryan-cox: Thanks, I missed that. It's been added.

pkg/network/node_identity.go
bryan-cox: Updated

pkg/hypershift/hypershift.go
bryan-cox: Updated this.

bindata/network/multus-admission-controller/001-service.yaml
bryan-cox: Added comments.

bindata/network/multus-admission-controller/monitor.yaml
```diff
@@ -26,8 +26,13 @@ spec:
     tlsConfig:
       ca:
         configMap:
+{{- if and (.HyperShiftEnabled) (ne .CAConfigMap "")}}
```
kyrtapz, line 29: I thought the `CAConfigMap` check is not needed?

pkg/network/node_identity.go
```diff
@@ -89,8 +89,12 @@ func renderNetworkNodeIdentity(conf *operv1.NetworkSpec, bootstrapResult *bootst
 	// HyperShift specific
 	if hcpCfg := hypershift.NewHyperShiftConfig(); hcpCfg.Enabled {
 		webhookCAClient = client.ClientFor(names.ManagementClusterName)
-		webhookCALookup = types.NamespacedName{Name: "openshift-service-ca.crt", Namespace: hcpCfg.Namespace}
-		caKey = "service-ca.crt"
+
+		data.Data["CAConfigMap"] = hcpCfg.CAConfigMap
```
kyrtapz, line 93: Is `CAConfigMap` and `CAConfigMapKey` used anywhere in the managed network node identity yaml templates?

kyrtapz:
@bryan-cox this PR is blocked by https://github.com/openshift/hypershift/pull/3688, right?

bindata/network/multus-admission-controller/monitor.yaml
bryan-cox: Fixed

pkg/network/node_identity.go
bryan-cox: I don't see it in the yaml templates when I do a search on that directory for either `NetworkNodeIdentityCABundle` or `openshift-service-ca.crt`

bryan-cox:
> @bryan-cox this PR is blocked by [openshift/hypershift#3688](https://github.com/openshift/hypershift/pull/3688), right?

No it shouldn't be blocked by #3688

bryan-cox:
> > @bryan-cox this PR is blocked by [openshift/hypershift#3688](https://github.com/openshift/hypershift/pull/3688), right?
> 
> No it shouldn't be blocked by #3688

@kyrtapz it's actually the other way around. This PR is holding up https://github.com/openshift/hypershift/pull/3688

kyrtapz:
> @kyrtapz it's actually the other way around. This PR is holding up [openshift/hypershift#3688](https://github.com/openshift/hypershift/pull/3688)

This PR disables the service annotation on hypershift, what is going to create the secrets if not [this commit](https://github.com/openshift/hypershift/pull/3688/commits/e69b3b8754457b451f35845be533fa683bc8d2fa)? 

Additionally the PR currently fails on hypershift with:
```
I0326 13:34:23.317646       1 log.go:245] Failed to render: failed to render manifests: failed to render file bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml: failed to render manifest bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml: template: bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml:26:17: executing "bindata/network/ovn-kubernetes/managed/monitor-control-plane.yaml" at <.CAConfigMapKey>: map has no entry for key "CAConfigMapKey"
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2309/pull-ci-openshift-cluster-network-operator-master-e2e-hypershift-ovn/1772605886221520896/artifacts/e2e-hypershift-ovn/dump/artifacts/namespaces/clusters-9344cdbee5473b30752a/core/pods/logs/cluster-network-operator-7d45bf9478-lc42l-cluster-network-operator.log

bindata/network/multus-admission-controller/001-service.yaml
```diff
@@ -13,7 +13,10 @@ metadata:
 {{- if .HyperShiftEnabled}}
     network.operator.openshift.io/cluster-name: {{.ManagementClusterName}}
 {{- end }}
+# Any new HCP deployment will automatically reconcile this secret in the HCP without the service-ca operator
+{{- if not .HyperShiftEnabled}}
```
kyrtapz, line 17: To work in a hypershift env that doesn't reconcile the secrets you also need the additional CAConfigMap check here.

PR #2311, opened 2024-03-14
OCPBUGS-31669: [release-4.14] ensure local networking deployments within hypershift use the client side load balancer to be resilient to control plane node failures

relyt0925:
In environments with no cloud load balancer health checking and removing bad control plane endpoints: requests to the control plane from data plane components like kube-proxy and mutlus using the control plane dns name directly will see intermittent failures. In hypershift: a node local haproxy load balancer runs that health checks and removes failed control plane ingress points dynamically. When the components move to using this: control plane failures do not lead to further failures in the data plane starting new workload. fixes: https://issues.redhat.com/browse/OCPBUGS-30103, https://issues.redhat.com/browse/OCPBUGS-31514

### DIFF
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -3,6 +3,7 @@ package platform
 import (
 	"context"
 	"fmt"
+	"k8s.io/utils/pointer"
 	"os"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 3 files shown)


hasueki:
/retitle OCPBUGS-31669: [release-4.14] ensure local networking deployments within hypershift use the client side load balancer to be resilient to control plane node failures

pkg/platform/platform.go
```diff
@@ -115,6 +116,18 @@ func InfraStatus(client cnoclient.Client) (*bootstrap.InfraStatus, error) {
 			return nil, fmt.Errorf("failed to retrieve HostedControlPlane %s: %v", types.NamespacedName{Namespace: hc.Namespace, Name: hc.Name}, err)
 		}
 		res.HostedControlPlane = hcp
+		if res.HostedControlPlane.Spec.Networking.APIServer == nil {
+			res.HostedControlPlane.Spec.Networking.APIServer = &hyperv1.APIServerNetworking{
+				AdvertiseAddress: pointer.String("172.20.0.1"),
```
kyrtapz, line 121: Why does the backport ignore ipv6?

relyt0925: Backport is only for IBM Cloud where we don't have IPV6

relyt0925:
Just to note: this has been validated in IBM Cloud!

relyt0925:
@kyrtapz could I get your help getting this in the next level of approvals from a openshift prow perspective?

PR #2316, opened 2024-03-16
ensure local networking deployments within hypershift use the client side load balancer to be resilient to control plane node failures (all hypershift environments)

relyt0925:
In environments with no cloud load balancer health checking and removing bad control plane endpoints: requests to the control plane from data plane components like kube-proxy and mutlus using the control plane dns name directly will see intermittent failures. In hypershift: a node local haproxy load balancer runs that health checks and removes failed control plane ingress points dynamically. When the components move to using this: control plane failures do not lead to further failures in the data plane starting new workload.


relyt0925:
@kyrtapz can you confirm for ovn specifically: for hypershift the ultimate variable that takes precidence is the K8s_LOCAL_APISERVER variable? If so we are good but I noted we define that in a config map but in the deployment itself KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT is still defined.

kyrtapz:
@relyt0925 Thanks for following up on this and apologies for the delay in reviewing it.
Looking at the HyperShift implementation it seems to be always setting one backed:
```
backend remote_apiserver
  mode tcp
  log global
  option httpchk GET /version
  option log-health-checks
  default-server inter 10s fall 3 rise 3
  server controlplane {{ .ExternalAPIAddress }}:{{ .ExternalAPIPort }}
```
can you share a deployment example in which we gain something from the haproxy health checks?
The code looks good and the only concern I have is the additional dependency on the local HAProxy instance.

PR #2318, opened 2024-03-19
SDN-4519: Configure CNO as per new API changes for join and transit switch subnet

arghosh93:
This PR makes it possible for users to configure join and transit switch subnet as a day 1 operation keeping in mind of the new API changes for both subnets.

Change Log:

- validateOVNKubernetes() function now validates whether configurable subnets belong to same IP family as ClusterNetwork CIDR. This function also checks whether provided subnet have enough IPs to assign to all nodes schedulable as per ClusterNetwork CIDR and hostPrefix.

- V4InternalSubnet and V6InternalSubnet will be deprecated soon but until then values of those must be same as InternalJoinSubnet for IPv4 and IPv6 correspondingly. This PR adds a check to reassure that CIDR provided for join subnet in both fields are same.

- We used to check IP overlapping between configurable subnets at validateOVNKubernetes() function only. This PR moves all those IP overlapping checks to validateIPPools() function.

- This PR moves some unit test from TestValidateOVNKubernetes() to TestValidateIPPools(). TestValidateIPPools() consist of unit tests checking overlapping between configurable subnets.


arghosh93:
/hold
Need to test after conflicts in https://github.com/openshift/api/pull/1626 are resolved.

arghosh93:
I was able to create a cluster with unmerged API changes vendored into this CNO repository. For testing purpose I have pushed modified/deleted files changed after running 'go mod vendor' with a separate commit. I will remove this extra commit as soon as the API PR gets merged.

 I ran intended tests and received expected result. I would like to make this PR available for review. However I am not sure whether CI would pass or not. so, keeping the 'hold' label intact.
cc: @jcaamano

pkg/network/ovn_kubernetes.go
```diff
@@ -1761,9 +1778,8 @@ func setOVNObjectAnnotation(objs []*uns.Unstructured, key, value string) error {
 	return nil
 }
 
-func isV4InternalSubnetLargeEnough(conf *operv1.NetworkSpec) bool {
+func isV4SubnetLargeEnough(conf *operv1.NetworkSpec, subnet string) bool {
```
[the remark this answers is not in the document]
arghosh93: Ack

pkg/network/render.go
```diff
@@ -530,6 +530,80 @@ func validateIPPools(conf *operv1.NetworkSpec) []error {
 		errs = append(errs, errors.Errorf("spec.clusterNetwork and spec.serviceNetwork must either both be IPv4-only, both be IPv6-only, or both be dual-stack"))
 	}
 
+	oc := conf.DefaultNetwork.OVNKubernetesConfig
```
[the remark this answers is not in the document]
arghosh93: I have moved all configurable subnet checks to function validateOVNKubernetesSubnets(). Thanks for this nice suggestion.

pkg/network/ovn_kubernetes.go
```diff
@@ -1836,3 +1726,191 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+				}
+				_, ipv4JoinSwitchCidr, err := net.ParseCIDR(oc.IPv4.InternalJoinSubnet)
+				if err != nil {
+					out = append(out, errors.Errorf("v4InternalJoinSubnet is invalid: %s", err))
+				} else if !isV4NodeSubnetLargeEnough(conf.ClusterNetwork, oc.IPv4.InternalJoinSubnet) {
+					out = append(out, errors.Errorf("v4InternalJoinSubnet %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", oc.IPv4.InternalJoinSubnet))
+				}
+				if err := pool.Add(*ipv4JoinSwitchCidr); err != nil {
+					out = append(out, errors.Errorf("Whole or subset of v4InternalJoinSubnet CIDR %s is already in use: %s", oc.IPv4.InternalJoinSubnet, err))
+				}
```
jcaamano, line 1809: Maybe consider to have an adhoc aux function for this section of code that repeats 8 times, like
```
func validateOVNKubernetesSubnets(name, subnet string, otherSubnets *iputil.IPPool, cn []operv1.ClusterNetworkEntry)
```

pkg/network/ovn_kubernetes.go
```diff
@@ -1836,3 +1726,191 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+				out = append(out, errors.Errorf("v4InternalSubnet will be deprecated soon, until then it must be same as v4InternalJoinSubnet %s ", oc.IPv4.InternalJoinSubnet))
+			}
+		} else if oc.V4InternalSubnet != "" && oc.IPv4 == nil {
+			out = append(out, errors.Errorf("value of v4InternalSubnet needs to be same as v4InternalJoinSubnet"))
+		} else if oc.V4InternalSubnet == "" && oc.IPv4 != nil {
+			if oc.IPv4.InternalJoinSubnet != "" {
+				out = append(out, errors.Errorf("value of v4InternalSubnet needs to be same as v4InternalJoinSubnet"))
+			}
+		}
+
```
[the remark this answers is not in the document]
arghosh93: yes, makes sense. I hadn't considered upgrade scenario.

I was thinking too much about inconsistency. Anyway ovnKubernetesConfig.ipv4.internalJoinSubnet takes precedence over ovnKubernetesConfig.internalSubnet, so we should be good. Thanks for the suggestion!

pkg/network/ovn_kubernetes.go
arghosh93: Agreed, I have made changes and pushed. Thanks!

pkg/network/ovn_kubernetes.go
```diff
@@ -376,6 +406,21 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	err = setOVNObjectAnnotation(objs, names.TransitSwitchSubnetAnnotation, strings.Join(transitSwitchSubnet, ","))
+	if err != nil {
+		return nil, progressing, errors.Wrapf(err, "failed to set transit switch subnet %s annotation on ovnkube daemonset and deployment", strings.Join(transitSwitchSubnet, ","))
+	}
+
+	err = setOVNObjectAnnotation(objs, names.InternalMasqueradeSubnetAnnotation, strings.Join(internalMasqueradeSubnet, ","))
+	if err != nil {
+		return nil, progressing, errors.Wrapf(err, "failed to set masquerae subnet %s annotation on ovnkube daemonset and deployment", strings.Join(internalMasqueradeSubnet, ","))
+	}
```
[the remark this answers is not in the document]
arghosh93: Done, I have moved away from annotation based rollout. Now ovnkube deployment and daemonset should be rolled out based on change in hash value of 008-script-lib.yaml.

I have also added a separate commit for day 2 configuration change related implementation.
Thanks for all your help :)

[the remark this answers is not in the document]
arghosh93: Allowing day 2 configuration of masquerade subnet  was a mistake on my side. Sorry about it.

anuragthehatter:
@arghosh93 Is this PR ready for QE pre-merge test? With latest changes though I still see `100.88.0.0/16' still being used despite of forcing day0 to `10.99.0.0/16` say

pkg/network/ovn_kubernetes.go
```diff
@@ -1840,3 +1731,164 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+func validateOVNKubernetesSubnet(name, subnet string, otherSubnets *iputil.IPPool, cn []operv1.ClusterNetworkEntry) error {
+	out := []error{}
+	_, cidr, err := net.ParseCIDR(subnet)
+	if err != nil {
+		out = append(out, errors.Errorf("%s is invalid: %s", name, err))
+	} else if name != "v4InternalMasqueradeSubnet" && !utilnet.IsIPv6CIDRString(subnet) && !isV4NodeSubnetLargeEnough(cn, subnet) {
+		out = append(out, errors.Errorf("%s %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", name, subnet))
+	} else if name != "v6InternalMasqueradeSubnet" && utilnet.IsIPv6CIDRString(subnet) && !isV6NodeSubnetLargeEnough(cn, subnet) {
+		out = append(out, errors.Errorf("%s %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", name, subnet))
+	}
```
jcaamano, line 1890: I don't understand. You are not actually aggregating the errors, you just return the first error you encounter.

This is fine but no need for append or anything and you can follow a simple `if ... { return fmt.Errorf(...) }` strategy.

Also, for the subnets that not need a size check, you can pass a nil `cn` to signal that it is not needed. If properly documented, is probably more maintainable than having to check for specific names which are just passed through for error formatting

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -575,6 +575,15 @@ data:
         "
       fi
 
+      ovn_v4_join_subnet_opt=
+      if [[ "{{.V4JoinSubnet}}" != "" ]] ; then
```
jcaamano, line 579: nit: space in `]] ;` not used elsewhere so probably remove it for consistency. Many instances of this.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -603,5 +612,56 @@ data:
...
+{{- if eq .ProductFlavor "managed" }}
+        --k8s-token-file=/var/run/secrets/hosted_cluster/token \
+        --node-server-privkey=/etc/pki/tls/metrics-cert/tls.key \
+        --node-server-cert=/etc/pki/tls/metrics-cert/tls.crt \
+{{- end }}
+        ${ovn_v4_join_subnet_opt} \
+        ${ovn_v6_join_subnet_opt} \
+        ${ovn_v4_transit_switch_subnet_opt} \
+        ${ovn_v6_transit_switch_subnet_opt}
+    }
```
jcaamano, line 667: add empty line at the end

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -33,6 +33,7 @@ spec:
       annotations:
         hypershift.openshift.io/release-image: {{.ReleaseImage}}
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        network.operator.openshift.io/ovnkube-script-lib-hash: "{{.OVNKubeConfigHash}}"
```
jcaamano, line 36: So this is interesting because now control plane might rollout together with node for config changes that have nothing to do with it.

I can't think of an actual problem right now but this may indeed bring them about.

In general I think that the parameters should be set as env variables in the specific pod specs and that script lib should just use those env variables; instead of the using the template variables directly from the script lib. This is however out of scope and a job for a different PR.

Tagging @kyrtapz for awareness and to check if he has ideas on how this can be a problem.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -603,5 +612,56 @@ data:
...
+        --config-file=/run/ovnkube-config/ovnkube.conf \
+        --loglevel "${OVN_KUBE_LOG_LEVEL}" \
+        --metrics-bind-address "${metrics_bind_ip}:${metrics_port}" \
+        --metrics-enable-pprof \
+        --metrics-enable-config-duration \
+{{- if eq .ProductFlavor "managed" }}
+        --k8s-token-file=/var/run/secrets/hosted_cluster/token \
+        --node-server-privkey=/etc/pki/tls/metrics-cert/tls.key \
+        --node-server-cert=/etc/pki/tls/metrics-cert/tls.crt \
+{{- end }}
```
jcaamano, line 662: So I guess that another option for things like metrics config and cert config is to set them as variables before calling this function, and check in this function their values and act accordingly.

Not strong opinion but what I propose would be play better with a future option of using pod env variables and would be consistent for all parameters (instead of passing arguments for some, and checking template variables for others)

pkg/network/ovn_kubernetes.go
```diff
@@ -123,6 +123,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["RoutableMTU"] = nil
 	// v4 and v6 join subnet are used when the user wants to use the addresses that we reserve for the join subnet in ovn-k
 	// TODO: this field is being deprecated and will turn into c.GatewayConfig.IPv4/6.InternalJoinSubnet when we introduce the transit switch config into the api
+	data.Data["V4JoinSubnet"] = ""
+	data.Data["V6JoinSubnet"] = ""
+	data.Data["V4TransitSwitchSubnet"] = ""
+	data.Data["V6TransitSwitchSubnet"] = ""
+
```
jcaamano, line 130: curious, why do we need this initialization?

jcaamano (commented):
@arghosh93 can you rephrase the commit title `This commit contains a workaround` to somethign like `Avoid setting networkDiagnostics cluster config`

anuragthehatter:
> I forgot to remove old network.operator CRDs from manifests directory after running update-codegen.sh. I have removed old CRDs and force pushed, @anuragthehatter can you try again?

@arghosh93 Thanks, Sure

pkg/network/ovn_kubernetes.go
arghosh93: I thought adding all anomalies during the validation stage is a good idea and hence kept on appending errors instead of returning immediately upon encountering an error. Returning immediately also seems reasonable though.

I can start sending cn as nil for masquerade subnet.

If we dont pass name while formatting error then wont it be little bit confusing for customers? However I do agree that as per checks in place we are not allowing any overlaps with any configurable subnets. so, chances of any confusion is less.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: Ack

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: Ack

pkg/network/ovn_kubernetes.go
arghosh93: If we dont initialize and c.IPv4.InternalTransitSwitchSubnet is "" then V4TransitSwitchSubnet does not get set to any value. as a result, RenderTemplate() starts throughing variable undefined error as I am using V4TransitSwitchSubnet in 008.script-lib.yaml. 
Same will also happen for join subnet when we would deprecate c.V4InternalSubnet.

pkg/network/ovn_kubernetes.go
jcaamano: Yeah, continue passing the name to log it into the error, but do not use it to treat exceptions.

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
jcaamano: Although undesirable, I don't think it would be an actual problem. The same would happen when for example the ovn-k image changes during an upgrade.

kyrtapz: > So this is interesting because now control plane might rollout together with node for config changes that have nothing to do with it.
> I can't think of an actual problem right now but this may indeed bring them about.

So `OVNKubeConfigHash` contains the hash value of `ovnkube-script-lib` configmap that is only used in ovnkube-node. I think that introducing this here is confusing. 
Additionally, this would cause unnecessary rollouts of  ovnkube-control-plane every time a  ovnkube-node specific setting changes.

> In general I think that the parameters should be set as env variables in the specific pod specs and that script lib should just use those env variables; instead of the using template variables directly from the script lib. This is however out of scope and a job for a different PR.

Using env vars sounds reasonable but I would avoid introducing yet another way of rolling out changes just for the purpose of this PR.

jcaamano: > So `OVNKubeConfigHash` contains the hash value of `ovnkube-script-lib` configmap that is only used in ovnkube-node. I think that introducing this here is confusing. Additionally, this would cause unnecessary rollouts of ovnkube-control-plane every time a ovnkube-node specific setting changes.

I think using `ovnkube-script-lib` for ovn-control-plane also is reasonable. As I mention, the equivalent consideration would be when the ovn-k image changes on upgrade.

> Using env vars sounds reasonable but I would avoid introducing yet another way of rolling out changes just for the purpose of this PR.

Unsure what you mean with `another way`. I guess I am talking about **replacing** one way (using configuration template variables directly in script lib) with another (using env variables) so that we don't trigger rollouts of unnecessary components when certain configuration parameters are changed. Of course, it might be unreasonable to make the change in one go so we might need to assume doing this progressively.

pkg/network/ovn_kubernetes.go
```diff
@@ -1836,3 +1726,191 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+			out = append(out, errors.Wrapf(err, "could not parse spec.serviceNetwork %s", snet))
+			continue
+		}
+		if err := pool.Add(*cidr); err != nil {
+			out = append(out, errors.Errorf("Whole or subset of ServiceNetwork CIDR %s is already in use: %s", snet, err))
+		}
+	}
+
+	oc := conf.DefaultNetwork.OVNKubernetesConfig
+	if oc != nil {
```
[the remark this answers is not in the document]
arghosh93: Done

pkg/network/ovn_kubernetes.go
arghosh93: Done

pkg/network/ovn_kubernetes.go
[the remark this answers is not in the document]
arghosh93: Updated as per suggestions, Thanks!

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: This logic is no more part of the implementation. Thanks for the suggestion though :)

pkg/network/ovn_kubernetes.go
```diff
@@ -1840,3 +1731,196 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+	} else if oc.V4InternalSubnet == "" && oc.IPv4 != nil {
+		if oc.IPv4.InternalJoinSubnet != "" {
+			if !cnHasIPv4 {
+				out = append(out, errors.Errorf("JoinSubnet %s and ClusterNetwork must have matching IP families", oc.IPv4.InternalJoinSubnet))
+			}
+			if err := validateOVNKubernetesSubnet("v4InternalJoinSubnet", oc.IPv4.InternalJoinSubnet, &pool, conf.ClusterNetwork); err != nil {
+				out = append(out, err)
+			}
+		}
+	}
```
jcaamano, line 1819: We can go for a shorter approach if we use an aux variable to store the subnet
```
v4InternalSubnet := oc.V4InternalSubnet
if oc.IPv4 != nil && oc.IPv4.InternalJoinSubnet != "" {
    if v4InternalSubnet != "" && v4InternalSubnet != oc.IPv4.InternalJoinSubnet {
        out = append(out, errors.Errorf("v4InternalSubnet will be deprecated soon, until then it must be same as v4InternalJoinSubnet %s ", oc.IPv4.InternalJoinSubnet))
    }
    v4InternalSubnet = oc.IPv4.InternalJoinSubnet
}
if v4InternalSubnet != "" {
    if !cnHasIPv4 {
        out = append(out, errors.Errorf("JoinSubnet %s and ClusterNetwork must have matching IP families", oc.IPv4.InternalJoinSubnet))
    }
    if err := validateOVNKubernetesSubnet("v4InternalJoinSubnet", v4InternalSubnet, &pool, conf.ClusterNetwork); err != nil {
        out = append(out, err)
    }
}
```

And looking at the section of code above, it doesn't look like we are validating if a v4InternalSubnet is present when it should, as if we should do something else with the cnHasIPv4 check:
```
v4InternalSubnet := oc.V4InternalSubnet
if oc.IPv4 != nil && oc.IPv4.InternalJoinSubnet != "" {
    if v4InternalSubnet != "" && v4InternalSubnet != oc.IPv4.InternalJoinSubnet {
        out = append(out, errors.Errorf("v4InternalSubnet will be deprecated soon, until then it must be same as v4InternalJoinSubnet %s ", oc.IPv4.InternalJoinSubnet))
    }
    v4InternalSubnet = oc.IPv4.InternalJoinSubnet
}
if cnHasIPv4 && v4InternalSubnet == "" {
    out = append(out, errors.Errorf("JoinSubnet %s and ClusterNetwork must have matching IP families", oc.IPv4.InternalJoinSubnet))
}
if v4InternalSubnet != "" {
    if err := validateOVNKubernetesSubnet("v4InternalJoinSubnet", v4InternalSubnet, &pool, conf.ClusterNetwork); err != nil {
        out = append(out, err)
    }
}
```

Same approach on the other instances.

jcaamano: > Thank you so much @jcaamano for this really nice suggestion. first code block seems sufficient to me.
> 
> ```
> if cnHasIPv4 && v4InternalSubnet == "" {
>     out = append(out, errors.Errorf("JoinSubnet %s and ClusterNetwork must have matching IP families", oc.IPv4.InternalJoinSubnet))
> }
> ```
> 
> Neither oc.V4InternalSubnet nor oc.IPv4.InternalJoinSubnet is a mandatory field and can be empty. So above check seems not required to me. What do you think?

Okey. I got that wrong and I guess we are assuming defaults on ovn-k.

arghosh93: yes, right

pkg/network/ovn_kubernetes.go
```diff
@@ -1840,3 +1731,164 @@ func isIPSecEnabledInPod(pod v1.PodTemplateSpec, containerName string) bool {
...
+	} else if name != "v6InternalMasqueradeSubnet" && utilnet.IsIPv6CIDRString(subnet) && !isV6NodeSubnetLargeEnough(cn, subnet) {
+		out = append(out, errors.Errorf("%s %s is not large enough for the maximum number of nodes which can be supported by ClusterNetwork", name, subnet))
+	}
+	// If both InternalSubnet and InternalJoinSubnet has been added then CNO should not exhibit
+	// an overlap error. So, avoiding addition of InternalSubnet to otherSubnets IPPpool
+	if name != "v4InternalSubnet" && name != "v6InternalSubnet" {
+		if err := otherSubnets.Add(*cidr); err != nil {
+			out = append(out, errors.Errorf("Whole or subset of %s CIDR %s is already in use: %s", name, subnet, err))
+		}
+	}
```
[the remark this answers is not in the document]
arghosh93: Done, resolving this conversation.

anuragthehatter:
Looks good to me on IPv4, IPv4/IPv6 and IPv6 cluster as both Day0 and Day 2 config. Thanks

anuragthehatter:
QE also observe interruption in service traffic as a part of Day2 operation For e.g https://privatebin.corp.redhat.com/?80c94dc0ce425dcc#6mP9VJ5DyoHtHkgWXT7viWynvKEmQM7hCtSXfptzxmMg

Trying to understand what are acceptable limits in such cases?

anuragthehatter:
> QE also observe interruption in service traffic as a part of Day2 operation For e.g https://privatebin.corp.redhat.com/?80c94dc0ce425dcc#6mP9VJ5DyoHtHkgWXT7viWynvKEmQM7hCtSXfptzxmMg
> 
> Trying to understand what are acceptable limits in such cases?

@jcaamano @arghosh93 wdyt

jcaamano:
> > QE also observe interruption in service traffic as a part of Day2 operation For e.g https://privatebin.corp.redhat.com/?80c94dc0ce425dcc#6mP9VJ5DyoHtHkgWXT7viWynvKEmQM7hCtSXfptzxmMg
> > Trying to understand what are acceptable limits in such cases?
> 
> @jcaamano @arghosh93 wdyt

@anuragthehatter I don't think this is the place to discuss this. We decided to add day-2 support in this PR because the disruption was deemed as acceptable elsewhere, we were not involved, that did not happen in the context of this PR. Perhaps @kyrtapz can help here.

tssurya:
@arghosh93: awesome job!! :) thank you!

PR #2319, opened 2024-03-25
Add write permission for EgressQoS status

pperiyasamy:
Adds required rbac permissions to make https://github.com/ovn-org/ovn-kubernetes/pull/4158 to work in d/s.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -149,6 +149,7 @@ rules:
   resources:
   - adminpolicybasedexternalroutes/status
   - egressfirewalls/status
+  - egressqoses/status
   verbs:
   - patch
 {{- if .OVN_ADMIN_NETWORK_POLICY_ENABLE }}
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
```diff
@@ -69,6 +69,7 @@ rules:
   resources:
     - adminpolicybasedexternalroutes
     - egressfirewalls
+    - egressqoses
```
tssurya, line 72: ah this is because we now have CM watching eqos for the status. ack.

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
tssurya, line 1: what about microshift? i guess we don't enable the feature there?
and this rbac is common for sno, ocp and hcp I guess?.. so we are covered there..

tssurya (approved):
lgtm
I'll leave final approval to @kyrtapz 
this blocks the DS merge on ovnkube side

PR #2321, opened 2024-03-26
Remove managed cluster checking for live migration

pliurh:
In 4.16, we want to support SDN live migration for unmanaged clusters.

### DIFF
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -116,9 +116,6 @@ func ValidateClusterConfig(clusterConfig *configv1.Network, client cnoclient.Cli
 		if infraRes.HostedControlPlane != nil {
 			return errors.Errorf("network type live migration is not supported on HyperShift clusters")
 		}
-		if !infraRes.StandaloneManagedCluster {
-			return errors.Errorf("network type live migration is not supported on self managed clusters")
-		}
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/network/cluster_config.go
```diff
@@ -116,9 +116,6 @@ func ValidateClusterConfig(clusterConfig *configv1.Network, client cnoclient.Cli
 		if infraRes.HostedControlPlane != nil {
 			return errors.Errorf("network type live migration is not supported on HyperShift clusters")
 		}
-		if !infraRes.StandaloneManagedCluster {
```
kyrtapz, line 119: I think there is no reason to keep the `StandaloneManagedCluster` detection if we want to remove the check.

martinkennelly: +1

martinkennelly:
TODO: Following this, we should remove the "fake" managed cluster hacks in origin tests also.

pkg/network/cluster_config.go
pliurh: I'm just wondering if this information can be useful for other use-case in the future.

pliurh: Done

kyrtapz:
/lgtm


> TODO: Following this, we should remove the "fake" managed cluster hacks in origin tests also.
@martinkennelly is there a jira tracking this?

martinkennelly:
> /lgtm
> 
> > TODO: Following this, we should remove the "fake" managed cluster hacks in origin tests also.
> 
> @martinkennelly is there a jira tracking this?

yes, in np-915.

PR #2322, opened 2024-03-26
CONSOLE-3952: Deploy networking-console-plugin by CNO

orenc1:
Use the CNO operator config controller to deploy and reconcile the resources required to enable the new Networking console plugin:
* Deployment (hosts the plugin pages contents via an nginx web server)
* Service (exposes the webserver to the console)
* ConfigMap (for nginx config)
* ConsolePlugin (binds the console with the dynamic plugin)

Also, enables the console plugin by adding it to the cluster Console plugins list.

Jira ticket: https://issues.redhat.com/browse/CONSOLE-3952

### DIFF
--- a/bindata/networking-console-plugin/001-config-map.yaml
+++ b/bindata/networking-console-plugin/001-config-map.yaml
@@ -1,0 +1,39 @@
+apiVersion: v1
+data:
+  nginx.conf: |
+    error_log /dev/stdout info;
+    events {}
+    http {
+      include            /etc/nginx/mime.types;
+      default_type       application/octet-stream;
+      keepalive_timeout  65;
+      server {
+        listen              9443 ssl;
+        ssl_certificate     /var/cert/tls.crt;
+        ssl_certificate_key /var/cert/tls.key;
+        root                /opt/app-root/src;
+
+        # Prevent caching for plugin-manifest.json
+        location = /plugin-manifest.json {
+          add_header Cache-Control 'no-cache, no-store, must-revalidate, proxy-revalidate, max-age=0';
+          add_header Pragma 'no-cache';
+          add_header Expires '0';
+        }
+
+        # Prevent caching for plugin-entry.js
+        location = /plugin-entry.js {
+          add_header Cache-Control 'no-cache, no-store, must-revalidate, proxy-revalidate, max-age=0';
+          add_header Pragma 'no-cache';
+          add_header Expires '0';
+        }
+      }
+    }
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/component: networking-console-plugin
+    app.kubernetes.io/managed-by: cluster-network-operator
+    app.kubernetes.io/name: networking-console-plugin
+    app.kubernetes.io/part-of: cluster-network-operator
+  name: networking-console-plugin
+  namespace: openshift-network-console
--- a/bindata/networking-console-plugin/002-deployment.yaml
+++ b/bindata/networking-console-plugin/002-deployment.yaml
@@ -1,0 +1,116 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: networking-console-plugin
+  namespace: openshift-network-console
+  annotations:
+    kubernetes.io/description: |
+      This deployment deploys the Networking console plugin pod which serves
+      the contents of the Networking section in OpenShift Console
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+    networkoperator.openshift.io/non-critical: ""
+  labels:
+    app.kubernetes.io/component: networking-console-plugin
+    app.kubernetes.io/managed-by: cluster-network-operator
+    app.kubernetes.io/name: networking-console-plugin
+    app.kubernetes.io/part-of: cluster-network-operator
+spec:
+  replicas: {{.Replicas}}
+  selector:
+    matchLabels:
+      app.kubernetes.io/component: networking-console-plugin
+      app.kubernetes.io/managed-by: cluster-network-operator
+      app.kubernetes.io/name: networking-console-plugin
+      app.kubernetes.io/part-of: cluster-network-operator
+{{- if gt .Replicas 1 }}
+  strategy:
+    rollingUpdate:
+      maxUnavailable: 1
+    type: RollingUpdate
+{{- end }}
+  template:
+    metadata:
+      annotations:
+        target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+      labels:
+        app.kubernetes.io/component: networking-console-plugin
+        app.kubernetes.io/managed-by: cluster-network-operator
+        app.kubernetes.io/name: networking-console-plugin
+        app.kubernetes.io/part-of: cluster-network-operator
+    spec:
+{{- if gt .Replicas 1 }}
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app.kubernetes.io/component: networking-console-plugin
+                  app.kubernetes.io/managed-by: cluster-network-operator
+                  app.kubernetes.io/name: networking-console-plugin
+                  app.kubernetes.io/part-of: cluster-network-operator
+              namespaces:
+              - openshift-network-console
+              topologyKey: kubernetes.io/hostname
+            weight: 90
+{{- end }}
+      automountServiceAccountToken: false
+      containers:
+      - command:
+        - /bin/sh
+        - -c
+        - |
+          if echo "$POD_IP" | grep -qE '^([0-9]{1,3}\.){3}[0-9]{1,3}$'; then
+            LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="9443"
+          else
+            LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME="[::]:9443"
+          fi
+          sed "s/LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/$LISTEN_ADDRESS_PORT_REPLACED_AT_RUNTIME/g" /etc/nginx/nginx.conf > /tmp/nginx.conf
+          exec nginx -c /tmp/nginx.conf -g 'daemon off;'
+        env:
+        - name: POD_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        image: "{{.NetworkingConsolePluginImage}}"
+        imagePullPolicy: IfNotPresent
+        name: networking-console-plugin
+        ports:
+        - containerPort: 9443
+          name: https
+        resources:
+          requests:
+            cpu: 10m
+            memory: 50Mi
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+        terminationMessagePolicy: FallbackToLogsOnError
+        volumeMounts:
+        - mountPath: /var/cert
+          name: networking-console-plugin-cert
+          readOnly: true
+        - mountPath: /etc/nginx/nginx.conf
+          name: nginx-conf
+          readOnly: true
+          subPath: nginx.conf
+      dnsPolicy: ClusterFirst
+      nodeSelector:
+        kubernetes.io/os: linux
+      priorityClassName: system-cluster-critical
+      restartPolicy: Always
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      volumes:
+      - configMap:
+          defaultMode: 420
+          name: networking-console-plugin
+        name: nginx-conf
+      - name: networking-console-plugin-cert
+        secret:
+          defaultMode: 420
+          secretName: networking-console-plugin-cert
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"net"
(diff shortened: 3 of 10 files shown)


pkg/network/render.go
```diff
@@ -833,6 +856,110 @@ func renderNetworkPublic(manifestDir string) ([]*uns.Unstructured, error) {
...
 }
 
+// renderNetworkingConsolePlugin renders the common objects related to the networking console plugin resources
+func renderNetworkingConsolePlugin(manifestDir string, cl cnoclient.Client) ([]*uns.Unstructured, error) {
+	if !isConsolePluginCRDExist(cl) {
+		log.Printf("consoleplugins.console.openshift.io CRD does not exist yet")
+		return nil, nil
+	}
+
+	tmpFallbackConsoleImage := "quay.io/orenc/networking-console-plugin:4.16"
```
orenc1, line 866: looks like in an hypershift hosted cluster, the CNO deployment on the hosted control plane is not getting populated with the `NETWORKING_CONSOLE_PLUGIN_IMAGE` env var.
could you please advise how can we inject it in that case?

gouyang:
@orenc1 could you catch up on this? It's a test blocker.

pkg/network/render.go
```diff
@@ -886,6 +908,110 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
+		return errors.Wrapf(err, "Could not get consoles.operator.openshift.io resource")
+	}
+
+	if stringInSlice(consoleObj.Spec.Plugins, pluginName) {
+		return nil
+	}
+
+	var patches []cnoclient.JsonPatch
+
+	if consoleObj.Spec.Plugins == nil {
```
kyrtapz, line 967: Please use applyconfigurations instead of a raw JsonPath. You might even skip the `Get` to only do one API call.

pkg/network/render.go
```diff
@@ -886,6 +908,110 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
+	}
+
+	manifests, err := render.RenderDir(filepath.Join(manifestDir, "networking-console-plugin"), &data)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render networking-console-plugin manifests")
+	}
+	return manifests, nil
+}
+
+type Patch struct {
```
kyrtapz, line 936: This doesn't seem to be used anywhere

pkg/network/render.go
```diff
@@ -886,6 +908,110 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
+	Type types.PatchType
+	Data []byte
+}
+
+// registerNetworkingConsolePlugin enables console plugin for networking-console if not already enabled
+func registerNetworkingConsolePlugin(cl cnoclient.Client) error {
```
kyrtapz, line 942: What will happen when the console plugin is not deployed and we register the plugin?
A more general question: why doesn't the console just consume the plugins that were created?

pkg/network/render.go
```diff
@@ -886,6 +908,110 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
 }
 
+// renderNetworkingConsolePlugin renders the common objects related to the networking console plugin resources
+func renderNetworkingConsolePlugin(manifestDir string, cl cnoclient.Client) ([]*uns.Unstructured, error) {
+	if !isConsolePluginCRDExist(cl) {
+		log.Printf("consoleplugins.console.openshift.io CRD does not exist yet")
+		return nil, nil
+	}
+
+	tmpFallbackConsoleImage := "quay.io/orenc/networking-console-plugin:4.16"
```
kyrtapz, line 918: Please move that to a separate commit.
Note: https://github.com/openshift/cluster-network-operator/pull/2332#issuecomment-2250034263

kyrtapz (commented):
@orenc1 thanks for the PR!

Are there any E2E tests for these changes?

bindata/networking-console-plugin/001-config-map.yaml
```diff
@@ -0,0 +1,39 @@
...
+    }
+kind: ConfigMap
+metadata:
+  labels:
+    app.kubernetes.io/component: networking-console-plugin
+    app.kubernetes.io/managed-by: cluster-network-operator
+    app.kubernetes.io/name: networking-console-plugin
+    app.kubernetes.io/part-of: cluster-network-operator
+  name: networking-console-plugin
+  namespace: openshift-network-operator
```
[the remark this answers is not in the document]
orenc1: because it feels as the most suitable namespace to be at, avoiding opening a new namespace to host these resources.
as a reference, the monitoring-plugin resources are also deployed as the same namespace as the cluster-monitoring-operator.
do you have another suggestion?

pkg/network/render.go
```diff
@@ -886,6 +908,110 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
+		return errors.Wrap(err, fmt.Sprintf("registering console-plugin %q with console %q failed", pluginName, clusterConsole))
+	}
+	return nil
+}
+
 func isSupportedDualStackPlatform(platformType configv1.PlatformType) bool {
 	return dualStackPlatforms.Has(string(platformType))
 }
+
+func isConsolePluginCRDExist(cl cnoclient.Client) bool {
```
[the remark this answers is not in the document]
orenc1: It is created by the console-operator, but since both are cluster-operators, we want to make sure the CRD has been deployed beforehand. At cluster creation, both console-operator and CNO are running and creating resources.

pkg/network/render.go
orenc1: >What will happen when the console plugin is not deployed and we register the plugin?
nothing disastrous, the console will just drop some errors the the plugin could not be loaded in its logs and in the browser's console.

> A more general question: why doesn't the console just consume the plugins that were created?
because in the general case of a regular operator with a custom console plugin, the admin should have an option to disable it if s/he wants.
in our case, we're auto-registering it because it should be always enabled, as the networking section is a core section in the UI.

pkg/network/render.go
orenc1: of course.
that was just to be sure it's working in the CI.

bindata/networking-console-plugin/001-config-map.yaml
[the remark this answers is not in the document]
kyrtapz: This should have its own namespace, CNO deploys a lot of different components in the same manner. One example would be network-diagnostics. I suggest to move it to the `network-console-plugin` namespace.

pkg/network/render.go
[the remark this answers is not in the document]
kyrtapz: At which operator level in CVO is the console operator deployed?
Additionally to avoid unnecessary API calls please move the call to `consolePluginCRDExists` into the bootstrap part of CNO and introduce a variable that can be read in render, example:
https://github.com/openshift/cluster-network-operator/blob/6aa2f0fbf2b06ce52db1a2a039bbdbcaf588c409/pkg/platform/platform.go#L145-L149

pkg/network/render.go
```diff
@@ -139,6 +142,20 @@ func Render(operConf *operv1.NetworkSpec, clusterConf *configv1.NetworkSpec, man
 	}
 	objs = append(objs, o...)
 
+	// render networking console plugin
+	o, err = renderNetworkingConsolePlugin(manifestDir, bootstrapResult)
+	if err != nil {
+		return nil, progressing, err
+	}
+	if o != nil {
```
kyrtapz, line 150: NIT: this check is not necessary.

pkg/network/render.go
```diff
@@ -888,6 +905,49 @@ func renderIPTablesAlerter(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.
...
+	return manifests, nil
+}
+
+// registerNetworkingConsolePlugin enables console plugin for networking-console if not already enabled
+func registerNetworkingConsolePlugin(bootstrapResult *bootstrap.BootstrapResult, cl cnoclient.Client) error {
+	if !bootstrapResult.Infra.ConsolePluginCRDExists {
+		return nil
+	}
+	pluginName := "networking-consoleOperator-plugin"
+	consoleOperator := applyoperv1.Console(names.OPERATOR_CONFIG)
```
kyrtapz, line 941: `consoleOperator.Spec` is nil after this line, try this instead:
```suggestion
	consoleOperator := applyoperv1.Console(names.OPERATOR_CONFIG).WithSpec(applyoperv1.ConsoleSpec().WithPlugins(pluginName))
```

kyrtapz (changes requested):

pkg/platform/platform.go
```diff
@@ -252,3 +256,13 @@ func getMachineConfigPoolStatuses(ctx context.Context, client cnoclient.Client,
 	}
 	return mcpStatuses, nil
 }
+
+func consolePluginCRDExists(cl cnoclient.Client) bool {
+	consolePluginCrdKey := crclient.ObjectKey{Name: "consoleplugins.console.openshift.io"}
+	consolePluginCrdObj := &apiextensionsv1.CustomResourceDefinition{}
+	err := cl.Default().CRClient().Get(context.TODO(), consolePluginCrdKey, consolePluginCrdObj)
+	if err != nil {
```
kyrtapz, line 264: This error should be returned if it is not of type "IsNotFound"

kyrtapz:
/retitle CONSOLE-3952: Deploy networking-console-plugin by CNO

pkg/network/render.go
orenc1: I did it (with the addition of the required `FieldManager` in `ApplyOptions{}`, but i got a conflict error:
```
I0805 11:37:01.006859 1 log.go:245] Failed to render: registering networking-console-plugin "networking-console-plugin" with ConsoleOperator "cluster" has failed: Apply failed with 1 conflict: conflict with "operator" using operator.openshift.io/v1: .spec.plugins
```
could you please advise how can we avoid the conflict?

kyrtapz: You might need to set `ApplyOptions.Force` to true.

orenc1: ok, i've managed to get it working.
since `Plugins` is a list, the console object must be fetched before we attempt an apply operation, otherwise it will overwrite the list and remove the existing plugins.

kyrtapz: ugh :( The `Spec.Plugins` doesn't have a listType configured which means it is atomic. This doesn't make much sense in a scenario where multiple operators are supposed to edit it but that's the legacy API we need to handle.

To avoid racing and overriding others changes lets go back to a form of your initial approach. Something like this should do it:
```
func registerNetworkingConsolePlugin(bootstrapResult *bootstrap.BootstrapResult, cl cnoclient.Client) error {
	if !bootstrapResult.Infra.ConsolePluginCRDExists {
		return nil
	}

	console, err := cl.Default().OpenshiftOperatorClient().OperatorV1().Consoles().Get(context.TODO(), "cluster", metav1.GetOptions{})
	if err != nil {
		return errors.Wrap(err, "Failed to get Console Operator resource")
	}

	if sets.NewString(console.Spec.Plugins...).Has(pluginName) {
		return nil
	}
	console.Spec.Plugins = append(console.Spec.Plugins, pluginName)

	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
		_, err := cl.Default().OpenshiftOperatorClient().OperatorV1().Consoles().Update(context.TODO(), console, metav1.UpdateOptions{})
		return err
	})
}

```
P.S make pluginName a global const.

pkg/platform/platform.go
```diff
@@ -252,3 +256,17 @@ func getMachineConfigPoolStatuses(ctx context.Context, client cnoclient.Client,
...
+		if apierrors.IsNotFound(err) {
+			log.Printf("consoleplugins.console.openshift.io CRD was not found: %v", err)
+		} else {
+			log.Printf("ERROR: unable to get consoleplugins.console.openshift.io CRD: %v", err)
```
kyrtapz, line 268: The function should return this error, log it.
The error here should stop the bootstrap process.

orenc1: Done.
This function now returns an error in case the error is other than IsNotFound, and it is getting propagated into reconciliation error.

pkg/network/render.go
orenc1: changed the function back to use `Update()` instead of `Patch()` :+1:

bindata/networking-console-plugin/002-deployment.yaml
```diff
@@ -0,0 +1,110 @@
...
+      the contents of the Networking section in OpenShift Console
+    release.openshift.io/version: "{{.ReleaseVersion}}"
+    networkoperator.openshift.io/non-critical: ""
+  labels:
+    app.kubernetes.io/component: networking-console-plugin
+    app.kubernetes.io/managed-by: cluster-network-operator
+    app.kubernetes.io/name: networking-console-plugin
+    app.kubernetes.io/part-of: cluster-network-operator
+spec:
+  replicas: 2
```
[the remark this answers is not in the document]
orenc1: right, the second pod will remain Pending.
i changed the anti affinity rule to "preferred".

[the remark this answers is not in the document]
kyrtapz: Is there a point of deploying two replicas on the same node?
This is being flagged by CI: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2322/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-single-node/1825532768075911168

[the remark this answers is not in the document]
orenc1: looks like it worked now:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2322/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-single-node/1825818573545672704

item #55 here:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2322/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-single-node/1825818573545672704/artifacts/e2e-aws-ovn-single-node/gather-extra/artifacts/deployments.json

kyrtapz:
Have you verified that the PR works with hypershift?
Looking at the logs the plugin never gets enabled:
```
I0820 10:02:07.089080       1 log.go:245] consoleplugins.console.openshift.io CRD does not exist yet
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2322/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-hypershift-conformance/1825818566218223616/artifacts/e2e-aws-ovn-hypershift-conformance/dump/artifacts/namespaces/clusters-316a2c2da8b3b8900e71/core/pods/logs/cluster-network-operator-6ff4dc4cb7-phfrf-cluster-network-operator.log

upalatucci:
I've created an hyper shift cluster based on this pr and the console plugin is working properly. Can this depends on something else? Maybe we need to re-run the ci?

kyrtapz:
Most of the CI failures are caused by the known issue being fixed here https://github.com/openshift/origin/pull/29028.
/lgtm

orenc1:
https://github.com/openshift/origin/pull/29028 is merged and included in payload.
/retest-required

orenc1:
#2478 needs to be rebased.
/cherry-pick release-4.17

PR #2323, opened 2024-03-27
infraconfig: additional test cases for VIP mutation

mkowalski:
Contributes-to: OCPBUGS-31424


mkowalski:
/retest-required

This only adds unit tests so I don't care too much what and why failed

PR #2325, opened 2024-03-29
OCPBUGS-30948: Use Patch() instead of Update() for updating the status of network.operator

pliurh:
As we don't enable status subresource in the CRD
networks.operator.openshift.io. We used to use the Update() func to update the status of network.operator. It might mistakenly override some content in the spec. The func ApplyObject with Patch() underneath can avoid such a problem.


kyrtapz:
> As we don't enable status subresource in the CRD networks.operator.openshift.io. We used to use the Update() func to update the status of network.operator. It might mistakenly override some content in the spec. The func ApplyObject with Patch() underneath can avoid such a problem.

I don't see how it is possible for he Update function to change the spec here. I think that the resourceVersion wouldn't match and the request would get rejected. Do you have an example of how this happened in the wild?

pliurh:
> I don't see how it is possible for he Update function to change the spec here. I think that the resourceVersion wouldn't match and the request would get rejected. Do you have an example of how this happened in the wild?

The patch only contains the `status`. The resourceVersion won't be checked for the patch call.
The bug OCPBUGS-30948 is an example. The `egressIP` field will be mistakenly overwritten when set to `false`.  It may be related to  https://github.com/openshift/api/blob/abd990ce290b7e9481e33a1a71e045b39cc77ca1/operator/v1/types_network.go#L175.

kyrtapz:
> The patch only contains the `status`. The resourceVersion won't be checked for the patch call. The bug OCPBUGS-30948 is an example. The `egressIP` field will be mistakenly overwritten when set to `false`. It may be related to https://github.com/openshift/api/blob/abd990ce290b7e9481e33a1a71e045b39cc77ca1/operator/v1/types_network.go#L175.

I do not see how the current approach (with `Update`) could cause any overwriting with the resourceVersion check, I don't think it is possible. Could you share a an order of events/reproducer steps that could help me understand how this happened?

pliurh:
> I do not see how the current approach (with Update) could cause any overwriting with the resourceVersion check, I don't think it is possible. Could you share a an order of events/reproducer steps that could help me understand how this happened?

The reproduction steps are described at https://issues.redhat.com/browse/OCPBUGS-30948. After executing the steps, You will find the `spec.migration.feature.egressIP` field is switched to `true` by CNO, specifically by the `set` function I modified. I suspect it's a bug in the `Update` function when a boolean json field has `true` as the default value with `omitempty`.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -387,7 +388,18 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
...
+		}
+		// make sure we only update the status field
+		patchData, err := json.Marshal(&st{
+			Status: oc.Status,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to create patch: %v", err)
+		}
+
+		if err := status.client.ClientFor("").CRClient().Patch(context.TODO(), oc, crclient.RawPatch(types.MergePatchType, patchData)); err != nil {
```
kyrtapz, line 402: I don't think that `retry.RetryOnConflict` makes sense with this change?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -316,16 +317,21 @@ func (status *StatusManager) writeHypershiftStatus(operStatus *operv1.NetworkSta
 // Set updates the operator and clusteroperator statuses with the provided conditions.
 func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv1.OperatorCondition) {
 	var operStatus *operv1.NetworkStatus
+	oc := &operv1.Network{ObjectMeta: metav1.ObjectMeta{Name: names.OPERATOR_CONFIG}}
 
 	// Set status on the network.operator object
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
kyrtapz, line 323: Is there a point of running the `RetryOnConflict` on a Get?

PR #2326, opened 2024-04-02
SDN-4725:Bump openshift/api

bpickard22:
Brings in new transitSwitch and updated joinSwitch fields


jcaamano:
@bpickard22 we will need a new way to vendor the manifests. @arghosh93 has an attempt [here](https://github.com/openshift/cluster-network-operator/pull/2318/files#diff-ce07b69075d5f4237197bb90b721f2b357c8567a1a8945068239b32e57d85552) but he is missing the 
```
//go:build tools
// +build tools
```
tags.

Which would then enable you to execute `hack/update-codegen.sh` which we need I think.

PR #2327, opened 2024-04-03
OCPBUGS-31528: infraconfig: handle vSphere UPI by not updating any fields

mkowalski:
vSphere UPI is a special platform which uses `platform: vsphere` but its PlatformStatus is `nil`. Because of this, whenever we need to differentiate the migration scenario for vSphere IPI and UPI

* IPI starts upgrade with `nil` PlatformSpec and ends with values copied from PlatformStatus
* UPI starts upgrade with `nil` PlatformSpec and ends with `nil`

Important to note that serialization of `nil` is different than serialization of an empty struct.

Important to note that having Platform defined and empty/nil PlatformStatus is a good enough indicator that we use UPI.

Fixes: OCPBUGS-31528


zhaozhanqi:
pre-merge tested this PR with UPI vsphere and upgrade pass

PR #2328, opened 2024-04-03
SDN-4596: Update (B)ANP CRDs; prepare for GA

tssurya:
This PR updates the ANP/BANP CRDs to reflect the latest changes, i.e egress support and this reflects the release v0.1.3 upstream: https://github.com/kubernetes-sigs/network-policy-api/releases/tag/v0.1.3 (had to change bundle version manually since release is not cut yet but this is just aesthetics) + adds support for `isCIDR()` in the networks peer which is OCP carry only HACK because it hits k/k only in 1.31 release but we have support for this in OCP thanks to https://github.com/openshift/kubernetes/pull/1828

The manual changes are a second commit on top with OCPHACKs title, we should hopefully sync with upstream once we cut a release + we become 1.31 so that network-policy-api repo can also leverage isCIDR()

source-code: https://github.com/tssurya/network-policy-api/tree/final-ocp-crds-4.16-ga

### DIFF
--- a/bindata/network/ovn-kubernetes/common/001-crd.yaml
+++ b/bindata/network/ovn-kubernetes/common/001-crd.yaml
@@ -862,8 +862,9 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/network-policy-api/pull/106
-    policy.networking.k8s.io/bundle-version: v0.1.0
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/network-policy-api/pull/30
+    policy.networking.k8s.io/bundle-version: v0.1.1
+    policy.networking.k8s.io/channel: experimental
   creationTimestamp: null
   name: adminnetworkpolicies.policy.networking.k8s.io
 spec:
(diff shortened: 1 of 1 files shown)


kyrtapz:
Would it be possible to automate pulling in the CRDs from https://github.com/kubernetes-sigs/network-policy-api/tree/main/config/crd/standard?
`hack/update-codegen.sh` is the right place for that.

tssurya:
> Would it be possible to automate pulling in the CRDs from https://github.com/kubernetes-sigs/network-policy-api/tree/main/config/crd/standard? `hack/update-codegen.sh` is the right place for that.

yes but that is a larger change I think? meaning we have the same problem for syncing crds from ovnk upstream like efw, eip, apbr etc in cno which are done manually. We can have a tech debt cleanup?

kyrtapz:
> > Would it be possible to automate pulling in the CRDs from https://github.com/kubernetes-sigs/network-policy-api/tree/main/config/crd/standard? `hack/update-codegen.sh` is the right place for that.
> 
> yes but that is a larger change I think? meaning we have the same problem for syncing crds from ovnk upstream like efw, eip, apbr etc in cno which are done manually. We can have a tech debt cleanup?

ovn-k upstream CRDs are not versioned so that would be tricky to achieve, it should be doable here if we use a proper version from the upstream repository.
I am ok with doing the automation as a followup considering the CRD is already in CNO anyway. 

/lgtm

PR #2329, opened 2024-04-03
SDN-4114: Add iptables-alerter daemonset

danwinship:
This adds a daemonset that runs a script that periodically:
- lists all local non-`hostNetwork` pods
- checks if iptables rules have been created in that pod's network namespace (other than the rules created by ovn-kubernetes/openshift-sdn)
- logs an Event if so

There was an enhancement for this which never went anywhere, but you can see some discussion of ideas in https://github.com/openshift/enhancements/pull/1509/files.

I actually originally implemented this in MCO via a systemd timer that ran a locally-installed script. But this ran into problems because it had to use kubelet's serviceAccount, and kubelet doesn't have permission to list/create events in all namespaces. So I would have had to create a separate service account and get the credentials installed onto the node but I don't think we have any simple way to do that. So I went with the privileged-DaemonSet-while-loop approach instead.

The script is currently sort of a mess because I wasn't sure if there's a pre-built image that has the binaries I want (particularly `crictl`)...

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
@@ -94,6 +94,8 @@ spec:
           value: quay.io/openshift/origin-cluster-network-operator:latest
         - name: CLOUD_NETWORK_CONFIG_CONTROLLER_IMAGE
           value: quay.io/openshift/origin-cloud-network-config-controller:latest
+        - name: CLI_IMAGE
+          value: quay.io/openshift/origin-cli:latest
         - name: POD_NAME
           valueFrom:
             fieldRef:
--- a/bindata/network/iptables-alerter/001-rbac.yaml
+++ b/bindata/network/iptables-alerter/001-rbac.yaml
@@ -1,0 +1,32 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: openshift-iptables-alerter
+rules:
+- apiGroups: ["", "events.k8s.io"]
+  resources:
+  - events
+  verbs:
+  - get
+  - list
+  - create
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: iptables-alerter
+  namespace: openshift-network-operator
+---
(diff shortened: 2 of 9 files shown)


danwinship:
/hold
this is no longer WIP but it is still blocked on https://github.com/openshift/ovn-kubernetes/pull/1946 and https://github.com/openshift/sdn/pull/581

qiowang721:
Hi @danwinship , as per our discussion, currently the script can not detect if i use the iptables-legacy binaries, added a comment about it here for later planning, thanks.

manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
```diff
@@ -94,6 +94,8 @@ spec:
           value: quay.io/openshift/origin-cluster-network-operator:latest
         - name: CLOUD_NETWORK_CONFIG_CONTROLLER_IMAGE
           value: quay.io/openshift/origin-cloud-network-config-controller:latest
+        - name: CLI_IMAGE
```
kyrtapz, line 97: HyperShift configures the CLI image here: https://github.com/openshift/hypershift/blob/f085b411118247a77d6cb5be15b5db4d23a15ad3/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L534

I think that it should be the same one?

kyrtapz (commented):
Looks good to me overall.

manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
danwinship: yes... uh... wait, several other controllers are already using `os.Getenv("CLI_IMAGE")` without us having defined it in CNO's environment previously... how does that work outside of hypershift?

tssurya:
sdn and ovnk PRs have merged
@danwinship we should be good here to remove hold?

kyrtapz:
> sdn and ovnk PRs have merged @danwinship we should be good here to remove hold?

Lets see if it picks anything in CI!
/test e2e-aws-ovn
/test e2e-aws-sdn-upgrade 
/test e2e-hypershift-ovn

PR #2332, opened 2024-04-09
CONSOLE-3952: Add networking-console-plugin image to release payload

orenc1:
This is required so hypershift's control-plane-operator will be able to deploy CNO with this image, extracted from the release payload.

This PR is a prerequisite for #2322

### DIFF
--- a/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
+++ b/manifests/0000_70_cluster-network-operator_03_deployment-ibm-cloud-managed.yaml
@@ -96,6 +96,8 @@ spec:
           value: quay.io/openshift/origin-cli:latest
         - name: FRR_K8S_IMAGE
           value: quay.io/openshift/origin-metallb-frr:latest
+        - name: NETWORKING_CONSOLE_PLUGIN_IMAGE
+          value: quay.io/openshift/origin-networking-console-plugin:latest
         - name: POD_NAME
           valueFrom:
             fieldRef:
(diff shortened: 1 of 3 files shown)


kyrtapz:
> This is required so hypershift's control-plane-operator will be able to deploy CNO with this image, extracted from the release payload.

This change is not really HyperShift specific? We need the image in the release payload for standalone openshift as well.

You are going to need to include the new image ENV for CNO deployment in HyperShift which is done from HyperShift here:
https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go

orenc1:
> > This is required so hypershift's control-plane-operator will be able to deploy CNO with this image, extracted from the release payload.
> 
> This change is not really HyperShift specific? We need the image in the release payload for standalone openshift as well.
> 
> You are going to need to include the new image ENV for CNO deployment in HyperShift which is done from HyperShift here: https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go

Yes, but in order to do so, we'll need to add that image to the OCP payload first, right?

kyrtapz:
> Yes, but in order to do so, we'll need to add that image to the OCP payload first, right?

Correct, I was clarifying that this change is not HyperShift specific.

/lgtm

kyrtapz:
/retitle CONSOLE-3952: Add networking-console-plugin image to release payload

kyrtapz:
@orenc1 `ci/prow/okd-scos-images`  is not reporting any status and the PR seems stuck.

orenc1:
> @orenc1 `ci/prow/okd-scos-images` is not reporting any status and the PR seems stuck.

ok, i'm rebasing the PR.

kyrtapz:
/hold
lets figure out how are we going to handle okd and if we are going to be able to merge all the other pieces in time for feature freeze

orenc1:
looks like `okd-scos-images` is broken.
/test okd-scos-images

PR #2335, opened 2024-04-12
OCPBUGS-31878: Propogate hypershift control plane priority class override to multus and preserve container resource requests

lucaspalm:
**What this PR does / why we need it**:
The multus-admission-controller component currently does not honor the hypershift custom priority class overrides specified on the hostedcluster.  The custom priority class override feature was first introduced to all other hypershift hosted control plane components in https://github.com/openshift/hypershift/pull/2661.  Multus was left out at the time due to it not being directly managed by the CPO.  These changes extend the hypershift custom priority class override feature to the multus-admission-controller to ensure that this component honors the custom class if it is set.  These changes also extend the hypershift control plane component resource preservation support to the multus container as well, where the expected behavior is that any modifications to the container's resource requests are not overwritten on subsequent reconciliations.

**Which issue(s) this PR fixes**:
https://issues.redhat.com/browse/OCPBUGS-31878

### DIFF
--- a/pkg/hypershift/hypershift.go
+++ b/pkg/hypershift/hypershift.go
@@ -58,6 +58,7 @@ type HostedControlPlane struct {
 	NodeSelector                 map[string]string
 	AdvertiseAddress             string
 	AdvertisePort                int
+	PriorityClass                string
 }
 
 // AvailabilityPolicy specifies a high level availability policy for components.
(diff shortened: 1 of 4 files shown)


riendeau:
> tide Pending — Not mergeable. Needs lgtm label.

I _think_ all required tests are now passing (`e2e-hypershift-ovn` [doesn't really exist](https://github.com/openshift/cluster-network-operator/pull/2335#issuecomment-2122950766) any more?) , and we just need an `lgtm`?

riendeau (changes requested):
Please hold for now, pending investigation of default value specification in the template.

riendeau (approved):

riendeau:
I've gone through and now understand the usage of the `default` func provided by sprig here.

/remove-hold

rtheis (approved):
/lgtm

rtheis:
@riendeau please fix the Jira problems.

pkg/hypershift/hypershift.go
```diff
@@ -141,6 +142,11 @@ func ParseHostedControlPlane(hcp *unstructured.Unstructured) (*HostedControlPlan
 		return nil, fmt.Errorf("failed to extract controllerAvailabilityPolicy: %v", err)
 	}
 
+	controlPlanePriorityClassAnnotation, _, err := unstructured.NestedString(hcp.UnstructuredContent(), "metadata", "annotations", "hypershift.openshift.io/control-plane-priority-class")
```
enxebre, line 145: nit: it'd be nice to have a comment linking the counterpart source code. So when someone with less context reads the context can understand where this is coming from.

PR #2341, opened 2024-04-17
OCPBUGS-32347: Improve IPsec MachineConfig readiness check

pperiyasamy:
Currently CNO checks only machineConfigStatus.Configuration.Source for IPsec Machine Config are rolled out completely on cluster nodes, but from CI test with 24 nodes, it's found that source is populated even before UpdatedMachineCount reaches to MachineCount. So CNO proceeds with rendering IPsec DaemonSet which makes its pods into crashlooping on those node because ipsec service are not running yet. Hence this commit also checks MachineCount with UpdatedMachineCount to ensure IPsec machine config is available on all nodes.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -1461,22 +1461,23 @@ func isIPsecMachineConfigActive(infra bootstrap.InfraStatus) bool {
 		// One of the IPsec MachineConfig is not created yet, so return false.
 		return false
 	}
-	ipSecPluginOnMasterNodes := hasSourceInMachineConfigStatus(infra.MasterMCPStatus, infra.MasterIPsecMachineConfigs)
-	ipSecPluginOnWorkerNodes := hasSourceInMachineConfigStatus(infra.WorkerMCPStatus, infra.WorkerIPsecMachineConfigs)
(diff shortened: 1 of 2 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1472,7 +1472,8 @@ func hasSourceInMachineConfigStatus(machineConfigStatus mcfgv1.MachineConfigPool
 		sourceNames.Insert(machineConfig.Name)
 	}
 	for _, source := range machineConfigStatus.Configuration.Source {
-		if sourceNames.Has(source.Name) {
+		if sourceNames.Has(source.Name) &&
+			machineConfigStatus.MachineCount == machineConfigStatus.UpdatedMachineCount {
```
jcaamano, line 1476: I think this check should be done in `isIPsecMachineConfigActive`

Also, reading this code now it is confusing this method returns true if just one of the passed machineConfigs is present, vs all of them.

pperiyasamy: yes @jcaamano , changed method name with a comment to reflect what it's intended to achieve, hope it's ok now.

jcaamano: I meant to say move the check to the other method, not change this method's name.

pperiyasamy: ok, that makes sense. done.

PR #2343, opened 2024-04-18
[DNM] Art+patch

tssurya:
We need good developer docs in CNO so that other teams can contribute properly :/

Helping out with https://github.com/openshift/cluster-network-operator/pull/2277 ignore this PR


tssurya:
@kyrtapz : PTAL, I think you mentioned some concerns over this PR right? cc @ashwindasr

kyrtapz:
> @kyrtapz : PTAL, I think you mentioned some concerns over this PR right? cc @ashwindasr

Thanks for the ping. So I have two concerns:
 - Some of the changed files are vendored in from the API, why can't we add the annotation directly in the API repo? @ashwindasr 
 - We need to update the `hack/update-codegen.sh` script so it doesn't overwrite the changes in this pr

PR #2344, opened 2024-04-19
NP-974: Validation for live migration from OpenShiftSDN to OVNKubernetes 

kyrtapz:
Adds the following validations for OpenShiftSDN to OVNKubernetes live migration:
 - Do not allow migrating from OpenShiftSDN in multitenant mode.
 - Check whether any of the `ClusterNetwork` and `ServiceNetwork` CIDRs conflict with the internal OVNKubernetes subnets: `v[4|6]InternalSubnet`, `v[4|6]InternalTransitSwithSubnet` and `v[4|6]InternalMasquerateSubnet`.
 - Do not allow migrating from OpenShiftSDN if there are any pods
   with pod.network.openshift.io/assign-macvlan annotation present in the cluster.

### DIFF
--- a/pkg/network/cluster_config.go
+++ b/pkg/network/cluster_config.go
@@ -1,16 +1,22 @@
 package network
 
 import (
+	"context"
+	"fmt"
 	"net"
 	"strings"
 
 	configv1 "github.com/openshift/api/config/v1"
+	v1 "github.com/openshift/api/network/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/cluster-network-operator/pkg/platform"
 	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	utilnet "k8s.io/utils/net"
@@ -28,10 +34,10 @@ func ValidateClusterConfig(clusterConfig *configv1.Network, client cnoclient.Cli
 	if err != nil {
 		return err
 	}
-	return validateClusterConfig(clusterConfig, infraRes)
+	return validateClusterConfig(clusterConfig, infraRes, client)
 }
 
-func validateClusterConfig(clusterConfig *configv1.Network, infraRes *bootstrap.InfraStatus) error {
+func validateClusterConfig(clusterConfig *configv1.Network, infraRes *bootstrap.InfraStatus, client cnoclient.Client) error {
 
 	// Check all networks for overlaps
 	pool := iputil.IPPool{}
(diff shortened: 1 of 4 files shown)


pkg/network/cluster_config.go
```diff
@@ -117,9 +123,128 @@ func validateClusterConfig(clusterConfig *configv1.Network, infraRes *bootstrap.
...
+			return err
+		}
+
+		// Detect pods with pod.network.openshift.io/assign-macvlan (Used by egress router)
+		macVlanPods, err := cnoclient.ListAllPodsWithAnnotationKey(context.TODO(), client, v1.AssignMacvlanAnnotation)
+		if err != nil {
+			return errors.Wrapf(err, "error listing macvlan pods")
+		}
+		if len(macVlanPods) != 0 {
+			return fmt.Errorf("network type live migration is not supported for pods with %q annotation", v1.AssignMacvlanAnnotation)
```
pliurh, line 247: Shall we just tell the user to remove the egress router configuration before starting the live migration?

pkg/network/cluster_config.go
```diff
@@ -117,9 +123,128 @@ func validateClusterConfig(clusterConfig *configv1.Network, infraRes *bootstrap.
...
+		})
+	}
+	for _, svcNet := range clusterCofnig.Spec.ServiceNetwork {
+		subnets = append(subnets, subnet{
+			name: "serviceNetwork",
+			cidr: svcNet,
+		})
+	}
+
+	v4InternalSubnet, v6InternalSubnet := GetInternalSubnets(operConfig.Spec.DefaultNetwork.OVNKubernetesConfig)
```
pliurh, line 154: Do we need to check the IPv6 subnet? We don't support IPv6 in live migration.

pkg/network/cluster_config.go
```diff
@@ -117,9 +123,128 @@ func validateClusterConfig(clusterConfig *configv1.Network, infraRes *bootstrap.
...
+	subnets = append(subnets, subnet{
+		name: "v4InternalMasqueradeSubnet",
+		cidr: v4InternalMasqueradeSubnet,
+	})
+	subnets = append(subnets, subnet{
+		name: "v6InternalMasqueradeSubnet",
+		cidr: v6InternalMasqueradeSubnet,
+	})
+
+	for i, subnetA := range subnets {
```
pliurh, line 184: shall we put the ovn-k internal subnets and the cluster subnets in separate lists and compare them with each other?

pkg/network/ovn_kubernetes.go
```diff
@@ -65,8 +65,14 @@ const OVN_NODE_IDENTITY_CERT_DURATION = "24h"
 const OVN_EGRESSIP_HEALTHCHECK_PORT = "9107"
 
 const (
-	OVSFlowsConfigMapName   = "ovs-flows-config"
-	OVSFlowsConfigNamespace = names.APPLIED_NAMESPACE
+	OVSFlowsConfigMapName        = "ovs-flows-config"
+	OVSFlowsConfigNamespace      = names.APPLIED_NAMESPACE
```
pliurh, line 69: Are they used in this patch?

pkg/network/cluster_config.go
kyrtapz: I think it doesn't hurt to have if here, there is a chance that a creative user sets the IPv6 CIDRs to overlapping values before kicking of the migration right?

pkg/network/cluster_config.go
kyrtapz: I think we can keep it as is, it would detect overlaps between ovn-k networks if they were specified

pkg/network/ovn_kubernetes.go
kyrtapz: I haven't changed anything here, this is only space alignment

pkg/network/cluster_config.go
kyrtapz: Does this work for you?
`network type live migration is not supported for pods with %q annotation. Please remove all egress router pods`

pkg/network/cluster_config.go
pliurh: We don't support SDN to dual-stack OVN migration. We may reject it here.

kyrtapz: The user can change the cluster to dual-stack after migrating from SDN, there is no way to make the cluster dual-stack during the migration.
I don't think that there is any harm in having the ipv6 checks in place here although I agree that this is an unlikely scenario.

pkg/network/cluster_config.go
kyrtapz: Done.

kyrtapz:
Fixed unit tests.
@pliurh ptal

PR #2345, opened 2024-04-19
OCPBUGS-27925: Tighten conditions on shouldUpdateOVNKonInterConnectZoneModeChange

jluhrsen:
shouldUpdateOVNKonInterConnectZoneModeChange should only hold off the update of ovnkube components if an upgrade to IC is ongoing. In all other cases it should be return true, true, true.


jluhrsen:
closing this because we really do want to see [this](https://github.com/openshift/cluster-network-operator/pull/2207) land instead.

PR #2349, opened 2024-04-22
OCPBUGS-32525: Check every MachineConfigPool for IPsec plugin existence

pperiyasamy:
While testing IPsec upgrade on 24 node cluster in CI, it's noticed there are more than two (master, worker) machine config pools present. For example infra and workload pools were present which belong worker role. Hence this PR queries all machine config pools based on master and worker roles and check ipsec plugin existence on every pool.

### DIFF
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -13,6 +13,7 @@ import (
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	types "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


pkg/platform/platform.go
```diff
@@ -26,6 +26,11 @@ var cloudProviderConfig = types.NamespacedName{
 	Name:      "kube-cloud-config",
 }
 
+var (
+	MasterRoleMachineConfigLabel = "machineconfiguration.openshift.io/role=master"
+	WorkerRoleMachineConfigLabel = "machineconfiguration.openshift.io/role=worker"
+)
+
```
jcaamano, line 33: nit: make this const and unexport

pkg/platform/platform.go
```diff
@@ -234,3 +235,21 @@ func isMachineConfigClusterOperatorReady(client cnoclient.Client) (bool, error)
...
+func GetMachineConfigPoolStatuses(ctx context.Context, client cnoclient.Client, mcLabel string) ([]mcfgv1.MachineConfigPoolStatus, error) {
+	mcpList := &mcfgv1.MachineConfigPoolList{}
+	if err := client.Default().CRClient().List(ctx, mcpList); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+	}
+	var mcpStatuses []mcfgv1.MachineConfigPoolStatus
+	for _, mcp := range mcpList.Items {
+		for lKey, lValue := range mcp.Spec.MachineConfigSelector.MatchLabels {
```
jcaamano, line 248: What if `MatchExpressions` is used instead of `MatchLabels`?
```
  machineConfigSelector:
    matchExpressions:
      - {key: machineconfiguration.openshift.io/role, operator: In, values: [worker,infra]}
```
from  https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md

I guess we need to use some kind of kubernetes library for label matching here.

pkg/platform/platform.go
pperiyasamy: ok, done.

PR #2353, opened 2024-04-23
infraconfig: use client-side apply

mkowalski:
Infrastructure CR is an object managed by at least 2 actors,
cluster-bootstrap and network-operator. The former initializes the
object with values provided at install-time, the latter manages
the object according to the user's requests.

Because of the type used in the API definition of PlatformStatus (`set`
and not `atomic`), it is not trivial for the controller to perform
server-side apply that modifies values owned by cluster-bootstrap.

In order to overcome the issue, we are moving to client-side apply and
we are adding additional error handling (`RetryOnConflict`) to be able
to fully control all the fields we may need to update.


mkowalski:
/hold
https://github.com/openshift/cluster-network-operator/pull/2355 will be more elegant as it uses server-side apply

mkowalski:
/close

Fixing via https://github.com/openshift/api/pull/1873

PR #2355, opened 2024-04-24
OPNET-512: infraconfig: grab ownership from cluster-bootstrap

mkowalski:
Infrastructure CR is an object managed by at least 2 actors, cluster-bootstrap and network-operator. The former initializes the object with values provided at install-time, the latter manages the object according to the user's requests.

Because of the type used in the API definition of PlatformStatus (`set` and not `atomic`), it is not trivial for the controller to perform server-side apply that modifies values owned by cluster-bootstrap.

In order to overcome the issue, we are adding a logic that removes existing fieldManagers and makes network-operator the owner. This way we are able to not only add values but modify and delete already existing ones.

### DIFF
--- a/pkg/controller/infrastructureconfig/infrastructureconfig_controller.go
+++ b/pkg/controller/infrastructureconfig/infrastructureconfig_controller.go
@@ -6,38 +6,51 @@ import (
 	"log"
 	"reflect"
 
-	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


mkowalski:
/hold

https://github.com/openshift/api/pull/1873 will be more elegant way to fix it

mkowalski:
/close

Fixing via https://github.com/openshift/api/pull/1873

PR #2359, opened 2024-04-26
NP-975: Stop OpenShift SDN control plane during live migration

pliurh:
During SDN live migration, both OpenShift and OVNKubernetes run in parallel within a cluster. If a new node is added to the cluster, the control planes of both CNIs attempt to allocate a node subnet for it, resulting in a conflict. To address this, the OpenShift SDN control plane is halted during live migration, and the migration script within the ovnkube-node pod takes responsibility for creating the hostsubnet for the new node, and make the allocated node subnet aligned in both CNIs.

### DIFF
--- a/bindata/network/openshift-sdn/controller.yaml
+++ b/bindata/network/openshift-sdn/controller.yaml
@@ -28,9 +28,8 @@ spec:
         - -c
         - |
 {{- if .IsNetworkTypeLiveMigration }}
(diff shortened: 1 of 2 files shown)


pliurh:
Stop OpenShift SDN controll plane will cause issues when new namespaces are created during the migration. We need to find another approach.

PR #2360, opened 2024-04-26
OCPBUGS-32979: Update network operator status for IPsec

pperiyasamy:
This PR updates network operator progressing condition for rolling out IPsec daemonset for cluster. It reflects appropriate status for network cluster operator which would make openshift-install to exit gracefully after installing all of IPsec components for the cluster.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -268,6 +268,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVNIPsecDaemonsetEnable"] = OVNIPsecDaemonsetEnable
 	data.Data["OVNIPsecEnable"] = OVNIPsecEnable
 
+	// Set progressing to true until IPsec DaemonSet is rendered when EW IPsec config is enabled.
+	// TODO Do a poor man's job mapping machine config pool status to CNO progressing state for now.
+	// This has two problems:
+	// - Not a great feedback to the user on why we are progressing other than `Waiting to render manifests`.
+	// - If pool status degrades due to CNO's changes, CNO stays progressing where it would be
+	//   potentially better to report it as degraded as well.
+	// Overall, mapping machine config pool status to CNO status should better be done in status manager.
+	// Future efforts on this are tracked in https://issues.redhat.com/browse/SDN-4829.
+	progressing = OVNIPsecDaemonsetEnable && !renderIPsecHostDaemonSet && !renderIPsecContainerizedDaemonSet
+
 	klog.V(5).Infof("IPsec: is MachineConfig enabled: %v, is East-West DaemonSet enabled: %v", data.Data["IPsecMachineConfigEnable"], data.Data["OVNIPsecDaemonsetEnable"])
 
 	if c.GatewayConfig != nil && c.GatewayConfig.RoutingViaHost {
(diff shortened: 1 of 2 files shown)


jcaamano (commented):
PTAL @kyrtapz @pliurh 

I would like commit 2 general opinion here and also in context of live migration which is another functionality that depends on pool status.

I think we should reconcile on pool status changes, but regardless this makes sense.

jcaamano:
@pperiyasamy please note that this will delay installs if ipsec is enabled from the get go, possibly delaying them over an installer global timeout and making them fail. We might need to account for this in ipsec CI jobs and also in docs if the default timeout is not big enough.

pperiyasamy:
> @pperiyasamy please note that this will delay installs if ipsec is enabled from the get go, possibly delaying them over an installer global timeout and making them fail. We might need to account for this in ipsec CI jobs and also in docs if the default timeout is not big enough.

yes @jcaamano , The `ci/prow/e2e-ovn-ipsec-step-registry`  job is so far looking fine with change, but it would be also good to get this release [PR](https://github.com/openshift/release/pull/50740) merged (as this makes ipsec jobs to adapt with new API changes) and look for few more runs how it behaves. cc @jluhrsen

pperiyasamy:
This [1788117489796780032](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2360/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1788117489796780032) run shows change is in effect ([here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2360/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1788117489796780032/artifacts/e2e-ovn-ipsec-step-registry/ipi-install-install/artifacts/.openshift_install-1715158847.log)) and COs are back to running in 28mins. Would try for few more runs.

pkg/network/ovn_kubernetes.go
```diff
@@ -268,6 +268,9 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["OVNIPsecDaemonsetEnable"] = OVNIPsecDaemonsetEnable
 	data.Data["OVNIPsecEnable"] = OVNIPsecEnable
 
+	// Set progressing to true until IPsec DaemonSet is rendered when EW IPsec config is enabled.
```
jcaamano, line 271: @pperiyasamy can you add a TODO here explaining what we want to do in the future and/or pointing to the jira issue? thanks!

pperiyasamy: Done. Thanks a lot @jcaamano and @kyrtapz for a quick offline reviews on this PR.

pkg/network/ovn_kubernetes.go
```diff
@@ -268,6 +268,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	// TODO Do a poor man's job mapping machine config pool status to CNO progressing state for now.
+	// This has two problems:
+	// - Not a great feedback to the user on why we are progressing other than `Waiting to render manifests`.
+	// - If pool status degrades due to CNO's changes, CNO stays progressing where it would be
+	//   potentially better to report it as degraded as well.
+	// Overall, mapping machine config pool status to CNO status should better be done in status manager.
+	// Future effort on this are tracked in https://issues.redhat.com/browse/SDN-4829.
```
jcaamano, line 278: nit: `efforts`

PR #2362, opened 2024-04-29
OCPBUGS-33080: Update EQoS CRD to latest

tssurya:
Bring in the changes from https://github.com/ovn-org/ovn-kubernetes/pull/4158 for 4.16

cc @pperiyasamy

### DIFF
--- a/bindata/network/ovn-kubernetes/common/001-crd.yaml
+++ b/bindata/network/ovn-kubernetes/common/001-crd.yaml
@@ -320,8 +320,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
-  creationTimestamp: null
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -316,12 +316,12 @@ status:
   conditions: []
   storedVersions: []
 ---
+---
```
pperiyasamy, line 319: nit: can you remove this extra separator ?

tssurya: oh woops haha

tssurya: done! thanks for catching that Peri!

kyrtapz:
@pperiyasamy we could use some downstream e2e tests in origin for this feature to avoid bugs like this one.
/lgtm

PR #2364, opened 2024-05-01
OCPBUGS-28230: enforce termination message policy on all platform pods

deads2k:
maybe the last?


deads2k:
just like the others, tagging

PR #2367, opened 2024-05-08
NP-975: Align new node subnet allocation between OVN-K and SDN during live migration

pliurh:
During live migration, we run ovn-k and sdn in parallel. When new nodes are added to the cluster, ovn-k and sdn may assign different subnets to the nodes. This patch lets ovn-k subject to sdn for the node subnet allocation when both CNIs are deployed.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -67,26 +67,45 @@ spec:
           #!/bin/bash
           set -euo pipefail
 {{- if .IsNetworkTypeLiveMigration }}
-          # In case of rollback, openshift-sdn need to reuse the host subnet used by ovn-kube, so we need to create the hostsubnet CR for the local node if it doesn't exist.
+          set -x
+          NODE_CNI="OpenShiftSDN"
+          if ip link show br-ex; then
+            echo "br-ex exists"
+            NODE_CNI="OVNKubernetes"
+          else
+            echo "br-ex doesn't exist"
+          fi
+
           # retry loop
           RETRY_INTERVAL=3
-          MAX_RETRIES=10
+          MAX_RETRIES=100
           for ((retry_count=1; retry_count<=MAX_RETRIES; retry_count++)); do
-              # Get SDN subnet
-              SDN_SUBNET=$(oc get hostsubnet ${K8S_NODE_NAME} -o jsonpath="{.subnet}"||true)
+            # Get SDN subnet
+            SDN_SUBNET=$(oc get hostsubnet ${K8S_NODE_NAME} -o jsonpath="{.subnet}"||true)
 
-              # Get OVN subnet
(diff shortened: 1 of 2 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -67,26 +67,45 @@ spec:
...
+                break
+              fi
+            else
+              if [ -z "${OVN_SUBNET}" ]; then
+                echo "OVN node subnet is not found, retry..."
+                sleep $RETRY_INTERVAL
+              else
+                # When rolling back, SDN controll plane will not start until the hostsubnet has been created here.
+                echo "Use OVN-K subnet ${OVN_SUBNET}, when the node CNI is OVN-K"
+                if [ -z "${SDN_SUBNET}" ]; then
```
kyrtapz, line 103: Why are you no longer verifying that SDN_SUBNET is equal to OVN_SUBNET if both are set?
Wouldn't that cause issues when  a node is added during the migration to SDN and it is still using ovn-k but the sdn-controller allocates a subnet for it?

pliurh: In that case, ovnkube-node will use the node subnet allocated by SDN. So we don't care whether the SDN_SUBNET is equal to the OVN_SUBNET. It will be overwritten regardless. And here, the SDN pod will wait until the OVN_SUBNET is set by the ovnkube-node pod.

bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -67,26 +67,45 @@ spec:
...
+            echo "br-ex exists"
+            NODE_CNI="OVNKubernetes"
+          else
+            echo "br-ex doesn't exist"
+          fi
+
           # retry loop
           RETRY_INTERVAL=3
-          MAX_RETRIES=10
+          MAX_RETRIES=100
```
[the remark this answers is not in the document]
pliurh: This is to reduce the number of pod timeouts during the migration. By retaining more logs, it aids in debugging. I should increase the timeout for OVN-K pods too.

bindata/network/openshift-sdn/sdn.yaml
kyrtapz: This is about OVN-K -> SDN migration.
- A new node is added during the migration and it still uses ovn-k
- Both CNIs allocate the subnet
- In sdn pod `if [ -z "${SDN_SUBNET}" ]; then` is not going to be true and SDN will use the subnet allocated by SDN which can conflict with a subnet allocate by ovn-k.

Does this make sense? In the current code we are not going to overwrite the hostsubnet if there is already one specified.

kyrtapz: We synced offline, here is how it would look like for sdn->ovn-k migration:
- Migration starts and both controllers started after syncing on current subnets (sdn container created all hostsubnets from ovn subnets)
- A new node is added, both controllers allocate a subnet (it shouldn't conflict with any existign subnets but can be different for each CNI)
- With the changes from this PR the ovn-k pod will first overwrite the ovn-k subnet with the value allocated by SDN

martinkennelly:
/retitle OCPBUGS-34659,NP-975: Align new node subnet allocation between OVN-K and SDN during live migration

PR #2370, opened 2024-05-15
OCPBUGS-33728: use v4-masquerade-subnet config str

jluhrsen:
ovn-kubernetes expects v4-masquerade-subnet not
v4-internal-masquerade-subnet:

  https://github.com/ovn-org/ovn-kubernetes/blob/f5d4dfb6acef57d9a4cad95fedd672dfe4112f9a/go-controller/pkg/config/config.go#L444


jluhrsen:
let's use @arghosh93 PR here:
https://github.com/openshift/cluster-network-operator/pull/2371/

PR #2372, opened 2024-05-16
Remove ip xfrm state when IPsec is disabled

pperiyasamy:
The ovn-ipsec-host daemonset doesn't remove ipsec state when IPsec for east west traffic is disabled. This might still keep running data plane traffic with ipsec encrypted. So this PR cleans up ipsec state so that east west traffic restored with only geneve header.


bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -589,6 +589,19 @@ data:
         ovn_v6_join_subnet_opt="--gateway-v6-join-subnet {{.V6JoinSubnet}}"
       fi
 
+{{- if not .OVNIPsecEnable }}
+      # flush out xfrm state and policy entries so that it cleans up
+      # entries created for ovs. Since it's cleaning up whole entries,
+      # it may cause slight interruption in case of external ipsec
+      # configuration.
+      echo "remove ovs ipsec state if present"
```
yuvalk, line 597: problem is, this also removes the N-S state, not just ovs
so need to verify OVNIPsecEnable actually means both?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -589,6 +589,19 @@ data:
...
+      ip x s flush
+      ip x p flush
+      rm -f /etc/ipsec.d/openshift.conf || true
+      # if pluto is running, we need to restart it after the flush
+      chroot /proc/1/root ipsec restart || true
```
yuvalk, line 602: @bengal I remember there used to be a NM issue around this, not sure if fixed?

bengal: AFAIK, restarting ipsec while there are active connections created by NM-libreswan would break them. You need to reactivate such connections manually to restore connectivity.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
pperiyasamy: yes, it also does ipsec service restart and wouldn't this get ipsec state for NS back into the system ?
Moreover this logic is similar to what ovn-ipsec container does [here](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-host.yaml#L231-L238)  and same thing happens for container restart scenarios as well. 

The `OVNIPsecEnable` flag is used only to identify if EW IPsec is enabled or not.

The IPsec mode is changed in following order `Full -> Disabled -> External`, EW ipsec connections in `openshift.conf ` occasionally not cleaned up even after IPsec disabled, then external mode gets back those connections active again.  we may delete only `openshift.conf `file to handle this case,  but not sure about the behavior for `Full -> External` . Hence this change is introduced to handle all the cases.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
yuvalk: doesnt feels very production grade.
if pluto daemon restart, for any reason (error? crash? etc)
all the connections will die and wont even attempt to restart

bengal: Right, in case of crashes that's a problem. Is this issue already filed somewhere? If not please do, so that we can plan how to improve this in NM-libreswan.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
yuvalk, line 597: another important issue with this, is that AFAIK, it'll create connectivity issues for workloads!

removing the state, forces ipsec off, but if the other node is not updated at exact the same moment, it'll still try to do ipsec (according to it's policy)

I dont think there's a good solution for this besides leaving ipsec on until next reboot

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -397,6 +333,125 @@ spec:
...
+              ip x s flush
+              ip x p flush
+              rm -f /etc/ipsec.d/openshift.conf
+              # since pluto is on the host, we need to restart it after the flush
+              chroot /proc/1/root ipsec restart
+            fi
+          }
+
+          # Function to handle SIGTERM
+          cleanup() {
```
trozet, line 409: when a pod is deleted and all containers receive sigterm, i guess this will go ahead and flush before ovs-monitor-ipsec exits. Maybe you should wait a maximum of like 15 seconds or something for ovs-monitor script to exit before you do the ipsec cleanup. I think default termination grace period is 30 seconds, so waiting 15 should be ok.

pperiyasamy: yes @trozet , good find. done.

pperiyasamy (commented):
The `e2e-ovn-ipsec-step-registry` CI lane is continuously failing last 3 times on this PR. The ipsec logs are clean from collected mg report and IKE SAs establishments exist across all the nodes. Not sure this [change](https://github.com/openshift/cluster-network-operator/pull/2372/files#diff-60c1502ee55d2c965b4452d8adb3b8ad5f305eeea705fe65300902f1b68b80ddL233-L238) causes a problem during ipsec pod reboots while running e2e tests.  need to look at this further.

trozet (approved):
/approve

trozet:
@pperiyasamy you need to set proper requests for the new container you added:

```
  apps/v1/DaemonSet/openshift-ovn-kubernetes/ovn-ipsec-host/container/ovn-ipsec-cleanup does not have a cpu request (rule: "apps/v1/DaemonSet/openshift-ovn-kubernetes/ovn-ipsec-host/container/ovn-ipsec-cleanup/request[cpu]")
  apps/v1/DaemonSet/openshift-ovn-kubernetes/ovn-ipsec-host/container/ovn-ipsec-cleanup does not have a memory request (rule: "apps/v1/DaemonSet/openshift-ovn-kubernetes/ovn-ipsec-host/container/ovn-ipsec-cleanup/request[memory]")
```

pperiyasamy:
created a bug for `e2e-ovn-ipsec-step-registry` failure: https://issues.redhat.com/browse/OCPBUGS-45205

jcaamano:
/lgtm
/approve
/hold

Waiting for CI

pperiyasamy:
The ipsec test failures with `e2e-aws-ovn-ipsec-serial` lane are known failures, part of it is fixed with PR https://github.com/openshift/origin/pull/28797 (external mode rollout completeness check) and remaining failures are due to bug [OCPBUGS-41823](https://issues.redhat.com/browse/OCPBUGS-41823), seeing the error from pluto logs.

```
Dec 04 17:06:59.881622 ip-10-0-1-22 pluto[27549]: "ovn-e0a41e-0-out-1" #49: No IKEv2 connection found with compatible Traffic Selectors
Dec 04 17:06:59.881637 ip-10-0-1-22 pluto[27549]: "ovn-e0a41e-0-out-1" #49: responding to CREATE_CHILD_SA message (ID 2) from 10.0.108.160:500 with encrypted notification TS_UNACCEPTABLE
```

huiran0826:
Pre-merge tested this PR, after changing ipsec mode from full to external. I checked below items
1. ovn-ipsec-host pods  disappeared
2. ipsec state/policy and /etc/ipsec.d/openshift.conf was cleaned up
Regarding upgrade, I did an upgrade from 4.18.0-ec.3 to the PR build, upgrade succeeded
Before upgrade
%  oc get pods -n openshift-ovn-kubernetes -l app=ovn-ipsec
ovn-ipsec-host-crzgp   1/1     Running   0          15m
ovn-ipsec-host-fcr2m   1/1     Running   0          15m
ovn-ipsec-host-gj7n4   1/1     Running   0          15m
ovn-ipsec-host-h84n2   1/1     Running   0          15m
ovn-ipsec-host-rwssh   1/1     Running   0          15m
ovn-ipsec-host-smkdk   1/1     Running   0          15m 
After ugprade
% oc get pods -n openshift-ovn-kubernetes
NAME                                     READY   STATUS    RESTARTS   AGE
ovn-ipsec-host-8d7x5                     2/2     Running   0          71m
ovn-ipsec-host-h6k9s                     2/2     Running   0          71m
ovn-ipsec-host-h9wd6                     2/2     Running   0          71m
ovn-ipsec-host-j8ksn                     2/2     Running   0          71m
ovn-ipsec-host-smxjn                     2/2     Running   0          71m
ovn-ipsec-host-xv8gq                     2/2     Running   0          71m

/label qe-approved

PR #2378, opened 2024-05-22
OCPBUGS-34214: Add conditions for ignored-namespaces

s1061123:
This code changes to introduce additional condition check for ignored-namespaces, whether the ns is openshift component or not


s1061123:
Checked that security issue is not related to my changes. just FYI.

PR #2382, opened 2024-05-23
OCPBUGS-34359: Ensure that the node-identity webhook address contains colons for IPv6

kyrtapz:
For clusters with IPv6 as the primary network the `network-node-identity.openshift.io` admission webhook should have the IP address in brackets.
The only actual change is that we include brackets in the webhook address when using ipv6 (`url: https://::1:9743/node` changes to `url: https://[::1]:9743/node`)

The issue was observed during testing of the k8s 1.30 rebase in which the webhook client started using http2 for loopback IPs: https://github.com/kubernetes/kubernetes/pull/122558.
It looks like the issue is caused by how a http2 client handles this invalid address, I verified this change by setting up a cluster with https://github.com/openshift/kubernetes/pull/1953 and this pr.

### DIFF
--- a/bindata/network/node-identity/self-hosted/node-identity-pki.yaml
+++ b/bindata/network/node-identity/self-hosted/node-identity-pki.yaml
@@ -7,4 +7,4 @@ metadata:
   namespace: openshift-network-node-identity
 spec:
   targetCert:
-    commonName: {{.NetworkNodeIdentityAddress}}
+    commonName: {{.NetworkNodeIdentityIP}}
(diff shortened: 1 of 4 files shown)


soltysh (approved):
Confirmed this manually running a cluster-bot cluster from this PR combined with https://github.com/openshift/kubernetes/pull/1953. Previously the worker nodes would be stuck with `NotReady` state, with this fix they become immediately `Ready`.

/lgtm

kyrtapz:
Both e2e-gcp-ovn and e2e-azure-ovn-upgrade seem to be perma failing across different PRs.
e2e-gcp-ovn failures are already tracked here: https://issues.redhat.com/browse/TRT-1680.
I've looked into the azure failures and there is nothing indicating they are related to this change.
This PR only affects ipv6 clusters so it is highly unlikely that it contributes the the failures. 

/override ci/prow/e2e-gcp-ovn

PR #2383, opened 2024-05-23
Add machine config pool watcher for status manager

pperiyasamy:
This enhances status manager to watch for machine config pool object, retrieves machine config pool status when ipsec is enabled and when any one of the machine config pool goes into degraded state, then makes network operator also into degraded state.


pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,88 @@
...
+	c, err := controller.New("machineconfigs-watcher", mgr, controller.Options{Reconciler: pw})
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Informer{Informer: s.mcInformer,
+		Handler: handler.EnqueueRequestsFromMapFunc(enqueueRP)})
+	if err != nil {
+		return err
+	}
```
jcaamano, line 77: Any reason you wouldn't be able to use the generic `kind` source?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -52,6 +52,7 @@ const (
 	CertificateSigner
 	InfrastructureConfig
 	DashboardConfig
+	MachineConfig
```
jcaamano, line 55: I think this list is ordered from from more important to least important problems. So I would say this probably needs to be above PodDeployment.

pkg/controller/statusmanager/ipsec_status.go
```diff
@@ -0,0 +1,151 @@
...
+	defer status.Unlock()
+
+	master, err := status.ipsecMachineConfigExists(platform.MasterRoleMachineConfigLabel)
+	if err != nil {
+		log.Printf("failed to retrieve machine configs for master role: %v", err)
+	}
+	worker, err := status.ipsecMachineConfigExists(platform.WorkerRoleMachineConfigLabel)
+	if err != nil {
+		log.Printf("failed to retrieve machine configs for worker role: %v", err)
+	}
```
jcaamano, line 26: This should be more generic. The status manager should become aware of the specific MachineConfigs that we need rendered regardless of them being for ipsec or any other potential thing in the future, just as, for example, it is aware of the needed daemon sets. deployments, etc. So other controllers need to add that information to the status manager or we can rely on the owner reference. 

It should also check if those machine configs are been rendered into the pool or no, not just if the pool is degraded or progressing.

jcaamano (changes requested):
Some generic comments first.

With this, I also think we should no longer return progressing from the render method. That was a trick we should not have used.

And would also like @kyrtapz to take a look eventually.

pkg/controller/statusmanager/machineconfig_watcher.go
jcaamano: Sorry for the lack of detail. I meant use the kind source instead of the informer source so you don't have to deal with the informers:
https://github.com/kubernetes-sigs/controller-runtime/blob/3b032e16c0dc19d656626a288cd417d36beaebad/pkg/source/source.go#L73

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,88 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+type MachineConfigsWatcher struct {
+	status *StatusManager
+}
+
+func (s *StatusManager) initInformerForMachineConfigs(cfg *rest.Config) error {
+	configClient, err := configv1client.NewForConfig(cfg)
```
jcaamano, line 26: If you use a new client, you won't share the cache with the manager. You wouldn't have this problem with the kind source since it forces you to think about what cache you are using, However if you want to use infromers anyway, use the same client used for the manager.

jcaamano (changes requested):

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,110 @@
...
+		CreateFunc: func(e event.CreateEvent) bool {
+			mc := e.Object.(*mcfgv1.MachineConfig)
+			return hasRequiredLabel(mc)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			mc := e.ObjectNew.(*mcfgv1.MachineConfig)
+			return hasRequiredLabel(mc)
+		},
+	}
+}
```
jcaamano, line 71: Why is this predicate not based on the owner reference?
Why is this predicate not filtering delete events?

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,110 @@
...
+		CreateFunc: func(e event.CreateEvent) bool {
+			mcp := e.Object.(*mcfgv1.MachineConfigPool)
+			return hasRequiredMachineConfigSelector(mcp)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			mcp := e.ObjectNew.(*mcfgv1.MachineConfigPool)
+			return hasRequiredMachineConfigSelector(mcp)
+		},
+	}
+}
```
jcaamano, line 84: As is, you could have just used the `predicate.LabelSelectorPredicate`.
However, we are not handling updates that remove the label. Are we not expecting that to happen?

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,156 @@
...
+	workerMachineConfigs, err := status.networkMachineConfigs(ctx, platform.WorkerRoleMachineConfigLabel)
+	if err != nil {
+		log.Printf("failed to retrieve machine configs for worker role: %v", err)
+	}
+	// When both master and worker roles don't have any machine configs owned by network operator, then return now.
+	if len(masterMachineConfigs) == 0 && len(workerMachineConfigs) == 0 {
+		status.setNotDegraded(MachineConfig)
+		status.unsetProgressing(MachineConfig)
+		return
+	}
```
jcaamano, line 33: Not sure this is entirely true. Don't we also need to check that the pool status is not referencing any machine config that had been added by us? Otherwise it would mean they are still progressing towards removing our machine configs.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,156 @@
...
+			progressing = true
+			break
+		}
+	}
+	return progressing
+}
+
+func (status *StatusManager) findMachineConfigPoolsForLabel(ctx context.Context, mcLabel labels.Set) ([]*mcfgv1.MachineConfigPool, error) {
+	mcPoolList := &mcfgv1.MachineConfigPoolList{}
+	err := status.client.Default().CRClient().List(ctx, mcPoolList, &client.ListOptions{})
```
jcaamano, line 126: Any reason not to use the cache?

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,156 @@
...
+		if mcSelector.Matches(mcLabel) {
+			mcPools = append(mcPools, &mcPoolList.Items[i])
+		}
+	}
+	return mcPools, nil
+}
+
+func (status *StatusManager) networkMachineConfigs(ctx context.Context, mcLabel labels.Set) ([]*mcfgv1.MachineConfig, error) {
+	mcs := &mcfgv1.MachineConfigList{}
+	err := status.client.Default().CRClient().List(ctx, mcs, &client.ListOptions{LabelSelector: mcLabel.AsSelector()})
```
jcaamano, line 145: Any reason not to use the cache?

pkg/network/ovn_kubernetes.go
```diff
@@ -2050,3 +2014,33 @@ func GetMasqueradeSubnet(conf *operv1.OVNKubernetesConfig) (v4Subnet, v6Subnet s
...
+		}
+	}
+	return false
+}
+
+// AreMachineConfigsRenderedOnPool returns true if machineConfigs are completely rendered on the given machine config
+// pool status, otherwise returns false.
+func AreMachineConfigsRenderedOnPool(status mcfgv1.MachineConfigPoolStatus, machineConfigs []*mcfgv1.MachineConfig) bool {
+	return status.MachineCount == status.UpdatedMachineCount && hasSourceInMachineConfigStatus(status, machineConfigs)
+}
```
jcaamano, line 2034: Should we find a better place for these? They are not `ovn-kubernetes` specific.

jcaamano (changes requested):

pkg/controller/statusmanager/machineconfig_watcher.go
pperiyasamy: The label check has to be done on the `mcp.Spec.MachineConfigSelector`, so can't use `predicate.LabelSelectorPredicate` directly. yes, need to handle delete event for mcp as well, just adding predicate sufficient for this, would do it in next update.

pkg/network/ovn_kubernetes.go
pperiyasamy: moved these functions into platform module, not seeing any other modules that would fit for now.

pperiyasamy:
can not see `master role machine config pool in progressing state` (or) `worker role machine config pool in progressing state` messages from latest ipsec job's install [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2383/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1864654444914806784/artifacts/e2e-ovn-ipsec-step-registry/ipi-install-install/artifacts/.openshift_install-1733407652.log), but when i tested this [change](https://github.com/pperiyasamy/cluster-network-operator/commit/9f34f8471d01b1bcf68504676a36733fb88ed169) on 4.16, it was working, let me debug this on a bot cluster.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -103,6 +105,9 @@ type StatusManager struct {
 
 	relatedObjects []configv1.ObjectReference
 
+	// local cache to store deployed network operator machine configs.
+	availableMachineConfigs map[string]sets.Set[string]
```
jcaamano, line 109: nit: 

what do you mean with `available`? The term gives the feeling that they are already in effect. `Maybe use `rendered`?

BTW, wouldn't these machine configs be among the `relatedObjects` above as well. And an additional consideration, for pods there is no cache, they are listed every reconcile. I guess we could have done the same as an alternative although if we are filtering by owner reference it woudl be not as efficient as listing by the explicit label cNop is setting on pod related resources. I guess I don't have anything against the current approach but I am curious if you considered other options.

pkg/platform/platform.go
```diff
@@ -298,3 +303,44 @@ func IsUserDefinedIPsecMachineConfig(machineConfig *mcfgv1.MachineConfig) bool {
...
+}
+
+func checkSourceInMachineConfigStatus(machineConfigStatus mcfgv1.MachineConfigPoolStatus, machineConfigs sets.Set[string],
+	test func(sets.Set[string], sets.Set[string]) bool) bool {
+	sourceNames := sets.New[string]()
+	for _, source := range machineConfigStatus.Configuration.Source {
+		sourceNames.Insert(source.Name)
+	}
+	return test(sourceNames, machineConfigs)
+}
```
jcaamano, line 346: I guess another option would have been to place these somewhere in the util package

jcaamano (changes requested):
Looks good, minor comments.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,242 @@
...
+		return deleteMachineConfigs
+	}
+	lsbytes := co.Annotations[lastSeenMachineConfigAnnotation]
+	if lsbytes == "" {
+		return deleteMachineConfigs
+	}
+	out := machineConfigState{}
+	err = json.Unmarshal([]byte(lsbytes), &out)
+	if err != nil {
+		// No need to return error; just move on
```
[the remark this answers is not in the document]
pperiyasamy: it needed CO object to be created in the unit test beforehand, added it there.

pkg/controller/statusmanager/status_manager.go
jcaamano: `relatedObjects` would be the cache, you wouldn't need `availableMachineConfigs`.

So the thing is: 

what is our state from the time we render a MachineConfig (more-less the same time it is added to related objects) to the time we receive the pool update event (when the update our status accordingly)?

In between we have to wait to receive an event for the machine config itself. And this event can be out of order with respect to the pool event.

So you should consider:
* would it be better to update our status as soon as a machine config is rendered?
* otherwise, should we update our status  when we receive a machine config event as well?

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,255 @@
...
+		status.setProgressing(MachineConfig, "MachineConfig", fmt.Sprintf("%s machine config pool in progressing state", progressingPool))
+		return true
+	}
+	for _, pool := range pools {
+		if pool.Spec.Paused {
+			// When a machine config pool is in paused state, then it skips rolling out (or) removing any eligible machine configs.
+			// It is not appropriate to consider this pool for reporting status, so just ignore it.
+			continue
+		}
+		rendered := test(pool.Status, machineConfigs)
```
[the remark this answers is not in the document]
pperiyasamy: Every machine config is labeled with a role, The `machineConfigs`  at here is already filtered out in `status.findMachineConfigPoolsForLabel`, so we are testing against relevant pool. so i think we are safer here.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,255 @@
...
+
+	progressingPool := status.isMachineConfigPoolProgressing(pools)
+	if progressingPool != "" {
+		status.setProgressing(MachineConfig, "MachineConfig", fmt.Sprintf("%s machine config pool in progressing state", progressingPool))
+		return true
+	}
+	for _, pool := range pools {
+		if pool.Spec.Paused {
+			// When a machine config pool is in paused state, then it skips rolling out (or) removing any eligible machine configs.
+			// It is not appropriate to consider this pool for reporting status, so just ignore it.
```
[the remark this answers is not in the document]
pperiyasamy: updated the comment now about why we are not reporting status for a pool when it's in paused state, that is because progressing state would block network upgrade. we anyway need to get https://github.com/openshift/cluster-network-operator/pull/2449 to move into newer ipsec daemonset is immediately irrespective of mcp state, this is going to really unblock the network upgrade.  Let's work on that PR once this is merged.

pkg/controller/statusmanager/status_manager.go
pperiyasamy: yes @jcaamano , nice suggestion, this helps to report the status timely as soon as machine configs are rendered. as discussed, introduced `SetMachineConfigs` method in status manager, updating caches and reporting status accordingly.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,315 @@
...
+			needStateUpdate = true
+			deletedMachineConfigMap[role].Delete(recreatedMachineConfigs.UnsortedList()...)
+		}
+	}
+	if !needStateUpdate {
+		return
+	}
+	if err := status.setLastDeletedMachineConfigState(deletedMachineConfigMap); err != nil {
+		klog.Errorf("failed to set deleted machine config state: %v", err)
+	}
```
jcaamano, line 170: My idea was a bit different here. 

Annotate the rendered machine configs role->name. Have a getter for this that reads the annotation if renderedMachineConfigs is not initialized yet. And a setter to both set renderedMachineConfigs and annotate.

The deleted machine configs are the difference between `newRenderedMachineConfigs` and renderedMachineConfigs. 
Set the deleted machine configs on a separate field, no need to annotate them.

Then call SetFromMachineConfigPool, and it will just set the status.

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil && apierrors.IsNotFound(err) {
+		p.status.processDeletedMachineConfig(request.NamespacedName.Name)
+		return reconcile.Result{}, nil
+	}
+	if err != nil {
+		klog.Errorf("failed to retrieve machine config: %v", err)
+		return reconcile.Result{}, nil
+	}
+	p.status.processCreatedMachineConfig(*mc)
+	return reconcile.Result{}, nil
```
jcaamano, line 71: This should just call SetFromMachineConfigPool. We don't really care about anything else.

pkg/controller/statusmanager/machineconfig_status.go
pperiyasamy: @jcaamano awesome idea on keeping rendered machine configs in the annotation cache until it gets deleted from the cluster. This is making those functions very clean. updated the commit bf88169 now. PTAL.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	s.Lock()
+	defer s.Unlock()
+	if s.renderedMachineConfigs == nil {
+		s.renderedMachineConfigs, err = s.getLastRenderedMachineConfigState()
+		if err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jcaamano: Also no need to check `s.renderedMachineConfigs == nil`, right? Because this is part of the startup code.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+	for _, pool := range pools {
+		if mcfgv1.IsMachineConfigPoolConditionTrue(pool.Status.Conditions, mcfgv1.MachineConfigPoolDegraded) {
+			degradedPool = pool.Name
+			break
+		}
+	}
+	return degradedPool
+}
+
+func (status *StatusManager) isMachineConfigPoolProgressing(pools []mcfgv1.MachineConfigPool) string {
```
jcaamano, line 225: nit: `isAnyMachineConfigPoolProgressing`

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+						status.machineConfigsBeingRemoved[role].Delete(machineConfig)
+						status.renderedMachineConfigs[role].Delete(machineConfig)
+						if err := status.setLastRenderedMachineConfigState(status.renderedMachineConfigs); err != nil {
+							klog.Errorf("failed to update rendered machine config state: %v", err)
+						}
+					}
+				} else {
+					rendered = mcutil.AreMachineConfigsRenderedOnPool(pool.Status, mcSet)
+				}
+				if !rendered {
```
jcaamano, line 161: The use of rendered is a bit misleading.
I would probably do 
```
removed := ...
...
added := ...
...
if !added || !removed {
    ...
}
```

pkg/controller/statusmanager/status_manager.go
```diff
@@ -117,6 +118,11 @@ type StatusManager struct {
 
 	relatedObjects []configv1.ObjectReference
 
+	// local cache to store rendered network operator machine configs.
+	renderedMachineConfigs map[string]sets.Set[string]
```
jcaamano, line 122: If you take into accound my other comments, I think we can initialize this to an empty map for our peace of mind.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+	mcPools := &mcfgv1.MachineConfigPoolList{}
+	err = status.client.ClientFor("").CRClient().List(ctx, mcPools)
+	if err != nil {
+		klog.Errorf("failed to retrieve machine config pools: %v", err)
+		return
+	}
+	status.SetFromMachineConfigPool(mcPools.Items)
+}
+
+func (status *StatusManager) SetFromMachineConfigPool(mcPools []mcfgv1.MachineConfigPool) {
```
jcaamano, line 117: Since this is the main reconciliation loop, I think it is a good place to add doctext for the method giving a brief description on how it works. For example, is not easy to grasp how, annotating the MCs even if they are being removed until we actually confirm they have been removed, is key to handle restarts gracefully.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+			renderedMachineConfigMap[role].Insert(new.UnsortedList()...)
+			annotateUpdate = true
+		}
+	}
+	if annotateUpdate {
+		status.renderedMachineConfigs = renderedMachineConfigMap
+		if err := status.setLastRenderedMachineConfigState(renderedMachineConfigMap); err != nil {
+			klog.Errorf("failed to set rendered machine config state: %v", err)
+		}
```
jcaamano, line 105: Why not return an error?
```
		if err := status.setLastRenderedMachineConfigState(renderedMachineConfigMap); err != nil {
			klog.Errorf("failed to set rendered machine config state: %v", err)
		}
        status.renderedMachineConfigs = renderedMachineConfigMap
```

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+func (status *StatusManager) SetFromMachineConfigPool(mcPools []mcfgv1.MachineConfigPool) {
+	status.Lock()
+	defer status.Unlock()
+	// The status.renderedMachineConfigs is a non-nil map at the time when SetFromMachineConfigPool method is invoked.
+	for role, machineConfigs := range status.renderedMachineConfigs {
+		pools, err := status.findMachineConfigPoolsForLabel(mcPools, map[string]string{platform.MachineConfigLabelRoleKey: role})
+		if err != nil {
+			klog.Errorf("failed to get machine config pools for the role %s: %v", role, err)
```
jcaamano, line 124: Please return the errors.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+	status.Lock()
+	var err error
+	if status.renderedMachineConfigs == nil {
+		status.renderedMachineConfigs, err = status.getLastRenderedMachineConfigState()
+		if err != nil {
+			status.Unlock()
+			klog.Errorf("failed to get rendered machine config state: %v", err)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
pperiyasamy: good catch on handling status.renderedMachineConfigs cache. updated the code accordingly.

pkg/controller/statusmanager/machineconfig_status.go
```diff
@@ -0,0 +1,248 @@
...
+	if annotateUpdate {
+		status.renderedMachineConfigs = renderedMachineConfigMap
+		if err := status.setLastRenderedMachineConfigState(renderedMachineConfigMap); err != nil {
+			klog.Errorf("failed to set rendered machine config state: %v", err)
+		}
+	}
+	status.Unlock()
+	mcPools := &mcfgv1.MachineConfigPoolList{}
+	err = status.client.ClientFor("").CRClient().List(ctx, mcPools)
```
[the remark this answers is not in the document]
pperiyasamy: as discussed, let's keep using kube client to list objects in status manager for now.

pkg/controller/statusmanager/machineconfig_status.go
pperiyasamy: done.

pkg/names/names.go
```diff
@@ -172,6 +172,19 @@ const (
...
+	MachineConfigLabelRoleKey = "machineconfiguration.openshift.io/role"
+)
+
+var (
+	// When user deploys their own machine config for installing and configuring specific version of libreswan, then
+	// corresponding master and worker role machine configs annotation must have `user-ipsec-machine-config: true`.
+	// When CNO finds machine configs with the annotation, then it skips rendering its own IPsec machine configs
+	// and reuse already deployed user machine configs for the ovn-ipsec-host daemonset.
+	UserDefinedIPsecMachineConfigAnnotation = map[string]string{"user-ipsec-machine-config": "true"}
+)
```
jcaamano, line 186: Why didn't you move 	masterRoleMachineConfigLabel nad workerRoleMachineConfigLabel here as well.

And we can also make the trick of the method with UserDefinedIPsecMachineConfigAnnotation

pkg/controller/statusmanager/machineconfig_watcher.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type MachineConfigPoolWatcher struct {
+	status *StatusManager
+	cache  cache.Cache
+}
+
+// AddMachineConfigWatcher wires up the MachineConfigWatcher and MachineConfigPoolWatcher
+// to the controller-manager.
+func (s *StatusManager) AddMachineConfigWatcher(mgr manager.Manager) error {
```
jcaamano, line 36: nit: AddMachineConfigWatchers

pkg/names/names.go
pperiyasamy: yes, done.

jcaamano:
/lgtm
/approve
/hold

@pperiyasamy feel free to unhold when you are satisfied with the test results

pperiyasamy:
/test e2e-aws-ovn-ipsec-upgrade
/test e2e-aws-ovn-ipsec-serial

some ipsec jobs are failing due to olm deployment failures which happening very recently, not related to PR change. triggering it agin.

pperiyasamy:
The commit [b99b606](https://github.com/openshift/cluster-network-operator/pull/2383/commits/b99b6062f6767af9bb1c2c83672df655b602009a) is updated with following change to fix an issue (Network status is still reflecting machine config pool status even after IPsec is disabled) found from @huiran0826's testing.

```
$ git diff
diff --git a/pkg/controller/statusmanager/machineconfig_status.go b/pkg/controller/statusmanager/machineconfig_status.go
index fa61f2dff..858293c22 100644
--- a/pkg/controller/statusmanager/machineconfig_status.go
+++ b/pkg/controller/statusmanager/machineconfig_status.go
@@ -157,7 +157,16 @@ func (status *StatusManager) SetFromMachineConfigPool(mcPools []mcfgv1.MachineCo
                                        removed = mcutil.AreMachineConfigsRemovedFromPool(pool.Status, mcSet)
                                        if removed {
                                                status.machineConfigsBeingRemoved[role].Delete(machineConfig)
+                                               // Delete map entry from status cache if role doesn't have machine configs. By deleting the entry,
+                                               // there won't be any unnecessary processing of pools in the reconcile loop when it's not dealing
+                                               // with network operator machine configs anymore.
+                                               if status.machineConfigsBeingRemoved[role].Len() == 0 {
+                                                       delete(status.machineConfigsBeingRemoved, role)
+                                               }
                                                status.renderedMachineConfigs[role].Delete(machineConfig)
+                                               if status.renderedMachineConfigs[role].Len() == 0 {
+                                                       delete(status.renderedMachineConfigs, role)
+                                               }
                                                if err := status.setLastRenderedMachineConfigState(status.renderedMachineConfigs); err != nil {
                                                        return fmt.Errorf("failed to update rendered machine config state: %v", err)
                                                }
diff --git a/pkg/controller/statusmanager/status_manager_test.go b/pkg/controller/statusmanager/status_manager_test.go
index 8a1930dcf..60ce0fe84 100644
--- a/pkg/controller/statusmanager/status_manager_test.go
+++ b/pkg/controller/statusmanager/status_manager_test.go
@@ -737,6 +737,39 @@ func TestStatusManagerSetFromIPsecConfigs(t *testing.T) {
        if len(co.Status.Versions) > 0 {
                t.Fatalf("Status.Versions unexpectedly already set: %#v", co.Status.Versions)
        }
+
+       // Set master mcp into degraded state, check network operator status condition is
+       // doesn't go into degraded state.
+       masterIPsecmachineConfigPool.Status = mcfgv1.MachineConfigPoolStatus{Conditions: []mcfgv1.MachineConfigPoolCondition{{Type: mcfgv1.MachineConfigPoolDegraded,
+               Status: v1.ConditionTrue}}}
+       err = status.SetFromMachineConfigPool([]mcfgv1.MachineConfigPool{masterIPsecmachineConfigPool,
+               workerIPsecMachineConfigPool})
+       if err != nil {
+               t.Fatalf("error processing machine config pools: %v", err)
+       }
+       co, oc, err = getStatuses(client, "testing")
+       if err != nil {
+               t.Fatalf("error getting ClusterOperator: %v", err)
+       }
+       if !conditionsInclude(oc.Status.Conditions, []operv1.OperatorCondition{
+               {
+                       Type:   operv1.OperatorStatusTypeDegraded,
+                       Status: operv1.ConditionFalse,
+               },
+       }) {
+               t.Fatalf("unexpected network Status.Conditions: %#v", oc.Status.Conditions)
+       }
+       if !conditionsInclude(oc.Status.Conditions, []operv1.OperatorCondition{
+               {
+                       Type:   operv1.OperatorStatusTypeProgressing,
+                       Status: operv1.ConditionFalse,
+               },
+       }) {
+               t.Fatalf("unexpected Status.Conditions: %#v", oc.Status.Conditions)
+       }

```
Did few tests with changes, The issue is also solved with the fix.

PR #2384, opened 2024-05-23
SDN-4896: Drop openshift-sdn support

danwinship:
It's time...

~I removed live migration and offline migration in two separate pieces because I realized belatedly I had to be somewhat careful about it because we still need _some_ of the migration code in place to support "MTU migration"...~

UPDATED: this now leaves most of the sdn code there, it just removes support for actually _deploying_ sdn (and in particular, removes the references to the SDN image, which is needed so we can start the process of removing the image from the release). We can do followups to clean up the dead code.

### DIFF
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -568,7 +568,7 @@ func validateMultus(conf *operv1.NetworkSpec) []error {
 func validateDefaultNetwork(conf *operv1.NetworkSpec) []error {
 	switch conf.DefaultNetwork.Type {
 	case operv1.NetworkTypeOpenShiftSDN:
-		return validateOpenShiftSDN(conf)
+		return []error{errors.Errorf("unsupported network type %q", conf.DefaultNetwork.Type)}
 	case operv1.NetworkTypeOVNKubernetes:
 		return validateOVNKubernetes(conf)
 	default:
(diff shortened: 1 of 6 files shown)


pkg/network/render.go
```diff
@@ -568,7 +568,7 @@ func validateMultus(conf *operv1.NetworkSpec) []error {
 func validateDefaultNetwork(conf *operv1.NetworkSpec) []error {
 	switch conf.DefaultNetwork.Type {
 	case operv1.NetworkTypeOpenShiftSDN:
-		return validateOpenShiftSDN(conf)
+		return []error{errors.Errorf("unsupported network type %q", conf.DefaultNetwork.Type)}
```
kyrtapz, line 571: End of an era!

kyrtapz:
/lgtm
We are going to need to remove the reference from the hypershift repo:
https://github.com/openshift/hypershift/blob/main/control-plane-operator/controllers/hostedcontrolplane/cno/clusternetworkoperator.go#L534

pliurh:
A process question: If we remove the SDN support from CNO in 4.17. How can we fix and verify the live migration-related code in the master branch, and then backport?

danwinship:
> A process question: If we remove the SDN support from CNO in 4.17. How can we fix and verify the live migration-related code in the master branch, and then backport?

Future migration fixes (or general openshift-sdn fixes) will start from `release-4.16` and be backported from there.

PR #2385, opened 2024-05-24
OCPBUGS-23758: Set global IP forwarding sysctl parameters while starting ovnkube-node

arghosh93:
This PR is to set IPv4 and IPv6 IP forwarding sysctl parameter when ipForwarding mode is set to Restricted.

If ipForwarding is set to 'Restricted':

sysctl -w net.ipv4.ip_forward=0
sysctl -w net.ipv6.conf.all.forwarding=0

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -562,13 +562,19 @@ data:
         dns_name_resolver_enabled_flag="--enable-dns-name-resolver"
       fi
 
-      # If IP Forwarding mode is global set it in the host here.
+      # If IP Forwarding mode is global set it in the host here. IPv6 IP Forwarding shuld be
+      # enabled for all interfaces at all times if cluster is configured as single stack IPv6
+      # or dual stack. This will be taken care by ovn-kubernetes(ovn-org/ovn-kubernetes#4376).
+      # Setting net.ipv6.conf.all.forwarding to '0' when ipForwarding is Restricted to make 
+      # sure IPv6 IP Forwarding is disabled when cluster is configured as single stack IPv4.
       ip_forwarding_flag=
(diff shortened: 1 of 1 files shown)


trozet (approved):
/approve

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -569,6 +569,8 @@ data:
         sysctl -w net.ipv6.conf.all.forwarding=1
       else
         ip_forwarding_flag="--disable-forwarding"
+        sysctl -w net.ipv4.ip_forward=0
+        sysctl -w net.ipv6.conf.all.forwarding=0
```
kyrtapz, line 573: We will need to have the ipv6 enabled even if `IP_FORWARDING_MODE` is not Global.
For more context: https://github.com/ovn-org/ovn-kubernetes/pull/4376
Lets remove that line from here.

arghosh93: Okay, Understood. Thanks @kyrtapz 
IPv6 forwarding is also disabled by default just like IPv4 as per https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt.

I do see that you are enabling IPv6 forwarding in your PR(#4376) if either single stack IPv6 or dual stack is configured.

so, I think its okay to have net.ipv6.conf.all.forwarding set to default(0) while we are not using single stack IPv6 or dual stack. Or, should I set net.ipv6.conf.all.forwarding=1 if ipForwarding mode is 'Restricted' instead of removing line 573 completely.

arghosh93: @kyrtapz , I have removed the line where I was setting net.ipv6.conf.all.forwarding to '0' when ipForwarding mode is set to 'Restricted'. Please take a look. 
Thanks in advance!

PR #2387, opened 2024-05-27
OCPBUGS-34524: Fix DNSResolver feature gate enablement

npinaeva:
Put feature-gate flag to the control-plane pod spec to make sure control-plane pods are restarted on feature gate enablement. ovnkube-node is fine, as this flag is enabled via script-lib.


npinaeva:
should succeed now after https://github.com/openshift/release/pull/52726 is merged

PR #2389, opened 2024-05-29
Add ipsec state metric

JoshSalomon:
This commit adds a metric for the ipsec mode. The metric created is a Gauge named "openshift_network_operator_ipsec_state" with a constant value of 1 and 2 labels:
1. "mode" which lists the ipsec mode, one of "Disabled" "External" or "Full"
2. "is_legacy_api" which states the flavor of the API used - either "true" for legacy API (up to OCP V4.14) or "false" for the new API (OCP 4.15+)
   -- If the cluster uses non ovn-k network (a network that does not support ipsec) the value of this label is "N/A - ipsec not supported (non-OVN network)" The telemetry is collected by a rule
"openshift:openshift_network_operator_ipsec_state:sum" that strips unwanted dimentions from the Gauge

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -517,12 +517,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	return objs, progressing, nil
 }
 
-// getIPsecMode return the ipsec mode accounting for upgrade scenarios
+// GetIPsecMode return the ipsec mode accounting for upgrade scenarios
 // Find the IPsec mode from Ipsec.config
 // Ipsec.config == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
 // ipsecConfig.mode == "" (bw compatibility) || ipsec.Config == Full ==> ipsec is enabled for NS and EW
 // ipsecConfig.mode == External ==> ipsec is enabled for NS only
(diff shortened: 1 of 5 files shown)


tssurya:
raised a question on slack: https://redhat-internal.slack.com/archives/CDCP2LA9L/p1717002867332349?thread_ts=1717001558.745169&cid=CDCP2LA9L on why we are backporting a "new metric" as a "bugfix"

martinkennelly (approved):
/lgtm

pkg/network/ovn_kubernetes.go
```diff
@@ -906,8 +911,8 @@ func getOVNEncapOverhead(conf *operv1.NetworkSpec) uint32 {
 	const geneveOverhead = 100
 	const ipsecOverhead = 46 // Transport mode, AES-GCM
 	var encapOverhead uint32 = geneveOverhead
-	mode := getIPsecMode(conf.DefaultNetwork.OVNKubernetesConfig)
-	if mode == operv1.IPsecModeFull || mode == "" {
+	mode := GetIPsecMode(conf.DefaultNetwork.OVNKubernetesConfig)
```
martinkennelly, line 914: fyi for others: conflict needed to be resolved here when OP cherry-picked.

kyrtapz:
/retitle [release-4.15] OCPBUGS-34596: Add ipsec state metric
/lgtm
/label backport-risk-assessed

PR #2390, opened 2024-05-29
OCPBUGS-34885:[release-4.14] Fix 4.13->4.14 upgrade with ipsec enabled

ricky-rav:
Ipsec is currently not functional all throughout the 4.13->4.14 upgrade, until the new ovnkube-node pods are deployed on each node.  

4.13 deploys `ovn-ipsec` daemonset, while 4.14 deploys `ovn-ipsec-host` and `ovn-ipsec-containerized` daemonsets, which require the certificates issued by ovnkube-node-identity, otherwise they are in crashloopbackoff
until the new ovnkube-node pods are running.

The idea is that, during 4.13->4.14 upgrades,we should run the 4.13 ovn-ipsec daemonset until phase 2 of the upgrade to IC is done, that is until all multizone ovnkube-node pods are deployed, at which point the new 4.14 daemonsets can be created and the 4.13 one is removed.

```                    
        │ 4.13          │   4.14              │  4.14                  │ 4.14                    
────────┼───────────────┼─────────────────────┼────────────────────────┼─────────────────────────
        │               │   phase1            │  phase2                │ upgrade is done                      
        │               │   single-zone folder│  multi-zone-tmp folder │ multi-zone folder       
        │               │                     │                        │                         
 OVN    │ 3 ovnk-master │   3 ovnk-master     │  3 ovnk-master         │ /                       
        │ 3 ovnk-node   │   3 ovnk-node       │  6 ovnk-node           │ 6 ovnk-node             
        │               │                     │  2 ovnk-control-plane  │ 2 ovnk-control-plane    
        │               │                     │                        │                         
 IPSEC  │ ovn-ipsec     │   ovn-ipsec         │  ovn-ipsec             │ ovn-ipsec-host          
                        │                                                ovn-ipsec-containerized    
```                                                            

https://issues.redhat.com/browse/OCPBUGS-33500

Fixes #OCPBUGS-33500

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -557,7 +557,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	// created and the 4.13 one is removed.
 	if ongoingUpgradeToInterconnect {
 		// don't create ovn-ipsec-host until upgrade to IC is done
-		k8s.UpdateObjByGroupKindName(objs, "apps", "DaemonSet", util.OVN_NAMESPACE, "ovn-ipsec-host", func(o *uns.Unstructured) {
+		k8s.UpdateObjByGroupKindName(objs, "apps", "DaemonSet", util.OVN_NAMESPACE, util.OVN_IPSEC_HOST, func(o *uns.Unstructured) {
 			anno := o.GetAnnotations()
 			if anno == nil {
 				anno = map[string]string{}
@@ -566,7 +566,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 			o.SetAnnotations(anno)
 		})
 		// don't create ovn-ipsec-containerized until upgrade to IC is done
-		k8s.UpdateObjByGroupKindName(objs, "apps", "DaemonSet", util.OVN_NAMESPACE, "ovn-ipsec-containerized", func(o *uns.Unstructured) {
+		k8s.UpdateObjByGroupKindName(objs, "apps", "DaemonSet", util.OVN_NAMESPACE, util.OVN_IPSEC_CONTAINERIZED, func(o *uns.Unstructured) {
 			anno := o.GetAnnotations()
 			if anno == nil {
 				anno = map[string]string{}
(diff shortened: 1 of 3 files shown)


ricky-rav:
I've successfully tested my changes with an upgrade from 4.13 to 4.14 on aws to check that the 4.14 ipsec pods are deployed only at the end of the upgrade, as outlined in the schema from my first message.

pperiyasamy:
@ricky-rav does it worth to add a unit test to cover this case ?

ricky-rav:
> @ricky-rav does it worth to add a unit test to cover this case ?

Sure, PTAL. I had to add a few more tweaks for the unit test. :)

pkg/network/ovn_kubernetes.go
```diff
@@ -1624,18 +1624,24 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
...
 
-	ipsecHostDaemonSet := &appsv1.DaemonSet{
-		TypeMeta: metav1.TypeMeta{
-			Kind:       "DaemonSet",
-			APIVersion: appsv1.SchemeGroupVersion.String(),
-		},
-	}
-	nsn = types.NamespacedName{Namespace: util.OVN_NAMESPACE, Name: "ovn-ipsec-host"}
-	if err := kubeClient.ClientFor("").CRClient().Get(context.TODO(), nsn, ipsecHostDaemonSet); err != nil {
+	ipsecHostDaemonSet, err := kubeClient.Default().Kubernetes().AppsV1().DaemonSets(util.OVN_NAMESPACE).Get(context.TODO(), util.OVN_IPSEC_HOST, metav1.GetOptions{})
```
pperiyasamy, line 1627: any specific reason why ovn-ipsec-host DS retrieved twice from API ? we could just try retrieve it, if it exists, set ipsecStatus parameters, otherwise set ipsecStatus to nil.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -2148,6 +2149,220 @@ func TestRenderOVNKubernetesDualStackPrecedenceOverUpgrade(t *testing.T) {
...
+
+	// move to the end of the upgrade: ovnk master is gone, ovnk node, ovnk control plane are up.
+	// The 4.13 ipsec daemonset is gone; the two 4.14 ipsec daemonsets are finally rendered without
+	// the create-wait annotation.
+	fmt.Println("\n============= end of phase 2, will remove ovnkube master")
+	errMsg = "end of phase 2"
+
+	bootstrapResult.OVN.ControlPlaneUpdateStatus.Progressing = false // roll out of control plane is done
+	bootstrapResult.OVN.IPsecUpdateStatus.Name = util.OVN_IPSEC_HOST // ovn-ipsec-host now replaced ovn-ipsec
+	bootstrapResult.OVN.IPsecUpdateStatus.Version = "4.14.0"
```
pperiyasamy, line 2333: one question: i guess so far ovn-ipsec-host is rendered with create-wait-annotation ? so how it suddenly pops up from `IPsecUpdateStatus` ?

pperiyasamy (commented):
Thanks @ricky-rav , some minor nits.

pkg/network/ovn_kubernetes_test.go
ricky-rav: It's related to the answer to your other question, my bad for not making it clear from the start. In the line you highlighted I'm setting the name of the ipsec update status to `ovn-ipsec-host`, while until then it was `ovn-ipsec`. This reflects what the bootstrap code in CNO would store in  `IPsecUpdateStatus`, but it won't make any difference in the running code, since the name of `IPsecUpdateStatus` is never checked. It's just to remind us what changes are supposed to be happening.

pkg/network/ovn_kubernetes.go
pperiyasamy: ok, i understand it now, this sets up `ipsecStatus` appropriately.

PR #2401, opened 2024-06-05
OCPBUGS-33497: Limit iptables-alerter's cpu

danwinship:
It's supposed to be a background process, but it's not acting that way right now

### DIFF
--- a/bindata/network/iptables-alerter/003-daemonset.yaml
+++ b/bindata/network/iptables-alerter/003-daemonset.yaml
@@ -44,6 +44,8 @@ spec:
           requests:
             cpu: 10m
             memory: 65Mi
+          limits:
+            cpu: 10m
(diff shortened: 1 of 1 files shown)


danwinship:
@sdodson so this is failing because apparently we're not allowed to set CPU limits on `openshift-*` pods without an exception. Does this seem like something exception-worthy? I mean, I literally want to just throttle its CPU...

sdodson:
/retest-required
Those all look like build infra problems

danwinship:
/override ci/prow/4.16-upgrade-from-stable-4.15-e2e-gcp-ovn-rt-upgrade

danwinship:
/override ci/prow/4.16-upgrade-from-stable-4.15-e2e-gcp-ovn-rt-upgrade

PR #2404, opened 2024-06-06
NO-JIRA: iptables-alerter streamlining

danwinship:
I'm not really sure why iptables-alerter is using a lot of CPU (or at least, why some people _think_ it's using a lot of CPU), but looking at what it does, the most likely CPU suck is `jq`. (Especially maybe if `crictl` returns A LOT of JSON for certain pods for some reason?)

So this fixes it to use golang template support instead, and also fixes it to make only a single call to `crictl` per pod. Also, while I was tracing the pod to try to figure out what was being slow, I realized there's really no point in checking if we already logged an event before even deciding if we want to log an event. (Running `iptables-save` is more work on the node itself, but running `kubectl` is probably more work for the cluster as a whole.)

### DIFF
--- a/bindata/network/iptables-alerter/002-script.yaml
+++ b/bindata/network/iptables-alerter/002-script.yaml
@@ -47,13 +47,6 @@ data:
             fi
             netns=$(basename "${netns_path}")
 
-            # Check if we already logged an event for it
-            events=$(kubectl get events -n "${pod_namespace}" -l pod-uid="${pod_uid}" 2>/dev/null)
-            if [[ -n "${events}" ]]; then
-                echo "Skipping pod ${pod_namespace}/${pod_name} which we already logged an event for."
-                continue
-            fi
-
             # Set iptables_output to the first iptables rule in the pod's network
             # namespace, if any. (We use `awk` here rather than `grep` intentionally
             # to avoid awkwardness with grep's exit status on no match.)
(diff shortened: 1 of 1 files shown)


danwinship:
/assign @kyrtapz 

from #2451:
> as a future improvement we might consider storing the `crictl inspectp "${id}"` result in a variable to avoid multiple crio calls.

done! (but I felt like the simpler fix in 2451 was more backportable...)

bindata/network/iptables-alerter/002-script.yaml
```diff
@@ -37,36 +24,31 @@ data:
...
+                     * as a go template (eg, to fill in {{.ReleaseVersion}} above), but
+                     * it also needs to contain a go template string (to pass to crictl
+                     * below). We do this by putting the crictl argument as a string
+                     * literal inside an argument of the outer template:
+                     *     ... --template {{`'BLAH BLAH BLAH'`}} ...
+                     * becomes
+                     *     ... --template 'BLAH BLAH BLAH' ...
+                     * in the actual ConfigMap.
+                     */
+                -}}
```
kyrtapz, line 39: The template doesn't work:
```
template: ../../bindata/network/iptables-alerter/002-script.yaml:29: comment ends before closing delimiter"
```
This should work:
```suggestion
                {{- /* 
                     * bindata/network/iptables-alerter/002-script.yaml is processed
                     * as a go template (eg, to fill in {{.ReleaseVersion}} above), but
                     * it also needs to contain a go template string (to pass to crictl
                     * below). We do this by putting the crictl argument as a string
                     * literal inside an argument of the outer template:
                     *     ... --template {{`'BLAH BLAH BLAH'`}} ...
                     * becomes
                     *     ... --template 'BLAH BLAH BLAH' ...
                     * in the actual ConfigMap.
                */ -}}
```

kyrtapz (commented):
Looks good, there is just one issue with the template.

PR #2408, opened 2024-06-11
SDN-5029: linter: removed deprecated linters

maiqueb:
This commit removes the linters deprecated from 0.53.0 to 0.59.0. 

It is required for a follow-up PR which will update CNO to use golang 1.22 (which requires golangci-lint >= 0.56.0).

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -6,7 +6,6 @@ run:
 linters:
   disable-all: true #so that lint version bumps don't change results
   enable:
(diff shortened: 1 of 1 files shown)


jcaamano:

Not possibly affected as linting does not run in that job.

kyrtapz:
/override ci/prow/4.16-upgrade-from-stable-4.15-e2e-gcp-ovn-rt-upgrade

PR #2410, opened 2024-06-12
OCPBUGS-35316: Create the configmap mtu if not found

pliurh:
The configmap mtu is required by SDN live migration. It should be created by the mtu-prober job at the installation time since 4.11. But if the cluster was upgrade from a very early releases, it may be absent.

### DIFF
--- a/pkg/controller/operconfig/mtu_probe.go
+++ b/pkg/controller/operconfig/mtu_probe.go
@@ -58,6 +58,12 @@ func (r *ReconcileOperConfig) probeMTU(ctx context.Context, oc *operv1.Network, 
 	if err != nil {
 		return 0, fmt.Errorf("failed to deploy mtu prober: %w", err)
 	}
+	r.mtuProberCleanedUp = false
+	defer func() {
+		if err := r.deleteMTUProber(ctx, infra); err != nil {
+			klog.Errorf("Failed to clean up mtu prober: %v", err)
+		}
+	}()
 
 	klog.Info("MTU prober deployed, waiting for result ConfigMap")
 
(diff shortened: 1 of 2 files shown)


kyrtapz:
The configmap is used to communicate the mtu probe job result to CNO and the value is set in the network.operator CR.
I think that the migration should not rely on the configmap but read the value from  `cluster/network.operator.openshift.io`.

pliurh:
@kyrtapz We need the host MTU at https://github.com/openshift/cluster-network-operator/blob/69bdb71fe4928bf203e17a17aa1586f8e246e1df/pkg/controller/clusterconfig/migration.go#L161

Assuming the host MTU is cluster MTU + 50 for an SDN cluster is not always a valid guess. Generating the configMap can also help align the available info among clusters installed from different releases. It can also be helpful for other scenarios where the host MTU is needed.

kyrtapz:
> @kyrtapz We need the host MTU at
> 
> https://github.com/openshift/cluster-network-operator/blob/69bdb71fe4928bf203e17a17aa1586f8e246e1df/pkg/controller/clusterconfig/migration.go#L161
> 
> Assuming the host MTU is cluster MTU + 50 for an SDN cluster is not always a valid guess. Generating the configMap can also help align the available info among clusters installed from different releases. It can also be helpful for other scenarios where the host MTU is needed.

I think this change won't work with clusters without any worker nodes (possible in hypershift).

pliurh:
I think this change won't work with clusters without any worker nodes (possible in hypershift).
> This PR is not to address the scenario where the configmap is not created for clusters without worker nodes. This patch is to create the configmap where it should have been created at installation but not.

Additionally, can you remind me why do we need the machine MTU set for live migration?
> IIUC, the machine MTU is a must-have for changing the cluster routable MTU.

kyrtapz:
/lgtm

We have discussed an alternative approach of calling the `probeMTU` in `calculateRoutableMTU`, opted against it as it would add a potential extra step during the migration and would require a minor refactor.

PR #2421, opened 2024-06-25
SDN-5070: Allow day 2 customization of masquerade subnet

arghosh93:
This PR is to rollout ovnkube-node when there is a change in either ipv4 or ipv6 internalMasqueradeSubnet and done after cluster installation.

This PR also stops updating ovnkube-config configmap with ipv4 or ipv6 internalMasqueradeSubnet and instead updates ovnkube-script-lib configmap in openshift-ovn-kubernetes namespace so that CNO inbuilt logic can trigger ovnkube-node rollout.

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/managed/004-config.yaml
@@ -56,12 +56,6 @@ data:
     [gateway]
     mode={{.OVN_GATEWAY_MODE}}
     nodeport=true
-{{- if (index . "V4InternalMasqueradeSubnet")}}
-    v4-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
-{{- end }}
-{{- if (index . "V6InternalMasqueradeSubnet")}}
-    v6-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
-{{- end }}
 {{- if .OVNHybridOverlayEnable }}
 
     [hybridoverlay]
@@ -147,12 +141,6 @@ data:
     [gateway]
     mode={{.OVN_GATEWAY_MODE}}
     nodeport=true
-{{- if (index . "V4InternalMasqueradeSubnet")}}
-    v4-masquerade-subnet="{{.V4InternalMasqueradeSubnet}}"
-{{- end }}
-{{- if (index . "V6InternalMasqueradeSubnet")}}
-    v6-masquerade-subnet="{{.V6InternalMasqueradeSubnet}}"
-{{- end }}
 
 
     [logging]
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -132,8 +132,8 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["V6JoinSubnet"] = ""
 	data.Data["V4TransitSwitchSubnet"] = ""
 	data.Data["V6TransitSwitchSubnet"] = ""
-	data.Data["V4InternalMasqueradeSubnet"] = ""
-	data.Data["V6InternalMasqueradeSubnet"] = ""
+	data.Data["V4MasqueradeSubnet"] = ""
+	data.Data["V6MasqueradeSubnet"] = ""
 
 	data.Data["V4JoinSubnet"] = c.V4InternalSubnet
 	data.Data["V6JoinSubnet"] = c.V6InternalSubnet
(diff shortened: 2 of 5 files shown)


kyrtapz:
We are going to have to update the API:
https://github.com/openshift/api/blob/c299a673ec729b7e63a6c363d67e961454218e09/operator/v1/types_network.go#L576
and docs:
https://github.com/openshift/openshift-docs/blob/05e9773d92c7123590929b19224a62f83e565ee0/networking/cidr-range-definitions.adoc#L25

pkg/network/ovn_kubernetes.go
```diff
@@ -833,6 +849,22 @@ func bootstrapOVNConfig(conf *operv1.Network, kubeClient cnoclient.Client, hc *h
...
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("Could not determine stale masquerade subnets: %w", err)
+		}
+	} else {
+		spec := &operv1.NetworkSpec{}
+		err = json.Unmarshal([]byte(cm.Data["applied"]), spec)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to unmarshal data from configmap %s in %s namespace: %w", names.APPLIED_PREFIX+names.OPERATOR_CONFIG, names.APPLIED_NAMESPACE, err)
+		}
+		ovnConfigResult.V4OldMasqueradeSubnet = spec.DefaultNetwork.OVNKubernetesConfig.GatewayConfig.IPv4.InternalMasqueradeSubnet
```
kyrtapz, line 864: Have you tested that this change works? I am concerned that CNO will cleanup the `V[4/6]OldMasqueradeSubnet` on the next reconciliation and that might happen before ovn-k has a chance to cleanup the stale entries.

pkg/network/ovn_kubernetes.go
```diff
@@ -132,6 +132,10 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["V6JoinSubnet"] = ""
 	data.Data["V4TransitSwitchSubnet"] = ""
 	data.Data["V6TransitSwitchSubnet"] = ""
+	data.Data["V4InternalMasqueradeSubnet"] = ""
```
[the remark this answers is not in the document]
arghosh93: Ack

pkg/network/ovn_kubernetes.go
arghosh93: yaa, as @trozet mentioned at https://github.com/ovn-org/ovn-kubernetes/pull/4496#issuecomment-2251451338 , we need to document that user should refrain from changing masquerade subnet again until all ovnkube-node PODs have rolled out upon changing once. User should also refrain from changing masquerade subnet before completion of cluster upgrade.

If user is not changing masquerade subnet before rollout of all ovnkube-node PODs then there should not be a problem as first bootstrap happens then we run Render and then we apply new config to the applied-cluster config map.
https://github.com/openshift/cluster-network-operator/blob/master/pkg/controller/operconfig/operconfig_controller.go#L368-L417

arghosh93: Hmm, I understand your concern. very nice catch :+1: . I think we need to annotate the daemonset with stale masquerade subnet only with valid IP in that case. wdyt?

arghosh93: @kyrtapz I am planning following to tackle this problem:

- Instead of updating 008-script-lib.yaml with new and stale masquerade subnet, I am going to update 004-config.yaml. In that case ovnkube-node daemonset is not going to be updated directly when masquerade subnet is changed at day 2. Instead ovnkube-config configmap will be updated.
- Since changes to ovnkube-config CM does not rollout ovnkube daemonset, I am going to add an annotation to the daemonset with a bool value to rollout ovnkube-node only when V[4][6]StaleMasqueradeSubnet is not empty.

If consecutive reconciliation happens then it is not going to update the daemonset and would stop a new rollout. In documentation, we would also need to mention that v4/v6 masquerade subnet should not be changed again until last rollout of ovnkube-node completes. 

Do you have any other suggestions? Thanks in advance!

arghosh93: Closing this conversation as we are not anymore passing on old masquerade subnet to ovnkube.

bindata/network/ovn-kubernetes/managed/004-config.yaml
```diff
@@ -147,12 +141,6 @@ data:
     [gateway]
     mode={{.OVN_GATEWAY_MODE}}
     nodeport=true
-{{- if (index . "V4InternalMasqueradeSubnet")}}
```
kyrtapz, line 150: Prior to this change `V[4|6]InternalMasqueradeSubnet` was set in the configmap so it was read by both ovnkube-node daemonset and the cluster-manager deployment. With your PR only ovnkube-node gets the argument.
I think it would be good to add it to control-plane too, for consistency, even if it is not currently used.

transit cidr example:
https://github.com/openshift/cluster-network-operator/blob/0be5bd93a6cba203b77313b3444022330a61b58a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L118-L125
https://github.com/openshift/cluster-network-operator/blob/4776726dee1d22b6b06bbaa7024fc0d3776b6f97/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml#L158-L165

arghosh93: @kyrtapz  I omitted it as we dont need masquerade subnet while running cluster-manager. Similarly we dont provide transit switch subnet details to ovnkube-controller. 
However if its about consistency then I agree that adding masquerade subnet to ovnkube-control-plane.yaml makes sense.

kyrtapz: Thanks! so currently:
- JoinSubnet - passed to both
- TransitSwitchSubnet - only passed to cluster-manager
- MasqueradeSubnet - only passed to ovnkube-node

Prior to your change MasqueradeSubnet was configured through the configmap for both, I am ok to keep it as is for now if we were not using the  MasqueradeSubnet in cluster-manager anyway.

PR #2433, opened 2024-07-09
SDN-4934: Enable TechPreview FeatureGate for NetworkSegmentation

tssurya:
Enables TechPreview feature gate for NetworkSegmentation
Builds on top of https://github.com/openshift/cluster-network-operator/pull/2428
We will GA in the last sprint but we don't have to do anything given
we just need to bump it to default ON in API repo.

The new network CRD can also use this gate to be installed if the flag is ON

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -547,6 +547,12 @@ data:
         multi_network_enabled_flag="--enable-multi-network"
       fi
 
+      network_segmentation_enabled_flag=
+      if [[ "{{.OVN_NETWORK_SEGMENTATION_ENABLE}}" == "true" ]]; then
+        multi_network_enabled_flag="--enable-multi-network"
+        network_segmentation_enabled_flag="--enable-network-segmentation"
+      fi
+
       multi_network_policy_enabled_flag=
       if [[ "{{.OVN_MULTI_NETWORK_POLICY_ENABLE}}" == "true" ]]; then
         multi_network_policy_enabled_flag="--enable-multi-networkpolicy"
@@ -615,6 +621,7 @@ data:
         --disable-snat-multiple-gws \
         ${export_network_flows_flags} \
         ${multi_network_enabled_flag} \
+        ${network_segmentation_enabled_flag} \
         ${multi_network_policy_enabled_flag} \
         ${admin_network_policy_enabled_flag} \
         ${dns_name_resolver_enabled_flag} \
--- a/bindata/network/ovn-kubernetes/managed/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/managed/004-config.yaml
@@ -42,6 +42,12 @@ data:
 {{- if .OVN_MULTI_NETWORK_ENABLE }}
     enable-multi-network=true
 {{- end }}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
+    {{- if not .OVN_MULTI_NETWORK_ENABLE }}
+    enable-multi-network=true
+    {{- end }}
+    enable-network-segmentation=true
+{{- end }}
 {{- if .OVN_MULTI_NETWORK_POLICY_ENABLE }}
     enable-multi-networkpolicy=true
 {{- end }}
@@ -132,6 +138,12 @@ data:
 {{- end }}
 {{- if .OVN_MULTI_NETWORK_ENABLE }}
     enable-multi-network=true
+{{- end }}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
+    {{- if not .OVN_MULTI_NETWORK_ENABLE }}
+    enable-multi-network=true
+    {{- end }}
+    enable-network-segmentation=true
 {{- end }}
 {{- if .OVN_MULTI_NETWORK_POLICY_ENABLE }}
     enable-multi-networkpolicy=true
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
@@ -174,6 +174,15 @@ spec:
             persistent_ips_enabled_flag="--enable-persistent-ips"
           fi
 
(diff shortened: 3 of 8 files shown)


bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -547,6 +547,11 @@ data:
         multi_network_enabled_flag="--enable-multi-network"
       fi
 
+      network_segmentation_enabled_flag=
+      if [[ "{{.OVN_NETWORK_SEGMENTATION_ENABLE}}" == "true" ]]; then
+        network_segmentation_enabled_flag="--enable-multi-network --enable-network-segmentation"
```
kyrtapz, line 552: NIT: Can we do this instead to avoid potentially providing `--enable-multi-network` twice?
```suggestion
        multi_network_enabled_flag="--enable-multi-network"
        network_segmentation_enabled_flag="--enable-network-segmentation"
```

bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
```diff
@@ -48,6 +48,12 @@ data:
 {{- if .OVN_MULTI_NETWORK_ENABLE }}
     enable-multi-network=true
 {{- end }}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
```
kyrtapz, line 51: When OVN_NETWORK_SEGMENTATION_ENABLE becomes true control-plane won't restart, is this expected?

bindata/network/ovn-kubernetes/managed/004-config.yaml
```diff
@@ -133,6 +139,12 @@ data:
 {{- if .OVN_MULTI_NETWORK_ENABLE }}
     enable-multi-network=true
 {{- end }}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
+    {{- if not .OVN_MULTI_NETWORK_ENABLE }}
+        enable-multi-network=true
```
kyrtapz, line 144: NIT: this indentation is not correct, this is at the same level as other fields below and above

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
tssurya: yeah nice suggestion

bindata/network/ovn-kubernetes/managed/004-config.yaml
tssurya: hmmm I don't see it.. I must be blind...what I have is:
```
{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
    {{- if not .OVN_MULTI_NETWORK_ENABLE }}
        enable-multi-network=true
    {{- end }}
    enable-network-segmentation=true
{{- end }}
```
enable-multi-network needs to be same level as which line?

bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
tssurya: do you mean if someone disables the feature? why so?

bindata/network/ovn-kubernetes/managed/004-config.yaml
kyrtapz: The indentation of the rendered value is not at the same level as others, this would make it the same:
```
{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
    {{- if not .OVN_MULTI_NETWORK_ENABLE }}
    enable-multi-network=true
    {{- end }}
    enable-network-segmentation=true
{{- end }}
```
I know that your approach looks better but in the rendered file we would end up with:
```
    [ovnkubernetesfeature]
    enable-egress-ip=true
    enable-egress-firewall=true
    enable-egress-qos=true
    enable-egress-service=true
    egressip-node-healthcheck-port=666
        enable-multi-network=true
    enable-network-segmentation=true
```

tssurya: ahhhhh I see :) gotcha

bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
tssurya: pat explained offline this is for users going from GA to TP (can't think why someone would support this as a day2 thing but here we are); so this needs to also be done on control plane config-maps and I have fixed this.

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -174,6 +174,14 @@ spec:
             persistent_ips_enabled_flag="--enable-persistent-ips"
           fi
 
+          # This is needed so that converting clusters from GA to TP
+          # will rollout control plane pods as well
+          network_segmentation_enabled_flag=
+          if [[ "{{.OVN_NETWORK_SEGMENTATION_ENABLE}}" == "true" ]]; then
+            multi_network_enabled_flag="--enable-multi-network"
```
kyrtapz, line 181: `multi_network_enabled_flag` is not set when `OVN_NETWORK_SEGMENTATION_ENABLE` is not true

tssurya: ofc it isn't (Facepalm), some features got the free pass to be GA without feature gates huh.
nice catch pat!

kyrtapz:
/lgtm

```
+ network_segmentation_enabled_flag=
+ [[ true == \t\r\u\e ]]
+ multi_network_enabled_flag=--enable-multi-network
+ network_segmentation_enabled_flag=--enable-network-segmentation
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2433/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn-techpreview/1813125539452424192/artifacts/e2e-gcp-ovn-techpreview/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovnkube-node-9dg7j_ovnkube-controller.log

PR #2437, opened 2024-07-15
DNM - testing things with CloudConfig

EmilienM:
The kube-cloud-config ConfigMap should be retrieved from the HCP
namespace when Hypershift deployed the CNO.


EmilienM:
/hold
This PR is wrong. But our problem is that the ConfigMap data is empty, we're investigating.

PR #2438, opened 2024-07-16
OCPBUGS-36688: - WIP - looking for ipsec in all worker mcps

yuvalk:
we need to make sure all mcps are aligned with the ipsec on host before
we can remove the daemonset

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -1488,14 +1488,24 @@ func isIPsecMachineConfigActive(infra bootstrap.InfraStatus) bool {
 		return status.MachineCount == status.UpdatedMachineCount &&
 			hasSourceInMachineConfigStatus(status, machineConfigs)
 	}
+
 	for _, masterMCPStatus := range infra.MasterMCPStatuses {
 		if !ipSecPluginOnPool(masterMCPStatus, infra.MasterIPsecMachineConfigs) {
 			return false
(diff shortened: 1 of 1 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1488,14 +1488,24 @@ func isIPsecMachineConfigActive(infra bootstrap.InfraStatus) bool {
...
+	for _, mcp := range mcpList.Items {
+		if mcp.Name != "master" {
+			for _, source := range mcp.Status.Configuration.Source {
+				isContainIpsec := false
+				if strings.Contains(source.Name, "ipsec") {
+					isContainIpsec = true
+				}
+				if !isContainIpsec {
+					return false
+				}
```
ricky-rav, line 1507: I'm trying to familiarize myself with the surrounding code, but here we can simplify it with just:
```
				if !strings.Contains(source.Name, "ipsec") {
					return false
				}
```

yuvalk: I agree, but this PR is just a manual backport of https://github.com/openshift/cluster-network-operator/pull/2349,

IMHO all such changes (as you propose) should go through upstream (master) first..

ricky-rav: All right, no problem then :)

yuvalk:
/close
wrong branch!

PR #2441, opened 2024-07-18
CNV-43973: HCP custom tolerations integration 

davidvossel:
The HostedCluster object for hypershift recently got a [new feature](https://github.com/openshift/hypershift/pull/4339) that allows users creating HCP clusters to set custom tolerations on their HCP pods. The CNO needs to integrate this feature in the same way it integrated the NodeSelector feature for HCP.

### DIFF
--- a/pkg/hypershift/hypershift.go
+++ b/pkg/hypershift/hypershift.go
@@ -10,6 +10,8 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	operv1 "github.com/openshift/api/operator/v1"
+	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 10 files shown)


pkg/hypershift/hypershift.go
```diff
@@ -252,3 +309,25 @@ func SetHostedControlPlaneConditions(hcp *unstructured.Unstructured, operStatus
 	hcp.Object["status"].(map[string]interface{})["conditions"] = conditions
 	return conditions, nil
 }
+
+func TolerationsToStringArray(tolerations []corev1.Toleration) ([]string, error) {
```
kyrtapz, line 313: Missing doc block

pkg/network/ovn_kubernetes.go
```diff
@@ -703,6 +708,13 @@ func bootstrapOVNHyperShiftConfig(hc *hypershift.HyperShiftConfig, kubeClient cn
 
 	ovnHypershiftResult.ClusterID = hcp.ClusterID
 	ovnHypershiftResult.HCPNodeSelector = hcp.NodeSelector
+
+	tolerations, err := hypershift.TolerationsToStringArray(hcp.Tolerations)
```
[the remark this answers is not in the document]
davidvossel: Good suggestion. i stored the string array and got rid of the `!= 0` checks.

pkg/hypershift/hypershift.go
davidvossel: fixed

PR #2442, opened 2024-07-19
SDN-5085: Add support for route advertisements

jcaamano:


pkg/network/render.go
```diff
@@ -654,6 +654,9 @@ func renderCRDForMigration(conf *operv1.NetworkSpec, manifestDir string, feature
 		data.Data["OVN_ADMIN_NETWORK_POLICY_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateAdminNetworkPolicy)
 		data.Data["OVN_NETWORK_SEGMENTATION_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateNetworkSegmentation)
 		data.Data["OVN_OBSERVABILITY_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateOVNObservability)
+		data.Data["OVN_ROUTE_ADVERTISEMENTS_ENABLE"] = conf.DefaultNetwork.OVNKubernetesConfig != nil &&
```
arghosh93, line 657: is it required if we are not supporting route advertisements in OpenshiftSDN ?

bindata/network/ovn-kubernetes/common/001-crd.yaml
```diff
@@ -3583,3 +3583,301 @@ spec:
...
+  name: routeadvertisements.k8s.ovn.org
+spec:
+  group: k8s.ovn.org
+  names:
+    kind: RouteAdvertisements
+    listKind: RouteAdvertisementsList
+    plural: routeadvertisements
+    shortNames:
+    - ra
+    singular: routeadvertisements
```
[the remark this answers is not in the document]
jcaamano: The `kind` is already plural so it makes no sense to make the singular any different. However. let's discuss this first on the OVN-K side
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4472

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3780,3 +3782,88 @@ func networkOwnerRef() []metav1.OwnerReference {
...
+		{
+			name: "default",
+			args: args{
+				conf:            fakeNetworkConf,
+				bootstrapResult: fakeBootstrapResultOVN,
+				manifestDir:     manifestDirOvn,
+				client:          cnofake.NewFakeClient(),
+				featureGates:    noFeatureGates,
+			},
+			expectNumObjs: 37,
```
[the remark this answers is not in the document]
jcaamano: I guess just counting. However the purpose of this specific test is to establish a baseline to which the other feature specific tests can relatively compare to. 

Yes, it is impacted if we render more objects by default but it should be just fairly easy to add or remove according to changes.

Better than nothing.

pkg/network/render.go
jcaamano: OpenshiftSDN is deprecated, what we are doing here has nothing to do with that. This is just an optional feature. Why would you ask?

bindata/network/ovn-kubernetes/common/001-crd.yaml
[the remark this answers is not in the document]
arghosh93: yes, lets discuss on the ovn-kubernetes PR. Will add a similar comment there to start the conversation.

pkg/network/render.go
arghosh93: Okay, Now I understand he purpose. This comment can be ignored.

pkg/network/ovn_kubernetes.go
```diff
@@ -327,6 +327,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["DNS_NAME_RESOLVER_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateDNSNameResolver)
 	data.Data["OVN_NETWORK_SEGMENTATION_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateNetworkSegmentation)
 	data.Data["OVN_OBSERVABILITY_ENABLE"] = featureGates.Enabled(apifeatures.FeatureGateOVNObservability)
+	data.Data["OVN_ROUTE_ADVERTISEMENTS_ENABLE"] = c.RouteAdvertisements == operv1.RouteAdvertisementsEnabled
```
pliurh, line 330: Shall we also check the featureGate here?

jcaamano: I don't see the need. The API is hidden if the feature gate is not set.

jcaamano:
e2e-gcp-ovn-upgrade is failing because of infra issues

trozet (approved):
/lgtm

pkg/network/ovn_kubernetes.go
trozet: hmm then how come we check other features by feature gate here and not just assume the API is not available?

jcaamano: because they are using the feature gate to enable or disable the feature itself

for routeadvertisements we have a dedicated flag

bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
```diff
@@ -212,6 +212,41 @@ rules:
...
+  - create
+  - delete
+  - update
+  - patch
+- apiGroups:
+  - k8s.cni.cncf.io
+  resources:
+  - network-attachment-definitions
+  verbs: 
+  - patch
```
pliurh, line 248: Shall we also give `create` permission for creating the NAD for the default network?

jcaamano: Yeah, let me try to express that better

pkg/network/ovn_kubernetes.go
tssurya: for my own understanding is the RA API installation totally controlled by this flag alone and how does this tie in with the actual feature gate for this feature? meaning did this PR enable RA feature by default in OCP? Or is this dedicated flag still behind the main `RouteAdvertisements` FG in the api level?

PR #2444, opened 2024-07-21
[release-4.14] OCPBUGS-37468: Backport ipsec state metric

JoshSalomon:
This is a backport of PR #[2270](https://github.com/openshift/cluster-network-operator/pull/2270) which added ipsec state metric to OCP V4.16, and PR #[2389](https://github.com/openshift/cluster-network-operator/pull/2389) which backported this to V4.15. Both these PRs are based on API change that was added to V4.15 - therefore this PR is a manual backport and not a cherry-pick since the API does not exist in V4.14 and will not be ported to this version due to compatibiilty reasons.

### DIFF
--- a/pkg/util/ipsec/metrics.go
+++ b/pkg/util/ipsec/metrics.go
@@ -1,0 +1,50 @@
+package ipsec
+
+import (
+	"strconv"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+	"k8s.io/klog/v2"
+)
+
+const (
+	cnoNamespace  = "openshift_network_operator"
+	ipsecStateNA  = "N/A - ipsec not supported (non-OVN network)"
+	ipsecDisabled = "Disabled"
+	ipsecInternal = "Internal"
+)
+
+var (
+	ipsecStateGauge *metrics.GaugeVec
+	trueStr         string
+)
+
+func init() {
+	ipsecStateGauge = metrics.NewGaugeVec(&metrics.GaugeOpts{
+		Namespace: cnoNamespace,
+		Name:      "ipsec_state",
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster, " +
+			"and the API that invoked it, legacy (pre OCP 4.14) or new. " +
+			"In case the network doesn't support ipsec (non-OVN network), " +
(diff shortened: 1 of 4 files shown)


pkg/util/ipsec/metrics.go
```diff
@@ -0,0 +1,51 @@
...
+		Help: "A metric with a constant '1' value labeled by the latest ipsecMode of the cluster, " +
+			"and the API that invoked it, legacy (pre OCP 4.14) or new. " +
+			"In case the network doesn't support ipsec (non-OVN network), " +
+			"the 'is_legacy_api' value is set to '" + ipsecStateNA + "'.",
+	}, []string{"mode", "is_legacy_api"})
+	legacyregistry.MustRegister(ipsecStateGauge)
+	trueStr = strconv.FormatBool(true)
+}
+
+// TODO: Check for the external case (limited to one customer only)???
```
kyrtapz, line 36: Is this TODO still valid? If so can you create a jira card to track the followup work and add it in this comment?

kyrtapz (commented):
Overall looks good.
IIUC on 4.14->4.15 upgrade of an ipsec enabled cluster the mode will change from `Internal` to `Full`, is that correct?

kyrtapz:
Tide got confused with https://prow.ci.openshift.org/github-link?dest=openshift/cluster-network-operator/pull/2443 and is now reporting failures on jobs that do not exist in 4.14. You might try to re-open the PR to clean it up.

JoshSalomon:
> Overall looks good. IIUC on 4.14->4.15 upgrade of an ipsec enabled cluster the mode will change from `Internal` to `Full`, is that correct?

Partially - the reporting of the mode in this metric for the telemetry will change from 'Internal' to 'Full'. From API perspective there is no named mode in 4.14 just a boolean (whether the `OVNKubernetesConfig.IPsecConfig` exists (non nil)) so it is just a reporting issue. I made this different in the metric since in the telemetry it may be important to understand whether the user can have a NS ipsec or alternatively to easily assess ipsec users in 4.14 vs 4.15+

pkg/util/ipsec/metrics.go
JoshSalomon: Good catch - I will remove it - no reason to add telemetry for a single customer.

kyrtapz:
/retitle [release-4.14] OCPBUGS-37468: Backport ipsec state metric

PR #2448, opened 2024-07-23
SDN-4919: OVN-Kubernetes UDN: EndpointSlice mirror controller RBAC permissions

kyrtapz:
Add RBAC permissions for the new EndpointSlice mirror controller. 
The services/finalizer update is required because the controller copies the ownerReference that might block serivce deletion.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
@@ -156,6 +156,24 @@ rules:
     - ipamclaims/status
   verbs: [ "patch", "update" ]
 {{- end}}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
+- apiGroups:
+    - discovery.k8s.io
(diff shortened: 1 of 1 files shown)


bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
```diff
@@ -156,6 +156,24 @@ rules:
     - ipamclaims/status
   verbs: [ "patch", "update" ]
 {{- end}}
+{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}
+- apiGroups:
+    - discovery.k8s.io
+  resources:
+    - endpointslices
```
tssurya, line 163: this would also give us power to change eps created by primary k8s controller a well, we need to be careful but we are consciously doing this

tssurya (approved):
/lgtm

tssurya:
/retitle SDN-4919: OVN-Kubernetes UDN: EndpointSlice mirror controller RBAC permissions

PR #2451, opened 2024-07-30
OCPBUGS-37713: Handle random crictl errors in iptables-alerter

danwinship:
Apparently under heavy node load, crictl may (sporadically) return errors like "timed out". In the linked bug, it is apparently failing on the "newtork namespace type" check and getting back `""`, which is not `"NODE"`, so it keeps going, and then apparently succeeds on the following crictl calls, and ends up doing an iptables rule check in the host network namespace and logging messages because it seems the host-network ovn-kubernetes iptables rules there.


kyrtapz:
/lgtm

as a future improvement we might consider storing the`crictl inspectp "${id}"` result in a variable to avoid multiple crio calls.

PR #2449, opened 2024-07-26
Move to use newer IPsec DaemonSets when MCP is in paused state

pperiyasamy:
When MCP is in paused state, network operator continues to render older IPsec daemonsets which blocks network cluster operator not getting upgraded to 4.15. Hence this commit renders both newer 4.15 IPsec daemonsets for intermediate period. When MCPs are moved to unpaused state and IPsec machine configs are installed on it, then it goes ahead with rendering only host flavored IPsec daemonset.

PR #2454, opened 2024-08-01
OCPBUGS-52280: Move to use newer IPsec DaemonSets irrespective of MCP state

pperiyasamy:
When a machine config pool is in a paused state, then it doesn't process any machine config. so during legacy IPsec upgrade (`4.14->4.15`), IPsec machine configs may not be installed on the nodes when its pool is in paused state.
In those cases the network operator continues to render older IPsec daemonsets which blocks network components from not getting upgraded to newer versions.

Hence this PR renders newer IPsec daemonsets immediately, with new `IPsecCheckForLibreswan` check ensures one of the pods serves IPsec for the node. When MCPs are fully rolled out with ipsec machine config, then it goes ahead with rendering only host flavored IPsec daemonset.

It brings in new behavior on IPsec daemonset rendering during IPsec deployment, upgrade and node reboot scenarios.

1. Users would notice both daemonsets being rendered at the time of IPsec install (or) upgrade for a temporary period until IPsec machine configs are fully deployed.
2. At the time of node reboot or machine config pool goes into progressing state, both daemonsets being rendered. In this scenario, the containerized ipsec daemonset pods are dormant.
3. It removes legacy upgrade case as every upgrade would be considered as the same with this approach.
4. It now mounts top level system directories `/usr/sbin` and `/usr/libexec` instead of specific ipsec host paths. The ipsec paths are available only when libreswan is installed on the node (as mentioned in step 1). 
5. For the upgrades `4.14->4.15` which moves IPsec from container to host deployment, so adding mco version to be at least 4.15 to start rendering IPsec machine configs.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
@@ -49,6 +49,14 @@ spec:
         - |
           #!/bin/bash
           set -exuo pipefail
+
+{{ if .IPsecCheckForLibreswan }}
+          if rpm --dbpath=/usr/share/rpm -q libreswan; then
+            echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container doesnt need to init anything"
+            exit 0
+          fi
+{{ end }}
+
 {{ if .NETWORK_NODE_IDENTITY_ENABLE }}
           # When NETWORK_NODE_IDENTITY_ENABLE is true, use the per-node certificate to create a kubeconfig
           # that will be used to talk to the API
@@ -189,6 +197,9 @@ spec:
           name: signer-ca
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
+        - mountPath: /usr/share/rpm
+          name: host-usr-share-rpm
+          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -219,6 +230,13 @@ spec:
             exit 0
           }
           trap cleanup SIGTERM
+
+{{ if .IPsecCheckForLibreswan }}
+          if rpm --dbpath=/usr/share/rpm -q libreswan; then
+            echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container will sleep to infinity"
+            sleep infinity
+          fi
+{{ end }}
 
           # Don't start IPsec until ovnkube-node has finished setting up the node
           counter=0
@@ -276,6 +294,9 @@ spec:
           name: host-var-log-ovs
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
+        - mountPath: /usr/share/rpm
+          name: host-usr-share-rpm
+          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -288,6 +309,12 @@ spec:
             - -c
             - |
               #!/bin/bash
+{{ if .IPsecCheckForLibreswan }}
+              if rpm --dbpath=/usr/share/rpm -q libreswan; then
+                echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container is always \"alive\""
+                exit 0
+              fi
+{{ end }}
               if [[ $(ipsec whack --trafficstatus | wc -l) -eq 0 ]]; then
                 echo "no ipsec traffic configured"
                 exit 10
@@ -321,6 +348,10 @@ spec:
       - name: host-cni-netd
         hostPath:
           path: "{{.CNIConfDir}}"
+      - name: host-usr-share-rpm
+        hostPath:
+          path: /usr/share/rpm
+          type: DirectoryOrCreate
       tolerations:
       - operator: "Exists"
 {{end}}
--- a/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
@@ -37,6 +37,9 @@ spec:
               - key: network.operator.openshift.io/dpu-host
                 operator: DoesNotExist
       serviceAccountName: ovn-kubernetes-node
+{{ if .IPsecServiceCheckOnHost }}
+      hostPID: true
+{{ end }}
       hostNetwork: true
       dnsPolicy: Default
       priorityClassName: "system-node-critical"
@@ -50,9 +53,9 @@ spec:
           #!/bin/bash
           set -exuo pipefail
 
-{{ if .IPsecCheckForLibreswan }}
-          if rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container doesnt need to init anything"
+{{ if .IPsecServiceCheckOnHost }}
+          if chroot /proc/1/root systemctl is-active --quiet ipsec.service; then
+            echo "host has ipsec.service running and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container doesnt need to init anything"
             exit 0
           fi
 {{ end }}
@@ -197,9 +200,6 @@ spec:
           name: signer-ca
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -231,9 +231,9 @@ spec:
           }
           trap cleanup SIGTERM
 
-{{ if .IPsecCheckForLibreswan }}
-          if rpm --dbpath=/usr/share/rpm -q libreswan; then
-            echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container will sleep to infinity"
+{{ if .IPsecServiceCheckOnHost }}
+          if chroot /proc/1/root systemctl is-active --quiet ipsec.service; then
+            echo "host has ipsec.service running and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container will sleep to infinity"
             sleep infinity
           fi
 {{ end }}
@@ -294,9 +294,6 @@ spec:
           name: host-var-log-ovs
         - mountPath: /etc/openvswitch
           name: etc-openvswitch
-        - mountPath: /usr/share/rpm
-          name: host-usr-share-rpm
-          readOnly: true
         resources:
           requests:
             cpu: 10m
@@ -309,9 +306,9 @@ spec:
             - -c
             - |
               #!/bin/bash
-{{ if .IPsecCheckForLibreswan }}
-              if rpm --dbpath=/usr/share/rpm -q libreswan; then
-                echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container is always \"alive\""
+{{ if .IPsecServiceCheckOnHost }}
+              if chroot /proc/1/root systemctl is-active --quiet ipsec.service; then
+                echo "host has ipsec.service running and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container is always \"alive\""
                 exit 0
               fi
 {{ end }}
@@ -348,10 +345,6 @@ spec:
       - name: host-cni-netd
         hostPath:
           path: "{{.CNIConfDir}}"
-      - name: host-usr-share-rpm
-        hostPath:
-          path: /usr/share/rpm
-          type: DirectoryOrCreate
       tolerations:
       - operator: "Exists"
 {{end}}
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -49,8 +49,6 @@ type OVNUpdateStatus struct {
 // OVNIPsecStatus contains status of current IPsec configuration
 // in the cluster.
 type OVNIPsecStatus struct {
-	// LegacyIPsecUpgrade true if IPsec in 4.14.x cluster is upgraded to 4.15.x version.
-	LegacyIPsecUpgrade bool
 	// IsOVNIPsecActiveOrRollingOut set to true unless we are sure it is not. Note that this is
 	// set to true when ovnkube-node daemonset is in progressing state which is not reflecting
 	// actual ovn ipsec state. so must be precautious in making decisions at the time of machine
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -46,6 +46,7 @@ import (
 	iputil "github.com/openshift/cluster-network-operator/pkg/util/ip"
 	"github.com/openshift/cluster-network-operator/pkg/util/k8s"
 	mcutil "github.com/openshift/cluster-network-operator/pkg/util/machineconfig"
+	"github.com/openshift/cluster-network-operator/pkg/version"
 )
 
 const CLUSTER_CONFIG_NAME = "cluster-config-v1"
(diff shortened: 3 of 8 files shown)


bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
```diff
@@ -288,6 +309,12 @@ spec:
             - -c
             - |
               #!/bin/bash
+{{ if .IPsecCheckForLibreswan }}
+              if rpm --dbpath=/usr/share/rpm -q libreswan; then
+                echo "host has libreswan and therefore ipsec will be configured by ipsec host daemonset, this ovn ipsec container is always \"alive\""
```
jcaamano, line 314: What do you mean here with `is always alive`?

pkg/network/ovn_kubernetes.go
```diff
@@ -270,10 +270,11 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 
 	IPsecMachineConfigEnable, OVNIPsecDaemonsetEnable, OVNIPsecEnable, renderIPsecHostDaemonSet, renderIPsecContainerizedDaemonSet,
-		renderIPsecDaemonSetAsCreateWaitOnly := shouldRenderIPsec(c, bootstrapResult)
+		renderIPsecDaemonSetAsCreateWaitOnly, renderBothIPsecDemonSetsWhenAPoolPausedState := shouldRenderIPsec(c, bootstrapResult)
 	data.Data["IPsecMachineConfigEnable"] = IPsecMachineConfigEnable
 	data.Data["OVNIPsecDaemonsetEnable"] = OVNIPsecDaemonsetEnable
 	data.Data["OVNIPsecEnable"] = OVNIPsecEnable
+	data.Data["IPsecCheckForLibreswan"] = renderBothIPsecDemonSetsWhenAPoolPausedState
```
jcaamano, line 277: Couldn't this just be
```
data.Data["IPsecCheckForLibreswan"] = renderIPsecHostDaemonSet && renderIPsecContainerizedDaemonSet
```

pkg/network/ovn_kubernetes.go
```diff
@@ -620,7 +621,9 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// On upgrade, we will just remove any existing ipsec deployment without making any
 	// change to them. So during upgrade, we must keep track if IPsec MachineConfigs are
 	// active or not for non Hybrid hosted cluster.
-	isIPsecMachineConfigActive := isIPsecMachineConfigActive(bootstrapResult.Infra)
+	machineConfigPoolPaused := isThereAnyMachineConfigPoolPaused(bootstrapResult.Infra)
+	isIPsecMachineConfigActiveInUnPausedPools := isIPsecMachineConfigActive(bootstrapResult.Infra, true)
```
jcaamano, line 625: I would move these two variables to the same block where `renderBothIPsecDemonSetsWhenAPoolPausedState` is defined. And then I would elaborate a bit more the comment of that block saying that if there are unpaused pools, we wait until thos poolls have the ipsec machine config active before deploying both daemonsets

pkg/network/ovn_kubernetes.go
```diff
@@ -653,7 +664,7 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 
 	// While OVN ipsec is being upgraded and IPsec MachineConfigs deployment is in progress
 	// (or) IPsec config in OVN is being disabled, then ipsec deployment is not updated.
-	renderIPsecDaemonSetAsCreateWaitOnly = isIPsecMachineConfigNotActiveOnUpgrade || (isOVNIPsecActive && !renderIPsecOVN)
+	renderIPsecDaemonSetAsCreateWaitOnly = (isIPsecMachineConfigNotActiveOnUpgrade && !renderBothIPsecDemonSetsWhenAPoolPausedState) || (isOVNIPsecActive && !renderIPsecOVN)
```
jcaamano, line 667: This condition is counter-intuitive.

What about 
```
...isIPsecMachineConfigNotActiveOnUpgrade || !isIPsecMachineConfigActiveInUnPausedPools ... 
```

Also since you changed the condition, please update the comment

pkg/network/ovn_kubernetes.go
```diff
@@ -629,16 +632,24 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
...
 	// at the time of upgrade though user created IPsec Machine Config is not
 	// present/active.
-	renderIPsecHostDaemonSet = (renderIPsecDaemonSet && isIPsecMachineConfigActive && !isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade
+	renderIPsecHostDaemonSet = (renderIPsecDaemonSet && isIPsecMachineConfigActive && !isHypershiftHostedCluster) ||
+		isIPsecMachineConfigNotActiveOnUpgrade || renderBothIPsecDemonSetsWhenAPoolPausedState
 
 	// The containerized ipsec deployment is only rendered during upgrades or
 	// for hypershift hosted clusters.
-	renderIPsecContainerizedDaemonSet = (renderIPsecDaemonSet && isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade
+	renderIPsecContainerizedDaemonSet = (renderIPsecDaemonSet && isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade ||
```
jcaamano, line 651: Since you changed the condition, please update the comment

pkg/network/ovn_kubernetes.go
```diff
@@ -629,16 +632,24 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
...
+	renderBothIPsecDemonSetsWhenAPoolPausedState = machineConfigPoolPaused && isIPsecMachineConfigActiveInUnPausedPools &&
+		!isIPsecMachineConfigActive
+
 	// If ipsec is enabled, we render the host ipsec deployment except for
 	// hypershift hosted clusters and we need to wait for the ipsec MachineConfig
 	// extensions to be active first. We must also render host ipsec deployment
 	// at the time of upgrade though user created IPsec Machine Config is not
 	// present/active.
-	renderIPsecHostDaemonSet = (renderIPsecDaemonSet && isIPsecMachineConfigActive && !isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade
+	renderIPsecHostDaemonSet = (renderIPsecDaemonSet && isIPsecMachineConfigActive && !isHypershiftHostedCluster) ||
```
jcaamano, line 646: Since you changed the condition, please update the comment

jcaamano (commented):
I am fine with checking the "paused" spec field of the pools for now.

pkg/network/ovn_kubernetes.go
pperiyasamy: the existing condition `(isIPsecMachineConfigNotActiveOnUpgrade && !renderBothIPsecDemonSetsWhenAPoolPausedState)` helps the case at which both daemonsets can be rendered without create-wait annotation, it can't be done with the suggested approach.

pkg/network/ovn_kubernetes.go
pperiyasamy: done

pkg/network/ovn_kubernetes.go
pperiyasamy: yes, done.

pkg/network/ovn_kubernetes.go
jcaamano: So I guess you mean to say is that you want to **update** the daemonsets if machine config is active in unpaused pools and inactive in paused pools.  But I am missing the reasoning.

- Updating the deamonsets is something we didn't want to do if the machine config was not updated. Why? 
- And why can we update them in the case the pools are paused? Are both these reasonings independent?

pperiyasamy: @jcaamano as discussed offline, updated 4.15 PR (https://github.com/openshift/cluster-network-operator/pull/2449) with following:

1. Update with both daemonsets as long as ipsec machine config is not active in any of the pool(s).
2. Get rid of checking 'paused' pool.
3. Remove `LegacyIPsecUpgrade` checks as it's not needed anymore due to update of both daemonsets at the start of upgrade itself.

Would update this PR once IPsec upgrade CI looks clean there.

pkg/network/ovn_kubernetes.go
```diff
@@ -645,15 +625,19 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
...
-	// hypershift hosted clusters and we need to wait for the ipsec MachineConfig
-	// extensions to be active first. We must also render host ipsec deployment
-	// at the time of upgrade though user created IPsec Machine Config is not
-	// present/active.
-	renderIPsecHostDaemonSet = (renderIPsecDaemonSet && isIPsecMachineConfigActive && !isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade
-
-	// The containerized ipsec deployment is only rendered during upgrades or
-	// for hypershift hosted clusters.
-	renderIPsecContainerizedDaemonSet = (renderIPsecDaemonSet && isHypershiftHostedCluster) || isIPsecMachineConfigNotActiveOnUpgrade
+	// hypershift hosted clusters. We must also render host ipsec daemonset
```
jcaamano, line 628: Have you checked that the comment block for the method itself (lines 594-608) is accurate?

pkg/bootstrap/types.go
```diff
@@ -111,11 +110,11 @@ type InfraStatus struct {
...
 
-	// MasterMCPStatus contains machine config pool statuses for pools having master role.
-	MasterMCPStatuses []mcfgv1.MachineConfigPoolStatus
+	// MasterMCPs contains machine config pools having master role.
+	MasterMCPs []mcfgv1.MachineConfigPool
 
-	// WorkerMCPStatus contains machine config pool statuses for pools having worker role.
-	WorkerMCPStatuses []mcfgv1.MachineConfigPoolStatus
+	// WorkerMCPs contains machine config pools having worker role.
+	WorkerMCPs []mcfgv1.MachineConfigPool
```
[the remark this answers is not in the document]
pperiyasamy: yes, right, but now we need to rely on `MachineCongPool` for a new unit test covering MachineConfigPool in paused and unpaused states. updated commit message to reflect this.

pkg/network/ovn_kubernetes.go
pperiyasamy: yes, updated the method comment about new upgrade behavior.

pperiyasamy:
The ipsec upgrade is successful in the run [ci/prow/e2e-aws-ovn-ipsec-upgrade](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2454/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-ipsec-upgrade/1835623511058550784), pluto logs look clean. there are still two api connection related disruption tests are failing. rerunning the job...

pkg/bootstrap/types.go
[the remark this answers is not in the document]
jcaamano: Why would you need to unit test that if the functionality does not depend on that anymore? You are not really testing any new code path or anything. That should be an e2e test instead.

pperiyasamy:
@jcaamano I think this PR now depends on https://github.com/openshift/cluster-network-operator/pull/2383/ because network co is reported with success as soon as ipsec daemonsets are deployed successfully, but ipsec machine configs are being rolled out a bit later and no status is reported for that. Because of this openshift-install returns prematurely without waiting for ipsec machine config rollout to complete, I think this caused failure on this ipsec ci [run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2454/pull-ci-openshift-cluster-network-operator-master-e2e-ovn-ipsec-step-registry/1846962310304763904), but it doesn't have mg report to prove this explanation.

jcaamano:
/lgtm
/approve
/hold

Held for premerge testing

pkg/network/ovn_kubernetes.go
```diff
@@ -287,11 +287,12 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 
 	IPsecMachineConfigEnable, OVNIPsecDaemonsetEnable, OVNIPsecEnable, renderIPsecHostDaemonSet, renderIPsecContainerizedDaemonSet,
-		renderIPsecDaemonSetAsCreateWaitOnly := shouldRenderIPsec(c, bootstrapResult)
+		renderIPsecDaemonSetAsCreateWaitOnly, isUserDefinedIPsecMachineConfigPresent := shouldRenderIPsec(c, bootstrapResult)
```
jcaamano, line 291: Why dont you call `isUserDefinedIPsecMachineConfigInstalled` directly from here instead of going though `shouldRenderIPsec`?

jcaamano (commented):
Two additional  questions.

This is what I see we are doing with the mounts:
```
          /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
          # Check kernel modules
          /usr/libexec/ipsec/_stackmanager start
          # Check nss database status
          /usr/sbin/ipsec --checknss
```
How does this help into avoiding the incompatibility between host and continer?

For the last PR
`so CNOs 4.15 machine configs are unnecessarily rolled out with 4.14 machine config operator`
I thought that CNO was upgraded last, so by the time CNO is 4.15, MCO should be 4.15 as well?

pkg/network/ovn_kubernetes.go
pperiyasamy: yes, right, done.

pperiyasamy:
> Two additional questions.
> 
> This is what I see we are doing with the mounts:
> 
> ```
>           /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
>           # Check kernel modules
>           /usr/libexec/ipsec/_stackmanager start
>           # Check nss database status
>           /usr/sbin/ipsec --checknss
> ```
> 
> How does this help into avoiding the incompatibility between host and continer?

The `ovn-ipsec` container also runs `ipsec whack --trafficstatus` command for liveness probe which is what crashing pluto daemon due to version mismatch. but this never occurs when CNO installs its own IPsec machine config,  we ensure libreswan installed on the ovn-ipsec container and host are always same for every OCP release. so mounting those ipsec paths are not required and it also helps avoiding error that I explained in the commit [f529788](https://github.com/openshift/cluster-network-operator/pull/2454/commits/f529788faf21322880046c00564e6ad16fd3ac97).

> For the last PR `so CNOs 4.15 machine configs are unnecessarily rolled out with 4.14 machine config operator` I thought that CNO was upgraded last, so by the time CNO is 4.15, MCO should be 4.15 as well?

yes, network upgrade happened before machine-config CO upgrade and @huiran0826 found it while testing the fix https://github.com/openshift/cluster-network-operator/pull/2658 with 4.15.

pperiyasamy:
> yes, network upgrade happened before machine-config CO upgrade, @huiran0826 found it while testing the fix #2658 with 4.15.

@jcaamano added timeline info about `4.14 -> 4.15` upgrade scenario that came out of a detailed discussion.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

anuragthehatter:
QE tested following and it looks good

```
Upgrade test:
1. Install 4.18, Enable IPsec, make any one of the machine config pool into paused state.
2. Start the OCP upgrade to the PR build.
3. You may notice both ovn-ipsec-containerized and ovn-ipsec-host ds are rendered when mcp goes into progressing state.
3. When OCP upgrade finishes, cluster is rendered only with ovn-ipsec-host daemonset (because even paused pool is also installed with ipsec machine config).

IPsec enable test:
1. Bring up OCP cluster from PR.
2. make a machine config pool into paused state.
3. Enable IPsec.
5. Ensure IPsec machine configs are installed on unpaused pools, both ipsec DS are rendered.
6. Unpause the mc pool, IPsec machine config will be installed on the unpaused pool now.
7. At the end cluster is rendered only with ovn-ipsec-host ds.
```

jcaamano:
/lgtm

I have some concerns running host binaries against container libraries, particularly when combined with RHEL workers where our coverage might not be as good. I suggested @pperiyasamy to crosscheck with the libreswan team if this is safe to do.

pperiyasamy:
> particularly when combined with RHEL workers where our coverage might not be as good.

yes @jcaamano when I just think now about our 4.15 testing that we did with RHEL workers, we would hit the issue https://issues.redhat.com//browse/OCPBUGS-28676 due to `/usr/share/rpm` path. will have to fix this. thanks for bringing this up.

huiran0826:
I checked with installer team, starting from 4.19 including 4.19, OCP will not support RHEL worker.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854

pperiyasamy:
The latest ipsec upgrade [job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-network-operator-2454-openshift-machine-config-operator-4854-e2e-aws-ovn-ipsec-upgrade/1902049228331421696) with `testwith` command is failing with below monitor error (though upgrade went through fine)

`static pod lifecycle failure - static pod: "kube-apiserver" in namespace: "openshift-kube-apiserver" for revision: 2 on node: "ip-10-0-80-248.ec2.internal" didn't show up, waited: 5m44s`

This seems to be a known bug https://issues.redhat.com/browse/OCPBUGS-36867. Checked ipsec-connect-wait service and ovn-ipsec-host pod logs, those are clean.

Triggered e2e-aws-ovn-ipsec-upgrade job directly now because https://github.com/openshift/machine-config-operator/pull/4854 is already merged (hope this change is effective in CI builds)

PR #2456, opened 2024-08-01
drop networkType migration support

danwinship:
More followup from openshift-sdn removal (#2384).

This gets rid of all of the code (and templating) related to migration (both live and offline).

In theory, we could try to leave the migration "framework" behind, while only getting rid of the openshift-sdn-specific bits, but this didn't really seem like it would work; the generic and the specific are all mixed together. (And at any rate, we don't expect to ever support networkType migration ever again anyway.)

There was discussion before about whether we should leave this code behind since we'll still need to fix bugs in it in older releases, but since there's no way to actually _test_ this code any more in 4.17 and later, it really makes more sense to drop it. (Fixes for migration bugs will now have to start in `release-4.16` rather than `master`.)

This _could_ merge in 4.17, but it doesn't really need to, and even though it's 99% just deletion, it needs to be reviewed somewhat carefully to make sure I didn't accidentally remove something we need to keep. (In particular, we still need to support MTU migration, though CI should catch that if I screwed it up.)

(This is split into multiple smaller commits to make it a little easier to review, but it's really all one piece. In particular, the first commit, "Reject networkType migration" is _not_ something we need to merge for 4.17; 4.17 will already reject a "migrate to openshift-sdn" configuration because we don't support openshift-sdn. The commit here just changes it so it would instead reject it because we don't support migration.)


danwinship:
/verified by @danwinship 

This is removing code, so there's nothing to test

danwinship:
/override ci/prow/e2e-aws-ovn-hypershift-conformance

appear to be permafailing

danwinship:
/verified bypass
this is just removing code, not adding anything new that needs to be tested

tssurya:
> /verified bypass this is just removing code, not adding anything new that needs to be tested

even if its code removal.. I think it should have some level of regression testing and QE should be using `verified` this label AFAIK? so bypass should also come from QE right?

I can see this effort is tracked in Tech Debt epic for 4.22 (thanks for taking care of this loooong standing cleanup) but it wasn't considered towards planning and not sure if rest of the team is aware.. would be good to check with QE if anything needs to be done for this card and the FUPs you have..
all tech debt work this (and nftables porting) should have regression testing at the very least or at least some level of evaluation on whether we need QE or not..

danwinship:
Don't 20 e2e jobs count as "some level of regression testing"?

The docs say that `/verified` can be used by either engineering or QE, but don't offer much guidance as to which is appropriate when. But I will avoid `/verified bypass` in the future.

PR #2460, opened 2024-08-05
SDN-4919: Change the masquerade subnet default value for new clusters 

kyrtapz:
Set the following default masquerade CIDRs in OVNKubernetes for new clusters:
IPv4: `169.254.0.0/17`
IPv6: `fd69::/112`
If the masquerade subnet is provided through the API it takes precedence over
the defaults.

There is no change for upgraded clusters. If not specified through the API the
default masquerade CIDRs are decided by OVNKubernetes and currently are:
IPv4: `169.254.169.0/29`
IPv6: `fd69::/125`

We went with  `169.254.0.0/17` instead of  `169.254.0.0/16` to avoid catching 169.254.169.254(cloud metadata IP).

Introduced the `networkoperator.openshift.io/default-masquerade-network-cidrs` annotation on the ovnkube-node daemonset that stores the new default CIDRs. This is needed to ensure that the values set during cluster creation are retained. The value of this annotation can differ from what is set in the `OVNKubernetesConfig.GatewayConfig.IPv[4|6].InternalMasqueradeSubnet` API field, in this case the API value takes precedence. It's done this way to ensure that if the API field is cleared we fallback to the new default value.

Note that the change to `defaultV4MasqueradeSubnet/defaultV6MasqueradeSubnet` consts is safe since it was only used in live migration checks and live migration is not supported in 4.17 (SDN is deprecated). There is a plan to cleanup this code and this is tracked elsewhere.

### DIFF
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
@@ -16,6 +16,9 @@ metadata:
     {{ if .OVNIPsecEnable }}
     networkoperator.openshift.io/ipsec-enabled: "true"
     {{ end }}
+    {{ if .DefaultMasqueradeNetworkCIDRs }}
+    networkoperator.openshift.io/default-masquerade-network-cidrs: "{{.DefaultMasqueradeNetworkCIDRs}}"
+    {{ end }}
 spec:
   selector:
     matchLabels:
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -13,6 +13,9 @@ metadata:
     kubernetes.io/description: |
       This daemonset launches the ovn-kubernetes per node networking components.
     release.openshift.io/version: "{{.ReleaseVersion}}"
+    {{ if .DefaultMasqueradeNetworkCIDRs }}
+    networkoperator.openshift.io/default-masquerade-network-cidrs: "{{.DefaultMasqueradeNetworkCIDRs}}"
+    {{ end }}
     {{ if .OVNIPsecEnable }}
     networkoperator.openshift.io/ipsec-enabled: "true"
     {{ end }}
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -2,8 +2,9 @@ package bootstrap
 
 import (
 	configv1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+
+	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 )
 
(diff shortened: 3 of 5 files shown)


kyrtapz:
After upgrade there was no change to the masq subnet:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2460/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-upgrade/1820776719288635392/artifacts/e2e-aws-ovn-upgrade/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovnkube-node-2xhmk_ovnkube-controller.log
new cluster used the wider default masq subnet:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2460/pull-ci-openshift-cluster-network-operator-master-e2e-gcp-ovn/1820776719322189824/artifacts/e2e-gcp-ovn/gather-extra/artifacts/pods/openshift-ovn-kubernetes_ovnkube-node-8tc4f_ovnkube-controller.log

kyrtapz:
@arghosh93 @trozet PTAL

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
     {{ if .OVNIPsecEnable }}
     networkoperator.openshift.io/ipsec-enabled: "true"
     {{ end }}
+    {{ if .DefaultMasqueradeNetworkCIDRs }}
+    networkoperator.openshift.io/default-masquerade-network-cidrs: "{{.DefaultMasqueradeNetworkCIDRs}}"
```
trozet, line 20: why is it necessary to annotate the DS? I see you use it in bootstrap to see if there was already a value set. Why is that necessary if we have the default/API values already?

Also, why is this only done in self-hosted and not managed?

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -16,6 +16,9 @@ metadata:
     {{ if .OVNIPsecEnable }}
     networkoperator.openshift.io/ipsec-enabled: "true"
     {{ end }}
+    {{ if .DefaultMasqueradeNetworkCIDRs }}
```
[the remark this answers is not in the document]
kyrtapz: Yes we do, thanks!
Fixed

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
kyrtapz: > why is it necessary to annotate the DS? I see you use it in bootstrap to see if there was already a value set. Why is that necessary if we have the default/API values already?

On new clusters we need to set the annotation so the default values are not wiped on the next reconciliation.
Consider this example:
1. Fresh cluster starts up, CNO configures the default masq subnets
2. OVN-k daemonset is rolled out with the new default masq subnets
3. On next reconciliation we need to ensure we preserve the defaults from step 1

PR #2466, opened 2024-08-08
OCPBUGS-38176: multus, Add openshift-cnv to globalNamespace

qinqon:
For kubevirt passt binding to work a global nad is needed to be configured, this change allow nads being created at openshift-cnv namespace to be read from pods at different namespace kubevirt VMs can use passt with the nad at openshift-cnv


EdDev:
This looks like a very big change, exposing some specific add-on namespace to all other namespaces in the cluster. I.e. why would a non CNV related pod have access to the NADs defined in that CNV specific namespace?

PR #2495, opened 2024-09-10
OCPBUGS-39209: Add configurable subnets while running hybrid-overlay-node binary

arghosh93:
This commit is to append configurable subnets while running hybrid-overlay-node binary in the process of live migration.

This change will prevent subnet overlap check to emmit an error when clusterNetwork CIDR overlps with either join or transit switch or masquerade default subnets even after those subnets have been customized by patching network.operator CR.

### DIFF
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
@@ -482,9 +482,43 @@ spec:
                 --cert-duration={{.NodeIdentityCertDuration}}
               "
             fi
+
+            ovn_v4_masquerade_subnet_opt=
+            if [[ "{{.V4MasqueradeSubnet}}" != "" ]]; then
+              ovn_v4_masquerade_subnet_opt="--gateway-v4-masquerade-subnet {{.V4MasqueradeSubnet}}"
+            fi
+            ovn_v6_masquerade_subnet_opt=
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
Met the following error with build https://github.com/openshift/cluster-network-operator/pull/2498 on 4.16 live migration testing. 

```
status:
  conditions:
  - lastTransitionTime: "2024-09-12T09:43:30Z"
    message: 'Internal error while rendering operator configuration: failed to render
      manifests: failed to render file bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml:
      failed to render manifest bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml:
      template: bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml:482:21:
      executing "bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml" at
      <.V4MasqueradeSubnet>: map has no entry for key "V4MasqueradeSubnet"'
    reason: RenderError
    status: "True"
    type: Degraded
```

arghosh93:
Clean cherry pick to 4.16 is not possible due to change in masquerade subnet holder.
master: https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/ovn_kubernetes.go#L141
4.16: https://github.com/openshift/cluster-network-operator/blob/release-4.16/pkg/network/ovn_kubernetes.go#L157

PR #2501, opened 2024-09-12
OCPBUGS-42004: Add adapter init & server containers for ARO HCP

bryan-cox:
Add the Microsoft adapter init and adapter server sidecar containers to the cloud network config controller deployment.


bryan-cox:
/hold

Microsoft is requesting to not use the sidecar containers.

bryan-cox:
/close

An official decision was made on Monday to not use sidecars to authenticate with Azure. Closing out this PR.

PR #2507, opened 2024-09-19
HOSTEDCP-1981: Add controlplane cli image envar for use with hypershift

Patryk-Stefanski:
Add CLI_CONTROL_PLANE_IMAGE enavr to
cloud-network, multus-admission-controller and node-identity to avoid
registryOverrides propagating to data plane components


Patryk-Stefanski:
@dougbtv could you ptal.

kyrtapz:
Can we do it the other way around to keep it consistent with the already existing [OVN-K image approach](https://github.com/openshift/cluster-network-operator/blob/2419499e3767f8efc4af04c92e693ef103ae21c8/pkg/platform/hypershift.go#L37)?
In HyperShift we would read `CLI_CONTROL_PLANE_IMAGE` and we would use the `CLI_IMAGE` in all other cases. 
Note that the iptables alerter was introduced in 4.16 so less backports! :)

Patryk-Stefanski:
@kyrtapz done

kyrtapz:
@Patryk-Stefanski once the HyperShift PR merges it might be a good idea to remove the if condition for this change to ensure we never end up using the wrong image unknowingly.

Patryk-Stefanski:
/cherry-pick release-4.17 release-4.16 release-4.15 release-4.14

PR #2515, opened 2024-09-30
OCPBUGS-42449: Use CNIConfDir for mounting directory to ovn-ipsec-host pod

pperiyasamy:
The `ovn-ipsec-host` DaemonSet manifest file hardcoded `host-cni-netd` mount point with `/var/run/multus/cni/net.d `directory, but this won't work when multi network is disabled via `disableMultiNetwork` parameter as ovn-k uses `/etc/kubernetes/cni/net.d` directory to write cni config file for this case. So this PR fixes host path to use `CNIConfDir` tag.


kyrtapz:
/retitle OCPBUGS-42449: Use CNIConfDir for mounting directory to ovn-ipsec-host pod

PR #2518, opened 2024-10-02
CNF-14257, CNF-14251: multi-networkpolicy: Enable for `ipvlan` and `bond` networkds

zeeke:
Make the multus-networkpolicies iptables implementation work with `ipvlan` and `bond` network types.


zeeke:
Thanks @pliurh @cgoncalves for reviewing this.

@dougbtv @trozet I believe we need your approval here. Can you please take a look?

PR #2522, opened 2024-10-07
HOSTEDCP-2035: Set up Azure authentication with cert for ARO HCP

bryan-cox:
Passes the appropriate env var in order to authenticate with Azure with certs for ARO HCP

### DIFF
--- a/bindata/cloud-network-config-controller/managed/controller.yaml
+++ b/bindata/cloud-network-config-controller/managed/controller.yaml
@@ -92,8 +92,6 @@ spec:
             value: "{{.KubernetesServicePort}}"
           - name: KUBERNETES_SERVICE_HOST
             value: "{{.KubernetesServiceHost}}"
-          - name: AZURE_MSI_AUTHENTICATION
-            value: "{{.AzureMSIAuthentication}}"
       containers:
       # hosted-cluster-token creates a token with a custom path(/var/run/secrets/hosted_cluster/token)
       # The token path is included in the kubeconfig used by cncc containers to talk to the hosted clusters API server
@@ -185,6 +183,14 @@ spec:
         - name: "NO_PROXY"
           value: "{{ .NO_PROXY}}"
 {{ end }}
+{{- if not (eq .AzureManagedClientID "")}}
+        - name: "ARO_HCP_MI_CLIENT_ID"
+          value: "{{ .AzureManagedClientID }}"
+        - name: "ARO_HCP_TENANT_ID"
+          value: "{{ .AzureManagedTenantID }}"
+        - name: "ARO_HCP_CLIENT_CERTIFICATE_PATH"
+          value: "{{ .AzureManagedCertPath}}"
+{{ end }}
         resources:
           requests:
             cpu: 10m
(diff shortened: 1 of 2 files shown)


bryan-cox:
I tested this PR with the following PRs and was able to deploy an Azure Hosted Cluster from an AKS management cluster:
- https://github.com/openshift/hypershift/pull/4811
- https://github.com/openshift/hypershift/pull/4877
- https://github.com/openshift/hypershift/pull/4903
- https://github.com/openshift/cloud-network-config-controller/pull/156

bindata/cloud-network-config-controller/managed/controller.yaml
```diff
@@ -184,6 +182,14 @@ spec:
 {{ if .NO_PROXY }}
         - name: "NO_PROXY"
           value: "{{ .NO_PROXY}}"
+{{ end }}
+{{- if not (eq .AroHcpSecretProviderClass "")}}
```
kyrtapz, line 186: In your CNCC [PR](https://github.com/openshift/cloud-network-config-controller/pull/156/files#diff-d3dbddd1556958367344cf05b435a98ffdc18ec9c060843fa1a5f4582d22f3b7R594) you are using `ARO_HCP_MI_CLIENT_ID` to determine whether to use Managed Identity or not. Can you use the same condition here?

kyrtapz (commented):
Some minor comments.

bindata/cloud-network-config-controller/managed/controller.yaml
```diff
@@ -203,6 +209,11 @@ spec:
         - name: cloud-token
           mountPath: /var/run/secrets/openshift/serviceaccount
           readOnly: true
+{{- if not (eq .AzureManagedClientID "")}}
+        - name: cncc-cert
+          mountPath: /mnt/certs
```
kyrtapz, line 214: Can we mount this under `/var/run/secrets/` instead? I am not really digging using `/mnt` here

pkg/network/cloud_network.go
```diff
@@ -102,7 +102,11 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 		data.Data["HTTP_PROXY"] = os.Getenv("MGMT_HTTP_PROXY")
 		data.Data["HTTPS_PROXY"] = os.Getenv("MGMT_HTTPS_PROXY")
 		data.Data["NO_PROXY"] = os.Getenv("MGMT_NO_PROXY")
-		data.Data["AzureMSIAuthentication"] = os.Getenv("AZURE_MSI_AUTHENTICATION")
+		data.Data["AzureManagedClientID"] = os.Getenv("ARO_HCP_MI_CLIENT_ID")
+		data.Data["AzureManagedTenantID"] = os.Getenv("ARO_HCP_TENANT_ID")
+		data.Data["AzureManagedCertDirectory"] = "/var/run/secrets/azure/"
```
kyrtapz, line 107: can you define the path as a constant and then do
```
data.Data["AzureManagedCertPath"] = filepath.Join(constant, os.Getenv("ARO_HCP_CLIENT_CERTIFICATE_PATH"))
```

ARO_HCP_CLIENT_CERTIFICATE_PATH could maybe change to ARO_HCP_CLIENT_CERTIFICATE_NAME? just a nit

PR #2524, opened 2024-10-08
Adds UDN list/watch to ovnknode rbac

trozet:
Required due to:
https://github.com/ovn-org/ovn-kubernetes/commit/d1ebf1c6e335cf095458a3beafd1a601d3eb5a71#diff-4daa87967cc1e7e6e9b9a025289ea7576881ce9d8e3a85750a8e0db0ee65512bL173

### DIFF
--- a/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
+++ b/bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
@@ -142,6 +142,7 @@ rules:
     - egressips
     - egressqoses
     - egressservices
+    - userdefinednetworks
   verbs:
     - get
     - list


maiqueb (approved):
Thank you.

tssurya (approved):

tssurya:
/retitle SDN-4930: Adds UDN list/watch to ovnknode rbac

bindata/network/ovn-kubernetes/common/002-rbac-node.yaml
```diff
@@ -142,6 +142,7 @@ rules:
     - egressips
     - egressqoses
     - egressservices
+    - userdefinednetworks
```
kyrtapz, line 145: This should be under `{{- if .OVN_NETWORK_SEGMENTATION_ENABLE }}`.
We can do that as a follow up

tssurya: oh good point, we won't have this CRD if the feature gate is not enabled.. maybe we should re push?
i don't know what it means to add rbac for a non-existent resource..

kyrtapz: > i don't know what it means to add rbac for a non-existent resource..
That should be fine, the RBAC doesn't validate that the resource exists already

tssurya: jolly

PR #2526, opened 2024-10-10
NO-JIRA: OVN-Kubernetes node RBAC tightening

kyrtapz:
This is a follow up to https://github.com/openshift/cluster-network-operator/pull/2524.
Allow to read userdefinednetworks only when the featuregate is enabled.

/cc @tssurya @trozet


kyrtapz:
/hold cancel
the RBAC is present in techpreview lanes only

PR #2531, opened 2024-10-14
CNF-13671: Adjust frr-k8s metrics

fedepaol:
Few adjustments required to make the frr-k8s metrics to appear under prometheus.

Tested locally, will be tested under CI.

### DIFF
--- a/bindata/network/frr-k8s/monitor.yaml
+++ b/bindata/network/frr-k8s/monitor.yaml
@@ -36,22 +36,20 @@ metadata:
     networkoperator.openshift.io/ignore-errors: ""
 spec:
   endpoints:
-    - port: metricshttps
-      honorLabels: true
-      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
-      scheme: https
-      tlsConfig:
-        caFile: /etc/prometheus/configmaps/prometheus.io/scrape/service-ca.crt
-        serverName: openshift-frr-k8s-metrics-service.openshift-frr-k8s.svc
-        insecureSkipVerify: false
-    - port: frrmetricshttps
-      honorLabels: true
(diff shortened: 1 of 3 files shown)


fedepaol:
cc @jcaamano

bindata/network/frr-k8s/monitor.yaml
```diff
@@ -36,26 +36,59 @@ metadata:
...
-        insecureSkipVerify: false
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    honorLabels: true
+    port: metricshttps
+    scheme: https
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      certFile: /etc/prometheus/secrets/metrics-client-certs/tls.crt
+      keyFile: /etc/prometheus/secrets/metrics-client-certs/tls.key
+      serverName: frr-k8s-monitor-service.openshift-frr-k8s.svc
```
jcaamano, line 47: Can you rationalize the `certFile`, `keyFile` and `serverName` changes with me? Would like to understand their reason to be because I don't see a pattern with the other ServiceMonitors we define in CNO.

fedepaol: tbh, I just looked at other service monitors in a running cluster, as mimicking what was in CNO didn't work (turned out the service name wasn't the right one). 

Let me retry that, if it works I'll keep it consistent.

fedepaol: ok, it seems happy with caFile only, updated the PR

jcaamano: OK, I was missing what the `serverName` was supposed to be but I see the connection now. All good.

PR #2532, opened 2024-10-14
OCPBUGS-42605: OCPBUGS-42244: Exporting environment varialbe NODE_CNI for live migration

pliurh:
By checking the environment varialbe NODE_CNI, the CNI processes can know if it shall run in the live migration mode.

### DIFF
--- a/bindata/network/openshift-sdn/sdn.yaml
+++ b/bindata/network/openshift-sdn/sdn.yaml
@@ -77,6 +77,7 @@ spec:
           else
             echo "br-ex doesn't exist"
           fi
+          export NODE_CNI
 
           # retry loop
           RETRY_INTERVAL=3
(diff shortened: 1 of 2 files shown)


bindata/network/openshift-sdn/sdn.yaml
```diff
@@ -70,7 +70,7 @@ spec:
           set -euo pipefail
 {{- if .IsNetworkTypeLiveMigration }}
           set -x
-          NODE_CNI="OpenShiftSDN"
+          export NODE_CNI="OpenShiftSDN"
```
kyrtapz, line 73: this is a bit confusing:
```     
         export NODE_CNI="OpenShiftSDN"
         if ip link show br-ex; then
            echo "br-ex exists"
            NODE_CNI="OVNKubernetes"
         else
            echo "br-ex doesn't exist"
         fi
```
can we keep `NODE_CNI` exported in both cases?

pliurh: Done

PR #2544, opened 2024-10-25
Run IPsec service on the container

pperiyasamy:
The libreswan 4.9 is having a regression issue which causes intermittent traffic failure between ipsec tunnels. So let us make ipsec service to run on the container which would help to change libreswan back into 4.5 as an intermeditate solution.


pkg/platform/platform.go
```diff
@@ -30,6 +30,7 @@ var cloudProviderConfig = types.NamespacedName{
 var (
 	masterRoleMachineConfigLabel = map[string]string{"machineconfiguration.openshift.io/role": "master"}
 	workerRoleMachineConfigLabel = map[string]string{"machineconfiguration.openshift.io/role": "worker"}
+	EWIPsecOnly                  = true
```
jcaamano, line 33: Why can't this be a constant?

Also add a comment explaining what this does and intentions. I guess you are doing it this way to have an easy way to revert in the future. But I am only guessing, it would be nice to have a comment explaining.

PR #2545, opened 2024-10-25
OCPBUGS-39189: fix annotation from deployment to pod

upalatucci:
The annotation should not be added to the deployment but should be added to the pod.


kramaranya (commented):
/lgtm

kyrtapz:
CI looks unhealthy but doesn't seem to be related to this change.
/approve

PR #2549, opened 2024-10-29
OCPBUGS-43740: Pass transit_switch_subnet options in ovnkube-node pod

pliurh:
In OVN-K, we use function CheckForOverlaps() to check if there is any overlapping among the subnet CIDRs in the configuration. If we don't pass the transit_switch_subnet options, ovnkube will use the default transit_switch_subnet in this check instead of the real one. So we pass this options to in ovnkube-node pod, although it's not actually used there.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -615,6 +615,15 @@ data:
         ovn_v6_masquerade_subnet_opt="--gateway-v6-masquerade-subnet {{.V6MasqueradeSubnet}}"
       fi
 
+      ovn_v4_transit_switch_subnet_opt=
+      if [[ "{{.V4TransitSwitchSubnet}}" != "" ]]; then
+        ovn_v4_transit_switch_subnet_opt="--cluster-manager-v4-transit-switch-subnet {{.V4TransitSwitchSubnet}}"
+      fi
+      ovn_v6_transit_switch_subnet_opt=
+      if [[ "{{.V6TransitSwitchSubnet}}" != "" ]]; then
+        ovn_v6_transit_switch_subnet_opt="--cluster-manager-v6-transit-switch-subnet {{.V6TransitSwitchSubnet}}"
+      fi
+
       exec /usr/bin/ovnkube \
         --init-ovnkube-controller "${K8S_NODE}" \
         --init-node "${K8S_NODE}" \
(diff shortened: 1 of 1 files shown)


jcaamano:
@pliurh since this will cause ovnkube-node daemonset to rollout, have you checked for potential doc impacts?

Also, this is kinda broken. This overlap check itself should be done in CNO and not in OVN-K. Otherwise nothing prevents a day 2 update on the transit switch conflicting with existing networks. cc @arghosh93 @tssurya

pliurh:
It the [doc](https://docs.openshift.com/container-platform/4.17/networking/ovn_kubernetes_network_provider/configure-ovn-kubernetes-subnets.html#nw-ovn-kubernetes-change-transit-subnet_configure-ovn-kubernetes-subnets), we didn't mention which pods will restarted. So I think we're fine with the doc.

But I agree that we shall check it in CNO.

arghosh93:
We are checking all kinds of overlaps in CNO but the problem is that we dont check overlap with default subnets if any configurable subnet is not mentioned at network CR.
If we hard code default subnets at CNO then any change in OVN kubernetes side also needs to be propagated at CNO side at the same time. This may or may not be a problem and depends on awareness of the change in OVN Kubernetes side.

pliurh:
Well, I think this patch will fix the issue if we already have the subnets overlapping check in CNO.

jcaamano:
> We are checking all kinds of overlaps in CNO but the problem is that we dont check overlap with default subnets if any configurable subnet is not mentioned at network CR. If we hard code default subnets at CNO then any change in OVN kubernetes side also needs to be propagated at CNO side at the same time. This may or may not be a problem and depends on awareness of the change in OVN Kubernetes side.

I think CNO should take control of what the default subnets are for OCP. Basically if no subnets are specified then CNO should explicitly chose the default. Which need to be the same as the current OVNK defaults and does not mean that if they change in OVN kubernetes side it also needs to be propagated at CNO, we wouldn't be able to do that anyway.

However, I thought that the real reason was to check overlaps with other primary UDNs networks. I guess this is not the case after your comment?

pliurh:
@jcaamano I agree with you that CNO shall always render the subnets explicitly. But that is not the issue this PR is trying to fix. Can we get this one merged first?

zhaozhanqi:
pre-merge tested this patch by

1. update the internalTransitSwitchSubnet to another subnet. 
oc patch network.operator.openshift.io cluster --type='merge' -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalTransitSwitchSubnet": "100.69.0.0/16"}}}}}'
2. then update the internalJoinSubnet to the default value of internalTransitSwitchSubnet 100.88.0.0/16

oc patch network.operator.openshift.io cluster --type='merge' -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalJoinSubnet": "100.88.0.0/16"}}}}}'

that means when updating internalJoinSubnet which is validating with new subnet 100.69.0.0/16 other than the default value `100.88.0.0/16`
and ovn-kubernetes pod can be worked 

/label qe-approved

PR #2552, opened 2024-10-31
[release-4.16] OCPBUGS-43803: Start feature migration when the cluster CNI is changed to the target type

pliurh:
The first step in an offline migration is settingspec.migration.networkType, which triggers the cluster to update the MachineConfigPools. It also triggers feature migration, the egress IP previously assigned by the SDN will be released. At this stage, however, OVN-K is not yet deployed; the SDN continues to manage networking, allowing the egress IP functionality to remain operational. Therefore, we postpone the feature migration until after spec.default.type has been
modified. It can help to reduce the downtime for egressIP during SDN migration. The impact for live migration is similar.

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -540,26 +540,29 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 			return reconcile.Result{}, err
 		}
 
-		migration := operConfig.Spec.Migration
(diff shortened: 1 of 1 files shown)


kyrtapz:
@pliurh is this improvement specific to offline migration? Could you please add some context in the PR description about how is this helping reduce downtime?

pliurh:
> @pliurh is this improvement specific to offline migration? Could you please add some context in the PR description about how is this helping reduce downtime?

I was wrong, it can also help live migration. I updated the commit message, PTAL.

PR #2553, opened 2024-10-31
SDN-5436: Use host ipsec binary inside ovn-ipsec container 

pperiyasamy:
The ovn-ipsec container for the ovn-ipsec-host pod has libreswan installed and uses its ipsec command to configure ipsec service running on the host. But there is a chance versions of libreswan installed on the host and container may be different. This causes coredump on the ipsec service while running ipsec whack command from the container. 
This PR mounts ipsec binary and its dependent libraries from the host to avoid incompatibility behavior.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -305,6 +305,10 @@ spec:
           name: host-var-lib
         - mountPath: /etc
           name: host-etc
+        - mountPath: /usr/sbin/ipsec
+          name: ipsec-bin
+        - mountPath: /usr/libexec/ipsec
+          name: ipsec-lib
         resources:
           requests:
             cpu: 10m
@@ -359,6 +363,14 @@ spec:
           path: /etc
           type: Directory
         name: host-etc
+      - hostPath:
+          path: /usr/sbin/ipsec
+          type: File
+        name: ipsec-bin
+      - hostPath:
+          path: /usr/libexec/ipsec
+          type: Directory
+        name: ipsec-lib
       tolerations:
       - operator: "Exists"
 {{end}}


bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -359,6 +363,14 @@ spec:
           path: /etc
           type: Directory
         name: host-etc
+      - hostPath:
+          path: /usr/sbin/ipsec
+          type: FileOrCreate
```
trozet, line 368: why add create? wouldn't this pod only be launched if the machine extension is enabled?

pperiyasamy:
> /test e2e-aws-ovn-ipsec-serial
> /test e2e-aws-ovn-ipsec-upgrade

@jcaamano i see above two jobs failed because of other reasons though ipsec upgrade and install are eventually successful. with the upgrade, there are api connection failures during the upgrade which theoretically shouldn't happen though nodes are rebooted. This [change](https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-host.yaml#L233-L237) may introduce traffic failure for temporary period, trying to investigate the issue with a release [PR](https://github.com/openshift/release/pull/55842) on 4.15.
on the serial job failure, we do have ipsec failure on the external mode, there is a potential fix is available on the [PR](https://github.com/openshift/origin/pull/28797), we must test it. the NS traffic is failed on another test, we may need to investigate on that. looks like we must collect ipsec logs at every test failure. must gather being collected at the end which is not more useful to troubleshoot an individual test. so would add another commit  on the [PR](https://github.com/openshift/origin/pull/28797) to do this.

So these CI failures are not related to specific changes from this PR.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: yes @trozet , correct. the `ovn-ipsec-host ` pod are launched only when ipsec machine configs are active.
updated those hostPath volume types accordingly.

PR #2554, opened 2024-11-04
SDN-5436: Provide support for user owned IPsec machine configs

pperiyasamy:
There was a regression issue found with libreswan 4.9 which breaks pod-to-pod traffic over ipsec tunnels. The libreswan and its dependent packages are installed by ipsec machine config extension which is rolled out by network operator.   So we must rollback to libreswan 4.5 to fix the problem, but this is not directly possible with ipsec machine config extension.

As an immediate workaround (until it gets fixed properly in next libreswan release), we must skip network operator from rendering its own machine config and allow users to install their own machine config and it can install previous version of libreswan (say 4.5) which is working fine now with OCP. 

This PR adds required changes to allow for user owned ipsec machine configs to be present with a condition that it must have `ipsec-machine-config: true` set on it's annotation and skips rendering operator owned ipsec machine configs to avoid conflict.

### DIFF
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -34,6 +34,7 @@ import (
 	cnofake "github.com/openshift/cluster-network-operator/pkg/client/fake"
 	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/cluster-network-operator/pkg/platform"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


pkg/network/ovn_kubernetes_test.go
```diff
@@ -3260,8 +3400,10 @@ func TestRenderOVNKubernetesDisableIPsecWithUserInstalledIPsecMachineConfigs(t *
 	bootstrapResult.Infra = bootstrap.InfraStatus{}
 	bootstrapResult.Infra.MasterIPsecMachineConfigs = []*mcfgv1.MachineConfig{{}}
 	bootstrapResult.Infra.MasterIPsecMachineConfigs[0].Name = masterMachineConfigIPsecExtName
+	bootstrapResult.Infra.MasterIPsecMachineConfigs[0].Annotations = platform.UserDefinedIPsecMachineConfigAnnotation
```
kyrtapz, line 3403: Why is this required in the existing test?

pkg/platform/platform.go
```diff
@@ -31,8 +31,9 @@ var cloudProviderConfig = types.NamespacedName{
 }
 
 var (
-	masterRoleMachineConfigLabel = map[string]string{"machineconfiguration.openshift.io/role": "master"}
-	workerRoleMachineConfigLabel = map[string]string{"machineconfiguration.openshift.io/role": "worker"}
+	masterRoleMachineConfigLabel            = map[string]string{"machineconfiguration.openshift.io/role": "master"}
+	workerRoleMachineConfigLabel            = map[string]string{"machineconfiguration.openshift.io/role": "worker"}
+	UserDefinedIPsecMachineConfigAnnotation = map[string]string{"ipsec-machine-config": "true"}
```
kyrtapz, line 36: NIT: should the annotation include some clue about this being a user-defined ipsec configuration?

pkg/platform/platform.go
```diff
@@ -275,3 +277,15 @@ func consolePluginCRDExists(cl cnoclient.Client) (bool, error) {
 	}
 	return true, nil
 }
+
+func IsUserDefinedIPsecMachineConfig(machineConfig *mcfgv1.MachineConfig) bool {
```
kyrtapz, line 281: NIT: I think this could use a docblock explaining what does a UserDefineIPsecMachineConfig cause.

kyrtapz (commented):
Overall I am ok with the proposed change but it is very specific to machineconfigs for IPsec. I was thinking about a more generic solution that would allow a power user to disable reconciliation of specific resources managed by CNO.

I would like for @jcaamano to take a look as well as he reviewed the IPsec logic(especially the machineconfig reconciliation).

/approve

pkg/platform/platform.go
jcaamano: I think @kyrtapz is referring to the key of the annotation itself.

so `user-ipsec-machine-config` vs `ipsec-machine-config`
or `custom-ipsec-machine-config`
or `ipsec-machine-config-override`

pperiyasamy: oh! yes @jcaamano , i missed it. using `user-ipsec-machine-config` as the key for annotation now.

PR #2557, opened 2024-11-07
OCPBUGS-44344: The sysctl allowlist daemonset should set requests for memory and cpu

dougbtv:
Otherwise, sometimes this fails intermittently in CI. Borrowed request amount pattern from other CNI plugin installations.


dougbtv:
Example reference: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus/multus.yaml#L521-L524

dgoodwin:
/retitle OCPBUGS-44344: The sysctl allowlist daemonset should set requests for memory and cpu

PR #2565, opened 2024-11-13
[release-4.16] OCPBUGS-44338: Prevent live migration process from flapping between step-2 and step-3

pliurh:
When a MCP is paused, the NetworkTypeMigrationTargetCNIInUse condition cannot become true. However NetworkTypeMigrationMTUReady will be false in that case, as the mtu-migration.service has been removed from unpaused MCPs. The migration will be flapping between step-2 and step-3. Checking the spec.defaultNetwork.Type before entering step-2 to ensure we don't go back from step-3.

### DIFF
--- a/pkg/controller/clusterconfig/migration.go
+++ b/pkg/controller/clusterconfig/migration.go
@@ -112,19 +112,21 @@ func (r *ReconcileClusterConfig) prepareOperatorConfigForNetworkTypeMigration(ct
 		return nil
 	}
 	if !mtuApplied && cniReady {
-		if isRollback {
-			klog.Infof("step-2: propagate the network type change to network.config")
-			operConfig.Spec.DefaultNetwork.Type = operv1.NetworkType(clusterConfig.Spec.NetworkType)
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterconfig/migration.go
```diff
@@ -111,7 +111,7 @@ func (r *ReconcileClusterConfig) prepareOperatorConfigForNetworkTypeMigration(ct
 		}
 		return nil
 	}
-	if !mtuApplied && cniReady {
+	if !mtuApplied && cniReady && operConfig.Spec.DefaultNetwork.Type != operv1.NetworkType(clusterConfig.Spec.NetworkType) {
 		if isRollback {
```
kyrtapz, line 115: Should this if block happen after mtuMigration retrieval? Currently if calculateRoutableMTU fails we will not re-enter the step 2 block on rollback.

pliurh: Good point. Fixed.

PR #2570, opened 2024-11-19
Add nodeslicepool

bpickard22:
Fast ranges needs the nodeslicepool crd, this adds that and updates the multus unit test to accommodate it


bpickard22:
cc @dougbtv

PR #2584, opened 2024-12-05
[release-4.16] OCPBUGS-45593: Pass transit_switch_subnet options in ovnkube-node pod

arghosh93:
In OVN-K, we use function CheckForOverlaps() to check if there is any overlapping among the subnet CIDRs in the configuration. If we don't pass the transit_switch_subnet options, ovnkube will use the default transit_switch_subnet in this check instead of the real one. So we pass this options to in ovnkube-node pod, although it's not actually used there.

(cherry picked from commit 82c53dcfcc990ec449a217c47625bd8dccca8a87)

Conflict: Fixed a merge conflict and it was due to masquerade subnet not being added as a flag to ovnkube running in ovnkube-controller in release-4.16.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -595,6 +595,15 @@ data:
         ovn_v6_join_subnet_opt="--gateway-v6-join-subnet {{.V6JoinSubnet}}"
       fi
 
+      ovn_v4_transit_switch_subnet_opt=
+      if [[ "{{.V4TransitSwitchSubnet}}" != "" ]]; then
+        ovn_v4_transit_switch_subnet_opt="--cluster-manager-v4-transit-switch-subnet {{.V4TransitSwitchSubnet}}"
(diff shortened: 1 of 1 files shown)


zhaozhanqi:
pre-merge tested 

update the internalTransitSwitchSubnet to another subnet.
oc patch network.operator.openshift.io cluster --type='merge' -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalTransitSwitchSubnet": "100.69.0.0/16"}}}}}'
then update the internalJoinSubnet to the default value of internalTransitSwitchSubnet 100.88.0.0/16
oc patch network.operator.openshift.io cluster --type='merge' -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalJoinSubnet": "100.88.0.0/16"}}}}}'

that means when updating internalJoinSubnet which is validating with new subnet 100.69.0.0/16 other than the default value 100.88.0.0/16
and ovn-kubernetes pod can be worked

/label qe-approved

PR #2605, opened 2024-12-13
WIP: OCPBUGS-46422: Add a ValidatingAdmissionPolicy blocking ServiceCIDR changes

danwinship:
The ServiceCIDR API is beta (and thus available by default) in k8s 1.31, but it will not work correctly in OCP, because (among other things) we do not support dynamically updating ovn-kubernetes with changes to the service CIDRs. Thus, we need to block admins from using this API (while not interfering with kube-apiserver's own use of the API).


kyrtapz:
> (The reason that the policy allows changes to the default (`"kubernetes"`) ServiceCIDR is to allow single-stack to dual-stack migration; the apiserver itself will modify the default ServiceCIDR in that case when it is reconfigured.)

Do you think we could also check the user? To ensure only the apiserver can change the default ServiceCIDR.

danwinship:
> Do you think we could also check the user? To ensure only the apiserver can change the default ServiceCIDR.

ah... yeah. I guess the upstream examples don't do that because not all clusters use RBAC...

danwinship:
updated with validation expression from https://github.com/kubernetes/website/pull/51748

bindata/cluster-network-operator/servicecidr-vap.yaml
```diff
@@ -0,0 +1,25 @@
...
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["networking.k8s.io"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["servicecidrs"]
+  validations:
+    # Allow apiserver to change the 'kubernetes' ServiceCIDR (when doing single/dual-stack
```
kyrtapz, line 14: The comment mentions the 'kubernetes' ServiceCIDR but he CEL doesn't check for that.
Is it intentional or missing?

danwinship: Right, the effect is "allow the apiserver to make whatever changes it wants", but the only change the apiserver ever wants to make is to modify `kubernetes` if you change single/dual-stack. So the comment is just providing some additional clarification about _why_ we want this exception.

danwinship:
This fails because there's a test of just "you can do basic kubernetes API operations on ServiceCIDR objects" which fails in the presence of the VAP. And we can't skip the test, because it's Conformance. I filed https://github.com/kubernetes/kubernetes/issues/133613...

danwinship:
/hold cancel

it might take a bit to propagate through, but https://github.com/openshift/kubernetes/pull/2444 fixes the conformance test failure, so once e2e passes, this can merge

danwinship:
which is nearly permafailing and not affected by this PR anyway

danwinship:
/verified by "ServiceCIDR should be blocked"

(new test from https://github.com/openshift/origin/pull/30234, which can merge after this does. Test run at https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/1965905253861691392.)

PR #2610, opened 2024-12-27
OCPBUGS-46564: Update egressfirewall CRD to be consistent with ovn-kubernetes repo

arkadeepsen:
This PR updates the egressfirewall CRD with the latest CRD from the ovk-kubernetes repo. This fixes the issue of mismatch of regex validation of the DNS name field between the two CRDs.


npinaeva:
pattern looks correct
/lgtm

PR #2611, opened 2025-01-09
SDN-4168: Cleanup ipsec state only when ipsec is not full mode

pperiyasamy:
This PR does the following to fixes to prevent unnecessary ipsec service restart, ip xfrm state policy cleanups while bringing up ipsec-host pod. This would potentially avoid reestablishment of IKE SAs during ipsec pod restarts and let OVN networking pods traffic go on without any packet drops.

1. There is an incorrect check in ipsec pod clean up logic which removes `/etc/ipsec.d/openshift.conf` file, ip xfrm state and policy entries in all cases, but these must be removed only when ipsec mode is changed from full to external or disabled.
2. We don't need narrowing=yes option to be set explicitly anymore because system default crypto policies are commented out now, otherwise `TS_UNACCEPTABLE` error is seen temporarily at the time of ipsec service restart.
3. The IPsec service restart is needed only at the time of specific IPsec config changes, so doing ipsec service only at the time commenting out default crypto-policies conf file.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -407,7 +407,7 @@ spec:
           # When east-west ipsec is not disabled, then do not flush xfrm states and
           # policies in order to maintain traffic flows during container restart.
           ipsecflush() {
-            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] || \
+            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] && \
                [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig}')" != "{}" ]; then
               ip x s flush
               ip x p flush
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -239,23 +239,22 @@ spec:
           defaultcpinclude="include \/etc\/crypto-policies\/back-ends\/libreswan.config"
           if ! grep -q "# ${defaultcpinclude}" /etc/ipsec.conf; then
             sed -i "/${defaultcpinclude}/s/^/# /" /etc/ipsec.conf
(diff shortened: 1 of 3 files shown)


pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-ovn-ipsec-step-registry https://github.com/openshift/origin/pull/29232

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-ovn-ipsec-step-registry https://github.com/openshift/origin/pull/29232

pperiyasamy:
The  `[sig-arch][Late] operators should not create watch channels very often [apigroup:apiserver.openshift.io] [Suite:openshift/conformance/parallel] ` test is failing in the run https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-network-operator-2611-openshift-origin-29232-e2e-ovn-ipsec-step-registry/1877412979897536512 upon ipsec pod reboots. 
It happens even when there is no ipsec state/policy cleanup, no pluto restart. Needs investigation...

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-ovn-ipsec-step-registry https://github.com/openshift/origin/pull/29232

pperiyasamy:
> The `[sig-arch][Late] operators should not create watch channels very often [apigroup:apiserver.openshift.io] [Suite:openshift/conformance/parallel] ` test is failing in the run https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-network-operator-2611-openshift-origin-29232-e2e-ovn-ipsec-step-registry/1877412979897536512 upon ipsec pod reboots. It happens even when there is no ipsec state/policy cleanup, no pluto restart. Needs investigation...

it may be a flaky test, tracking it via bug https://issues.redhat.com/browse/OCPBUGS-46414.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -407,7 +392,7 @@ spec:
           # When east-west ipsec is not disabled, then do not flush xfrm states and
           # policies in order to maintain traffic flows during container restart.
           ipsecflush() {
-            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] || \
+            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] && \
```
jcaamano, line 395: What if mode is "Disabled"? Are you sure the comment above matches what you are doing?

PR #2612, opened 2025-01-09
OCPBUGS-42609: Disable adding/removing the UDN namespace label

trozet:
UDN namespace label will be required when namespaces are created in order to use a primary UDN in that namespace.

Upstream PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4912

### DIFF
--- a/bindata/network/ovn-kubernetes/common/udn-admission-policy.yaml
+++ b/bindata/network/ovn-kubernetes/common/udn-admission-policy.yaml
@@ -1,0 +1,32 @@
+{{if .OVN_NETWORK_SEGMENTATION_ENABLE}}
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: user-defined-networks-namespace-label
+spec:
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["v1"]
+        operations:  ["UPDATE"]
+        resources:   ["namespaces"]
+  failurePolicy: Fail
+  validations:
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/common/udn-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: disable-udn-label-change-binding
+spec:
+  policyName: disable-udn-label-change
+  validationActions: [Deny]
+  matchResources:
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["v1"]
+        operations:  ["UPDATE"]
+        resources:   ["namespaces"]
```
kyrtapz, line 30: NIT: missing an empty line

bindata/network/ovn-kubernetes/common/udn-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: admissionregistration.k8s.io/v1
```
kyrtapz, line 1: We should only apply this if UDN is enabled

bindata/network/ovn-kubernetes/common/udn-admission-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: disable-udn-label-change
```
kyrtapz, line 4: Can we make it more user friendly? something like:
`user-defined-networks-namespace-label`

PR #2617, opened 2025-01-14
OCPBUGS-48323: Pass transit_switch_subnet options in ovnkube-node pod for single-zone

pliurh:
The single-zone mode is a phase in the multi-step upgrade process for IC. To ensure the subnet overlapping check uses the customized value, it is necessary to pass the transit_switch_subnet parameter to ovn-k when configuring this mode.

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/common/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/managed/common/004-config.yaml
@@ -60,6 +60,14 @@ data:
 {{- end }}
 {{- if .OVNHybridOverlayEnable }}
 
+    [clustermanager]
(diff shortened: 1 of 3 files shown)


ricky-rav:
@arghosh93 PTAL

zhaozhanqi:
pre-merge test pass with this fix

Upgrade from 4.13 build with fix https://github.com/openshift/cluster-network-operator/pull/2618 to this fixed version
by `oc patch network.operator.openshift.io cluster --type='merge' -p='{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"ipv4":{"internalTransitSwitchSubnet": "100.78.0.0/16"}}}}}'` and then do upgrade. 

/label qe-approved

PR #2621, opened 2025-01-20
SDN-5330: Keep IPsec in OVN enabled during the upgrade

pperiyasamy:
The following change on the machine pool:
 
`status.MachineCount == status.UpdatedMachineCount && hasSourceInMachineConfigStatus(status, machineConfigs)` 

is introduced with PR https://github.com/openshift/cluster-network-operator/pull/2349 which ensures IPsec machine config is always installed on all the nodes in the cluster,
So this is deleting the IPsec daemonset as per the CNO state machine for IPsec when the condition is not met. But this is also accidentally disabling IPsec in OVN which is not an expected behavior. This causes ovs-monitor-ipsec to refresh existing ipsec connections unnecessarily when IPsec pod comes up as it is not able to find remote_name from the tunnel. This may also trigger deleting IPsec connection entries from openshift.conf file if ovs-monitor-ipsec is not killed timely when ipsec daemonset is removed. So this PR enables ovn ipsec option as long as the API is set with Full mode.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -613,69 +613,72 @@ func IsIPsecLegacyAPI(conf *operv1.OVNKubernetesConfig) bool {
 	return conf.IPsecConfig == nil || conf.IPsecConfig.Mode == ""
 }
 
-// shouldRenderIPsec method ensures the have following IPsec states for upgrade path from 4.14 to 4.15 or later versions:
-// When 4.14 cluster is already installed with MachineConfig for IPsec extension and ipsecConfig is set in network operator
-// config (i.e. IPsec for NS+EW), then render CNO's IPsec MC extension and ipsec-host daemonset.
-// When 4.14 cluster is just running with ipsecConfig set in network operator config (i.e. IPsec for EW only), then activate
-// IPsec MachineConfig and render ipsec-host daemonset.
-// When 4.14 cluster is just installed with MachineConfig for IPsec extension (i.e. IPsec for NS only), then just keep MachineConfig
-// to be in the same state without rendering IPsec daemonsets.
-// When 4.14 cluster is Hypershift cluster running with ipsecConfig set, then just render ovn-ipsec-containerized daemonset as
-// MachineConfig kind is not supported there.
-// For Upgrade path from pre-4.14 to 5.15 or later versions:
-// When pre-4.14 cluster is just running with ipsecConfig set in network operator config (i.e. IPsec for EW only), then activate
-// IPsec MachineConfig and render ipsec-host daemonset.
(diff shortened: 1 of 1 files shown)


pperiyasamy:
/assign @jcaamano @igsilya @trozet 

Thanks to @igsilya for catching this issue while troubleshooting CI issues for IPsec.

pkg/network/ovn_kubernetes.go
```diff
@@ -668,10 +668,8 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 	// Wait for MCO to be ready unless we had already rendered the IPsec MachineConfig.
 	renderCNOIPsecMachineConfig = renderCNOIPsecMachineConfig && (isCNOIPsecMachineConfigPresent || isMachineConfigClusterOperatorReady)
 
-	// We render OVN IPsec if East-West IPsec is enabled or it's upgrade is in progress.
-	// If NS IPsec is enabled as well, we need to wait to IPsec MachineConfig
-	// to be active if it's not an upgrade and not a hypershift hosted cluster.
-	renderIPsecOVN = (renderIPsecHostDaemonSet || renderIPsecContainerizedDaemonSet) && mode == operv1.IPsecModeFull
+	// We render OVN IPsec as long as East-West IPsec is enabled.
+	renderIPsecOVN = mode == operv1.IPsecModeFull
```
jcaamano, line 672: Why not
```
renderIPsecOVN = (renderIPsecHostDaemonSet || renderIPsecContainerizedDaemonSet || isOVNIPsecActive) && mode == operv1.IPsecModeFull
```

Why don't we need to wait anymore for the NS IPSec to be deployed before activating OVN IPSec?

Also, it's best if the comment explains why we are doing what we are doing, not just what we are doing.

jcaamano (commented):
> Due to known reasons as introduced in the PR https://github.com/openshift/cluster-network-operator/pull/2349, The IPsec daemonset is removed at the time of OCP upgrade or MCP in progressing state.

Can you please elaborate?

pkg/network/ovn_kubernetes.go
pperiyasamy: >Why don't we need to wait anymore for the NS IPSec to be deployed before activating OVN IPSec?

activating OVN IPSec at this case won't trigger anything because ovs-monitor-ipsec script is not started yet. but still your suggestion on checking `isOVNIPsecActive` makes more sense as it enables IPsec OVN only when needed. changed it. added some info about explaining 'why', hope that's ok now.

pperiyasamy: Thanks @jcaamano for your time and patience, it was a great long discussion to update IPsec state machine with up to  date information. Updated the PR as we discussed.

PR #2606, opened 2024-12-18
SDN-5330: Wait enough for IPsec tunnels to establish before kubelet start

pperiyasamy:
When node goes for a reboot on an IPsec enabled cluster, once it comes up, libreswan parses /etc/ipsec.d/openshift.conf file and establishes SAs with peers and it may be still in progress even after kubelet is started, pod scheduled on this node would fail communicating with other pods until IPsec tunnels are established. So this commit makes ipsecenabler systemd service to wait upto 2 minutes until IPsec tunnels are established and this would give better chance for ipsec CI lanes to run fine without disruptive events.

The butane configs are translated into machine configs with command `butane <file>.bu -o <file>.yaml` and only storage portion is copied manually into IPsec machine config files.

PR #2628, opened 2025-01-27
SDN-5330: Keep rendering OVN IPsec when its daemonset not available

pperiyasamy:
The commit https://github.com/openshift/cluster-network-operator/commit/4e57dcd195992dd6d19b2d1a731d6673c9ee2e4b is not complete because `OVNIPsecStatus` is still not set when none of the IPsec daemonset exists on the cluster at the time of machine config pools are updating
(or) node is rebooted. Hence fixing it by `OVNIPsecStatus` is always set to reflect ipsec deployment state of the cluster
and update the render pipeline to render ovn ipsec for the above mentioned scenarios.

It renders ovn ipsec even before ipsec daemonsets are deployed when IPsec is freshly enabled on the cluster. That's ok	because It will be effective only when the `ovs-monitor-ipsec `script is started and that's going to be done only when the ipsec pod is running. so we are safe to ignore it now.

When IPsec is disabled from API, ovn ipsec is disabled followed by stop	rendering ipsec machine config and ipsec daemonset. While ipsec machine configs are removed which	would make `ovnkube-node` daemonset into progressing state and `OVNIPsecActive` condition becomes true again. Hence this pr considers machine config
status as well so that the rendering pipeline will not render IPsec machine configs again.

It also removes dead code related to IPsec 4.13 upgrade which is no longer valid for upgrade scenarios beyond >= 4.15. This change is intended to be backported until 4.15 so still keeping legacy upgrade scenarios.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -610,9 +610,16 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
 
 	isHypershiftHostedCluster := bootstrapResult.Infra.HostedControlPlane != nil
 	isIpsecLegacyUpgrade := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.LegacyIPsecUpgrade
-	isOVNIPsecActive := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.OVNIPsecActive
+	isOVNIPsecActiveOrRollingOut := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.IsOVNIPsecActiveOrRollingOut
+	isCNOIPsecMachineConfigPresent := isCNOIPsecMachineConfigPresent(bootstrapResult.Infra)
+	isUserDefinedIPsecMachineConfigPresent := isUserDefinedIPsecMachineConfigPresent(bootstrapResult.Infra)
+	isMachineConfigClusterOperatorReady := bootstrapResult.Infra.MachineConfigClusterOperatorReady
 
 	mode := GetIPsecMode(conf)
+
+	// when OVN is rolling out, OVN IPsec might be fully or partially active or inactive.
+	// If MachineConfigs are not present, we know its inactive since we only stop rendering them once inactive.
+	isOVNIPsecActive := isOVNIPsecActiveOrRollingOut && (isCNOIPsecMachineConfigPresent || isUserDefinedIPsecMachineConfigPresent || isHypershiftHostedCluster)
 
 	// We render the ipsec deployment if IPsec is already active in OVN
 	// or if EW IPsec config is enabled.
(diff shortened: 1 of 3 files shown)


pperiyasamy:
/hold

this breaks ipsec disable, needs fixing

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/cluster-network-operator/pull/2606

pkg/network/ovn_kubernetes.go
```diff
@@ -1214,77 +1195,53 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
...
-	if ipsecStatus != nil {
-		ipsecStatus.OVNIPsecActive = ovnIPsecStatus.OVNIPsecActive
+	if ipsecContainerizedDaemonSet != nil && ipsecHostDaemonSet != nil {
+		// Both IPsec daemonset versions exist, so this is an upgrade from 4.14.
+		ovnIPsecStatus.LegacyIPsecUpgrade = true
+	} else if ipsecContainerizedDaemonSet == nil && ipsecHostDaemonSet == nil &&
+		mode == operv1.IPsecModeDisabled {
+		// set OVN IPsec status to nil when none of the IPsec daemonset(s) exists
+		// and IPsec is disabled from API.
+		ovnIPsecStatus = nil
```
jcaamano, line 1237: Why consider mode here? We keep making changes I have no explanation for.
This should be decided elsewhere, as this is not the place to make decisions based on configuration, this should just be a representation of the state of the system.

pperiyasamy: ok, moved this to `shouldRenderIPsec` method now.

pkg/network/ovn_kubernetes.go
```diff
@@ -633,18 +614,20 @@ func shouldRenderIPsec(conf *operv1.OVNKubernetesConfig, bootstrapResult *bootst
...
-	// We render the ipsec deployment if IPsec is already active in OVN
-	// or if EW IPsec config is enabled.
-	renderIPsecDaemonSet = isOVNIPsecActive || mode == operv1.IPsecModeFull
+	// We render the ipsec deployment if IPsec is already active in OVN or if EW IPsec config is enabled.
+	// When it's not a hypershift hosted cluster we also need to ensure machine configs present though
+	// isOVNIPsecActive is true.
+	isCNOIPsecMachineConfigPresent := isCNOIPsecMachineConfigPresent(bootstrapResult.Infra)
+	isUserDefinedIPsecMachineConfigPresent := isUserDefinedIPsecMachineConfigPresent(bootstrapResult.Infra)
+	renderIPsecDaemonSet = (isOVNIPsecActive && (isCNOIPsecMachineConfigPresent || isUserDefinedIPsecMachineConfigPresent ||
+		isHypershiftHostedCluster)) || mode == operv1.IPsecModeFull
```
jcaamano, line 623: can you rather tweak isOVNIPsecActive above?
```
isOVNIPsecActiveOrRollingOut := bootstrapResult.OVN.IPsecUpdateStatus != nil && bootstrapResult.OVN.IPsecUpdateStatus.OVNRolloutInProgress
// when OVN is rolling out, OVN IPsec might be fully or partially active or inactive.  If MachineConfigs are not present, we know its inactive since we only stop rendering them once inactive.
isOVNIPsecActive := isOVNIPsecActiveOrRollingOut && (isCNOIPsecMachineConfigPresent || isUserDefinedIPsecMachineConfigPresent || isHypershiftHostedCluster)

```

PR #2641, opened 2025-02-05
OCPBUGS-49662: Prevent resetting masquerade subnet

arghosh93:
Prevent resetting masquerade subnet to default value(169.254.169.0/29)
set at upstream ovn-kubernetes when ovnkube-node daemonset is removed
from the cluster.
This PR have an additional commit to update masquerade subnet to
ovnkube-control-plane deployment. This is required as masquerade
subnet is currently configurable at day 2.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -1306,7 +1306,7 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
 	}
 
 	// set the default masquerade CIDR for new clusters while ignoring upgrades
-	if res.ControlPlaneUpdateStatus == nil && res.NodeUpdateStatus == nil {
+	if res.ControlPlaneUpdateStatus == nil || res.NodeUpdateStatus == nil {
 		klog.Infof("Configuring the default masquerade subnets to %q and %q", defaultV4MasqueradeSubnet, defaultV6MasqueradeSubnet)
 		res.DefaultV4MasqueradeSubnet = defaultV4MasqueradeSubnet
 		res.DefaultV6MasqueradeSubnet = defaultV6MasqueradeSubnet
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
@@ -171,6 +171,15 @@ spec:
             ovn_v6_transit_switch_subnet_opt="--cluster-manager-v6-transit-switch-subnet {{.V6TransitSwitchSubnet}}"
           fi
 
+          ovn_v4_masquerade_subnet_opt=
+          if [[ "{{.V4MasqueradeSubnet}}" != "" ]]; then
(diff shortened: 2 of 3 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -1306,7 +1306,7 @@ func bootstrapOVN(conf *operv1.Network, kubeClient cnoclient.Client, infraStatus
 	}
 
 	// set the default masquerade CIDR for new clusters while ignoring upgrades
-	if res.ControlPlaneUpdateStatus == nil && res.NodeUpdateStatus == nil {
+	if res.ControlPlaneUpdateStatus == nil || res.NodeUpdateStatus == nil {
```
kyrtapz, line 1309: What if someone removes both the controlplane and the dataplane components?

arghosh93: If I remove both ovnkube-node DS and ovnkube-control-plane deployment then also masquerade subnet does not get reverted back to old default subnets.
Do you anticipate this change may override old default masquerade subnets during a upgrade? Then I would say this change is not sufficient to address the issue.

arghosh93: I have tried to upgrade from one version to other(4.17-4.18) and I dont see ovnkube-node DS and ovnkube-control-plane deployment being deleted and instead all images get changed.

arghosh93: @kyrtapz should we move forward with this change?

kyrtapz: Upgrades are not removing the daemonset either. I thought this fix is to address the corner case of users removing the ovn-k daemonset?

arghosh93: yes, correct. Masquerade subnet gets reset when either ovnkube-node daemonset or ovnkube-control-plane deployment is removed manually.

kyrtapz:
/lgtm

NOTE: This PR does not address the case where both the control-plane and the dataplane components were removed.

arghosh93:
@asood-rh can you please run a pre-merge test with this PR? Following is what we need to test:

- Create a cluster(>4.18) and verify whether masquerade subnet is set to 169.254.0.0/17 or not.
- Remove ovnkube-node daemonset and verify whether masquerade subnet remains same or not.
- Remove ovnkube-control-plane deployment and verify whether masquerade subnet remains same or not.
- Remove both aforementioned deployment and daemonset and verify whether masquerade subnet remains same or not.

The above test also makes sense after changing masquerade subnet at day 2.
- Change masquerade subnet to 169.254.0.0/20.
- Remove ovnkube-node daemonset and verify whether masquerade subnet remains same or not.
- Remove ovnkube-control-plane deployment and verify whether masquerade subnet remains same or not.
- Remove both aforementioned deployment and daemonset and verify whether masquerade subnet remains same or not.

Regarding upgrades. We are checking those in CI.

arghosh93:
/override ci/prow/4.19-upgrade-from-stable-4.18-e2e-azure-ovn-upgrade
It's not a required job, but tide is still complaining about this one, it seems.

tssurya:
/override ci/prow/4.19-upgrade-from-stable-4.18-e2e-azure-ovn-upgrade

since this PR is very old its stuck on the older version jobs, overriding this to get this in

PR #2644, opened 2025-02-12
[release-4.18] OCPBUGS-50582: Graceful cleanup of IPsec states

pperiyasamy:
This PR does the backport of following fixes to prevent unnecessary ipsec service restart, ip xfrm state policy cleanups while bringing up ipsec-host pod. This would potentially avoid reestablishment of IKE SAs during ipsec pod restarts and let OVN networking pods traffic go on without any packet drops.

1. There is an incorrect check in ipsec pod clean up logic which removes `/etc/ipsec.d/openshift.conf` file, ip xfrm state and policy entries in all cases, but these must be removed only when ipsec mode is changed from full to external or disabled.
2. We don't need narrowing=yes option to be set explicitly anymore because system default crypto policies are commented out now, otherwise `TS_UNACCEPTABLE` error is seen temporarily at the time of ipsec service restart.
3. The IPsec service restart is needed only at the time of specific IPsec config changes, so doing it only at the time commenting out default crypto-policies conf file.
4. After IPsec is deployed, when machine config pools goes into progressing/degraded state while processing some other machine machine configs (or) node reboot scenarios, CNO deletes IPsec pods which is known behavior as per changes in the commit 6aa2f0fbf2b06ce52db1a2a039bbdbcaf588c409, but it also accidentally disabling IPsec in OVN which is not an expected behavior. This causes ovs-monitor-ipsec to refresh existing ipsec connections unnecessarily when IPsec pod comes up (once machine config pool settles) as it is not able to find remote_name from the tunnel.
5.  It also removes dead code related to IPsec 4.13 upgrade which is no longer valid for upgrade scenarios beyond >= 4.15. This change is intended to be backported until 4.15 so still keeping legacy upgrade (4.14->4.15) scenarios.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/80-ipsec-master-extensions.yaml
+++ b/bindata/network/ovn-kubernetes/common/80-ipsec-master-extensions.yaml
@@ -20,6 +20,7 @@ spec:
 
          [Service]
          Type=oneshot
+         ExecStartPre=rm -f /etc/ipsec.d/cno.conf
          ExecStart=systemctl enable --now ipsec.service
 
          [Install]
--- a/bindata/network/ovn-kubernetes/common/80-ipsec-worker-extensions.yaml
+++ b/bindata/network/ovn-kubernetes/common/80-ipsec-worker-extensions.yaml
@@ -20,6 +20,7 @@ spec:
 
          [Service]
          Type=oneshot
+         ExecStartPre=rm -f /etc/ipsec.d/cno.conf
          ExecStart=systemctl enable --now ipsec.service
 
          [Install]
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -407,7 +407,7 @@ spec:
           # When east-west ipsec is not disabled, then do not flush xfrm states and
           # policies in order to maintain traffic flows during container restart.
           ipsecflush() {
-            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] || \
+            if [ "$(kubectl get networks.operator.openshift.io cluster -ojsonpath='{.spec.defaultNetwork.ovnKubernetesConfig.ipsecConfig.mode}')" != "Full" ] && \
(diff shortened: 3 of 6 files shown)


huiran0826:
Following verification steps in https://issues.redhat.com/browse/OCPBUGS-50616 and pre-merge testing result looks good.
/label qe-approved

trozet:
/hold

until 2454 merges

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4926

pperiyasamy:
/testwith openshift/cluster-network-operator/release-4.18/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4926

huiran0826:
Pre-merge tested https://github.com/openshift/cluster-network-operator/pull/2644 and https://github.com/openshift/machine-config-operator/pull/4926 together.
1. Verified upgrade from 4.17-> 4.18 PR with worker mcp paused, upgrade successfully.
2. Verified upgrade from 4.18.5-> 4.18 PR without mcp paused, upgrade successfully.
3. Verified EUS upgrade from 4.16->4.17 PR-> 4.18 PR, upgrade successfully.
4. Enable IPSec/Disable IPsec was tested and worked as expected.
5. Rebooting node, network co will render mcp status 
6. full-> external, external-> full, mode switch worked.

pperiyasamy:
> /testwith openshift/cluster-network-operator/release-4.18/e2e-aws-ovn-ipsec-upgrade [openshift/machine-config-operator#4926](https://github.com/openshift/machine-config-operator/pull/4926)

This ipsec upgrade [run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cluster-network-operator-2644-openshift-machine-config-operator-4926-e2e-aws-ovn-ipsec-upgrade/1903010132288081920) failed with disruptive events, it happened because in 4.18, reboot happening twice for the nodes during the upgrade, the 1st reboot happened even before mcp changes [#4926](https://github.com/openshift/machine-config-operator/pull/4926) are in place on the node. so these events are expected (the event timings also matches with 1st reboot timings).

pperiyasamy:
> /label acknowledge-critical-fixes-only

this is a critical fix and must be backported into 4.15 ASAP for a customer bug https://issues.redhat.com/browse/OCPBUGS-36688,

PR #2646, opened 2025-02-12
[DNM] test rbac for whereabouts nodeslice controller

dougbtv:


dougbtv:
```
$ cat /tmp/nad.yaml 
apiVersion: "k8s.cni.cncf.io/v1"
metadata:
  name: wb-ipam-two
  namespace: openshift-multus
spec:
  config: '{
    "cniVersion": "0.3.0",
    "name": "wb-ipam-two-cni-name",
    "type": "bridge",
    "bridge": "cni0",
    "ipam": {
      "type": "whereabouts",
      "range": "10.5.0.0/20",
      "node_slice_size": "/24"
    }
  }'
$ cat /tmp/pod.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: samplepod2
  annotations:
    k8s.v1.cni.cncf.io/networks: openshift-multus/wb-ipam-two
spec:
  containers:
  - name: samplepod
    command: ["/bin/ash", "-c", "trap : TERM INT; sleep infinity & wait"]
    image: alpine
```

dougbtv:
"works" in current status, but has a DRY problem with the whereabouts flatfile config being repeated.

PR #2651, opened 2025-02-20
[Release 4.16] OCPBUGS-50712: Not update status.migration of the network.config CR to empty

pliurh:
If spec.migration.features of the network.operator CR, CNO will update the status.migration of the network.config CR to an empty value '{}'. It prevents CNO from proceeding with live migration. This shall be fixed in CNO.

Only update status.migration.network of the network.config CR when the spec.migration.network of network.operator CR is set.

### DIFF
--- a/pkg/controller/clusterconfig/migration.go
+++ b/pkg/controller/clusterconfig/migration.go
@@ -103,7 +103,7 @@ func (r *ReconcileClusterConfig) prepareOperatorConfigForNetworkTypeMigration(ct
 		return nil
 	}
 
-	if !cniReady && clusterConfig.Status.Migration == nil {
+	if !cniReady && (clusterConfig.Status.Migration == nil || clusterConfig.Status.Migration.NetworkType == "") {
 		klog.Infof("step-1: deploy target CNI: %s", clusterConfig.Spec.NetworkType)
 		operConfig.Spec.Migration = &operv1.NetworkMigration{
 			Mode:        operv1.LiveNetworkMigrationMode,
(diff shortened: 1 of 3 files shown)


martinkennelly (approved):
/lgtm

martinkennelly:
added to team agenda for approver

martinkennelly:
What about 4.17? not needed cause no sdn?

pliurh:
> What about 4.17? not needed cause no sdn?

Yes. We shall remove the migration code from CNO 4.17 onwards.

pliurh:
> Why not remove that code from later releases first then ?

We have a story for this work. But It is not the priority right now.

pkg/controller/clusterconfig/migration.go
```diff
@@ -103,7 +103,7 @@ func (r *ReconcileClusterConfig) prepareOperatorConfigForNetworkTypeMigration(ct
 		return nil
 	}
 
-	if !cniReady && clusterConfig.Status.Migration == nil {
+	if !cniReady && (clusterConfig.Status.Migration == nil || clusterConfig.Status.Migration.NetworkType == "") {
```
kyrtapz, line 106: Is `clusterConfig.Status.Migration.NetworkType == ""` still possible with https://github.com/openshift/cluster-network-operator/pull/2651/files#diff-0490e0f88e14c4f85aff51f6745b80b3bdfb7867b5b6bbb890945cac20cec481R362?

pliurh: No, it's solely for safety.

kyrtapz:
@pliurh Since we don't block it on the API level, wouldn't it make sense to add a CNO pre-flight check for this?

pliurh:
> @pliurh Since we don't block it on the API level, wouldn't it make sense to add a CNO pre-flight check for this?

This issue happened in the `status`. When the user set all the boolean fields under `spec.migration.features` to false, which is not the default value.  Since users normally won't change status fields. I think this patch is sufficient.

PR #2653, opened 2025-02-25
[Release 4.15] OCPBUGS-51208: Use applyconfigurations for updating network.oprerator status

pliurh:
Use applyconfigurations to avoid setting any fields that are not managed by the status-manager.

This is a manual cherry-pick of #2413 by resolving merge conflict in `status_manager_test.go`.


pperiyasamy:
/assign @kyrtapz 

This is needed for https://github.com/openshift/cluster-network-operator/pull/2383 when it lands in 4.15 which we plan to do it in coming days for a customer bug.

PR #2657, opened 2025-03-02
OCPBUGS-48305: frr-k8s: align manifests with operator

oribon:
cherry-pick from 4.19 https://github.com/openshift/cluster-network-operator/pull/2615, without the unnumbered api changes that are not supported in 4.18


fedepaol:
The frrk8s job failed with two tests (vrf leaking) because we didn't backport lowering the vrf id yet.

/approve
/lgtm

PR #2662, opened 2025-03-10
CORENET-5972: Consume openvswitch-ipsec systemd service for OVN IPsec deployment

pperiyasamy:
This ipsec machine config extension is now installing `openvswitch3.5-ipsec` package on the node, so this PR consumes that package to configure, enable and start `openvswitch-ipsec` systemd service which basically moves away running ovs-monitor-ipsec process from container to host. 

It fixes following issues.

1. Getting rid of running host binaries inside ovn-ipsec container, for example `addconn` which depends on `glibc` which creates version incompatibility issues between host and container. 
2. There is no need for patching `auto=start` parameter into each IPsec connections which is introduced by the PR https://github.com/openshift/machine-config-operator/pull/4854.
3. The ovs-monitor-ipsec is now solving auth problem while using default crypto policies loaded by libreswan on the host.

In order to consume openvswitch-ipsec systemd service, this PR does the following:

1. Stop spawning ovs-monitor-ipsec as foreground process in the ovn-ipsec container. Instead setup required IPsec configuration parameters in the /etc/sysconfig/openvswitch file, enable and start the openvswitch-ipsec service
on the host. This is done at the time when ovn-ipsec-host pod is coming up for the first time. For the pod restart scenarios, it just checks openvswitch-ipsec service is running on the host, otherwise exit from the container with error.

2. Keep running an ovn-ipsec container that waits for the ovs-monitor-ipsec.log file to be created, monitors the openvswitch-ipsec service health, and redirects /var/log/openvswitch/ovs-monitor-ipsec.log to the ovn-ipsec container's stdout console. The container fails if the systemd service stops unexpectedly, ensuring
proper pod lifecycle management.

4. There is no necessity of doing ipsec state and policy cleanup in ovn-ipsec-cleanup container when OVN IPsec is handled via openvswitch-ipsec systemd service.

5. During the OCP upgrade, the new ipsec os extension takes a while to deploy with openvswitch3.5-ipsec package, so by the time ovn-ipsec-host daemonset is rendered, we need to handle that scenario by running ovs-monitor-ipsec in the container. So this PR is also considering the transition phase of the process that is moving
from container to host.

6. The ovn-keys init container configures ovs with IPsec certificate paths, so the container uses same host directory path to store and configure ovs with certificates because the ovs-monitor-ipsec process is running on the host now.

/assign @igsilya 

## Release Notes

* **Bug Fixes**
  * Improved IPsec certificate/key regeneration so the IPsec service restarts more reliably.
  * Enhanced coordination between the IPsec components in the pod and the host IPsec service for smoother startup and clean shutdown.

* **Refactor**
  * Updated IPsec asset storage and runtime configuration to use host-backed paths for more consistent behavior.
  * Refined volume mounts and command checks so health/liveness operations rely on the host IPsec tooling when available.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -109,8 +109,9 @@ spec:
           # authentic.
           echo "Configuring IPsec keys"
 
-          cert_pem=/etc/openvswitch/keys/ipsec-cert.pem
+          cert_pem=/var/lib/openvswitch/etc/keys/ipsec-cert.pem
 
+          ovs_ipsec_restart_needed="false"
           # If the certificate does not exist or it will expire in the next 6 months
           # (15770000 seconds), we will generate a new one.
           if ! openssl x509 -noout -dates -checkend 15770000 -in $cert_pem; then
@@ -118,25 +119,25 @@ spec:
             # is a requirement by OVN.
             cn=$(ovs-vsctl --retry -t 60 get Open_vSwitch . external-ids:system-id | tr -d "\"")
 
-            mkdir -p /etc/openvswitch/keys
+            mkdir -p /var/lib/openvswitch/etc/keys
 
             # Generate an SSL private key and use the key to create a certitificate signing request
-            umask 077 && openssl genrsa -out /etc/openvswitch/keys/ipsec-privkey.pem 2048
+            umask 077 && openssl genrsa -out /var/lib/openvswitch/etc/keys/ipsec-privkey.pem 2048
             if ! openssl req -new -text \
                         -extensions v3_req \
                         -addext "subjectAltName = DNS:${cn}" \
                         -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
-                        -key /etc/openvswitch/keys/ipsec-privkey.pem \
-                        -out /etc/openvswitch/keys/ipsec-req.pem; then
+                        -key /var/lib/openvswitch/etc/keys/ipsec-privkey.pem \
+                        -out /var/lib/openvswitch/etc/keys/ipsec-req.pem; then
               echo "openssl req failed with -extensions v3_req, retrying without it"
               openssl req -new -text \
                           -addext "subjectAltName = DNS:${cn}" \
                           -subj "/C=US/O=ovnkubernetes/OU=kind/CN=${cn}" \
-                          -key /etc/openvswitch/keys/ipsec-privkey.pem \
-                          -out /etc/openvswitch/keys/ipsec-req.pem
+                          -key /var/lib/openvswitch/etc/keys/ipsec-privkey.pem \
+                          -out /var/lib/openvswitch/etc/keys/ipsec-req.pem
             fi
 
-            csr_64=$(base64 -w0 /etc/openvswitch/keys/ipsec-req.pem) # -w0 to avoid line-wrap
+            csr_64=$(base64 -w0 /var/lib/openvswitch/etc/keys/ipsec-req.pem) # -w0 to avoid line-wrap
 
             # Request that our generated certificate signing request is
             # signed by the "network.openshift.io/signer" signer that is
@@ -176,7 +177,10 @@ spec:
             # kubectl delete csr/$(hostname)
 
             # Get the CA certificate so we can authenticate peer nodes.
-            openssl x509 -in /signer-ca/ca-bundle.crt -outform pem -text -out /etc/openvswitch/keys/ipsec-cacert.pem
+            openssl x509 -in /signer-ca/ca-bundle.crt -outform pem -text -out /var/lib/openvswitch/etc/keys/ipsec-cacert.pem
+
+            # New certs are generated, so openvswitch-ipsec service may need restart to load new certificates.
+            ovs_ipsec_restart_needed="true"
           fi
 
           # Configure OVS with the relevant keys for this node. This is required by ovs-monitor-ipsec.
@@ -185,8 +189,13 @@ spec:
           # the will get read and loaded into NSS by the ovs-monitor-ipsec process
           # which has not started yet.
           ovs-vsctl --retry -t 60 set Open_vSwitch . other_config:certificate=$cert_pem \
-                                                     other_config:private_key=/etc/openvswitch/keys/ipsec-privkey.pem \
-                                                     other_config:ca_cert=/etc/openvswitch/keys/ipsec-cacert.pem
+                                                     other_config:private_key=/var/lib/openvswitch/etc/keys/ipsec-privkey.pem \
+                                                     other_config:ca_cert=/var/lib/openvswitch/etc/keys/ipsec-cacert.pem
+
+          # Restart openvswitch-ipsec systemd service when OVS is configured with new certificate when old one expired.
+          if grep -q "openshift.conf" /etc/sysconfig/openvswitch && [ "$ovs_ipsec_restart_needed" = "true" ]; then
+            chroot /proc/1/root systemctl restart openvswitch-ipsec
+          fi
         env:
         - name: K8S_NODE
           valueFrom:
@@ -203,8 +212,8 @@ spec:
           name: host-var-run
         - mountPath: /signer-ca
           name: signer-ca
-        - mountPath: /etc/openvswitch
-          name: etc-openvswitch
+        - mountPath: /var/lib
+          name: host-var-lib
         - mountPath: /etc
           name: host-etc
         resources:
@@ -247,6 +256,38 @@ spec:
           if ! pgrep pluto; then
             echo "pluto is not running, enable the service and/or check system logs"
             exit 2
+          fi
+
+          # Configure IPsec using the host's openvswitch-ipsec systemd service if present.
+          # This avoids running host binaries from the container and solves compatibility issues.
+          # Fall back to container-based ovs-monitor-ipsec when the openvswitch-ipsec systemd service
+          # is not yet available on the node.
+          ovsipsecservice="openvswitch-ipsec"
+          if chroot /proc/1/root systemctl list-unit-files --type=service | grep "$ovsipsecservice" &> /dev/null; then
+            if ! grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+              sed -i 's|OPTIONS=\"\"|OPTIONS=\"--no-restart-ike-daemon --ovs-monitor-ipsec-options='\''--ipsec-conf=\/etc\/ipsec.d\/openshift.conf --root-ipsec-conf=\/etc\/ipsec.conf --ipsec-d=\/var\/lib\/ipsec\/nss --use-default-crypto'\''\"|' /etc/sysconfig/openvswitch
+              if ! grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+                echo "failed to configure /etc/sysconfig/openvswitch"
+                exit 1
+              fi
+              chroot /proc/1/root systemctl enable --now $ovsipsecservice
+            fi
+
+            # Capture PID and InvocationID atomically, verify no restart occurred
+            invocation_id=$(chroot /proc/1/root systemctl show -p InvocationID --value openvswitch-ipsec)
+            main_pid=$(chroot /proc/1/root systemctl show -p MainPID --value openvswitch-ipsec)
+            file_pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
+            if [ "$main_pid" != "$file_pid" ]; then
+              echo "openvswitch-ipsec PID mismatch: systemd reports $main_pid, PID file has $file_pid"
+              exit 1
+            fi
+            echo "$main_pid" > /tmp/ovs-monitor-ipsec-main.pid
+
+            while true; do
+              chroot /proc/1/root journalctl -f "_SYSTEMD_INVOCATION_ID=$invocation_id" -n all
+              echo "journalctl exited unexpectedly, retrying..."
+              sleep 2
+            done
           fi
 
           # The ovs-monitor-ipsec doesn't set authby, so when it calls ipsec auto --start
@@ -305,6 +346,10 @@ spec:
                  - |
                    #!/bin/bash
                    set -exuo pipefail
+                   if grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+                     echo "openvswitch-ipsec service is configured on the host, no action needed in pre stop"
+                     exit 0
+                   fi
                    # In order to maintain traffic flows during container restart, we
                    # need to ensure that xfrm state and policies are not flushed.
 
@@ -332,8 +377,6 @@ spec:
           name: host-var-run
         - mountPath: /var/log/openvswitch/
           name: host-var-log-ovs
-        - mountPath: /etc/openvswitch
-          name: etc-openvswitch
         - mountPath: /var/lib
           name: host-var-lib
         - mountPath: /etc
@@ -360,9 +403,24 @@ spec:
                 exit 0
               fi
 {{ end }}
-              if [[ $(ipsec whack --trafficstatus | wc -l) -eq 0 ]]; then
-                echo "no ipsec traffic configured"
-                exit 10
+              if grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+                if ! chroot /proc/1/root systemctl is-active --quiet openvswitch-ipsec; then
+                  echo "openvswitch-ipsec service is not active"
+                  exit 1
+                fi
+                if [ -f /tmp/ovs-monitor-ipsec-main.pid ]; then
+                  file_pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
+                  main_pid=$(cat /tmp/ovs-monitor-ipsec-main.pid)
+                  if [ "$main_pid" != "$file_pid" ]; then
+                    echo "openvswitch-ipsec service restarted (PID changed from $main_pid to $file_pid)"
+                    exit 1
+                  fi
+                fi
+              else
+                if [[ $(ipsec whack --trafficstatus | wc -l) -eq 0 ]]; then
+                  echo "no ipsec traffic configured"
+                  exit 10
+                fi
               fi
           initialDelaySeconds: 15
           periodSeconds: 60
@@ -440,7 +498,15 @@ spec:
 
           # Function to handle SIGTERM
           cleanup() {
-            echo "received SIGTERM, flushing ipsec config"
+            echo "received SIGTERM, flushing ipsec config if required"
+            # The openvswitch-ipsec service is started by this pod. When active, it updates
+            # /etc/ipsec.d/openshift.conf for IPsec connection changes, and the pluto daemon
+            # manages xfrm state and policy entries accordingly. Skip flushing IPsec state
+            # on exit.
+            if grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+              echo "openvswitch-ipsec service is configured on the host, no ipsec flush needed"
+              exit 0
+            fi
             # Wait upto 15 seconds for ovs-monitor-ipsec process to terminate before
             # cleaning up ipsec entries.
             counter=0
@@ -472,14 +538,19 @@ spec:
           done
           echo "ovs-monitor-ipsec is started"
 
-          # Monitor the ovs-monitor-ipsec process.
-          while kill -0 "$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)"; do
-            sleep 1
-          done
+          # When ovs-monitor-ipsec process is not running via openvswitch-ipsec systemd
+          # service, then keep monitoring ovn-ipsec container process until it terminates
+          # in the ovn-ipsec container's pre stop hook.
+          if ! grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+            # Monitor the ovs-monitor-ipsec process.
+            while kill -0 "$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)"; do
+              sleep 1
+            done
 
-          # Once the ovs-monitor-ipsec process terminates, execute the cleanup command.
-          echo "ovs-monitor-ipsec is terminated, flushing ipsec config"
-          ipsecflush
+            # Once the ovs-monitor-ipsec process terminates, execute the cleanup command.
+            echo "ovs-monitor-ipsec is terminated, flushing ipsec config"
+            ipsecflush
+          fi
 
           # Continue running until SIGTERM is received (or exit naturally)
           while true; do
@@ -518,10 +589,6 @@ spec:
           defaultMode: 420
           name: signer-ca
         name: signer-ca
-      - hostPath:
-          path: /var/lib/openvswitch/etc
-          type: DirectoryOrCreate
-        name: etc-openvswitch
       - hostPath:
           path: "{{.CNIConfDir}}"
         name: host-cni-netd
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -499,7 +499,8 @@ spec:
             # Wait upto 15 seconds for ovs-monitor-ipsec process to terminate before
             # cleaning up ipsec entries.
             counter=0
-            while kill -0 "$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)"; do
+            pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
+            while [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; do
               counter=$((counter+1))
               sleep 1
               if [ $counter -gt 15 ];
@@ -532,7 +533,8 @@ spec:
           # in the ovn-ipsec container's pre stop hook.
           if ! grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
             # Monitor the ovs-monitor-ipsec process.
-            while kill -0 "$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)"; do
+            pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
+            while [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; do
               sleep 1
             done
 
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -326,8 +326,6 @@ spec:
           if [ -e /usr/libexec/ipsec/_stackmanager ]; then
             /usr/libexec/ipsec/_stackmanager start
           fi
-          # Check nss database status
-          /usr/sbin/ipsec --checknss
 
           # Start ovs-monitor-ipsec which will monitor for changes in the ovs
           # tunnelling configuration (for example addition of a node) and configures
@@ -381,8 +379,6 @@ spec:
           name: host-var-lib
         - mountPath: /etc
           name: host-etc
-        - mountPath: /usr/sbin
-          name: usr-sbin
         - mountPath: /usr/libexec
           name: usr-libexec
         resources:
@@ -417,8 +413,10 @@ spec:
                   fi
                 fi
               else
-                if [[ $(ipsec whack --trafficstatus | wc -l) -eq 0 ]]; then
-                  echo "no ipsec traffic configured"
+                # Check if ovs-monitor-ipsec is running and has configured IPsec tunnels
+                status=$(ovs-appctl -t ovs-monitor-ipsec ipsec/status 2>/dev/null)
+                if [ -z "$status" ] || [ "$status" = "{}" ]; then
+                  echo "no ipsec tunnels configured"
                   exit 10
                 fi
               fi
@@ -604,10 +602,6 @@ spec:
           path: /etc
           type: Directory
         name: host-etc
-      - hostPath:
-          path: /usr/sbin
-          type: Directory
-        name: usr-sbin
       - hostPath:
           path: /usr/libexec
           type: Directory
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -320,13 +320,6 @@ spec:
           # Workaround for https://github.com/libreswan/libreswan/issues/373
           ulimit -n 1024
 
-          /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules only for libreswan version <= 5.2. The _stackmanager binary is
-          # removed from 5.3 onwards, so this check is not needed on later versions.
-          if [ -e /usr/libexec/ipsec/_stackmanager ]; then
-            /usr/libexec/ipsec/_stackmanager start
-          fi
-
           # Start ovs-monitor-ipsec which will monitor for changes in the ovs
           # tunnelling configuration (for example addition of a node) and configures
           # libreswan appropriately.
@@ -379,8 +372,6 @@ spec:
           name: host-var-lib
         - mountPath: /etc
           name: host-etc
-        - mountPath: /usr/libexec
-          name: usr-libexec
         resources:
           requests:
             cpu: 10m
@@ -602,10 +593,6 @@ spec:
           path: /etc
           type: Directory
         name: host-etc
-      - hostPath:
-          path: /usr/libexec
-          type: Directory
-        name: usr-libexec
       tolerations:
       - operator: "Exists"
 {{end}}


pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854 https://github.com/openshift/os/pull/1718 https://github.com/openshift/machine-config-operator/pull/4878 https://github.com/openshift/ovn-kubernetes/pull/2472

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854 https://github.com/openshift/os/pull/1718 https://github.com/openshift/machine-config-operator/pull/4878 https://github.com/openshift/ovn-kubernetes/pull/2472

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854 https://github.com/openshift/os/pull/1718 https://github.com/openshift/machine-config-operator/pull/4878 https://github.com/openshift/ovn-kubernetes/pull/2472

pperiyasamy:
The `testwith` doesn't take https://github.com/openshift/os/pull/1718 changes, ipsec os extension is still not installing `openvswitch-ipsec` package, so we must get 1718 landed first for testing CNO changes.

igsilya:
@pperiyasamy I agree, we should get OVS 3.5 first into rhcos / ovn-k / microshift.  We can install openvswitch3.5-ipsec at the same time, it should not be a problem since the service is disabled until CNO activates it.  We need OVS 3.5 either way for other purposes (rhel 10 support, for example).   Once we have OVS 3.5 and the openvswitch-ipsec service we can more easily test CNO and other changes.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4854 https://github.com/openshift/os/pull/1718 https://github.com/openshift/machine-config-operator/pull/4878 https://github.com/openshift/ovn-kubernetes/pull/2472

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade openshift/os#1718 openshift/machine-config-operator#4878 openshift/ovn-kubernetes#2472

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade openshift/ovn-kubernetes#2472 openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/ovn-kubernetes/pull/2472 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
The run https://github.com/openshift/cluster-network-operator/pull/2662/checks?check_run_id=44182797463 is failing at cluster installation with few cluster operators in degraded state including machine config, master mcp in degraded state, 

```
2025-06-16T16:04:29.143452024Z I0616 16:04:29.143424    2754 daemon.go:2230] Validating against current config rendered-master-e97d82c84329b93800782246bdc1ecd5
2025-06-16T16:04:29.143770751Z I0616 16:04:29.143751    2754 daemon.go:2142] SSH key location ("/home/core/.ssh/authorized_keys.d/ignition") up-to-date!
2025-06-16T16:04:29.143770751Z I0616 16:04:29.143767    2754 image_manager_helper.go:92] Running captured: rpm-ostree kargs
2025-06-16T16:04:29.195997188Z I0616 16:04:29.195952    2754 update.go:2789] "Validated on-disk state"
2025-06-16T16:04:29.199894354Z I0616 16:04:29.199867    2754 daemon.go:2341] System state unchanged: MachineConfig: rendered-master-e97d82c84329b93800782246bdc1ecd5
2025-06-16T16:04:56.210717425Z E0616 16:04:56.210660    2754 writer.go:231] Marking Degraded due to: "could not set drain annotation: unable to update node \"ip-10-0-57-152.us-east-2.compute.internal\": etcdserver: request timed out"
2025-06-16T16:05:23.224348638Z E0616 16:05:23.224314    2754 writer.go:247] Error setting Degraded annotation for node ip-10-0-57-152.us-east-2.compute.internal: unable to update node "ip-10-0-57-152.us-east-2.compute.internal": etcdserver: request timed out
2025-06-16T16:05:23.224406540Z E0616 16:05:23.224391    2754 daemon.go:620] Could not update annotation: unable to update node "ip-10-0-57-152.us-east-2.compute.internal": etcdserver: request timed out
2025-06-16T16:05:23.225977866Z I0616 16:05:23.225959    2754 daemon.go:2032] Running: /run/machine-config-daemon-bin/nmstatectl persist-nic-names --root / --kargs-out /tmp/nmstate-kargs800206539 --cleanup
2025-06-16T16:05:23.227468209Z [2025-06-16T16:05:23Z INFO  nmstatectl] Nmstate version: 2.2.45
2025-06-16T16:05:23.227525150Z
```

This can't be an issue related to the changes. triggering the job again. Since [OCPBUGS-56555](https://issues.redhat.com/browse/OCPBUGS-56555) is fixed now, the upgrade run must use `openvswitch-ipsec systemd` service after upgrade is complete.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

igsilya:
The change looks correct to me from the OVS and service ordering perspective, also tried it with cluster bot together with https://github.com/openshift/machine-config-operator/pull/4878 and it worked as expected (didn't test upgrades).
/lgtm
/retest

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

jluhrsen:
/testwith openshift/cluster-network-operator/master/5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

pperiyasamy:
Two issues being noticed with e2e-aws-ovn-upgrade-ipsec run:
1. it uses build (which broke ipsec due to glibc incompatibility issue ) for the initial cluster, so ovn-ipsec-host pod is crashlooping until pod is upgraded to PR version.
2.  machine config daemon never deployed openvswitch3.5-ipsec package during upgrade and failing with below error.
```
2026-06-24T18:11:11.339388625Z W0624 18:11:11.339342   48949 update.go:1946] Extension package openvswitch3.5-ipsec not found in RPM database
2026-06-24T18:11:11.339444737Z E0624 18:11:11.339419   48949 writer.go:231] Marking Degraded due to: "extension package verification failed: the following extension packages are missing from the RPM database: [openvswitch3.5-ipsec]"
2026-06-24T18:11:11.392881358Z I0624 18:11:11.392825   48949 daemon.go:643] Error syncing node ip-10-0-16-239.ec2.internal (retries 2): extension package verification failed: the following extension packages are missing from the RPM database: [openvswitch3.5-ipsec]
2026-06-24T18:11:13.353789060Z I0624 18:11:13.353753   48949 daemon.go:796] Transitioned from state: Done -> Degraded

```
We had bug https://redhat.atlassian.net/browse/OCPBUGS-56555 for similar issue before, but looks like it's not fixed completely or don't know if it's a rhcos image issue for the ocp upgrade. will check with mco team.

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/release-5.0/5.0-upgrade-from-stable-4.22-e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

jluhrsen:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec openshift/machine-config-operator#4878

jluhrsen:
/testwith openshift/cluster-network-operator/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

jluhrsen:
/testwith openshift/cluster-network-operator/release-5.0/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

jluhrsen:
/testwith openshift/cluster-network-operator/master/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec openshift/ovn-kubernetes#3254 openshift/machine-config-operator#4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec openshift/ovn-kubernetes#3254 openshift/machine-config-operator#6250

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec openshift/machine-config-operator#6250

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/ovn-kubernetes/pull/3254 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/ovn-kubernetes/pull/3254 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -305,6 +364,10 @@ spec:
                  - |
                    #!/bin/bash
                    set -exuo pipefail
+                   if chroot /proc/1/root systemctl is-active --quiet openvswitch-ipsec; then
```
jcaamano: Why are our checks inconsistent? On some places we check if the service is defined, but here and in other places we check if the service is active? I might be missing something here.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,56 @@ spec:
...
+              sleep 1
+            done
+
+            # Tail the log
+            while true; do
+              if ! chroot /proc/1/root systemctl is-active --quiet $ovsipsecservice; then
+                echo "$ovsipsecservice service stopped unexpectedly"
+                exit 1
+              fi
+              tail -F /var/log/openvswitch/ovs-monitor-ipsec.log
```
jcaamano: I think `tail -F` is pretty resilient and wont fail if ipsec service stops. Have you tested it?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,56 @@ spec:
...
+              fi
+              sleep 1
+            done
+
+            # Tail the log
+            while true; do
+              if ! chroot /proc/1/root systemctl is-active --quiet $ovsipsecservice; then
+                echo "$ovsipsecservice service stopped unexpectedly"
+                exit 1
+              fi
```
jcaamano: Wouldn't it be simpler to just have a subshell periodically checking if the service is active and sending a SIGTERM to the parent shell when not?

Also, I am not sure how tight you want to align the state of the service to the state of the pod. The service will restart by itself and you might miss it.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,56 @@ spec:
             exit 2
           fi
 
+          # Configure IPsec using the host's openvswitch-ipsec systemd service if present.
+          # This avoids running host binaries from the container and solves compatibility issues.
+          # Fall back to container-based ovs-monitor-ipsec for older nodes without the service.
```
jcaamano: Why keep the fallback? Is not working in all circumstances anyway, is it?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -440,7 +501,15 @@ spec:
 
           # Function to handle SIGTERM
           cleanup() {
-            echo "received SIGTERM, flushing ipsec config"
+            echo "received SIGTERM, flushing ipsec config if required"
+            # When IPsec connections are managed by openvswitch-ipsec systemd service,
+            # It means ovs-monitor-ipsec process is running on the host always, so ip
+            # xfrm state and policy entries are managed properly according to ipsec
+            # configuration on the OVN. so skip flushing ipsec for this case.
+            if chroot /proc/1/root systemctl is-active --quiet openvswitch-ipsec; then
```
jcaamano:
```
When IPsec connections are managed by openvswitch-ipsec systemd service
```
When is this? When `openvswitch-ipsec` is running, or when it is present? It should be a static configuration, so it should be when it is present even if it is not running which would indicate a different circumstance?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -338,13 +338,6 @@ spec:
           # Workaround for https://github.com/libreswan/libreswan/issues/373
           ulimit -n 1024
 
-          /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules only for libreswan version <= 5.2. The _stackmanager binary is
-          # removed from 5.3 onwards, so this check is not needed on later versions.
-          if [ -e /usr/libexec/ipsec/_stackmanager ]; then
-            /usr/libexec/ipsec/_stackmanager start
-          fi
-
```
jcaamano, line 290: So I don't understand a couple of things:

Commit message says
```
- Use container's /usr/libexec/platform-python binary to run ovs-monitor-ipsec
script.
```
But I don't see such change.

Also if we manage to keep the fallback working with commits `Remove /usr/sbin host directory dependency from ovn-ipsec-host pod` and `Remove /usr/libexec mount and avoid GLIBC compatibility issues` I don't understand why we need anything else.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
jcaamano: Maybe to keep the upgrade working , can remove after?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: you are right, `tail -F` doesn't fail upon `openvswitch-ipsec` service stop, so service crash would never get unnoticed. now checking service activeness with subshell, going to test this change with a cluster.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: The service definition is checked only once in the ovn-ipsec container before configuring and starting `openswitch-ipsec` service (it happens only once), once service is started, it remains to active during the lifetime of the node. so all other places we only check if the service is active.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: The `openvswitch-ipsec` service is just a replacement for running `ovs-monitor-ipsec` process from container and it's managing entries only with `openshift.conf` file, pluto is still managing ip xfrm state/policy entries. updated the comment to reflect this, is that clear now ?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: we run `ovs-monitor-ipsec` as a ovn-ipsec container foreground process, container will be restarted when script fails, so wanted to reflect the same for `openvswitch-ipsec` systemd service. hope that should be ok.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: that's right, this `ovs-monitor-ipsec` fallback is needed at the time of upgrade until node is upgraded to use `openvswitch-ipsec` service. updated the comment.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: > - Use container's /usr/libexec/platform-python binary to run ovs-monitor-ipsec
script.
But I don't see such change.

I just meant existing script with the above. since /usr/libexec from host is not mounted now, it uses container's  `/usr/libexec/platform-python` binary to run `ovs-monitor-ipsec` script.

```
          # Start ovs-monitor-ipsec which will monitor for changes in the ovs
          # tunnelling configuration (for example addition of a node) and configures
          # libreswan appropriately.
          # We are running this in the foreground so that the container will be restarted when ovs-monitor-ipsec fails.
          /usr/libexec/platform-python /usr/share/openvswitch/scripts/ovs-monitor-ipsec \
            --pidfile=/var/run/openvswitch/ovs-monitor-ipsec.pid --ike-daemon=libreswan --no-restart-ike-daemon \
            --ipsec-conf /etc/ipsec.d/openshift.conf --ipsec-d /var/lib/ipsec/nss \
            --log-file --monitor unix:/var/run/openvswitch/db.sock
```
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/common/ipsec-host.yaml#L291-L298

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
jcaamano: so why not do `if grep -q "openshift.conf" /etc/sysconfig/openvswitch` here?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
jcaamano: Again, why not check `grep -q "openshift.conf" /etc/sysconfig/openvswitch` here?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -472,14 +541,20 @@ spec:
...
           echo "ovs-monitor-ipsec is started"
 
-          # Monitor the ovs-monitor-ipsec process.
-          while kill -0 "$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)"; do
-            sleep 1
-          done
+          # When ovs-monitor-ipsec process is not running via openvswitch-ipsec systemd
+          # service, then keep monitoring ovn-ipsec container process until it terminates
+          # in the ovn-ipsec container's pre stop hook.
+          if ! chroot /proc/1/root systemctl is-active --quiet openvswitch-ipsec; then
```
jcaamano: why not check grep -q "openshift.conf" /etc/sysconfig/openvswitch here?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,66 @@ spec:
...
+            counter=0
+            while [ ! -f /var/log/openvswitch/ovs-monitor-ipsec.log ]; do
+              counter=$((counter+1))
+              if [ $counter -gt 30 ]; then
+                echo "ovs-monitor-ipsec.log file not created after 30 seconds"
+                exit 1
+              fi
+              sleep 1
+            done
+
```
jcaamano: I don't think we need any of this?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,66 @@ spec:
...
+                sleep 5
+                if ! chroot /proc/1/root systemctl is-active --quiet $ovsipsecservice; then
+                  echo "$ovsipsecservice service stopped unexpectedly"
+                  kill -TERM 0
+                  exit
+                fi
+              done
+            ) &
+            watcher_pid=$!
+            trap 'kill $watcher_pid 2>/dev/null; exit 1' TERM
```
jcaamano: Why do you need this?

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,66 @@ spec:
...
+              fi
+              sleep 1
+            done
+
+            # Tail the log, with a subshell periodically checking service health
+            (
+              set +x
+              while true; do
+                sleep 5
+                if ! chroot /proc/1/root systemctl is-active --quiet $ovsipsecservice; then
```
jcaamano: if you want to tightly align the lifecycle of the pod with the lifecycle of the service, you might want to actually monitor PID changes

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,66 @@ spec:
...
+                  echo "$ovsipsecservice service stopped unexpectedly"
+                  kill -TERM 0
+                  exit
+                fi
+              done
+            ) &
+            watcher_pid=$!
+            trap 'kill $watcher_pid 2>/dev/null; exit 1' TERM
+            while true; do
+              tail -F /var/log/openvswitch/ovs-monitor-ipsec.log
```
jcaamano: the problem here is that if the service exists and restarts, and the the pod exists and restarts, some of the logs of the new service run will not be on the new pod

you might want to ensure when tailing that you get all the contents of the file instead of just the last 10 lines with `tail -n +1`
But that might also include past service runs. Do you know if the service restart purges the log file?

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/ovn-kubernetes/pull/3254 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: removed unwanted checks now, but still waiting for `ovs-monitor-ipsec.pid` to be created and can copied to shared file for liveness probe.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: done.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: It's the good suggestion,thanks.  as we discussed, changed it with  `journalctl -f _SYSTEMD_INVOCATION_ID="$(systemctl show -p InvocationID --value openvswitch-ipsec)" -n all` to show only logs since last service restart.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: sure, done

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: removed it, as per your suggestion, doing service activeness check via liveness probe.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: liveness probe uses PID changes now.

`Warning  Unhealthy  8s                 kubelet            Liveness probe failed: openvswitch-ipsec service restarted (PID changed from 25992 to 35600)`

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
pperiyasamy: openvswitch-ipsec service restart doesn't purge the log file, but we are good now with new journalctl command.

pperiyasamy:
/testwith openshift/ovn-kubernetes/main/5.0-upgrade-from-stable-4.22/e2e-aws-ovn-upgrade-ipsec https://github.com/openshift/ovn-kubernetes/pull/3254 https://github.com/openshift/machine-config-operator/pull/4878

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-serial-ipsec https://github.com/openshift/machine-config-operator/pull/4878

igsilya:
@pperiyasamy I didn't test this code and I'm not an expert in pod lifecycles, but the change seems correct to me.  There is a ton of fallback code for the old behavior, which I hope can be removed in 5.1 (users can't upgrade to 5.1 without upgrading to 5.0 first, right?).  Otherwise,
/lgtm

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,41 @@ spec:
             exit 2
           fi
 
+          # Configure IPsec using the host's openvswitch-ipsec systemd service if present.
+          # This avoids running host binaries from the container and solves compatibility issues.
+          # Fall back to container-based ovs-monitor-ipsec when the openvswitch-ipsec systemd service
+          # is not yet available on the node.
+          ovsipsecservice="openvswitch-ipsec"
+          if chroot /proc/1/root systemctl list-unit-files --type=service | grep "$ovsipsecservice"; then
```
jcaamano: use `grep -q`

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,41 @@ spec:
...
+              # Wait for ovs-monitor-ipsec PID file to be created
+              counter=0
+              while [ ! -f /var/run/openvswitch/ovs-monitor-ipsec.pid ]; do
+                counter=$((counter+1))
+                if [ $counter -gt 30 ]; then
+                  echo "ovs-monitor-ipsec.pid not created after 30 seconds"
+                  exit 1
+                fi
+                sleep 1
+              done
```
jcaamano: I don't think we need to check this, if the service is enabled, the pid needs to be there. And if it is not, we can just bail out with the suggestion in my next comment

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -249,6 +258,41 @@ spec:
...
+            fi
+
+            # Store initial PID for liveness probe to detect service restarts
+            cp /var/run/openvswitch/ovs-monitor-ipsec.pid /tmp/ovs-monitor-ipsec-base.pid
```
jcaamano: I guess I would do something like this:
```
              # Capture PID and InvocationID atomically, verify no restart occurred
              invocation_id=$(chroot /proc/1/root systemctl show -p InvocationID --value openvswitch-ipsec)
              main_pid=$(chroot /proc/1/root systemctl show -p MainPID --value openvswitch-ipsec)
              file_pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
              if [ "$main_pid" != "$file_pid" ]; then
                echo "openvswitch-ipsec PID mismatch: systemd reports $main_pid, PID file has $file_pid"
                exit 1
              fi
              echo "$main_pid" > /tmp/ovs-monitor-ipsec-base.pid

              while true; do
                chroot /proc/1/root journalctl -f "_SYSTEMD_INVOCATION_ID=$invocation_id" -n all
                echo "journalctl exited unexpectedly, retrying..."
                sleep 2
              done
```

So we capture invocation_id and pid right away, to frame what service run we are monitoring.

pperiyasamy: nice, updated it.

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -360,9 +390,26 @@ spec:
...
+              if grep -q "openshift.conf" /etc/sysconfig/openvswitch; then
+                if ! chroot /proc/1/root systemctl is-active --quiet openvswitch-ipsec; then
+                  echo "openvswitch-ipsec service is not active"
+                  exit 1
+                fi
+                if [ -f /tmp/ovs-monitor-ipsec-main.pid ]; then
+                  file_pid=$(cat /var/run/openvswitch/ovs-monitor-ipsec.pid 2>/dev/null)
+                  main_pid=$(cat /tmp/ovs-monitor-ipsec-main.pid)
+                  if [ "$main_pid" != "$file_pid" ]; then
+                    echo "openvswitch-ipsec PID mismatch: systemd reports $main_pid, PID file has $file_pid"
```
jcaamano: I would keep the original message you had here

jcaamano (commented):
small nit, otherwise lgtm

bindata/network/ovn-kubernetes/common/ipsec-host.yaml
```diff
@@ -332,16 +368,10 @@ spec:
...
-        - mountPath: /etc/openvswitch
-          name: etc-openvswitch
         - mountPath: /var/lib
           name: host-var-lib
         - mountPath: /etc
           name: host-etc
-        - mountPath: /usr/sbin
-          name: usr-sbin
-        - mountPath: /usr/libexec
-          name: usr-libexec
```
igsilya, line 344: Hmm.  The code still supports both modes - systemd on the host or `ovs-monitor-ipsec` daemon inside the container.  But if we do not mount these, then `ovs-monitor-ipsec` inside the container will use `ipsec` command from the inside of the container that is potentially incompatible with the `pluto` running on the host.  Is that right?  Should the mounts be preserved until the fallback code is removed?

jcaamano: I guess this comment is inline with the comment [here](https://github.com/openshift/cluster-network-operator/pull/2662#discussion_r3579438607) 
```
Also if we manage to keep the fallback working with commits Remove /usr/sbin host directory dependency from ovn-ipsec-host pod and Remove /usr/libexec mount and avoid GLIBC compatibility issues I don't understand why we need anything else.
```

I talked with @pperiyasamy offline and what I understood is this is picking between two evils: either we live RHEL10 glibc problems or we live with our host/container protocol problems and he chose the latter as the most unlikely. Please @pperiyasamy confirm.

pperiyasamy: yes, mounting `/usr/sbin` also broke with `chroot` command (which pod uses it heavily) due to glibc compatibility issues. so had to get rid of mounting host `/usr/sbin` directory completely.

jcaamano: And I guess this would be the only thing that would work in an hypothetical RHEL9 -> RHEL10 upgrade case, which I am not sure its a thing.

igsilya: Yeah, I see, it's hard to actually have both cases 100% working without doing terrible things inside the container like building a separate chroot or something.  Since RHEL10 is TP in 5.0, I suppose, sacrificing the 4.x RHEL 9 -> 5.0 RHEL 10 upgrade path is more reasonable than having breakages in common cases.  May need some documentation around that, but otherwise sounds fine to me.

mkowalski:
/override-sticky ci/prow/e2e-aws-ovn-upgrade-ipsec
/override-sticky ci/prow/e2e-aws-ovn-fdp-qe
/override-sticky ci/prow/e2e-metal-ipi-ovn-dualstack-bgp-local-gw

PR #2665, opened 2025-03-11
OCPBUGS-52972: Correct namespace ref and cert dir

bpickard22:
cc @dougbtv 

correct paste error on cert dir and point controller daemonest to correct whereabouts namespace


dougbtv (approved):
/lgtm

PR #2670, opened 2025-03-25
OCPBUGS-5241: Auto opt in for whereabouts-controller

bpickard22:
Using the shim to deploy the controller causes the controller to error out, auto-deploy works


dougbtv:
I'm ok moving forward on this for the time being, but we'd  like to ideally have it as opt-in, tracking in https://issues.redhat.com/browse/CORENET-4481

PR #2673, opened 2025-03-31
[release-4.16] OCPBUGS-53317: Fix feature migration for EgressIP

pliurh:
1. In the migration case from SDN to OVN, set egress-assignable node
   annotation if 'egressCIDRs' is configured in the hostsubnet. The
   user may configure more hostsubnet with 'egressCIDRs' then the
   egress IPs.

2. In the rollback case, verify the existence of the node annotation
   `cloud.network.openshift.io/egress-ipconfig` before using it. In
   non-cloud platforms, use the node annotation 'node-primary-ifaddr'
   to derive the 'egressCIDRs'.

### DIFF
--- a/pkg/controller/operconfig/migration.go
+++ b/pkg/controller/operconfig/migration.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"reflect"
 	"strings"
 	"sync"
@@ -42,6 +43,7 @@ const egressAssignable = "k8s.ovn.org/egress-assignable"
 const multicastEnabledSDN = "netnamespace.network.openshift.io/multicast-enabled"
 const multicastEnabledOVN = "k8s.ovn.org/multicast-enabled"
 const ovnAnnotationPrefix = "k8s.ovn.org"
+const ovnAnnotationNodeIfAddr = ovnAnnotationPrefix + "/node-primary-ifaddr"
 
 var networkTypeMigrationConditionTypes = []string{
 	names.NetworkTypeMigrationMTUReady,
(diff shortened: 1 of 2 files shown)


pkg/controller/operconfig/migration.go
```diff
@@ -516,8 +541,8 @@ func netNamespaceHasEgressIpConfig(nns uns.Unstructured) bool {
 }
 
 func hostSubnetHasEgressIpConfigAutomatic(hsn uns.Unstructured) bool {
-	if hsn.Object["egressIPs"] != nil && hsn.Object["egressCIDRs"] != nil {
-		return len(hsn.Object["egressIPs"].([]interface{})) > 0 && len(hsn.Object["egressCIDRs"].([]interface{})) > 0
+	if hsn.Object["egressCIDRs"] != nil {
```
martinkennelly, line 544: can you explain why the changes here?

martinkennelly (commented):
/lgtm

martinkennelly:
is qe using eip disruption tester during migration to determine this ok? Ask Sai for the tool. Its somewhere in the org https://github.com/cloud-bulldozer

pkg/controller/operconfig/migration.go
pliurh: The customer may configure more nodes that can host egress IP (by setting `egressCIDRs` to hostsubnets) than the number of egress IP. So we shouldn't assume that all these nodes have an egress IP allocated. We shall just label those nodes with `egress-assignable` while migrating the configuration.

pkg/controller/operconfig/migration.go
```diff
@@ -408,18 +410,41 @@ func convertOvnEgressIpToSdnEgressIp(ctx context.Context, client cnoclient.Clien
...
+					}
+					nodeIfAddr := make(map[string]string)
+					if err := json.Unmarshal([]byte(nodeIfAddrAnno), &nodeIfAddr); err != nil {
+						return err
+					}
+					address, ok := nodeIfAddr["ipv4"]
+					if !ok {
+						return fmt.Errorf("unexpected error: node annotation missing ipv4 entry")
+					}
+					_, cidr, err := net.ParseCIDR(address)
```
kyrtapz, line 441: Why are you parsing the CIDR here but not doing that for the cloud case?

kyrtapz:
Thanks for the unit tests!

pkg/controller/operconfig/migration.go
pliurh: In cloud case, we can use node annotation `cloud.network.openshift.io/egress-ipconfig` to derive the node underlay subnet, which is not available in non-cloud platforms.

PR #2676, opened 2025-04-01
CORENET-4481: update install-cni script

bpickard22:
update install-cni script to match whereabouts bin so that we get cert rotatation in openshift

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -567,7 +567,7 @@ spec:
 
             CNI_BIN_DIR=${CNI_BIN_DIR:-"/host/opt/cni/bin/"}
             WHEREABOUTS_KUBECONFIG_FILE_HOST=${WHEREABOUTS_KUBECONFIG_FILE_HOST:-"/etc/cni/net.d/whereabouts.d/whereabouts.kubeconfig"}
-            CNI_CONF_DIR=${CNI_CONF_DIR:-"/host/etc/cni/net.d"}
+            CNI_CONF_DIR=${CNI_CONF_DIR:-"/host{{ .SystemCNIConfDir }}"}
             WHEREABOUTS_RECONCILER_CRON=${WHEREABOUTS_RECONCILER_CRON:-30 4 * * *}
 
             # Make a whereabouts.d directory (for our kubeconfig)
@@ -655,8 +655,8 @@ spec:
             EOF
 
             # Copy the config from ConfigMap to the desired directory
-            cp /etc/whereabouts/config/whereabouts.conf $WHEREABOUTS_GLOBALCONFIG
-            chmod 600 $ WHEREABOUTS_CONF_FILE
+            cp /etc/whereabouts/config/whereabouts.conf $WHEREABOUTS_CONF_FILE
+            chmod 600 $WHEREABOUTS_CONF_FILE
 
             else
               warn "Doesn't look like we're running in a kubernetes environment (no serviceaccount token)"
@@ -680,24 +680,12 @@ spec:
 
             }
 
-            function get_token_md5sum {
-              md5sum "$SERVICE_ACCOUNT_TOKEN_PATH" | awk '{print $1}'
-            }
-
-            function get_ca_file_md5sum {
-              if [ ! -f "$KUBE_CA_FILE" ]; then
-                echo ""
-                return
-              fi
-              md5sum "$KUBE_CA_FILE" | awk '{print $1}'
-            }
             generateKubeConfig
-            export LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
-            export LAST_KUBE_CA_FILE_MD5SUM="$(get_ca_file_md5sum)"
             # ------------------ end Generate a "kube-config"
 
             # ----------------- Generate a whereabouts conf
-            generateWhereaboutsConf
+            # removed because we have the configmap
+            #generateWhereaboutsConf
             # ---------------- End generate a whereabouts conf
 
 
@@ -706,22 +694,6 @@ spec:
             # cp -f /whereabouts $CNI_BIN_DIR
 
             # ---------------------- end generate a "kube-config".
-
-            echo "Sleep and Watching for service account token and CA file changes..."
-            # enter sleep/watch loop
-            while true; do
-              # Check the md5sum of the service account token and ca.
-              svcaccountsum="$(get_token_md5sum)"
-              casum="$(get_ca_file_md5sum)"
-              if [ "$svcaccountsum" != "$LAST_SERVICEACCOUNT_MD5SUM" ] || ! [ "$SKIP_TLS_VERIFY" == "true" ] && [ "$casum" != "$LAST_KUBE_CA_FILE_MD5SUM" ]; then
-                log "Detected service account or CA file change, regenerating kubeconfig..."
-                generateKubeConfig
-                LAST_SERVICEACCOUNT_MD5SUM="$svcaccountsum"
-                LAST_KUBE_CA_FILE_MD5SUM="$casum"
-              fi
-
-              sleep 1s
-            done
 
             # Unless told otherwise, sleep forever.
             # This prevents Kubernetes from restarting the pod repeatedly.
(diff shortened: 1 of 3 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -585,8 +585,30 @@ spec:
...
+
+            function warn()
+            {
+                log "WARN: {$1}"
+            }
+
+
+            function generateKubeConfig {
+              # Check if we're running as a k8s pod.
             if [ -f "$SERVICE_ACCOUNT_PATH/token" ]; then
```
dougbtv, line 612: hate to be this guy -- but, is this funky indentation expected?

bindata/network/multus/multus.yaml
```diff
@@ -630,19 +658,58 @@ spec:
...
 
+            }
+
+            generateKubeConfig
+
+            # ------------------ end Generate a "kube-config"
+
+            # ----------------- Generate a whereabouts conf
+
+            function generateWhereaboutsConf {
```
dougbtv, line 673: Do we actually want this? Like... How were we generating this before? We should make sure it's the same as we had it.

bindata/network/multus/multus.yaml
```diff
@@ -630,19 +658,58 @@ spec:
...
+                svcaccountsum=$(md5sum $SERVICE_ACCOUNT_TOKEN_PATH | awk '{print $1}')
+                casum=$(md5sum $KUBE_CA_FILE | awk '{print $1}')
+                if [ "$svcaccountsum" != "$LAST_SERVICEACCOUNT_MD5SUM" ] || [ "$casum" != "$LAST_KUBE_CA_FILE_MD5SUM" ]; then
+                  # log "Detected service account or CA file change, regenerating kubeconfig..."
+                  generateKubeConfig
+                fi
 
-            # ---------------------- end Generate a "kube-config".
+                sleep 1
+              done
```
dougbtv, line 712: the above is money, I know this is the thing we need, but... also check funky indents here on `done`

bindata/network/multus/multus.yaml
bpickard22: ah, might have missed a line when mass tabbing over

bindata/network/multus/multus.yaml
bpickard22: its functionally the same, i think when we implemented this in whereabouts we agreed that moving the logic into functions was cleaner and easier to understand

mlguerrero12: but still have the config map here (somewhere above this file), so we should delete it if we go for this approach

mlguerrero12:
In general, the script is different than the current version upstream. That's fine since we don't need to the token-watcher.sh script separately. However, we need to try to sync the rest

bindata/network/multus/multus.yaml
```diff
@@ -585,8 +585,30 @@ spec:
...
+            LAST_SERVICEACCOUNT_MD5SUM=""
+            LAST_KUBE_CA_FILE_MD5SUM=""
 
-            # Check if we're running as a k8s pod.
+
+            # Setup our logging routines
+
+            function log()
+            {
+                echo "$(date --iso-8601=seconds) ${1}"
```
mlguerrero12, line 596: this is different upstream. I tested it during the review and I got an error. It is now

    echo "$(date -Iseconds) ${1}"

bindata/network/multus/multus.yaml
```diff
@@ -630,19 +658,58 @@ spec:
...
+
+            # ---------------------- end generate a "kube-config".
+
+            # enter sleep/watch loop
+
+            while true; do
+              # Check the md5sum of the service account token and ca.
+                svcaccountsum=$(md5sum $SERVICE_ACCOUNT_TOKEN_PATH | awk '{print $1}')
+                casum=$(md5sum $KUBE_CA_FILE | awk '{print $1}')
+                if [ "$svcaccountsum" != "$LAST_SERVICEACCOUNT_MD5SUM" ] || [ "$casum" != "$LAST_KUBE_CA_FILE_MD5SUM" ]; then
```
mlguerrero12, line 706: LAST_SERVICEACCOUNT_MD5SUM, this is never updated. Fixed upstream as well

bindata/network/multus/multus.yaml
dougbtv: Yeah, these lines: https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/multus/multus.yaml#L108-L125

I introduced that in this commit: https://github.com/openshift/cluster-network-operator/commit/68e28f829ecbd65b85407c3d5004470f02c3ab33 -- which should've been cleaned up, and I missed it. So, my b.

Actually, it's critical that we remove the config map

bindata/network/multus/multus.yaml
```diff
@@ -362,12 +344,7 @@ spec:
         - name: os-release
           hostPath:
             path: /etc/os-release
-            type: File
-        - name: cni-binary-copy
-          configMap:
-            name: cni-copy-resources
-            defaultMode: 0744
-        - name: multus-socket-dir-parent
+            type: Filehttps://github.com/openshift/cluster-network-operator/pull/2676
```
dougbtv, line 347: that's got a copy/paste error

bindata/network/multus/multus.yaml
```diff
@@ -868,11 +899,8 @@ spec:
...
-          configMap:
-            name: whereabouts-config
-            items:
-            - key: whereabouts.conf
-              path: whereabouts.conf
-{{- end}}
+        - hostPath:
+            path: {{ .SystemCNIConfDir }}/whereabouts.d/whereabouts.conf
+            type: File
+          name: whereabouts-flat-config
```
dougbtv, line 905: this doesn't look like it's getting mounted anywhere?

bindata/network/multus/multus.yaml
```diff
@@ -630,19 +633,54 @@ spec:
...
+            }
+            generateKubeConfig
+            export LAST_SERVICEACCOUNT_MD5SUM="$(get_token_md5sum)"
+            export LAST_KUBE_CA_FILE_MD5SUM="$(get_ca_file_md5sum)"
+            # ------------------ end Generate a "kube-config"
+
+            # ----------------- Generate a whereabouts conf
+            generateWhereaboutsConf
+            # ---------------- End generate a whereabouts conf
+
```
mlguerrero12, line 677: The for loop was removed upstream. It is now in the token-watcher script. I think we should use what we have upstream as a reference but modify everything according to our needs. There are differences everywhere as @dougbtv is also pointing out.

dougbtv: +1. I think we're on the right track for now. I think this recommendation is a perfect step one -- let's get the contents of what's upstream in here for now.

then, let's make modifications upstream so that  we can parameterize what we need (e.g. skipping binary copy, setting directories, etc), and then we can use what's upstream directly as a next step after this first step.

dougbtv:
I wound up filing https://issues.redhat.com/browse/CORENET-5835 to track the use of the upstream scripts :+1:

mlguerrero12:
> I wound up filing https://issues.redhat.com/browse/CORENET-5835 to track the use of the upstream scripts 👍

During the upstream PR review, it was concluded that having another script was the quickest solution. However, rotating the token in code might be a better approach. If we do this, we could simplify the logic of the upstream scripts again. Just to keep in mind.

/lgtm

bindata/network/multus/multus.yaml
```diff
@@ -874,5 +924,174 @@ spec:
...
+              log "Detected service account or CA file change, regenerating kubeconfig..."
+              generateKubeConfig
+              LAST_SERVICEACCOUNT_MD5SUM="$svcaccountsum"
+              LAST_KUBE_CA_FILE_MD5SUM="$casum"
+            fi
+
+            sleep 1s
+          done
+
+          # Unless told otherwise, sleep forever.
```
dougbtv, line 1063: remove all this stuff after the `while true` loop as it doesn't get exec'd :+1:

bindata/network/multus/multus.yaml
```diff
@@ -874,5 +924,174 @@ spec:
             items:
             - key: whereabouts.conf
               path: whereabouts.conf
+---
+apiVersion: apps/v1
+kind: Daemonset
+metadata:
+  name: whereabouts-watcher
```
dougbtv, line 931: let's rename this `whereabouts-token-watcher` -- as we have some `token-watcher` nomenclature upstream (and maybe we can directly use that script here, in follow up), and it'll be more specific

bindata/network/multus/multus.yaml
```diff
@@ -874,5 +924,166 @@ spec:
...
+            context:
+              cluster: local
+              user: whereabouts
+              namespace: ${WHEREABOUTS_NAMESPACE}
+          current-context: whereabouts-context
+          EOF
+
+          # Copy the config from ConfigMap to the desired directory
+          cp /etc/whereabouts/config/whereabouts.conf $WHEREABOUTS_GLOBALCONFIG
+          chmod 600 $ WHEREABOUTS_CONF_FILE
```
dougbtv, line 1024: peep this `$ WHEREABOUTS_CONF_FILE` -- that space shouldn't be there after the dollar sign

bindata/network/multus/multus.yaml
```diff
@@ -874,5 +924,166 @@ spec:
...
+              LAST_SERVICEACCOUNT_MD5SUM="$svcaccountsum"
+              LAST_KUBE_CA_FILE_MD5SUM="$casum"
+            fi
+
+            sleep 1s
+          done
+
+      resources:
+        requests:
+          cpu: 10mi
```
dougbtv, line 1065: that CPU request looks wrong?

dougbtv: yeah I think  this should be `10m`

bindata/network/multus/multus.yaml
```diff
@@ -874,5 +924,166 @@ spec:
...
+          memory: 10mi
```
dougbtv, line 1066: And 10Mi

dougbtv:

This appears to be an unrelated failure, and is appearing on other PRs, and we have significant signal from other jobs.

see: [job history](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-network-operator-master-hypershift-e2e-aks)
and [example failure during this PR](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2685/pull-ci-openshift-cluster-network-operator-master-hypershift-e2e-aks/1910687306906341376) and [the same build failure on another PR](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2676/pull-ci-openshift-cluster-network-operator-master-hypershift-e2e-aks/1910717307877855232)

PR #2677, opened 2025-04-01
OCPBUGS-54159: frr-k8s: tolerate all taints

jcaamano:
Just as ovnkube-node, frr-k8s daemon set is an infrastructure component and should be deployed regardless of taints. This solves frr-k8s pods not being deployed on nodes labeled with node-role.kubernetes.io/infra and tainted with NoSchedule.

### DIFF
--- a/bindata/network/frr-k8s/frr-k8s.yaml
+++ b/bindata/network/frr-k8s/frr-k8s.yaml
@@ -276,9 +276,4 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       tolerations:
-      - key: node-role.kubernetes.io/master
-        effect: NoSchedule
-        operator: Exists
-      - key: node-role.kubernetes.io/control-plane
-        effect: NoSchedule
-        operator: Exists
+      - operator: "Exists"


jcaamano:
Windows job fails because WMCO manager pod fails to run, logs:
```
2025-04-02T08:44:49Z	ERROR	controllers.secret	Unable to retrieve private key, please ensure it is created	{"error": "the cache is not started, can not read objects"}
github.com/openshift/windows-machine-config-operator/controllers.(*SecretReconciler).SetupWithManager
	/build/windows-machine-config-operator/controllers/secret_controller.go:72
main.main
	/build/windows-machine-config-operator/cmd/operator/main.go:231
runtime.main
	/usr/lib/golang/src/runtime/proc.go:272
2025-04-02T08:44:49Z	ERROR	setup	unable to create ConfigMap reconciler	{"error": "error creating ignition object: failed to parse MachineConfig ignition: unsupported config version\nReport: "}
main.main
	/build/windows-machine-config-operator/cmd/operator/main.go:242
runtime.main
	/usr/lib/golang/src/runtime/proc.go:272
```

Seems unrelated to CNO so will override. I did advise on the appropriate slack forum.

PR #2681, opened 2025-04-04
OCPBUGS-53180: Add IPv6 NGINX configuration

orenc1:
Add an additional listen directive that will catch IPv6-originated requests towards the nginx server that serves the networking-console-plugin manifests.

### DIFF
--- a/bindata/networking-console-plugin/001-config-map.yaml
+++ b/bindata/networking-console-plugin/001-config-map.yaml
@@ -9,6 +9,7 @@ data:
       keepalive_timeout  65;
       server {
         listen              9443 ssl;
+        listen              [::]:9443 ssl;
         ssl_certificate     /var/cert/tls.crt;
         ssl_certificate_key /var/cert/tls.key;
         root                /opt/app-root/src;


bindata/networking-console-plugin/001-config-map.yaml
```diff
@@ -9,6 +9,7 @@ data:
       keepalive_timeout  65;
       server {
         listen              9443 ssl;
+        listen              [::]:9443 ssl;
```
spadgett, line 12: I believe this will break on clusters have have IPv6 disabled. You probably need a fix like https://github.com/openshift/console-plugin-template/pull/79

spadgett: It looks like this is an unsupported configuration. Disregard

kyrtapz, line 12: The code below is currently not doing anything, do you want to address that as a part of this PR?
https://github.com/openshift/cluster-network-operator/blob/a8648b35c0e45671be50c1171549e2b28121f35d/bindata/networking-console-plugin/002-deployment.yaml#L64-L70

kyrtapz (commented):
I guess that is acceptable but please note that the net-console is single-stack and will always pick only the primary network family. It means that on a dualstack cluster the service will be IPv4.

bindata/networking-console-plugin/001-config-map.yaml
kyrtapz: sure but the code I highlighted is not doing anything. In case you don't want to configure the IP family dynamically you should just remove it. It can be done as a followup.

kyrtapz:

https://issues.redhat.com/browse/OCPBUGS-54499

PR #2684, opened 2025-04-10
OCPBUGS-54837: Add missing backendAddressPools read permission on azure

kyrtapz:
Although the issue was found while testing on AKS it is not specific to it.
This also affects standalone Azure deployments.


kyrtapz:
@ricky-rav ptal as I you were looking into backendAddressPools with CNCC in the past.

cadenmarchese (approved):

kyrtapz:
/cherry-pick release-4.18 release-4.17 release-4.16 release-4.15 release-4.14

PR #2691, opened 2025-04-22
Use same conf file both host and containerized ipsec deployment

pperiyasamy:
The host and containerized ipsec deployments are not sharing same conf file containing ipsec connection entries populated by ovs-monitor-ipsec process. when ipsec is moved from containerized to host deployment at the time of ipsec installation, this file won't be present after ipsec machine config installed, node reboot, so ipsec can't be established before crio is started. Hence this PR makes containerized ipsec deployment to use `/etc/ipsec.d/openshift.conf` file from the host so that it can be used later by pluto daemon to establish IKE SAs before kubelet/crio is started.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
@@ -261,6 +261,18 @@ spec:
           # Workaround for https://github.com/libreswan/libreswan/issues/373
           ulimit -n 1024
 
+          # The ovs-monitor-ipsec doesn't set authby, so when it calls ipsec auto --start
+          # the default ones defined at Libreswan's compile time will be used. On restart,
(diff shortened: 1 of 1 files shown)


pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-ovn-ipsec-step-registry https://github.com/openshift/machine-config-operator/pull/5071

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial https://github.com/openshift/machine-config-operator/pull/5071

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-aws-ovn-ipsec-serial openshift/machine-config-operator#5071

pperiyasamy:
/testwith openshift/cluster-network-operator/master/e2e-ovn-ipsec-step-registry https://github.com/openshift/machine-config-operator/pull/5071

PR #2693, opened 2025-04-23
[DNM] Use correct regex for IPv6 IPsec connections

pperiyasamy:


pperiyasamy:
The fix is in effect, e2e-metal-ipi-ovn-ipv6-ipsec [run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2693/pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-ipv6-ipsec/1915124964176957440) failed because of job timeout, so we must get https://github.com/openshift/release/pull/64174 also landed.

kyrtapz:
@pperiyasamy is this a workaround? Please add a comment to the code and a description of the issue to the PR/commit.
If there is an issue with the script is it tracked as a bug somewhere?

igsilya:
IMO, this code should not be merged.  It is a good test to verify the fix, but the fix should be shipped in the OVS package instead.
/hold

PR #2697, opened 2025-04-29
OCPBUGS-54806: Add recording rules for UDN telemetry

danwinship:
(The metric is designed to have only the things we want for telemetry anyway, but the rules for telemetry metrics say we need an explicit recording rule anyway, so that if more info gets added to the metric later it doesn't get automatically sucked in to telemetry.)

/cc

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
@@ -23,6 +23,10 @@ spec:
       expr: max by(direction, action)(ovnkube_controller_admin_network_policies_rules)
     - record: cluster:ovnkube_controller_baseline_admin_network_policies_rules:max
       expr: max by(direction, action)(ovnkube_controller_baseline_admin_network_policies_rules)
+    - record: cluster:ovnkube_clustermanager_user_defined_networks:max
+      expr: max by(role, topology)(ovnkube_clustermanager_user_defined_networks)
+    - record: cluster:ovnkube_clustermanager_cluster_user_defined_networks:max
+      expr: max by(role, topology)(ovnkube_clustermanager_cluster_user_defined_networks)
     # OVN kubernetes cluster manager functional alerts
     - alert: V4SubnetAllocationThresholdExceeded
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/managed/alert-rules-control-plane.yaml
```diff
@@ -23,6 +23,10 @@ spec:
       expr: max by(direction, action)(ovnkube_controller_admin_network_policies_rules)
     - record: cluster:ovnkube_controller_baseline_admin_network_policies_rules:max
       expr: max by(direction, action)(ovnkube_controller_baseline_admin_network_policies_rules)
+    - record: cluster:ovnkube_clustermanager_user_defined_networks:max
+      expr: max by(role, topology)(ovnkube_clustermanager_user_defined_networks)
+    - record: cluster:ovnkube_clustermanager_cluster_user_defined_networks:max
+      expr: max by(role, topology)(ovnkube_clustermanager_cluster_user_defined_networks)
```
danwinship, line 29: @tssurya I based this on the ANP metrics, but why do we do "max" here rather than returning the current value of the metric?

jan--f: Yeah this will work ok. We use aggregation functions like this for telemetry to reduce the number of labels. This gets rid of any automatically added labels (container, pod, namespace and such). Since we don't really aggregate values here (since we preserve the labels that matter) it can feel a bit awkward.

Other options are `sum` or `group_by` like in https://github.com/openshift/cluster-network-operator/blob/2ca824bfbd32ad196d1ac8e4167bfd369f464f66/bindata/cluster-network-operator/prometheus.yaml#L15-L18

jan--f: See also https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

tssurya (approved):
/lgtm

danwinship:
/verified bypass

no way to validate telemetry until it's fully merged

PR #2705, opened 2025-05-09
Add cert-rotation script into cno

bpickard22:
update install-cni script to match whereabouts bin so that we get cert rotatation in openshift

we have to break the tokenwatcher section of the whereabouts script out because it was running in an init container, so it would never "complete", hanging the rollout

to accomplish this I created a new whereabouts watcher ds to house this section of the script

IMPORTANT: to enable the token watcher to run you must have the the whereabouts shim in your network object
https://docs.openshift.com/container-platform/4.12/networking/multiple_networks/configuring-additional-network.html#nw-multus-creating-whereabouts-reconciler-daemon-set_configuring-additional-network

(cherry picked from commit 6e82842306a2722188b7af48d3a73b4bdd36c0e4) (cherry picked from commit 89d7afc3b2b80608de39eb1b67644136d13173c5)


bpickard22:
going to pick these changes from a different pr to make sure i get all the changes needed for cert rotation

PR #2708, opened 2025-05-14
frr-k8s: stop listening for incoming connection in the bgp daemon

karampok:
frr-k8s will not listen to port 179, and BGP peering can only be established from OCP to external.

### DIFF
--- a/bindata/network/frr-k8s/config.yaml
+++ b/bindata/network/frr-k8s/config.yaml
@@ -46,7 +46,7 @@ data:
     #
     vtysh_enable=yes
     zebra_options="  -A 127.0.0.1 -s 90000000 --limit-fds 100000"
-    bgpd_options="   -A 127.0.0.1 --limit-fds 100000"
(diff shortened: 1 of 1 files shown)


karampok:
cc @fedepaol

fedepaol:
@karampok I think this deserves do be a bug as we did it non - intentionally (the behavior was reverted upstream and on metallb itself).

karampok:
The frr-k8s fail with
```
Summarizing 1 Failure:
  [FAIL] BGPSessionState BGP and BFD Each node manages its statuses [It] IPV4
  /root/dev-scripts/frr/e2etests/tests/bgp_session_status.go:209
Ran 47 of 105 Specs in 3455.224 seconds
FAIL! -- 46 Passed | 1 Failed | 0 Pending | 58 Skipped
--- FAIL: TestE2E (3455.28s)
FAIL
```
which is known issue

fedepaol:
@karampok please un - hold when we manage to get a green frrk8s run (iiuc, it's failing due to external factors, which should sort out hopefully)

PR #2710, opened 2025-05-19
Fix live migration with feature migration configured explicitly

pliurh:
The function isMigrationChangeSafe reports error when feature migration configured explicitly. It prevent the live migration from being completed at the last step, when the clusterconfig controller remove the flags spec.migration.networkType and spec.migration.mode.


pkg/network/render.go
```diff
@@ -689,7 +689,9 @@ func isMigrationChangeSafe(prev, next *operv1.NetworkSpec, infraStatus *bootstra
 	if next.Migration != nil && next.Migration.Mode == operv1.LiveNetworkMigrationMode && infraStatus.HostedControlPlane != nil {
 		return []error{errors.Errorf("live migration is unsupported in a HyperShift environment")}
 	}
-	if prev.Migration != nil && next.Migration != nil && prev.Migration.NetworkType != next.Migration.NetworkType && next.Migration.Mode != operv1.LiveNetworkMigrationMode {
+	if prev.Migration != nil && next.Migration != nil && next.Migration.NetworkType != "" &&
```
kyrtapz, line 692: This now allows changing the NetworkType anytime during the live migration.
Should we at least limit it to going from "set" to "unset"?

pkg/network/render_test.go
```diff
@@ -611,3 +611,104 @@ func Test_renderNetworkDiagnostics(t *testing.T) {
...
+		{
+			name:        "changing network type during offline live migration should error",
+			prev:        makeNetSpec("OpenShiftSDN", ""),
+			next:        makeNetSpec("OVNKubernetes", ""),
+			infraStatus: &bootstrap.InfraStatus{},
+			wantErr:     true,
+			errMsg:      "cannot change migration network type after migration has started",
+		},
+		{
+			name:        "changing network type during live migration should be valid",
```
kyrtapz, line 662: We don't support such scenario do we?

pkg/network/render.go
pliurh: This patch is actually to allow it to go from "set" to "unset". This happens at the last step of the live migration. When no `migration.featureMigration` is configured, the `migration` field will be removed; therefore, it won't report an error here. However, when the user configures `migration.featureMigration` manually, the `migration` field will not be removed by the server-side apply. Instead, only the `migration.networkType` and `migration.mode` will be unset.

kyrtapz:
/override ci/prow/e2e-gcp-ovn-upgrade
https://issues.redhat.com/browse/TRT-2133

zhaozhanqi:
pre-merge tested this PR and this can fix this issue 
/label qe-approved
/label cherry-pick-approved

PR #2714, opened 2025-05-21
OCPBUGS-55962: Provide config map to force loose isolation for UDN networks

pperiyasamy:
This PR adds support for configuring loose isolation mode for the BGP advertised UDN networks. The config map with name `openshift-network-operator/udn-config-overrides` must be created with `force-loose-isolation` key set to `true` which rolls out the loose mode by recreating `ovnkube-node` daemonset pods.

Steps to roll out loose isolation mode:

1. Create a below config map.
```
kind: ConfigMap
apiVersion: v1
metadata:
  name: udn-config-overrides
  namespace: openshift-network-operator
data:
  force-loose-isolation: "true"
```
2. Wait for ovnkube-node DS rollout to complete.
3. Create BGP advertised UDN networks.

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
@@ -430,6 +430,10 @@ spec:
         - name: "NO_PROXY"
           value: "{{ .NO_PROXY}}"
 {{ end }}
+        {{ if .IsLooseUDNIsolationEnabled }}
+        - name: UDN_ISOLATION_MODE
+          value: "loose"
+        {{ end }}
         - name: K8S_NODE
           valueFrom:
             fieldRef:
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -126,6 +126,11 @@ type InfraStatus struct {
 
 	// ConsolePluginCRDExists set to true when the consoleplugins.console.openshift.io has been deployed.
 	ConsolePluginCRDExists bool
+
+	// LooseUDNIsolationModeEnabled set to true when loose isolation mode is enabled between two BGP advertised
+	// UDN networks. In loose isolation mode, those network pods can communicate with each other accoding to
+	// provider network configuration.
+	LooseUDNIsolationModeEnabled bool
 }
 
 // APIServer is the hostname & port of a given APIServer. (This is the
--- a/pkg/platform/platform.go
+++ b/pkg/platform/platform.go
@@ -62,6 +62,22 @@ func isNetworkNodeIdentityEnabled(client cnoclient.Client, infra *bootstrap.Infr
 	return true, nil
 }
 
+// isLooseUDNIsolationEnabled determines if loose udn isolation mode should be enabled.
+// It checks the `force-loose-isolation` key in the openshift-network-operator/udn-config-overrides configmap.
+// If the configmap doesn't exist, it returns false (the UDN isolation is protected by default).
+func isLooseUDNIsolationEnabled(client cnoclient.Client) (bool, error) {
+	configMap := &corev1.ConfigMap{}
+	if err := client.ClientFor("").CRClient().Get(context.TODO(),
+		types.NamespacedName{Name: "udn-config-overrides", Namespace: names.APPLIED_NAMESPACE}, configMap); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("unable to bootstrap OVN, unable to retrieve udn-config-overrides config: %s", err)
+	}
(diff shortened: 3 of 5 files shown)


pkg/platform/platform.go
```diff
@@ -62,6 +62,22 @@ func isNetworkNodeIdentityEnabled(client cnoclient.Client, infra *bootstrap.Infr
 	return true, nil
 }
 
+// isLooseUDNIsolationEnabled determines if loose udn isolation mode should be enabled.
+// It checks the `force-loose-isolation` key in the openshift-network-operator/udn-config-overrides configmap.
+// If the configmap doesn't exist, it returns false (the UDN isolation is protected by default).
+func isLooseUDNIsolationEnabled(client cnoclient.Client) (bool, error) {
+	configMap := &corev1.ConfigMap{}
+	if err := client.ClientFor("").CRClient().Get(context.TODO(),
```
arkadeepsen, line 70:
```suggestion
	if err := client.Default().CRClient().Get(context.TODO(),
```

pkg/platform/platform.go
```diff
@@ -62,6 +62,22 @@ func isNetworkNodeIdentityEnabled(client cnoclient.Client, infra *bootstrap.Infr
...
+		types.NamespacedName{Name: "udn-config-overrides", Namespace: names.APPLIED_NAMESPACE}, configMap); err != nil {
```
arkadeepsen, line 71: Should `udn-config-overrides` be added as a constant? Then the constant can be used here.

pkg/bootstrap/types.go
```diff
@@ -126,6 +126,11 @@ type InfraStatus struct {
 
 	// ConsolePluginCRDExists set to true when the consoleplugins.console.openshift.io has been deployed.
 	ConsolePluginCRDExists bool
+
+	// LooseUDNIsolationModeEnabled set to true when loose isolation mode is enabled between two BGP advertised
+	// UDN networks. In loose isolation mode, those network pods can communicate with each other accoding to
```
arkadeepsen, line 131:
```suggestion
	// UDN networks. In loose isolation mode, those network pods can communicate with each other according to
```
nit

bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
```diff
@@ -430,6 +430,10 @@ spec:
         - name: "NO_PROXY"
           value: "{{ .NO_PROXY}}"
 {{ end }}
+        {{ if .IsLooseUDNIsolationEnabled }}
+        - name: UDN_ISOLATION_MODE
+          value: "loose"
+        {{ end }}
```
jcaamano, line 436: I don't think this matches what https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5276/commits/3dd614922518bb513795f9c1f7951f9e13563ddc expects. The setting there is called
`ROUTED_UDN_ISOLATION` instead of `UDN_ISOLATION_MODE`. Also the values it takes are `Disabled` or `Enabled` not `loose`.

This should preferably be passed as a command line argument in `script-lib` rather than an environment variable for consistency.

I think `UDN_ISOLATION_MODE` is fine, I would probably use `Strict` and `Loose` for values rather than Disabled` or `Enabled`.

pkg/platform/platform.go
```diff
@@ -62,6 +62,22 @@ func isNetworkNodeIdentityEnabled(client cnoclient.Client, infra *bootstrap.Infr
...
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("unable to bootstrap OVN, unable to retrieve udn-config-overrides config: %s", err)
+	}
+	isLooseIsolationEnabled := configMap.Data["force-loose-isolation"]
+	return isLooseIsolationEnabled == "true", nil
```
jcaamano, line 78: Maybe just use the same key/value that we pass to ovnk for consistency? So `UDNRoutedIsolationMode` and `Strict`,`Loose` values? Any non expected value should return error.

pkg/platform/platform.go
jcaamano: I was thinking if we should go more generic with the name and use `ovn-kubernetes-config-overrides` and this be a more generic `getOVNKConfigOverrides` returning a map. Just to prevent us from spreading on config map names.

pliurh: Can we just use the existing `env-overrides` configmap instead of creating a new configmap?

pliurh: I created a new PR https://github.com/openshift/cluster-network-operator/pull/2752 according to @jcaamano 's comments. PTAL.

pperiyasamy:
closing it in favor of https://github.com/openshift/cluster-network-operator/pull/2752.

PR #2715, opened 2025-05-26
[release-4.15] OCPBUGS-56649: Fix live migration with feature migration configured explicitly

pliurh:
The function isMigrationChangeSafe reports error when feature migration configured explicitly. It prevent the live migration from being completed at the last step, when the clusterconfig controller remove the flags spec.migration.networkType and spec.migration.mode.

Manual cherry-pick, resolve the conflicts in pkg/network/render_test.go by removing the func `Test_renderNetworkDiagnostics`.


kyrtapz:
@pliurh is the commit cherry-picked? if you do cherry-pick -x it will preserve the info of the original one.

PR #2716, opened 2025-05-30
CNF-18236: FRRK8s webhook: webhook liveness / readiness from metrics to webhook

fedepaol:
Instead of relying on the metrics port to understand when the webhook is ready, we try to hit the endpoint of the webhook itself, so the webhook pod is going to be ready only when it is effectively able to serve.

This is aligned with the frr-k8s binary change where we add a new endpoint "healtz" to the webhook server.

This is also required for https://github.com/openshift/frr/pull/93 where we disable the metrics endpoint as part of making the webhook hostnetworked.

### DIFF
--- a/bindata/network/frr-k8s/webhook.yaml
+++ b/bindata/network/frr-k8s/webhook.yaml
@@ -72,6 +72,7 @@ spec:
         - --disable-cert-rotation=true
         - --namespace=$(NAMESPACE)
         - --metrics-bind-address=:7572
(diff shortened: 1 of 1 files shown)


fedepaol:
frrk8s lane: 4 failures, 2 node selector (which should not run here) and 2 to be investigated in frr.
Not related to this change.

oribon:
/hold
we might want to enforce the webhook-port flag

PR #2718, opened 2025-06-04
OCPBUGS-56086: Validate NAD name and spec only in multus admission controller

jcaamano:
This should save up calls to the webhook and reduce latency when a NAD is updated to add labels or annotations.

The root cause of the change is the OVNK BGP feature: when BGP is enabled for the cluster default network, reconfiguration might cause temporary disruptions. As part of this reconfiguration and necessary to complete it, OVNK depends on annotating an internal NAD. We want to avoid having to reach the webhook for this annotation because the temporary disruption might prevent it and in that case the reconfiguration won't complete.

Another possibility would be to filter out from validation the specific internal NAD but this current approach might be more beneficial overall.

/cc @kyrtapz @dougbtv

### DIFF
--- a/bindata/network/multus-admission-controller/003-webhook.yaml
+++ b/bindata/network/multus-admission-controller/003-webhook.yaml
@@ -27,6 +27,10 @@ webhooks:
         apiGroups: ["k8s.cni.cncf.io"]
         apiVersions: ["v1"]
         resources: ["network-attachment-definitions"]
+    matchConditions:
+      # On updates, only validate if the Spec changes
+      - name: CreateDeleteOrUpdatedSpec
+        expression: oldObject == null || object == null || object.spec != oldObject.spec
     sideEffects: NoneOnDryRun
     admissionReviewVersions:
     - v1


bindata/network/multus-admission-controller/003-webhook.yaml
```diff
@@ -27,6 +27,10 @@ webhooks:
         apiGroups: ["k8s.cni.cncf.io"]
         apiVersions: ["v1"]
         resources: ["network-attachment-definitions"]
+    matchConditions:
+      - name: 'nameOrSpec'
```
kyrtapz: `nameOrSpec` I don't think you are using the name.

kyrtapz (commented):
I think this change make a lot of sense, especially considering the fact that the webhook only looks at the spec anyway:
https://github.com/openshift/multus-admission-controller/blob/main/pkg/webhook/webhook.go#L104-L157

kyrtapz:
> Another possibility would be to filter out from validation the specific internal NAD but this current approach might be more beneficial overall.

Yet another possibility would be to skip the webhook for NADs generated/updated by OVN-Kubernetes. I don't think that it adds a lot of value most of the time as we are generating the NADs synthetically. This can be considered separately to the BGP issue

jcaamano:
/payload-job periodic-ci-openshift-ovn-kubernetes-release-4.20-periodics-e2e-metal-ipi-ovn-dualstack-bgp-techpreview

jcaamano:
> /payload-job periodic-ci-openshift-ovn-kubernetes-release-4.20-periodics-e2e-metal-ipi-ovn-dualstack-bgp-techpreview

Seem like this isn't gonna work. I am not particularly concerned, BGP is not the only use case annotating NADs. We already annotate them with network IDs and network names.

jcaamano:
/hold

want to bring this up in the NPWG call next week

jcaamano:
/retitle OCPBUGS-56086: Validate NAD name and spec only in multus admission controller

PR #2727, opened 2025-06-18
OCPBUGS-57643: Add release version annotation to whereabouts-controller/whereabouts-token-watcher

bverschueren:
Include a `release.openshift.io/version` annotation on both `whereabouts-controller` deployment and `whereabouts-token-watcher` daemonset. The CNO [compares this against the target version](https://github.com/openshift/cluster-network-operator/blob/2a1d6159a44bf8aa26bba9981f144006a1d99990/pkg/controller/statusmanager/pod_status.go#L114-L116) during an upgrade and never updates its version if the check fails.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -862,6 +862,8 @@ kind: Deployment
 metadata:
   name: whereabouts-controller
   namespace: openshift-multus
+  annotations:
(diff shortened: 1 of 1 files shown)


pliurh:
/lgtm

@kyrtapz PTAL

kyrtapz:
We really need some automation that checks for this...

/lgtm

PR #2730, opened 2025-06-25
CNF-18237: Align frrk8s manifests to upstream

fedepaol:
Align to upstream:

- FRRK8s webhook becomes hostnetworked and we change the port to one in the allowed range in https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md (still need to be registered)
- The DisableMP is deprecated, the default behavior is single stack ipfamily, a dualstackIPFamily flag is added for backward compatibility.

### DIFF
--- a/bindata/network/frr-k8s/webhook.yaml
+++ b/bindata/network/frr-k8s/webhook.yaml
@@ -61,7 +61,7 @@ spec:
         component: frr-k8s-webhook-server
       annotations:
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
-        openshift.io/required-scc: restricted-v2
+        openshift.io/required-scc: privileged
     spec:
       containers:
       - command:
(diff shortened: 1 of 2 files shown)


bindata/network/frr-k8s/webhook.yaml
```diff
@@ -81,12 +80,8 @@ spec:
...
         name: frr-k8s-webhook-server
         ports:
-        - containerPort: 7572
-          name: monitoring
-        - containerPort: 9443
-          name: webhook          
-        securityContext:
-         runAsNonRoot: true
+        - containerPort: 9123
+          name: webhook
```
oribon, line 84: do we need to update this in https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md ?

fedepaol: yep, I'll followup as soon as this gets in

PR #2731, opened 2025-06-26
OCPBUGS-58155: CNO must skip watching IngressController if Ingress capability is disabled in HyperShift

linoyaslan:
This PR updates the CNO to conditionally watch the IngressController resource only when the Ingress capability is enabled. When Ingress is disabled in HyperShift, the IngressController CRD is not available. Without this change, CNO attempts to watch a non-existent resource and repeatedly restarts. This update prevents that by skipping the watch when the capability is disabled.

**Assisted-by:** Cursor

### DIFF
--- a/pkg/controller/ingressconfig/ingressconfig_controller.go
+++ b/pkg/controller/ingressconfig/ingressconfig_controller.go
@@ -2,14 +2,17 @@ package ingressconfig
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"time"
 
 	operv1 "github.com/openshift/api/operator/v1"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/controller/statusmanager"
+	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"k8s.io/apimachinery/pkg/api/meta"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -46,8 +49,55 @@ func newIngressConfigReconciler(client crclient.Client, status *statusmanager.St
 	return &ReconcileIngressConfigs{client: client, status: status}
 }
 
+// isIngressCapabilityEnabled checks if the Ingress capability is enabled in the cluster
+func isIngressCapabilityEnabled(client crclient.Client) (bool, error) {
+	hcpCfg := hypershift.NewHyperShiftConfig()
+	return isIngressCapabilityEnabledWithConfig(client, hcpCfg)
+}
+
+// isIngressCapabilityEnabledWithConfig checks if the Ingress capability is enabled in the cluster
+func isIngressCapabilityEnabledWithConfig(client crclient.Client, hcpCfg *hypershift.HyperShiftConfig) (bool, error) {
+	// Handle nil client
+	if client == nil {
+		return false, fmt.Errorf("client cannot be nil")
+	}
+
+	// Check if this is a hypershift cluster
+	if !hcpCfg.Enabled {
+		// For non-hypershift clusters, assume ingress capability is always enabled
+		return true, nil
+	}
+
+	// For hypershift clusters, check if the IngressController CRD exists
+	gvk := operv1.SchemeGroupVersion.WithKind("IngressController")
+	mapping, err := client.RESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		if meta.IsNoMatchError(err) {
+			log.Printf("IngressController CRD not found, assuming Ingress capability is disabled: %v", err)
+			return false, nil
+		}
+		return false, err
+	}
+
+	// If we got a mapping, the CRD exists
+	log.Printf("IngressController CRD found, Ingress capability is enabled (Resource: %v)", mapping.Resource)
+	return true, nil
+}
+
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r *ReconcileIngressConfigs) error {
+	// Check if Ingress capability is enabled before setting up the controller
+	ingressEnabled, err := isIngressCapabilityEnabled(mgr.GetClient())
+	if err != nil {
+		log.Printf("Error checking if Ingress capability is enabled: %v", err)
+		return err
+	}
+
+	if !ingressEnabled {
+		log.Printf("Ingress capability is disabled, skipping ingress-config-controller creation entirely")
+		return nil
+	}
+
 	// create a controller and register watcher for ingresscontroller resource
 	c, err := controller.New("ingress-config-controller", mgr, controller.Options{Reconciler: r})
 	if err != nil {
--- a/pkg/controller/ingressconfig/ingressconfig_controller_test.go
+++ b/pkg/controller/ingressconfig/ingressconfig_controller_test.go
@@ -1,0 +1,126 @@
+package ingressconfig
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/cluster-network-operator/pkg/client/fake"
+	"github.com/openshift/cluster-network-operator/pkg/hypershift"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func TestIsIngressCapabilityEnabled(t *testing.T) {
+	testCases := []struct {
+		name              string
+		hypershiftEnabled bool
+		crdExists         bool
+		restMapperError   error
+		expectedResult    bool
+		expectedError     bool
+	}{
+		{
+			name:              "Non-hypershift cluster should always return true (IngressController CRD doesn't exists)",
+			hypershiftEnabled: false,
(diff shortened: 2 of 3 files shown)


pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -46,8 +48,48 @@ func newIngressConfigReconciler(client crclient.Client, status *statusmanager.St
 	return &ReconcileIngressConfigs{client: client, status: status}
 }
 
+// isIngressCapabilityEnabled checks if the Ingress capability is enabled in the cluster
+func isIngressCapabilityEnabled(client crclient.Client) (bool, error) {
+	// Handle nil client
+	if client == nil {
```
kyrtapz, line 61: This is quite defensive. While I don't think it is necessary I am fine to keep it :) keeps AI calm :laughing:

pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -46,8 +48,48 @@ func newIngressConfigReconciler(client crclient.Client, status *statusmanager.St
...
+
+	// If we got a mapping, the CRD exists
+	log.Printf("IngressController CRD found, Ingress capability is enabled (Resource: %v)", mapping.Resource)
+	return true, nil
+}
+
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r *ReconcileIngressConfigs) error {
+	// Check if Ingress capability is enabled before setting up the controller
+	ingressEnabled, err := isIngressCapabilityEnabled(mgr.GetClient())
```
wking, line 90: Checking at container-start like this helps, but [capabilities can be enabled at any point after cluster install][1], so if you only check at container-start time, you're exposed to situations like:

1. Cluster is running without the `Ingress` capability.
2. Network operator comes up, notices the lack of `IngressController`, and skips its `ingress-config-controller`.
3. Cluster-admin enables the `Ingress` capability.
4. `IngressController` CustomResourceDefinition created.
5. Here you'd want the network operator to notice, and start running its `ingress-config-controller`.  But with your current code, you'll only notice whenever the network operator container restarts, which on a quiet, rarely-updated cluster, could be an unbounded amount of time.

I'm not networking to know how important it is to run the `ingress-config-controller`; maybe (5) is an acceptable risk?  But the alternative is to build a CRD-watcher or `isIngressCapabilityEnabled` poller, or some such that somehow gets `ingress-config-controller` running if later cluster changes lead to `IngressController` existing.  The easiest way to get the controller running might just be logging "I see that `IngressController` exists now!" and exiting 0, so the container runtime creates a new container, which would see `IngressController` on start-up, and work with your current start-up-time detection.

[1]: https://github.com/openshift/enhancements/blob/0f916a52af1a6fbdab0c5b80ae0e66c7a27efb6a/enhancements/installer/component-selection.md#capabilities-can-be-installed

linoyaslan: Thanks for the review @wking!

From what I see, in HyperShift the spec.capabilities field is immutable after the HostedCluster is created. I tried editing and patching it to re-enable Ingress, but got this error:

`Capabilities is immutable. Changes might result in unpredictable and disruptive behavior.`

So maybe in regular OCP it’s possible to enable capabilities post-install? but in HyperShift it doesn’t seem allowed (Or maybe there is a way to do that which I'm not familiar with?)
That makes me think your suggestion about watching for IngressController becoming available might not apply here.

Let me know what you think!

linoyaslan: **Note**: I focused on HyperShift because disabling Ingress isn’t  supported in standalone OpenShift.
https://github.com/openshift/api/blob/master/config/v1/types_cluster_version.go#L394

pkg/controller/ingressconfig/ingressconfig_controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+			crdExists:      true, // doesn't matter for non-hypershift
+			expectedResult: true,
+			expectedError:  false,
+		},
+	}
+
+	// Note: We can't easily test hypershift scenarios because the hypershift detection
+	// is based on package-level variables that are set at initialization time.
+	// In a real hypershift environment, the HYPERSHIFT env var would be set before
+	// the process starts, so the package would initialize correctly.
```
ricky-rav: So here in `pkg/controller/ingressconfig/ingressconfig_controller_test.go` we're only testing a non-hypershift scenario, where ingress can be only enabled, so `isIngressCapabilityEnabled=true`. Is that right? If so, let's say that explicitly. Can you change `we can't easily test scenarios` with `we're not testing hypershift scenarios`, so that when we go back to this months from now it'll be crystal clear? Thanks! :)

ricky-rav (changes requested):
Just a nit, all the rest looks good.

pkg/controller/ingressconfig/ingressconfig_controller_test.go
linoyaslan: @ricky-rav  Done :) Thank you!

ricky-rav:
@kyrtapz PTAL

pkg/controller/ingressconfig/ingressconfig_controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/openshift/cluster-network-operator/pkg/client/fake"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// mockRESTMapper is a fake RESTMapper for testing
+type mockRESTMapper struct {
```
kyrtapz: NIT: Can you extract the mock into a separate file?

pkg/controller/ingressconfig/ingressconfig_controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+	// In non-hypershift mode, this should still return true regardless of REST mapper errors
+	result, err = isIngressCapabilityEnabled(mockClient)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(result).To(BeTrue()) // Non-hypershift always returns true
+}
+
+// TestIsIngressCapabilityEnabledIntegrationNote documents the limitation of unit testing
+// hypershift scenarios and explains how this would work in practice
+func TestIsIngressCapabilityEnabledIntegrationNote(t *testing.T) {
+	t.Log("=== Integration Testing Note ===")
```
kyrtapz: This can be easily fixed by passing the hypershit config to `isIngressCapabilityEnabledWithConfig`.
Can you please do that?

pkg/controller/ingressconfig/ingressconfig_controller.go
kyrtapz: @linoyaslan can you provide the Hypershift API validation that prevents capabilities editing?

kyrtapz (changes requested):
Thanks @linoyaslan! 
Are there any CI lanes testing disabling capabilities in HyperShift?

pkg/controller/ingressconfig/ingressconfig_controller.go
linoyaslan: @kyrtapz https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L730

linoyaslan:
> Thanks @linoyaslan!
Are there any CI lanes testing disabling capabilities in HyperShift?

Hi, yes. for each disabled capability we implement, we write a dedicated e2e test named Ensure<Cap-name>CapabilityDisabled, and it runs as part of the e2e-aws CI job.

linoyaslan:
> @linoyaslan can you provide the Hypershift API validation that prevents capabilities editing?

https://github.com/openshift/hypershift/blob/main/api/hypershift/v1beta1/hostedcluster_types.go#L730

linoyaslan:
@ricky-rav @kyrtapz Hi, I’ve pushed new changes based on @kyrtapz’s review. I’d appreciate it if you could take another look. Thanks!

kyrtapz:
/hold
one more comment.

pkg/controller/ingressconfig/ingressconfig_controller.go
```diff
@@ -46,8 +48,53 @@ func newIngressConfigReconciler(client crclient.Client, status *statusmanager.St
...
+	// Check if this is a hypershift cluster
+	if !hcpCfg.Enabled {
+		// For non-hypershift clusters, assume ingress capability is always enabled
+		return true, nil
+	}
+
+	// For hypershift clusters, check if the IngressController CRD exists
+	gvk := operv1.SchemeGroupVersion.WithKind("IngressController")
+	mapping, err := client.RESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
```
kyrtapz, line 74: This should actually check if the error is what you expect for a missing rest mapping and if it is not the function should return an error.

linoyaslan: Agree. let me fix it

linoyaslan: @kyrtapz  Done. Hope it's what you meant to

kyrtapz:
@linoyaslan you will need to [fix](https://github.com/openshift/cluster-network-operator/pull/2731#issuecomment-3008529382) the attached bug.

kyrtapz:
e2e-aws-ovn-upgrade is really flaky, the changes in this PR are only affecting hypershift deployments and CNO logs suggest no issues.

PR #2738, opened 2025-07-10
OCPBUGS-57354: block upgrade if a CNI migration is in progress

jluhrsen:


jluhrsen:
/jira refresh
/hold
want to test locally on a real cluster upgrade/migration

jluhrsen:
/retest
/hold cancel
I did some local testing and this looks like it's working as expected
/assign @kyrtapz 
/assign @pliurh

pliurh:
@jluhrsen Thanks for this PR. We only support CNI migration in releases 4.16 and earlier. We plan to remove all migration code from CNO for 4.17+ releases. So it makes no sense to create a patch on master and backport for this feature. Could you propose this change to 4.16 branch?

jluhrsen:
> @jluhrsen Thanks for this PR. We only support CNI migration in releases 4.16 and earlier. We plan to remove all migration code from CNO for 4.17+ releases. So it makes no sense to create a patch on master and backport for this feature. Could you propose this change to 4.16 branch?

@pliurh , thanks for the info. did not know this.

[here](https://github.com/openshift/cluster-network-operator/pull/2760) is the 4.16 version

PR #2739, opened 2025-07-10
OCPBUGS-58306: Disable network segmentation when multus is disabled

ricky-rav:
Network segmentation requires multus to be enabled. If the user disables multus (multi networks), we should also disable network segmentation. 
Since network segmentation is on by default, we have to disable it explicitly if we find `disableMultiNetwork=true`


ricky-rav:
@kyrtapz PTAL

kyrtapz:
@ricky-rav @anuragthehatter do we have any coverage for `disableMultiNetwork=true`?
Do we need to add a CI job?

ricky-rav:
> @ricky-rav @anuragthehatter do we have any coverage for `disableMultiNetwork=true`? Do we need to add a CI job?

There's nothing in origin for `disableMultiNetwork`. Do we care enough about this to have a dedicated lane for it?

PR #2754, opened 2025-07-28
MGMT-21314: CNO enable advanced gateway detection in ovnkube in dpu host mode

tsorya:
MGMT-21314: CNO enable advanced gateway detection in ovnkube in dpu host mode

Adding required gateway value for ovnk in dpu-host mode This commit enables usage of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5327/

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -512,10 +512,19 @@ data:
 
       echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node"
 
+
+      if [ "{{.OVN_NODE_MODE}}" == "dpu-host" ]; then
+        // this is required for the dpu-host mode to configure right gateway interface
+        // https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5327/files
+        gateway_interface=derive-from-mgmt-port
+      else
+        gateway_interface=br-ex
+      fi
+
       if [ "{{.OVN_GATEWAY_MODE}}" == "shared" ]; then
-        gateway_mode_flags="--gateway-mode shared --gateway-interface br-ex"
+        gateway_mode_flags="--gateway-mode shared --gateway-interface ${gateway_interface}"
       elif [ "{{.OVN_GATEWAY_MODE}}" == "local" ]; then
-        gateway_mode_flags="--gateway-mode local --gateway-interface br-ex"
+        gateway_mode_flags="--gateway-mode local --gateway-interface ${gateway_interface}"
       else
         echo "Invalid OVN_GATEWAY_MODE: \"{{.OVN_GATEWAY_MODE}}\". Must be \"local\" or \"shared\"."
         exit 1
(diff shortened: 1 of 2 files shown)


bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -512,10 +512,21 @@ data:
 
       echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node"
 
+      # Use OVN_NODE_MODE environment variable, default to "full" if not set
+      OVN_NODE_MODE=${OVN_NODE_MODE:-full}
+      if [ "${OVN_NODE_MODE}" == "dpu-host" ]; then
```
wizhaoredhat, line 519: I expect to see logic surrounding dpu mode as well. But I only see dpu-host and full mode

tsorya: I know but we don't need it for TP and i am not fully aware how it will look when we will run with 2 ovnks, prefer to add minimal change for now

wizhaoredhat:
We need to be careful of release cycles if DPF Operator takes over OVN-K

DPF Operator is not built by RH so we can't control the release cycle. 
If OVN-K API moves then it would be very difficult to backport API changes or not allow those API changes. Because it maybe that the DPF Operator is using a certain commits of OVN-K. DPF Operator may not backport certain changes that are needed as well.

I suggest we also have some control of the OVN-K daemonset running in DPU mode. Maybe tell the DPF Operator that it is externally managed?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
ricky-rav: Uhm... so at least we should leave a TODO message in a comment.

ricky-rav (changes requested):
The second commit should describe that you're adding `OVN_NODE_MODE` to the yamls.  
As for for `OVN_NODE_MODE=dpu`, I'd be happier if we could propagate it to OVNK, even if not yet tested, but at the very least we should have a TODO comment saying it's missing for now.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
tsorya: I am not sure we will actually need it to say the truth. Really depends on how it will go further, that's why i didn't add anything regarding smart nic mode as it is not relevant for now and not sure it will be in the future

ricky-rav: Fine by me! @wizhaoredhat what do you think?

wizhaoredhat: I would prefer there was some documentation of the assumptions we are making here.
- So when the next person reads the code, they won't be confused with different components outside controlling the DPU mode.
If not TODO, at least some documentation explaining the design.

tsorya: Let me write comment in this case.

tsorya:
> The second commit should describe that you're adding `OVN_NODE_MODE` to the yamls. As for for `OVN_NODE_MODE=dpu`, I'd be happier if we could propagate it to OVNK, even if not yet tested, but at the very least we should have a TODO comment saying it's missing for now.

We propagate it, this is part of the change 
https://github.com/openshift/cluster-network-operator/pull/2754/files#diff-189477d33f414b8df5d55f565d29f4d25fed670e5c055e8952206dfedfd4153fR521

Let write a description for the second commit in commit message

PR #2757, opened 2025-07-31
CORENET-6247: Set runAsUser for ovnkube-control-plane

bradbehle:
Now that IBM ROKS is working on OVN support for Hypershift cluster, we need to run the ovnkube-control-plane pods as a non-root user if at all possible.  This change will run the ovnkube-control-plane pods as the RUN_AS_USER from the cluster-network-operator the same way that multus-admission-controller pods currently do.

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -255,16 +255,6 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	} else {
 		data.Data["OVNPlatformAzure"] = false
 	}
-	if bootstrapResult.Infra.PlatformType == configv1.IBMCloudPlatformType { // TODO: Find a way to exclude IPI clusters from this???
-		// OVN is only supported in IBMCloud on Hypershift clusters, so get RunAsUser from HyperShiftConfig
-		if bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.RunAsUser != "" {
-			data.Data["OVNRunAsUser"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.RunAsUser
-		} else {
-			data.Data["OVNRunAsUser"] = "1001" // We do not want to run this as root, so if nothing is specified, use 1001
-		}
-	} else {
-		data.Data["OVNRunAsUser"] = ""
-	}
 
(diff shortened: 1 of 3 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -255,6 +255,16 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	} else {
 		data.Data["OVNPlatformAzure"] = false
 	}
+	if bootstrapResult.Infra.PlatformType == configv1.IBMCloudPlatformType { // TODO: Find a way to exclude IPI clusters from this???
```
rtheis: I think this need to come from the HyperShift hosted control plane configuration for any cloud provider when it is set. This is how the multus admission controller works. Doing this will ensure that IPI clusters are not changed.

bradbehle: Ok.  I updated this to apply to all providers to match what is being done with multus admission controller

rtheis (approved):
/lgtm

bradbehle:
I checked the results of the two failing required tests, and the failures do not look like they are at all related to this change, so I triggered a retest of those.

kyrtapz:
@bradbehle the changes are not IBMCloud specific, can you update the PR title/description?
@csrwng do we have CI coverage for `RunAsUser` in the hypershift conformance jobs?

kyrtapz:
As a side note, I think it should be possible to set
```
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
```
on the ovn-k controllplane in general, lmk if that is something you can try to include as part of this effort.

bradbehle:
/retitle CORENET-6247: Set runAsUser for ovnkube-control-plane

rtheis:
/retitle CORENET-6247: Set runAsUser for ovnkube-control-plane

bradbehle:
@kyrtapz I took your suggestion and added

```
        runAsNonRoot: true
        seccompProfile:
          type: RuntimeDefault
```

to the deployment as well.  I only did this when the RunAsUser was set, because @rtheis and I think at least the `runAsNonRoot: true` would cause the pod to crash otherwise, because the image likely is set to run as the root user by default.  Let me know what you think.

kyrtapz:
Thanks @bradbehle @rtheis 
/approve

PR #2725, opened 2025-06-16
NO-JIRA: remove unused clusterrolebinding default-account-cluster-network-operator

lance5890:
1. The default-account-cluster-network-operator ClusterRoleBinding only worke when upgrade from 4.15
2. cleanup the unused yaml in newest(4.20) version
3. kind of subsequent optimization measures of https://github.com/openshift/cluster-network-operator/pull/2084

address https://github.com/openshift/cluster-version-operator/issues/1207

PR #2758, opened 2025-08-01
Configure CVO to delete default-account role binding

clarku:
Previously the default-account-cluster-network-operator ClusterRoleBinding wouldn't be deleted because it had invalid annotations of

>  include.release.openshift.io/(profile): "false"

in #2084 .  

This manifest was removed in https://github.com/openshift/cluster-network-operator/pull/2725, but this PR reverts that change by adding this ClusterRoleBinding back to the CVO manifests with fixed annotations so that it will be deleted properly. The plan is to reapply the changes in that other PR once this change has been in a full version of OpenShift.


jcaamano:
@clarku do you have access to any reference documentation that states `include.release.openshift.io/(profile): "false"` is an invalid annotation and how that in any case prevents the binding to be deleted?

jcaamano:
@lance5890 did you have any chance to crosscheck that the binding was indeed removed before pushing #2725? 

Is the problem that we are hitting here platform specific?

clarku:
@jcaamano, 

Thanks for taking a look at this PR. After your comments, I took another look at this. 

Firstly, for reference, here is the [enhancement](https://github.com/openshift/enhancements/blob/a271717ac0ddec489caf3ee1caefd0a07bb77802/enhancements/update/cluster-profiles.md#cluster-profile-annotation) for cluster profiles.

Regarding your specific question, I think I was slightly incorrect earlier when I referred to `include.release.openshift.io/(profile): "false"` as an "invalid" annotation. It is probably more accurate to say that it is redundant as leaving the annotation out would have the same effect as using "false". This can be seen in the code references below. 

I believe the real issue is that when the `release.openshift.io/delete: "true"` annotation is used to delete a resource, the `include.release.openshift.io/(profile): "true"` annotation would also have to be set for the relevant cluster profiles. If this isn't done, then the manifest would be excluded from the reconcilation process for those profiles, and the deletion wouldn't go through. From testing, I can see the message of this exclusion in the Cluster Version Operator (CVO) logs with the default cluster profile (self-managed-high-availability) and the current manifest. 
~~~
payload.go:210] excluding Filename: "0000_70_cluster-network-operator_02_rbac.yaml" Group: "rbac.authorization.k8s.io" Kind: "ClusterRoleBinding" Name: "default-account-cluster-network-operator": unrecognized value include.release.openshift.io/self-managed-high-availability=false
~~~
This error is from [here](https://github.com/openshift/cluster-version-operator/blob/9de00ba8fb8330b3a1b440396dc2c5dbe974119b/vendor/github.com/openshift/library-go/pkg/manifest/manifest.go#L226-L233) and called by [this](https://github.com/openshift/cluster-version-operator/blob/9de00ba8fb8330b3a1b440396dc2c5dbe974119b/pkg/payload/payload.go#L214-L224).

lance5890:
> @jcaamano,
> 
> Thanks for taking a look at this PR. After your comments, I took another look at this.
> 
> Firstly, for reference, here is the [enhancement](https://github.com/openshift/enhancements/blob/a271717ac0ddec489caf3ee1caefd0a07bb77802/enhancements/update/cluster-profiles.md#cluster-profile-annotation) for cluster profiles.
> 
> Regarding your specific question, I think I was slightly incorrect earlier when I referred to `include.release.openshift.io/(profile): "false"` as an "invalid" annotation. It is probably more accurate to say that it is redundant as leaving the annotation out would have the same effect as using "false". This can be seen in the code references below.
> 
> I believe the real issue is that when the `release.openshift.io/delete: "true"` annotation is used to delete a resource, the `include.release.openshift.io/(profile): "true"` annotation would also have to be set for the relevant cluster profiles. If this isn't done, then the manifest would be excluded from the reconcilation process for those profiles, and the deletion wouldn't go through. From testing, I can see the message of this exclusion in the Cluster Version Operator (CVO) logs with the default cluster profile (self-managed-high-availability) and the current manifest.
> 
> ```
> payload.go:210] excluding Filename: "0000_70_cluster-network-operator_02_rbac.yaml" Group: "rbac.authorization.k8s.io" Kind: "ClusterRoleBinding" Name: "default-account-cluster-network-operator": unrecognized value include.release.openshift.io/self-managed-high-availability=false
> ```
> 
> This error is from [here](https://github.com/openshift/cluster-version-operator/blob/9de00ba8fb8330b3a1b440396dc2c5dbe974119b/vendor/github.com/openshift/library-go/pkg/manifest/manifest.go#L226-L233) and called by [this](https://github.com/openshift/cluster-version-operator/blob/9de00ba8fb8330b3a1b440396dc2c5dbe974119b/pkg/payload/payload.go#L214-L224).

yes, I think @clarku find the issue that https://github.com/openshift/cluster-network-operator/pull/2084 has not set the `true` annotation value, this will not add the yaml into cvo Appropriately, and i forgot to double check the file too

manifests/0000_70_cluster-network-operator_02_rbac.yaml
```diff
@@ -24,3 +24,21 @@ roleRef:
...
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/single-node-developer: "true"
+subjects:
+- kind: ServiceAccount
+  name: default
+  namespace: openshift-network-operator
+roleRef: 
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
```
jcaamano: Can you add a new line?

clarku: Done

jcaamano:
/lgtm
/approve

Should this be considered a critical fix? If so, why?

clarku:
> Should this be considered a critical fix? If so, why?

I don't think so and here is my reasoning. 

- There is a easy workaround by deleting the role binding. 
- This has no direct impact on the cluster. 
- Although this bug means that there can be a unused role binding that binds to cluster admin, I don't believe that this represents a major security risk by itself.

PR #2760, opened 2025-08-01
OCPBUGS-57354: block upgrade if a CNI migration is in progress

jluhrsen:


jluhrsen:
@pliurh , PTAL

pkg/controller/statusmanager/status_manager.go
```diff
@@ -379,8 +380,37 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 			)
 		}
 
-		if oc.Spec.DefaultNetwork.Type == operv1.NetworkTypeOpenShiftSDN {
-			// OpenShiftSDN is removed in 4.17, so block the upgrade if we have OpenShiftSDN in the spec.
+		migrationInProgress := false
+		netCfg := &configv1.Network{}
+		if err := status.client.ClientFor("").CRClient().
+			Get(context.TODO(), types.NamespacedName{Name: names.OPERATOR_CONFIG}, netCfg); err == nil {
```
pliurh: If `err != nil`, we should return it instead of ignoring it.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -379,8 +380,37 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
...
+			migrationInProgress = meta.IsStatusConditionPresentAndEqual(
+				netCfg.Status.Conditions,
+				names.NetworkTypeMigrationInProgress,
+				metav1.ConditionTrue,
+			)
+		}
+
+		if oc.Spec.DefaultNetwork.Type == operv1.NetworkTypeOpenShiftSDN && migrationInProgress {
```
pliurh: I think we shall remove the check `oc.Spec.DefaultNetwork.Type == operv1.NetworkTypeOpenShiftSDN`. Upgrade shall be blocked when `migrationInProgress` is true.

jluhrsen: done

jluhrsen:
thanks for the review @pliurh . I think I addressed both of your comments. PTAL

jluhrsen:
/hold
Looks like I broke UT

jluhrsen:
/retest

@pliurh I think this is ready for review now

/hold cancel

jluhrsen:
/retest

@pliurh the windows job is failing every time right now, but doesn't look like the same reason each time. I don't see the 4.16 windows job running anywhere else so not sure if the job is totally broken or not. Any thoughts?

pkg/controller/statusmanager/status_manager.go
```diff
@@ -379,8 +382,29 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 			)
 		}
 
-		if oc.Spec.DefaultNetwork.Type == operv1.NetworkTypeOpenShiftSDN {
-			// OpenShiftSDN is removed in 4.17, so block the upgrade if we have OpenShiftSDN in the spec.
+		netCfg := &configv1.Network{}
+		if err := status.client.ClientFor("").CRClient().
+			Get(context.Background(), types.NamespacedName{Name: names.OPERATOR_CONFIG}, netCfg); err != nil {
+			if !apierrors.IsNotFound(err) {
```
pliurh: Why not return for other kinds of error?

jluhrsen: if it's catchall, returning for all errors, the UTs break since this code path gets executed and there is not really even a cluster available.

do you want me to return for all errors and figure out some fix for the UTs?

pliurh: Yes, as long as it doesn't cost too much effort.

jluhrsen: @pliurh I just updated this to account for your review. can you PTAL when you have a chance?

jluhrsen:
@kyrtapz I think we are good here

jluhrsen:
@pliurh the [windows job is in horrible state](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-network-operator-release-4.16-e2e-aws-ovn-windows). can you `/override` it please?

the only place it runs is in CNO so I doubt anyone even knows or cares about it.

PR #2764, opened 2025-08-05
OCPBUGS-60126: HCP: fix handling of tolerations with no key/value

csrwng:
Fixes rendering of tolerations that do not include a key/value.
Ref: [OCPBUGS-60126](https://issues.redhat.com/browse/OCPBUGS-60126)

### DIFF
--- a/pkg/hypershift/hypershift.go
+++ b/pkg/hypershift/hypershift.go
@@ -331,19 +331,52 @@ func tolerationsToStringSliceYaml(tolerations []corev1.Toleration) ([]string, er
 		return nil, nil
 	}
 
-	yamlBytes, err := yaml.Marshal(tolerations)
-	if err != nil {
-		return nil, err
-	}
+	var result []string
 
-	yamlStrs := []string{}
-	for _, arg := range strings.Split(string(yamlBytes), "\n") {
-
-		// filter out null and empty strings
-		if strings.Contains(arg, ": null") || strings.Contains(arg, ": \"\"") {
-			continue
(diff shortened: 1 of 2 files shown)


kyrtapz:
@csrwng what was the failure? Is the yaml marshaller failing to output a correct yaml? 
I am not a huge fan of having to parse the tolerations... Swapping the yaml parser to `sigs.k8s.io/yaml` seems to be giving more promising results but maybe we can simplify it even further considering we parse annotations from a string to then marshall it back to a yaml...

csrwng:
@kyrtapz so what was happening is that we were doing the serialization in one shot for all tolerations, and then removing any lines that contained nil or empty string value. So for a no key/value toleration, the initial serialization looked something like:
```
- key: ""
  operator: Exists
  value: ""
```

Then we removed the lines with empty strings, and ended up with only:
```
  operator: Exists
```
which is not valid YAML list syntax

csrwng:
I simply changed serialization to a single toleration at a time, instead of the entire slice, and prepending the proper '- ' before each entry.

kyrtapz:
@csrwng [verify-deps](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2764/pull-ci-openshift-cluster-network-operator-master-verify-deps/2008909643396419584) is failing, are we missing `go mod tidy`?

ricky-rav:
@csrwng it's been a long time, do we still care about this PR?

PR #2752, opened 2025-07-28
OCPBUGS-55962: Allow overriding OVN-Kubernetes configuration

pliurh:
This change introduces a mechanism to provide hidden or experimental configuration to OVN-Kubernetes through a ConfigMap. This provides a flexible way to enable advanced or experimental OVN-Kubernetes features without modifying the CNO API.

A new ConfigMap 'ovn-kubernetes-config-overrides', is now read by the operator during the bootstrap process. The key-value data from this ConfigMap is passed down to the OVN-K components.

As an initial implementation, this is used to support the flag 'udn-isolation-mode' which is to enable user to select the UDN isolation mode between 'loose' and 'strict'. If unset, the default mode is 'strict'.

PR #2769, opened 2025-08-06
CORENET-6276: Use priorityClass from HostedControlPlane for OVN and NNI

bradbehle:
priorityClass can be specified in the HostedControlPlane for hypershift managed clusters.  When it is specified, it should be used by the control plane components.  Currently the cluster-network-operator deployment and the multus-admission-controller deployment do use the priority class that is specified, but the ovnkube-control-plane and network-node-identity deployments do not, they use a hardcoded value.

This change updates the ovnkube-control-plane and network-node-identity deployments to use teh priorityClass from the HostedControlPlane if it is specified

### DIFF
--- a/pkg/network/ovn_kubernetes.go
+++ b/pkg/network/ovn_kubernetes.go
@@ -234,7 +234,6 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["HCPNodeSelector"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPNodeSelector
 	data.Data["HCPLabels"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPLabels
 	data.Data["HCPTolerations"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPTolerations
-	data.Data["PriorityClass"] = bootstrapResult.Infra.HostedControlPlane.PriorityClass
 	data.Data["OVN_NB_INACTIVITY_PROBE"] = nb_inactivity_probe
 	data.Data["OVN_CERT_CN"] = OVN_CERT_CN
 	data.Data["OVN_NORTHD_PROBE_INTERVAL"] = os.Getenv("OVN_NORTHD_PROBE_INTERVAL")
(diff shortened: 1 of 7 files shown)


pkg/network/ovn_kubernetes.go
```diff
@@ -234,6 +234,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	data.Data["HCPNodeSelector"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPNodeSelector
 	data.Data["HCPLabels"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPLabels
 	data.Data["HCPTolerations"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.HCPTolerations
+	data.Data["PriorityClass"] = bootstrapResult.Infra.HostedControlPlane.PriorityClass
```
rtheis: I think this needs to be `bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.PriorityClass` to match the rest of the code and may explain why PR tests are failing.

bradbehle: I changed the code as you suggested and added the PriorityClass to the OVNHyperShiftBootstrapResult struct and used it from there.

pkg/network/ovn_kubernetes.go
```diff
@@ -420,6 +420,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 		productFlavor = "managed"
 		data.Data["CAConfigMap"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.CAConfigMap
 		data.Data["CAConfigMapKey"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.CAConfigMapKey
+		data.Data["PriorityClass"] = bootstrapResult.OVN.OVNKubernetesConfig.HyperShiftConfig.PriorityClass
```
rtheis: It seems a bit odd to set this here rather than where `data.Data["HCPTolerations"]` is set but I don't think we get different results. I'll defer to Red Hat on the correct thing to do.

rtheis (approved):
/lgtm

kyrtapz:
> priorityClass can be specified in the HostedControlPlane for hypershift managed clusters. When it is specified, it should be used by the control plane components. Currently the cluster-network-operator deployment and the multus-admission-controller deployment do use the priority class that is specified, but the ovnkube-control-plane and network-node-identity deployments do not, they use a hardcoded value.
> 
> This change updates the ovnkube-control-plane and network-node-identity deployments to use teh priorityClass from the HostedControlPlane if it is specified

@bradbehle @csrwng  Don't we also need it for cloud-network-config-controller? It is currently hard-coded as: `hypershift-control-plane`

pkg/network/ovn_kubernetes.go
kyrtapz: Yeah, @bradbehle would you mind moving all 3 vars to the block that [sets the HCP vars from bootstrap](https://github.com/openshift/cluster-network-operator/blob/bb745971fceb366de50a75b3f373850af230fea1/pkg/network/ovn_kubernetes.go#L228)?

kyrtapz:
https://issues.redhat.com/browse/OCPBUGS-60455

kyrtapz:
/hold
We need to get https://github.com/openshift/cluster-network-operator/pull/2752 in first.
I will unhold it once that is done and try to help merge this one quickly.

kyrtapz:
This is a hypershift specific change, overrding unrelated jobs (they passed in previous runs):
/override ci/prow/e2e-gcp-ovn-upgrade
/override ci/prow/e2e-metal-ipi-ovn-dualstack-bgp

kyrtapz:
https://redhat-external.slack.com/archives/C01C8502FMM/p1755787091649519

PR #2778, opened 2025-08-17
MGMT-21485: Enable dpu-host mode that matches DPF requirements

tsorya:
MGMT-21485: Implement per-node feature enforcement via OVN_NODE_MODE
    
    This commit introduces OVN_NODE_MODE environment variable to enable per-node
    feature enforcement, particularly for DPU host mode where certain features
    must be disabled regardless of cluster-wide configuration.
    
    - Move feature toggles from ConfigMap (004-config.yaml) to startup script (008-script-lib.yaml)
    - ConfigMap values cannot be reliably overridden per-node, but script logic can be conditional
    - Implement OVN_NODE_MODE-based conditional feature enablement in startup script
    
    - Add 'dpu-host' mode that automatically disables incompatible features:
      - Egress IP and related features (egress firewall, egress QoS, egress service)
      - Multicast support
      - Multi-network policies and admin network policies
      - Network segmentation features
    - Set gateway_interface='derive-from-mgmt-port' for DPU host nodes
    - Add ovnkube_node_mode='--ovnkube-node-mode dpu-host' flag
    
    From bindata/network/ovn-kubernetes/*/004-config.yaml:
    - enable-egress-ip=true
    - enable-egress-firewall=true
    - enable-egress-qos=true
    - enable-egress-service=true
    - enable-multicast=true
    - enable-multi-network=true (conditionally)
    - enable-admin-network-policy=true (conditionally)
    - enable-network-segmentation=true (conditionally)
    
    - Add conditional blocks in 008-script-lib.yaml based on OVN_NODE_MODE
    - Full mode (default): All features enabled as configured
    - DPU host mode: Incompatible features force-disabled
    - Maintain backward compatibility for existing deployments
    
    - Add comprehensive TestOVNKubernetesScriptLibCombined test covering:
      - DPU host mode feature gating and disabling
      - Full mode with multi-network features enabled/disabled
      - Non-mode-gated features (route advertisements, DNS resolver, etc.)
      - Gateway interface variable usage validation
    - Remove redundant individual test functions after consolidation
    - Update existing config rendering tests for new ConfigMap content
    
    - Create docs/ovn_node_mode.md with detailed technical explanation
    - Update docs/operands.md with OVN-Kubernetes node modes section
    - Update docs/architecture.md with per-node configuration explanation
    - Update README.md with DPU host mode support information
    - Add implementation details, feature mapping tables, and migration notes
    
    ConfigMap-based feature control cannot be overridden per-node, making it
    impossible to disable features on specific node types (like DPU hosts) while
    keeping them enabled cluster-wide. Moving the logic to startup scripts allows
    the same cluster configuration to work across heterogeneous node types.
    
    This change ensures that DPU host nodes automatically have incompatible
    features disabled, preventing runtime failures and enabling mixed-mode
    cluster deployments.
    
    - Existing clusters continue to work without changes
    - Default behavior (full mode) remains unchanged
    - Migration is automatic during upgrade process
    - No manual intervention required

    MGMT-21314: CNO enable advanced gateway detection in ovnkube in dpu host mode
    Adding required gateway value for ovnk in dpu-host mode
    This commit enables usage of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5327/


bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -512,13 +512,31 @@ data:
 
       echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node"
 
+      # enable egress ip, egress firewall, egress qos, egress service
+      egress_ip_enable_flag="--enable-egress-ip=true --enable-egress-firewall=true --enable-egress-qos=true --enable-egress-service=true"
```
ricky-rav: maybe `egress_features_enable_flag`?

bindata/network/ovn-kubernetes/managed/004-config.yaml
```diff
@@ -146,12 +133,6 @@ data:
 {{- if .OVN_PRE_CONF_UDN_ADDR_ENABLE }}
     enable-preconfigured-udn-addresses=true
 {{- end }}
-{{- end }}
-{{- if .OVN_MULTI_NETWORK_POLICY_ENABLE }}
-    enable-multi-networkpolicy=true
-{{- end }}
-{{- if .OVN_ADMIN_NETWORK_POLICY_ENABLE }}
-    enable-admin-network-policy=true
```
ricky-rav, line 154: Further below in this file the ovnkube-config for the hosted cluster isn't modified. Was this intentional? Do we need to do anything about it?

tsorya: Hosted cluster is out of scope for tech preview. I believe when we will have an option to run 2 ovnk a we will see what is required from configuration point of view

ricky-rav: All right! Can you add that to the commit message? Thanks!

ricky-rav:
@tsorya could you have a look at the failures in `pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-dualstack-bgp` and `pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-dualstack-bgp-local-gw`?  I'm focusing on these two because these jobs have been reporting a problem with CNO in particular:

>  **operator conditions network** 
> {Operator degraded (RolloutHung): Deployment "/openshift-ovn-kubernetes/ovnkube-control-plane" rollout is not making progress - pod ovnkube-control-plane-5df7cbb79c-gb2jf is in CrashLoopBackOff State  Operator degraded (RolloutHung): Deployment "/openshift-ovn-kubernetes/ovnkube-control-plane" rollout is not making progress - pod ovnkube-control-plane-5df7cbb79c-gb2jf is in CrashLoopBackOff State}


A quick look at what's going on with openshift-ovn-kubernetes/ovnkube-control-plane should tell you whether we missed anything in the input parameters.

tsorya:
> @tsorya could you have a look at the failures in `pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-dualstack-bgp` and `pull-ci-openshift-cluster-network-operator-master-e2e-metal-ipi-ovn-dualstack-bgp-local-gw`? I'm focusing on these two because these jobs have been reporting a problem with CNO in particular:
> 
> > **operator conditions network**
> > {Operator degraded (RolloutHung): Deployment "/openshift-ovn-kubernetes/ovnkube-control-plane" rollout is not making progress - pod ovnkube-control-plane-5df7cbb79c-gb2jf is in CrashLoopBackOff State  Operator degraded (RolloutHung): Deployment "/openshift-ovn-kubernetes/ovnkube-control-plane" rollout is not making progress - pod ovnkube-control-plane-5df7cbb79c-gb2jf is in CrashLoopBackOff State}
> 
> A quick look at what's going on with openshift-ovn-kubernetes/ovnkube-control-plane should tell you whether we missed anything in the input parameters.

Looking

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
```diff
@@ -538,6 +538,8 @@ spec:
           value: "{{.OVN_CONTROLLER_INACTIVITY_PROBE}}"
         - name: OVN_KUBE_LOG_LEVEL
           value: "4"
+        - name: OVN_NODE_MODE
```
kyrtapz, line 541: Your commit message states that `ConfigMap values cannot be reliably overridden per-node, but script logic can be conditional`
How is that solved with setting the env vars for the whole daemonset?

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
```diff
@@ -135,8 +135,13 @@ spec:
           # will rollout control plane pods as well
           network_segmentation_enabled_flag=
           multi_network_enabled_flag=
-          if [[ "{{.OVN_NETWORK_SEGMENTATION_ENABLE}}" == "true" ]]; then
+          if [[ "{{.OVN_MULTI_NETWORK_ENABLE}}" == "true" ]]; then
             multi_network_enabled_flag="--enable-multi-network"
+          fi
+          if [[ "{{.OVN_NETWORK_SEGMENTATION_ENABLE}}" == "true" ]]; then
```
kyrtapz, line 141: This seems like a logic change. Was there a reason behind it?
@ricky-rav you are good with that right? I vaguely remember you were working on some bug around this.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
tsorya: This one provides OVN_NODE_MODE per daemonset. 
In case of DPU-HOST mode CNO will create new daemonset with DPU HOST mode value.
So each different OVN_NODE_MODE will have it's own daemonset

tsorya: In general it allows to provide the mode we running with.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
tsorya: In general took it from here 
https://github.com/openshift/cluster-network-operator/pull/2778/files#diff-633cbda8a04c3338bc6d456f1046b08e1fe1a1b09d165e74cc313335ca7b62b2R47 but i can remove it

tsorya: In general i leaved as is OVN_NETWORK_SEGMENTATION_ENABLE logic

ricky-rav: @kyrtapz the bug you probably have in mind in the end was for DPU GA (4.21), so the PR is still open and we can get to it once we're done with 4.20: https://github.com/openshift/cluster-network-operator/pull/2739

@tsorya so the difference between the original code and your code is that when OVN_MULTI_NETWORK_ENABLE=true and OVN_NETWORK_SEGMENTATION_ENABLE=false, we don't set any corresponding flags in the old code, while you set the multi network flag. 
Let's keep the original behaviour for the sake of not introducing any new behaviour that hasn't been tested :) 

Sorry for not realizing before.

tsorya: Hmm, not really. Look on configmap part i remove
`{{- if .OVN_MULTI_NETWORK_ENABLE }}
    enable-multi-network=true
{{- end }}`
So in case OVN_MULTI_NETWORK_ENABLE was set control-plane was getting this flag through CM in anycase

tsorya: From the running setup 
multi_network_enabled_flag=
if [[ "true" == "true" ]]; then
  multi_network_enabled_flag="--enable-multi-network"
fi
if [[ "true" == "true" ]]; then
  if [[ "true" != "true" ]]; then
    multi_network_enabled_flag="--enable-multi-network"
  fi
  network_segmentation_enabled_flag="--enable-network-segmentation"
fi

So in case --enable-network-segmentation we will always have enable-multi-network.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml
tsorya: https://github.com/openshift/cluster-network-operator/pull/2778/files#diff-140eef3ba41bdcf401d507408084181f2c0ac627532b61e0f7906ea7cc926782R148

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
kyrtapz: Thanks for confirming!

ricky-rav: @kyrtapz so I've just discussed this with @tsorya. The way that configuration parameters are handled in ovnk is not as intuitive as I thought. 

The key point is that non-default values that are present in the configmap take precedence over everything else (https://github.com/ovn-kubernetes/ovn-kubernetes/blame/7c047281d46a6e5f154a0f5d3feb8866af3619f4/go-controller/pkg/config/config.go#L750), also over input arguments that we pass here in the yaml. 

So that's why Igal is removing a set of parameters from the configmap and adding them as input args in the yaml. These parameters are disabled by default in upstream ovn-k and simply enabling/disabling them through input arguments while keeping them enabled in the configmap wouldn't have worked.

In the particular case above, the logic that sets the multi network and network segmentation flags is being moved from the configmap to the yaml. The fact that the logic from the original yaml is more restrictive than the one from the original configmap didn't really matter in the end, because it was the non-default values in the configmap (multi network enabled, udn network segmentation enabled) that were taking precedence over everything else.

PR #2783, opened 2025-08-25
Update linter

bpickard22:
update golang-cli linter to use non-deprecated linters
update test context refs to t.Context() to satisfy usetesting linter

we need this in order to run make check as it will error out in its current state due to fully deprecated linters


zhaozhanqi:
verified by "pre-merge testing"

PR #2787, opened 2025-08-27
CORENET-6311: Skip running _stackmanager for libreswan 5.3+

pperiyasamy:
The `_stackmanager` binary was removed in libreswan 5.3 and is no longer required for prerequisite checks. This commit conditionally runs `_stackmanager` only when the binary is present on the system.

cc @huiran0826

### DIFF
--- a/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
@@ -262,8 +262,11 @@ spec:
           ulimit -n 1024
 
           /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules
-          /usr/libexec/ipsec/_stackmanager start
+          # Check kernel modules only for libreswan version <= 5.2. The _stackmanager binary is
+          # removed from 5.3 onwards, so this check is not needed on later versions.
+          if [ -e /usr/libexec/ipsec/_stackmanager ]; then
+            /usr/libexec/ipsec/_stackmanager start
+          fi
           # Check nss database status
           /usr/sbin/ipsec --checknss
           # Start the pluto IKE daemon
--- a/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
+++ b/bindata/network/ovn-kubernetes/common/ipsec-host.yaml
@@ -273,8 +273,11 @@ spec:
           ulimit -n 1024
 
           /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules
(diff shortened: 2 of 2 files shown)


bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
```diff
@@ -262,8 +262,12 @@ spec:
           ulimit -n 1024
 
           /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules
-          /usr/libexec/ipsec/_stackmanager start
+          # Check kernel modules only for libreswan version <= 5.2. The _stackmanager binary is
+          # removed from 5.3 onwards, so this check is not needed on later versions.
+          libreswan_version=$(ipsec --version | awk '{print $2}')
+          if [ "$(printf '%s\n' "$libreswan_version" "5.2" | sort -V | head -n1)" = "$libreswan_version" ]; then
```
igsilya: I know this PR is marked as DNM, so not really intended for review. :)  But maybe it's better to simplify the check to just `if [ -e /usr/libexec/ipsec/_stackmanager ]; then` ?  Comparison of version strings seems harder to maintain.

pperiyasamy: thanks @igsilya for the review, you are right, simplified the check now.

bindata/network/ovn-kubernetes/common/ipsec-containerized.yaml
```diff
@@ -262,8 +262,11 @@ spec:
           ulimit -n 1024
 
           /usr/libexec/ipsec/addconn --config /etc/ipsec.conf --checkconfig
-          # Check kernel modules
-          /usr/libexec/ipsec/_stackmanager start
+          # Check kernel modules only for libreswan version <= 5.2. The _stackmanager binary is
+          # removed from 5.3 onwards, so this check is not needed on later versions.
+          if [  -e /usr/libexec/ipsec/_stackmanager ]; then
```
igsilya: One space should be enough. :)
Otherwise, lgtm.

pliurh:
@pperiyasamy if libreswan 5.3 is what we have in the OCP master, why do we still keep the code for versions <= 5.2?

pperiyasamy:
> @pperiyasamy if libreswan 5.3 is what we have in the OCP master, why do we still keep the code for versions <= 5.2?

@pliurh currently master is still with libreswan 5.2. even if we bump libreswan version t0 5.3 in master, this script is still getting used when upgrading it from 5.2 to 5.3.

igsilya:
Note: will also need this change down to at least 4.19 as we'll get 5.3 there as soon as it is released.

PR #2789, opened 2025-08-29
CORS-4200: Accept the GCP service endpoints

barbacbd:
pkg/network/cloud_network.go:
** Accept the GCP service endpoints. Use the compute endpoint override (if it exists) to set the API URL. This value is passed to cloud-network-config-controller to be used when the GCP compute service is created, so that all requests to the api are sent via this url rather than the default.

### DIFF
--- a/pkg/network/cloud_network.go
+++ b/pkg/network/cloud_network.go
@@ -80,6 +80,14 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 		data.Data["PlatformAzureEnvironment"] = cloudBootstrapResult.PlatformStatus.Azure.CloudName
 	}
 
+	if cloudBootstrapResult.PlatformType == v1.GCPPlatformType {
+		for _, ep := range cloudBootstrapResult.PlatformStatus.GCP.ServiceEndpoints {
+			if ep.Name == configv1.GCPServiceEndpointNameCompute {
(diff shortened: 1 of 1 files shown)


pkg/network/cloud_network.go
```diff
@@ -80,6 +80,14 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 		data.Data["PlatformAzureEnvironment"] = cloudBootstrapResult.PlatformStatus.Azure.CloudName
 	}
 
+	if cloudBootstrapResult.PlatformType == v1.GCPPlatformType {
+		for _, ep := range cloudBootstrapResult.PlatformStatus.GCP.ServiceEndpoints {
+			if ep.Name == configv1.GCPServiceEndpointNameCompute {
+				apiurl = ep.URL
```
ricky-rav, line 86: So here we're setting `apiurl` to the last `cloudBootstrapResult.PlatformStatus.GCP.ServiceEndpoints` whose name is `configv1.GCPServiceEndpointNameCompute`. Are we expecting multiple endpoints to match? Can we just break out of the loop after the first match?

barbacbd:
Closing in favor of a solution that does not require explicit api endpoint overrides for each new service.

barbacbd:
Closing in favor of a solution that does not require explicit api endpoint overrides for each new service.

PR #2798, opened 2025-09-16
set ovnkube-cluster-manager as default container

lance5890:


ricky-rav (changes requested):
Nice, I didn't know about this annotation! Could you apply the same changes also to `bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml`? 

Bonus point: 
Could we also do the same for ovnkube-node.yaml (self-hosted and managed) and set ovnkube-controller as the default container? That would be amazing.

Thanks! :)

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -33,6 +33,7 @@ spec:
       annotations:
         hypershift.openshift.io/release-image: {{.ReleaseImage}}
         target.workload.openshift.io/management: '{"effect": "PreferredDuringScheduling"}'
+        kubectl.kubernetes.io/default-container: ovnkube-control-plane
```
lance5890, line 36: add more default containers as suggested

lance5890:
> Nice, I didn't know about this annotation! Could you apply the same changes also to `bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml`?
> 
> Bonus point: Could we also do the same for ovnkube-node.yaml (self-hosted and managed) and set ovnkube-controller as the default container? That would be amazing.
> 
> Thanks! :)

that makes sense, done!

bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
```diff
@@ -40,6 +40,7 @@ spec:
         network.operator.openshift.io/ovnkube-script-lib-hash: "{{.OVNKubeConfigHash}}"
         # prevent blocks when node critical pods get evicted prior to workloads
         cluster-autoscaler.kubernetes.io/enable-ds-eviction: "false"
+        kubectl.kubernetes.io/default-container: ovnkube-controller
```
ricky-rav, line 43: Thanks! Let's do the same for ovnkube-controller also in `bindata/network/ovn-kubernetes/self-hosted/ovnkube-node.yaml` and we're done. Thanks again!

ricky-rav (changes requested):

bindata/network/ovn-kubernetes/managed/ovnkube-node.yaml
lance5890: done

ricky-rav (approved):
/LGTM

ricky-rav:
@kyrtapz PTAL

kyrtapz:
Thanks @lance5890! 
/ok-to-test
/approve

kyrtapz:
/retitle NO-JIRA: set ovnkube-cluster-manager as default container

PR #2741, opened 2025-07-15
OCPBUGS-42303: Add drop flows for GARP

martinkennelly:
/cc
/hold

PR #2801, opened 2025-09-17
OCPBUGS-42303: Add drop flows for GARP

pperiyasamy:
/cc
/hold

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -241,6 +241,23 @@ data:
         --tls-cert-file=${clientcert}
     }
 
+    # Adds drop flows for GARPs coming from patch port to br-int for fn arg bridge name.
+    # Remove when https://issues.redhat.com/browse/FDP-1537 lands in OVN core.
+    add_garp_drop_flow() {
+      local bridge="$1"
+      local cookie="0x0305"
+      local priority="499"
+      # FIXME: can we generate the exact name. Its possible we add these flows to the incorrect port when selecting on substring
+      for port_name in $(ovs-vsctl list-ports "$bridge"); do
+        if [[ "$port_name" == *to-br-int ]]; then
+          local of_port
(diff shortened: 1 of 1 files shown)


cgoncalves:
Dup of https://github.com/openshift/cluster-network-operator/pull/2741 ?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -498,6 +516,13 @@ data:
         exit 1
       fi
 
+      # start temp work around
+      # remove when https://issues.redhat.com/browse/FDP-1537 is available
+      if ovs-vsctl br-exists "br-ex" && ! retry "60" "add-GARP-drop-flows" "add_garp_drop_flow br-ex"; then
+          echo "GARP drop flows addition failed - expected during initial OVS setup"
```
pliurh: If the `*to-br-int` has not been created after a node reboot, we will have to wait for 2 mins here. I think it's better to just call `add_garp_drop_flow` without retry. Like what we do at u/s https://github.com/ovn-kubernetes/ovn-kubernetes/blob/0a11e5c9f625425dd54be53a3d96f77b81112910/dist/images/ovnkube.sh#L1817

pperiyasamy: @pliurh  do you mean wait for `ovsdb-server, ovs-vswitchd, ovn-northd and ovn-controller` processes to be ready and then invoke `add_garp_drop_flow `without a retry ? anyway retry logic also waits for 2 mins until patch port is created.

pperiyasamy: ok @pliurh , as discussed this drop flow is needed only for pod/container restart case, the node reboot is taken care with https://github.com/openshift/machine-config-operator/pull/5123. removed retry logic now.

martinkennelly:
DS merge is in - this is ready to go

martinkennelly:
/retest

cluster init issues

martinkennelly:
CI is borked :/ Trying to understand why - unrelated to these changes.

PR #2802, opened 2025-09-17
OCPBUGS-61215: Tweak iptables-alerter to try to avoid crictl bug

danwinship:
We keep getting reports of iptables-alerter causing CPU usage alerts... we debugged this at one point and it was because crictl was suddenly using a ton of CPU and RAM for no apparent reason. We weren't able to reproduce beyond that and it's not really worth spending a lot of time trying to fix since crictl is not in the critical path of any normal pods anyway. I had tried improving things by using _less_ crictl before (#2404) but we're still getting reports. This fixes it to not use crictl until after we've determined that _some_ pod, somewhere on the node is using iptables, and that should hopefully be "never".


danwinship:
/verified by @danwinship 

no e2e test, tested by hand

bindata/network/iptables-alerter/002-script.yaml
```diff
@@ -19,10 +19,36 @@ data:
...
-        date
-        # Iterate over local pods
+    function check_pods {
+        # We need to use crictl to be able to map pod information to network namespace
+        # information, but there seems to be some bug in crictl that causes excessive CPU
+        # usage on some hosts, for unknown reasons. Since we expect that most nodes won't
+        # have any iptables-using pods anyway, do a pre-scan of all (non-hostnetwork)
+        # namespaces without using crictl, and bail out early if we don't find anything
+        iptables_output=""
+        for netns_pid in $(lsns -t net -o pid -nr | sort -u | grep -v '^1$'); do
```
martinkennelly, line 33: not for others; `'^1$'` excludes pid 1 :)

martinkennelly:
/test e2e-aws-ovn-serial-2of2

Unrelated disruption

martinkennelly:
/test e2e-aws-ovn-upgrade

Unrelated - job reached timeout limit.

danwinship:
/override ci/prow/e2e-aws-ovn-serial-2of2

unrelated, very failure-prone jobs

PR #2804, opened 2025-09-23
CORS-4180: Allow AWS and Azure as platforms that support dual-stack on Day-0

sadasu:
1. Added AWS and Azure as platforms that support DualStack on Day-0
2. Added AWS, Azure and OpenStack as platforms that do not support conversion to DualStack on Day-2.
3. Updated tests to reflect the newly supported platforms and when they support DualStack.
4. Currently DualStack for AWS and Azure are behind featuregates. Updated to check that these featuregates are enabled before DualStack support is extended to these platforms.

### DIFF
--- a/pkg/network/cluster_config_test.go
+++ b/pkg/network/cluster_config_test.go
@@ -2,10 +2,10 @@ package network
 
 import (
 	"fmt"
-	"strings"
 	"testing"
 
 	configv1 "github.com/openshift/api/config/v1"
+	apifeatures "github.com/openshift/api/features"
 	v1 "github.com/openshift/api/network/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -17,6 +17,7 @@ import (
 	"github.com/openshift/cluster-network-operator/pkg/client/fake"
 	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 
 	. "github.com/onsi/gomega"
 )
@@ -35,6 +36,20 @@ var ClusterConfig = configv1.NetworkSpec{
 	ServiceNetwork: []string{"192.168.0.0/20"},
 
 	NetworkType: "None",
+}
+
+func getFeatureGatesWithDualStack() featuregates.FeatureGate {
+	return featuregates.NewFeatureGate(
+		[]configv1.FeatureGateName{apifeatures.FeatureGateAdminNetworkPolicy,
+			apifeatures.FeatureGateDNSNameResolver,
+			apifeatures.FeatureGateNetworkSegmentation,
+			apifeatures.FeatureGateOVNObservability,
+			apifeatures.FeatureGateAWSDualStackInstall,
+			apifeatures.FeatureGateAzureDualStackInstall},
+		[]configv1.FeatureGateName{
+			apifeatures.FeatureGatePreconfiguredUDNAddresses,
+		},
+	)
 }
 
 func TestValidateClusterConfig(t *testing.T) {
@@ -57,12 +72,13 @@ func TestValidateClusterConfig(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 
 	cc := *ClusterConfig.DeepCopy()
-	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+	featureGates := getFeatureGatesWithDualStack()
+	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	g.Expect(err).NotTo(HaveOccurred())
 
 	haveError := func(cfg configv1.NetworkSpec, substr string) {
 		t.Helper()
-		err = ValidateClusterConfig(&configv1.Network{Spec: cfg}, client)
+		err = ValidateClusterConfig(&configv1.Network{Spec: cfg}, client, featureGates)
 		g.Expect(err).To(MatchError(ContainSubstring(substr)))
 	}
 
@@ -88,7 +104,7 @@ func TestValidateClusterConfig(t *testing.T) {
 
 	cc = *ClusterConfig.DeepCopy()
 	cc.ClusterNetwork[1].HostPrefix = 0
-	res := ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+	res := ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	// Since the NetworkType is None, and the hostprefix is unset we don't validate it
 	g.Expect(res).Should(BeNil())
 
@@ -289,7 +305,8 @@ func TestValidClusterConfigLiveMigration(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			client := fake.NewFakeClient(tt.objects...)
-			err := validateClusterConfig(tt.config, tt.infraRes, client)
+			fg := getFeatureGatesWithDualStack()
+			err := validateClusterConfig(tt.config, tt.infraRes, client, fg)
 			if tt.expectErr {
 				g.Expect(err).To(HaveOccurred())
 				if tt.expectedErrorMsg != "" {
@@ -301,6 +318,21 @@ func TestValidClusterConfigLiveMigration(t *testing.T) {
 			}
 		})
 	}
+}
+
+func getFeatureGatesWithIncompleteDualStack() featuregates.FeatureGate {
+	return featuregates.NewFeatureGate(
+		[]configv1.FeatureGateName{apifeatures.FeatureGateAdminNetworkPolicy,
+			apifeatures.FeatureGateDNSNameResolver,
+			apifeatures.FeatureGateNetworkSegmentation,
+			apifeatures.FeatureGateOVNObservability,
+			apifeatures.FeatureGateAWSDualStackInstall,
+		},
+		[]configv1.FeatureGateName{
+			apifeatures.FeatureGatePreconfiguredUDNAddresses,
+			apifeatures.FeatureGateAzureDualStackInstall,
+		},
+	)
 }
 
 func TestValidateClusterConfigDualStack(t *testing.T) {
@@ -323,12 +355,13 @@ func TestValidateClusterConfigDualStack(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 
 	cc := *ClusterConfig.DeepCopy()
-	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+	featureGates := getFeatureGatesWithDualStack()
+	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	g.Expect(err).NotTo(HaveOccurred())
 
 	haveError := func(cfg configv1.NetworkSpec, substr string) {
 		t.Helper()
-		err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+		err = ValidateClusterConfig(&configv1.Network{Spec: cfg}, client, featureGates)
 		g.Expect(err).To(MatchError(ContainSubstring(substr)))
 	}
 
@@ -368,11 +401,14 @@ func TestValidateClusterConfigDualStack(t *testing.T) {
 		CIDR:       "fd01::/48",
 		HostPrefix: 64,
 	})
-	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	g.Expect(err).NotTo(HaveOccurred())
 
-	// You can't use dual-stack if this is anything else but BareMetal or NonePlatformType
-	infrastructure.Status.PlatformStatus.Type = configv1.AzurePlatformType
+	// You can't use dual-stack if enabled on an unsupported platform
+	infrastructure.Status.PlatformStatus.Type = configv1.GCPPlatformType
+	infrastructure.Status.PlatformStatus.GCP = &configv1.GCPPlatformStatus{
+		Region: "us-west1",
+	}
 	client = fake.NewFakeClient(infrastructure)
 	err = createProxy(client)
 	g.Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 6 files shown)


patrickdillon:
This looks good to me. AWS tests are failing to find a lease... i think we're hitting ci infra problems.

pkg/network/render.go
```diff
@@ -166,6 +172,8 @@ func Render(operConf *operv1.NetworkSpec, clusterConf *configv1.NetworkSpec, man
 		return nil, progressing, err
 	}
 
+	updateDualStackPlatforms(featureGates)
```
tthvo: Should we move the function to earlier in the func `Render` so that any rendering, if any later on, can always pick up the most up-to-date list of supported dualstack platform?

pkg/network/render.go
```diff
@@ -999,3 +1011,14 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
...
+
+func updateDualStackPlatforms(featureGates featuregates.FeatureGate) {
+	for _, fg := range featureGates.KnownFeatures() {
+		if fg == apifeatures.FeatureGateAWSDualStackInstall && featureGates.Enabled(apifeatures.FeatureGateAWSDualStackInstall) {
+			dualStackPlatforms.Insert(string(configv1.AWSPlatformType))
+		}
+		if fg == apifeatures.FeatureGateAzureDualStackInstall && featureGates.Enabled(apifeatures.FeatureGateAzureDualStackInstall) {
+			dualStackPlatforms.Insert(string(configv1.AzurePlatformType))
+		}
+	}
```
tthvo: question: Do we need to loop through all feature gates at all :eyes:? Can we:

```golang
func updateDualStackPlatforms(featureGates featuregates.FeatureGate) {
    if featureGates.Enabled(apifeatures.FeatureGateAWSDualStackInstall) {
        dualStackPlatforms.Insert(string(configv1.AWSPlatformType))
    }
    if featureGates.Enabled(apifeatures.FeatureGateAzureDualStackInstall) {
        dualStackPlatforms.Insert(string(configv1.AzurePlatformType))
    }
}
```

tthvo: I was thinking the number of feature gates might be large and we can save some compute time :D

pkg/network/render.go
```diff
@@ -38,6 +38,12 @@ var dualStackPlatforms = sets.NewString(
 	string(configv1.OpenStackPlatformType),
 )
 
+var conversionToDualStackPlatforms = sets.NewString(
+	string(configv1.BareMetalPlatformType),
+	string(configv1.NonePlatformType),
+	string(configv1.VSpherePlatformType),
+)
```
tthvo: question: Previously, only `OpenStack` is rejected. Just wondering if we confirmed this is OK?

tthvo: The code comments seem to indicate so, but I have 0 clue :disappointed:

sadasu: This is the "allow" list so it contains the list of platforms that allow conversion to DualStack on Day-2. The code earlier was looking at OpenStack platform as the reject list. It could be implemented either ways.

tthvo: Oh sounds good! I was just unsure if there are any other platforms that allow dualstack, which is not yet accounted for here...

pkg/network/render.go
sadasu: Done. Just looked at how featuregates.Enabled() is implemented and we don't need to iterate though the existing feature gates.

tthvo:
Maybe another rebase?

pkg/network/render.go
```diff
@@ -47,6 +53,9 @@ func Render(operConf *operv1.NetworkSpec, clusterConf *configv1.NetworkSpec, man
 	var progressing bool
 	objs := []*uns.Unstructured{}
 
+	// Update the list of supported DualStack platforms based on enabled feature gates.
+	updateDualStackPlatforms(featureGates)
+
```
tthvo: @sadasu this is added in the `Render` function is, I believe, is called when rendering manifests for bootstrap.

This means when the cluster-network-operator is running in the cluster, such function is not called, right? And the list is not modified at all?

tthvo: /lgtm cancel

Pending question above :eyes:

tthvo: I think we should have the function `isSupportedDualStackPlatform` keeps the list of supported platforms (i.e. both static and feature-gated items), instead of having a global variable and modifying them at runtime.

WDYT?

pkg/network/render.go
```diff
@@ -999,3 +1012,16 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
 
 	return out, nil
 }
+
+// updateDualStackPlatforms checks for enabled feature gates and adds to the list
+// of platforms that support DualStack on Day-0. Currently the 2 platforms added here
+// donot support conversion to DualStack on Day-2. When that happens, we will need to
+// update `conversionToDualStackPlatforms` too.
+func updateDualStackPlatforms(featureGates featuregates.FeatureGate) {
```
kyrtapz: I don't like the fact that we are modifying the global variable through this function. My proposal is to set the supported platforms as part of the `bootstrap.InfraStatus` and read it from there. Same applies to `conversionToDualStackPlatforms`.

pkg/network/render.go
```diff
@@ -47,6 +53,9 @@ func Render(operConf *operv1.NetworkSpec, clusterConf *configv1.NetworkSpec, man
 	var progressing bool
 	objs := []*uns.Unstructured{}
 
+	// Update the list of supported DualStack platforms based on enabled feature gates.
+	updateDualStackPlatforms(featureGates)
```
kyrtapz: ~~Calling it here would mean adding elements to `dualStackPlatforms` every time `Render` is called, we cannot do that.~~

Edit: this isn't a problem since we use a set. I still don't like the fact that it tries to modify the global variable so often.

pkg/network/render.go
```diff
@@ -400,13 +409,13 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
...
-	// Validate that this is either a BareMetal or None PlatformType. For all other
-	// PlatformTypes, migration to DualStack is prohibited
+	// Validate that this is a platform that supports DualStack. If it does, then check if
+	// migration to DualStack on day-2 is allowed.
 	if len(prev.ServiceNetwork) < len(next.ServiceNetwork) {
 		if !isSupportedDualStackPlatform(infraRes.PlatformType) {
 			return errors.Errorf("%s is not one of the supported platforms for dual stack (%s)", infraRes.PlatformType,
 				strings.Join(dualStackPlatforms.List(), ", "))
-		} else if string(configv1.OpenStackPlatformType) == string(infraRes.PlatformType) {
+		} else if !isConversionSupportedDualStackPlatform(infraRes.PlatformType) {
```
kyrtapz: I know you didn't introduce this but we can simplify the code and ditch the `else if` since the if returns.
something like this:
```
if !isSupportedDualStackPlatform(infraRes.PlatformType) {
    return ...
}
if !isConversionSupportedDualStackPlatform(infraRes.PlatformType) {
    return ...
}
```

pkg/network/render.go
tthvo: Whoops, nvmind... It is also called during reconcilation :sweat: so its fine. Please ignore it :D

pkg/network/render.go
sadasu: @kyrtapz happy to make that change but I am not sure how future proof that is. As long as the `dualStackPlatforms` is in place, we run the risk of someone using that global instead of the InfraStatus.

kyrtapz: With the change I proposed we should get rid of the global `dualStackPlatforms` variable.

sadasu: @kyrtapz and @tthvo I have added a WIP commit that gets rid of the global `dualStackPlatforms` but doesn't use `InfraStatus`. With `InfraStatus`, I found the blast radius to be too large for what we want to achieve here. PTAL and provide feedback.

pkg/network/render.go
```diff
@@ -403,10 +396,7 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	// Validate that this is either a BareMetal or None PlatformType. For all other
 	// PlatformTypes, migration to DualStack is prohibited
 	if len(prev.ServiceNetwork) < len(next.ServiceNetwork) {
-		if !isSupportedDualStackPlatform(infraRes.PlatformType) {
-			return errors.Errorf("%s is not one of the supported platforms for dual stack (%s)", infraRes.PlatformType,
-				strings.Join(dualStackPlatforms.List(), ", "))
-		} else if string(configv1.OpenStackPlatformType) == string(infraRes.PlatformType) {
+		if !isConversionToDualStackSupported(infraRes.PlatformType) {
```
sadasu, line 398: @kyrtapz please note here that I am only checking if the conversion to DualStack is possible. I removed the check `!isSupportedDualStackPlatform()` because the method `isNetworkChangeSafe()` is checking for Day-2 changes.

pkg/network/render.go
```diff
@@ -999,3 +985,25 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
 
 	return out, nil
 }
+
+func isSupportedDualStackPlatform(platformType configv1.PlatformType, featureGates featuregates.FeatureGate) bool {
+	switch platformType {
+	case configv1.BareMetalPlatformType, configv1.NonePlatformType, configv1.VSpherePlatformType, configv1.OpenStackPlatformType:
```
sadasu: Do you want me to add Platform: External to this list? It is treated identical to Platform:None in other places.

pkg/network/render.go
```diff
@@ -999,3 +985,25 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
...
+	case configv1.AzurePlatformType:
+		return featureGates.Enabled(apifeatures.FeatureGateAzureDualStackInstall)
+	default:
+		return false
+	}
+}
+
+func isConversionToDualStackSupported(platformType configv1.PlatformType) bool {
+	switch platformType {
+	case configv1.BareMetalPlatformType, configv1.NonePlatformType, configv1.VSpherePlatformType:
```
sadasu, line 1009: Ditto:Do you want me to add Platform: External to this list? It is treated identical to Platform:None in the Installer.

pkg/network/cluster_config_test.go
```diff
@@ -323,12 +340,13 @@ func TestValidateClusterConfigDualStack(t *testing.T) {
...
+	featureGates := getFeatureGatesWithDualStack()
+	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	g.Expect(err).NotTo(HaveOccurred())
 
 	haveError := func(cfg configv1.NetworkSpec, substr string) {
 		t.Helper()
-		err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client)
+		err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 		g.Expect(err).To(MatchError(ContainSubstring(substr)))
 	}
```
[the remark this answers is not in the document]
tthvo: I think there won't be any build error here, especially this is just a test file :grin: 

But it is a valid suggestion to avoid referencing a captured variable. I guess that's why there was a parameter `cfg configv1.NetworkSpec` in the first place.

pkg/network/render.go
```diff
@@ -403,10 +396,7 @@ func isNetworkChangeSafe(prev, next *operv1.NetworkSpec, infraRes *bootstrap.Inf
 	// Validate that this is either a BareMetal or None PlatformType. For all other
 	// PlatformTypes, migration to DualStack is prohibited
```
tthvo: We should probably update this comment too (or remove it) since Vsphere also support dualstack conversion.

pkg/network/cluster_config_test.go
```diff
@@ -368,11 +386,14 @@ func TestValidateClusterConfigDualStack(t *testing.T) {
...
+	err = ValidateClusterConfig(&configv1.Network{Spec: cc}, client, featureGates)
 	g.Expect(err).NotTo(HaveOccurred())
 
-	// You can't use dual-stack if this is anything else but BareMetal or NonePlatformType
-	infrastructure.Status.PlatformStatus.Type = configv1.AzurePlatformType
+	// You can't use dual-stack if enabled on an unsupported platform
+	infrastructure.Status.PlatformStatus.Type = configv1.GCPPlatformType
+	infrastructure.Status.PlatformStatus.GCP = &configv1.GCPPlatformStatus{
+		Region: "us-west1",
+	}
```
tthvo, line 411: Should we also have test cases that validate if there is no error when a platform:
- Support dualstack out of the box.
- Support dualstack with Feature gate.

pkg/network/render.go
tthvo: I guess this is fine under the assumption: if a platform supports converting to dual-stack, it really should support dual-stack, right :grin:?

The difference here is just error message :D

pkg/network/render.go
tthvo: Thanks @sadasu!

The new approach makes sense to me and it worked when testing with https://github.com/openshift/installer/pull/9930. Though obviously I am not the expert here, maybe @kyrtapz can have the final thought :pray:

pkg/network/cluster_config_test.go
sadasu: > Should we also have test cases that validate if there is no error when a platform:
> 
> * Support dualstack out of the box.

This already exists

> * Support dualstack with Feature gate.

Added a test for that.

pkg/network/render.go
kyrtapz: We can consider it as a separate effort to keep the changes contained here.
Feel free to raise a separate PR!

sadasu:
@orenc1 Hi! I noticed that you added kubevirt as a platform that supports DualStack. This PR changes how we determine which platforms support DualStack and also distinguishes between platforms that support only on Day-0 vs also in Day-2. Could you please review to make sure it is OK not to include kubevirt as a platform that supports DualStack on Day-2?

tthvo (commented):
/lgtm

Doing another install to verify ... 👀

tthvo:
/verified by @tthvo 

Same results from testing. Rebase looks good to me :D

sadasu:
@kyrtapz Hi! The jobs [e2e-aws-ovn-hypershift-conformance](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2804/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-hypershift-conformance/2009269954074906624) and [hypershift-e2e-aks](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2804/pull-ci-openshift-cluster-network-operator-master-hypershift-e2e-aks/2009269954062323712) are showing a higher incidence of failures lately. They had passed earlier before my rebase. Is it OK overriding them?

tthvo:
/test hypershift-e2e-aks

Let's give it another try. Latest runs from other PRs look OK...

tthvo:
/hold cancel

Test all passed. Retesting is a bit flaky :(

PR #2846, opened 2025-11-27
  NO-JIRA:  cleanup unused code and refactor for allowlist controller

karampok:
- Removes unused code (scheme field, configmap template)
  - Refactors/simplifies (inlines single-use helpers, better naming)
  - No functional changes, just QOL improvements


zeeke (commented):
/lgtm

Great cleaning IMO

jrvaldes:
/test e2e-aws-ovn-windows


PR merged
- https://github.com/openshift/windows-machine-config-operator/pull/3603

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -36,21 +38,15 @@ import (
 )
 
 const (
-	allowlistDsName      = "cni-sysctl-allowlist-ds"
-	allowlistAnnotation  = "app=cni-sysctl-allowlist-ds"
-	manifestDir          = "../../bindata/allowlist/daemonset"
-	allowlistManifestDir = "../../bindata/network/multus/004-sysctl-configmap.yaml"
+	dsName       = "cni-sysctl-allowlist-ds"
+	dsAnnotation = "app=cni-sysctl-allowlist-ds"
+	dsManifest   = "../../bindata/allowlist/daemonset/daemonset.yaml"
```
jcaamano: Can you explain the motivation of the `dsManifest` value change? I am missing it from the commit message. This is being passed to a `renderDir` function so it wouldn't naturally make sense.

jcaamano (commented):
@karampok 

Can you address the @coderrabitai comments?
Is there any testing coverage for this controller?

karampok:
> @karampok
> 
> Can you address the @coderrabitai comments? 

I will fix the typo for the first comment. The second comment is actually code/comment which I do not want to touch since many comments are in that way.

>Is there any testing coverage for this controller?

Unfortunately there is no unit test, there are some high level test in origin 

```
"[sig-network][Feature:tuning] pod should not start for sysctls not on whitelist [apigroup:k8s.cni.cncf.io] net.ipv4.conf.IFNAME.arp_filter [Suite:openshift/conformance/parallel]"

```
and here https://github.com/openshift-kni/cnf-features-deploy/blob/master/cnf-tests/testsuites/e2esuite/security/tuning.go

pkg/controller/allowlist/allowlist_controller.go
karampok: Since that dir has only one manifest
```
tree bindata/allowlist/daemonset/
bindata/allowlist/daemonset/
└── daemonset.yaml
```
having a `Dir` is more confusing, and there is no path (AFAIK) that will be any other manifest there.

This PR is minimal refactoring, in a next PR I will also refactor the `renderDir` as this is what does not make sense

jcaamano: I am not sure I agree. The daemonset is represented by the contents of `bindata/allowlist/daemonset/`; the fact that this is represented by a single or multiple manifests is a detail that is best left out from the controller code. Otherwise, if in the future we want to add more manifests (let's say a config map) then we would need to change the code. Rendering directories is the common pattern we follow in CNO to abstract the code from this fact.

karampok: no strong opinion, i changed to `dsManifestDir = "../../bindata/allowlist/daemonset"`

karampok:
@jcaamano is that something we can merge now? Thanks

jcaamano:
@karampok can you do me a favor and address the first comment of the bot? I thin just replacing `!exists` with `err == nil && !exists` would do. I know it's not in your diff but it's so egregious...

karampok:
@jcaamano I pushed another commit which I change that code part.

karampok:
AFAICT the ci failures are not due to my PR? can we merge? Need verified label

jcaamano:
> AFAICT the ci failures are not due to my PR? can we merge? Need verified label

We need this fix I think
https://github.com/openshift/cluster-network-operator/pull/2889

karampok:
@jcaamano still needs  labels `Not mergeable. Needs lgtm, verified labels.`

karampok:
hi, @jcaamano small reminder here, thanks

jcaamano:
@karampok in theory you would add the verify label yourself if it is not a bug fix, or QE if it is a bug fix.

karampok:
@jcaamano  I rebased, maybe that will help with the ci

PR #2851, opened 2025-12-04
NO-JIRA: Added global overrides for OVN-Kubernetes

palonsoro:
So far, one can override log levels in specific nodes via `env-overrides` configmap. However, only specific nodes could be overridden, there was not a global override. This modifies `ovnkube-lib.sh` to allow specifying a `_global` node whose overrides are sourced in all nodes when present. If `_global` and node-specific keys are present in the configmap, values on the node-specific one should take precedence.

This would be useful for scenarios where there is machine autoscaling, because you cannot know the node names in advance and manually maintain the `env-overrides` configmap if the node machines change automatically.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -11,6 +11,13 @@ data:
   ovnkube-lib.sh: |-
     #!/bin/bash
     set -x
+    # Add global overrides if the container has mounted any
+    if [[ -f "/env/_global" ]]; then
+      set -o allexport
+      source "/env/_global"
+      set +o allexport
+    fi
+
     # Add node-specific overrides if the container has mounted any
     K8S_NODE=${K8S_NODE:-}
     if [[ -n "${K8S_NODE}" && -f "/env/${K8S_NODE}" ]]; then


palonsoro:
Tested on a 4.21 nightly. It works as expected:
- If a `_global` key is added to `env-overrides` configmap, it is applied on all nodes.
- If there is a `_global` key and a node name key in `env-overrides` configmap, the node name one takes precedence.

So the PR should be ready for review.

palonsoro:
This PR would also address what is requested in [RFE-8572](https://issues.redhat.com/browse/RFE-8572)

palonsoro:
The failed test complains about stuff not related to the PR.

So @bpickard22 @pliurh not sure if you could please help me reviewing and getting this PR. Not in a rush, but it would help us troubleshooting if we could set this globally, specially in autoscaling scenarios (as exemplified in the RFE).

Thanks in advance.

pliurh:
The script `ovnkube-lib.sh` is only used by ovnkube-node pods. Shall we also set this for the ovnkube-control-plane pods?
You shall also update the `README.md` to include an example of the `_global` key.

palonsoro:
Hi

> The script ovnkube-lib.sh is only used by ovnkube-node pods. Shall we also set this for the ovnkube-control-plane pods?

I originally thought of nodes only, but it makes sense to include the control plane too. Let me update the control plane too.

> You shall also update the README.md to include an example of the _global key.

Got it. I'll be doing it shortly.

palonsoro:
By the way, while reviewing this, I find it confusing that we still use `_master` for nbdb, sbdb, northd... This comes from the pre-IC era. However, it is really confusing that we set configurations in the nodes with the `_master` key.

So I see 2 alternatives:
- Add `_global` also on the containers of `ovnkube-node` namespace that use `_master` for the log levels.
- Make the containers that use `_master` on the `ovnkube-node` use the node-specific configmap key and `_global`. That way, all the containers in `ovnkube-node` daemonset would use the node-specific setting or `global` and only the `ovnkube-control-plane` deployment would use `_master` and `_global`.

Which one would you prefer? I'd definitely advocate for the latter. It is a bit of a wider change (maybe I should re-title the PR, then), but IMHO it makes much more sense to have the node-specific setting impact everything in the node, `_master` to impact only what now is the real control plane (`ovnkube-control-plane` deployment ) and `_global` impact everything.

What do you think?

palonsoro:
@pliurh not sure if you could please comment on the above. Thanks.

pliurh:
Option 2 makes more sense to me.
However, it could be a breaking change to existing clusters that have already turned on debug. @kyrtapz WDYT?

palonsoro:
I'd say that extra logging is something that should be turned on only for specific situations, not to be kept left on. And as this is a new feature, if we don't backport it, we can just document the behavior change (in the customer portal solutions where we document OVN-Kubernetes log level changes).

bpickard22:
@palonsoro if you can get around to fixing the merge conflicts ill try to give this a review so we can get it merged

palonsoro:
Thx @bpickard22 . I rebased the PR.

While reviewing, it would be good to address the points raised during the discussion with @pliurh , mainly these:
- https://github.com/openshift/cluster-network-operator/pull/2851#issuecomment-3640728797 
- https://github.com/openshift/cluster-network-operator/pull/2851#issuecomment-3640933922

It would be good to make a decision on what to do with these before making changes/improvements to the PR.

asadawar:
Nice addition. One thing I noticed while investigating OCPBUGS-61688: the `ovnkube-control-plane` deployment mounts `/env` from the `env-overrides` configmap, but its containers do not source `ovnkube-lib.sh`. So neither the existing per-node overrides nor the new `_global` key will affect `ovnkube-control-plane` pods. The `OVN_KUBE_LOG_LEVEL` there is set to `"4"` in the pod spec and nothing reads `/env/${K8S_NODE}`.

Not a blocker for this PR since it was already the case before, but worth noting as a follow-up if the intent is for `_global` to truly apply everywhere.

palonsoro:
Thx for the comment @asadawar . Maybe a follow-up PR would make sense once this one can make it.

kyrtapz:
/lgtm
/retitle NO-JIRA: Added global overrides for OVN-Kubernetes

palonsoro:
Thx @kyrtapz

PR #2858, opened 2025-12-11
OCPBUGS-60532: Allowlist controller: add node reconciler to sync allowlist on node creation

karampok:
When a node joins the cluster, create a Job on that specific node to sync
    the CNI sysctl allowlist configuration.

    Implementation:
    - Watch node creation events
    - Skip any action if ConfigMap unchanged from the default (multus daemon
      installs default)
    - Create Job with nodeSelector targeting specific node
    - If job AlreadyExists, follow up the result
    - Delete successful Jobs, preserve failed Jobs for debugging

    Why not the daemonset approach:
     1. Jobs run only on the new node (1 pod) while DaemonSet runs on all
        nodes (100+ pods), so 10 new nodes create 10 Jobs vs 1000 pods.
     2. Each Job succeeds or fails independently with logs kept for 24
        hours, while DaemonSet waits for all pods to be ready so one stuck
        pod blocks the entire update.
     3. Multiple Jobs run in parallel without conflicts, while DaemonSet
        needs complex logic to handle multiple node events with delays and
        state checking.
     4. ConfigMap changes use DaemonSet (update all nodes), node additions
        use Jobs (update one node), each approach fits its purpose.

    Known issues:
    1. On restart both those jobs and the daemonset will at the same time.
    2. On restart (or if new node is master node) job will fail after 10
       minutes because we do not tolerate by design


karampok:
This PR stacks on top of https://github.com/openshift/cluster-network-operator/pull/2846 (just for the name refactoring)

pkg/controller/allowlist/nodeReconciler.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+func (r *ReconcileNode) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+	defer utilruntime.HandleCrash(r.status.SetDegradedOnPanicAndCrash)
+
+	defaultCM := &corev1.ConfigMap{}
+	if err := r.client.Default().CRClient().Get(ctx,
+		types.NamespacedName{Name: names.DefaultAllowlistConfigName, Namespace: names.MultusNamespace},
+		defaultCM); err != nil {
+		klog.Infof("%s no default ConfigMap %v found", reconcilerID, err)
```
zeeke, line 79: Can we improve this logging? Looking at the CI, it appears many times:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2858/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-serial-1of2/2000492588125130752/artifacts/e2e-aws-ovn-serial/gather-extra/artifacts/pods/openshift-network-operator_network-operator-68b5c9bb7b-2x96h_network-operator_previous.log

Maybe we can reconcile on the node creation event only?

karampok: Thanks @zeeke 

The current implementation  filters for node creation events only via the `nodePredicate()` (lines 145-159). 
However, the logging issue you're observing is a timing/ordering problem on controller startup:

  1. Synthetic CREATE events: When the controller starts/restarts, controller-runtime sends synthetic CREATE events for all existing nodes in the cache (standard behavior).  What we discussed and which documented in the commit as known issue.

  2. ConfigMap not ready: The `default-cni-sysctl-allowlist` ConfigMap doesn't exist yet because multus daemon hasn't created it. This causes every reconciliation to fail and log errors. So is a race condition that multus takes time to create the configmap, and that is just transient errors.

Let me think about it, but probably just doing nothing until there is default-config map might make sense (like if api does not exists, return not expo backoff retry)

zeeke: >The current implementation filters for node creation events only via the nodePredicate() (lines 145-159).
I missed that. that's fine.

>This causes every reconciliation to fail and log errors. So is a race condition that multus takes time to create the configmap, and that is just transient errors.

yeah, it's about having a cleaner startup logs

PR #2894, opened 2026-01-28
CORS-4332: Add GCP to the allowed platforms for dual stack

barbacbd:
Added GCP as a platform that supports DualStack on Day-0
Added GCP as a platform that does not support conversion to DualStack on Day-2
Added tests for GCP

** GCP is behind a featuregate. Ensure that these featuregates are enablede before dual stack support is extended to GCP.


sadasu:
/lgtm

@barbacbd after you test, could you please add the verified label?

tthvo (commented):
/lgtm

too :D

PR #2910, opened 2026-02-20
OCPBUGS-77015: wait for patch port to apply drop garp flows

bpickard22:
There is currently a race in the ovnkube-lib script where we try to add some drop flows to prevent duplicate garps to the patch port, but if the patch port isnt ready, the ovn-controller can connect to the db and create the patch port without the flow.

We need to wait for the patch port to exist before we try to add the GARP drop flow

also added some logs for better visibility

Assisted-By: Claude Sonnet 4.5

cc @pperiyasamy 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added automatic retry-and-wait for bridge integration setup with extended retry limit and 2s delays.
  * Starts the retry process in the background so initialization remains non-blocking.
  * Improved INFO logs when waiting and when a drop-flow is added; emits a WARN if retries exhaust without success and clearly reports failure when the required port never appears.

### DIFF
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -247,9 +247,38 @@ data:
           local of_port
           of_port=$(ovs-vsctl get interface "$port_name" ofport)
           ovs-ofctl add-flow "$bridge" "cookie=$cookie,table=0,priority=$priority,in_port=$of_port,arp,arp_op=1,actions=drop" > /dev/null
-          break
+          echo "$(date -Iseconds) - INFO - Added GARP drop flow on $bridge port $port_name (ofport=$of_port)"
+          return 0
         fi
       done
+
+      # Port not found
+      return 1
+    }
+
+    # Wait for patch port to exist and add GARP drop flow.
+    wait-and-add-garp-drop-flow() {
+      local bridge="$1"
+      local max_retries=90
+      local retries=0
+
+      while [[ "${retries}" -lt "${max_retries}" ]]; do
+        if add_garp_drop_flow "${bridge}"; then
+          return 0
+        fi
+
+        (( retries += 1 ))
+
+        # Log once after 30 seconds to indicate we're still waiting
+        if [[ "${retries}" -eq 15 ]]; then
+          echo "$(date -Iseconds) - INFO - Waiting for patch port on ${bridge} to be ready..."
+        fi
+
+        sleep 2
+      done
+
+      echo "$(date -Iseconds) - WARN - Failed to add GARP drop flow on ${bridge} after ${max_retries} attempts (best-effort workaround)"
+      return 1
     }
 
     # quit-nbdb() will cleanly shut down the northbound dbserver. It is intended
@@ -512,7 +541,7 @@ data:
       # start temp work around
       # remove when https://issues.redhat.com/browse/FDP-1537 is available
       if ovs-vsctl br-exists "br-ex"; then
-          add_garp_drop_flow br-ex
+          (wait-and-add-garp-drop-flow br-ex &)
       fi
       # end temp work around
 


bpickard22:
ci is failing because the wait-and-add-drop-flows function I added is blocking, it currently has a retry window of 60 seconds, but its blocking, and in a ci cluster it takes longer than 60 seconds for the patch port to come up, so the container errors out, causing the failures

I think there are two ways that we could go about this fix
1. run the garp flow check in the background so it waits for the patch port to come up after ovnkube node starts up 
this will work but i dont love this. we can make it fail gracefully, but then we could hit the same issue where we are trying to add the drop flows to a patch port that doesnt exist, and when ovn-controller creates the patch port from the db, it will not have the flows
also i think this will be difficult to debug 

2. fix this in ovn-k 
we maybe could remove the drop flows logic here and add the drop flows there where it is easier to track the state of the patch port. I will work on a pr for this in the meantime anyways, but would like your thoughts @pperiyasamy @arghosh93 

I think that we can avoid the race between the ovn-controller and the ovnkube script more easily with this solution, but it will be a bigger change

bpickard22:
I think for now I am going to go with a mix. I will make the check run in the background and fail gracefully. this should provide some relief. In the meantime I will work on a fix in ovn-k, then when that merges we can remove the bash workaround here for the drop flows

bpickard22:
none of the failing jobs have passed for some time, so i think it should be safe to override right @jluhrsen

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -247,9 +247,35 @@ data:
...
       done
+
+      # Port not found
+      return 1
+    }
+
+    # Wait for patch port to exist and add GARP drop flow.
+    wait-and-add-garp-drop-flow() {
+      local bridge="$1"
+      local max_retries=90
```
arghosh93: Did you check comment https://github.com/openshift/cluster-network-operator/pull/2801#discussion_r2358233987 on the PR which added this GARP drop flow? 

This comment basically says that this workaround is for hard reboots or container restarts and should not require a retry, as during a graceful node reboot we add GARP drop flows while stopping the nodeControllerManager.

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/bridgeconfig/bridgeflows.go#L65-L76

bpickard22: Thats the whole problem I believe. We have to ensure (or in this case best-effort) attempt to get the drop flow into the patch port.
There is a race between this script (where we write the flow) and the ovn-controller creating the patch port, and if the port is not up when we try to write the flow, we will just continue, and then the ovn-controller will create the patch port without the flow.

bpickard22:
Also to address the code-rabbit comment, initially I was trying to gate the script to ensure that the drop flows returns successfully, but in CI clusters it takes to long for the patch port to come up, so it would always timeout causing the container to fail and ci to error out. This workaround should provide relief via more success in adding the drop flows to the port, but there are scenarios in which it will not. The fix here -> https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6011 will ensure that it always works, and we can remove the workaround here when that merges, but in the meantime I believe this is our best option

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -247,9 +247,35 @@ data:
...
+
+      while [[ "${retries}" -lt "${max_retries}" ]]; do
+        if add_garp_drop_flow "${bridge}"; then
+          return 0
+        fi
+
+        (( retries += 1 ))
+        if [[ "${retries}" -lt "${max_retries}" ]]; then
+          echo "$(date -Iseconds) - WARN - Patch port not found on ${bridge}, retry ${retries}/${max_retries}"
+          sleep 2
```
arghosh93: Sleeping for 2 seconds seems aggressive to me. Can we increase the number of retries instead?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -247,9 +247,35 @@ data:
...
+        fi
+      done
+
+      echo "$(date -Iseconds) - ERROR - Failed to add GARP drop flow on ${bridge} after ${max_retries} attempts"
```
arghosh93: Is there any way to bypass this part of the script if the node dont have an egress-assignable label? Not sure if this is possible but worth evaluating.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: Please be mindful about info level logs if we are setting max_retries to a very high number.

bpickard22: yeah good catch, I will reduce the level to a warning and reduce the amount of logs we send

bpickard22: i still think the 2 second sleep is the right number to give us the best shot at the patch port being up, but i reduced the number of retries from 90 - 60 so at the worst case we only sit here for 2 mins

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
bpickard22: that would be a racey check so i dont want to add it. Ther is a chance that the node gets labeled after we check but then its too late to write to the patch port.
Id rather keep this running on all nodes at startup as as preventative measure until we can implement the real fix in ovn core

jluhrsen:
/test e2e-gcp-ovn

this working yet?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: Okay

arghosh93 (approved):

bpickard22:
just need an approver to take a look at this point. The two required jobs will need an override

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -248,9 +247,42 @@ data:
...
+
+      while [[ "${retries}" -lt "${max_retries}" ]]; do
+        if add_garp_drop_flow "${bridge}"; then
+          return 0
+        fi
+
+        (( retries += 1 ))
+
+        # Log once after 30 seconds to indicate we're still waiting
+        if [[ "${retries}" -eq 15 && "${logged_waiting}" == "false" ]]; then
```
danwinship: `logged_waiting` doesn't end up doing anything

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -248,9 +247,42 @@ data:
...
+      done
+
+      # Port not found
+      return 1
+    }
+
+    # Wait for patch port to exist and add GARP drop flow.
+    wait-and-add-garp-drop-flow() {
+      local bridge="$1"
+      local max_retries=60
```
danwinship: > ci is failing because the wait-and-add-drop-flows function I added is blocking, it currently has a retry window of 60 seconds, but its blocking, and in a ci cluster it takes longer than 60 seconds for the patch port to come up, so the container errors out, causing the failures

I don't see where the existing 60 second window you're talking about comes from, but assuming that's correct, is merely doubling it to 120 seconds sufficient?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -248,9 +247,42 @@ data:
...
+
+        (( retries += 1 ))
+
+        # Log once after 30 seconds to indicate we're still waiting
+        if [[ "${retries}" -eq 15 && "${logged_waiting}" == "false" ]]; then
+          echo "$(date -Iseconds) - INFO - Waiting for patch port on ${bridge} to be ready..."
+          logged_waiting=true
+        fi
+
+        if [[ "${retries}" -lt "${max_retries}" ]]; then
```
danwinship: this is kind of just adding complexity to let it log the error message 2 seconds sooner than it would have otherwise...

danwinship (commented):
Ben poked me on Slack to look at this. I'm not familiar with the original bug and am just commenting on whether this PR does what it claims, not whether it's actually the right fix...

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
bpickard22: removed

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
bpickard22: removed the logic here as well

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
bpickard22: in my first iteration I was retrying for 30 seconds but if the script failed to write the flow in that time I would return error instead of a warning. CI cluster was routinely taking ~60 seconds before the patch port would be created. I think that 120 seconds _should_ be sufficient, but I bumped it up to 180 just to be conservative. This runs in the background anyways so the only downside is now it will just spin longer before we get the failure message, but this does give odds to ensure that we get the flow on the patch port

PR #2915, opened 2026-02-27
GCP-430: Wire GCP WIF credentials for CNCC in HyperShift HCP mode

apahim:
- Read GCP_CNCC_CREDENTIALS_FILE env var (set by CPO) and set GOOGLE_APPLICATION_CREDENTIALS on the CNCC container so that CNCC's auto-detection (cloud-network-config-controller PR [#206](https://github.com/openshift/cloud-network-config-controller/pull/206)) picks up the WIF external_account credential.
- Add --token-audience=openshift to the cloud-token minter so GCP STS can validate the projected service-account token against the OIDC provider's allowed audience.
- Add unit tests for template rendering.

## Summary by CodeRabbit

* **New Features**
  * Added support for conditional Google Cloud Workload Identity credentials in cloud network configuration.
  * Enabled token-audience configuration for OpenShift token minter.

* **Tests**
  * Added tests to verify GCP credential injection/omission and path handling.
  * Added tests to verify token-audience is present in the cloud token minter.

### DIFF
--- a/pkg/network/cloud_network.go
+++ b/pkg/network/cloud_network.go
@@ -109,6 +109,15 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 		data.Data["AzureManagedCertDirectory"] = azureCertPath
 		data.Data["AzureManagedCredsPath"] = filepath.Join(azureCertPath, os.Getenv("MANAGED_AZURE_HCP_CREDENTIALS_FILE_PATH"))
 		data.Data["AzureManagedSecretProviderClass"] = os.Getenv("ARO_HCP_SECRET_PROVIDER_CLASS")
+		// GCP WIF credential path for HCP deployments.
+		gcpCredsFile := os.Getenv("GCP_CNCC_CREDENTIALS_FILE")
+		if gcpCredsFile == "" {
+			data.Data["GCPCredentialsPath"] = ""
+		} else if filepath.Base(gcpCredsFile) != gcpCredsFile {
+			return nil, errors.Errorf("invalid GCP_CNCC_CREDENTIALS_FILE %q: must be a filename", gcpCredsFile)
+		} else {
+			data.Data["GCPCredentialsPath"] = filepath.Join("/etc/secret/cloudprovider", gcpCredsFile)
+		}
 		caOverride.ObjectMeta = metav1.ObjectMeta{
 			Namespace:   hcpCfg.Namespace,
 			Name:        "cloud-network-config-controller-kube-cloud-config",
--- a/pkg/network/cloud_network_test.go
+++ b/pkg/network/cloud_network_test.go
@@ -1,0 +1,173 @@
+package network
+
+import (
+	"testing"
+
+	"github.com/openshift/cluster-network-operator/pkg/render"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func makeManagedControllerRenderData() render.RenderData {
+	data := render.MakeRenderData()
+	data.Data["ReleaseVersion"] = "4.18.0"
+	data.Data["PlatformType"] = "GCP"
+	data.Data["PlatformRegion"] = "us-central1"
+	data.Data["PlatformTypeAWS"] = "AWS"
+	data.Data["PlatformTypeAzure"] = "Azure"
+	data.Data["PlatformTypeGCP"] = "GCP"
+	data.Data["CloudNetworkConfigControllerImage"] = "test-image"
+	data.Data["KubernetesServiceURL"] = "https://localhost:6443"
+	data.Data["ExternalControlPlane"] = true
+	data.Data["PlatformAzureEnvironment"] = ""
+	data.Data["PlatformAWSCAPath"] = ""
+	data.Data["PlatformAPIURL"] = ""
+	data.Data["CLIImage"] = "cli-image"
+	data.Data["TokenMinterImage"] = "token-minter-image"
+	data.Data["TokenAudience"] = "https://issuer.example.com"
+	data.Data["ManagementClusterName"] = "test-cluster"
+	data.Data["HostedClusterNamespace"] = "test-ns"
+	data.Data["ReleaseImage"] = "release-image"
+	data.Data["HCPNodeSelector"] = map[string]string{}
+	data.Data["HCPLabels"] = map[string]string{}
+	data.Data["HCPTolerations"] = []string{}
+	data.Data["RunAsUser"] = ""
+	data.Data["PriorityClass"] = ""
+	data.Data["HTTP_PROXY"] = ""
+	data.Data["HTTPS_PROXY"] = ""
+	data.Data["NO_PROXY"] = ""
+	data.Data["AzureManagedCertDirectory"] = ""
+	data.Data["AzureManagedCredsPath"] = ""
+	data.Data["AzureManagedSecretProviderClass"] = ""
+	data.Data["GCPCredentialsPath"] = ""
+	return data
+}
+
+// getEnvVar looks up an env var by name from a container map and returns its value.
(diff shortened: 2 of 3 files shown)


pkg/network/cloud_network.go
```diff
@@ -109,6 +109,13 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 		data.Data["AzureManagedCertDirectory"] = azureCertPath
 		data.Data["AzureManagedCredsPath"] = filepath.Join(azureCertPath, os.Getenv("MANAGED_AZURE_HCP_CREDENTIALS_FILE_PATH"))
 		data.Data["AzureManagedSecretProviderClass"] = os.Getenv("ARO_HCP_SECRET_PROVIDER_CLASS")
+		// GCP WIF credential path for HCP deployments.
+		gcpCredsFile := os.Getenv("GCP_CNCC_CREDENTIALS_FILE")
```
ricky-rav, line 113: In the commit message you say that GCP_CNCC_CREDENTIALS_FILE is set by CPO. What's CPO? You're setting this env var in this other PR, right? https://github.com/openshift/hypershift/pull/7824/changes

pkg/network/cloud_network_test.go
```diff
@@ -0,0 +1,158 @@
...
+	data.Data["HCPTolerations"] = []string{}
+	data.Data["RunAsUser"] = ""
+	data.Data["PriorityClass"] = ""
+	data.Data["HTTP_PROXY"] = ""
+	data.Data["HTTPS_PROXY"] = ""
+	data.Data["NO_PROXY"] = ""
+	data.Data["AzureManagedCertDirectory"] = ""
+	data.Data["AzureManagedCredsPath"] = ""
+	data.Data["AzureManagedSecretProviderClass"] = ""
+	data.Data["GCPCredentialsPath"] = ""
```
ricky-rav: You can probably use a helper function to set up `data.Data` for your tests

pkg/network/cloud_network_test.go
```diff
@@ -0,0 +1,158 @@
...
+				if obj.GetKind() != "Deployment" {
+					continue
+				}
+				yaml, err := obj.MarshalJSON()
+				if err != nil {
+					t.Fatalf("failed to marshal deployment: %v", err)
+				}
+				yamlStr := string(yaml)
+
+				hasGoogleAppCreds := strings.Contains(yamlStr, "GOOGLE_APPLICATION_CREDENTIALS")
```
ricky-rav: Here and below, every time you call `strings.Contains`, you're just looking for a substring in the entire yaml. I think it's more appropriate to actually parse the object and look for exactly the parameters you're interested in and check their value. For example, existing unit tests use `uns.NestedSlice` and `uns.NestedString` to do that.

ricky-rav (changes requested):

pkg/network/cloud_network.go
```diff
@@ -109,6 +109,13 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
...
+		if gcpCredsFile != "" {
+			data.Data["GCPCredentialsPath"] = filepath.Join("/etc/secret/cloudprovider", gcpCredsFile)
+		} else {
+			data.Data["GCPCredentialsPath"] = ""
```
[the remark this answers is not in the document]
apahim: Added filepath.Base() validation, it now rejects values that aren't plain filenames

pkg/network/cloud_network_test.go
apahim: Created makeManagedControllerRenderData() helper, used by both tests

pkg/network/cloud_network_test.go
apahim: Replaced raw JSON string matching with uns.NestedSlice, uns.NestedString, and uns.NestedStringSlice to properly parse unstructured objects and look up specific containers, env vars, and args

pkg/network/cloud_network.go
apahim: Fixed the commit message.

Yes,the GCP_CNCC_CREDENTIALS_FILE comes from hypershift

pkg/network/cloud_network.go
```diff
@@ -109,6 +109,15 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
...
 		data.Data["AzureManagedSecretProviderClass"] = os.Getenv("ARO_HCP_SECRET_PROVIDER_CLASS")
+		// GCP WIF credential path for HCP deployments.
+		gcpCredsFile := os.Getenv("GCP_CNCC_CREDENTIALS_FILE")
+		if gcpCredsFile == "" {
+			data.Data["GCPCredentialsPath"] = ""
+		} else if filepath.Base(gcpCredsFile) != gcpCredsFile {
+			return nil, errors.Errorf("invalid GCP_CNCC_CREDENTIALS_FILE %q: must be a filename", gcpCredsFile)
+		} else {
+			data.Data["GCPCredentialsPath"] = filepath.Join("/etc/secret/cloudprovider", gcpCredsFile)
+		}
```
danwinship, line 120: So... if the feature is configured, hypershift passes the literal string `"application_default_credentials.json"` to CNO, which then prepends the literal string `"/etc/secret/cloudprovider/"` and passes it to CNCC, which uses it as a filename.

This would make sense if hypershift knew the credentials filename but not the directory it was in, and CNO knew the directory but not the filename. But I'm guessing that's not the case.

Where do the creds come from? Who writes them there?

If HyperShift actually knows for sure what the credentials filename is (and nobody else does know for sure) then it should pass the whole path to CNO, which should pass it to CNCC.

If HyperShift doesn't know where the creds are, but only knows whether or not it wants CNCC to use them, then it should just pass a true/false value to CNO, and then likewise, CNO can either pass the full pathname or a true/false value to CNCC.

Also, it seems like it would be better to configure CNCC via a command-line argument...

apahim: Hypershift creates the secret cloud-network-config-controller-creds containing the WIF credential JSON under the key application_default_credentials.json

- https://github.com/openshift/hypershift/pull/7824

Here specifically:

- https://github.com/apahim/hypershift/blob/2eef5fe23aa36a1469d1d6b8bd463c9dcf152f4b/hypershift-operator/controllers/hostedcluster/internal/platform/gcp/gcp.go#L434-L441

But that's becuase the secret name is hardcoded in CNO: 

- https://github.com/openshift/cluster-network-operator/blob/bc5af87/bindata/cloud-network-config-controller/managed/controller.yaml#L251-L253

Then CNO mounts it in /etc/secret/cloudprovider:

- https://github.com/openshift/cluster-network-operator/blob/bc5af87/bindata/cloud-network-config-controller/managed/controller.yaml#L202-L204

Then we have to create the env var for CNO with the filename:

- https://github.com/apahim/hypershift/blob/2eef5fe23aa36a1469d1d6b8bd463c9dcf152f4b/control-plane-operator/controllers/hostedcontrolplane/v2/cno/deployment.go#L162-L169

The filename-only approach follows the existing Azure pattern at line 110, where MANAGED_AZURE_HCP_CREDENTIALS_FILE_PATH similarly passes a filename that CNO combines with a mount path (/var/run/secrets/azure).

Each component owns its part of the path: HyperShift controls the secret key name, CNO controls the mount point.

The env var also doubles  as a feature flag - empty means GCP WIF is not configured.

Passing the full path would couple HyperShift to CNO's mount point; passing a boolean would require hardcoding the filename in CNO, coupling it to HyperShift's secret key choice. The current split keeps both sides independent.

Regarding configuring CNCC: the env var set on the CNCC container is GOOGLE_APPLICATION_CREDENTIALS, which is a standard GCP SDK convention. The client libraries read it  automatically to locate credentials... It's not an arbitrary env var, so a CLI argument would add an extra layer that just sets this env var anyway.

danwinship: huh... ok

PR #2955, opened 2026-04-06
OCPBUGS-62144: Makes sure rendering for egress IP reachabilityTimeout triggers restart of ovnkube (node/control) pods

raphaelvrosa:
Issue: EgressIP `ReachabilityTotalTimeoutSeconds` is rendered by ovnkube-config (configmap), but ovnkube node and control pods are not restarted to have the new value applied in their settings.

Solution: Make sure the spec templates of ovnkube node and control-plane pods take into account the changes in the `ReachabilityTotalTimeoutSeconds` setting. This is implemented by rendering the cli flags in their spec template command. When this value is changed, it then triggers the  restart of those pods, reloading the parameter in their configuration.

Adds test cases to evaluate the rendering and the hash changes in ovnkube node and control pod specs.
Updates README to document that reachabilityTimeoutSeconds is enabled for runtime updates.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring egress IP reachability timeout as a runtime-modifiable setting in OVN Kubernetes deployments.

* **Tests**
  * Added comprehensive test coverage for the reachability timeout configuration across different deployment scenarios.

* **Documentation**
  * Updated configuration documentation to reflect the new runtime-configurable reachability timeout field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ Other values are ignored. If you wish to use use a third-party network provider 
 
 
 ### Configuring OVNKubernetes
-OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed except for `gatewayConfig` and `IPsec` which can be changed at runtime:
+OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed except for `gatewayConfig`, `IPsec` and `reachabilityTotalTimeoutSeconds` which can be changed at runtime:
 * `MTU`: The MTU to use for the geneve overlay. The default is the MTU of the node that the cluster-network-operator is first run on, minus 100 bytes for geneve overhead. If the nodes in your cluster don't all have the same MTU then you may need to set this explicitly.
 * `genevePort`: The UDP port to use for the Geneve overlay. The default is 6081.
 * `hybridOverlayConfig`: hybrid linux/windows cluster (see below).
--- a/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
+++ b/bindata/network/ovn-kubernetes/common/008-script-lib.yaml
@@ -575,6 +575,11 @@ data:
 
       echo "I$(date "+%m%d %H:%M:%S.%N") - starting ovnkube-node"
 
+      ovn_eip_reachability_timeout_opt=
+      {{- if .ReachabilityTotalTimeoutSeconds }}        
+        ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
+      {{ end }}
+
       init_ovnkube_controller="--init-ovnkube-controller ${K8S_NODE}"
       gateway_interface="br-ex"
       OVN_NODE_MODE=${OVN_NODE_MODE:-full}
@@ -754,5 +759,6 @@ data:
         ${ovn_v6_masquerade_subnet_opt} \
         ${ovn_v4_transit_switch_subnet_opt} \
         ${ovn_v6_transit_switch_subnet_opt} \
-        ${dpu_lease_flags}
+        ${dpu_lease_flags} \
+        ${ovn_eip_reachability_timeout_opt}
     }
--- a/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
@@ -156,6 +156,11 @@ spec:
             fi
           done
 
+          ovn_eip_reachability_timeout_opt=
+          {{- if .ReachabilityTotalTimeoutSeconds }}                 
+            ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
+          {{ end }}
+
           ovn_v4_join_subnet_opt=
           if [[ "{{.V4JoinSubnet}}" != "" ]]; then
             ovn_v4_join_subnet_opt="--gateway-v4-join-subnet {{.V4JoinSubnet}}"
@@ -216,7 +221,8 @@ spec:
             ${ovn_v6_masquerade_subnet_opt} \
             ${persistent_ips_enabled_flag} \
             ${route_advertisements_enable_flag} \
-            ${evpn_enable_flag}
+            ${evpn_enable_flag} \
+            ${ovn_eip_reachability_timeout_opt}
         volumeMounts:
         - mountPath: /run/ovnkube-config/
           name: ovnkube-config
--- a/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
@@ -102,6 +102,11 @@ spec:
             source "/env/_master"
             set +o allexport
           fi
+          
+          ovn_eip_reachability_timeout_opt=
+          {{- if .ReachabilityTotalTimeoutSeconds }}        
+            ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
+          {{ end }}
 
           ovn_v4_join_subnet_opt=
           if [[ "{{.V4JoinSubnet}}" != "" ]]; then
@@ -170,7 +175,8 @@ spec:
             ${persistent_ips_enabled_flag} \
             ${gateway_mode_flags} \
             ${route_advertisements_enable_flag} \
-            ${evpn_enable_flag}
+            ${evpn_enable_flag} \
+            ${ovn_eip_reachability_timeout_opt}
         volumeMounts:
         - mountPath: /run/ovnkube-config/
           name: ovnkube-config
--- a/pkg/network/ovn_kubernetes_test.go
+++ b/pkg/network/ovn_kubernetes_test.go
@@ -411,7 +411,6 @@ enable-egress-ip=true
 enable-egress-firewall=true
 enable-egress-qos=true
 enable-egress-service=true
-egressip-reachability-total-timeout=3
 egressip-node-healthcheck-port=9107
 enable-multi-network=true
 enable-network-segmentation=true
@@ -473,7 +472,6 @@ enable-egress-ip=true
 enable-egress-firewall=true
 enable-egress-qos=true
 enable-egress-service=true
-egressip-reachability-total-timeout=0
 egressip-node-healthcheck-port=9107
 enable-multi-network=true
 enable-network-segmentation=true
@@ -4029,6 +4027,151 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
 	g.Expect(progressing).To(BeFalse())
 
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("CustomResourceDefinition", "", "ipamclaims.k8s.cni.cncf.io")))
+}
+
+// TestRenderOVNKubernetesReachability tests egress IP reachability timeout rendering
+func TestRenderOVNKubernetesReachability(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	testCases := []struct {
+		name                                string
+		reachabilityTimeout                 *uint32
+		expectKubernetesFeatureReachability bool
+		expectErr                           bool
+	}{
+		{
+			name:                                "No reachability timeout (nil)",
+			reachabilityTimeout:                 nil,
+			expectKubernetesFeatureReachability: false,
+			expectErr:                           false,
+		},
+		{
+			name:                                "Reachability timeout set to 0",
+			reachabilityTimeout:                 ptrToUint32(0),
+			expectKubernetesFeatureReachability: true,
+			expectErr:                           false,
+		},
+		{
+			name:                                "Reachability timeout changed to 10",
+			reachabilityTimeout:                 ptrToUint32(10),
+			expectKubernetesFeatureReachability: true,
+			expectErr:                           false,
+		},
+		{
+			name:                                "Reachability timeout unchanged to 10",
+			reachabilityTimeout:                 ptrToUint32(10),
+			expectKubernetesFeatureReachability: true,
+			expectErr:                           false,
+		},
+		{
+			name:                                "Reachability timeout changed to 5",
+			reachabilityTimeout:                 ptrToUint32(5),
+			expectKubernetesFeatureReachability: true,
+			expectErr:                           false,
+		},
+		{
+			name:                                "Reachability timeout disabled",
+			reachabilityTimeout:                 nil,
+			expectKubernetesFeatureReachability: false,
+			expectErr:                           false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			crd := OVNKubernetesConfig.DeepCopy()
+			config := &crd.Spec
+			config.DefaultNetwork.OVNKubernetesConfig.EgressIPConfig.ReachabilityTotalTimeoutSeconds = tc.reachabilityTimeout
+
+			errs := validateOVNKubernetes(config)
+			g.Expect(errs).To(HaveLen(0))
+			fillDefaults(config, nil)
+
+			// at the same time we have an upgrade
+			t.Setenv("RELEASE_VERSION", "2.0.0")
+
+			bootstrapResult := fakeBootstrapResult()
+			bootstrapResult.OVN = bootstrap.OVNBootstrapResult{
+				ControlPlaneReplicaCount: 3,
+				OVNKubernetesConfig: &bootstrap.OVNConfigBoostrapResult{
+					DpuHostModeLabel:     OVN_NODE_SELECTOR_DEFAULT_DPU_HOST,
+					DpuModeLabel:         OVN_NODE_SELECTOR_DEFAULT_DPU,
+					SmartNicModeLabel:    OVN_NODE_SELECTOR_DEFAULT_SMART_NIC,
+					MgmtPortResourceName: "",
+					HyperShiftConfig: &bootstrap.OVNHyperShiftBootstrapResult{
+						Enabled: false,
+					},
+				},
+			}
+
+			featureGatesCNO := getDefaultFeatureGates()
+			fakeClient := cnofake.NewFakeClient()
+			// Set is as Hypershift hosted control plane.
+			bootstrapResult.Infra = bootstrap.InfraStatus{}
+			bootstrapResult.Infra.HostedControlPlane = &hypershift.HostedControlPlane{}
+			objs, _, err := renderOVNKubernetes(config, bootstrapResult, manifestDirOvn, fakeClient, featureGatesCNO)
+			if tc.expectErr {
+				g.Expect(err).To(HaveOccurred())
+				return
+			}
+			g.Expect(err).NotTo(HaveOccurred())
+
+			var configMap *uns.Unstructured
+			var scriptCP string
+			var scriptNode string
+			for _, obj := range objs {
+				// Gets script that starts ovnkube-control-plane pod, ovnkube-cluster-manager container
+				if obj.GetKind() == "Deployment" && obj.GetName() == "ovnkube-control-plane" && obj.GetNamespace() == "openshift-ovn-kubernetes" {
+					containers, found, err := uns.NestedSlice(obj.Object, "spec", "template", "spec", "containers")
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(found).To(BeTrue())
+					for _, c := range containers {
+						cm := c.(map[string]interface{})
+						if name, ok := cm["name"]; ok && name == "ovnkube-cluster-manager" {
+							command, found, err := uns.NestedSlice(cm, "command")
+							g.Expect(err).NotTo(HaveOccurred())
+							g.Expect(found).To(BeTrue())
+							g.Expect(len(command)).To(BeNumerically(">", 2)) // Command options are set in the 3rd slice
+							scriptCP = command[2].(string)
+							break
+						}
+					}
+				}
+
+				// Gets script that starts ovnkube-node pod, ovnkube-controller container
+				if obj.GetKind() == "ConfigMap" && obj.GetName() == "ovnkube-script-lib" {
+					configMap = obj
+					configMapData, found, err := uns.NestedStringMap(configMap.Object, "data")
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(found).To(BeTrue(), "ovnkube-script-lib ConfigMap should have data field")
+					scriptNode = configMapData["ovnkube-lib.sh"]
+				}
+			}
+			g.Expect(scriptCP).NotTo(BeEmpty())
+			g.Expect(scriptNode).NotTo(BeEmpty())
+
+			if tc.expectKubernetesFeatureReachability {
+				g.Expect(scriptCP).To(
+					ContainSubstring(fmt.Sprintf("--egressip-reachability-total-timeout %d", *tc.reachabilityTimeout)),
+					"ovnkube-control-plane pod template should contain the configured reachability timeout value",
+				)
+				g.Expect(scriptNode).To(
+					ContainSubstring(fmt.Sprintf("--egressip-reachability-total-timeout %d", *tc.reachabilityTimeout)),
+					"ovnkube-node pod template should contain the configured reachability timeout value",
+				)
+
+			} else {
+				g.Expect(scriptCP).NotTo(
+					ContainSubstring("--egressip-reachability-total-timeout"),
+					"ovnkube-control-plane pod template should not contain the configured reachability timeout value",
+				)
+				g.Expect(scriptNode).NotTo(
+					ContainSubstring("--egressip-reachability-total-timeout"),
+					"ovnkube-node pod template should not contain the configured reachability timeout value",
+				)
+			}
+		})
+	}
 }
 
 type fakeClientReader struct {
--- a/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml
@@ -40,9 +40,6 @@ data:
     enable-egress-firewall=true
     enable-egress-qos=true
     enable-egress-service=true
-    {{- if .ReachabilityTotalTimeoutSeconds }}
-    egressip-reachability-total-timeout={{.ReachabilityTotalTimeoutSeconds}}
-    {{- end }}
 {{- if .ReachabilityNodePort }}
     egressip-node-healthcheck-port={{.ReachabilityNodePort}}
 {{- end }}


pkg/network/ovn_kubernetes.go
```diff
@@ -504,7 +504,9 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	// Compute a separate hash for no-overlay node config (outboundSNAT).
 	// This allows ovnkube-node to restart when outboundSNAT changes
 	nodeNoOverlayHash := sha1.New()
-	nodeNoOverlayHashData := fmt.Sprintf("outboundSNAT=%v", data.Data["NoOverlayOutboundSNAT"])
+	nodeNoOverlayHashData := fmt.Sprintf("outboundSNAT=%v,reachabilityTimeout=%v",
+		data.Data["NoOverlayOutboundSNAT"],
+		data.Data["ReachabilityTotalTimeoutSeconds"])
```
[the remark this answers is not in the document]
raphaelvrosa: Implemented.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,170 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+			ovnkubeConf := configMapData["ovnkube.conf"]
+			if tc.expectKubernetesFeatureReachability {
+				g.Expect(ovnkubeConf).To(ContainSubstring("[ovnkubernetesfeature]"),
+					"ConfigMap should contain [ovnkubernetesfeature] section when enabled")
+				g.Expect(ovnkubeConf).To(ContainSubstring("egressip-reachability-total-timeout="),
+					"ConfigMap should contain egressip-reachability-total-timeout=")
+			} else {
+				g.Expect(ovnkubeConf).NotTo(ContainSubstring("egressip-reachability-total-timeout="),
+					"ConfigMap should not contain egressip-reachability-total-timeout= section when disabled")
+			}
```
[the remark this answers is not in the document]
raphaelvrosa: Implemented.

pkg/network/ovn_kubernetes.go
```diff
@@ -501,10 +501,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	}
 	data.Data["OVNKubeConfigHash"] = hex.EncodeToString(h.Sum(nil))
 
+	var reachabilityTimeoutStr string
```
raphaelvrosa: Make sure hash is consistent when values do not change.

raphaelvrosa:
Fix in some CI tests depend on: https://github.com/openshift/origin/pull/30961/changes

raphaelvrosa:
> @raphaelvrosa: This pull request references [Jira Issue OCPBUGS-62144](https://redhat.atlassian.net/browse/OCPBUGS-62144), which is invalid:
> 
>     * expected the bug to target the "4.22.0" version, but no target version was set
> 
> 
> Comment `/jira refresh` to re-evaluate validity if changes to the Jira bug are made, or edit the title of this pull request to link to a different bug.
> 
> The bug has been updated to refer to the pull request using the external bug tracker.
> Details
> 
> In response to [this](https://github.com/openshift/cluster-network-operator/pull/2955):
> 
> > Implements rendering of EgressIP `ReachabilityTotalTimeoutSeconds`.
> > Adds test cases to evaluate rendering.
> > Updates README, enable runtime update of setting.
> 
> Instructions for interacting with me using PR comments are available [here](https://prow.ci.openshift.org/command-help?repo=openshift%2Fcluster-network-operator). If you have questions or suggestions related to my behavior, please file an issue against the [openshift-eng/jira-lifecycle-plugin](https://github.com/openshift-eng/jira-lifecycle-plugin/issues/new) repository.

/jira refresh

pkg/network/ovn_kubernetes.go
```diff
@@ -501,10 +501,17 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
...
+	if timeout, ok := data.Data["ReachabilityTotalTimeoutSeconds"].(*uint32); ok && timeout != nil {
+		reachabilityTimeoutStr = strconv.FormatUint(uint64(*timeout), 10)
+	}
+
 	// Compute a separate hash for no-overlay node config (outboundSNAT).
 	// This allows ovnkube-node to restart when outboundSNAT changes
 	nodeNoOverlayHash := sha1.New()
-	nodeNoOverlayHashData := fmt.Sprintf("outboundSNAT=%v", data.Data["NoOverlayOutboundSNAT"])
+	nodeNoOverlayHashData := fmt.Sprintf("outboundSNAT=%v,reachabilityTimeout=%s",
```
arghosh93, line 512: This will only help in having a new hash data when timeout value changes or get updated. Based on that it will also redeploy ovnkube-node PODs as the hash added to the annotation of the daemonset would get updated. 
However there is no way ovnkube-node can know about the exact timeout value to pass it to the application so that egressIP healthchecks can be done accordingly. 
You have to find a way such that CNO can update https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L1201 cli flag and the application can configure further things as per the new value upon application/POD restart.

pkg/network/ovn_kubernetes.go
```diff
@@ -514,9 +521,10 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 	// This allows ovnkube-control-plane to restart when bgpManagedConfig changes,
 	// without restarting ovnkube-node pods.
 	cpHash := sha1.New()
-	cpHashData := fmt.Sprintf("asNumber=%v,topology=%v",
+	cpHashData := fmt.Sprintf("asNumber=%v,topology=%v,reachabilityTimeout=%s",
```
arghosh93, line 524: Above comment applies here as well. This only gonna make sure that the new hash value get updated at the ovnkube-control-plane deployment. But ovnkube application is not gonna be abe to decode. 
This mechanism is only to trigger a redeployment of PODs and not to render data to application.

arghosh93 (changes requested):
You will have to find a way to render timeout values to both ovnkube-node and ovnkube-control-plane, and also make sure that both gets redeployed. When ovnkube gets to know about the new timeout value either through an environment variable or through CLI flag, then it can take care of setting things up on the ovnkube side, as implementation for that is already in place.

pkg/network/ovn_kubernetes.go
raphaelvrosa: The value is already rendered in the data structure (https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/ovn_kubernetes.go#L418) and then set in the ovnkube-config ConfigMap (https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/ovn_kubernetes.go#L486). 
The test case added in this PR validates both the changes in the value of the rendered setting, with and without hash changes.  
The change was also tested in an actual ovn-k8s deployment, where runtime changes in settings triggered the restart of the ovnkube-* pods, re-initialized with the updated values of the field.

pkg/network/ovn_kubernetes.go
raphaelvrosa: Same as above. 
The cli config of both ovnkube-node and ovnkube-control-plane pods are extracted from ovnkube-config ConfigMap. When CNO renders the data (as mentioned above) it triggers the restart of these pods.
This PRs addresses the bug, before the value was being changed in the ConfigMap, but the pods were not being restarted.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,175 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(found).To(BeTrue(), "ConfigMap should have data field")
+			ovnkubeConf := configMapData["ovnkube.conf"]
+			if tc.expectKubernetesFeatureReachability {
+				g.Expect(ovnkubeConf).To(ContainSubstring("[ovnkubernetesfeature]"),
+					"ConfigMap should contain [ovnkubernetesfeature] section when enabled")
+				g.Expect(tc.reachabilityTimeout).NotTo(BeNil())
+				g.Expect(ovnkubeConf).To(
+					ContainSubstring(fmt.Sprintf("egressip-reachability-total-timeout=%d", *tc.reachabilityTimeout)),
+					"ConfigMap should contain the configured reachability timeout value",
```
raphaelvrosa, line 4065: @arghosh93 this validates the rendering of the changed value.

pkg/network/ovn_kubernetes.go
arghosh93: wonderful! Thanks for pointing me to the right section of the code. Now I understand the issue better. Earlier I thought both ovnkube-control-plane and ovnkube-node were not being redeployed upon changing the timeout. But now it seems that ovnkube-node is being redeployed but ovnube-control-plane is not. 

Rendering timeout value to ovnkube-config configmap only would not trigger rollout and hence you are adding it for hash creation and annotate ovnube-control-plane so that it redeploys.

This method would work but there is a better way of implementing this. To trigger the redeployment of ovnube-control-plane we generally export  a variable to ovnube-control-plane deployment and pass it using the CLI flag to the ovnkube application. https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml#L106-L212

We did not do it for No-Overlay mode as we dont have corresponding CLI flags for those fields. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L898-L1877

We dont follow a similar way for ovnkube-node as it needs a lot more data than ovnube-control-plane and the daemonset file was getting bigger and bigger and hence we have moved command part of the daemonset to a configmap and annotate the daemonset so that redeployment occurs.

pkg/network/ovn_kubernetes.go
arghosh93: i think ovnkube-control-plane is not being restarted with this bug.

raphaelvrosa: Before this PR, none of the pods (ovnkube-node and ovnkube-control-plane) were being restarted.
Now, with this PR, both changes in the hashes of ovnkube-node and ovnkube-control-plane specs trigger the restart of the pods. This was validated in a running cluster.

pkg/network/ovn_kubernetes.go
raphaelvrosa: Before this PR, none of the pods (ovnkube-node and ovnkube-control-plane) were being restarted when the setting (eip timeoutSeconds) was changed.
With this PR, the hash annotations are updated in both pods, enabling their restart.

Why the variable export and passing of a cli flag is a better approach? 

ovnkube-control-plane retrieves the cli config from the ConfigMap settings when it is restarted already, this is the default behavior of the pod.
And the hash exists to signal the ConfigMap settings have changed, so the PR implements the hash approach. 
And in ovnkube-node the same approach is already enabled by the current implementation, the hash approach signals the need to restart the pod.

raphaelvrosa: @arghosh93 the implementation is updated using a cli flag for ovnkube-control-plane. Please check.
How should it be done for ovnkube-node?

pkg/network/ovn_kubernetes.go
raphaelvrosa: Removed the hash implementation.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: Removed the hash implementation.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
```diff
@@ -102,6 +102,13 @@ spec:
             source "/env/_master"
             set +o allexport
           fi
+          
+{{- if .ReachabilityTotalTimeoutSeconds }}        
+          ovn_eip_reachability_timeout_opt=
+          if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
+            ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
+          fi
+{{- end }}
```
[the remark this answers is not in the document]
raphaelvrosa: Fixed.

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -156,6 +156,13 @@ spec:
             fi
           done
 
+{{- if .ReachabilityTotalTimeoutSeconds }}        
+          ovn_eip_reachability_timeout_opt=
+          if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
+            ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
+          fi
+{{- end }}
```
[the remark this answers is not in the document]
raphaelvrosa: Fixed.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,159 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+			} else {
+				g.Expect(ovnkubeConf).NotTo(ContainSubstring("egressip-reachability-total-timeout="),
+					"ConfigMap should not contain egressip-reachability-total-timeout= section when disabled")
+
+				g.Expect(scriptCP).NotTo(
+					ContainSubstring("--egressip-reachability-total-timeout"),
+					"ovnkube-control-plane pod template should not contain the configured reachability timeout value",
+				)
+			}
+		})
```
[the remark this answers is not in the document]
raphaelvrosa: The goal of the changes is no longer to change the hash, but to change the command inside the pod spec templates.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -574,6 +571,13 @@ data:
 
       # enable egress ip, egress firewall, egress qos, egress service
       egress_features_enable_flag="--enable-egress-ip=true --enable-egress-firewall=true --enable-egress-qos=true --enable-egress-service=true"
+
+      {{- if .ReachabilityTotalTimeoutSeconds }}
+      if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
+        egress_features_enable_flag="$egress_features_enable_flag --egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
```
arghosh93: I understand your confusion and I would like to clarify that egressip-reachability-total-timeout  is not a feature and rather a configuration parameter. We better not add it to egress_features_enable_flag and keep it independent.

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -156,6 +156,13 @@ spec:
             fi
           done
 
+          {{- if .ReachabilityTotalTimeoutSeconds }}                 
+          ovn_eip_reachability_timeout_opt=
```
arghosh93, line 159: We are not initializing ovn_eip_reachability_timeout_opt unless ReachabilityTotalTimeoutSeconds is true. However while starting ovnkube we are always using ovn_eip_reachability_timeout_opt.  I think it would create a problem when the Egress IP reachability timeout is not mentioned at network.operator.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: I dont see use of egress_features_enable_flag while starting ovnkube at the bottom.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
```diff
@@ -102,6 +102,13 @@ spec:
             source "/env/_master"
             set +o allexport
           fi
+          
+          {{- if .ReachabilityTotalTimeoutSeconds }}        
+          ovn_eip_reachability_timeout_opt=
```
arghosh93, line 106: same as above.

README.md
```diff
@@ -146,7 +146,7 @@ Other values are ignored. If you wish to use use a third-party network provider
 
 
 ### Configuring OVNKubernetes
-OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed except for `gatewayConfig` and `IPsec` which can be changed at runtime:
+OVNKubernetes supports the following configuration options, all of which are optional and once set at cluster creation, they can't be changed except for `gatewayConfig`, `IPsec` and `reachabilityTotalTimeoutSeconds` which can be changed at runtime:
```
arghosh93, line 149: I think reachabilityTotalTimeoutSeconds can be changed after setting it once.

arghosh93 (commented):
Please remove https://github.com/raphaelvrosa/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml#L40 if thats appropriate.

README.md
raphaelvrosa: Yes, and ` they can't be changed except for` this is what the text above mentions.
Can you clarify your comment?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
raphaelvrosa: Agreed, fixed.

raphaelvrosa: The new flag is added.

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
raphaelvrosa: Agreed, fixed.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
raphaelvrosa: Agreed, fixed.

README.md
arghosh93: Never mind, I misinterpreted.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
```diff
@@ -574,6 +571,14 @@ data:
 
       # enable egress ip, egress firewall, egress qos, egress service
       egress_features_enable_flag="--enable-egress-ip=true --enable-egress-firewall=true --enable-egress-qos=true --enable-egress-service=true"
+
+      ovn_eip_reachability_timeout_opt=
+      {{- if .ReachabilityTotalTimeoutSeconds }}        
+      if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
```
arghosh93: I think you can simply do it like below:
```
 node_mgmt_port_netdev_flags=
 if [[ -n "${ReachabilityTotalTimeoutSeconds}" ]] ; then
      ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout ${ReachabilityTotalTimeoutSeconds}"
 fi
```

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -156,6 +156,13 @@ spec:
             fi
           done
 
+          ovn_eip_reachability_timeout_opt=
+          {{- if .ReachabilityTotalTimeoutSeconds }}                 
+          if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
```
arghosh93: same as above.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
```diff
@@ -102,6 +102,13 @@ spec:
             source "/env/_master"
             set +o allexport
           fi
+          
+          ovn_eip_reachability_timeout_opt=
+          {{- if .ReachabilityTotalTimeoutSeconds }}        
+          if [[ "{{.ReachabilityTotalTimeoutSeconds}}" != "" ]]; then
```
arghosh93: same as above.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,179 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+			bootstrapResult.Infra.HostedControlPlane = &hypershift.HostedControlPlane{}
+			objs, _, err := renderOVNKubernetes(config, bootstrapResult, manifestDirOvn, fakeClient, featureGatesCNO)
+			if tc.expectErr {
+				g.Expect(err).To(HaveOccurred())
+				return
+			}
+			g.Expect(err).NotTo(HaveOccurred())
+
+			// Find the ovnkube-config ConfigMap and check the template data
+			var configMap *uns.Unstructured
```
arghosh93, line 4119: why we are checking for the configmap? Did you consider removing https://github.com/raphaelvrosa/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/self-hosted/004-config.yaml#L40 ?

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,179 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+			var scriptNode string
+			for _, obj := range objs {
+				// Gets script that starts ovnkube-control-plane pod, ovnkube-cluster-manager container
+				if obj.GetKind() == "Deployment" && obj.GetName() == "ovnkube-control-plane" && obj.GetNamespace() == "openshift-ovn-kubernetes" {
+					containers, found, err := uns.NestedSlice(obj.Object, "spec", "template", "spec", "containers")
+					g.Expect(err).NotTo(HaveOccurred())
+					g.Expect(found).To(BeTrue())
+					for _, c := range containers {
+						cm := c.(map[string]interface{})
+						if name, ok := cm["name"]; ok && (name == "ovnkube-cluster-manager" || name == "ovnkube-control-plane") {
```
arghosh93: name would be 'ovnkube-cluster-manager' I guess. Do you need to use or here?

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,179 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+							command, found, err := uns.NestedSlice(cm, "command")
+							g.Expect(err).NotTo(HaveOccurred())
+							g.Expect(found).To(BeTrue())
+							g.Expect(len(command)).To(BeNumerically(">", 2))
+							scriptCP = command[2].(string)
```
arghosh93, line 4135: why 2?

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,179 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
...
+
+			if tc.expectKubernetesFeatureReachability {
+				g.Expect(ovnkubeConf).To(ContainSubstring("[ovnkubernetesfeature]"),
+					"ConfigMap should contain [ovnkubernetesfeature] section when enabled")
+				g.Expect(tc.reachabilityTimeout).NotTo(BeNil())
+				g.Expect(ovnkubeConf).To(
+					ContainSubstring(fmt.Sprintf("egressip-reachability-total-timeout=%d", *tc.reachabilityTimeout)),
+					"ConfigMap should contain the configured reachability timeout value",
+				)
+				g.Expect(ovnkubeConf).To(ContainSubstring("egressip-reachability-total-timeout="),
```
arghosh93: why we are checking both `ContainSubstring(fmt.Sprintf("egressip-reachability-total-timeout=%d", *tc.reachabilityTimeout)),` and `ContainSubstring("egressip-reachability-total-timeout=")`.

pkg/network/ovn_kubernetes_test.go
```diff
@@ -3944,6 +3944,179 @@ func TestRenderOVNKubernetesEnablePersistentIPs(t *testing.T) {
 	g.Expect(objs).To(ContainElement(HaveKubernetesID("CustomResourceDefinition", "", "ipamclaims.k8s.cni.cncf.io")))
 }
 
+// TestRenderOVNKubernetesReachability tests egress IP reachability timeout rendering
+func TestRenderOVNKubernetesReachability(t *testing.T) {
```
arghosh93, line 4033: if we are doing this, should we do this for all features and not only for egressIP reachability timeout? We probably cant afford to have similar block for all features. WDYT?

arghosh93 (changes requested):

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: The goal of this PR is to trigger the reboot of ovnkube-* pods. Adding the verification of rendering for all the features in CNO is another topic. There are tests checking the rendering, but I guess not all the dynamic/run-time changes.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: Indeed, forgot to clean this from the previous changes, fixed.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: `control-plane` comes from previous mode. In interconnect mode only `cluster-manager` is used. This was copied from other test cases.
Fixed to keep `cluster-manager` only.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: In the spec template of the pod, the command is defined as follows:
```
        command:
        - /bin/bash
        - -c
        - |
```
Meaning the string containing the options come only in the 3rd part of the string slice of command.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: The goal is just to reboot the pods in this PR.
Besides, the goal is to keep all the configs consistent, so there is no reason to remove the value from the ConfigMap.

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
raphaelvrosa: The first check `{{- if .ReachabilityTotalTimeoutSeconds }}` is important to validate if the value is present (not nil) in the values of the spec template. 
The inner verification cannot validate the difference between the nil value and the "zero" value.

arghosh93: Do you think the suggestion I gave would not work in this case?  Can you see how it is done for similar variables in the same file?

pkg/network/ovn_kubernetes_test.go
arghosh93: Can you tell me what is the point of keeping it in the configmap? Earlier the configuration was being added to the configmap but redeployment was not being triggered. Now to redeploy, we are going to feed this config to ovnkube using cli. so, why dont we get rid of adding it to the configmap?

pkg/network/ovn_kubernetes_test.go
arghosh93: Could you do a follow-up PR, please?

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
raphaelvrosa: No, it does not work. 
It won't pass the test case ` TestRenderOVNKubernetesReachability/No_reachability_timeout_(nil)` .
You can try it, remove the {{ if ... }} {{ end }} statements in 008-script-lib.yaml and run the tests `go test -v -count=1 ./pkg/network/ -run TestRenderOVNKubernetesReachability`. 

In that case, when the value of the setting is nil (not present) it is going to set the keyword in the args of the command. This should not happen. So the test fails.

From my understanding, as this is a integer the nil value is considered as a zero value, so it adds the flag (the test checks if the flag is added and if it contains a value). By doing the check of the nil value first it makes sure the zero value is considered in the inner if statement.
I.e., statements like `if [[ {{.ReachabilityTotalTimeoutSeconds}} ]]; then` or `if [[ "{{.ReachabilityTotalTimeoutSeconds}}" ]]; then` do not work either (the test fails).

Makes sense?

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: Ok, I'm opening a bug ticket and do a follow up PR.

pkg/network/ovn_kubernetes_test.go
raphaelvrosa: Makes sense. But that is not in the scope of the bug.
Now it is removed from the config map and the tests are changed accordingly.

arghosh93 (approved):
LGTM

bindata/network/ovn-kubernetes/common/008-script-lib.yaml
arghosh93: On a second thought, it seems that the conditional check can be further simplified. Can you tell me what can be the problem if we keep it like below?
```
  ovn_eip_reachability_timeout_opt=
  {{- if .ReachabilityTotalTimeoutSeconds }}
  ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
  {{- end }}
```

raphaelvrosa: Indeed. Done.

bindata/network/ovn-kubernetes/self-hosted/ovnkube-control-plane.yaml
raphaelvrosa: Done

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
raphaelvrosa: Done

arghosh93 (approved):

bindata/network/ovn-kubernetes/managed/ovnkube-control-plane.yaml
```diff
@@ -156,6 +156,11 @@ spec:
             fi
           done
 
+          ovn_eip_reachability_timeout_opt=
+          {{- if .ReachabilityTotalTimeoutSeconds }}                 
+            ovn_eip_reachability_timeout_opt="--egressip-reachability-total-timeout {{.ReachabilityTotalTimeoutSeconds}}"
```
kyrtapz, line 161: Huh, this actually fixes a bug of egressip-reachability-total-timeout not being set at all for hypershift...

kyrtapz (approved):
/lgtm
While the node side controlled by `008-script-lib.yaml` doesn't require the flag I think it is good to have it there for consistency.
This needs a manual verification probably to ensure it works as expected in a live cluster.

raphaelvrosa:
verified by execution of ocp cluster with cluster-bot, checking  `--egressip-reachability-total-timeout` parsing of values in initialization of ovnkube in ovnkube-node and ovnkube-control-plane when operator config is updated.

arghosh93:
/verified by @arghosh93 
```
:~/home/$ oc get network.operator -oyaml|grep -i timeout
          reachabilityTotalTimeoutSeconds: 10

:~/home/$ oc logs ovnkube-control-plane-7d6499867c-zgb74 -n openshift-ovn-kubernetes -c ovnkube-cluster-manager|grep timeout
+ ovn_eip_reachability_timeout_opt=
+ ovn_eip_reachability_timeout_opt='--egressip-reachability-total-timeout 0'
+ exec /usr/bin/ovnkube --enable-multicast --init-cluster-manager ip-10-0-8-119.us-west-2.compute.internal --config-file=/run/ovnkube-config/ovnkube.conf --loglevel 4 --metrics-bind-address 127.0.0.1:29108 --metrics-enable-pprof --metrics-enable-config-duration --gateway-v4-masquerade-subnet 169.254.0.0/17 --gateway-v6-masquerade-subnet fd69::/112 --enable-persistent-ips --gateway-mode shared --egressip-reachability-total-timeout 10

:~/home/$ oc logs ovnkube-node-rwfth -c ovnkube-controller -n openshift-ovn-kubernetes|grep timeout
+ ovn_eip_reachability_timeout_opt=
+ ovn_eip_reachability_timeout_opt='--egressip-reachability-total-timeout 10'
+ exec /usr/bin/ovnkube --init-ovnkube-controller ip-10-0-98-4.us-west-2.compute.internal --init-node ip-10-0-98-4.us-west-2.compute.internal --config-file=/run/ovnkube-config/ovnkube.conf --ovn-empty-lb-events --loglevel 4 --inactivity-probe=180000 --gateway-mode shared --gateway-interface br-ex --metrics-bind-address 127.0.0.1:29103 --ovn-metrics-bind-address 127.0.0.1:29105 --metrics-enable-pprof --metrics-enable-config-duration --export-ovs-metrics --disable-snat-multiple-gws --zone ip-10-0-98-4.us-west-2.compute.internal --enable-multicast --acl-logging-rate-limit 20 --disable-forwarding --bootstrap-kubeconfig=/var/lib/kubelet/kubeconfig --cert-dir=/etc/ovn/ovnkube-node-certs --cert-duration=24h --gateway-v4-masquerade-subnet 169.254.0.0/17 --gateway-v6-masquerade-subnet fd69::/112 --egressip-reachability-total-timeout 10
```

raphaelvrosa:
@kyrtapz I think you need to override all of them in a single comment?
```
/override ci/prow/e2e-metal-ipi-ovn-ipv6-ipsec
```

kyrtapz:
azure job is doing badly, the failure is unrelated(rbac) test failing https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/2955/pull-ci-openshift-cluster-network-operator-master-e2e-azure-ovn-upgrade/2075510176793759744

PR #2960, opened 2026-04-14
OCPBUGS-83406 OCPBUGS-83407:  fix issue for managed mode of no-overlay

arghosh93:
there are two issues for no-overlay managed routing mode:

With -p 0, the BGP daemon does not listen for incoming connections, which breaks the full-mesh iBGP topology used in managed routing mode. Each node must be able to accept BGP sessions from all other nodes, requiring the standard port 179 to be active.

When no-overlay transport is configured with managed routing (NoOverlayManagedEnabled),
the ovn-kubernetes-control-plane service account runs the managed BGP controller which
needs to create, delete, patch and update RouteAdvertisements CRs. The existing
OVN_ROUTE_ADVERTISEMENTS_ENABLE block only grants list/get/watch, causing the cluster
manager to fail on startup with a forbidden error.

cherry-picked from https://github.com/openshift/cluster-network-operator/pull/2954

## Summary by CodeRabbit

* **New Features**
  * BGP daemon will accept incoming BGP sessions on port 179 when managed overlay routing is enabled (otherwise listens on the default isolated port).
  * Control-plane permissions now optionally allow creating/updating route advertisement resources when managed overlay routing is enabled.

### DIFF
--- a/bindata/network/frr-k8s/config.yaml
+++ b/bindata/network/frr-k8s/config.yaml
@@ -46,7 +46,7 @@ data:
     #
     vtysh_enable=yes
     zebra_options="  -A 127.0.0.1 -s 90000000 --limit-fds 100000"
-    bgpd_options="   -A 127.0.0.1 -p 0 --limit-fds 100000"
+    bgpd_options="   -A 127.0.0.1 -p {{ if .NoOverlayManagedEnabled }}179{{ else }}0{{ end }} --limit-fds 100000"
     ospfd_options="  -A 127.0.0.1"
     ospf6d_options=" -A ::1"
     ripd_options="   -A 127.0.0.1"
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -861,6 +861,8 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
 			data.Data["FRRK8sImage"] = os.Getenv("FRR_K8S_IMAGE")
 			data.Data["KubeRBACProxyImage"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
 			data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+			data.Data["NoOverlayManagedEnabled"] = conf.DefaultNetwork.OVNKubernetesConfig != nil &&
+				conf.DefaultNetwork.OVNKubernetesConfig.BGPManagedConfig.BGPTopology != ""
 			objs, err := render.RenderDir(filepath.Join(manifestDir, "network/frr-k8s"), &data)
 			if err != nil {
 				return nil, fmt.Errorf("failed to render frr-k8s manifests: %w", err)
--- a/pkg/network/render.go
+++ b/pkg/network/render.go
@@ -860,6 +860,8 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
 			data := render.MakeRenderData()
 			data.Data["FRRK8sImage"] = os.Getenv("FRR_K8S_IMAGE")
 			data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+			data.Data["NoOverlayManagedEnabled"] = conf.DefaultNetwork.OVNKubernetesConfig != nil &&
+				conf.DefaultNetwork.OVNKubernetesConfig.BGPManagedConfig.BGPTopology != ""
 			objs, err := render.RenderDir(filepath.Join(manifestDir, "network/frr-k8s"), &data)
 			if err != nil {
 				return nil, fmt.Errorf("failed to render frr-k8s manifests: %w", err)
--- a/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
+++ b/bindata/network/ovn-kubernetes/common/004-rbac-control-plane.yaml
@@ -243,7 +243,7 @@ rules:
   verbs:
   - patch
   - update
-- apiGroups: 
+- apiGroups:
   - frrk8s.metallb.io
   resources:
   - frrconfigurations
@@ -255,6 +255,17 @@ rules:
   - delete
   - update
   - patch
+{{- end}}
+{{- if .NoOverlayManagedEnabled }}
+- apiGroups:
+  - k8s.ovn.org
+  resources:
+  - routeadvertisements
+  verbs:
+  - create
+  - delete
+  - patch
+  - update
 {{- end}}
 {{- if .OVN_EVPN_ENABLE }}
 - apiGroups:


arghosh93:
/retitle OCPBUGS-83406 OCPBUGS-83407:  fix issue for managed mode of no-overlay

bindata/network/frr-k8s/config.yaml
```diff
@@ -46,7 +46,11 @@ data:
     #
     vtysh_enable=yes
     zebra_options="  -A 127.0.0.1 -s 90000000 --limit-fds 100000"
-    bgpd_options="   -A 127.0.0.1 -p 0 --limit-fds 100000"
+    # With -p 0, the BGP daemon does not listen for incoming connections,
+    # which breaks the full-mesh iBGP topology used in managed routing mode.
+    # Each node must be able to accept BGP sessions from all other nodes,
+    # requiring the standard port 179 to be active.
+    bgpd_options="   -A 127.0.0.1 -p {{ if .NoOverlayManagedEnabled }}179{{ else }}0{{ end }} --limit-fds 100000"
```
danwinship, line 49: This comment is written as an explanation of *why we had to change this*, but someone reading the file in the future isn't going to know about the previous state. Also, in this case, the comment makes sense in the template file in the source tree, but it won't make sense in the actual generated object in an OCP cluster (after the template has been resolved). So I would just remove the comment.

danwinship, line 49: Listening on port 179 is the default, so this can just be

```suggestion
    bgpd_options="   -A 127.0.0.1 {{ if not .NoOverlayManagedEnabled }}-p 0{{ end }} --limit-fds 100000"
```

However, is this really correct? Won't dropping the `-p 0` bring back whatever bug #2708 was trying to fix? (Presumably, it means we'll also accept BGP messages from hosts _other than_ other OCP nodes, which we don't want?)

pkg/network/render.go
```diff
@@ -861,6 +861,8 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, manifestDir s
 			data.Data["FRRK8sImage"] = os.Getenv("FRR_K8S_IMAGE")
 			data.Data["KubeRBACProxyImage"] = os.Getenv("KUBE_RBAC_PROXY_IMAGE")
 			data.Data["ReleaseVersion"] = os.Getenv("RELEASE_VERSION")
+			data.Data["NoOverlayManagedEnabled"] = conf.DefaultNetwork.OVNKubernetesConfig != nil &&
```
danwinship, line 864: I know we call it "no overlay mode" but is there some other name that could be used here? Having the variable name start with "No" makes it seem like it's talking about the absence of a feature (i.e., `NoOverlayManagedEnabled: true` means that we don't want `OverlayManagedEnabled`)

maybe just `EnableNoOverlayManaged` would work? (Or just `EnableNoOverlay`? I'm not sure what the "managed" refers to here.)

asood-rh:
/verified by @asood-rh 

The CI workflow where cluster installed successfully with routing: managed. Will add post install step to verify success in https://github.com/openshift/release/pull/77768

bindata/network/frr-k8s/config.yaml
arghosh93: As per https://github.com/openshift/cluster-network-operator/pull/2708 PR description:
```
-p, --bgp_port <port>
Set the BGP protocol’s port number. If set to 0, bgpd will not listen on any BGP port
```
However, we want to listen on port 179 to establish full mesh. 
cc: @fedepaol

arghosh93: yaa, makes sense. Let me remove it from here and add to the commit message so that any developer revisiting this part of the code understands why it was done.

pkg/network/render.go
arghosh93: we have been calling this as NoOverlay from the very beginning. This naming pattern spans over both CNO and OvnKubernetes. 
I can do a follow-up PR, maybe to change the naming convention.

bindata/network/frr-k8s/config.yaml
danwinship: With managed no-overlay, there are *some* BGP messages that we want to accept, but that doesn't meant that we want to accept *all* BGP messages from *all* sources. The fact that we previously disabled this implies that either (a) in some clusters there are "stray" BGP messages that we want to ignore, or (b) there are security concerns with accepting all BGP messages. So, we may need some sort of firewall rule or bgpd configuration option in addition to enabling port 179.

pkg/network/render.go
danwinship: Yes, I wasn't saying we should change the name of the feature, I meant, can we call the _template variable_ something other than `NoOverlayManagedEnabled`

bindata/network/frr-k8s/config.yaml
arghosh93: @danwinship [initially](https://github.com/openshift/cluster-network-operator/commit/37c651821e7f97ec82090844e83f81d81aef96d7), we were listening on all ports. Then, some issue arose due to listening on all ports and we blocked traffic by mentioning `-p 0`. But to establish full mesh in NoOverlay managed mode, we would have to listen on port 179 for BGP peer communication. 
@jcaamano Are you aware of any security concerns due to listening on port 179 when NoOverlay mode is enabled?

pkg/network/render.go
arghosh93: We have been following this naming convention from the beginning. For example `NoOverlayManagedASNumber`  and `NoOverlayManagedTopology` are already in use. But I do agree to your point that calling it `underlay` could have been better instead of `NoOverlay`.
Can I address this in a separate PR if you are okay with it?

bindata/network/frr-k8s/config.yaml
fedepaol: @danwinship it's not about messages, it's about who initiates the bgp session.
BGP is bidirectional and runs over tcp, which means the first session that is able to connect wins, the other side stops trying to connect and all the messages flow over that single tcp session in one way or the other.
When we implemented metallb we replicated the original behavior which was to always connect from the cluster to the external router (which made sense as a security precaution I guess). The behavior was then mutuated by frrk8s, but of course can't stay the same if we need full mesh among frrk8s instances, where each one will also have to accept incoming connections.

pkg/network/render.go
danwinship: ah, I didn't notice there were already other `NoOverlay` variables. Yeah, can be a follow-up

bindata/network/frr-k8s/config.yaml
jcaamano: @fedepaol @karampok I think there might have been an issue, maybe related to graceful restart: https://redhat.atlassian.net/browse/OCPBUGS-48521?focusedCommentId=3095119
https://redhat.atlassian.net/browse/OCPBUGS-48521?focusedCommentId=3095121
Is this something we need to be concerned about?

Other than that...

What @danwinship suggested [here](https://github.com/openshift/cluster-network-operator/pull/2960/changes#r3092777145) is good right? There is no need to explicitly set `-p 179` as that's default?

I think we should register port 179 here: https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md

In terms of hardening, I guess the API&docs should describe that this opens up port 179 to accept incoming BGP connections from other cluster nodes and leave it up to the customer to install whatever firewall rules it deems necessary? I am not aware of other instances where we apply hardening on user behalf (might completely be missing it).
Also maybe document the potential disruption when redeploying frr-k8s equivalent to upgrades, if any?

danwinship: > I think we should register port 179 here: https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md

Yes.

We probably also need to deal with the "commatrix" stuff (the thing that makes sure that all listening ports are properly declared as Services). (eg, see [OCPBUGS-077019](https://redhat.atlassian.net/browse/OCPBUGS-77019))

> In terms of hardening, I guess the API&docs should describe that this opens up port 179 to accept incoming BGP connections from other cluster nodes

It's not just nodes. It's any host that can reach nodes. In particular, it might mean we accept BGP connections from our upstream router?

> and leave it up to the customer to install whatever firewall rules it deems necessary? I am not aware of other instances where we apply hardening on user behalf (might completely be missing it).

We install a lot of firewall rules in IPI clusters (in the form of AWS/GCP/Azure/etc security policies). For UPI clusters, we document what ports need to be left open, and who they need to be left open to (https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/installation_configuration/configuring-firewall).

jcaamano: > It's not just nodes. It's any host that can reach nodes. In particular, it might mean we accept BGP connections from our upstream router?

Yeah. what I meant to say is document the purpose so that a user can apply correct policies: i.e. block any incoming connection that does not come from other cluster node/subnet.

> We install a lot of firewall rules in IPI clusters (in the form of AWS/GCP/Azure/etc security policies). 

Is that done from installers? Do we own those policies? Anything for baremetal?

> For UPI clusters, we document what ports need to be left open, and who they need to be left open to (https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/installation_configuration/configuring-firewall).

Good place to doc!

karampok: The GR (Gracefull Restart) was dropped by openshift so we should be okay. Nevertheless if in the future there is a another property that is set only during the BGP session, then that might be have the same bug (= the external side setups the session before the ocp side writed the frr config use this property)

My preference would be to have `-p 179` even if is default, just to be clear.

arghosh93:
/verified by @asood-rh
Code [comment](https://github.com/openshift/cluster-network-operator/compare/c82594dbe5bbedf213c70f7becfe3b879d8c4f81..b98074ea3a2465d4d06d9a5a61b0a65f1a2f2560) was only modified after Arti added the verified label. Readding the label.

PR #2968, opened 2026-04-22
OCPBUGS-81741: Watch Network and Infrastructure in proxyconfig controller

jluhrsen:
The proxyconfig controller reads Network.Status.ClusterNetwork and Infrastructure.Status to compute Proxy.Status.NoProxy, but only watched Proxy and ConfigMaps. Network or Infrastructure changes would not trigger reconciliation, leaving proxy status stale.

Add watches for Network and Infrastructure resources to ensure reconciliation occurs when these resources change.

Also add Proxy status subresource support to fake client and unit tests covering reconciliation logic.

Co-authored-by: Claude Code <noreply@anthropic.com)

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for proxy reconciliation, validating status updates when network or infrastructure change, empty-spec behavior, and handling missing resources.

* **Improvements**
  * Controller now watches Network and Infrastructure and more predictably handles cluster-scoped events to keep proxy status in sync.

* **Chores**
  * Test harness and fake client extended to include proxy status subresource for more realistic test scenarios.


jluhrsen:
@danwinship @tssurya @kyrtapz @pliurh looking for someone to check on this. it's been sitting idle on my open bug list for a while now. haven't really followed up to find a reviewer yet.

I've verified the fix manually and passed this through coderabbit review (locally). should be good for a human to look now.

pkg/controller/proxyconfig/controller.go
```diff
@@ -79,6 +79,18 @@ func add(mgr manager.Manager, r *ReconcileProxyConfig) error {
 		return err
 	}
 
+	// Watch for changes to the network resource.
+	err = c.Watch(source.Kind[crclient.Object](mgr.GetCache(), &configv1.Network{}, &handler.EnqueueRequestForObject{}))
```
danwinship, line 83: This should require a change to `Reconcile()` too shouldn't it? (It checks what kind of object it's reconciling.)

Assuming that's correct, that means this patch doesn't actually work at all, which makes me feel like it probably ought to have had a corresponding e2e test?

danwinship (commented):
(oops, I reviewed this the other day but I guess I forgot to hit "submit")

pkg/controller/proxyconfig/controller.go
jluhrsen: Thanks @danwinship . I tested this fix with a live cluster so I am pretty confident in the change, but obviously  don't know this code base like you do.

But, I think it works because [Reconcile() is interested in the "Name"](https://github.com/openshift/cluster-network-operator/pull/2968/changes/74e43ba69af6680984e7be14006e9590dd0ffdc0#diff-60364f77d3208108d4e05739cdf035439a123ec619f06651202dae56f059dcaaL104) which I guess always will be "[cluster](https://github.com/openshift/cluster-network-operator/blob/6dc18040e7c214f6a1db25b6f5ef4642c6c6a186/pkg/names/names.go#L13-L19)" for these.

I think an e2e might be complicated, but if you need one figured out to validate this even further, I try to figure it out.

danwinship: huh... you should fix `Reconcile` so that the `switch` does something that actually looks like it should work, rather than than something that just coincidentally happens to work. Possibly just split out the "configmap" and "not a configmap" cases.

jluhrsen: ok, I think I see what you mean. I gave it another try to hopefully make it a little more clear

jluhrsen: @danwinship, WDYT?

pkg/controller/proxyconfig/controller.go
```diff
@@ -92,24 +104,28 @@ type ReconcileProxyConfig struct {
...
 	defer utilruntime.HandleCrash(r.status.SetDegradedOnPanicAndCrash)
 	validate := true
 	trustBundle := &corev1.ConfigMap{}
 
-	switch {
-	case request.NamespacedName == names.Proxy():
+	switch request.Namespace {
+	case "":
+		if request.Name != names.CLUSTER_CONFIG {
+			log.Printf("Ignoring unknown cluster-scoped object, reconciliation will be skipped: request=%v", request)
```
danwinship: Yeah, so what I meant before was, change `Reconcile()` so that basically it checks "does `request` point to a `ConfigMap` in `ADDL_TRUST_BUNDLE_CONFIGMAP_NS`", and if it does, then it does that case, and if not, then it doesn't look at `request` again, it just assumes it's supposed to re-resolve the proxy config from the top down.

Maybe split those out into two functions, `reconcileTrustBundle(request)` and `reconcileProxyConfig()`.

pkg/controller/proxyconfig/controller.go
```diff
@@ -92,24 +104,28 @@ type ReconcileProxyConfig struct {
...
 		var err error
 		proxyConfig := &configv1.Proxy{}
 		infraConfig := &configv1.Infrastructure{}
 		netConfig := &configv1.Network{}
 		clusterConfig := &corev1.ConfigMap{}
 
-		log.Printf("Reconciling proxy '%s'", request.Name)
-		if err := r.client.Get(ctx, request.NamespacedName, proxyConfig); err != nil {
+		log.Printf("Reconciling proxy '%s'", names.PROXY_CONFIG)
+		if err := r.client.Get(ctx, names.Proxy(), proxyConfig); err != nil {
```
danwinship, line 127: (This change and the one below are correct.)

danwinship:
So it's hard to point out in the diff, but it seems that a bunch of code got deleted in the rewrite.

Before, the trust bundle case was

```
		log.Printf("Reconciling additional trust bundle configmap '%s/%s'", request.Namespace, request.Name)
		// Only proceed if request matches the configmap referenced by proxy trustedCA.
		// Validate the trust bundle configmap.
		// Create a configmap containing the merged proxy.trustedCA/system bundles.
		// A nil proxy is generated by upgrades and installs not requiring a proxy.
			// Create a configmap containing the system trust bundle.
			// Validate trustedCA of proxy spec.
		log.Printf("Reconciling additional trust bundle configmap '%s/%s' complete", request.Namespace, request.Name)
```

now it's

```
        log.Printf("Reconciling additional trust bundle configmap '%s/%s'", request.Namespace, request.Name)
        // Only proceed if request matches the configmap referenced by proxy trustedCA.
        // Validate the trust bundle configmap.
        // Create a configmap containing the merged proxy.trustedCA/system bundles.
        log.Printf("Reconciling additional trust bundle configmap '%s/%s' complete", request.Namespace, request.Name)
```

That is, previously, half of the "Reconciling proxy" case was duplicated into the "Reconciling additional trust bundle" case, and now it isn't any longer. (The duplication of the code before was bad and ought to have been done in a better way, but we need to make sure we're doing all of the necessary steps. Or were those steps unnecessary?)

(Sorry, I'm not really familiar with this code, so I keep discovering more things as the diff touches more parts of the file...)

jluhrsen:
> So it's hard to point out in the diff, but it seems that a bunch of code got deleted in the rewrite.
> 
> Before, the trust bundle case was
> 
> ```
> 		log.Printf("Reconciling additional trust bundle configmap '%s/%s'", request.Namespace, request.Name)
> 		// Only proceed if request matches the configmap referenced by proxy trustedCA.
> 		// Validate the trust bundle configmap.
> 		// Create a configmap containing the merged proxy.trustedCA/system bundles.
> 		// A nil proxy is generated by upgrades and installs not requiring a proxy.
> 			// Create a configmap containing the system trust bundle.
> 			// Validate trustedCA of proxy spec.
> 		log.Printf("Reconciling additional trust bundle configmap '%s/%s' complete", request.Namespace, request.Name)
> ```
> 
> now it's
> 
> ```
>         log.Printf("Reconciling additional trust bundle configmap '%s/%s'", request.Namespace, request.Name)
>         // Only proceed if request matches the configmap referenced by proxy trustedCA.
>         // Validate the trust bundle configmap.
>         // Create a configmap containing the merged proxy.trustedCA/system bundles.
>         log.Printf("Reconciling additional trust bundle configmap '%s/%s' complete", request.Namespace, request.Name)
> ```
> 
> That is, previously, half of the "Reconciling proxy" case was duplicated into the "Reconciling additional trust bundle" case, and now it isn't any longer. (The duplication of the code before was bad and ought to have been done in a better way, but we need to make sure we're doing all of the necessary steps. Or were those steps unnecessary?)
> 
> (Sorry, I'm not really familiar with this code, so I keep discovering more things as the diff touches more parts of the file...)

@danwinship , you are many times more familiar with this code base than I am, so no worries there.

I think the short answer is, yes those duplication steps were unnecessary, but maybe it wasn't perfectly de-duped 🤷🏻 

the problem is that the original fix here was small and worked, but I get the confusion. so now we are in more of a refactor mode on top of that fix, and claude is recognizing that duplication and refactoring as it sees fit. I've spun the changes through codex and coderabbit double and triple checking that they believe functionality is not affected here. and yes, apparently there was unneeded duplication that got removed.

However, I've posted a new try. keeping the duplication so the review is easier to look at. But, it's not breaking things out in to your idea of two functions, `reconcileTrustBundle(request)` and `reconcileProxyConfig()`. When I try that we end up with the ugly diff.  Maybe can think about a de-dup and split out refactor later if that seems important to you.

Does it, at least, make what's happening in Reconcile() a little more clear?

jluhrsen:
also, @danwinship , I created [this](https://github.com/openshift/cluster-network-operator/pull/3033) as maybe a better idea to do this kind of refactoring you are noticing we need. It tries to make smaller refactor commits for easier review. please don't spend time on it yet, but let me know if you like this idea better and I'll take it further. I need to review it myself, let the e2e's run and who knows what the rabbit is gonna say about it.

pkg/controller/proxyconfig/controller.go
```diff
@@ -92,40 +104,40 @@ type ReconcileProxyConfig struct {
...
+// "openshift-config" or another watched resource indicating that the cluster
+// Proxy should be reconciled from current state.
 func (r *ReconcileProxyConfig) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
 	defer utilruntime.HandleCrash(r.status.SetDegradedOnPanicAndCrash)
 	validate := true
 	trustBundle := &corev1.ConfigMap{}
 
 	switch {
-	case request.NamespacedName == names.Proxy():
+	case request.Namespace != names.ADDL_TRUST_BUNDLE_CONFIGMAP_NS:
```
danwinship, line 116:
```suggestion
	case request.Namespace != names.ADDL_TRUST_BUNDLE_CONFIGMAP_NS:
		// Additional trust bundles are handled below; for all other config changes,
		// regardless of which object changed, we just re-resolve everything starting
		// from the Proxy config.
```

danwinship:
From what I was seeing when I looked at it before, it seemed like it might all boil down to

```
if the reconciliation is caused by a trust bundle change {
    update trust bundle stuff
}
re-resolve the proxy config from the top down
```

but anyway, the current simple fix (with the added explanatory comment) is good enough I guess

pkg/controller/proxyconfig/controller.go
jluhrsen: done

jluhrsen:
> From what I was seeing when I looked at it before, it seemed like it might all boil down to
> 
> ```
> if the reconciliation is caused by a trust bundle change {
>     update trust bundle stuff
> }
> re-resolve the proxy config from the top down
> ```
> 
> but anyway, the current simple fix (with the added explanatory comment) is good enough I guess

ok, thank you. if you think it's worth following up with the [full refactor idea](https://github.com/openshift/cluster-network-operator/pull/3033), let me know and I'll chase that down after this PR lands.

danwinship:
/lgtm

I don't think the full refactor is worth it, given that neither of us really knows the code well enough to know exactly how it should be refactored :grimacing:

jluhrsen:
@danwinship can you `/override ci/prow/e2e-aws-ovn-upgrade-ipsec` which is [permafailing](https://sippy.dptools.openshift.org/sippy-ng/jobs/5.0/analysis?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22name%22%2C%22operatorValue%22%3A%22equals%22%2C%22value%22%3A%22periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-upgrade-ipsec%22%7D%5D%7D) because of [OCPBUGS-89238](https://redhat.atlassian.net/browse/OCPBUGS-89238)

danwinship:
/override ci/prow/e2e-metal-ipi-ovn-dualstack-bgp-local-gw ci/prow/e2e-ovn-ipsec-step-registry ci/prow/e2e-metal-ipi-ovn-ipv6-ipsec

danwinship:
/override ci/prow/e2e-metal-ipi-ovn-dualstack-bgp-local-gw ci/prow/e2e-ovn-ipsec-step-registry ci/prow/e2e-metal-ipi-ovn-ipv6-ipsec ci/prow/e2e-aws-ovn-upgrade-ipsec

PR #3011, opened 2026-05-18
CORENET-7154: Fix debounce timer for OperatorConfig level being incorrectly cleared

tpantelis:
The debounce timer for the `OperatorConfig` status level was being incorrectly cleared when the operator config exists. The code was using the `OperatorConfig` status level internally to track the debounce timer for when the operator config is missing. However it also blindly cleared the timer when the operator config exists, which is the normal case, potentially interfering with a caller's usage of the `OperatorConfig` status level.

This commit removes the debouncing for this case as the operconfig should never actually be missing so it was likely some other error that occurred that facilitated adding the debounce. Since it is unknown what caused the original scenario, we shouldn't assume that delaying the degraded status is the correct solution.

## Summary by CodeRabbit

* **Bug Fixes**
  * Cluster operator is now marked degraded immediately when required operator configuration is missing, ensuring prompt and consistent degraded-state reporting.

* **Tests**
  * Tests updated to verify the immediate degraded behavior and to ensure debounce/timer state does not delay or hide missing-configuration reporting.

### DIFF
--- a/pkg/controller/statusmanager/status_manager.go
+++ b/pkg/controller/statusmanager/status_manager.go
@@ -473,13 +473,6 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 		}
 
 		if operStatus == nil {
-			if _, exists := status.failureFirstSeen[OperatorConfig]; !exists {
-				status.failureFirstSeen[OperatorConfig] = status.clock.Now()
-				return nil
-			}
-			if status.clock.Since(status.failureFirstSeen[OperatorConfig]) < degradedFailureDurationThreshold {
-				return nil
-			}
 			cohelpers.SetStatusCondition(&co.Status.Conditions, configv1.ClusterOperatorStatusCondition{
 				Type:    configv1.OperatorDegraded,
 				Status:  configv1.ConditionTrue,
@@ -487,7 +480,6 @@ func (status *StatusManager) set(reachedAvailableLevel bool, conditions ...operv
 				Message: "Failed to get networks.operator.openshift.io cluster",
 			}, clock.RealClock{})
 		} else {
-			delete(status.failureFirstSeen, OperatorConfig)
 			if reachedAvailableLevel {
 				co.Status.Versions = []configv1.OperandVersion{
 					{Name: "operator", Version: operStatus.Version},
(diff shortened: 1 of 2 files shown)


danwinship:
I don't understand the description. What's the exact scenario and what's the incorrect behavior in that scenario? What is the user-visible impact of the bug?

tpantelis:
> I don't understand the description. What's the exact scenario and what's the incorrect behavior in that scenario? What is the user-visible impact of the bug?

I outlined the scenario in https://redhat.atlassian.net/browse/CORENET-7154.

pkg/controller/statusmanager/status_manager.go
```diff
@@ -57,6 +57,7 @@ const (
 	CertificateSigner
 	InfrastructureConfig
 	DashboardConfig
+	operConfigMissing
```
danwinship: OK, so the problem with this is that operconfig should never actually be missing, and if it was, we'd be fine with immediately going degraded. According to the discussion in https://github.com/openshift/cluster-network-operator/pull/2896#discussion_r2765762732, it seemed like some _other_ error was occurring, which the code was mistakenly classifying as "operator config doesn't exist", and this sporadically happened in some real job, but in the example @jluhrsen linked to, the pod logs have been deleted now so there's no further information.

Anyway, I think we should remove the debouncing for this case, reproduce the failure that led to Jamo adding this check, and then figure out the _right_ fix for it (because it may actually indicate a real problem with the cluster which we are just masking by delaying the degraded status).

tpantelis: Sounds good.

danwinship: (you can remove `operConfigMissing` now too)

danwinship:
/assign @jluhrsen 

we can test this to try to reproduce the original failure here by doing payload jobs, can't we?

pkg/controller/statusmanager/status_manager.go
tpantelis: Done

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-ci-4.22-upgrade-from-stable-4.21-e2e-aws-ovn-upgrade 10

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-ci-4.22-upgrade-from-stable-4.21-e2e-aws-ovn-upgrade 10

jluhrsen:
/hold
wanna wait to see the next set of 10 upgrade jobs to see if our Progressing=True bug shows up

jluhrsen:
/verified by 20 4.22-upgrade-from-stable-4.21-e2e-aws-ovn-upgrade jobs not finding the network operator ever going Progressing=True

/hold cancel

PR #3014, opened 2026-05-26
CORENET-7114: Add OCP centralized TLS profile support

tpantelis:
See commits for details.

## Summary by CodeRabbit

* **New Features**
  * Reconcile triggers for API server TLS/security profile changes, including HyperShift-hosted clusters
  * Support passing a custom manifest path for operator deployment

* **Improvements**
  * MTU probe polls more frequently for faster detection
  * Bootstrap-populated TLS profile is used to populate rendered manifests (min version & ciphers)

* **Tests**
  * Extensive new test coverage for rendering, HyperShift, MTU probing, reconciliation and failure scenarios

* **Chores**
  * Linter and dependency configuration updates

### DIFF
--- a/pkg/client/fake/fake_client.go
+++ b/pkg/client/fake/fake_client.go
@@ -107,9 +107,20 @@ func NewFakeClient(objs ...crclient.Object) cnoclient.Client {
 		crclient:     crfake.NewClientBuilder().WithStatusSubresource(co, proxy).WithObjects(objs...).Build(),
 		osOperClient: osoperfakeclient.NewClientset(),
 	}
+
+	// Create a management cluster client for HyperShift scenarios
+	// This represents a separate cluster, so it starts empty
+	managementClient := FakeClusterClient{
+		kClient:      faketyped.NewClientset(),
+		dynclient:    fakedynamic.NewSimpleDynamicClient(scheme.Scheme),
+		crclient:     crfake.NewClientBuilder().WithStatusSubresource(co).Build(),
+		osOperClient: osoperfakeclient.NewClientset(),
+	}
+
 	return &FakeClient{
 		clusterClients: map[string]*FakeClusterClient{
-			names.DefaultClusterName: &fc,
+			names.DefaultClusterName:    &fc,
+			names.ManagementClusterName: &managementClient,
 		},
 	}
 }
--- a/vendor/sigs.k8s.io/controller-runtime/.gitignore
+++ b/vendor/sigs.k8s.io/controller-runtime/.gitignore
@@ -1,0 +1,30 @@
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, build with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# editor and IDE paraphernalia
+.idea
+*.swp
+*.swo
+*~
+
+# Vscode files
+.vscode
+
+# Tools binaries.
+hack/tools/bin
+
+# Release artifacts
+tools/setup-envtest/out
+
+junit-report.xml
+/artifacts
--- a/cmd/cluster-network-operator/main.go
+++ b/cmd/cluster-network-operator/main.go
@@ -6,18 +6,30 @@ import (
 	"fmt"
 	"os"
 
+	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
+	"github.com/openshift/cluster-network-operator/pkg/hypershift"
 	"github.com/openshift/cluster-network-operator/pkg/names"
+	"github.com/openshift/cluster-network-operator/pkg/network"
 	"github.com/openshift/cluster-network-operator/pkg/operator"
 	"github.com/openshift/cluster-network-operator/pkg/version"
+	libgoclient "github.com/openshift/library-go/pkg/config/client"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/crypto"
+	"github.com/openshift/library-go/pkg/operator/events"
+	"github.com/openshift/library-go/pkg/serviceability"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"k8s.io/apiserver/pkg/server"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
-
+	"k8s.io/client-go/rest"
 	utilflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/logs"
+	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 )
+
+const componentName = "network-operator"
 
 func main() {
 	pflag.CommandLine.SetNormalizeFunc(utilflag.WordSepNormalizeFunc)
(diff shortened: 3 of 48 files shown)


pkg/client/fake/fake_client.go
```diff
@@ -70,16 +82,93 @@ func (fc *FakeClient) Clients() map[string]cnoclient.ClusterClient {
 	return out
 }
 
-func isOpenShiftObject(obj crclient.Object) bool {
+// isKubernetesObject returns true if the object is a core Kubernetes type
+// that can be handled by the fake Kubernetes typed client.
+func isKubernetesObject(obj crclient.Object) bool {
```
danwinship: Hm... so the prior code was just completely wrong? It was adding all of the `openshift.io` objects to the typed k8s client, which was then silently ignoring them, and the tests only worked because it was also adding everything to the dynamic client, which is what actually got used?

pkg/client/fake/fake_client.go
```diff
@@ -70,16 +82,93 @@ func (fc *FakeClient) Clients() map[string]cnoclient.ClusterClient {
...
 	for _, v := range kKind {
-		if strings.HasSuffix(v.Group, "openshift.io") {
+		// Include core Kubernetes API groups
+		if v.Group == "" { // core group (pods, services, etc)
+			return true
+		}
+		// Include k8s.io groups, but exclude extension APIs that aren't part of the typed client
+		if strings.HasSuffix(v.Group, ".k8s.io") &&
+			v.Group != "apiextensions.k8s.io" && // CRDs are not in typed client
+			!strings.HasPrefix(v.Group, "sigs.k8s.io") { // SIG projects use dynamic client
```
danwinship: This seems fragile. There is presumably a way to determine exactly what types are part of the typed client...

Alternatively, since it seems that `faketyped.NewClientset` just silently ignores everything it doesn't own, you could just not bother filtering at all...

pkg/client/fake/fake_client.go
```diff
@@ -70,16 +82,93 @@ func (fc *FakeClient) Clients() map[string]cnoclient.ClusterClient {
...
+		}
+		gvk := gvr.GroupVersion().WithKind(kind)
+		unstructuredObj.SetGroupVersionKind(gvk)
+
+		// Try to get the existing object
+		_, err = dynClient.Tracker().Get(gvr, namespace, name)
+		if err == nil {
+			// Object exists, update it by replacing with the patch data
+			// For Server-Side Apply, we just replace the whole object
+			err = dynClient.Tracker().Update(gvr, unstructuredObj, namespace)
```
danwinship: That definitely can't work in all cases... that assumes that either every user of SSA in the unit tests is managing the entire object, or else that nothing ever depends on any of the fields of an object that aren't being managed by SSA.

pkg/controller/fake/cache.go
```diff
@@ -0,0 +1,160 @@
...
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/watch"
+	toolscache "k8s.io/client-go/tools/cache"
+	crCache "sigs.k8s.io/controller-runtime/pkg/cache"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Cache struct {
```
danwinship: What is this needed for? What are we doing in our unit tests that is not already supported by the existing abstractions, that requires us to write all of this?

pkg/controller/operconfig/controller_applied_test.go
```diff
@@ -0,0 +1,115 @@
+package operconfig_test
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
```
danwinship: We don't normally use ginkgo for unit tests. And given that there are already lots of unit tests that _don't_ use ginkgo, using it here is extra confusing (and makes it hard to understand the test runner output, and hard to run a single test, because you have to use different syntax to run a single ginkgo test vs running a single non-ginkgo test).

(We don't normally use gomega for unit tests either, but that's less objectionable.)

pkg/controller/operconfig/controller_applied_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	When("a previous applied config exists", func() {
+		BeforeEach(func() {
+			t.operConfig.Spec.DefaultNetwork.OVNKubernetesConfig.MTU = nil
+			prevSpec.DefaultNetwork.OVNKubernetesConfig.MTU = ptr.To[uint32](1401)
+		})
+
+		It("should detect and apply changes", func(ctx context.Context) {
+			appliedSpec := awaitAppliedSpec(ctx)
```
danwinship: The flow of control here is extremely non-obvious. How can you wait for the spec to be applied? You didn't tell anything to apply it... The use of `JustBeforeEach` to do that makes the test harder to understand.

pkg/controller/operconfig/mtu_probe.go
```diff
@@ -68,7 +69,7 @@ func (r *ReconcileOperConfig) probeMTU(ctx context.Context, oc *operv1.Network,
 	klog.Info("MTU prober deployed, waiting for result ConfigMap")
 
 	// wait up to 100 seconds for Job to report back.
-	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 100*time.Second, false, func(ctx context.Context) (bool, error) {
+	err = wait.PollUntilContextTimeout(ctx, time.Second, 100*time.Second, false, func(ctx context.Context) (bool, error) {
```
danwinship: why?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -58,25 +58,24 @@ var ManifestPath = "./bindata"
...
-	return add(mgr, rc)
+	return AddWithManifestPath(mgr, status, c, featureGates, ManifestPath)
 }
 
-const ControllerName = "operconfig"
-
-// newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, status *statusmanager.StatusManager, c cnoclient.Client, featureGates featuregates.FeatureGate) (*ReconcileOperConfig, error) {
-	return &ReconcileOperConfig{
+func AddWithManifestPath(mgr manager.Manager, status *statusmanager.StatusManager, c cnoclient.Client,
```
danwinship: document that this is for unit tests

pkg/controller/statusmanager/status_manager.go
```diff
@@ -165,6 +165,12 @@ func New(client cnoclient.Client, name, cluster string) *StatusManager {
 	return status
 }
 
+func NewWithClock(client cnoclient.Client, name, cluster string, clock clock.PassiveClock) *StatusManager {
```
danwinship: document that this is for unit tests

vendor/sigs.k8s.io/controller-runtime/.gitignore
```diff
@@ -0,0 +1,30 @@
+# Binaries for programs and plugins
```
danwinship, line 1: why did all of these files get pulled in?

pkg/controller/operconfig/controller_hypershift_test.go
```diff
@@ -61,16 +62,35 @@ func testHypershiftMode() {
...
 		})
 
-		// Set infrastructure name for HyperShift mode
 		t.infrastructure.Status.InfrastructureName = infraName
 	})
 
 	JustBeforeEach(func(ctx context.Context) {
-		// Create HostedControlPlane object in management cluster
+		// Create the HostedCluster object in the management cluster.
+		hc := &uns.Unstructured{
```
danwinship: why does this use Unstructured rather than the actual API objects?

pkg/controller/operconfig/controller_applied_test.go
tpantelis: yeah I noticed that in this repo. But in other openshift projects, Ginkgo is used. Eg, in **ovn-kubernetes** , there is a mix of Ginkgo and Go testing but the trend for new code seems to be Ginkgo.

Personally, I think Go testing framework is old and klunky. I'm a little surprised using Ginkgo and Gomega would actually be objectionable in this repo but, if that's the case, then I can convert it to Go testing (tho perhaps Gomega could be allowed as a compromise, that would really be helpful). I could also just remove these tests altogether to simplify the PR and review.

pkg/client/fake/fake_client.go
tpantelis: I did this a while ago and trying to remember the details... but essentially yes. I vaguely remember a test was failing and this change addressed it (or at least I thought) but, after reverting the change now, all the tests pass. So it doesn't seem to be needed now - I can revert it.

pkg/client/fake/fake_client.go
tpantelis: See above

pkg/client/fake/fake_client.go
tpantelis: yes, this is a simplified SSA implementation. It works for the current test usage pattern in the oper config controller because it sets `Force: true` in the `PatchOptions` and it sends the full object on every apply. There's no other uses of SSA in current unit tests and we could adjust it if we encounter test scenarios in the future that need proper field-level SSA semantics. Or I could move this specific implementation to the oper config controller unit tests.

pkg/controller/operconfig/mtu_probe.go
tpantelis: To speed it up a bit, especially for the unit tests.  I don't know how long the job typically takes in a real environment but I'd imagine it's typically a few sec at most in which case a 5 sec interval seems like a lot. But if there's a specific reason 5 sec was chosen, I'll revert it and adjust the unit tests.

pkg/controller/operconfig/controller_hypershift_test.go
tpantelis: The production code avoids importing the HyperShift API types, see https://github.com/openshift/cluster-network-operator/blob/6dc18040e7c214f6a1db25b6f5ef4642c6c6a186/pkg/hypershift/hypershift.go#L52, and uses `Unstructured`.

pkg/controller/operconfig/controller_applied_test.go
tpantelis: yeah line 53 triggers reconciliation. I agree that can be hard to follow. I could move the trigger into the `It` blocks for clarity and also add a `triggerReconciliation() `function.

pkg/controller/fake/cache.go
tpantelis: The fake `Cache` implementation bridges the informers with the controller's reconcile queue. The production code calls `controller.Watch` with a `source.Kind` which uses the manager's `Cache` instance to obtain the `Informer` (via `GetInformer()`) and add an event handler. But the real `Cache` implementation watches the API Server directly (via an HTTP connection) and doesn't watch the associated client. So when an object is written to the fake client, the controller's event handler is not invoked. The fake `Cache` implementation's `GetInformer` method returns an informer that intercepts watch requests and returns a fake `watch.Interface`. The tests then inject events into the fake watchers to trigger the controller's event handler.
 
We could do direct reconciler invocation via `Reconcile()` but that wouldn't test  informer setup, watch predicates/filters and verification of "reconciliation should NOT trigger" scenarios.

vendor/sigs.k8s.io/controller-runtime/.gitignore
tpantelis: `sigs.k8s.io/controller-runtime` was bumped to _v0.23.3_ which pulled these files in via `go mod vendor`. Looking at other dependencies in the vendor dir, there are many other such non _.go_ files as well so that's the general behavior.

tpantelis:
I removed the oper config unit tests to simplify the PR and change the other unit tests I added in the `network` package to use Go testing.

vendor/sigs.k8s.io/controller-runtime/.gitignore
danwinship: They were included in the commit before you bumped `go.mod` though. ANyway, it's fixed now

danwinship:
You pushed a bunch of unrelated commits along with your branch...

tpantelis:
@danwinship So it turns out CNO uses `controllercmd.NewControllerCommandConfig`

https://github.com/openshift/cluster-network-operator/blob/6dc18040e7c214f6a1db25b6f5ef4642c6c6a186/cmd/cluster-network-operator/main.go#L49-L51

which automatically sets up an HTTPS metrics server but with hard-coded TLS min version (TLS 1.2) and cipher suites, which, of course, isn't TLS profile compliant. I missed that before. It also does not (directly) provide a way to customize the TLS config, although it seems there is a clunky way to do it. Assuming that's true, we'll also need to watch for TLS profile changes and restart the process (there's no way to hot-reload the metrics server). Given that, the changes I made to the `ReconcileOperConfig` controller to watch the `APIServer` (and `HostedCluster` for hypershift) would basically move to a higher-level and trigger process restart.

Other options are:
- use plain HTTP for operator metrics
- continue to use the hard-coded TLS  defaults and don't honor the centralized TLS profile for operator metrics

but I assume neither of those is acceptable.

danwinship:
> So it turns out CNO uses `controllercmd.NewControllerCommandConfig` ...
which automatically sets up an HTTPS metrics server but with hard-coded TLS min version (TLS 1.2) and cipher suites, which, of course, isn't TLS profile compliant.

I assume someone is going to have to fix that as part of all the 5.0 TLS work anyway though?

> use plain HTTP for operator metrics

That's what we do for most components, and then add `kube-rbac-proxy` to export via HTTPS, so this ought to be easy / similar to other patches?

tpantelis:
> > So it turns out CNO uses `controllercmd.NewControllerCommandConfig` ...
> > which automatically sets up an HTTPS metrics server but with hard-coded TLS min version (TLS 1.2) and cipher suites, which, of course, isn't TLS profile compliant.
> 
> I assume someone is going to have to fix that as part of all the 5.0 TLS work anyway though?

I asked about it on the **#forum-ocp-tls-strict-obedience** channel but no answer (yet). There's no open PR but I suspect if `controllercmd` was going to be updated it would've been done by now. A prior post asking about `controllercmd` had no resolution other than to move anyway from using `controllercmd`. Looking at another openshift project using `controllercmd`, https://github.com/openshift/aws-ebs-csi-driver-operator/blob/master/cmd/aws-ebs-csi-driver-operator/main.go, it's not doing anything to use the centralized TLS profile.

Anyway, there is a way to configure the TLS settings in `controllercmd` although it requires duplicating some code from  `controllercmd`.  Another option is to convert to using controller runtime but that may be more work and we'd probably lose some functionality provided by `controllercmd`.

> 
> > use plain HTTP for operator metrics
> 
> That's what we do for most components, and then add `kube-rbac-proxy` to export via HTTPS, so this ought to be easy / similar to other patches?

I assume that would entail adding a `kube-rbac-proxy` sidecar container like https://github.com/openshift/ingress-node-firewall/blob/1c7880a7aa38192761538fb98446acca09509e09/config/manager/manager.yaml#L89. But then there's the issue of how to configure the values for the `--tls` params instead of hard-coding in the yaml which is even more problematic.

tpantelis:
@danwinship I added https://github.com/openshift/cluster-network-operator/pull/3014/changes/e0a9cc3be30b771a3d1a3c1defc66ef4a7fd2fe3  to configure the TLS settings for the CNO metrics server. I moved the TLS profile watching from `ReconcileOperConfig` to a new controller that triggers a restart on TLS profile update. I can squash the two commits once reviewed.

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -177,6 +179,96 @@ func add(mgr manager.Manager, r *ReconcileOperConfig) error {
...
+	err = c.Watch(source.Kind[crclient.Object](mgr.GetCache(), &configv1.APIServer{},
+		handler.EnqueueRequestsFromMapFunc(reconcileOperConfig),
+		predicate.Funcs{
+			CreateFunc: func(evt event.CreateEvent) bool {
+				// Don't reconcile on initial creation/add events
+				return false
+			},
+			UpdateFunc: func(evt event.UpdateEvent) bool {
+				newAPI, ok := evt.ObjectNew.(*configv1.APIServer)
+				if !ok || newAPI.GetName() != openshifttls.APIServerName {
```
danwinship: It seems weird that we're getting the name of the APIServer config object from `openshifttls`... is this not available from `library-go` or `openshift/api` or somethign?

danwinship:
> A prior post asking about controllercmd had no resolution other than to move anyway from using controllercmd

@tpantelis if you have any further information about that, can you update https://redhat.atlassian.net/browse/CORENET-7296 with it?

pkg/controller/operconfig/operconfig_controller.go
tpantelis: No it's not defined in `library-go` or `openshift/api.`

tpantelis:
> if you have any further information about that, can you update https://redhat.atlassian.net/browse/CORENET-7296 with it?

Done

tpantelis:
@danwinship The required `ci/prow/e2e-aws-ovn-upgrade-ipsec` job is failing but unrelated to this PR (this was mentioned on the team call plus it's failing for other PRs).  Can you override the failure? (I don't have permission)

tpantelis:
@weliang1 This PR needs the verified label. Can you  do that?

weliang1:
@tpantelis Analyze CI job failures for two HyperShift clusters, it looks like both failures were triggered by the new TLS code changes.

ci/prow/e2e-aws-ovn-hypershift-conformance 

```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/3014/pull-ci-openshift-cluster-network-operator-master-e2e-aws-ovn-hypershift-conformance/2069558971898793984/artifacts/e2e-aws-ovn-hypershift-conformance/dump/artifacts/namespaces/clusters-795b79bb96d298684998/core/pods/logs/cluster-network-operator-5d5d96f989-fgw9n-cluster-network-operator.log

Error:
CNO Pod Log (the actual crash):                                                                                                     
    F0623 23:55:30.294919 main.go:100] failed to apply cluster TLS profile:                                                                                                  
        failed to get TLS profile:                                                                                                                                           
        failed to fetch HostedCluster clusters-795b79bb96d298684998/795b79bb96d298684998:                                                                                    
        hostedclusters.hypershift.openshift.io "795b79bb96d298684998" is forbidden:                                                                                          
        User "system:serviceaccount:clusters-795b79bb96d298684998:cluster-network-operator"                                                                                  
        cannot get resource "hostedclusters"...   
  The NEW TLS code in this PR tries to read the HostedCluster CR at startup, but the CNO service account doesn't have RBAC permission to read it!

  Result:
  1. CNO starts → Tries to fetch HostedCluster CR
  2. RBAC denies: "Forbidden"
  3. Fatal error → Pod crashes
  4. Kubernetes restarts → Same error → Crashes again
  5. After 11 crashes → CrashLoopBackOff
  6. Network never reconciles → Cluster stuck → Timeout

```

ci/prow/hypershift-e2e-aks

```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cluster-network-operator/3014/pull-ci-openshift-cluster-network-operator-master-hypershift-e2e-aks/2069558972062371840/build-log.txt

Key Evidence Found:
CNO deployment unavailable - "cluster-network-operator deployment has 1 unavailable replicas"
Network operator not available - Listed in unavailable cluster operators
Nodes have no CNI - "no CNI configuration file in /etc/kubernetes/cni/net.d/"
Cluster stuck in "Partial" - "controlPlaneVersion state is Partial, expected Completed"
30 min timeout - "Failed to wait...in 30m0s: context deadline exceeded"
```

tpantelis:
@danwinship @weliang1 I submitted https://github.com/openshift/hypershift/pull/8826 to address the Hypershift RBAC issue. I think it's correct.

tpantelis:
> @danwinship @weliang1 I submitted [openshift/hypershift#8826](https://github.com/openshift/hypershift/pull/8826) to address the Hypershift RBAC issue. I think it's correct.

So it turns out we should be using the `HostedControlPlane` resource instead of `HostedCluster` for **hypershift**, both contain the `configv1.APIServerSpec` and CNO already has RBAC permission to access `HostedControlPlane`.

tpantelis:
@weliang1 The `ci/prow/hypershift-e2e-aks` job succeeded this time but `ci/prow/e2e-aws-ovn-hypershift-conformance` failed again. But the CNO started successfully at 17:08:28 with no errors and TLS profile logic executed correctly (logged: _TLS adherence policy is "", using default TLS settings_). 

The test that failed the job is unrelated to this PR:

  - Test: `[sig-network][Feature:Router][apigroup:route.openshift.io]` The HAProxy router converges when 
  multiple routers are writing conflicting status
  - Location: _github.com/openshift/origin/test/extended/router/stress.go:238_
  - Error: timed out waiting for the condition after 136 seconds
  - Description: This is a router stress test that verifies HAProxy router convergence when multiple router
  instances write conflicting status to route objects

I notice the job failing in other PRs as well so it appears flaky, probably why it's not required 😄 

@danwinship Due to rebase and `HostedControlPlane` changes, it needs the **lgtm** again.

weliang1:
The above eight failures were caused by test flakiness or CI environment issues, not by the new code.

/verified by @weliang

```

● ✅ Baseline scan completed successfully!
   
The TLS regression baseline scan finished and scanned 10 IPs across 14 ports in about 3 minutes. All results are saved in regression-test/baseline/.The scan shows all endpoints are compliant with Modern TLS profile (TLS 1.2/1.3 only with strong ciphers).
```

danwinship:
@tpantelis can you squash the HostedControlPlane/HostedCluster fix back into the original commit (rather than having it initially be wrong and then fixing it). And then I'll approve and then wrestle with CI overrides until I can get this to merge

tpantelis:
> @tpantelis can you squash the HostedControlPlane/HostedCluster fix back into the original commit (rather than having it initially be wrong and then fixing it). And then I'll approve and then wrestle with CI overrides until I can get this to merge

Done

danwinship:
/verified by @weliang
(just rebased around irrelevant changes)
/lgtm

danwinship:
/override ci/prow/e2e-metal-ipi-ovn-ipv6-ipsec ci/prow/e2e-ovn-ipsec-step-registry ci/prow/e2e-aws-ovn-upgrade-ipsec

danwinship:
@tpantelis serial test is failing because you added a new host-network listening socket that doesn't have a Service pointing to it...

tpantelis:
> @tpantelis serial test is failing because you added a new host-network listening socket that doesn't have a Service pointing to it...

So the refactoring to override the default `StartController` behavior bypassed the `--listen flag` handling so the bind address was on the default port 8443 instead of the intended exposed port 9104. 
 
Fixed by extracting and applying the `--listen` flag value in `startControllerWithTLSCustomization`.

PS: I think we should consider migrating to the controller runtime API in the next release instead of using library-go's controllercmd.

danwinship:
> PS: I think we should consider migrating to the controller runtime API in the next release instead of using library-go's controllercmd.

yeah, CNO needs a lot of refactoring...

danwinship:
/verified by @weliang
and by CI

danwinship:
/retest-required
/override ci/prow/e2e-ovn-ipsec-step-registry ci/prow/e2e-aws-ovn-upgrade-ipsec

PR #3030, opened 2026-06-15
OCPBUGS-88531: Propagate restart-date annotation to CNO operand pod templates

bryan-cox:
## Summary

When the `hypershift.openshift.io/restart-date` annotation is set on a HostedControlPlane CR, CNO-managed operands like `cloud-network-config-controller` were not being restarted. CNO never propagated this annotation to its operand pod templates, so there was nothing to trigger a rollout.

Previously, the CPO (control-plane-operator) handled restart-date propagation by directly patching individual deployments (multus-admission-controller, network-node-identity, ovnkube-control-plane), but this approach missed cloud-network-config-controller entirely — which is the bug reported in OCPBUGS-88531.

This PR fixes the issue by having CNO propagate the annotation itself:

- Reading the `hypershift.openshift.io/restart-date` annotation from the HostedControlPlane CR in `ParseHostedControlPlane()`
- Adding a `SetRestartDateAnnotation()` helper that injects the restart-date into pod template annotations of rendered Deployments, DaemonSets, and StatefulSets **within the HCP namespace only** (guest-cluster workloads are not affected)
- Calling this helper in the reconciler after `Render()` returns, gated on HyperShift mode and a non-empty restart-date value

The companion HyperShift PR ([#8751](https://github.com/openshift/hypershift/pull/8751)) removes the redundant CPO-side restart logic, since CNO now handles it directly.

### Annotation removal behavior

If the `restart-date` annotation is removed from the HostedControlPlane, CNO will render manifests without the annotation on pod templates. This changes the pod template spec, triggering another rollout that restores the pods to a clean state (no restart-date annotation). This is expected behavior — removing the annotation is an explicit operator action.

## Test plan

- [x] Unit tests for `ParseHostedControlPlane` with restart-date annotation
- [x] Unit tests for `SetRestartDateAnnotation` covering Deployments, DaemonSets, StatefulSets, non-apps objects, namespace filtering, annotation preservation, and multi-object handling
- [x] `go build ./...` passes
- [x] `go test ./pkg/... ./cmd/...` passes
- [x] `go vet` passes
- [x] Manual e2e: Set `hypershift.openshift.io/restart-date` annotation on a HostedCluster and verified all 4 CNO operands restart (cloud-network-config-controller, multus-admission-controller, network-node-identity, ovnkube-control-plane). Full [test verification report](https://github.com/bryan-cox/architectural-artifact-sharing/tree/main/test-verification-report-ocpbugs-88531).

Fixes: https://issues.redhat.com/browse/OCPBUGS-88531

### DIFF
--- a/pkg/controller/operconfig/operconfig_controller.go
+++ b/pkg/controller/operconfig/operconfig_controller.go
@@ -365,6 +365,15 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		r.status.UnsetProgressing(statusmanager.OperatorRender)
 	}
 
+	if hcp := bootstrapResult.Infra.HostedControlPlane; hcp != nil && hcp.RestartDate != "" {
+		if err := hypershift.SetRestartDateAnnotation(objs, hcp.Namespace, hcp.RestartDate); err != nil {
+			log.Printf("Failed to set restart-date annotation: %v", err)
+			r.status.MaybeSetDegraded(statusmanager.OperatorConfig, "RenderError",
+				fmt.Sprintf("Internal error while setting restart-date annotation: %v", err))
+			return reconcile.Result{}, err
+		}
+	}
+
 	// The first object we create should be the record of our applied configuration. The last object we create is config.openshift.io/v1/Network.Status
 	app, err := AppliedConfiguration(operConfig)
 	if err != nil {
--- a/pkg/hypershift/hypershift.go
+++ b/pkg/hypershift/hypershift.go
@@ -36,6 +36,8 @@ const (
 	ClusterIDLabel = "_id"
 	// HyperShiftConditionTypePrefix is a cluster network operator condition type prefix in hostedControlPlane status
 	HyperShiftConditionTypePrefix = "network.operator.openshift.io/"
+	// RestartDateAnnotation is used to trigger rolling restarts of control plane operands
+	RestartDateAnnotation = "hypershift.openshift.io/restart-date"
 )
 
 type RelatedObject struct {
@@ -49,6 +51,7 @@ type RelatedObject struct {
 
 // HostedControlPlane represents a subset of HyperShift API definition for HostedControlPlane
 type HostedControlPlane struct {
+	Namespace                    string
 	ClusterID                    string
 	ControllerAvailabilityPolicy AvailabilityPolicy
 	NodeSelector                 map[string]string
@@ -58,6 +61,7 @@ type HostedControlPlane struct {
 	AdvertisePort                int
 	PriorityClass                string
 	APIServerSpec                *configv1.APIServerSpec
+	RestartDate                  string
 }
 
 // AvailabilityPolicy specifies a high level availability policy for components.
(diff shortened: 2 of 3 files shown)


bryan-cox:
/verified by @bryan-cox 

https://bryan-cox.github.io/architectural-artifact-sharing/test-verification-report-ocpbugs-88531/

pkg/hypershift/hypershift.go
```diff
@@ -151,6 +154,11 @@ func ParseHostedControlPlane(hcp *unstructured.Unstructured) (*HostedControlPlan
 		return nil, fmt.Errorf("failed to extract control plane priority class annotation: %v", err)
 	}
 
+	restartDate, _, err := unstructured.NestedString(hcp.UnstructuredContent(), "metadata", "annotations", RestartDateAnnotation)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract restart date annotation: %v", err)
```
mgencur, line 153: Nit: It would be better to use `%w` to preserve the error chain but since all other calls in this file use `%v` let's keep it.

pkg/hypershift/hypershift.go
```diff
@@ -264,9 +272,37 @@ func ParseHostedControlPlane(hcp *unstructured.Unstructured) (*HostedControlPlan
 		AdvertiseAddress:             advertiseAddress,
 		AdvertisePort:                int(advertisePort),
 		PriorityClass:                controlPlanePriorityClassAnnotation,
+		RestartDate:                  restartDate,
 	}, nil
 }
 
+// SetRestartDateAnnotation sets the restart-date annotation on pod templates of all
+// Deployment, DaemonSet, and StatefulSet objects to trigger a rolling restart.
+func SetRestartDateAnnotation(objs []*unstructured.Unstructured, restartDate string) error {
```
kyrtapz: This will cause a rollout of ALL CNO managed Deployments, DaemonSets, and StatefulSets, even if they run only in the hosted cluster.
I was sure that this annotation only affects the control-plane components in hypershift, is that not the case?

pkg/controller/operconfig/operconfig_controller.go
```diff
@@ -365,6 +365,15 @@ func (r *ReconcileOperConfig) Reconcile(ctx context.Context, request reconcile.R
 		r.status.UnsetProgressing(statusmanager.OperatorRender)
 	}
 
+	if hcp := bootstrapResult.Infra.HostedControlPlane; hcp != nil && hcp.RestartDate != "" {
```
kyrtapz, line 368: When the annotation is removed there will be another rollout, is this the expected behavior?

kyrtapz (commented):
> When the hypershift.openshift.io/restart-date annotation is set on a HostedControlPlane CR, CNO-managed operands like cloud-network-config-controller were not being restarted. This happened because CNO uses server-side apply with deterministic manifests — applying the same manifest twice is a no-op, so operands don't roll out unless something in the pod template changes.


While the change is mostly I don't get the reasoning from the PR description.
CNO never propagated this annotation to its operand pod templates, so there was nothing to trigger a rollout. This has nothing to do with SSA.

bryan-cox:
@kyrtapz re: the PR description — you're right, the SSA framing is misleading. The actual issue is simply that CNO never propagated the `restart-date` annotation to its operand pod templates, so there was nothing to trigger a rollout. I'll update the description.

pkg/hypershift/hypershift.go
bryan-cox: Good catch — `SetRestartDateAnnotation` currently operates on all rendered objects from `Render()`, which includes guest-cluster DaemonSets (ovnkube-node, multus, etc.), not just the HCP-namespace control-plane Deployments. The original CPO restart logic only targeted multus-admission-controller, network-node-identity, and ovnkube-control-plane in the HCP namespace.

I'll scope this by filtering on the HCP namespace so only control-plane objects are annotated.

pkg/controller/operconfig/operconfig_controller.go
bryan-cox: Yes, removing the annotation would cause another rollout since the pod template spec changes (annotation removed). This is the expected behavior — removing the annotation is an explicit operator action, and the resulting rollout restores the clean state. I'll document this in the PR description.

bryan-cox:
/verified by @bryan-cox

https://bryan-cox.github.io/architectural-artifact-sharing/test-verification-report-ocpbugs-88531/

kyrtapz:
/hold cancel
@bryan-cox you need a rebase

bryan-cox:
/verified by @bryan-cox

https://bryan-cox.github.io/architectural-artifact-sharing/test-verification-report-ocpbugs-88531/

bryan-cox:
@kyrtapz could I get lgtm again please when you have a moment?

I rebased.

PR #3018, opened 2026-06-01
OCPBUGS-85677: avoid Progressing during node reboot pod recreation

jluhrsen:
The counter check (UpdatedNumberScheduled < CurrentNumberScheduled) correctly detects CNO-initiated rollouts but also triggers during node reboots when pods are recreated without any DaemonSet spec change.

Require generation > observedGeneration to confirm CNO recently updated the DaemonSet, filtering out transient pod churn from node lifecycle events.

Fixes false positives during MCO node reboots with ipsec machine configs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DaemonSet rollout monitoring so rollout status no longer stays “active” based solely on scheduled counts. Rollout state now accounts for whether a newer generation is pending and whether the DaemonSet is truly progressing (while avoiding false “in-progress” signals during stalled or transitional states).

PR #3034, opened 2026-06-20
OCPBUGS-85677, OCPBUGS-90541: Don't set Progressing=False until all pods available

mdbooth:
We don't want the ClusterOperator to report Progressing=True whenever a Node reboots. However, we also don't want to report Progressing=False during a CNI rollout until all pods are available. This change ensures both are covered.

Related to https://github.com/openshift/origin/pull/31320

Adds a `networkoperator.openshift.io/observed-stable-generation` annotation to CNO operands when they are observed stable. Unlike the copy of the object's status temporarily stored in a ClusterOperator annotation during a rollout, this is persistent state.

With this change, an operand is rolling out if:
* Its observed generation does not match its generation
* It is not currently available and it has never been observed stable at this generation

That is, a rollout completes once we have observed a stable rollout at a particular generation at least once. Therefore:
* We only go Progressing=False once all pods are available
* We do not subsequently go Progressing=False unless the operand's generation changes, i.e. a for a spec change

This is v2 of this PR, but I have left the original commit in place for comparison and in case we want to roll back to it. The reasons for v2 are explained in https://github.com/openshift/cluster-network-operator/pull/3034#issuecomment-4903530028

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status messaging and condition handling for DaemonSets, StatefulSets, and Deployments when pods are temporarily unavailable, separating rollout progress from reboot/churn scenarios.
  * Enhanced detection of CrashLoopBackOff during these phases to improve failure reporting.
  * Added persistence to “observed stable” tracking so stability is retained across status-manager updates.

* **Tests**
  * Expanded automated coverage for “awaiting pod/node availability,” including timing behavior and persistence across restarts.

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -94,33 +95,25 @@ func (status *StatusManager) SetFromPods() {
 	for _, ds := range daemonSets {
 		dsName := NewClusteredName(ds)
 		dsState, hadState := daemonsetStates[dsName]
-		dsRolloutActive := !status.installComplete || ds.Status.UpdatedNumberScheduled < ds.Status.CurrentNumberScheduled
+		dsNeverObservedStable := !isObservedStable(ds)
 
 		dsProgressing := false
 
 		if isNonCritical(ds) && ds.Status.NumberReady == 0 && !status.installComplete {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is waiting for other operators to become ready", dsName.String()))
 			dsProgressing = true
-		} else if ds.Status.UpdatedNumberScheduled < ds.Status.CurrentNumberScheduled {
+		} else if ds.Status.ObservedGeneration < ds.Generation || ds.Status.UpdatedNumberScheduled < ds.Status.CurrentNumberScheduled {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q update is rolling out (%d out of %d updated)", dsName.String(), ds.Status.UpdatedNumberScheduled, ds.Status.CurrentNumberScheduled))
 			dsProgressing = true
-		} else if ds.Status.NumberUnavailable > 0 && (hadState || dsRolloutActive) {
-			// Rollout in progress: either continuing a tracked rollout or a new/initial rollout
-			if hadState {
-				progressing = append(progressing, fmt.Sprintf("DaemonSet %q rollout is waiting for %d pods to become available", dsName.String(), ds.Status.NumberUnavailable))
-			} else {
-				progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
-			}
-			dsProgressing = true
-			if !isNonCritical(ds) {
-				clbo = append(clbo, status.CheckCrashLoopBackOffPods(dsName, ds.Spec.Selector.MatchLabels, "DaemonSet")...)
-			}
 		} else if ds.Status.NumberUnavailable > 0 {
-			// Reboot churn: unavailable pods but no active rollout and no tracked state
+			if dsNeverObservedStable {
+				progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not available (awaiting %d nodes)", dsName.String(), ds.Status.NumberUnavailable))
+				dsProgressing = true
+			}
 			if !isNonCritical(ds) {
 				clbo = append(clbo, status.CheckCrashLoopBackOffPods(dsName, ds.Spec.Selector.MatchLabels, "DaemonSet")...)
 			}
-		} else if ds.Status.NumberAvailable == 0 && dsRolloutActive {
+		} else if ds.Status.NumberAvailable == 0 && dsNeverObservedStable {
 			progressing = append(progressing, fmt.Sprintf("DaemonSet %q is not yet scheduled on any nodes", dsName.String()))
 			dsProgressing = true
 		}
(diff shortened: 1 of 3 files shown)


mdbooth:
Pods are crashlooping due to:
```
 2026-06-20T18:21:33.522142681Z /usr/libexec/ipsec/addconn: /lib64/libc.so.6: version `GLIBC_ABI_DT_RELR' not found (required by /usr/libexec/ipsec/addconn)
```
Tracked in https://issues.redhat.com/browse/OCPBUGS-89238

jluhrsen:
> /assign @danwinship @jluhrsen

@mdbooth @danwinship ,

according to claude this PR is also fixing [OCPBUGS-85677](https://github.com/jluhrsen/cluster-network-operator/tree/OCPBUGS-85677) which I was chasing with this #3018.

Should I close 3018 and dup OCPBUGS-85677 to OCPBUGS-90541 ?

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

/retest

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

/retest

mdbooth:
> @jluhrsen: trigger 1 job(s) for the /payload-(with-prs|job|aggregate|job-with-prs|aggregate-with-prs) command
> 
> * periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec
> 
> See details on https://pr-payload-tests.ci.openshift.org/runs/ci/e1ee8920-725c-11f1-90a8-66ef2ab22b57-0

@jluhrsen This was green.

jluhrsen:
> > @jluhrsen: trigger 1 job(s) for the /payload-(with-prs|job|aggregate|job-with-prs|aggregate-with-prs) command
> > 
> > * periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec
> > 
> > See details on https://pr-payload-tests.ci.openshift.org/runs/ci/e1ee8920-725c-11f1-90a8-66ef2ab22b57-0
> 
> @jluhrsen This was green.

yes. we have 50 results now from the aggregate jobs and now of them showed the network operator as Progressing=True when it should not have.

I marked OCPBUGS-85677 as a duplicate of OCPBUGS-90541 and closed my original PR in favor of this one

/lgtm

/verified by aggregate jobs

jluhrsen:
/retitle OCPBUGS-85677, OCPBUGS-90541: Don't set Progressing=False until all pods available

/retest

mdbooth:
The direct cause of the ipsec failures is https://redhat.atlassian.net/browse/OCPBUGS-89238

However, I've noticed an additional problem here. Other jobs are currently passing the install step in CI even though the ipsec DS never becomes stable. This is because of a related bug in the rollout code:

### How the old code was always broken:

The original rollout renders the `ovn-ipsec-host` with `IPsecServiceCheckOnHost=false`. This rollout is successful, presumably because the IPsec config isn't rendered. However, the DS has now been annotated with installComplete=True.

MCO is updated to add ipsec. Nodes reboot, etc, etc. CNO observes the new MachineConfig and re-renders the `ovn-ipsec-host` ds with ipsec included. This crash loops immediately.

HOWEVER, because the ds had previously rendered successfully, it is still annotated with installComplete=true. Also dsRolloutActive= false because:

```go
dsRolloutActive := !status.installComplete || ds.Status.UpdatedNumberScheduled < ds.Status.CurrentNumber
Scheduled
```

The pods are all scheduled: they're just crash looping.

So when we enter the `if ds.Status.NumberUnavailable > 0` branch we don't do anything. This is deterministically broken.

### How my new code usually works, but has a race:

My new patch uses `hadState || dsRolloutActive` as a 'currently rolling out' guard. It uses this to set Progressing if pods are unavailable while we're still rolling out, which is what we want.

However, it has a race. `hadState` will only be true if we were lucky enough to observe the mid-rollout state when `ds.Status.UpdatedNumberScheduled < ds.Status.CurrentNumberScheduled`. This has no other guards and will set us Progressing=True, which results in state being written.

If we're unlucky and don't observe that state, e.g. because CNO happen to be running on the control plane node which was rebooted first, we will never reset the state on the ds which enables us to know we're rolling out again.

### A better fix

I'm going to push a new proposed fix which adds a new annotation: `networkoperator.openshift.io/observed-stable-generation`. We only write this annotation in SetFromPods when the DS/SS/D is observed stable, and it's set to the object's Generation. This means it is invalidated by any spec change, which will ensure we go Progressing=True until we have observed that the new spec is stable. It is unaffected anything which doesn't change the generation, e.g. a node reboot, which is also what we want.

jluhrsen:
/payload-aggregate periodic-ci-openshift-release-main-nightly-4.22-e2e-aws-ovn-upgrade-ipsec 10

jluhrsen:
@mdbooth is this ready?

mdbooth:
The hypershift failure is suspicious.

/test hypershift-e2e-aks

mdbooth:
The HyperShift failure highlighted a real bug which is also happening on standalone but wasn't being caught. Adding the annotation containing observed stable generation to the deployment/ss/ds itself bumps the generation, which causes a continuous rollout. This is a quirk of these resources in apps/v1 which I wasn't aware of. This will need to be reworked.

/hold

PR #3023, opened 2026-06-09
NO-JIRA: Update openshift/api and openshift/client-go to latest

tssurya:
Bumps openshift/api to pick up the EVPN feature gate promotion to GA (Default + OKD feature sets)

- github.com/openshift/api: v0.0.0-20260320151444-324a1bcb9f55 → v0.0.0-20260609121705-d3390bd1109f
- github.com/openshift/client-go: v0.0.0-20260320040014-4b5fc2cdad98 → v0.0.0-20260603140539-6892dc3e1ffc

PR #3035, opened 2026-06-22
CORENET-7297: Add CodeRabbit configuration for automated PR reviews

tssurya:
See https://github.com/openshift/cluster-network-operator/pull/3023#issuecomment-4740241977 for org level context
See https://docs.google.com/document/d/1LdyFK6nqH2ZovXc4Q9Hvf7v4B5ujzhi8DgzeOFcX_QE/edit?tab=t.0 for repo level instructions

Since we have `inheritance` set to true, we will inherit everything at org level.
This is just the beginning, I will do two more follow up PRs for

1. custom checks/pre-merge-checks and
2. path-specific review instructions along with adding more global instructions 
 
moving forward we need to be adding E2E's and developer docs in CNO itself for every PR - its not going to be acceptable to let QE folks add them later since all that process has now changed.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,39 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+# Inherit settings from the organization-level .coderabbit.yaml configuration.
+inheritance: true
+# https://docs.coderabbit.ai/reference/configuration#reviews
+reviews:
+  high_level_summary: true
+  high_level_summary_in_walkthrough: true
+  poem: false
+  # https://docs.coderabbit.ai/configuration/path-instructions
+  path_instructions:
+    - path: '**'
+      instructions: |
+        This is the OpenShift Cluster Network Operator (CNO). It manages the lifecycle
+        of cluster networking components in OpenShift including OVN-Kubernetes, Multus,
+        kube-proxy, network-diagnostics, egress-router, cloud-network-config-controller,
+        network-metrics, node-identity, FRR-K8s, iptables-alerter, MTU-prober,
+        container-networking-plugins, bond CNI, whereabouts CNI, route-override CNI,
+        multi-network-policy, and the networking console plugin.
+        - Focus on major issues impacting performance, readability, maintainability and security.
+        - Consider backward compatibility, upgrade safety, and multi-platform support.
+        - Avoid nitpicks and avoid verbosity.
+  # https://docs.coderabbit.ai/configuration/path-instructions#path-filters
+  path_filters:
+    - "!vendor/**"
+    - "!_output/**"
+  # https://docs.coderabbit.ai/reference/configuration#auto-review
+  auto_review:
+    drafts: false
+    ignore_title_keywords:
+      - WIP
+      - DNM
+    # To exclude labels, be sure to negate the label with '!'.
+    # Non-negated values turn this into an allowlist.
+    labels:
+      - '!work-in-progress'
+  # https://docs.coderabbit.ai/reference/configuration#slop-detection
+  slop_detection:
+    enabled: true
+    label: "coderabbit/ai-slop"


.coderabbit.yaml
```diff
@@ -0,0 +1,34 @@
...
+    networking console plugin. Reviews should consider backward compatibility,
+    upgrade safety, and multi-platform support.
+  high_level_summary: true
+  high_level_summary_in_walkthrough: true
+  poem: false
+  path_instructions:
+    - path: '**'
+      instructions: >-
+        -Focus on major issues impacting performance, readability, maintainability and security.
+        Avoid nitpicks and avoid verbosity.
```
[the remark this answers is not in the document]
tssurya: @coderabbitai: check in my latest push

tssurya:
The E2Es being red on this PR have nothing to do with the contents of the PR. They will all be overridden by me to save money and avoid retests on something that doesn't test any changes on the PR
CoderabbitAI config yaml changes have no impact on the product or the CNO binary
This is just part of AI SDLC framework for OCP to have repo specific config setup for coderabbitai.

.coderabbit.yaml
```diff
@@ -0,0 +1,34 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+# Inherit settings from the organization-level .coderabbit.yaml configuration.
+inheritance: true
+reviews:
+  instructions: >-
```
taanyas: Just confirming CNO also manages whereabouts-cni, right? Should we add that as well in the list.

.coderabbit.yaml
```diff
@@ -0,0 +1,34 @@
...
+  poem: false
+  path_instructions:
+    - path: '**'
+      instructions: |
+        - Focus on major issues impacting performance, readability, maintainability and security.
+        - Avoid nitpicks and avoid verbosity.
+  path_filters:
+    - "!vendor/**"
+    - "!_output/**"
+  auto_review:
```
taanyas, line 27: The internal doc's sample config includes ignore_usernames for bot/service accounts so automated PRs dont get reviewed. 
Is this intentionally left out?

taanyas (approved):
LGTM overall. Left two nits inline.

.coderabbit.yaml
tssurya: that's right! the internal doc's sample config was just showing us a way to do it
in CNO we don't have that many bot PRs on master branch - its normally only the case for bug fix backports but that's on older branches

.coderabbit.yaml
tssurya: that's a good point!
looks like I missed:
CNI plugins, bond CNI, whereabouts CNI, Egress Router CNI, Route override CNI, 
will add these

tssurya: fixed in latest push

tssurya:
thanks @taanyas for the review, here is the diff: https://github.com/openshift/cluster-network-operator/compare/b34e28fcc0ae5815054b14b5f27a7efe2603a3d3..f3fe3106a1da7467a8a1d4d71f14829893204d2b after your previous review

.coderabbit.yaml
```diff
@@ -0,0 +1,35 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+# Inherit settings from the organization-level .coderabbit.yaml configuration.
+inheritance: true
+reviews:
+  instructions: >-
```
miheer: Should not instructions be under path: '**' entry under path_instructions ?

```
  path_instructions:
    - path: '**'
      instructions: |
        This is the OpenShift Cluster Network Operator (CNO). It manages the lifecycle
        of cluster networking components in OpenShift including OVN-Kubernetes, Multus,
        kube-proxy, network-diagnostics, egress-router, cloud-network-config-controller,
        network-metrics, node-identity, FRR-K8s, iptables-alerter, MTU-prober,
        container-networking-plugins, bond CNI, whereabouts CNI, route-override CNI,
        multi-network-policy, EVPN, and the networking console plugin. Reviews should
        consider backward compatibility, upgrade safety, and multi-platform support.
        - Focus on major issues impacting performance, readability, maintainability and security.
        - Avoid nitpicks and avoid verbosity.
```

tssurya: The global instructions are in outer path - applicable to the repo level context

Path instructions are for path specific aspects i.e code related things - Like mentioned on PR describe we plan to add path instructions and custom checks in follow up PRs

I think coderabbit docs should have this already maybe try to see if you can spot thr global instructions syntaxing in their docs.

miheer: my bad ignore previous comment.

Shall we do a separation in case we plan to add path-specific rules later ?

```
reviews:
  instructions: >-
    This is the OpenShift Cluster Network Operator (CNO). It manages the lifecycle
    of cluster networking components in OpenShift. Reviews should consider
    backward compatibility, upgrade safety, and multi-platform support.

  path_instructions:
    - path: "**"
      instructions: |
        - Focus on major issues impacting performance, readability, maintainability, and security.
        - Avoid nitpicks and avoid verbosity.
```

tssurya: @coderabbitai: is this correct config to have global instructions outlined this way ?

tssurya: @miheer: I think it's fine for now... in FUP when we do path instructions we can break with new line if required.. usually in yaml files we don't leave new lines between every field...

tssurya: @coderabbitai: show me exact section in coderabbit docs for "reviews.instructions" field

tssurya: @coderabbitai No I don't see any review.instructions
Be specific like you did with path instructions

tssurya: @coderabbitiai : How can you not know your own config ?

tssurya: @coderabbitai - don't you have a yaml syntax validator - why is this broken with 502 , how are users supposed to validate syntax?

tssurya: FTR: VScode didnt flag this either. So stupid

tssurya: @coderabbitai; What do you mean ? See https://docs.coderabbit.ai/configuration/yaml-validator

tssurya: @coderabbitai why is https://docs.coderabbit.ai/configuration/yaml-validator not working ?

tssurya: @coderabbitai: I am disappointed in you. Looks like I need to manually verify all configs about you since you don't know yourself too well.
See
image
how can you tell the validator is working well?
If validator isn't working well what are we supposed to do?

tssurya: @coderabbitai: But how can you say:

"The PR itself as a live validator — CodeRabbit has already been running reviews on this PR, which means the config is being parsed and applied. If the YAML had a fatal parse or schema error, reviews would fail or produce warnings."

You didn't catch the "instructions" field right? that's wrong schema error right?

miheer:
may be we should have a test for this? @coderabbitai @tssurya  WDYT ?
`Makefile`:
```
.PHONY: verify
verify: verify-update-codegen verify-coderabbit-config 

.PHONY: verify-coderabbit-config
verify-coderabbit-config:
	hack/verify-coderabbit-config.sh
```

`hack/verify-coderabbit-config.sh`:
```
#!/usr/bin/env bash
set -euo pipefail

ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
CONFIG="${ROOT}/.coderabbit.yaml"
SCHEMA="${ROOT}/hack/coderabbit-schema.v2.json"

# 1. Must be parseable YAML.
python3 -c 'import sys,yaml; yaml.safe_load(open(sys.argv[1]))' "${CONFIG}"

# 2. Must match CodeRabbit's schema.
check-jsonschema --schemafile "${SCHEMA}" "${CONFIG}"

echo "OK: .coderabbit.yaml parses and matches the schema"
```

tssurya:
but @coderabbitai: does it work though? I had run it locally and it still passed for me with `instructions` field set and it didn't complain:
```
 check-jsonschema --schemafile https://coderabbit.ai/integrations/schema.v2.json .coderabbit.yml
ok -- validation done
```

miheer:
@coderabbitai, So instead of relying on the schema's own strictness, if we write a small Go tool that walks .coderabbit.yaml against the schema and enforces known-keys itself at every level, plus value types, enum values, and required keys  and give an output like this:

I tested this:
clean file

`OK: matches the CodeRabbit schema`

file with mistakes:
```
.coderabbit.yaml: 5 problem(s):
  - reviews.instructions: unknown field
  - reviews.path_instructions[1]: missing required key "path"
  - reviews.poem: expected boolean, got string
  - reviews.profile: agressive is not an accepted value (expected one of [assertive chill])
  - reviews.profle: unknown field
```

I was proposing something like this:
```
// Command coderabbitvalidate checks .coderabbit.yaml against the CodeRabbit
// configuration schema. It validates more than a stock JSON-Schema run does,
// because schema.v2.json only sets "additionalProperties": false on the root
// object -- not on reviews/auto_review/tools/etc. A conformant validator
// therefore accepts unknown or misspelled keys inside those objects (e.g.
// reviews.instructions, or "profle" for "profile"), which CodeRabbit then
// silently drops. This tool enforces, at every level:
//
//   - unknown / misspelled keys  (the silent-drop bug)
//   - value types                (e.g. poem must be a boolean)
//   - enum values                (e.g. profile in {chill, assertive})
//   - required keys              (e.g. each path_instructions entry needs path + instructions)
//
// It deliberately does NOT implement the full JSON-Schema feature set (anyOf
// branch validation beyond key names, $ref, format, numeric bounds, ...). For
// the mistakes that realistically land in this file the checks above are
// enough, and staying on the standard library plus the already-vendored
// github.com/ghodss/yaml means no new dependency and identical behaviour
// locally and in CI.
//
// Run via: go run ./hack/tools/coderabbitvalidate
package main

import (
	"embed"
	"encoding/json"
	"fmt"
	"math"
	"os"
	"sort"

	"github.com/ghodss/yaml"
)

//go:embed schema.v2.json
var schemaFS embed.FS

// allowExtra lists dotted paths that CodeRabbit accepts but that are missing
// from the published schema (the resolved model can run ahead of
// schema.v2.json -- e.g. a top-level `focus` field). "" is the root. Extend
// this rather than loosening the checks if a valid field is ever flagged.
var allowExtra = map[string]bool{
	"focus": true,
}

func main() {
	const cfgPath = ".coderabbit.yaml"

	cfgRaw, err := os.ReadFile(cfgPath)
	if err != nil {
		fail("cannot read %s: %v", cfgPath, err)
	}
	var cfg interface{}
	if err := yaml.Unmarshal(cfgRaw, &cfg); err != nil {
		fail("%s is not valid YAML: %v", cfgPath, err)
	}

	schemaRaw, err := schemaFS.ReadFile("schema.v2.json")
	if err != nil {
		fail("cannot read embedded schema: %v", err)
	}
	var schema map[string]interface{}
	if err := json.Unmarshal(schemaRaw, &schema); err != nil {
		fail("embedded schema is not valid JSON: %v", err)
	}

	var problems []string
	walk(cfg, schema, "", &problems)

	if len(problems) > 0 {
		sort.Strings(problems)
		fmt.Fprintf(os.Stderr, "%s: %d problem(s):\n", cfgPath, len(problems))
		for _, p := range problems {
			fmt.Fprintf(os.Stderr, "  - %s\n", p)
		}
		fmt.Fprintln(os.Stderr, "\nCodeRabbit silently ignores keys it does not recognize, so unknown")
		fmt.Fprintln(os.Stderr, "fields take no effect. Fix the spelling/nesting/value, or (for a field")
		fmt.Fprintln(os.Stderr, "CodeRabbit genuinely accepts) add its path to allowExtra in this tool.")
		os.Exit(1)
	}
	fmt.Printf("OK: %s matches the CodeRabbit schema\n", cfgPath)
}

// walk validates value against the schema node and recurses into nested
// objects and array elements, appending any problems found.
func walk(value, schemaNode interface{}, path string, problems *[]string) {
	sn, ok := schemaNode.(map[string]interface{})
	if !ok {
		return
	}

	if t, ok := sn["type"].(string); ok && value != nil {
		if !typeOK(t, value) {
			*problems = append(*problems, fmt.Sprintf("%s: expected %s, got %s", at(path), t, jsonType(value)))
		}
	}
	if enum, ok := sn["enum"].([]interface{}); ok && value != nil {
		if !inEnum(enum, value) {
			*problems = append(*problems, fmt.Sprintf("%s: %v is not an accepted value%s", at(path), value, formatEnum(enum)))
		}
	}

	switch v := value.(type) {
	case map[string]interface{}:
		if req, ok := sn["required"].([]interface{}); ok {
			for _, r := range req {
				if rk, ok := r.(string); ok {
					if _, present := v[rk]; !present {
						*problems = append(*problems, fmt.Sprintf("%s: missing required key %q", at(path), rk))
					}
				}
			}
		}
		allowed := allowedProps(sn)
		if allowed == nil {
			return // schema does not constrain keys here; leave them alone
		}
		for key, child := range v {
			full := join(path, key)
			childSchema, known := allowed[key]
			if !known {
				if !allowExtra[full] {
					*problems = append(*problems, full+": unknown field")
				}
				continue
			}
			walk(child, childSchema, full, problems)
		}
	case []interface{}:
		if items, ok := sn["items"]; ok {
			for i, elem := range v {
				walk(elem, items, fmt.Sprintf("%s[%d]", path, i), problems)
			}
		}
	}
}

// allowedProps returns declared property names for an object node, merging
// anyOf/oneOf/allOf branches (so a value that may be a string OR an object is
// handled). Returns nil when no object properties are declared, meaning keys
// are unconstrained and must not be flagged.
func allowedProps(sn map[string]interface{}) map[string]interface{} {
	out := map[string]interface{}{}
	found := false
	if props, ok := sn["properties"].(map[string]interface{}); ok {
		found = true
		for k, v := range props {
			out[k] = v
		}
	}
	for _, branchKey := range []string{"anyOf", "oneOf", "allOf"} {
		branches, ok := sn[branchKey].([]interface{})
		if !ok {
			continue
		}
		for _, b := range branches {
			if bm, ok := b.(map[string]interface{}); ok {
				if sub := allowedProps(bm); sub != nil {
					found = true
					for k, v := range sub {
						out[k] = v
					}
				}
			}
		}
	}
	if !found {
		return nil
	}
	return out
}

func typeOK(t string, v interface{}) bool {
	switch t {
	case "object":
		_, ok := v.(map[string]interface{})
		return ok
	case "array":
		_, ok := v.([]interface{})
		return ok
	case "string":
		_, ok := v.(string)
		return ok
	case "boolean":
		_, ok := v.(bool)
		return ok
	case "number":
		_, ok := v.(float64)
		return ok
	case "integer":
		f, ok := v.(float64)
		return ok && f == math.Trunc(f)
	}
	return true // unknown/absent type: don't enforce
}

func jsonType(v interface{}) string {
	switch v.(type) {
	case map[string]interface{}:
		return "object"
	case []interface{}:
		return "array"
	case string:
		return "string"
	case bool:
		return "boolean"
	case float64:
		return "number"
	}
	return "null"
}

func inEnum(enum []interface{}, v interface{}) bool {
	for _, e := range enum {
		if e == v {
			return true
		}
	}
	return false
}

// formatEnum lists allowed values when the set is small, otherwise stays quiet
// to avoid dumping huge enums (e.g. the language list).
func formatEnum(enum []interface{}) string {
	if len(enum) == 0 || len(enum) > 12 {
		return ""
	}
	vals := make([]string, 0, len(enum))
	for _, e := range enum {
		vals = append(vals, fmt.Sprintf("%v", e))
	}
	sort.Strings(vals)
	return fmt.Sprintf(" (expected one of %v)", vals)
}

func at(path string) string {
	if path == "" {
		return "(root)"
	}
	return path
}

func join(path, key string) string {
	if path == "" {
		return key
	}
	return path + "." + key
}

func fail(format string, a ...interface{}) {
	fmt.Fprintf(os.Stderr, "coderabbit-validate: "+format+"\n", a...)
	os.Exit(1)
}
```


The code will added in this location:
hack/tools/coderabbitvalidate/
├── main.go            # the tool (attached)
└── schema.v2.json     # vendored copy: curl -sSfL https://coderabbit.ai/integrations/schema.v2.json -o

miheer:
Steps I used to test:

` git fetch upstream pull/3035/head:pr-3035`


`git switch pr-3035`

`mkdir -p hack/tools/coderabbitvalidate`

put main.go into hack/tools/coderabbitvalidate/main.go
`ls hack/tools/coderabbitvalidate/main.go`

`curl -sSfL https://coderabbit.ai/integrations/schema.v2.json \
  -o hack/tools/coderabbitvalidate/schema.v2.json`

 should print "{", not HTML
`head -c 1 hack/tools/coderabbitvalidate/schema.v2.json`   

`go run ./hack/tools/coderabbitvalidate`
`OK: .coderabbit.yaml matches the CodeRabbit schema
`

temporarily break it
```
printf '  instructions: test\n  profle: chill\n' >> .coderabbit.yaml   
go run ./hack/tools/coderabbitvalidate ; echo "exit=$?"
git checkout -- .coderabbit.yaml                                       
```
It will give the following output:
```
.coderabbit.yaml: N problem(s):
  - reviews.instructions: unknown field
  - reviews.profle: unknown field
exit=1
```

We can then wire it to Makefile
```
verify: verify-update-codegen verify-coderabbit-config

.PHONY: verify-coderabbit-config
verify-coderabbit-config:
	go run ./hack/tools/coderabbitvalidate
```

tssurya:
this makes no changes to prod code, so overriding the jobs to save money:

/override ci/prow/e2e-gcp-ovn

PR #2784, opened 2025-08-25
[release-4.17]: Manual feature backport of cert rotation for whereabouts

bpickard22:
Includes cert check script logic and delpoyment of token watcher daemonset in openshift-mutlus namespace to watch for token change

PR #3040, opened 2026-06-29
 [release-4.16] OCPBUGS-63233: Manual feature backport of cert rotation for whereabouts

bpickard22:
Manual backport of the whereabouts-token-watcher feature and follow-up fixes to release-4.16.

release-4.16 did not contain the token-watcher feature at all, so this required a manual backport of the feature plus accumulated fixes rather than a simple cherry-pick. Cherry-picked from 4.17 (PR #2784) and master (#2727, #2841). Future backports of follow-up fixes should be possible via the cherry-pick bot.

Equivalent to what is on release-4.17.

### DIFF
--- a/bindata/network/multus/multus.yaml
+++ b/bindata/network/multus/multus.yaml
@@ -106,6 +106,23 @@ data:
     fi
     rm -Rf $UPGRADE_DIRECTORY
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: whereabouts-flatfile-config
+  namespace: openshift-multus
+data:
+  whereabouts.conf: |
+    {
+      "datastore": "kubernetes",
+      "kubernetes": {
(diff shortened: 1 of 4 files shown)


bindata/network/multus/multus.yaml
```diff
@@ -157,7 +174,6 @@ data:
           "bootstrapKubeconfig": "{{ .KubeletKubeconfigPath }}",
           "certDir": "/etc/cni/multus/certs",
           "certDuration": "24h"
-        },
```
kyrtapz, line 160: why was this removed? isn't it wrong? You end up with:
```
      "perNodeCertificate": {
          "enabled": true,
          "bootstrapKubeconfig": "{{ .KubeletKubeconfigPath }}",
          "certDir": "/etc/cni/multus/certs",
          "certDuration": "24h"
```

bpickard22: yep thats broken :( 
Dont know how i missed that, good catch. We need to fix this in 4.17 as well

PR #3045, opened 2026-07-09
NO-JIRA: Update OWNERS file

maayanf24:
Remove martinkennelly from reviewers list


OWNERS_ALIASES
tssurya, line 10: let's remove @fepan @pliurh @JacobTanenbaum from this list and from reviewers

PR #3046, opened 2026-07-10
OPNET-783: statusmanager: a DaemonSet matching zero nodes is not progressing

mkowalski:
## What

A DaemonSet whose node selector or affinity currently matches no nodes (`status.desiredNumberScheduled == 0` with the generation observed) is fully rolled out — but the status manager reported it as `DaemonSet ... is not yet scheduled on any nodes` (Progressing) and, after the timeout, as a hung rollout (Degraded, `rollout is not making progress`), blocking co/network.

The guard keeps an *unreconciled* DaemonSet (`observedGeneration < generation`, i.e. all-zero status right after creation) counted as progressing, so freshly created DaemonSets do not flap Available during install.

## Why now

Hit by BGP-based VIP management (openshift/enhancements#1982): on compact clusters the frr-k8s DaemonSet legitimately schedules zero pods (control plane nodes run the frr-k8s static pod instead and are excluded by node affinity), which degraded co/network and stalled installation. The bug is feature-independent though — any optional DaemonSet whose affinity matches no nodes triggers it.

## Testing

New `TestStatusManagerDaemonSetMatchingZeroNodes` (TDD — reproduces the exact live symptom before the fix): desired=0/observed-current ⇒ Available, not Progressing, not Degraded, no rollout state tracked; includes a stale-rollout-state + clock-step case proving no hung-rollout Degraded fires and state is pruned. Validated live in the enhancement-1982 reference implementation (https://github.com/mkowalski/bgp-vip-demo).

Part of OPNET-595; CNO subtask: [OPNET-783](https://issues.redhat.com/browse/OPNET-783).

### DIFF
--- a/pkg/controller/statusmanager/pod_status.go
+++ b/pkg/controller/statusmanager/pod_status.go
@@ -112,7 +112,11 @@ func (status *StatusManager) SetFromPods() {
 			if !isNonCritical(ds) {
 				clbo = append(clbo, status.CheckCrashLoopBackOffPods(dsName, ds.Spec.Selector.MatchLabels, "DaemonSet")...)
 			}
-		} else if ds.Status.NumberAvailable == 0 && dsRolloutActive {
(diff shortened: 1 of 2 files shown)


mkowalski:
@coderabbitai full review

PR #3043, opened 2026-07-06
CORENET-7054: Use TLS profile to render CLI args for deployed components

tpantelis:
See commits for details

PR #3050, opened 2026-07-13
CORENET-7356: Add TLS profile support for network-check-source metrics

tpantelis:
Use the OCP API server TLS profile to configure TLS options for the metrics server and exit the process on TLS profile changes.

`library-go's controllercmd` framework does not have support for TLS profiles and does not easily allow TLS options to be customized. Therefore the code was migrated to use the controller-runtime `Manager` API which allows direct customization.

The code respects the TLS profile adherence setting and uses the `library-go` strong defaults when adherence is optional so as not to fall back to weaker Go default ciphers.

### DIFF
--- a/bindata/network-diagnostics/network-check-source.yaml
+++ b/bindata/network-diagnostics/network-check-source.yaml
@@ -48,8 +48,6 @@ spec:
         args:
           - --listen
           - 0.0.0.0:17698
-          - --namespace
-          - $(POD_NAMESPACE)
         env:
           - name: POD_NAME
             valueFrom:
--- a/pkg/cmd/checkendpoints/cmd.go
+++ b/pkg/cmd/checkendpoints/cmd.go
@@ -2,14 +2,22 @@ package checkendpoints
 
 import (
 	"context"
+	"crypto/tls"
+	"fmt"
 	"os"
 	"time"
 
+	configv1 "github.com/openshift/api/config/v1"
 	operatorcontrolplaneclient "github.com/openshift/client-go/operatorcontrolplane/clientset/versioned"
 	operatorcontrolplaneinformers "github.com/openshift/client-go/operatorcontrolplane/informers/externalversions"
+	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
+	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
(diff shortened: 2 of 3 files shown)


tpantelis:
@coderabbitai approve

danwinship:
/pipeline required

I don't really understand all of this, but it should fail pretty obviously if something is wrong

tpantelis:
> I don't really understand all of this, but it should fail pretty obviously if something is wrong

We'll see how many succeed. But the failures I've looked at (with claude's help) are infrastructure-related or caused by other issues not related to this PR. Also this PR changes the `network-check-source` which is for diagnostics so not sure what CI job, if any, actually tests it. I deployed and tested it manually locally. @weliang1 intends to do pre-merge testing with this PR.

danwinship:
@tpantelis 
> `F0715 17:29:50.341379       1 cmd.go:53] failed to fetch TLS profile: failed to get TLS profile: failed to fetch apiserver.config.openshift.io/cluster: apiservers.config.openshift.io "cluster" is forbidden: User "system:serviceaccount:openshift-network-diagnostics:network-diagnostics" cannot get resource "apiservers" in API group "config.openshift.io" at the cluster scope`

It might be better to just have CNO deploy the pods with `--tls-cipher-suites` args or whatever, rather than having them try to figure it out themselves...

tpantelis:
> It might be better to just have CNO deploy the pods with `--tls-cipher-suites` args or whatever, rather than having them try to figure it out themselves...

yeah that's a good idea. We'll still need the `controller-runtime` approach to apply the TLS args, unless we keep `controllercmd` and do similar ugliness that I did in CNO _main.go_. I'd rather use `controller-runtime`, also b/c I don't think `controllercmd` will be updated to support TLS curve/group preferences coming in 5.1. Also changes in https://github.com/openshift/cluster-network-operator/pull/3043 will be needed so I might as well add another commit to that PR. Sound good?

danwinship:
sounds good

tpantelis:
> sounds good

Submitted https://github.com/openshift/cluster-network-operator/pull/3043/changes/1117c8e634259c22449e99c12e26aee26d243b9b and https://github.com/openshift/cluster-network-operator/pull/3043/changes/cbf9d4e91f782637fd6e429e6f41fb9b35814014 to https://github.com/openshift/cluster-network-operator/pull/3043

PR #3058, opened 2026-07-15
OSASINFRA-4400: Add support for max_allowed_address_pairs OpenStack

danchild:
Following work in [openshift/cloud-network-config-controller/pull/230](https://github.com/openshift/cloud-network-config-controller/pull/230), a mechanism is required to pass the configured `max_allowed_address_pairs` as a command line argument to the controller.

New feature includes:

- Add a bootstrap result type for a new `cloud-network-config` ConfigMap,
  reading it if the platform type is OpenStack
- Validate `platform-os-max-allowed-address-pairs`, returning an error
  if it is negative, explicitly set to 0, or an invalid number (e.g.
  alphabetic string)
- Pass `platform-os-max-allowed-address-pairs` to the
  `cloud-network-config-controller` as a command line argument
- Check for changes to `cloud-network-config`, restarting the
  cloud-network-config-controller if there's a diff

Please note that [openshift/cloud-network-config-controller/pull/230](https://github.com/openshift/cloud-network-config-controller/pull/230) must be accepted and merged before accepting these changes.

### DIFF
--- a/bindata/cloud-network-config-controller/self-hosted/controller.yaml
+++ b/bindata/cloud-network-config-controller/self-hosted/controller.yaml
@@ -42,12 +42,16 @@ spec:
         image: {{.CloudNetworkConfigControllerImage}}
         imagePullPolicy: IfNotPresent
         command: ["/usr/bin/cloud-network-config-controller"]
-        args: [ "-platform-type", "{{.PlatformType}}",
-                "-platform-region={{.PlatformRegion}}",
-                "-platform-api-url={{.PlatformAPIURL}}",
-                "-platform-aws-ca-override={{.PlatformAWSCAPath}}",
-                "-platform-azure-environment={{.PlatformAzureEnvironment}}",
-                "-secret-name", "cloud-credentials"]
+        args:
+        - "-platform-type={{.PlatformType}}"
+        - "-platform-region={{.PlatformRegion}}"
+        - "-platform-api-url={{.PlatformAPIURL}}"
+        - "-platform-aws-ca-override={{.PlatformAWSCAPath}}"
+        - "-platform-azure-environment={{.PlatformAzureEnvironment}}"
+        - "-secret-name=cloud-credentials"
+{{- if .OpenStackMaxAllowedAddressPairs }}
+        - "-platform-os-max-allowed-address-pairs={{ .OpenStackMaxAllowedAddressPairs }}"
+{{- end }}
         env:
         - name: CONTROLLER_NAMESPACE
           valueFrom:
--- a/pkg/bootstrap/types.go
+++ b/pkg/bootstrap/types.go
@@ -97,12 +97,20 @@ type IPTablesAlerterBootstrapResult struct {
 	Enabled bool
 }
 
+// CloudNetworkConfigBootstrapResult contains bootstrap configuration
+// read from the cloud-network-config ConfigMap.
+type CloudNetworkConfigBootstrapResult struct {
+	OpenStackMaxAllowedAddressPairs int
+}
+
 type BootstrapResult struct {
 	Infra InfraStatus
 
 	OVN             OVNBootstrapResult
 	IPTablesAlerter IPTablesAlerterBootstrapResult
 	TLSProfile      TLSProfile
+
+	CloudNetworkConfig CloudNetworkConfigBootstrapResult
 }
 
 type InfraStatus struct {
--- a/pkg/network/bootstrap.go
+++ b/pkg/network/bootstrap.go
@@ -2,10 +2,14 @@ package network
 
 import (
 	"context"
+	"fmt"
+	"strconv"
 
+	configv1 "github.com/openshift/api/config/v1"
 	operv1 "github.com/openshift/api/operator/v1"
 	"github.com/openshift/cluster-network-operator/pkg/bootstrap"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
+	"github.com/openshift/cluster-network-operator/pkg/names"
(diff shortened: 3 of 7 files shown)


pkg/bootstrap/types.go
```diff
@@ -97,12 +97,28 @@ type IPTablesAlerterBootstrapResult struct {
 	Enabled bool
 }
 
+// OSMaxAllowedAddressPairs holds the configured max_allowed_address_pairs value
+// from the cloud-network-config ConfigMap.
+type OSMaxAllowedAddressPairs struct {
+	Value    int
+	IsSet    bool   // used to differentiate between a zero value and an explicitly set zero value
```
winiciusallan: I think you can use `Value` as a `*int` to differentiate between set and unset instead of adding a new field to the struct.

pkg/bootstrap/types.go
```diff
@@ -97,12 +97,28 @@ type IPTablesAlerterBootstrapResult struct {
...
+	RawValue string // used for error reporting if user provides a non-numeric string
```
winiciusallan: Do we need this? We use a typed flag and the user shouldn't be able to pass a non-integer value to this flag. 

https://github.com/openshift/cloud-network-config-controller/pull/230/changes#diff-b52c6361b0946b90f2d35912dbe72c9dc8a5acc2d94a43cc26ad640b45e2be9aR267

bindata/cloud-network-config-controller/self-hosted/controller.yaml
```diff
@@ -42,12 +42,18 @@ spec:
...
         command: ["/usr/bin/cloud-network-config-controller"]
-        args: [ "-platform-type", "{{.PlatformType}}",
-                "-platform-region={{.PlatformRegion}}",
-                "-platform-api-url={{.PlatformAPIURL}}",
-                "-platform-aws-ca-override={{.PlatformAWSCAPath}}",
-                "-platform-azure-environment={{.PlatformAzureEnvironment}}",
-                "-secret-name", "cloud-credentials"]
+        args:
+        - "-platform-type"
+        - "{{.PlatformType}}"
```
stephenfin: nit: 
```suggestion
        - "-platform-type={{.PlatformType}}"
```

bindata/cloud-network-config-controller/self-hosted/controller.yaml
```diff
@@ -42,12 +42,18 @@ spec:
...
+        - "-platform-region={{.PlatformRegion}}"
+        - "-platform-api-url={{.PlatformAPIURL}}"
+        - "-platform-aws-ca-override={{.PlatformAWSCAPath}}"
+        - "-platform-azure-environment={{.PlatformAzureEnvironment}}"
+        - "-secret-name"
+        - "cloud-credentials"
```
stephenfin: nit:

```suggestion
        - "-secret-name=cloud-credentials"
```

pkg/bootstrap/types.go
```diff
@@ -97,12 +97,20 @@ type IPTablesAlerterBootstrapResult struct {
 	Enabled bool
 }
 
+// CloudNetworkConfigBootstrapResult contains bootstrap configuration
+// read from the cloud-network-config ConfigMap.
+type CloudNetworkConfigBootstrapResult struct {
+	OSMaxAllowedAddressPairs *int
```
stephenfin: Any reason for this to be a pointer? It will default to 0, which seems okay?

pkg/network/bootstrap.go
```diff
@@ -67,3 +78,31 @@ func iptablesAlerterBootstrap(cl crclient.Reader) bootstrap.IPTablesAlerterBoots
...
+	raw, ok := cm.Data["platform-os-max-allowed-address-pairs"]
+	if !ok {
+		return result, nil
+	}
+
+	val, err := strconv.Atoi(raw)
+	if err != nil {
+		return result, fmt.Errorf("error parsing cloud-network-config platform-os-max-allowed-address-pairs=%q: %w", raw, err)
+	}
+
```
stephenfin, line 106: Any reason not to do our validation of value here, before we store?

pkg/network/cloud_network.go
```diff
@@ -49,6 +50,15 @@ func renderCloudNetworkConfigController(conf *operv1.NetworkSpec, bootstrapResul
 	data.Data["PlatformAzureEnvironment"] = ""
 	data.Data["PlatformAWSCAPath"] = ""
 
+	cnc := bootstrapResult.CloudNetworkConfig
+	if cnc.OSMaxAllowedAddressPairs != nil {
+		if *cnc.OSMaxAllowedAddressPairs <= 0 {
+			return nil, fmt.Errorf("invalid cloud-network-config: platform-os-max-allowed-address-pairs must be a non-zero, positive integer, got %d", *cnc.OSMaxAllowedAddressPairs)
+		}
+	}
```
stephenfin: See my comments above RE: validation and use/non-use of a pointer

stephenfin (changes requested):
Couple of small comments and some nits

pkg/network/cloud_network.go
danchild: valid moving to `cloudNetworkConfigBootstrap`

bindata/cloud-network-config-controller/self-hosted/controller.yaml
danchild: changed

bindata/cloud-network-config-controller/self-hosted/controller.yaml
danchild: changed

danchild:
@verified by @danchild

PR #3072, opened 2026-07-20
OCPBUGS-88063: ovn-kubernetes: Move MNP from ConfigMap to CLI flags

weliang1:
## Summary

MultiNetworkPolicy is not enforced on UDN secondary interfaces in OCP 5.0 because ovnkube-control-plane pods do not restart when the ConfigMap is updated with `enable-multi-networkpolicy=true`.

## Root Cause

PR #2944 moved MNP enablement to ConfigMap, but:
1. The ConfigMap hash only includes 008-script-lib.yaml, not 004-config.yaml
2. ovnkube-control-plane has no hash annotation to trigger restarts

## Solution

Follow the same pattern as commit f4734c557 (OCPBUGS-78731): move MNP back to CLI flags where pod restart happens automatically on spec changes.

This creates consistency with multicast (also a CLI flag) and avoids the ConfigMap hash timing issues.

## Changes

- **managed/004-config.yaml**: Remove MNP from both ConfigMaps (HyperShift nodes + control plane)
- **managed/ovnkube-control-plane.yaml**: Add `--enable-multi-networkpolicy` CLI flag
- **self-hosted/004-config.yaml**: Remove MNP from ConfigMap
- **self-hosted/ovnkube-control-plane.yaml**: Add `--enable-multi-networkpolicy` CLI flag
- **ovn_kubernetes_test.go**: Update test expectations
- **docs/ovn_node_mode.md**: Document ConfigMap vs CLI flag feature enablement mechanisms

### Manual Testing
1. Deploy cluster with `UseMultiNetworkPolicy=false` in `network.operator/cluster`
   - Verify ovnkube-control-plane pods lack `--enable-multi-networkpolicy` flag
   - `oc get pod -n openshift-ovn-kubernetes -l app=ovnkube-control-plane -o yaml | grep enable-multi-networkpolicy` (should return nothing)

2. Set `UseMultiNetworkPolicy=true` via `oc edit network.operator cluster`
   - Verify ovnkube-control-plane pods restart (watch rolling update)
   - Verify `--enable-multi-networkpolicy` flag is present after restart
   - `oc get pod -n openshift-ovn-kubernetes -l app=ovnkube-control-plane -o yaml | grep enable-multi-networkpolicy` (should show flag)

3. Test MNP enforcement on UDN Layer2 and Layer3 networks
   - Verify MNP rules are enforced on secondary interfaces
   - Test cases: OCP-77656, OCP-78125, OCP-78259, OCP-77657

[...]

PR #3084, opened 2026-07-22
OCPBUGS-92080: Adds rendering of enable-multi-network-policy in ovnkube-node

raphaelvrosa:
Sets the flag `--enable-multi-networkpolicy` to be properly rendered in ovnkube-node config flags when updated by CNO.

Modifies unit test to evaluate the rendering of multiple Flags to ovnkube node.

Previous PR https://github.com/openshift/cluster-network-operator/pull/3072 added rendering of the flag to ovnkube-control-plane but missed ovnkube-node.


arghosh93:
Don't we need ovnkube-cluster-manager to also rollout when multiNetworkPolicy is enabled?

raphaelvrosa:
> Don't we need ovnkube-cluster-manager to also rollout when multiNetworkPolicy is enabled?

In ovnk code, this flag is not being used by ovnkube-cluster-manager.

RamLavi (commented):
Hey @raphaelvrosa o/
This change makes `--enable-multi-networkpolicy` a node CLI flag in `008-script-lib.yaml`, which is the right fix IMO.

However, `docs/ovn_node_mode.md` still says ovnkube-node already restarts when the ConfigMap changes, so only control-plane-specific features need CLI flags. I think that sentence is what led [OCPBUGS-88063](https://redhat.atlassian.net/browse/OCPBUGS-88063) to add the flag only on control-plane and drop it from the ConfigMap, leaving node with EnableMultiNetworkPolicy=false.

Please update that doc in this PR: 008-script-lib is hashed (node restarts), 004-config is not, and MNP is a CLI flag on both control-plane and node.

raphaelvrosa:
Not needed anymore, done by https://github.com/openshift/cluster-network-operator/pull/3072/changes

docs/ovn_node_mode.md
```diff
@@ -19,6 +19,6 @@ Feature enablement is managed through two mechanisms:
 
 - **ConfigMap-based** (`004-config.yaml`): Most features (egress IP, multi-network, network segmentation, admin network policy, etc.) are configured in the cluster-wide ConfigMap which is passed to ovnkube via `--config-file`.
 
-- **CLI flags** (`ovnkube-control-plane.yaml`): Features that require ovnkube-control-plane pod restarts on configuration changes (multicast, multi-networkpolicy) are enabled via CLI flags (e.g., `--enable-multicast`, `--enable-multi-networkpolicy`) to ensure the control-plane pods restart automatically when the feature is toggled. Note that ovnkube-node pods already restart when the ConfigMap changes, so only control-plane-specific features require CLI flags.
+- **CLI flags** (`ovnkube-control-plane.yaml`): Features that require ovnkube-control-plane pod restarts on configuration changes (multicast, multi-networkpolicy) are enabled via CLI flags (e.g., `--enable-multicast`, `--enable-multi-networkpolicy`) to ensure the control-plane pods restart automatically when the feature is toggled.
```
RamLavi: Thanks for dropping the “only control-plane-specific features require CLI flags” line. 

I would argue that this still reads as if `--enable-multi-networkpolicy` is control-plane-only (`ovnkube-control-plane.yaml`). This PR also adds it in hashed `008-script-lib.yaml` so ovnkube-node rolls. 

Please consider mentioning that.

RamLavi (commented):
one last thing

docs/ovn_node_mode.md
raphaelvrosa: Sure, fixed.

RamLavi (approved):
LGTM

thanks!

RamLavi:
> @RamLavi: This pull request references [Jira Issue OCPBUGS-92080](https://redhat.atlassian.net/browse/OCPBUGS-92080), which is invalid:
> 
> * expected the bug to target either version "5.1.0._" or "openshift-5.1.0._", but it targets "5.0.0" instead
> 
> Comment `/jira refresh` to re-evaluate validity if changes to the Jira bug are made, or edit the title of this pull request to link to a different bug.
> 
> Details

@raphaelvrosa can you set the JIRA to target the correct version?

raphaelvrosa:
The only failing test is bad/red since August 13th.

raphaelvrosa:
@jcaamano can you override the failing test so we can merge this? It's consistently failing.

PR #3098, opened 2026-08-03
OCPBUGS-98149: use python3 instead of platform-python

pperiyasamy:
RHEL10 on s390x does not have platform-python. Switch to /usr/bin/python3 for ovs-monitor-ipsec invocation.


taanyas (approved):

mattedallo (approved):

pperiyasamy:
/close 

The https://github.com/openshift/cluster-network-operator/pull/2662 fixes the issue by using container's `/usr/libexec/platform-python` file.

PR #3104, opened 2026-08-04
OCPBUGS-64582: frr-k8s: use Recreate strategy for statuscleaner deployment

fedepaol:
The statuscleaner deployment uses hostNetwork with a fixed port (9123). With the default RollingUpdate strategy, upgrades on SNO clusters get stuck because the new pod cannot bind the host port already held by the old pod. Switching to Recreate ensures the old pod is terminated before the new one starts.

### DIFF
--- a/bindata/network/frr-k8s/node-status-cleaner.yaml
+++ b/bindata/network/frr-k8s/node-status-cleaner.yaml
@@ -9,6 +9,10 @@ metadata:
   annotations:
     release.openshift.io/version: "{{.ReleaseVersion}}"
 spec:
+{{- if .IsSNO }}
+  strategy:
+    type: Recreate
+{{- end }}
   selector:
     matchLabels:
       component: frr-k8s-statuscleaner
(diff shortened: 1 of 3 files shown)


bindata/network/frr-k8s/node-status-cleaner.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
   annotations:
     release.openshift.io/version: "{{.ReleaseVersion}}"
 spec:
+  strategy:
```
jcaamano, line 13: Wouldn't it be best to guard this by the `isSNO` variable?

fedepaol: ah thanks! Didn't know about its existance. I'll do that, although the effect of this change is a small window (during upgrades) where changes to CRDs are rejected, so I don't think it would hurt much in non - sno scenarios.

Anyways, adding it

jcaamano:

https://redhat-internal.slack.com/archives/CBN38N3MW/p1786420516017809

PR #3108, opened 2026-08-05
frr-k8s: sync CRD

oribon:
sync with https://github.com/openshift/frr/pull/135


oribon:
/verified by ci

in https://github.com/openshift/frr/pull/135

PR #3112, opened 2026-08-07
NO-ISSUE: Drop github.com/pkg/errors

skitt:
# Why

The module is deprecated and replaceable using stdlib constructs (with differences around stacktrace handling which shouldn't be significant here).

# What

This was mostly done using

```
git grep -l errors.Wrapf | xargs sed -i -E 's/errors.Wrapf\(err, "([^"]*)"([^)]*)\)/fmt.Errorf("\1: %w"\2, err)/'
git grep -l errors.Errorf | xargs sed -i 's/errors.Errorf/fmt.Errorf/'
git grep -l errors.Wrap | xargs sed -i -E 's/errors.Wrap\(err, "([^"]*)"\)/fmt.Errorf("\1: %w", err)/'
```

Some fmt.Errorf calls could be replaced by errors.New but the impact is minimal (I can't find the reference right now, but the intention of the Go authors is that fmt.Errorf can be used by default even when formatting isn't needed; it means one less thing to think about when the need for formatting changes).

# Verification

This is a refactor with no functional changes, so most existing tests don’t need adjusting and should be sufficient to verify correctness. There is one behaviour change, related to error-message capitalisation; the corresponding test is adjusted to suit.


skitt:
@coderabbitai resume

pkg/controller/allowlist/allowlist_controller.go
```diff
@@ -4,6 +4,7 @@ package allowlist
 
 import (
 	"context"
+	"errors"
```
danwinship: might as well replace the one `errors.New()` call with `fmt.Errorf()`

pkg/network/kube_proxy.go
```diff
@@ -1,15 +1,16 @@
...
+	"fmt"
 	"net"
 	"os"
 	"path/filepath"
 	"strconv"
 	"time"
 
-	"github.com/pkg/errors"
+	v1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-network-operator/pkg/hypershift"
```
danwinship: these are still out-of-place. (there are other CNO imports below)

skitt: I’ve grouped them for now — I wasn’t able to determine from the rest of the project whether CNO imports are supposed to be in their own section; is that the case?

PR #3113, opened 2026-08-07
NO-ISSUE: modernise the codebase using go fix

skitt:
# Why

Go now provides a built-in fix-up tool, `go fix`. It is intended to apply safe fixes which use newer code patterns to improve maintenability and/or performance.

# What

This is a one-time `go fix` run over the project, simplifying more regular maintenance using `go fix` in future.

See individual commits for details.

# Verification

This is a refactor with no behaviour changes, so existing tests don’t need adjusting and should be sufficient to verify correctness.


skitt:
@coderabbitai review

PR #3119, opened 2026-08-12
OCPBUGS-105887:Filter unsupported cipher suites to prevent ovnkube-identity crash

tpantelis:
When the Old TLS profile is configured with StrictAllComponents adherence, three cipher suites (ECDHE-ECDSA-AES256-SHA384, ECDHE-RSA-AES256-SHA384, AES256-SHA256) are included in the profile but are not supported by Go's crypto/tls package.

Previously, these OpenSSL cipher names were passed unconverted to ovnkube-identity's --tls-cipher-suites argument, causing the webhook container to crash with "Cipher suite ECDHE-ECDSA-AES256-SHA384 not supported or doesn't exist".

The fix silently filters out ciphers that cannot be converted from OpenSSL to IANA format, which indicates they are unsupported by Go's crypto/tls. This aligns with library-go's intentional design where OpenSSLToIANACipherSuites() already logs dropped ciphers at V(4).

NGINX components continue to receive the full OpenSSL cipher list since they support these ciphers natively.

### DIFF
--- a/pkg/network/tls.go
+++ b/pkg/network/tls.go
@@ -49,9 +49,9 @@ func addTLSInfoToRenderData(data map[string]interface{}, bootstrapResult *bootst
 		converted := crypto.OpenSSLToIANACipherSuites([]string{cipher})
 		if len(converted) > 0 {
 			ianaCiphers = append(ianaCiphers, converted...)
-		} else {
-			ianaCiphers = append(ianaCiphers, cipher)
 		}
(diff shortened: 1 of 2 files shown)


danwinship:
/retest
(@tpantelis this will also need an appropriate `/verified`)

tpantelis:
@danwinship The `ci/prow/e2e-metal-ipi-ovn-ipv6` job failure isn't related to this PR (_[sig-auth][Feature:ProjectAPI] TestProjectWatch should succeed)_. Can we override it?

PR #3131, opened 2026-08-21
OCPBUGS-112278: Add ConfigMap hash annotation to networking console plugin deployment

tpantelis:
When TLS parameters change, the nginx ConfigMap is updated but the pods don't automatically restart to pick up the new configuration. This adds a hash of the ConfigMap content as a pod template annotation, which triggers a rolling update when the ConfigMap changes.


tpantelis:
@danwinship Job `ci/prow/e2e-metal-ipi-ovn-dualstack-bgp-local-gw` is stubbornly failing not related to this PR. Can you override?

PR #2944, opened 2026-03-20
NVIDIA-554: DPU-host mode: use ConfigMap for OVN feature enablement instead of per-node script gating

tsorya:
NVIDIA-554: DPU-host mode: use ConfigMap for OVN feature enablement instead of per-node script gating

         All OVN-Kubernetes features (egress IP, egress firewall, multicast,
      multi-network, admin network policy, multi-external-gateway, etc.)
      are now enabled in DPU-host mode. The OVN controller on DPU-host
      nodes processes the configuration but does not offload egress IP
      datapath — traffic follows the regular kernel path instead.

        OVN_NODE_MODE remains used only for DPU-host structural differences:
        gateway interface, ovnkube-node-mode flag, and init-controller.

      Because these features are safe to enable cluster-wide, the per-node
      gating logic in the startup script (008-script-lib.yaml) is no
      longer needed. Feature flags are managed solely through the
      cluster-wide ConfigMap (004-config.yaml) passed to ovnkube via
      --config-file.

        Co-authored-by: Cursor <cursoragent@cursor.com>
        Made-with: Cursor

    This reverts commit 6f5697cb0b25fe1ad6cb48a9c969653475afa4b9

PR #3136, opened 2026-08-24
OCPBUGS-113532: ovn-kubernetes: Move MNP from ConfigMap to CLI flags

raphaelvrosa:
MultiNetworkPolicy is not enforced on UDN secondary interfaces in OCP 5.0 because ovnkube-control-plane pods do not restart when the ConfigMap is updated with enable-multi-networkpolicy=true.

Root cause: PR #2944 moved MNP enablement to ConfigMap, but:
1. The ConfigMap hash only includes 008-script-lib.yaml, not 004-config.yaml
2. ovnkube-control-plane has no hash annotation to trigger restarts

Solution: Follow the same pattern as commit f4734c557 (OCPBUGS-78731): move MNP back to CLI flags where pod restart happens automatically on spec changes.

This creates consistency with multicast (also a CLI flag) and avoids the ConfigMap hash timing issues that caused the original [NVIDIA-554](https://redhat.atlassian.net/browse/NVIDIA-554) fix to be reverted.

How to verify:
1. Deploy cluster with UseMultiNetworkPolicy=false
   - Verify ovnkube-control-plane pods lack --enable-multi-networkpolicy flag
2. Set UseMultiNetworkPolicy=true via network.operator/cluster
   - Verify ovnkube-control-plane pods restart with --enable-multi-networkpolicy
3. Test MNP enforcement on UDN Layer2/Layer3 networks
4. Verify DPU-host mode still works correctly
5. CI lanes: e2e-aws-ovn, e2e-gcp-ovn, e2e-metal-ipi-ovn-dualstack

Tested:
- MNP works on UDN Layer2 and Layer3 networks
- Pods restart when UseMultiNetworkPolicy changes
- DPU-host mode still works correctly
- Other ConfigMap features (egress-ip, etc.) unaffected

Jira: OCPBUGS-88063

docs: Clarify CLI flags are for control-plane pod restarts

Update ovn_node_mode.md to specify that CLI flags are used for features requiring ovnkube-control-plane pod restarts, not just any pod restarts. This clarifies that ovnkube-node pods already restart when the ConfigMap changes, so CLI flags are only needed for control-plane-specific features.

Addresses review feedback from danwinship.

(cherry picked from commit 17f08a709659914743be851d72bf70a39a9abbf5)

### DIFF
--- a/bindata/network/ovn-kubernetes/managed/004-config.yaml
+++ b/bindata/network/ovn-kubernetes/managed/004-config.yaml
@@ -43,10 +43,6 @@ data:
     enable-multi-network=true
     enable-network-segmentation=true
     enable-preconfigured-udn-addresses=true
-
-{{- if .OVN_MULTI_NETWORK_POLICY_ENABLE }}
-    enable-multi-networkpolicy=true
-{{- end }}
     enable-admin-network-policy=true
     enable-multi-external-gateway=true
 {{- if .DNS_NAME_RESOLVER_ENABLE }}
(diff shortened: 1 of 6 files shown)


raphaelvrosa:
@weliang1 can you add verify label? 
@danwinship can you review/approve? It's another backport.

PR #3141, opened 2026-08-25
[release-4.20] OCPBUGS-113989: frr-k8s: use Recreate strategy for webhook server deployment on SNO

jechen0648:
Manual backport of [#3135](https://github.com/openshift/cluster-network-operator/pull/3135) to release-4.20, superseding the failed auto cherry-pick [#3140](https://github.com/openshift/cluster-network-operator/pull/3140).

On SNO clusters, the frr-k8s webhook server Deployment uses hostNetwork with a fixed port (9123). With the default RollingUpdate strategy, upgrades get stuck because the new pod cannot bind the host port already held by the old pod. This PR:

Switches the frr-k8s-webhook-server Deployment to Recreate strategy on SNO clusters
Adds the networkoperator.openshift.io/pre-patch annotation so CNO can atomically switch strategy type and clear rollingUpdate fields via strategic-merge-patch before SSA (required on upgrade when the API server has already defaulted rollingUpdate)
Explicitly sets RollingUpdate with rollingUpdate fields on non-SNO clusters so CNO owns those fields going forward
Wires bootstrapResult into renderAdditionalRoutingCapabilities to determine SNO via ControlPlaneReplicaCount
Adds unit tests for the pre-patch apply path and deployment strategy rendering
4.20-specific changes
On release-4.20, the hostNetwork workload is frr-k8s-webhook-server (in webhook.yaml). The separate frr-k8s-statuscleaner Deployment (node-status-cleaner.yaml) was only introduced in 4.21.

The auto cherry-pick in #3140 failed unit tests because it cherry-picked the 4.21 test Test_renderFRRStatusCleanerStrategy, which looks for a Deployment that does not exist on 4.20. This PR fixes that by:

Renaming the test to Test_renderFRRWebhookServerStrategy
Targeting frr-k8s-webhook-server instead of frr-k8s-statuscleaner

Test plan
go test ./pkg/network/
go test ./pkg/apply/
CI unit job on PR

### DIFF
--- a/pkg/controller/statusmanager/status_manager_test.go
+++ b/pkg/controller/statusmanager/status_manager_test.go
@@ -250,30 +250,33 @@ func TestStatusManager_set(t *testing.T) {
 
 	obj := &uns.Unstructured{}
 	gvk := schema.GroupVersionKind{
-		Group:   "test",
-		Version: "test",
-		Kind:    "test",
+		Group:   "",
+		Version: "v1",
+		Kind:    "ConfigMap",
 	}
 	obj.SetGroupVersionKind(gvk)
 	obj.SetName("current")
+	obj.SetNamespace("default")
 	set(t, client, obj)
 
 	co.Status.RelatedObjects = []configv1.ObjectReference{
 		{
-			Group:    "test",
-			Resource: "test",
-			Name:     "current",
+			Group:     "",
+			Resource:  "configmaps",
+			Name:      "current",
+			Namespace: "default",
 		},
 	}
 	status.relatedObjects = []configv1.ObjectReference{
 		{
-			Group:    "test",
-			Resource: "test",
-			Name:     "related",
+			Group:     "",
+			Resource:  "configmaps",
+			Name:      "related",
+			Namespace: "default",
 		},
 	}
 	status.deleteRelatedObjectsNotRendered(co)
-	err = status.client.ClientFor("").CRClient().Get(context.TODO(), types.NamespacedName{Name: "current"}, obj)
+	err = status.client.ClientFor("").CRClient().Get(context.TODO(), types.NamespacedName{Namespace: "default", Name: "current"}, obj)
 	if err == nil {
 		t.Fatalf("unexpected related object in ClusterOperator object was not deleted")
 	}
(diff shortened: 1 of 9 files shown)


pkg/network/render_test.go
```diff
@@ -655,7 +655,8 @@ func Test_renderAdditionalRoutingCapabilities(t *testing.T) {
 	}
 }
 
-func Test_renderFRRStatusCleanerStrategy(t *testing.T) {
+func Test_renderFRRWebhookServerStrategy(t *testing.T) {
```
andreaskaris, line 658: We need to squash this commit into commit `frr-k8s: use Recreate strategy for statuscleaner deployment`  and add a note at the end of the commit about the change that was made.

pkg/network/render_test.go
```diff
@@ -655,7 +655,8 @@ func Test_renderAdditionalRoutingCapabilities(t *testing.T) {
...
+	// On 4.20 the hostNetwork workload is frr-k8s-webhook-server; statuscleaner was split out in 4.21.
```
andreaskaris, line 659: that comment here isn't necessary

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -250,30 +250,33 @@ func TestStatusManager_set(t *testing.T) {
 
 	obj := &uns.Unstructured{}
 	gvk := schema.GroupVersionKind{
-		Group:   "test",
-		Version: "test",
-		Kind:    "test",
+		Group:   "",
```
andreaskaris, line 253: this commit here makes no sense - the unit test is exactly the same in 4.21 where it passes without problems, meaning if there are problems, they lie elsewhere

andreaskaris: I found a better solution (IMO). I'll push it in a bit

andreaskaris: See https://github.com/openshift/cluster-network-operator/pull/3142

jechen0648:
prefer https://github.com/openshift/cluster-network-operator/pull/3142

PR #3146, opened 2026-09-02
CORENET-7275: Add AGENTS.md, point CodeRabbit knowledge base at it

kyrtapz:
Centralize CNO architecture/conventions/gotchas in AGENTS.md so both AI coding agents and CodeRabbit share one source of truth instead of duplicating knowledge into .coderabbit.yaml path_instructions.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,186 +1,78 @@
 # yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
 # Inherit settings from the organization-level .coderabbit.yaml configuration.
 inheritance: true
+
+# https://docs.coderabbit.ai/reference/configuration#knowledge-base
+# Feed AGENTS.md (architecture, conventions, gotchas) to CodeRabbit as review
+# context so it doesn't need to be duplicated into path_instructions below.
+knowledge_base:
+  code_guidelines:
+    enabled: true
+    filePatterns:
+      - "**/AGENTS.md"
+      - "docs/**/*.md"
+
 # https://docs.coderabbit.ai/reference/configuration#reviews
 reviews:
   high_level_summary: true
   high_level_summary_in_walkthrough: true
   poem: false
   # https://docs.coderabbit.ai/configuration/path-instructions
+  # Keep these short — they're path-scoping + reviewer-tone hints. Architecture
+  # facts and gotchas live in AGENTS.md (loaded via knowledge_base above); link
+  # to a section there instead of repeating it here.
   path_instructions:
     - path: '**'
       instructions: |
-        This is the OpenShift Cluster Network Operator (CNO). It manages the lifecycle
-        of cluster networking components in OpenShift including OVN-Kubernetes, Multus,
-        kube-proxy, network-diagnostics, egress-router, cloud-network-config-controller,
-        network-metrics, node-identity, FRR-K8s, iptables-alerter, MTU-prober,
-        container-networking-plugins, bond CNI, whereabouts CNI, route-override CNI,
-        multi-network-policy, EVPN, and the networking console plugin.
+        This is the OpenShift Cluster Network Operator (CNO) — see AGENTS.md for
+        project overview, architecture, and conventions.
         - Focus on major issues impacting performance, readability, maintainability and security.
         - Consider backward compatibility, upgrade safety, and multi-platform support.
         - Avoid nitpicks and avoid verbosity.
-        - In HyperShift, CNO is deployed by openshift/hypershift (not CVO). If changes
-          touch `manifests/` or the CNO deployment, check whether corresponding changes
-          are needed in the openshift/hypershift repo.
         - Dual-stack (IPv4+IPv6) must always be considered. Flag changes that only
-          handle single-stack addresses (e.g. missing net.JoinHostPort for IPv6).
-        - Two CRDs drive CNO configuration:
-          `network.config.openshift.io/cluster` is the source of truth set by the
-          installer. `network.operator.openshift.io/cluster` is the operational
-          config CNO actually renders from. On each reconcile, CNO copies
-          `clusterNetwork`, `serviceNetwork`, and `networkType` from config → operator,
-          unconditionally overwriting whatever is in the operator CR for those fields.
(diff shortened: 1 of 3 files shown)


kyrtapz:
@danwinship are you interested in taking a look? It's mostly moving things around since I realized the knowledge we captured in the review guidelines fits well in the AGENTS.md

.coderabbit.yaml
```diff
@@ -1,186 +1,84 @@
...
+        `managed/`/`self-hosted/`/`common/` split, restart-hash triggers, and
+        rollout ordering are the #1 source of HyperShift drift bugs here.
+        Also audit container command/args for duplicated or obsolete CLI flags.
+
     - path: 'bindata/network/node-identity/**'
       instructions: |
-        Node identity controller — has `managed/` and `self-hosted/` variants
-        like OVN-K. Ensure both variants stay aligned when making changes.
+        Has the same `managed/`/`self-hosted/` split as OVN-Kubernetes
+        (see AGENTS.md). Ensure both variants stay aligned.
```
danwinship, line 65: (Interesting... I assume this instruction didn't actually work reliably before, since the wording assumes that coderabbit LLM had also ingested the `instructions` from the `'bindata/network/ovn-kubernetes/**'` rule above, which it presumably wouldn't have if there were no changes in that directory. The new rule seems fine though.)

.coderabbit.yaml
```diff
@@ -1,186 +1,84 @@
...
-        - Every render function must set `data.Data["ReleaseVersion"]` from
-          `os.Getenv("RELEASE_VERSION")`. Missing this = the component's workloads
-          won't have the version annotation and upgrades stall.
-        - `daemonSetProgressing()` / `deploymentProgressing()` in ovn_kubernetes.go
-          drive upgrade status. Must handle zero DesiredNumberScheduled, rollout-hung
-          annotation, generation mismatch. Bugs here block cluster upgrades.
+        See AGENTS.md § "Render pipeline". Every render function must set
+        `data.Data["ReleaseVersion"]` or upgrades stall. Rollout-progress checks
+        (`daemonSetProgressing()`/`deploymentProgressing()`) must handle
+        zero-replica/zero-node HyperShift edge cases.
```
danwinship: It seems like extremely location-specific instructions like this would work better as comments in the code, where humans are more likely to see them too?

AGENTS.md
```diff
@@ -0,0 +1,195 @@
+# AGENTS.md — Cluster Network Operator
+
+Project context for AI coding agents. See [agents.md](https://agents.md/) for the open standard.
```
danwinship: (Is this going to cause agents to follow that link and suck https://agents.md/ into their context?)

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+
+## Project Overview
+
+The Cluster Network Operator (CNO) installs, configures, and upgrades the lifecycle of
+networking components on an OpenShift cluster. It follows the
+[Controller pattern](https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg#hdr-Controller):
+it watches `Network.operator.openshift.io/v1` and reconciles rendered manifests to match it.
```
danwinship: and `Network.config.openshift.io/v1`

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+make golangci-lint          # Lint only
+make verify-update-codegen  # Regenerate deepcopy/clientset code and diff-check
+```
+
+## Key Conventions
+
+- **Commit/PR titles**: imperative mood, under 72 chars, prefixed by affected component when
+  scoped (e.g. `ovn-kubernetes: Fix EgressIP race condition`). Include the Jira ref
+  (`OCPBUGS-12345`) in the title or description for bug fixes.
+- **DCO**: commits must be signed off (`git commit -s`).
```
danwinship: OCP does not generally use DCOs. That's an ovn-kubernetes thing.

Unless this is saying that _agents_ need to do a sign-off, but in that case you'd need clarification here to make sure that coderabbit doesn't try to require it for humans too.

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+- **Commits are logical units**: one self-contained change per commit; body explains *why*,
+  not a narration of the diff. No "address review comments" commits — squash into the
+  relevant commit. No merge commits — rebase onto target instead.
```
danwinship, line 59: can you call this out specifically in the coderabbit config?

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+## Go Code Quality (applies to code you write here)
+
+- Logging: klog only, never `fmt.Print*`/`log.*`.
+- Wrap errors with context (`fmt.Errorf`): `%w` when callers need `errors.Is/As`, `%v` when
+  the underlying error is an implementation detail. Don't return bare `err` from functions
+  doing meaningful work (thin wrappers/delegation are fine bare).
+  Watch for `err :=` in nested blocks shadowing an outer `err`.
+  Always use explicit `time.Duration` units (`10*time.Second`, never bare `10`).
+- **Dual-stack always**: never handle IPv4 only where IPv6 also applies (e.g. missing
+  `net.JoinHostPort` for IPv6 literals).
```
danwinship:
```suggestion
- **Dual-stack always**: never handle IPv4 only where IPv6 also applies (e.g. joining
  an IP address and port manually rather than using `net.JoinHostPort`).
```

(especially, it's not about "IPv6 literals", it's more often about variables that might contain IPv6 addresses)

AGENTS.md
```diff
@@ -0,0 +1,195 @@
...
+  Guard shared map/slice state accessed from multiple goroutines.
```
danwinship:
```suggestion
- Guard shared map/slice state accessed from multiple goroutines.
```

danwinship (commented):
Yeah, this basically looks good

AGENTS.md
kyrtapz: Removed it to be safe.

.coderabbit.yaml
kyrtapz: Moved the daemonSetProgressing/deploymentProgressing to code comments.

PR #3151, opened 2026-09-06
OCPBUGS-120940: quote YAML string fields in iptables-alerter Event

Devdatta-123:
## Summary
- Quote name/namespace interpolations in the iptables-alerter Event YAML so YAML 1.1 boolean keywords (`yes`, `no`, `true`, `false`, `on`, `off`) are treated as strings
- Prevents `kubectl create` from failing with `cannot unmarshal bool into Go struct field ObjectMeta.metadata.namespace of type string`, which CrashLoopBackOffs the DaemonSet and can leave cluster operator `network` Progressing
- Add a render unit test asserting those Event fields are quoted

Fixes: https://redhat.atlassian.net/browse/OCPBUGS-120940

## Test plan
- [ ] Unit test `TestRenderIPTablesAlerterQuotesYAMLStringFields` passes
- [ ] Create a namespace named `yes`, run a pod in it that installs legacy iptables rules, and confirm iptables-alerter logs the Event instead of crashing
- [ ] Confirm `kubectl get events -n yes` shows `IPTablesUsageObserved`
- [ ] Confirm the `network` ClusterOperator does not stay Progressing due to iptables-alerter CrashLoopBackOff

## Special notes for your reviewer
This change was prepared with AI assistance (Cursor). I reviewed the iptables-alerter script and the YAML 1.1 unmarshalling failure before opening this PR.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/network/iptables_alerter_test.go
+++ b/pkg/network/iptables_alerter_test.go
@@ -1,0 +1,40 @@
+package network
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	operv1 "github.com/openshift/api/operator/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestRenderIPTablesAlerterQuotesYAMLStringFields(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	bootstrapResult := fakeBootstrapResult()
+	bootstrapResult.IPTablesAlerter.Enabled = true
+
+	objs, err := renderIPTablesAlerter(&operv1.NetworkSpec{}, bootstrapResult, manifestDir)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	var scriptCM *uns.Unstructured
+	for _, obj := range objs {
+		if obj.GetKind() == "ConfigMap" && obj.GetName() == "iptables-alerter-script" {
+			scriptCM = obj
+			break
+		}
+	}
+	g.Expect(scriptCM).NotTo(BeNil())
+
+	script, found, err := uns.NestedString(scriptCM.Object, "data", "iptables-alerter.sh")
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(found).To(BeTrue())
+
+	// Unquoted YAML 1.1 booleans such as yes/no/true/false/on/off unmarshal as
+	// bool and crash kubectl when metadata.namespace/name must be strings.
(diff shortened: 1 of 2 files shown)


pkg/network/iptables_alerter_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	operv1 "github.com/openshift/api/operator/v1"
+	uns "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func TestRenderIPTablesAlerterQuotesYAMLStringFields(t *testing.T) {
```
danwinship, line 11: This is not doing anything useful. It's just validating "`bindata/network/iptables-alerter/002-script.yaml` contains the strings that we put into `bindata/network/iptables-alerter/002-script.yaml`". Notably, if someone added a new field to the event and failed to quote it, this test would not catch that.

It should probably just be removed. (`iptables-alerter` will be going away soon anyway.)

Devdatta-123: Agreed. That test only checked that the rendered ConfigMap still contained the quoted strings we already put in `002-script.yaml`, so it would not catch a newly added unquoted Event field.

Removed it in 169347d06.

pkg/network/iptables_alerter_test.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	g.Expect(scriptCM).NotTo(BeNil())
+
+	script, found, err := uns.NestedString(scriptCM.Object, "data", "iptables-alerter.sh")
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(found).To(BeTrue())
+
+	// Unquoted YAML 1.1 booleans such as yes/no/true/false/on/off unmarshal as
+	// bool and crash kubectl when metadata.namespace/name must be strings.
+	g.Expect(script).To(ContainSubstring(`namespace: "${pod_namespace}"`))
```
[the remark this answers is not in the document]
Devdatta-123: Dropping this test instead, per Dan's review: it only re-checked strings already present in the bindata script, so tightening the assertions would not make it more useful.

danwinship:
Please squash the two commits together. We don't want to merge a wrong commit followed by a fixup.

Devdatta-123:
Squashed the two commits into one: 621cbe5bd.

PR #3038, opened 2026-06-25
CORENET-7125: iptables to nftables

bpickard22:
Migrate from iptables to an nftables implementation

Migrates the NOTRACK rules to nftables, replaces Azure icmp drop rules, and added an nftables config for kubeproxy to allow for use of proxy-mode: nftables

assisted by: Cluade Opus 4.6

## Summary by CodeRabbit

* **New Features**
  * Migrated node networking helpers from iptables to nftables for Geneve UDP notrack handling and Azure-specific ICMP “fragmentation-needed” drops.
  * Enhanced kube-proxy configuration with additional nftables-related masquerade and sync timing options, plus improved conntrack TCP settings and node-port address override parsing.
* **Bug Fixes**
  * Improved nftables rule initialization with best-effort cleanup of legacy iptables rules to reduce stale configuration issues.
* **Tests**
  * Updated kube-proxy configuration generation tests to match current nftables output and defaults.

PR #3152, opened 2026-09-07
OCPBUGS-99652: prevent duplicate NOTRACK iptables rules on container restart

ssonigra:
## What does this PR do?
Prevents duplicate NOTRACK iptables rules from being added to the `raw` table on every ovnkube-controller container restart by checking if each rule already exists before appending it.

## Why is this important?
Without this fix, every ovnkube-controller restart blindly appends duplicate rules for the Geneve and hybrid overlay VXLAN ports. Because NOTRACK is a non-terminating target, all duplicates are evaluated sequentially, inflating packet counters and wasting CPU cycles.

## Implementation
The fix adds `iptables -C` (check) calls before each `iptables -A` (append) to ensure rules are only added when they don't already exist.

> **Note:** On `master` / OCP 5.x, this bug was already resolved as a side effect of the nftables migration (PR #3038), which uses `nft flush table` before re-adding rules. This fix targets the `release-4.x` branches that still use iptables.

## Jira
https://redhat.atlassian.net/browse/OCPBUGS-99652

## Related PRs
- #3093 (same fix for release-4.23)

## Test plan
- [x] `make build` passes
- [x] `make test-unit` passes
- [ ] Manual: restart ovnkube-node pod multiple times, verify `iptables -t raw -L PREROUTING -n -v --line-numbers` shows no duplicate NOTRACK rules

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stleerh:
Sorry, I accidentally changed the title. I changed it back.

PR #3163, opened 2026-09-10
NO-JIRA: fix(ovn-kubernetes): disable local DB probes

kyrtapz:
OVN Interconnect colocates NBDB, SBDB, northd, and ovn-controller on each node and connects them through Unix sockets. OVS detects peer closure directly and marks these transports as not requiring probes.

The configured intervals were carried over from the former TCP/TLS database topology. They can cause unnecessary disconnects when a healthy local daemon is busy and slow to reply.

Disable the probes and remove the obsolete CNO configuration knobs.

Assisted-by: Codex


dceara:
This makes sense to me, CC @almusil as we were also talking about this yesterday.

/lgtm

kyrtapz:
:disappointed: https://redhat-internal.slack.com/archives/CB95J6R4N/p1789042789395829

PR #3164, opened 2026-09-15
CORENET-7399: Expand golangci-lint configuration with comprehensive linter set

tpantelis:
Enabled most of the available linters covering code quality, bugs, performance, and style issues. Each linter is documented with its purpose and disabled linters are documented with a reason for exclusion (e.g., high noise-to-signal ratio, overly pedantic rules, or inapplicable to this codebase).
    
Also added exclusion rules to suppress false positives in specific contexts, particularly for test files and subjective style preferences.

Current violations were addressed in individual commits.


pkg/network/ovn_kubernetes.go
```diff
@@ -221,7 +221,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 				data.Data["RoutableMTU"] = conf.Migration.MTU.Network.From
 			}
 
-			// c.MTU is used to set the applied network configuration MTU
+			// c.MTU is used to set the applied network MTU configuration
```
danwinship: 1. The commit message says "Remove duplicate words in code comments", which this isn't?
2. I think the original text is closer to correct. It's talking about the "applied configuration", and there's no "MTU configuration", just an "MTU".

You could just remove this line though; the following 4 lines explain things suffiently

pkg/network/ovn_kubernetes.go
```diff
@@ -663,7 +663,7 @@ func renderOVNKubernetes(conf *operv1.NetworkSpec, bootstrapResult *bootstrap.Bo
 
 // GetIPsecMode return the ipsec mode accounting for upgrade scenarios
 // Find the IPsec mode from Ipsec.config
-// Ipsec.config == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
+// IpsecConfig == nil (bw compatibility) || ipsecConfig == Off ==> ipsec is disabled
 // ipsecConfig.mode == "" (bw compatibility) || ipsec.Config == Full ==> ipsec is enabled for NS and EW
 // ipsecConfig.mode == External ==> ipsec is enabled for NS only
```
danwinship: (again not "duplicate words")
This fixes "`Ipsec.config`" in one line but leaves it unchanged in the previous line, and "fixes" it to something wrong anyway. Should probably be `ipsecConfig` everywhere, as in the last 2 lines.

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -174,8 +175,10 @@ func (c *connectionChecker) getTCPConnectLatency(ctx context.Context, address st
 
 // isDNSError returns true if the cause of the net operation error is a DNS error
 func isDNSError(err error) bool {
-	if opErr, ok := err.(*net.OpError); ok {
-		if _, ok := opErr.Err.(*net.DNSError); ok {
+	var opErr *net.OpError
+	if errors.As(err, &opErr) {
+		var dnsErr *net.DNSError
+		if errors.As(opErr.Err, &dnsErr) {
```
danwinship: You don't need to do the intermediate unwrapping.

```suggestion
	if errors.Is(err, &net.DNSError{}) {
```

pkg/controller/observability/observability_controller.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	goerrors "errors"
```
danwinship: It's better to rename `"k8s.io/apimachinery/pkg/api/errors"` to `apierrors`

pkg/network/ovn_kubernetes_dpu_host_test.go
```diff
@@ -549,13 +549,11 @@ func TestDpuHostModeResourceCount(t *testing.T) {
...
-						_, found := ovnkubeController.Resources.Requests[resourceName]
-						g.Expect(found).To(BeFalse(), "resource request should not be set")
-						_, found = ovnkubeController.Resources.Limits[resourceName]
-						g.Expect(found).To(BeFalse(), "resource limit should not be set")
-					}
+				} else if tc.mgmtPortResourceName != "" {
+					_, found := ovnkubeController.Resources.Requests[resourceName]
+					g.Expect(found).To(BeFalse(), "resource request should not be set")
+					_, found = ovnkubeController.Resources.Limits[resourceName]
+					g.Expect(found).To(BeFalse(), "resource limit should not be set")
```
danwinship, line 555: I disagree with this rewrite in theory, but also `tc.mgmtPortResourceName` is always non-empty anyway, so just remove the ` if tc.mgmtPortResourceName != ""` part and it's good

pkg/network/render.go
```diff
@@ -881,8 +881,7 @@ func renderAdditionalRoutingCapabilities(conf *operv1.NetworkSpec, bootstrapResu
 	}
 	var out []*uns.Unstructured
 	for _, provider := range conf.AdditionalRoutingCapabilities.Providers {
-		switch provider {
-		case operv1.RoutingCapabilitiesProviderFRR:
+		if provider == operv1.RoutingCapabilitiesProviderFRR {
```
danwinship: should be 

```
if provider != operv1.RoutingCapabilitiesProviderFRR {
        continue
}
```

and then unindent the rest of the loop

pkg/controller/clusterconfig/clusterconfig_controller.go
```diff
@@ -27,7 +27,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// and Start it when the Manager is Started.
+// Add to the given Manger and start it when the Manager is started.
 func Add(mgr manager.Manager, status *statusmanager.StatusManager, c cnoclient.Client, featureGates featuregates.FeatureGate) error {
```
danwinship, line 31: the comment appears to have been missing the first half of the sentence ever since it was added...

Also, typo: "Manger"

pkg/controller/connectivitycheck/connectivity_check_controller.go
```diff
@@ -43,7 +43,7 @@ type NetworkConnectivityCheckController interface {
 	connectivitycheckcontroller.ConnectivityCheckController
 }
 
-// NetworkConnectivyCheckController consumes a series of clients, informers and a recorders.
+// NewNetworkConnectivityCheckController consumes a series of clients, informers and a recorders.
```
danwinship: The comment seems to be about the type, not its New function. Adding a blank line between the comment and the function should appease the linter.

pkg/controller/egress_router/egress_router_controller.go
```diff
@@ -8,17 +8,16 @@ import (
...
 
 	"github.com/openshift/cluster-network-operator/pkg/apply"
 	cnoclient "github.com/openshift/cluster-network-operator/pkg/client"
 	"github.com/openshift/cluster-network-operator/pkg/render"
 	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
 
-	"path/filepath"
-	"reflect"
-	"strings"
-	"time"
```
danwinship, line 20: seems to be in the wrong commit

pkg/names/names.go
```diff
@@ -194,32 +194,33 @@ func TrustedCABundleConfigMap() types.NamespacedName {
 }
 
 // constants for namespace and custom resource names
-// namespace in which ingress controller objects are created
+
+// IngressControllerNamespace in which ingress controller objects are created
```
danwinship: "... is the namespace in which ..."

pkg/names/names.go
```diff
@@ -194,32 +194,33 @@ func TrustedCABundleConfigMap() types.NamespacedName {
...
 const IngressControllerNamespace = "openshift-ingress-operator"
 
-// namespace representing host network traffic
+// HostNetworkNamespace representing host network traffic
```
danwinship: "... is the namespace representing ..."

pkg/names/names.go
```diff
@@ -194,32 +194,33 @@ func TrustedCABundleConfigMap() types.NamespacedName {
...
 // this is also the namespace where to set the ingress label
 const HostNetworkNamespace = "openshift-host-network"
 
-// label for ingress policy group
+// PolicyGroupLabelIngress label for ingress policy group
```
danwinship: etc, etc

pkg/cmd/checkendpoints/controller/connection_checker.go
```diff
@@ -157,6 +157,7 @@ func (c *connectionChecker) getTCPConnectLatency(ctx context.Context, address st
 
 	// perform tls handshake to avoid spamming the logs of tls endpoints
 	host, _, _ := net.SplitHostPort(address)
+	//nolint:gosec // G402: InsecureSkipVerify is intentional - this is a connectivity checker, not validating certs
```
danwinship, line 160: `gosec` tends to be a bit opinionated about the wrong things and it might be simpler to just not enable it rather than suppressing it everywhere...

pkg/controller/signer/signer.go
```diff
@@ -17,7 +19,15 @@ import (
 func newCertificateTemplate(certReq *x509.CertificateRequest, certDuration time.Duration) *x509.Certificate {
 	// Like in openshift/library-go/pkg/crypto/crypto.go, we will generate a random
 	// serial number
-	serialNumber := mathrand.New(mathrand.NewSource(time.Now().UTC().UnixNano())).Int63()
+	var serialNumber int64
+	serialNumberBig, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
```
danwinship: This does not need cryptographic randomness, and the library-go code it was copied from was intentionally not using `"crypto/rand"`

pkg/controller/proxyconfig/validation.go
```diff
@@ -36,7 +36,7 @@ const (
 )
 
 // ValidateProxyConfig ensures that httpProxy, httpsProxy and
-// noProxy fields of proxyConfig are valid.
+// noProxy fields of proxyConfig are valid.
```
danwinship: stray added space here

pkg/controller/configmap_ca_injector/controller.go
```diff
@@ -139,7 +139,7 @@ func (r *ReconcileConfigMapInjector) Reconcile(ctx context.Context, request reco
 	if err != nil {
 		log.Println(err)
 		r.status.SetDegraded(ctx, statusmanager.InjectorConfig, "InvalidInjectorConfig",
-			fmt.Sprintf("Failed to validate trusted CA certificates in %s", trustedCAbundleConfigMap.Name))
+			"Failed to validate trusted CA certificates in "+trustedCAbundleConfigMap.Name)
```
danwinship: I don't like this linter. Please disable it and remove this commit. "Performance" doesn't matter in any of these cases, and it's simpler to just consistently use `fmt.Sprintf` rather than worrying about when concatenation would be faster.

pkg/util/k8s/unstructured.go
```diff
@@ -31,7 +32,7 @@ func CalculateHash(obj any) (string, error) {
 		return "", err
 	}
 	configSum := sha256.Sum256(configStr)
-	return fmt.Sprintf("%x", configSum), nil
+	return hex.EncodeToString(configSum[:]), nil
```
danwinship: No, I don't know what that function does without looking it up

pkg/controller/proxyconfig/controller.go
```diff
@@ -387,7 +387,7 @@ func (r *ReconcileProxyConfig) mergeTrustBundlesToConfigMap(additionalData, syst
 		return nil, fmt.Errorf("failed to merge ca bundles, system trust bundle is empty")
 	}
 
-	combinedTrustData := []byte{}
+	combinedTrustData := make([]byte, 0, len(additionalData)+1+len(systemData))
```
danwinship: too much math, though if the alternative is needed to add a nolint comment, then keep this

pkg/controller/statusmanager/status_manager_test.go
```diff
@@ -486,7 +486,7 @@ func TestStatusManagerSetFromIPsecConfigs(t *testing.T) {
 	setOC(t, client, no)
 	setCO(t, client, "testing")
 
-	mcPools := []mcfgv1.MachineConfigPool{}
+	mcPools := make([]mcfgv1.MachineConfigPool, 0, 2)
```
danwinship: I don't really like this. You shouldn't have to read the entire rest of the function to confirm that the initial size is correct.

pkg/network/render.go
```diff
@@ -225,6 +225,7 @@ func DeprecatedCanonicalize(conf *operv1.NetworkSpec) {
 // Validate checks that the supplied configuration is reasonable.
 // This should be called after Canonicalize
 func Validate(conf *operv1.NetworkSpec) error {
+	//nolint:prealloc // We don't expect errors normally so no need to preallocate.
 	errs := []error{}
```
danwinship, line 227: Yeah, I'm thinking I don't like this linter. It's nice that it found the weird recurring `objs = append(objs, manifests...)` nonsense, but other than that, I don't think any of the fixes are necessarily much _better_.

(Keep the fixes I didn't comment on, but drop the linter.)

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
```diff
@@ -66,7 +66,7 @@ func NewPodNetworkConnectivityCheckController(podName, podNamespace string,
 
 // Sync ensures that the status updaters for each PodNetworkConnectivityCheck is started
 // and then performs each check.
-func (c *controller) Sync(ctx context.Context, syncContext factory.SyncContext) error {
+func (c *controller) Sync(ctx context.Context, _ factory.SyncContext) error {
```
danwinship, line 69: If we are implementing an interface, I feel like naming the variables is useful even if we don't _currently_ use them.

So I vote against this linter (at least as configured?)

pkg/controller/infrastructureconfig/sync_specstatus.go
```diff
@@ -2,17 +2,18 @@ package infrastructureconfig
 
 import (
 	"fmt"
+	"log"
+
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-network-operator/pkg/util/ip"
-	"log"
```
danwinship, line 7: in the wrong commit?

pkg/controller/infrastructureconfig/sync_specstatus.go
```diff
@@ -2,17 +2,18 @@ package infrastructureconfig
...
 )
 
 func (*synchronizer) SpecStatusSynchronize(infraConfig *configv1.Infrastructure) (*configv1.Infrastructure, error) {
 	updatedInfraConfig := infraConfig.DeepCopy()
 
 	var (
-		statusApiVips, statusIngressVips           *[]string
-		specApiVips, specIngressVips               *[]configv1.IP
+		statusAPIVips, statusIngressVips           *[]string
+		specAPIVips, specIngressVips               *[]configv1.IP
```
danwinship: "VIPs" not "Vips"

pkg/names/names.go
```diff
@@ -4,28 +4,28 @@ import "k8s.io/apimachinery/pkg/types"
 
 // some names
 
-// OPERATOR_CONFIG is the name of the CRD that defines the complete
+// OperatorConfig is the name of the CRD that defines the complete
```
danwinship, line 7: (I think we talked about fixing these back in the 4.3 timeframe, but didn't do it, to avoid backport conflicts :joy:.)

pkg/names/names.go
```diff
@@ -107,34 +107,34 @@ const ClusterNameAnnotation = "network.operator.openshift.io/cluster-name"
...
 const RelatedClusterObjectsAnnotation = "network.operator.openshift.io/relatedClusterObjects"
 
-// MULTUS_VALIDATING_WEBHOOK is the name of the ValidatingWebhookConfiguration for multus-admission-controller
+// MultusValidatingWebhook is the name of the ValidatingWebhookConfiguration for multus-admission-controller
 // that is used in multus admission controller deployment
-const MULTUS_VALIDATING_WEBHOOK = "multus.openshift.io"
+const MultusValidatingWebhook = "multus.openshift.io"
 
-// ADDL_TRUST_BUNDLE_CONFIGMAP_NS is the namespace for one or more
+// AddlTrustBundleConfigmapNs is the namespace for one or more
```
danwinship: "NS"
Also, just go with "Additional". A few more letters won't kill anyone. So `AdditionalTrustBundleConfigMapNS`. Oh, also "ConfigMap" because that's the properly-capitalized form of that type name.

pkg/names/names.go
```diff
@@ -107,34 +107,34 @@ const ClusterNameAnnotation = "network.operator.openshift.io/cluster-name"
...
 // ConfigMaps that contain user provided trusted CA bundles.
-const ADDL_TRUST_BUNDLE_CONFIGMAP_NS = "openshift-config"
+const AddlTrustBundleConfigmapNs = "openshift-config"
 
-// TRUSTED_CA_BUNDLE_CONFIGMAP_KEY is the name of the data key containing
+// TrustedCaBundleConfigmapKey is the name of the data key containing
```
danwinship: fix "CA" and "ConfigMap", likewise with other names below. ALso, more "NS" too.

pkg/network/ovn_kubernetes.go
```diff
@@ -49,36 +49,36 @@ import (
...
-const OVN_NODE_SELECTOR_DEFAULT_DPU = "network.operator.openshift.io/dpu="
-const OVN_NODE_SELECTOR_DEFAULT_SMART_NIC = "network.operator.openshift.io/smart-nic="
-const OVN_NODE_IDENTITY_CERT_DURATION = "24h"
-
-// DPU_NODE_LEASE_RENEW_INTERVAL_DEFAULT and DPU_NODE_LEASE_DURATION_DEFAULT are the default DPU health check lease configuration.
+const ClusterConfigName = "cluster-config-v1"
+const ClusterConfigNamespace = "kube-system"
+const OvnCertCn = "ovn"
+const OvnLocalGwMode = "local"
+const OvnSharedGwMode = "shared"
```
danwinship: OK, golang's rule is that acronyms are never TitleCase. They're either all lowercase or all caps. So:
`OVNCertCN`, `OVNLocalGWMode`, `OVNSharedGWMode`, `DPU`, `NIC`, `EgressIP`, etc

pkg/util/util.go
```diff
@@ -12,18 +12,17 @@ import (
...
-const OVN_NAMESPACE = "openshift-ovn-kubernetes"
-const OVN_CONTROL_PLANE = "ovnkube-control-plane"
-const OVN_NODE = "ovnkube-node"
-const OVN_CONTROLLER = "ovnkube-controller"
-const MTU_CM_NAMESPACE = "openshift-network-operator"
-const MTU_CM_NAME = "mtu"
-const OVN_NBDB = "nbdb"
+const OvnNamespace = "openshift-ovn-kubernetes"
+const OvnControlPlane = "ovnkube-control-plane"
+const OvnNode = "ovnkube-node"
```
danwinship: OVN

pkg/util/util.go
```diff
@@ -12,18 +12,17 @@ import (
...
+const MtuCmNamespace = "openshift-network-operator"
+const MtuCmName = "mtu"
```
danwinship: "MTU" and "CM". Or better yet, "ConfigMap"

pkg/util/util.go
```diff
@@ -12,18 +12,17 @@ import (
...
+const OvnNbdb = "nbdb"
```
danwinship: hrm. Well, that's "OVNNBDB"... maybe "OVNNBDBName"

cmd/cluster-network-operator/mtu_probe.go
```diff
@@ -77,10 +77,10 @@ func newMTUProberCommand() *cobra.Command {
 			if err == nil {
 				fmt.Println("Successfully set config map")
 				break
-			} else {
-				fmt.Printf("Failed to write ConfigMap: %v", err)
-				time.Sleep(10 * time.Second)
 			}
+
+			fmt.Printf("Failed to write ConfigMap: %v", err)
+			time.Sleep(10 * time.Second)
```
danwinship: I don't agree that these rewrites improve the code, though I don't think any of them make it worse either. But if you end up dropping `revive` entirely (or can configure it to avoid these rewrites specifically), then I'd drop this commit, to save on unnecessary backport conflicts.

pkg/client/list_pager.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 // ListAllOfSpecifiedType is for use with dynamic client
-func ListAllOfSpecifiedType(resourceType schema.GroupVersionResource, ctx context.Context, client Client) ([]*uns.Unstructured, error) {
+func ListAllOfSpecifiedType(ctx context.Context, resourceType schema.GroupVersionResource, client Client) ([]*uns.Unstructured, error) {
```
danwinship: The fact that no other changes were needed must mean that nothing is using this function...

Apparently it was previously used by the migration code, but it no longer has any callers and could just be removed.

pkg/controller/signer/signer_test.go
```diff
@@ -178,6 +178,7 @@ func TestDecodePrivateKey(t *testing.T) {
 		{
 			name: "RSA PKCS1 private key",
 			pemData: func(t *testing.T) []byte {
+				t.Helper()
```
danwinship: This seems wrong; if a test fails inside a helper defined _inside a test case_, then presumably we want to see that more specific failure location.

pkg/controller/proxyconfig/controller_test.go
```diff
@@ -183,6 +183,7 @@ func proxyWithSpec(spec configv1.ProxySpec) *configv1.Proxy {
 	}
 }
 
+//nolint:unparam // Ignore cidr always receives ...
 func networkWithClusterCIDR(cidr string) *configv1.Network {
```
danwinship, line 186: I might disable this linter for all `*_test.go` files...

danwinship:
(overall: yay!)

cmd/cluster-network-operator/mtu_probe.go
tpantelis: It saves unnecessary indentation which is purely style preference. `revive` has some opinionated, subjective style rules like this one (`superfluous-else`) but it also has a lot of useful rules as well that flagged other issues. I think we should keep `revive` but I disabled `superfluous-else` (and `indent-error-flow`).

pkg/client/list_pager.go
tpantelis: yeah it's not used any more. I had left it just in case it might have use. Actually none of the functions are used so I removed the file.

pkg/cmd/checkendpoints/controller/connection_checker.go
tpantelis: The intermediate unwrapping is unnecessary but we need to use `As` since `Is` uses value equality.

pkg/cmd/checkendpoints/controller/pod_network_connectivity_check_controller.go
tpantelis: This is from the `unused-parameter` rule of the `revive` linter. In this case the argument for keeping the name is valid. We could disable the `unused-parameter` rule but then it won't catch unused params in non-interface functions, which is useful. Unfortunately there's no way to distinguish between the two. I'm inclined to keep it - if there's a case where we now need to use a previously unused interface parameter then one can consult the interface docs for semantics (would probably have to do that anyway).

pkg/controller/configmap_ca_injector/controller.go
tpantelis: yeah fair enough. I disabled the `perfsprint` linter.

pkg/util/k8s/unstructured.go
tpantelis: I disabled the `perfsprint` linter.

pkg/controller/connectivitycheck/connectivity_check_controller.go
tpantelis: you're right - these comments should actually be above line 42.

pkg/controller/egress_router/egress_router_controller.go
tpantelis: GoLand moved them to the top on save to group them with the other standard library imports. Not intended for this commit but it's where they belong so I moved the changes to a separate commit.

pkg/controller/infrastructureconfig/sync_specstatus.go
tpantelis: Same as above re: GoLand - moved to a separate commit.

pkg/controller/observability/observability_controller.go
tpantelis: I agree and changed it. Originally I was trying to minimize the changes but it's not that much.

pkg/network/render.go
tpantelis: Fair enough - I disabled it. It's useful but also premature optimization and mostly negligible in this code base. Hence the linter's message "Consider preallocating ...".

pkg/controller/proxyconfig/controller.go
tpantelis: Removed the change.

pkg/controller/statusmanager/status_manager_test.go
tpantelis: Removed the change.

pkg/controller/signer/signer.go
tpantelis: ok - removed the changes and suppressed it.

pkg/controller/signer/signer_test.go
tpantelis: I was on the fence about this linter. I'll ignore it here but I'm also OK with disabling the linter.

pkg/network/ovn_kubernetes.go
tpantelis: Removed the line.

pkg/network/ovn_kubernetes.go
tpantelis: This is flagged by the `dupword` linter b/c `Ipsec.config` is repeated from the end of the previous line. Actually `Ipsec.config` isn't even correct. I changed it to:

```Go
// Find the IPsec mode from the IPsecConfig in the OVNKubernetesConfig.
// if IPsecConfig == nil (bw compatibility) || IPsecConfig ...
// if IPsecConfig.mode == "" (bw compatibility) || IPsecConfig ...
// if IPsecConfig.mode == External ==> ...
```

pkg/network/ovn_kubernetes_dpu_host_test.go
tpantelis: The rewrite saves indentation. There is one case on line 490 where `tc.mgmtPortResourceName` is empty.

pkg/cmd/checkendpoints/controller/connection_checker.go
tpantelis: This is a valid use of `InsecureSkipVerify: true`, but `gosec` is correctly flagging it - the setting is potentially security-sensitive and I think should require explicit justification.

`gosec` is opinionated about some things but does have a lot of useful rules so I think we should keep it. The biggest pain point in this code base (and others) is the "_G115: integer overflow conversion ..._" rule due to APIs using `uintX` types to represent  mask sizes, ports etc where we know overflow won't occur. I'm inclined to disable G115. Outside of that, there's only 3 other explicit suppressions which is low signal-to-noise.

I went ahead and removed the commit that suppressed G115 in code and disabled it in the _.golangci.yaml_.  I can disable `gosec` altogther if that's what we decide.

pkg/network/ovn_kubernetes_dpu_host_test.go
danwinship: oh, missed that.

So, I prefer the original (though not enough to want to `//nolint` it), because it expresses the actual logic better. The rewritten version is not especially explicit about the fact that we only care about `tc.mgmtPortResourceName` when `tc.expectResource` is `false`.

pkg/controller/egress_router/egress_router_controller.go
danwinship: oh... I assumed you had moved them to make some other linter happy... it might be good to fix the import ordering to be consistent everywhere, while we're totally breaking our cherry-pick-ability. I think golangci-lint has a module that lets you do that? (We want, generally, "core imports, blank line, non-core imports, blank line, imports from within cluster-network-operator", probably with "`.`" imports of ginkgo/gomega handled as a 4th group but I'm not sure what the most consistent location for them is.)

pkg/cmd/checkendpoints/controller/connection_checker.go
danwinship: > I'm inclined to disable G115. Outside of that, there's only 3 other explicit suppressions which is low signal-to-noise.

but there's only 1 actual _fix_ from `gosec` isn't there? (The timeouts in network-check-target.) So that's pretty terrible signal-to-noise, and presumably it will continue to be terrible in the future.

pkg/controller/signer/signer_test.go
danwinship: I like how it pointed out when _external_ helper functions were missing `t.Helper()`. But it seems like a bug to be flagging internal helpers.

pkg/controller/egress_router/egress_router_controller.go
tpantelis: `golangci-lint` has support for `gci` and `goimports` that do import ordering. But these are not linters and do not run during `golangci-lint run`. They're formatters that can be enabled in the _.golangci.yaml_ and are invoked via `golangci-lint fmt` and implement fixes. I ran it with `gci` and it orders them by standard lib then the rest and sorts each group (and removes blank lines). The changes it made were for the latter so the code base already has the standard imports listed first (with the exception of the 2 that GoLand auto-fixed). So I'd think by now all devs are using GoLand or a similar IDE that handles organizing imports. I'd prefer to possibly fix import ordering in a future PR if we decide to.

pkg/cmd/checkendpoints/controller/connection_checker.go
tpantelis: But the other 2 are legitimate issues that I wouldn't consider noise and I think are worth justifying via suppression. Plus they seem like isolated edge cases. If you really feel strongly enough to disable all of `gosec` then OK but I'd prefer to give it a chance first with the exclusions we have and see what future changes bring. We can always disable later if the signal-to-noise ratio becomes unreasonable.

pkg/names/names.go
```diff
@@ -107,34 +107,34 @@ const ClusterNameAnnotation = "network.operator.openshift.io/cluster-name"
...
 const RelatedClusterObjectsAnnotation = "network.operator.openshift.io/relatedClusterObjects"
 
-// MULTUS_VALIDATING_WEBHOOK is the name of the ValidatingWebhookConfiguration for multus-admission-controller
+// MultusValidatingWebhook is the name of the ValidatingWebhookConfiguration for multus-admission-controller
 // that is used in multus admission controller deployment
-const MULTUS_VALIDATING_WEBHOOK = "multus.openshift.io"
+const MultusValidatingWebhook = "multus.openshift.io"
 
-// ADDL_TRUST_BUNDLE_CONFIGMAP_NS is the namespace for one or more
+// AdditionalTrustBundleConfigMapNS is the namespace for one or more
```
danwinship, line 114: The "NS" is uglier now that the rest of the name is titlecase... can you change them all to "Namespace"?

pkg/controller/signer/signer_test.go
```diff
@@ -170,6 +170,7 @@ func TestSigner_reconciler_withInvalidUserName(t *testing.T) {
 }
 
 func TestDecodePrivateKey(t *testing.T) {
+	//nolint:thelper // Helper functions are inside test cases
```
danwinship: Comment could be clearer.

```suggestion
	//nolint:thelper // Don't suggest adding `t.Helper()` to single-use helper functions inside the test case.
```

danwinship, line 173: I had only mentioned this in one place, but there are _lots_ of places where I feel like it was suggesting this incorrectly. Like, every call to `testTLSArgRendering` (and `testTLSArgRendering` itself). And I don't like marking each one `//nolint`, because then the next time someone adds something using `testTLSArgRendering`, the linter will tell them to use `t.Helper()`, and the reviewer may not catch the fact that it's wrong.

So we should drop this linter, because it has too many false positives. (And, also, note that the failure mode of someone failing to use `t.Helper()` where they should is just that unit test error output isn't as helpful as it could have been, and it can always be fixed later.)

(I'm fine keeping the "good" `t.Helper()` additions if you want, but I haven't gone through every suggestion to figure out which ones would actually make the error output better and which would make it worse, so you'd have to do that.)

tpantelis: Agreed - let's drop it. I forced failures and I didn't even see it print line numbers, but maybe I was missing a  flag to enable them. I'm used to Ginkgo which prints the whole stack trace (when enabled).

PR #3167, opened 2026-09-17
[OCPBUGS-123509]multus: Refresh ignored namespaces on namespace changes

ssonigra:
## What does this PR do?

Adds event-driven reconciliation for namespace changes that affect the Multus admission controller ignore list.

The CNO now:
- Watches Namespace create, update, and delete events.
- Reconciles when a Namespace enters or leaves the set matching `openshift.io/cluster-monitoring=true` and `workload.openshift.io/allowed=management`.
- Refreshes the Multus `-ignore-namespaces` argument without restarting CNO.
- Keeps the existing periodic reconciliation as a fallback.
- Stops rendering when the Namespace list cannot be read, preserving the existing configuration for that reconciliation.

## Why?

ACM namespaces can be created after the network operator starts, particularly during IBU restore. Previously, the Multus ignore list could remain stale until the CNO was restarted.

## How to verify it

Local validation passed on commit `b3533a946`:
- `go test ./pkg/network ./pkg/controller/operconfig`
- `go vet ./...`
- `go build -buildvcs=false ./...`
- `git diff --check`

Automated CNO CI lanes cover unit, verify, verify-deps, lint, images, security, and upgrade/conformance testing. On the current commit, the following jobs are running and currently pending:
- `ci/prow/5.1-upgrade-from-stable-5.0-e2e-aws-ovn-upgrade` — AWS.
- `ci/prow/5.1-upgrade-from-stable-5.0-e2e-azure-ovn-upgrade` — Azure.
- `ci/prow/e2e-aws-ovn-fdp-qe`, `ci/prow/e2e-aws-ovn-hypershift-conformance`, `ci/prow/e2e-aws-ovn-serial-1of2`, `ci/prow/e2e-aws-ovn-serial-2of2`, `ci/prow/e2e-aws-ovn-upgrade`, and `ci/prow/e2e-aws-ovn-windows` — AWS.

No dedicated `test/e2e` test was added. This is a focused controller/rendering bug fix: unit tests directly verify the Namespace predicate and refreshed Multus ignore list. A dedicated E2E test would require provisioning an ACM-managed Namespace with internal platform labels and annotations and would not provide better coverage of the event predicate than these tests. Existing CNO upgrade and conformance lanes provide cluster-level regression coverage.

## Upgrade and rollback

This change has no API, storage, schema, or upgrade-hook changes. Existing periodic reconciliation remains as a fallback during upgrades. Rollback is safe by reverting this commit; no persistent data or migration cleanup is required.

## Jira

https://redhat.atlassian.net/browse/OCPBUGS-123509

Always review AI generated responses prior to use.
AI-assisted response via openshift-developer plugin


ssonigra:
@coderabbitai ignore pre-merge checks

pkg/network/multus_admission_controller.go
```diff
@@ -46,25 +55,23 @@ func getOpenshiftNamespaces(client cnoclient.Client) (string, error) {
 	}
 
 	for _, ns := range nsList.Items {
-		// add OpenShift components to ignored namespace
-		if metav1.HasAnnotation(ns.ObjectMeta, "workload.openshift.io/allowed") && ns.Annotations["workload.openshift.io/allowed"] == "management" {
+		if IsMultusAdmissionControllerIgnoredNamespace(&ns) {
```
danwinship, line 58: (`IsMultusAdmissionControllerIgnoredNamespace` also checks for `ns.Labels["openshift.io/cluster-monitoring"] == "true"`, but that has no additional effect here because `nsList` already has only the namespaces with that label. 👍 )

pkg/network/multus_admission_controller.go
```diff
@@ -46,25 +55,23 @@ func getOpenshiftNamespaces(client cnoclient.Client) (string, error) {
...
 			namespaces = append(namespaces, ns.Name)
 		}
 	}
+	sort.Strings(namespaces)
```
danwinship, line 62: What is this for?
If it's _just_ for unit test comparisons it would be better to make the unit tests comparisons non-order-dependent.

pkg/network/multus_admission_controller.go
```diff
@@ -46,25 +55,23 @@ func getOpenshiftNamespaces(client cnoclient.Client) (string, error) {
...
 	objs := []*uns.Unstructured{}
 	var err error
 
 	replicas := getMultusAdmissionControllerReplicas(bootstrapResult, hsc.Enabled)
-	if ignoredNamespaces == "" {
-		ignoredNamespaces, err = getOpenshiftNamespaces(client)
-		if err != nil {
-			klog.Warningf("failed to get openshift namespaces: %+v", err)
-		}
+	ignoredNamespaces, err := getOpenshiftNamespaces(client)
```
danwinship, line 72: you should make this a local variable now

danwinship (commented):
makes sense

pkg/network/multus_admission_controller_test.go
```diff
@@ -237,21 +237,65 @@ func TestRenderMultusAdmissionControllerGetNamespace(t *testing.T) {
...
+				},
+			},
+		},
+		metav1.CreateOptions{},
+	)
+	g.Expect(err).NotTo(HaveOccurred(), "failed to create the test namespace")
+
+	objs, err = renderMultusAdmissonControllerConfig(manifestDir, false, bootstrap, fakeClient, hsc, "", getDefaultFeatureGates())
+	g.Expect(err).NotTo(HaveOccurred(), "failed to render the refreshed Multus admission controller configuration")
+	g.Expect(findMultusWebhookExec(t, objs)).To(ContainSubstring("test-namespace"), "refreshed configuration should contain the test namespace")
```
Ultimate-etamitlU: Please start with one eligible namespace, render once, then add another and verify both; otherwise this test also passes against the pre-PR implementation and does  not prove the fix.

Ultimate-etamitlU (changes requested):

pkg/network/multus_admission_controller.go
ssonigra: I added the sort to keep the `-ignore-namespaces` argument stable. The Namespace API does not guarantee the order of results, so without sorting, the Deployment spec could change on a later reconcile even when the set of ignored namespaces has not changed. The test benefits from stable ordering too, but that is not the only reason for the sort.

pkg/network/multus_admission_controller.go
ssonigra: ignoredNamespaces is now local to the render function; I removed the package-level variable.

pkg/network/multus_admission_controller_test.go
ssonigra: this has been done , could you please cehck now

Ultimate-etamitlU: Test looks good to me with the changes.

Ultimate-etamitlU (approved):

Ultimate-etamitlU:
+1 to this PR based on review I did wrt to understanding to the bug, definitely there shouldn't be a need of restarting operator as per the issue noted in bug and aim of this code change.

pkg/network/multus_admission_controller.go
ssonigra: Hello Dan, you are Right, the label selector already filters this list. I left the check in IsMultusAdmissionControllerIgnoredNamespace because the Namespace event predicate also uses this helper, and those events aren’t pre-filtered. Keeping both checks here means rendering and event-triggered reconciliation use the same eligibility criteria.

